PR #857, opened 2014-01-21
Initial support for Zeiss .lms files

melissalinkert:
Fixes ticket https://trac.openmicroscopy.org.uk/ome/ticket/4150.

To test, verify that the dataset in data_repo/from_skyking/zeiss-lms/ looks reasonable when imported into OMERO and/or opened in ImageJ.


melissalinkert:
Not really sure why I didn't merge this back in March.  origin/develop merged in and exclude label removed.  Probably just needs one more quick sanity check before merging for real.

rleigh-dundee:
Looks OK in that it is the same as when last tested.  Still not convinced by the first series though.  I've never seen SEM data with that amount of what looks like noise with a high degree of regular structure in it.  Is it potentially 16-bit data being treated as 8-bit? (for the patterns along each scanline)?  Do we have any screenshot of what it's supposed to look like?

melissalinkert:
You're right - for some reason I thought I had tried setting the first image to uint16 previously without great results, but on further review it seems to give a better result.  39d8d94 should fix it, though we don't have screenshots or acquisition software for comparison.

rleigh-dundee:
This looks much better, thanks.  It certainly looks OK with showinf.  However, series0 won't open after import into OMERO.  For some reason it has Z=2 (rather than 1).  It has a thumbnail, but can't be viewed.

rleigh-dundee:
@jburel Any idea about the OMERO side here?  The Bio-Formats metadata looks correct, so quite why this is changed on import into OMERO is not clear.

jburel:
@rleigh-dundee: I will have to investigate

rleigh-dundee:
My fault.  That was with the import with the reader from last week's merge build.  Trying again with today's it's fine.  So looks good to merge.

melissalinkert:
@rleigh-dundee: any chance you were trying to view an image imported before 39d8d94 was in the merge build?  I would expect Z = 2 for the first series without that commit, and can't reproduce the problem when viewing the .lms file imported with the current merge build.

rleigh-dundee:
@melissalinkert Yes, looks that way.

PR #1014, opened 2014-03-31
Add initial structure for decompressing JPEG XR images (assist #8493).

bpindelski:
This PR adds a Bio-Formats module named `ome-jxr`. This module implements part of the ITU-T T.832 recommendation for "Still-image compression – JPEG XR". The main blocker encountered during work on the business logic was the lack of proper bit-level stream readers in Java (and Bio-Formats). The code implements the specification up to point 8.4.22.6. Comments have been left in places where the requirement of an enum or class was unknown before the decompression stage. Outstanding tasks can be seen in https://trac.openmicroscopy.org/ome/ticket/8493.

For discussion during code review:
- the ITU-T spec mentions global variables and codestream-level variables. I think a globally visible store of variables needed during the decompression process would make the code cleaner (reduced number of private class members),
- class hierarchy,
- test coverage and construction.

---

**WARNING: This implementation is incomplete. Please see [the comment](https://github.com/openmicroscopy/bioformats/pull/1014#issuecomment-190647919) below for the current status (March 2016)**

### DIFF
--- a/components/ome-jxr/src/ome/jxr/ifd/IFDMetadata.java
+++ b/components/ome-jxr/src/ome/jxr/ifd/IFDMetadata.java
@@ -203,7 +203,8 @@ public class IFDMetadata {
     if (colorSpace == null) {
       return null;
     } else {
-      return colorSpace.equals(1) || colorSpace.equals(0xffff) ? colorSpace : 1;
+      return colorSpace == 1 || colorSpace == 0xffff ? colorSpace
+          : (short) 0xffff;
     }
   }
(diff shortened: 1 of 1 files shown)


sbesson:
Following this initial effort, more information about the cost of trying to implement JPEG-XR can be read at http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-September/005616.html and http://blog.openmicroscopy.org/file-formats/community/2016/01/06/format-support/ .

PR #1015, opened 2014-03-31
Messy binary

qidane:
Is this an ok approach?

I think it would need OMETiffWriter.java and OMEBinaryOnlyTiffWriter.java defined from a common code (OMECoreTiffWriter.java?) that has almost all of the functions from OMETiffWriter.java. The current approach is a bit messy. (e.g. a function called closeParentWorkaroundToFix)

The other approach is putting a lot of conditionals into OMETiffWriter.java and having a function to set a flag, something like setSaveBinaryOnly(true).

Also at the moment is is saving 1 companion and 1 tiff with all the planes in it. Working on that bit now. The %z in the filename is currently being ignored.

--exclude

@melissalinkert @joshmoore

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -440,12 +440,15 @@ public final class ImageConverter {
       new Object[] {reader.getFormat(), out, format});
     long mid = System.currentTimeMillis();
 
+    // if the writer type supports stacks setup the variables for
+    //    the first and last series in the file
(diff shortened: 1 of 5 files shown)


ximenesuk:
Wow, this one is heading towards its first birthday. Anything need doing with it @qidane @melissalinkert or should we order a cake?

qidane:
Rebased onto current develop

jburel:
I will look at it post 5.1

jburel:
Closing this PR. Will spend looking at it during 5.1.3

PR #1177, opened 2014-06-24
specification: Add metadata for HDF5 embedding

rleigh-dundee:
- add Hdf5Data element for use in Pixels and Mask
- split TiffData out of ome.xsd and allow use in Mask

This is intended for comments, not merging at this time.  The intent isn't to get OME-HDF5 working or fully specified at this point, but to add the necessary bits to the schema to allow OME-XML to reference datasets (pixel data hypervolumes) in an HDF5 file.

--exclude


joshmoore:
Thanks for the proposal, @rleigh-dundee. I'm closing this for the moment, but when we ramp up on HDF5 we can re-evaluate.

PR #1309, opened 2014-09-02
Allow tile-based access of large JPEG-2000 images

melissalinkert:
--exclude

See QA 9464 and https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7568.  Tile based access largely works now without requiring an inordinate amount of memory, but this needs further work to fix tile reading beyond the coordinate (9216, 4096).


melissalinkert:
Closing for now to avoid confusion - this needs substantially more investigation than we can do right now.

PR #1381, opened 2014-10-21
Parse physical pixel sizes from CellSens .vsi datasets

melissalinkert:
Fixes http://trac.openmicroscopy.org.uk/ome/ticket/9805.  CellSens datasets (including test_images_good/cellsens) should now have PhysicalSizeX and PhysicalSizeY set for the largest resolution in each pyramid.  Smaller resolutions in the pyramid will not have physical sizes set, by design.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -826,6 +826,9 @@ public class CellSensReader extends FormatReader {
         if (!imageName.equals("Overview") && !imageName.equals("Label") && duplicate) {
           imageName += " #" + ii;
         }
+        if (imageName.equals("Overview") || imageName.equals("Label")) {
+          imageName = imageName.toLowerCase();
+        }
         store.setImageName(imageName, ii);
         store.setObjectiveSettingsID(MetadataTools.createLSID("Objective", 0, nextPyramid - 1), ii);
         store.setObjectiveSettingsRefractiveIndex(pyramid.refractiveIndex, ii);
@@ -844,7 +847,7 @@ public class CellSensReader extends FormatReader {
         }
       }
       else {
-        store.setImageName("Macro Image", ii);
+        store.setImageName("macro image", ii);
       }
 
       i += core.get(i).resolutionCount;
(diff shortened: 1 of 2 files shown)


melissalinkert:
Finally fixed the issues with reading .vsi files using Memoizer.  Removing the exclude label, as no more commits are planned and Travis is happy.

Note that this also fixes http://trac.openmicroscopy.org/ome/ticket/12433 now.  It would be best to spot check importing into OMERO the files from test_images_good, from_skyking/cellsens/olivier/small-tiles, and maybe one or two others.  Import should succeed, physical pixel sizes and exposure times should be present, and with the exception of test_images_good/anda pyramid generation should not occur.

components/formats-gpl/src/loci/formats/in/CellSensReader.java
```diff
@@ -177,17 +427,18 @@ public boolean isSingleFile(String id) throws FormatException, IOException {
...
   /* @see loci.formats.IFormatReader#getOptimalTileWidth() */
   @Override
   public int getOptimalTileWidth() {
     FormatTools.assertId(currentId, true, 1);
-    if (getCoreIndex() < core.size() - ifds.size()) {
+    if (getCoreIndex() < core.size() - 1) {
       return tileX.get(getCoreIndex());
     }
-    int ifdIndex = ifds.size() - (core.size() - getCoreIndex());
+    int ifdIndex = 1 - (core.size() - getCoreIndex());
```
rleigh-dundee: Is this likely to result in a negative index?  It's also done below here.

rleigh-dundee:
No import problems seen importing and viewing in OMERO.  Metadata looks good, and the expanded metadata coverage is a great improvement.  Looks good to merge.

components/formats-gpl/src/loci/formats/in/CellSensReader.java
melissalinkert: No - the ```if``` clause on line 438 ensures that ```getCoreIndex()``` will be >= ```core.size() - 1``` if line 441 is executed.

melissalinkert:
--no-rebase

PR #1407, opened 2014-11-06
TIFF: fix plane count calculation for pyramids with multiple planes

melissalinkert:
Fixes https://trac.openmicroscopy.org.uk/ome/ticket/12617.  To test, import the files from QA 10321 and verify that all Z sections (7 for the larger files, 1 for the smaller file) can be viewed without an exception in Blitz-0.log after pyramid generation completes.  Pyramids generated without this change should be read correctly as well, so reimport should not be strictly necessary if all files were previously imported into a develop server.

/cc @jburel, @zeb


melissalinkert:
Finally pushed a commit to fix the SizeT for the Gatan pyramid.  Likely needs one more quick round of review to make sure that nothing is broken by that change.

bramalingam:
Checked QA 10321. Everything works as expected (Pyramid generation happens without any errors, but it took around 4 hours to complete the process). Please merge if that is expected.

melissalinkert:
--no-rebase

PR #1421, opened 2014-11-17
Units mock

jburel:
In this PR: 
- handle case of bindata tag. (XMLReader does not generate the tag Bin:bindata)
- Convert cut in/out value from int to float.


jburel:
--rebased-to #1475 (partially)

PR #1426, opened 2014-11-19
Initial Rendering Definitions schema

qidane:
This introduces RenderingDefinations to the model.

----

--breaking
--exclude


qidane:
Java building, c++ is not. This could be used as a reference for future rendering work.

PR #1443, opened 2014-11-26
Genshi source (embedded)

joshmoore:
The genshi requirement added to the Bio-Formats build sometime ago has broken the IDE workflows that we advertise on the webpage, since the code-generation needed for a valid workspace was missing the Python library. In order to restore IDE usage, this PR adds a fallback version of genshi to the build.

Options looked at:
 * Add genshi zip -- unliked, see https://github.com/openmicroscopy/bioformats/pull/1443#issuecomment-64659772
 * Add genshi source -- current solution
 * Download genshi from Maven -- possibly next step

/cc @melissalinkert @rleigh-dundee @sbesson

### DIFF
--- a/ant/xsd-fu.xml
+++ b/ant/xsd-fu.xml
@@ -27,22 +27,22 @@ Include this in components generating sources using xsd-fu.
     <element name="args" implicit="yes"/>
     <sequential>
       <exec executable="python" failonerror="true">
-	<arg value="${root.dir}/components/xsd-fu/xsd-fu"/>
-	<args/>
-	<arg value="--ome-xml-model-package=@{omexml-model-package}"/>
+        <arg value="${root.dir}/components/xsd-fu/xsd-fu"/>
+        <args/>
+        <arg value="--ome-xml-model-package=@{omexml-model-package}"/>
         <arg value="--ome-xml-model-enums-package=@{omexml-model-enums-package}"/>
         <arg value="--ome-xml-model-enum-handlers-package=@{omexml-model-enum-handlers-package}"/>
-	<arg value="--metadata-package=@{metadata-package}"/>
-	<arg value="--ome-xml-metadata-package=@{omexml-metadata-package}"/>
-	<arg value="--language=@{lang}"/>
-	<arg value="--template-path=${xsdfu.templatepath}"/>
-	<arg value="--output-directory=@{output}/"/>
-	<arg value="--print-generated"/>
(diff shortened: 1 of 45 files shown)


joshmoore:
Took the middle ground and commited the source under `components/xsd-fu/python` which requires none of the previous changes to xsd-fu.xml or ome-xml/pom.xml. I'll see if I can pinpoint @dominikl's issues. Removing excluded.

joshmoore:
@dominikl : with Eclipse 4.4, I was asked to install `build-helper:1.4` and I'm still getting "Plugin execution not covered by lifecycle configuration: org.codehaus.mojo:build-helper-maven-plugin:1.4:add-source (execution: add-source, phase: generate-sources)" for ome-xml.  I can follow the suggestion from Eclipse and get rid of the errors by adding the following to 3 pom.xml files:
```
+                <pluginExecution>
+                  <pluginExecutionFilter>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>
+                      build-helper-maven-plugin
+                    </artifactId>
+                    <versionRange>[1.4,)</versionRange>
+                    <goals>
+                      <goal>add-source</goal>
+                    </goals>
+                  </pluginExecutionFilter>
+                  <action>
+                    <ignore></ignore>
+                  </action>
+                </pluginExecution>
```
but I don't know the implications. @melissalinkert? (Also it doesn't seem to be building the source after a `git clean -fdx`)

melissalinkert:
I think that POM change would be fine.

As for not rebuilding, maybe check that it's not trying to read from your ~/.m2/repository/ instead?

joshmoore:
Even though the eclipse support is not 100%, I'm still adding this for review so that we can at least get the genshi failures taken care of.

rleigh-dundee:
Looks fine, though it would be nice if xsd-fu would only import if if it was not already available on the existing python path so that it's only used as a fallback in the absence of a system copy.  It's currently overriding the system copy unconditionally.  The logic to check is already in xsd-fu, it just needs the path adding there, and moving to an alternative path e.g. xsd-fu/genshi rather than xsd-fu/python/genshi.  There may be a cleaner way.

joshmoore:
@rleigh-dundee : understood, but I'd really at least like to get us to the point where IDE users aren't completely broken. I'll rebase this and then someone can look into making this conditional.

joshmoore:
--rebased-to #1546

PR #1442, opened 2014-11-26
Update roi handler

bramalingam:
To test : 

1) Add bioformats_package.jar to your ImageJ plugins folder
2) Open any Image in ImageJ, draw roi's and add them to the ROI Manager.
3) Try exporting images using Plugins-->Bio-Formats-->Bio-Formats Exporter. 

This should write the ROI's from the ROI manager to your ome-tiff.

To check : 

Using Command line tools : 

---. /showinf -nopix -omexml /Users/bramalingam/Desktop/ImageJTest/finalTest.ome.tif

This should show all the ROI's , within "< ROI >" " < /ROI >"

Using ImageJ (Post Export):
_Note : Please replace the develop version of Bioformats_package.jar in your ImageJ folder to 5.0._ version of Bioformats_package.jar, the current Develop version has a known issue while reading xml files with unicode unit and will not show the ROI's that were written earlier.

Import the Image using Bio-formats Importer (select checkbox "Display ROI's" while import) and check if the ROI Manager is populated with the ROI's (that were drawn earlier).

Also : 
4) Try overwriting an existing file, while exporting.

PR #1470, opened 2014-12-08
Fix writing of tiff description (rebased onto dev_5_0)

jburel:
This is the same as gh-1459 but rebased onto dev_5_0.

----

To test the PR

* Drop bio-formats_tools into imageJ plugins
* Import an image e.g. png (This will generate some size with microns)
* export the image as OME-TIFF
* Check that the OME-TIFF is opened using the OMETiff reader either by using showinf or BF-importer in imageJ

To come: test but this PR should allow @bramalingam and @pwalczysko to review gh-1442

You will need to edit the description of the OME-TIFF and add a unicode character before import


pwalczysko:
The image described in the comment https://github.com/openmicroscopy/bioformats/pull/1470#issuecomment-67527575 and sent to me by @jburel would be worth preserving, wouldn't it ? At the moment it is at ``team/pwalczysko/ome-tiff-with-unicode/``. 
Should we kindly ask @melissalinkert to put it somewhere on ``test_images_good``  or is it too rudimentary or do we need something else in this line ? (should we sync this with the units-sample-files effort ? 
cc @qidane

PR #1475, opened 2014-12-09
Decode

jburel:
Cherry-pick commits from gh-1421
Improve XML-reader.
Do not include guava, Version on dev_5_0 is 13. (too old!)


jburel:
--rebased-from #1421 (partially)

PR #1492, opened 2014-12-15
Command line profiling

sbesson:
As we'll run more and more into performance testing, standardizing the way we collect and curate profiling traces across the board will become crucial. Since some of this profiling will be performed on remote machines, a requirement on JVisualVM is certainly overkill.
This PR makes a simple proposal: having a `BF_PROFILE` flag for command-line tools which enables `hprof` profiling and outputs an `.hprof` file to be examined via GUI tools.

Couple of remarks/questions at this point:
- is `hprof` the best candidate here? see http://www.brendangregg.com/blog/2014-06-09/java-cpu-sampling-using-hprof.html for a discussion and I know @joshmoore had other ideas
- where is the best place for documenting this (as well as other command-line tools environment variables), I started working on `users/comlinetools/index.txt` but then wondered wether we should not have a developers page for these features /cc @hflynn

### DIFF
--- a/components/autogen/src/gen-meta-support.sh
+++ b/components/autogen/src/gen-meta-support.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ###
 # #%L
--- a/components/test-suite/bftest
+++ b/components/test-suite/bftest
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # bftest: a script for running the Bio-Formats automated test suite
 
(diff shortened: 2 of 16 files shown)


melissalinkert:
All seems OK now.  Going to merge for now, and can rearrange the command line tools sections/pages later (since other work probably needs to be done there as well).

sbesson:
Is it worth rebasing or is it `develop` only?

sbesson:
--no-rebase

melissalinkert:
Sorry - assumed this fell under the docs improvement category, and so wasn't up for rebasing.

PR #1493, opened 2014-12-15
Add a page that gives an overview of each repository component

melissalinkert:
See https://trello.com/c/R7o2B8RV/99-add-a-page-on-which-component-is-responsible-for-what.   Almost certainly needs more work, but some feedback on what is missing or doesn't make sense would be helpful.


melissalinkert:
--no-rebase

PR #1499, opened 2014-12-18
C++: XML validation, writing and wrapper cleanup

rleigh-dundee:
- Refactor wrapper code; it could segfault under certain circumstances when the wrapper didn't refcount things properly.  It's now much simpler and much more robust, using std::shared_ptr deleters rather than templated manager wrappers to proxy for deleters.
- Add functions to write Document and Node objects to strings, streams and files.
- Add validateXML function to validate a string.
- Initial reading from file/stream/string throws if validation fails, so this is already self-validating.

--no-rebase

--------

Testing: there are unit tests for all changes, so if the tests pass they should be OK.


cpp/lib/ome/xerces/dom/Base.h
```diff
@@ -376,12 +84,30 @@ namespace ome
         {}
 
         /**
-         * Construct with initial wrapped value.
+         * Construct with initial wrapped value (managed).
          *
-         * @param base the value to wrap.
+         * @param wrapped the value to wrap.
+         * @param del the deleter to clean up the wrapped value
```
mtbc: missing period

cpp/test/ome-xerces/data/18x24y5z5t2c8b-text-invalid.ome
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0"?>
+<OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2012-06" xmlns:Bin="http://www.openmicroscopy.org/Schemas/BinaryFile/2012-06" xmlns:SPW="http://www.openmicroscopy.org/Schemas/SPW/2012-06" xmlns:SA="http://www.openmicroscopy.org/Schemas/SA/2012-06" xmlns:ROI="http://www.openmicroscopy.org/Schemas/ROI/2012-06" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2012-06
+     http://www.openmicroscopy.org/Schemas/OME/2012-06/ome.xsd">
```
mtbc: not the 2013 one?

mtbc:
I don't see any obvious problems with the code but would be good to get the CI job back to green before considering merging.

cpp/lib/ome/xerces/dom/Base.h
rleigh-dundee: Fixed now.

rleigh-dundee:
Fixed the tiny typo in the docstring.  Should be fine to merge now.

PR #1500, opened 2014-12-19
take new dimensions into account

bjoernthiel:
in addition to pull request #1480


melissalinkert:
Thanks again, @bjoernthiel.  We will need to add the Open Microscopy copyright block in a separate PR, but we'll make sure that the MPI copyright is not changed going forward.

--no-rebase

PR #1501, opened 2014-12-19
FormatReader/FormatTools Modulo improvements

rleigh-dundee:
When implementing a modulo-aware viewer, it became quickly clear that computing dimension coordinates when using Modulo was a real pain.  It requires all calculations to be done by the user, which quickly gets quite annoying.  This PR implements Modulo-aware variants of getIndex and getZCTCoords in both FormatReader and FormatTools, and updates the dependent classes to use them.  This allows direct computation of the full ZCTmZmCmT sextuple from an index and vice-versa.  Here, the ZCT coordinates are modified to take the Modulo sizes into account, so the user can work easily using the effective ZCTmZmCmT ranges.

Outstanding question: getIndex overloads naturally due to having extra arguments.  getZCTCoords does not, so was supplemented with getZCTModuloCoords.  In FormatTools, the getZCTCoords parameters allow overloading, but this could be renamed to ModuloCoords if desired.

--------

Testing:

The C++ implementation has a comprehensive set of testcases.  The Java code didn't have unit tests for ZCT, so ZCTmZmCmT variants were not added, but the logic is identical to the C++ implementation.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -315,7 +315,7 @@ public final class FormatTools {
 
   /**
    * Gets the rasterized index corresponding
-   * to the given Z, C and T coordinates.
+   * to the given Z, C and T coordinates (real sizes).
    */
   public static int getIndex(IFormatReader reader, int z, int c, int t) {
     String order = reader.getDimensionOrder();
@@ -328,7 +328,11 @@ public final class FormatTools {
 
   /**
    * Gets the rasterized index corresponding to the given Z, C, T,
-   * ModuloZ, ModuloC and ModuloT coordinates.
+   * ModuloZ, ModuloC and ModuloT coordinates (effective sizes).  Note
+   * that the Z, C and T coordinates take the modulo dimension sizes
+   * into account.  The effective size for each of these dimensions is
+   * limited to the total size of the dimension divided by the modulo
+   * size.
    */
   public static int getIndex(IFormatReader reader, int z, int c, int t,
                              int moduloZ, int moduloC, int moduloT) {
@@ -350,17 +354,17 @@ public final class FormatTools {
 
   /**
    * Gets the rasterized index corresponding
-   * to the given Z, C and T coordinates.
+   * to the given Z, C and T coordinates (real sizes).
    *
    * @param order Dimension order.
-   * @param zSize Total number of focal planes.
-   * @param cSize Total number of channels.
-   * @param tSize Total number of time points.
+   * @param zSize Total number of focal planes (real size).
+   * @param cSize Total number of channels (real size).
+   * @param tSize Total number of time points (real size).
    * @param num Total number of image planes (zSize * cSize * tSize),
    *   specified as a consistency check.
-   * @param z Z coordinate of ZCT coordinate triple to convert to 1D index.
-   * @param c C coordinate of ZCT coordinate triple to convert to 1D index.
-   * @param t T coordinate of ZCT coordinate triple to convert to 1D index.
+   * @param z Z coordinate of ZCT coordinate triple to convert to 1D index (real size).
+   * @param c C coordinate of ZCT coordinate triple to convert to 1D index (real size).
+   * @param t T coordinate of ZCT coordinate triple to convert to 1D index (real size).
    */
   public static int getIndex(String order, int zSize, int cSize, int tSize,
     int num, int z, int c, int t)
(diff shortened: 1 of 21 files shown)


components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -493,6 +567,37 @@ public static int getIndex(String order, int zSize, int cSize, int tSize,
...
+    int num, int index) {
+    int[] coords = getZCTCoords(order, zSize, cSize, tSize, num, index);
+
+    return new int[] {
+        coords[0] / moduloZSize,
+        coords[1] / moduloCSize,
+        coords[2] / moduloTSize,
+        coords[0] % moduloZSize,
+        coords[1] % moduloCSize,
+        coords[2] % moduloTSize
```
mtbc: now I know why we call them "modulo"!

cpp/lib/ome/test/main.cpp
```diff
@@ -45,3 +45,11 @@ main (int   argc,
   testing::InitGoogleTest(&argc, argv);
   return RUN_ALL_TESTS();
 }
+
+bool
+verbose ()
+{
+  static bool v(getenv("BIOFORMATS_TEST_VERBOSE")
+                && std::string(getenv("BIOFORMATS_TEST_VERBOSE")) == "true");
```
mtbc: I assume there's other code elsewhere that is equally picky about "true"'s case, etc., so this is good in context.

cpp/test/ome-bioformats/formatreader.cpp
```diff
@@ -563,6 +609,22 @@ TEST_P(FormatReaderTest, FlatSeries)
...
+      moddims(0,  1, 0, 0, 0, 0),
+      moddims(0,  0, 1, 0, 0, 0),
+      moddims(0,  1, 0, 1, 0, 0),
+      moddims(0,  0, 1, 1, 0, 0),
+      moddims(0,  1, 1, 0, 0, 0),
+      moddims(0,  1, 1, 1, 0, 0),
+      moddims(0,  2, 1, 3, 0, 0),
+      moddims(2,  3, 0, 2, 0, 0),
+      moddims(1,  2, 1, 3, 0, 0),
+      moddims(3,  3, 1, 4, 0, 0)
```
mtbc: the double space makes sense for some reason?

mtbc:
No obvious code problems, just need to get the build happy again.

cpp/test/ome-bioformats/formatreader.cpp
rleigh-dundee: No, just a spurious space propagated by copy-paste; pushed a fix now.

rleigh-dundee:
Fixed the minor alignment issue spotted.  Should be good to merge now.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -405,6 +427,42 @@ public static int getIndex(String order, int zSize, int cSize, int tSize,
...
+    int z, int c, int t,
+    int moduloZ, int moduloC, int moduloT) {
+    return getIndex(order,
+                    zSize,
+                    cSize,
+                    tSize,
+                    num,
+                    (z * moduloZSize) + moduloZ,
+                    (c * moduloCSize) + moduloC,
+                    (t * moduloTSize) + moduloT);
```
melissalinkert: I'm not 100% convinced of this logic.  I think either ```zSize```, ```cSize```, and ```tSize``` need to be the number of planes divided by the corresponding modulo size, or the last 3 arguments passed on to ```getIndex(String, int, int, int, int, int, int, int)``` should just be ```z```, ```c```, and ```t```.

Otherwise, if for example ```zSize``` is 100 (from ```getSizeZ()```), ```moduloZSize``` is 10, and ```z``` is 99 (since I expect based on the javadoc and other getIndex methods that ```z``` is in the range ```[0, zSize)```), then the Z index that is passed along would be 990.

rleigh-dundee: The "z" parameter here is the "effectiveSizeZ", i.e. the range is [0, zSize/moduloZSize) [half-open].  For all of the getIndex/get[Modulo]Coords methods added here, the z/c/t values are the effective values, accounting for the modulo dimension (since their intent is to allow the transparent use of the separated z/c/t and moduloz/c/t values); the user should never be exposed to the "real" z value; they just use the effective values and the plane index.

See the C++ testcases which do validate this logic.  If you use a z greater than (zSize/moduloZSize - 1) you'll get an exception.

The docstrings can certainly be updated to make this clearer.

melissalinkert: OK, but then you will need to fix lines 336-338, which are definitely not passing through the effective sizes.  Updating the Javadocs as well would certainly be helpful, as all of the other related methods are working with the absolute size Z/C/T.

rleigh-dundee: Sorry, I should have been clearer.  The FormatTools helper functions (both modulo and non-modulo) take the full real sizes for both zSize/cSize/tSize and the modulo sizes.  It needs the full size for the "num" check.  It could be passed the effective size and then compute the real size from it, but I retained the original intent to keep it similar to the original (and it can pass these values directly to the original), and because that's the information the caller has in practice; otherwise we'd need to compute the effective size, pass that and then immediately recompute the real size again.

The effective sizes are only used for the z/c/t indexes which follow.

I've pushed a commit which addresses this by clearly documenting what's used for each parameter for all the original and added methods.  Hope that's OK.

cpp/lib/ome/test/main.cpp
rleigh-dundee: This is only place this is used (at present).  It could be made more general if needed (anything that would parse as true), but at the moment it's purely for use by anyone running the tests by hand when debugging the test framework.

components/formats-api/src/loci/formats/FormatTools.java
rleigh-dundee: @melissalinkert If you feel that it makes more sense for the new functions in FormatTools to use the effective size rather than the real size, I'm certainly happy to adjust things.  It is certainly in some ways more logically consistent, but I'm not sure it's necessarily the most efficient way to implement it.  It's not likely to be a performance bottleneck, so happy to do things either way, whatever your preference is.

cpp/lib/ome/test/main.cpp
mtbc: I guess in part it depends on if you already have some convenient function handy analogous to Java's `Boolean.parseBoolean(String)`; suppose you could at least flatten the case.

melissalinkert: I assume any changes here are to happen in a separate PR?

mtbc: That would be okay with me.

PR #1502, opened 2014-12-19
Inveon: fix offsets for files larger than 2GB

melissalinkert:
See
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2014-December/004972.html

Builds should remain green with this change; the only thing to verify is that the offset passed to ```dat.seek``` is calculated using long (not int) arithmetic.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InveonReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InveonReader.java
@@ -110,7 +110,7 @@ public class InveonReader extends FormatReader {
   {
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
-    int planeSize = FormatTools.getPlaneSize(this);
+    long planeSize = (long) FormatTools.getPlaneSize(this);
     int index = getCoreIndex();
 
     RandomAccessInputStream dat = new RandomAccessInputStream(datFile);


components/formats-gpl/src/loci/formats/in/InveonReader.java
```diff
@@ -110,7 +110,7 @@ public boolean isThisType(RandomAccessInputStream stream) throws IOException {
   {
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
-    int planeSize = FormatTools.getPlaneSize(this);
+    long planeSize = (long) FormatTools.getPlaneSize(this);
```
rleigh-dundee: Given that FormatTools.getPlaneSize is returning an int and `sqrt(2^31) => 46340` we're likely to overflow for big histopathology images in general.  Should getPlaneSize (and the other helpers which use it) all not be switched to use `long`, saving the need for the cast here?  This will ensure we're not routinely overflowing for all readers using this and related helpers.

melissalinkert: That's up for debate I guess, but outside the scope of this PR.  Either way the return value of ```getPlaneSize(this)``` would need to be cast - if it returns ```long``` we're just casting when a byte array is constructed instead.

melissalinkert:
--no-rebase

components/formats-gpl/src/loci/formats/in/InveonReader.java
rleigh-dundee: Yes, it's certainly outside the scope of this PR, just thought it worth mentioning if it's a more general limitation we might want to address as a follow-on task.

PR #1503, opened 2014-12-19
ImageJ: enable min/max calculation if any series contains float data (rebased onto dev_5_0)

melissalinkert:
This is the same as gh-1495 but rebased onto dev_5_0.

----

Fixes the bug noted in https://github.com/openmicroscopy/bioformats/pull/1480#issuecomment-66973008.  The issue only appeared when the ```Autoscale``` option is turned off, the first image does not contain floating point data, and a subsequent image does contain floating point data and is the one selected to be opened.  

With this change, repeating the test from gh-1480 should result in the image being opened without an exception.

/cc @pwalczysko

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
@@ -523,17 +523,7 @@ public class ImportProcess implements StatusReporter {
     r = channelSeparator = new ChannelSeparator(r);
     r = dimensionSwapper = new DimensionSwapper(r);
 
-    boolean hasFloatingPoint = false;
-    for (int s=0; s<r.getSeriesCount(); s++) {
-      r.setSeries(s);
-      if (FormatTools.isFloatingPoint(r.getPixelType())) {
-        hasFloatingPoint = true;
-        break;
-      }
-    }
-    r.setSeries(0);
-
-    if (options.isAutoscale() || hasFloatingPoint) {
+    if (options.isAutoscale() || FormatTools.isFloatingPoint(r)) {
       r = minMaxCalculator = new MinMaxCalculator(r);
     }
     if (options.doStitchTiles()) {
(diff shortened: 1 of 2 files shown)


pwalczysko:
Sorry, but I am getting a different error in ImageJ when I perform the workflow.

```
loci.formats.FormatException: Unsupported stack format
	at loci.formats.in.OBFReader.initStack(OBFReader.java:369)
	at loci.formats.in.OBFReader.initFile(OBFReader.java:151)
	at loci.formats.FormatReader.setId(FormatReader.java:1317)
	at loci.plugins.in.ImportProcess.initializeFile(ImportProcess.java:494)
	at loci.plugins.in.ImportProcess.execute(ImportProcess.java:144)
	at loci.plugins.in.Importer.showDialogs(Importer.java:141)
	at loci.plugins.in.Importer.run(Importer.java:79)
	at loci.plugins.LociImporter.run(LociImporter.java:81)
	at ij.IJ.runUserPlugIn(IJ.java:195)
	at ij.IJ.runPlugIn(IJ.java:160)
	at ij.Executer.runCommand(Executer.java:128)
	at ij.Executer.run(Executer.java:64)
	at java.lang.Thread.run(Thread.java:695)
```
Workflow identical with https://github.com/openmicroscopy/bioformats/pull/1480#issuecomment-66973008. Used ``bioformats_package.jar`` in my ImageJ from https://ci.openmicroscopy.org/view/5.0/job/BIOFORMATS-5.0-merge-build/lastSuccessfulBuild/artifact/artifacts/

melissalinkert:
gh-1480 was not rebased, so that stack trace is expected.  Code review should be sufficient here, but I would also be OK with keeping this out of 5.0.7 if anyone is worried.

pwalczysko:
I am afraid that the proven benefit from this PR does not outweight the risk. Possibly if @rleigh-dundee could check the code and okay it, but still, we cannot improve the user experience on the original file which was problematic in 5.1, and we do not have any other files handy, so maybe we shouild leave it ?

pwalczysko:
(or do we know that there are many other files which will be affected / improved by this fix ?)

melissalinkert:
It's definitely an edge case, and can already be worked around by always checking the ```Autoscale``` box upon import.  Closing in the interests of reducing our collective stress levels.

PR #1506, opened 2014-12-22
LZW: fix one more off-by-one (rebased onto dev_5_0)

melissalinkert:
This is the same as gh-1497 but rebased onto dev_5_0.

----

See http://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7705.  For each of the three files linked in the forum thread, the values of the last column of pixels should be compared against the last column of pixels returned by ```imread``` in Matlab.  Without this change, each of the files should have a handful of rows for which the values differ; with this change, the values should match for all rows.

See also gh-1200.


sbesson:
Tested against the three example TIFF files linked in the community thread. Pixels off with the `5.0.6` version of Bio-Formats are fixed with this commit. Ready to merge.

PR #1507, opened 2014-12-26
Add note about running data tests on Windows

melissalinkert:
The original ```test-automated``` command will fail on Windows, but adding quotes around each ```-D``` argument should work.


sbesson:
Tested under Windows R2008 server. Quoting was only necessary if the path to the data contained spaces for my testing but recommending this as good practice is certainly sensible.

melissalinkert:
--no-rebase

PR #1508, opened 2014-12-26
Fix Leica .lei timestamp formatting

melissalinkert:
Millisecond parsing in Joda has different results than with Date/SimpleDateFormat.  If the number of ms digits in the format string exceeds the number of ms digits in the date string, then zero padding happens on the wrong side (e.g. 93 becomes 930 instead of 093).  If the number of ms digits in the date string exceeds the number of ms digits in the format string, then parsing fails.

See https://github.com/JodaOrg/joda-time/issues/62.

Fixes http://trac.openmicroscopy.org/ome/ticket/12117.

### DIFF
--- a/components/formats-common/src/loci/common/DateTools.java
+++ b/components/formats-common/src/loci/common/DateTools.java
@@ -34,6 +34,8 @@ package loci.common;
 
 import java.text.FieldPosition;
 import java.text.ParsePosition;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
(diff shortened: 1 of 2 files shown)


components/formats-common/src/loci/common/DateTools.java
```diff
@@ -259,20 +261,10 @@ public static String formatDate(String date, String[] formats,
...
-      LOGGER.debug("Invalid timestamp '{}'", date);
-    }
-    catch (UnsupportedOperationException e) {
-      LOGGER.debug("Error parsing timestamp '{}'", date, e);
-    }
-    if (timestamp == null) return -1;
-    return timestamp.getMillis();
+    SimpleDateFormat f = new SimpleDateFormat(format);
+    Date d = f.parse(date, new ParsePosition(0));
+    if (d == null) return -1;
```
rleigh-dundee: Don't we want to continue to throw an IllegalArgumentException in this case as before, rather than returning -1, which is a valid-but-incorrect date?  There's no way to distinguish between a valid date and an error condition.

rleigh-dundee:
The change looks reasonable, but I'll have to do some testing to check how this behaves WRT timezones given that it's passing through Date.  Ideally this should be resulting in a UTC timestamp, not offset to local time.

melissalinkert:
Pushed two commits to revert the DateTools change and fix this in the Leica reader.  I am concerned that we will run into the same problem elsewhere, so it will be important to keep an eye out for similar problems going forward.

rleigh-dundee:
The change looks fine, thanks.  The only minor detail is that with `leica/martin/samples-SP2/samples-SP2.lei` I see:

```
Timestamp #094: 2007:10:26,11:20:15:437
Timestamp #095: 2007:10:26,11:20:18:93
Timestamp #096: 2007:10:26,11:20:18:93
Timestamp #097: 2007:10:26,11:20:20:765
```

Would it be possible to pad the millisecond formatting appropriately here, to remove the ambiguity?  Or alternatively, use the normalised Timestamp value so it's a proper ISO-8601 date string.

melissalinkert:
Should be sorted out now.

rleigh-dundee:
The output is now fine, thanks.  Do we need two assignments of `timestamps[seriesIndex][j]` though, or could it just be a single assignment using getString(64) as a parameter to getTime?

melissalinkert:
One more commit pushed to address the last comment.  Hopefully nothing further that needs to be fixed?

rleigh-dundee:
The change looks fine, thanks.  Good to merge if green from me.

melissalinkert:
--no-rebase

PR #1509, opened 2014-12-29
Fix FilePattern grouping Windows

melissalinkert:
This makes sure that file patterns are detected consistently, regardless
of how the file separator is escaped.

To test, run ```ant "-Dtestng.directory=\path\to\test_images_good\eps" test-automated``` on Windows without this PR, and verify that the ```SizeZ``` test fails.  With this PR, the same test should pass.

All builds are expected to remain green.  See also https://trello.com/c/0Rdx4AMl/113-nfs-on-windows


sbesson:
Tested from Windows 2008 server. Without this PR the `sizeZ` test fails while all the tests pass for the `eps` subset file with this PR included. Ready to merge.

melissalinkert:
--rebased-to #1523

PR #1518, opened 2015-01-09
Fix link to Leica LAS AF Lite software

sbesson:
This should fix the broken link on the Leica LIF format page.

PR #1519, opened 2015-01-09
Fix link to Leica LAS AF Lite software (rebased onto dev_5_0)

sbesson:
This is the same as gh-1518 but rebased onto dev_5_0.

----

This should fix the broken link on the Leica LIF format page.


sbesson:
--rebased-from #1518

PR #1514, opened 2015-01-06
Parse SignificantBits from OME-XML/OME-TIFF files

melissalinkert:
The value of SignificantBits/```Bits per pixel``` shown by ```showinf``` should now match the value in the original file's OME-XML.

/cc @sbesson

PR #1520, opened 2015-01-10
Parse SignificantBits from OME-XML/OME-TIFF files (rebased onto dev_5_0)

melissalinkert:
This is the same as gh-1514 but rebased onto dev_5_0.

----

The value of SignificantBits/```Bits per pixel``` shown by ```showinf``` should now match the value in the original file's OME-XML.

/cc @sbesson


melissalinkert:
--rebased-from #1514

sbesson:
Tested with `OME-XML` and `OME-TIFF` files with and without `SignificantBits`. Works as expected. Ready to merge.

PR #1522, opened 2015-01-12
NFS performance

melissalinkert:
Lots of changes to hopefully improve performance, particularly when reading over NFS on Windows.  See https://trello.com/c/0Rdx4AMl/113-nfs-on-windows and http://trac.openmicroscopy.org/ome/ticket/6586.

In the end, this does not cache file attributes in Location.  The cost of caching and the high potential for cached values to be incorrect as a result of a high latency or unstable network connection resulted in caching having a negligible performance difference.  I had also tried enabling directory list caching by default, but with an even slightly flaky connection that is problematic, as the first call to list() may or may not have been complete.

This PR instead takes the approach of streamlining common operations (such as XML parsing and OME-XML creation), as well as reducing the number of times directories are listed or helper readers (for multi-file formats) are initialized.  177074c and 1cb786d are the only changes at the lowest level of I/O.  80ca51a and 1cb786d are the two most questionable commits.

In local tests, this showed at least an order of magnitude improvement when reading e.g. large TIFF, Zeiss CZI, and HCS datasets over NFS on Windows (both client and server were Windows machines).   I would expect some performance improvement on non-Windows systems, or at the very least performance should not be any worse.

I would prefer that this not be merged until the test_images_good test is running on the maxquant CI node, so that 1) we know that the tests actually pass, and 2) we know what the test performance looks like on a real system.  maxquant aside, the only testing needed is to verify that all jobs continue to pass without requiring more time than usual.


melissalinkert:
Few more commits pushed to fix issues noticed by BIOFORMATS-5.1-merge-full-repository.

mtbc:
I wondered what effect this PR has on http://trac.openmicroscopy.org/ome/ticket/12217, so I repeated the tests from that ticket, importing `from_skyking/mias/frans/siRNA_PRIM1_03102008/001-365700055641` on Linux over CIFS.

Without this PR, normal import takes 1m18, in-place ln_s import takes 9m15.
With this PR, normal import takes 1m3, in-place ln_s import takes 5m56.

From that point of view :+1:

components/formats-bsd/src/loci/formats/FileStitcher.java
```diff
@@ -668,6 +668,7 @@ public void setOriginalMetadataPopulated(boolean populate) {
     for (ExternalSeries s : externals) {
       String[] f = s.getFiles();
       for (String file : f) {
+        file = new Location(file).getAbsolutePath();
```
rleigh-dundee: New file variable shadows the existing file variable; consider using a different name?  Also repeated in the next hunk.

components/formats-common/src/loci/common/NIOFileHandle.java
```diff
@@ -263,12 +263,23 @@ public int read(ByteBuffer buf) throws IOException {
   @Override
   public int read(ByteBuffer buf, int off, int len) throws IOException {
```
rleigh-dundee: Noted due to the `int` cast below.  Why is the interface here using `int` rather than `long` for the offset and length?  Is is supposed to be restricted to small sizes?

components/formats-gpl/src/loci/formats/in/CellWorxReader.java
```diff
@@ -711,14 +714,16 @@ private IFormatReader getReader(String file)
...
 
-    IMetadata metadata;
-    try{
-      metadata = service.createOMEXMLMetadata();
-    }
-    catch (ServiceException exc) {
-      throw new FormatException("Could not create OME-XML store.", exc);
+    if (omexml) {
+      IMetadata metadata;
+      try {
```
rleigh-dundee: Now that OMEXMLServiceImpl is in formats-api, are there any circumstances in which createOMEXMLMetadata could throw?  I'm wondering if all use of the OMEXMLService API could be simplified if it would be possible to remove the `throws ServiceException` from their methods.

components/formats-gpl/src/loci/formats/in/HamamatsuVMSReader.java
```diff
@@ -171,7 +175,10 @@ public void close(boolean fileOnly) throws IOException {
     if (!fileOnly) {
       tileFiles = null;
       files.clear();
-      service.close();
+      if (service != null) {
+        service.close();
+      }
+      service = null;
```
rleigh-dundee: Should be inside the conditional?

rleigh-dundee:
Only a few minor cosmetic bits noted, which don't really need fixing, plus a couple of more general questions which if they require fixing, could be done in a separate PR.

If all the jobs are green, should be good to merge.

components/formats-common/src/loci/common/NIOFileHandle.java
melissalinkert: ByteBuffer is limited in size to Integer.MAX_VALUE; see http://docs.oracle.com/javase/6/docs/api/java/nio/ByteBuffer.html

components/formats-gpl/src/loci/formats/in/CellWorxReader.java
melissalinkert: If ome-xml.jar were missing, ServiceException could still be thrown.

rleigh-dundee: I might well be incorrect here, but isn't ome-xml.jar now a hard dependency here since the OMEXMLServiceImpl implementation is in formats-api as well as the interface?

melissalinkert: Arguably, yes, but whether it should remain that way is up for debate.  Either way, this is part of the decoupling discussion, and completely outside the scope of what this PR is intended to accomplish.

melissalinkert:
One more commit pushed for cosmetic fixes.  Note that we're still waiting on Windows CI results before merging.

mtbc:
Last commit looks fine to me.

melissalinkert:
Temporarily excluding for gh-1608.

melissalinkert:
Fixed merge conflicts and removed exclude/on hold labels, since it would be good to have this in 5.1.0 (even if we don't have full Windows test coverage).

rleigh-dundee:
The latest code changes look fine.  I'll try testing on Windows in a while.  Some jobs currently failing, but not necessarily due to this PR.

rleigh-dundee:
I spent some time testing a range of files from the changed readers, and timing their performance on Windows7/Java8.  I may have been unlucky in my choices, but the ones I picked didn't show any significant improvement e.g. scanr plates.  That's not to say it isn't significantly better for other examples though, and it doesn't appear to be regressing, so looks good to merge from that POV if green.

If you have any examples which do show significant improvement I'll be happy to repeat.

melissalinkert:
I assume you have squig mounted on the Windows machines, and aren't testing with local copies of datasets?  I'd expect ```/ome/data_repo/from_Biozentrum_Basel-Round_2``` to show a difference; I'd also expect a change in the overall time to run ```ant -Dtestng.directory=/ome/data_repo/test_images_good -Dtestng.memory=6144m -Dtestng.threadCount=4 test-automated```.

Keep in mind that OS-level caching can affect the results if you're testing a file, then immediately switching branches and testing the same file again.  For best results, you'll need to test all of the files with one branch, restart the machine, then switch branches and test all of the files again - that way there is an accurate comparison of the initial read time.

rleigh-dundee:
Yes, I was using files on squig; definitely not local.  I'll give these a try tomorrow.

rleigh-dundee:
On Windows, I get 69 vs 203 mins with and without the PR, respectively (single-threaded).  There were 458 failures and 952 skipped tests out of 22800 tests total.  Is that expected?

The Bizentrum files were too big for my system to process (both showinf and bfconvert); I could use showinf for metadata only, but there wasn't any noticeable difference in that case.

melissalinkert:
No, 458 failures in test_images_good is not expected at all, and I don't see this on Windows locally.

rleigh-dundee:
I'll try it with and without this PR; it's quite possibly unrelated.

sbesson:
Without this PR, running automated tests against `test_images_good` from the HEAD of develop takes 78 minutes with 446 failures and 952 skips. With this PR included, the same tests against the same target complete in 53 minutes with 158 failures and 0 skips.

Logs of both runs have been copied for examination (only 3 obvious failures). Given the size and history of this PR, my proposal is to merge it as it is since 
1- it does not destabilize the [non Windows full repository tests](http://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/)
2- it fixes reading failures under Windows
3- it improves performance by a factor of 30%

Improvements for failing tests can be carried out as separate PRs.

jburel:
@sbesson:
Agree, this is a major improvement.
The remaining tasks can be done in following PRs.

melissalinkert:
--no-rebase

PR #1525, opened 2015-01-12
ICS: re-open the correct file when reading from a memo file

melissalinkert:
If reopenFile() is not overridden, then the .ics file will always be
opened, even if it does not contain the pixel data.

Fixes http://trac.openmicroscopy.org/ome/ticket/12700.  To test, verify that importing ```test_images_good/ics/qdna1.ics``` into OMERO and viewing with Insight shows the same image as opening ```test_images_good/ics/qdna1.ics``` in ImageJ.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/ICSReader.java
+++ b/components/formats-bsd/src/loci/formats/in/ICSReader.java
@@ -1407,7 +1407,7 @@ public class ICSReader extends FormatReader {
         m.moduloC.typeDescription = "TCSPC";
         m.moduloC.start = 0;
         m.moduloC.step = 1;
-        m.moduloC.end = clen0;
+        m.moduloC.end = clen0 - 1;
       }
     }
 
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
The code change looks OK and I get the same image in OMERO and ImageJ (and ImageInfo).

However, there is a discrepancy.  ImageInfo and OMERO show a greyscale image while in ImageJ it's RGB with only a red channel so the greyscale image is red.  ImageInfo show:

```
        SizeC = 1 (0 null x 2 Channel)
```

The `null` looks a little suspicious here, is this correct?

melissalinkert:
That wasn't introduced by this PR, but should be fixed in 2ac24a9.  The red coloring in ImageJ is likely a result of whatever the "Color mode" option was set to - see if setting it to "Default" works.

rleigh-dundee:
The channel information looks correct now, thanks.  The colour was just due to the colour mode of the import; it's fine as Default.

Good to merge, thanks.

melissalinkert:
--no-rebase

PR #1526, opened 2015-01-13
RoiHandler :  display label text ROI's in ImageJ

bramalingam:
Add the latest build of Bioformats_package to the ImageJ plugin's folder

1) Open a Image from desktop in ImageJ.
2) Draw a Text ROI (add to ROI Manager).
3) Use Bioformats Exporter to export the image, with the export ROI's checkbox selected.

This exports the image as an ome-tiff with ROI's in the OME-XML

4) Now reopen the Image in ImageJ (using Bioformats Importer)
5) Select Checkboxes in the Bio-Formats Importer : Display Metadata, Display OME-XML Metadata, Display ROI's. & press OK.
6) This should open and display the Text ROI's in the Image and load the same to the ROI Manager.


jburel:
You may want to remove duplicate ```;;``` in convertPoint method.

components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
```diff
@@ -50,8 +48,9 @@
 import loci.formats.meta.IMetadata;
 import loci.formats.meta.MetadataStore;
 import loci.formats.ome.OMEXMLMetadata;
+import ome.model.units.Length;
```
melissalinkert: This should be ```ome.units.quantity.Length```; the Travis build currently fails as a result.

bramalingam: I've corrected for that, trying to fix a few other shape parameters for Label Text.

mtbc:
Travis failure.

jburel:
@bramalingam needs to push a fix. The problem was already highlighted by @melissalinkert

components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
```diff
@@ -311,19 +330,50 @@ else if(ijRoi.getTypeAsString().matches("Rectangle")){//Check if its a Rectangle
...
+    private static void storeText(Roi ijRoi, MetadataStore store, int roiNum, int shape) {
+        // TODO Auto-generated method stub
+        TextRoi roi = (TextRoi) ijRoi;
+
+        store.setLabelX(roi.getPolygon().getBounds().getX(), roiNum, shape);
+        store.setLabelY(roi.getPolygon().getBounds().getY(), roiNum, shape);
+
+        store.setLabelText(roi.getText(), roiNum, shape);
+        if (roi.getStrokeWidth() > 0) {
+            store.setLabelStrokeWidth( new ome.units.quantity.Length((roi.getStrokeWidth()), UNITS.PIXEL), roiNum, shape);
```
melissalinkert: Is there a reason for using the fully-qualified name of ```Length``` when that class is imported?

components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
```diff
@@ -311,19 +330,50 @@ else if(ijRoi.getTypeAsString().matches("Rectangle")){//Check if its a Rectangle
 
     // -- Helper methods --
 
+    private static void storeText(Roi ijRoi, MetadataStore store, int roiNum, int shape) {
+        // TODO Auto-generated method stub
```
melissalinkert: ```TODO``` comment no longer necessary?

components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
jburel: No, @bramalingam is learning!!

mtbc:
Ready for review again?

jburel:
added to standup

jburel:
Suggestion: you could import ome.xml.model.Label to make things easier to read

 * I draw a text, it is saved as a rectangle in OME-TIFF.
 * Another problem: if a file with the same name exists, I decided to overwrite it. When I re-import the generated image, it did not overwrite the contents.

bramalingam:
Please avoid the overwrite bug , will deal with that in a separate PR. 

The testing instructions remain the same. (Except that, create new files every time you export as an ome-tiff- Do not overwrite an existing image while exporting).

pwalczysko:
I think principally it works, except that the text region does not save the text. From the metadata output it looks like the region is saved as a simple rectangle.

Before export (text is ``hhh`` in the bottom left corner):

screen shot 2015-01-23 at 14 21 02

After export and reimport (text is identical with the number of the region in ROI Manager):

screen shot 2015-01-23 at 14 22 31

From the Metadata file opened after reimport in IJ:
```
...
<ROI ID="ROI:0:0" xmlns="http://www.openmicroscopy.org/Schemas/ROI/2013-10-dev-5">
<Union>
<Shape ID="Shape:0:0" StrokeColor="5" Text="0001-0040-0039">
<Rectangle Height="24.0" Width="35.0" X="22.0" Y="28.0"/>
```

pwalczysko:
Please note that the image formats showing the problem https://github.com/openmicroscopy/bioformats/pull/1526#issuecomment-71199930 are ``.png`` (screenshots made on Mac OS X).
The ```dv`` format opened and exported and reimported shows following flaw (addition of extra characters to the original text, which was ``hhh``):

screen shot 2015-01-23 at 14 46 03

bramalingam:
@jburel : there it is! the '#10;' comes from no where into the text.

pwalczysko:
Updated the comments - the problem is only as described in https://github.com/openmicroscopy/bioformats/pull/1526#issuecomment-71203210. - there is no other problem, and it is not format-specific.

bramalingam:
The above commit should fix the bug mentioned by Petr.

components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
```diff
@@ -311,19 +337,48 @@ else if(ijRoi.getTypeAsString().matches("Rectangle")){//Check if its a Rectangle
...
+
+        store.setLabelX(roi.getPolygon().getBounds().getX(), roiNum, shape);
+        store.setLabelY(roi.getPolygon().getBounds().getY(), roiNum, shape);
+
+        store.setLabelText(roi.getText().trim(), roiNum, shape);
+        if (roi.getStrokeWidth() > 0) {
+            store.setLabelStrokeWidth( new Length((roi.getStrokeWidth()), UNITS.PIXEL), roiNum, shape);
+        }
+        if (roi.getStrokeColor() != null) {
+            store.setLabelStrokeColor(new ome.xml.model.primitives.Color(roi.getStrokeColor().getColorSpace().getType()) , roiNum, shape);
```
melissalinkert: This and the other places where store.set*Color are called won't preserve the ROI's color - ```getColorSpace().getType()``` on a ```java.awt.Color``` returns the type of the ColorSpace (e.g. RGB), not the color itself.

The easiest thing to do is to define a method that turns a java.awt.Color into an ome.xml.model.primitives.Color by passing the getRed(), getGreen(), getBlue(), and getAlpha() values on java.awt.Color to the OME Color's constructor.  That method can then be called here instead of constructing a Color from the ColorSpace type.

bramalingam: Should I append all the getRed,getGreen,getBlue and getAlpha values to form one integer value, that can be parsed to ome.xml.model.primitives.Color(Integer)?
```
    private static ome.xml.model.primitives.Color toOMExmlColor(java.awt.Color color){
        
        ome.xml.model.primitives.Color test = new ome.xml.model.primitives.Color(test1);
        return test ;
        
    }
```

melissalinkert: No, there is an ome.xml.model.primitives.Color constructor that accepts the red, green, blue and alpha values individually and performs the conversion internally.

See http://downloads.openmicroscopy.org/bio-formats/5.0.6/api/ome/xml/model/primitives/Color.html#Color(int,%20int,%20int,%20int)

pwalczysko:
The problem with the additonal characters as described in https://github.com/openmicroscopy/bioformats/pull/1526#issuecomment-71203210 is fixed.

pwalczysko:
The properties of the text region and other regions are working as expected, except that the font size of text gets defaulted to be the same for all the text regions. For example, if the last drawn region was font size 13, all the regions saved have a displayed font size 13 after reimport, regardless of their orignally set and saved font size.

pwalczysko:
Tested the text ROI - the problem with the different font sizes is solved.

bramalingam:
@jburel @pwalczysko : Any final comments on this PR?

pwalczysko:
Discussed with @bramalingam : Fill color problems. @bramalingam will push a commit - this should be ready for testing tomorrow. 

  - Also test ROIs created in OMERO and opened in ImageJ.

pwalczysko:
The last commit works fine only when no properties were changed in OMERO. If I for example
   - create a rectangle in insigh-ij plugin on an image in OMERO
   - in ROI Manager of OMERO, go to Inspector tab
   - change the color of the line to red
   - change the line width to 12
   - save the ROI
   - open the image in ImageJ
   - the ROI will be shown on the image only as a "ghost", indicated just by number and the border selection (if I select it in the ROI Man. of IJ)

pwalczysko:
Works as expected. Tried also text ROIs, changes in font size and color, works fine.

melissalinkert:
This is OK to merge then?  No further objections?

pwalczysko:
No objections.

bramalingam:
--no-rebase

PR #1532, opened 2015-01-16
Compile MIPAV plugin

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/12420.

To test, verify that ```ant clean jars utils``` results in ```artifacts/mipav-stubs.jar``` being created and ```*.class``` files being present in ```components/formats-bsd/utils/mipav```.  All builds are expected to remain green.

As noted in the commit messages, it's unclear if deploying the MIPAV .jar to Artifactory would violate the license terms, so this just adds stubs for the few methods/classes needed by the plugin.  The stub jar is intentionally not added to loci_tools.jar/bioformats_package.jar, as doing so would cause problems for users who have the real MIPAV .jar on their classpath.


mtbc:
Travis failure.

melissalinkert:
Fix pushed.

mtbc:
Seems to work well: `mipav-stubs.jar` has classes, `PlugInBioFormatsImporter.class` is in `components/formats-bsd/utils/mipav`, MIPAV stubs maven module builds fine. Looks good to merge.

melissalinkert:
--no-rebase

PR #1531, opened 2015-01-16
Prevent NumberFormatException in ZeissCZIReader

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004988.html

To test this PR, check the samples files uploaded with the thread open correctly without throwing a `NumberFormatException` and that the full repository job stays green.

PR #1535, opened 2015-01-20
Prevent NumberFormatException in ZeissCZIReader (rebased onto dev_5_0)

sbesson:
This is the same as gh-1531 but rebased onto dev_5_0.

----

See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004988.html

To test this PR, check the samples files uploaded with the thread open correctly without throwing a `NumberFormatException` and that the full repository job stays green.


sbesson:
--rebased-from #1531

mtbc:
Guessing this is feedback #10434.

mtbc:
`serumstarve11214_1_confocal034Airy.czi` had one of the channels import as black, which obviously wasn't great for viewing, but once switched to red it all looked fine, and that issue's presumably outwith this PR. Looks good overall. :+1:

PR #1536, opened 2015-01-21
Deltavision: allow for ND filters stored as percent fractions

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/12692.  To test, verify that the ```NDFilter``` on ```Channel``` is ```0.5``` for test_images_good/dv/very_small.d3d.dv.  All builds should remain green.

/cc @will-moore

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -2377,7 +2377,7 @@ public class DeltavisionReader extends FormatReader {
         energyConvFactor = in.readFloat();
 
         // the stored value could be a percent fraction or a percentage
-        if (ndFilter > 1) {
+        if (ndFilter >= 1) {
           ndFilter /= 100;
         }
       }
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -2370,11 +2370,16 @@ public class DeltavisionReader extends FormatReader {
         in.skipBytes(4);
         expTime = in.readFloat();
 
-        ndFilter = in.readFloat() / 100;
+        ndFilter = in.readFloat();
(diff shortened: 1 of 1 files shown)


will-moore:
Hi @melissalinkert. Just looking at the code, I'm wondering if DVs ever have ND filters with 1% transmittance (which wouldn't be converted correctly if I've understood it right). Is it possible to actually read the % string that appears in the newer DVs, to be sure that we're dealing with percent?
If this is a real pain, maybe it would be better to assume that we are reading %, and always divide by 100 unless the value is below some cut-off. At least this way it'll only be the old files that have potential of being incorrect in edge cases. Cheers,

mtbc:
Test described for this PR passes:
info

mtbc:
"Download Original Metadata" disappeared from Insight in #3346. However, if I export as OME-TIFF I do see the string `NDFilter="0.5"` in that file. :+1:

melissalinkert:
@will-moore - newer DVs are likely reading the ND filter as text from a log file.  In this case, it's only being read as a single float field in the DV header itself, so there is no % to indicate whether or not to divide by 100.  1% transmittance seems unlikely, but I can certainly change the cut-off if you have an idea of what would be more appropriate.

will-moore:
In an e-mail from Michael a while back, he said:
"The DVs indeed show %T, with increments of 100%, 50%, 32%, 10% and I think maybe even 1% but I’m not confident about that."
So maybe the threshold is fine, but if it is 1 then still divide by 100:
```
    if (ndFilter >= 1) {
        ndFilter /= 100;
    }
```
so we get:
```
100 -> 1
1 -> 0.1
0.99 -> 0.99
0.5 -> 0.5
```
Thanks!

melissalinkert:
Commit pushed - thanks for clarifying.

will-moore:
Looks good and works fine too. Good to merge.

melissalinkert:
--rebased-to #1543

PR #1537, opened 2015-01-21
Metamorph: fix type of UIC1 tag 49 metadata

melissalinkert:
The metadata value's type is TiffRational, not double.

See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004995.html.  To test, check the original metadata of the attached .tif file with this PR included.  Unlike the attached original metadata file, there should not be any very small floating point values.  In particular, keys such as ```Channel #0 Camera Bit Depth```, ```Channel #0 _MagNA_```, and ```Channel #0 _MagRI_``` should have much more reasonable values.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -1798,11 +1798,10 @@ public class MetamorphReader extends BaseTiffReader {
 
     switch (type) {
       case 1:
-        in.skipBytes(1);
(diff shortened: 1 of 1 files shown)


sbesson:
Tested with the file attached to the ome-users thread as well as a couple of other sample metamorph files. The issue described above can be reproduced with 5.1.0-m3 and is fixed with this PR in. Good to merge once it has been tested by the full-repository job.

melissalinkert:
--rebased-to #1544

PR #1538, opened 2015-01-22
Metamorph: reset timepoint count if timelapse was not performed

melissalinkert:
See QA 10439.

To test, put the files from QA 10439 and QA 10442 in the same directory.  Open the .nd file (showinf, ImageJ, or OMERO) and verify that 15 planes are shown - 5 Z x 3 C.  Without this PR included, the same test should result in an exception as described in the QA issue.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -494,6 +494,9 @@ public class MetamorphReader extends BaseTiffReader {
       if (z != null) zc = Integer.parseInt(z);
       if (c != null) cc = Integer.parseInt(c);
       if (t != null) tc = Integer.parseInt(t);
+      else if (!doTimelapse) {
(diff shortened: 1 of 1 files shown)


mtbc:
Update copyright header?

Yes, 5×Z and 3×C. :+1:

In OMERO, one of the channels looks great, the other two have both min and max = 0 and appear featureless.

dominikl:
I'll check if that's an Insight problem, not sure where this min/max calculation actually happens.

melissalinkert:
It's not an Insight problem; only the first channel has data, so all zero for the last two channels is expected.

melissalinkert:
Leaving the copyright header alone, as snoopy is meant to update those now.

mtbc:
Oh, cool, that's neat.

snoopycrimecop:
No problem @mtbc.

melissalinkert:
--rebased-to #1545

PR #1540, opened 2015-01-24
C++: Complete OME-XML serialisation to and from OME model objects

rleigh-dundee:
- Fixes and additions to the Xerces-C wrappers
- Fixes and additions to the xsd-fu generator and the C++ templates
- Additional helper methods and functions added to ome-xml; parsing helpers move complexity out of the Genshi templates into real C++ templates, which is easier to edit and debug
- Addition of OME-XML output and validation to showinf
- Addition of model testcase to the ome-xml tests
- Add helpers to MetadataTools and XMLTools from Java MetadataTools, XMLTools and OMEXMLServiceImpl
- Revert model to 2013-06 for the 5.1.0 release

Note that the generated source code was changed by the template updates.

--no-rebase

--------

Testing: Unit tests (cpp/test/ome-xml/model) will test all the 2013-06 samples except for one with timestamps we don't currently handle [currently can't store dates prior to the POSIX epoch]; I added a custom version without pre-1970 dates for the time being.  Additional OME-XML could be tested by dropping it into the 2013-06 samples directory.

`showinf` should also display OME-XML metadata for TIFF files.  Run (in the build tree):

```
./bf bin/showinf/showinf --omexml yes --validate yes /path/to/tiff
```

Or in a downloaded binary build:

```
cd /path/to/unpacked/zip
export BIOFORMATS_HOME=$(pwd)
export LD_LIBRARY_PATH="${BIOFORMATS_HOME}/lib"
PATH="${BIOFORMATS_HOME}/bin:$PATH"
showinf --help
showinf --omexml yes --validate yes /path/to/tiff
```

Use `DYLD_LIBRARY_PATH` on MacOS X in place of `LD_LIBRARY_PATH`.

Note it is likely to be missing original metadata annotations at this point, but the OME-XML should validate and contain all the correct dimensions, etc.  (Multi-plane TIFFs will be timeseries and have SizeT==PlaneCount)

/cc @bramalingam @pwalczysko @emilroz

### DIFF
--- a/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
@@ -1132,15 +1132,20 @@ ${customUpdatePropertyContent[prop.name]}
 {% if not klass.isAbstractProprietary %}\
 {% if fu.SOURCE_TYPE == "header" %}\
         /// @copydoc ${lang.omexml_model_package}::OMEModelObject::asXMLElement
-        virtual xerces::dom::Element&
+        virtual xerces::dom::Element
         asXMLElement (xerces::dom::Document& document) const;
 {% end header %}\
 {% if fu.SOURCE_TYPE == "source" %}\
-      xerces::dom::Element&
+      xerces::dom::Element
       ${klass.name}::asXMLElement (xerces::dom::Document& document) const
       {
-        xerces::dom::Element nullelem;
-        return asXMLElementInternal(document, nullelem);
+{% if klass.name != "OME" %}\
+        xerces::dom::Element element(document.createElementNS(NAMESPACE, "${klass.name}"));
+{% end if %}\
+{% if klass.name == "OME" %}\
+        xerces::dom::Element element(document.getDocumentElement());
+{% end if %}\
+        return asXMLElementInternal(document, element);
       }
 {% end source %}\
 
@@ -1149,12 +1154,12 @@ ${customUpdatePropertyContent[prop.name]}
 
       protected:
         // Documented in base class.
-        virtual xerces::dom::Element&
+        virtual xerces::dom::Element
         asXMLElementInternal (xerces::dom::Document& document,
                               xerces::dom::Element&  element) const;
 {% end header %}\
 {% if fu.SOURCE_TYPE == "source" %}\
-      xerces::dom::Element&
+      xerces::dom::Element
       ${klass.name}::asXMLElementInternal (xerces::dom::Document& document,
                                            xerces::dom::Element&  element) const
       {
@@ -1170,11 +1175,6 @@ ${customUpdatePropertyContent[prop.name]}
             element = abstractElement;
           }
 {% end %}\
-        if (!element)
-          {
-            xerces::dom::Element newElement = document.createElementNS(NAMESPACE, "${klass.name}");
-            element = newElement;
-          }
 
 {% if klass.langBaseType is not None %}\
         // Element's text data
--- a/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
@@ -365,7 +365,7 @@ ${customContent}\
         }
 {% end %}\
 {% if not klass.isAbstractProprietary %}\
-        std::string tagName(element.getTagName());
+        std::string tagName(stripNamespacePrefix(element.getTagName()));
         if (tagName != "${klass.name}")
           {
             std::clog << "Expecting node name of ${klass.name} got " << tagName << std::endl;
--- a/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
@@ -300,7 +300,6 @@ namespace ome
         // Documented in superclass.
         const std::string&
         elementName() const;
-
 {% end header %}\
 {% if fu.SOURCE_TYPE == "source" %}\
       const std::string&
@@ -311,8 +310,8 @@ namespace ome
       }
 {% end source %}\
 {% end not Abstract %}\
-{% if len(customContent) > 0 %}\
 
+{% if len(customContent) > 0 %}\
 {% if fu.SOURCE_TYPE == "header" %}\
       private:
         /// Assignment operator (deleted).
--- a/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
@@ -69,6 +69,7 @@
 #include <ome/xerces/String.h>
 
 #include <ome/xml/model/ModelException.h>
+#include <ome/xml/model/detail/Parse.h>
 
 #include <boost/format.hpp>
 
@@ -348,15 +349,9 @@ ${customContent}\
 {% if klass.langBaseType != 'std::string' %}\
           if (!text.empty())
             {
-              std::istringstream is(text);
-              is.imbue(std::locale::classic());
-
-              if (!(is >> this->value))
-                {
-                  format fmt("Failed to parse ${klass.name} value ‘%1%’");
-                  fmt % text;
-                  throw ModelException(fmt.str());
-                }
+              set_value(text
+                        *this, &${klass.name}::setValue,
+                        "${klass.name}", "value");
             }
 {% end %}\
 {% if klass.langBaseType == 'std::string' %}\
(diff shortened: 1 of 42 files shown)


components/xsd-fu/templates-cpp/OMEXMLModelObject.template
```diff
@@ -1132,34 +1151,44 @@ ${customUpdatePropertyContent[prop.name]}
...
 {% if fu.SOURCE_TYPE == "source" %}\
-      xerces::dom::Element&
+      xerces::dom::Element
       ${klass.name}::asXMLElement (xerces::dom::Document& document) const
       {
-        xerces::dom::Element nullelem;
-        return asXMLElementInternal(document, nullelem);
+{% if klass.name != "OME" %}\
+        xerces::dom::Element element(document.createElementNS(NAMESPACE, "${klass.name}"));
+{% end if %}\
```
mtbc: there isn't an `else`?

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
```diff
@@ -1190,6 +1225,10 @@ ${customUpdatePropertyContent[prop.name]}
         element.setTextContent(this->value);
 {% end %}\
 {% end %}\
+{% if klass.isAnnotation %}\
+		// Ensure any base annotations add their Elements first
```
mtbc: it makes sense for the comment to be so far indented?

cpp/lib/ome/xerces/dom/Document.h
```diff
@@ -177,6 +178,60 @@ namespace ome
...
+        /// Validation scheme.
+        xercesc::XercesDOMParser::ValSchemes validationScheme;
+        /// Use namespaces?
+        bool doNamespaces;
+        /// Use schemas?
+        bool doSchema;
+        /// Handle multiple imports?
+        bool handleMultipleImports;
+        /// Do full checking during validation?
+        bool validationSchemaFullChecking;
```
mtbc: I assume that the meaning of "full checking" is fairly obvious to those using this kind of thing.

cpp/lib/ome/xerces/dom/NamedNodeMap.cpp
```diff
@@ -0,0 +1,58 @@
+/*
+ * #%L
+ * OME-XERCES C++ library for working with Xerces C++.
+ * %%
+ * Copyright © 2006 - 2014 Open Microscopy Environment:
```
mtbc: only 2014?

cpp/lib/ome/xml/meta/Convert.cpp
```diff
@@ -92,7 +92,6 @@ namespace
       convertFileAnnotations();
       convertListAnnotations();
       convertLongAnnotations();
-      convertMapAnnotations();
```
mtbc: the map annotation removal is interesting

cpp/lib/ome/xml/meta/Convert.cpp
```diff
@@ -229,6 +228,43 @@ namespace
...
+             P                         param3,
+             P                         param4)
+    {
+      bool ok = true;
+      try
+        {
+          (dest.*set)((src.*get)(param1, param2, param3, param4),
+                      param1, param2, param3, param4);
+        }
+      catch (const std::runtime_error& /* e */)
```
mtbc: ok to swallow actual error?

cpp/lib/ome/xerces/dom/NamedNodeMap.cpp
sbesson: Copyright update is now completely handled via an automatic script/CI job so reviewing these in Bio-Formats should not be necessary.

cpp/lib/ome/xml/meta/Convert.cpp
```diff
@@ -486,8 +521,12 @@ namespace
               for (index_type q = 0; q < binDataCount; ++q)
                 transfer(&MR::getPixelsBinDataBigEndian, &MS::setPixelsBinDataBigEndian, i, q);
 
-              index_type annotationRefCount(src.getImageAnnotationRefCount(i));
-              for (index_type q = 0; q < annotationRefCount; ++q)
+              index_type pixelsAnnotationRefCount(src.getPixelsAnnotationRefCount(i));
+              for (index_type q = 0; q < pixelsAnnotationRefCount; ++q)
+                transfer(&MR::getPixelsAnnotationRef, &MS::setPixelsAnnotationRef, i, q);
```
mtbc: Can we annotate pixels any more? Or is this for older models?

cpp/lib/ome/xerces/dom/NamedNodeMap.cpp
mtbc: Oh, yes, I think somebody already told me, thank you. :smiley:

cpp/test/ome-xerces/xerces.cpp
```diff
@@ -156,6 +162,15 @@ TEST_P(XercesTest, EmptyDocumentCreateElement)
...
 }
 
+TEST_P(XercesTest, EmptyDocumentCreateElementNS)
+{
+  xml::dom::Document document(ome::xerces::dom::createEmptyDocument("http://example.com/test/namespace1", "root"));
+  ASSERT_TRUE(document);
+  xml::dom::Element e(document.createElementNS("http://example.com/test/namespace2", "test"));
+  xml::dom::Element root(document.getDocumentElement());
+  root.appendChild(e);
+}
```
mtbc: anything further to assert here?

cpp/lib/ome/xml/meta/Convert.cpp
joshmoore: This will be tricky for a bit. C++ support is going to be one schema version behind so that we can get the work into a stable state ASAP.

cpp/lib/ome/xml/meta/Convert.cpp
joshmoore: See https://github.com/openmicroscopy/bioformats/pull/1540/files#r23686303

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
rleigh-dundee: No.  The OME root element is created with the document.

cpp/lib/ome/xerces/dom/Document.h
rleigh-dundee: It maps directly onto the Xerces API, so while I don't think it's really that obvious myself, it's directly matching the Xerces behaviour.  The entire ome-xerces wrapper is a thin layer over the Xerces-C API, so it's only intended to be used in combination with the Xerces API reference.

cpp/lib/ome/xml/meta/Convert.cpp
rleigh-dundee: The commits will be reverted after the 5.1 release so we can update to the new schema.  It's waiting on units to be done.

cpp/lib/ome/xml/meta/Convert.cpp
rleigh-dundee: This is all 2013-06 model reversion.

cpp/test/ome-xerces/xerces.cpp
rleigh-dundee: We could verify that the namespace is correct and all the elements are inserted correctly.  The main purpose is to ensure that the wrappers are valid and don't throw exceptions on any of the method calls, which doesn't require any assertions.

mtbc: That makes sense, thank you.

cpp/lib/ome/xml/meta/Convert.cpp
rleigh-dundee: Note that these changes are checked by ./cpp/test/ome-xml/metadata-convert-coverage which will flag up any missing calls.

qidane:
The .ome files look fine.

Would any of the changes to the xsd-fu files affect the java code generation? 
Could reverting it for C++ cause something to revert for Java?

rleigh-dundee:
The xsd-fu changes are restricted to the C++ templates so won't affect Java.  The exceptions are the tweaks to remove debug output but this won't affect the generated code.

joshmoore:
Looks like everyone's happy. Thanks @rleigh-dundee (oh, and @rleigh-codelibre too).

components/xsd-fu/xsd-fu
```diff
@@ -190,26 +192,6 @@ def main(model, opts):
             content = content.render(encoding='utf-8')
             customAsXMLElementPropertyContent[propertyName] = content
         # END Custom templates for asXMLElment()
-        parents = model.resolve_parents(obj.name)
```
joshmoore: Doh, perhaps a bit belatedly could you explain why this was removed?

cpp/lib/ome/xml/meta/Convert.cpp
rleigh-dundee: I didn't follow up to this one.  I don't really like this, but it's matching the behaviour of the Java implementation.  It also means that we can parse a document with minor errors.  It will need logging adding to log what the error was, once we've added a logging library (this is on the todo list).

components/xsd-fu/xsd-fu
rleigh-dundee: It's not removed; it's moved above so that the resolved parents are accessible in the ancillary template fragments.

joshmoore: Sorry, I intended to commend on `if opts.debug`

rleigh-dundee: Ah, OK.  In that case this was removed for a few reasons.  The cmake build captures stdout to introspect what's being generated, and enabling debugging breaks things horribly by spewing rubbish.  Debugging is now restricted to additional information expanded in the templates.  Additionally, the debug output is mostly just a copy of what's in the actual generated code, so writing it to a file and also to stdout isn't giving us anything we don't already have.  For the structured annoations, this is for debugging a single feature; this is included as a more general property list in the C++ templates, and the same can also be added to the Java templates if desired.

PR #1544, opened 2015-01-26
Metamorph: fix type of UIC1 tag 49 metadata (rebased onto dev_5_0)

melissalinkert:
This is the same as gh-1537 but rebased onto dev_5_0.

----

The metadata value's type is TiffRational, not double.

See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004995.html.  To test, check the original metadata of the attached .tif file with this PR included.  Unlike the attached original metadata file, there should not be any very small floating point values.  In particular, keys such as ```Channel #0 Camera Bit Depth```, ```Channel #0 _MagNA_```, and ```Channel #0 _MagRI_``` should have much more reasonable values.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -1776,11 +1776,10 @@ public class MetamorphReader extends BaseTiffReader {
 
     switch (type) {
       case 1:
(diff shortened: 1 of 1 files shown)


melissalinkert:
--rebased-from #1537

dominikl:
Values are looking sensible, good to merge.

PR #1545, opened 2015-01-26
Metamorph: reset timepoint count if timelapse was not performed (rebased onto dev_5_0)

melissalinkert:
This is the same as gh-1538 but rebased onto dev_5_0.

----

See QA 10439.

To test, put the files from QA 10439 and QA 10442 in the same directory.  Open the .nd file (showinf, ImageJ, or OMERO) and verify that 15 planes are shown - 5 Z x 3 C.  Without this PR included, the same test should result in an exception as described in the QA issue.


melissalinkert:
--rebased-from #1538

sbesson:
Tested with the merge builds. Works as expected. Relaunched Travis.

PR #1549, opened 2015-01-27
Zeiss CZI: set plane positions using pixel row/column values

melissalinkert:
If physical positions are not present, this means that the PositionX and
PositionY values will be set to the relative positions in pixels.

See QA 10445.  Without this PR, PositionX and PositionY on Plane are not populated for the QA file.  With this PR, ```showinf -omexml``` or ImageJ with ```Display OME-XML metadata``` checked should show both values populated for every plane.

Opening in ImageJ with ```Stitch tiles``` checked will arrange the tiles in order according to PositionX and PositionY, so this can be used to verify that the positions are correct - just note that the tile overlap is intentionally not removed, so the resulting images will not have completely smooth transitions between tiles.


bramalingam:

melissalinkert:
--no-rebase

PR #1550, opened 2015-01-27
Improve the copyright year update logic

sbesson:
This commit should now use the commit year from `git log -n1` per file to update the copyright header instead of using the current year unilaterally as presently.

### DIFF
--- a/tools/update_copyright
+++ b/tools/update_copyright
@@ -23,15 +23,17 @@ do
     echo "Updating $COPYRIGHT"
     for file in $(git grep -l "${COPYRIGHT}${YEAR}${RANGE}${YEAR}${AUTHOR}");
     do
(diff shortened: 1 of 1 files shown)


joshmoore:
Though I appreciate the comment that started this (https://github.com/openmicroscopy/bioformats/pull/1483#issuecomment-66972956), I think it's actually going to create more work than is needed at the moment. For example, the date on many files are actually those of the last time that the copyright was bumped, and so the results will still be skewed. The policy has been (for better or worse) to move to the current year, and since these PRs are about **reducing** the amount of time the team is worrying about headers, I'm closing this out of hand. At some point (far in the future), we can come back and re-evaluate this, but for the moment, I would really (__really__) like us to stop worrying about the copyright headers.

PR #1552, opened 2015-01-28
Save file

jburel:
Fix "overwrite" problem
To test the PR
test 1:
 * Open an image with mutli z and t in ImageJ
 * Click on BF-exporter
 * Select a name that already exists in chosen directory (if none, save first as ome-tiff)
 * to make sure the image is correctly updated, you should crop the original image
 * Click ok when the question to overwrite the file is asked.
 * Click ok for each of the following dialogs popping up
 * Check that the generated file is correct

test 2
* Repeat the steps from test 1 but this time select write z, t, c. 
* You will probably have to repeat twice to first generate the file with plane information.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -39,7 +39,6 @@ import ij.io.FileInfo;
 import ij.io.OpenDialog;
 import ij.measure.Calibration;
 import ij.plugin.frame.Recorder;
-import ij.plugin.frame.RoiManager;
 import ij.process.ByteProcessor;
 import ij.process.ColorProcessor;
 import ij.process.FloatProcessor;
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -154,6 +154,7 @@ public class Exporter {
             }
         }
 
+        File f = null;
         if (outfile == null || outfile.length() == 0) {
             // open a dialog prompting for the filename to save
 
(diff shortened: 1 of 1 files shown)


components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
```diff
@@ -560,6 +559,34 @@ else if (proc instanceof ShortProcessor) {
...
+            boolean in = false;
+            for (int i = 0; i < outputFiles.length; i++) {
+                String name = new File(outputFiles[i]).getName();
+                for (int j = 0; j < files.length; j++) {
+                    if (files[j].equals(name)) {
+                        in = true;
+                        break;
+                    }
+                }
+            }
```
melissalinkert: Could this be simplified a bit?  Something like:

```
boolean in = false;
for (String file : outputFiles) {
  if (new File(file).exists()) {
    in = true;
    break;
  }
}
```

jburel: Yes, I should have done that, I was testing the file output and did not simplify.
Too quick to move to other tasks

will-moore:
When you say "Open an image" - Using what software? / setup?

jburel:
@will-moore: sorry in ImageJ

bramalingam:
Everything works as expected. But a small fix might be required for the number of dialogs that pop up , when the full file name with extension is typed in while overwriting(was introduced by me previously). Fixing that in this PR itself would be great.

jburel:
@bramalingam: saving now fixed
Also stop the saving process when the user clicks on Cancel in The "split" dialog.

bramalingam:
Everything works as expected. Please merge.

jburel:
--no-rebase

PR #1551, opened 2015-01-28
Colorizer

joshmoore:
For supporting the application of lookup tables other than those provided in datasets, a new ReaderWrapper `Colorizer` has been added which takes an implementation of the new interface `LutSource` to convert one `byte[]` to another. The "standard" LUTs provided by ImageJ1's `LutLoader` are supported.

Testing:
 * `./bfconvert -colorize fire INPUT OUTPUT` should produce results similar to applying a LUT in ImageJ
 * `./showinf -colorize ice INPUT` show show an image with the alternative coloring.

Note: test OUTPUT should initially be of type `.ome.tiff`. `.png` is not yet working.

PR #1554, opened 2015-01-28
ND2: add the PFS offset and status to the original metadata table

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/12560.  To test, verify that ```PFS Offset``` and ```PFS Status``` entries appear in the original metadata table for .nd2 files that have PFS data listed in the acquisition metadata tab in NIS Elements Viewer.  I was testing with ```nd2/fiji-655/01.nd2```, though it might be a good idea to spot check one or two others.

I would expect builds to remain green.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -95,6 +95,7 @@ public class NativeND2Reader extends FormatReader {
   private int fieldIndex;
 
   private long xOffset, yOffset, zOffset;
+  private long pfsOffset, pfsStateOffset;
 
   private ArrayList<Length> posX;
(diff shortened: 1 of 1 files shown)


dominikl:
Tested with the same file nd2/fiji-655/01.nd2, couldn't find another nd2 file with PFS data. The PFS data is there, in the original metadata table. But I can't open the image in Insight, also thumbnail is black. Bioformats showinf also just shows a black image. So are all nd2 images in today's build.

dominikl:
Probably related to #1551 ... Will test again later, when this PR got excluded.

dominikl:
Tested again with today's build and image nd2/fiji-655/01.nd2. PFS data is there. Also was able to import and view the image. Looks good to merge to me.

melissalinkert:
--no-rebase

PR #1558, opened 2015-01-29
Add ROI's to an existing ome-tiff which already has ROI's 

bramalingam:
To test,

1) Import an ome-tiff which already has ROI's using the Bioformats_Importer.
2) Clear the list of ROI's in the ROIManger.
3) Draw new ROI's and add them to the ROIManager.
4) Export as ome-tiff using the the Bioformats exporter (with the export roi's checkbox selected).
5) Now reimport the saved image using Plugins > Bio-formats > Bio-formats Importer (“Please select the following checkboxes, "Display ROIs","Display OME-XML metadata" and "Display Metadata” in the bioformats-importer options)

This should have the old and new ROI's.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -177,7 +177,10 @@ public class ROIHandler {
 
         if (rois == null || rois.length == 0) return;
         List<String> discardList = new ArrayList<String>();
-        String roiID = null;int cntr = 0;
+        String roiID = null;
+        OME root = (OME) store.getRoot();
+        int roicount = root.sizeOfROIList();
+        int cntr = roicount;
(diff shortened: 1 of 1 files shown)


jburel:
When the image is opened, the rois are displayed in the data manager. 
If I deleted the rois from the data manager then export the file and override it.
The "deleted" rois are still there. This is confusing

Another point noticed, we should by default select Show all in roi manager when opening it with ROIs

bramalingam:
@jburel : I have pushed 2 commits to handle the issues stated above.

jburel:
@bramalingam: you have 2 branches with almost the same name
RoiHandler_ome-tiff_provenance and ROIHandler_ome-tiff_provenance
Are you sure you pushed to the correct one?

bramalingam:
replaced by https://github.com/openmicroscopy/bioformats/pull/1566

PR #1561, opened 2015-01-29
Fix file grouping and dimension swapping in ImageJ macros

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/12239 and http://trac.openmicroscopy.org/ome/ticket/12238.

To test, choose a single file with SizeZ > 1 or SizeT > 1 (the remaining dimensions should be 1).  I was using ```test_images_good/andor/VESSEL-DEC-Z.tif```.

For ticket 12239:
1. Start ImageJ, select ```Plugins > Macros > Record...```
2. Select ```Plugins > Bio-Formats > Bio-Formats Importer```
3. Select the test file
4. Check the ```Swap dimensions``` box, but leave everything else unchecked
5. Assign the > 1 dimension to something other than the default
6. Verify that the resulting image window reflects what was chosen in step 5.
7. Select the ```Create``` button in the macro recorder window, then ```Macros > Run Macro``` in the new macro window.
8. The image window created as a result of step 7 should match the one from step 5/6, and there should be no error message.

For ticket 12238:
1. Close ImageJ and repeat steps 1 - 3 above
2. Check the ```Group files with similar names``` box, but leave everything else unchecked
3. Check the ```Pattern``` box in the file grouping window
4. Repeat step 7 above
5. Verify that there are now two image windows open with matching contents.  There should be no error message.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
@@ -132,11 +132,10 @@ public class FilePatternDialog extends ImporterDialog {
       }
     };
     gd.addMessage(
-      "The list of files to be grouped can be specified in one of three ways:");
+      "The list of files to be grouped can be specified in one of the following ways:");
 
     // option one
 
-    gd.addCheckbox("", true);
     int len = id.length() + 1;
     if (len > 80) len = 80;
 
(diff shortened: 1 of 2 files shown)


bramalingam:
Both cases work as expected, Please merge.
```
run("Bio-Formats Importer", "open=/Users/bramalingam/Downloads/data_repo_good/stk/23srp5dksub.stk color_mode=Default view=Hyperstack stack_order=XYCZT swap_dimensions z_1=1 c_1=21 t_1=1");
```

```
run("Bio-Formats Importer", "open=/Users/bramalingam/Downloads/data_repo_good/andor/VESSEL-DEC-Z.tif color_mode=Default group_files view=Hyperstack stack_order=XYCZT contains=[] name=/Users/bramalingam/Downloads/data_repo_good/andor/VESSEL-DEC-Z.tif");
```

melissalinkert:
--no-rebase

PR #1570, opened 2015-02-03
Fix remaining DM_BOXED_PRIMITIVE_FOR_PARSING warnings

sbesson:
With this commit, Bio-Formats top-level findbugs target should be compliant with Findbugs 3.0.0

--no-rebase


melissalinkert:
:+1:

PR #1571, opened 2015-02-03
FlowSight: default to 1 channel if channel tags are missing

melissalinkert:
Fixes the file in QA 10450.

/cc @zeb

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FlowSightReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FlowSightReader.java
@@ -34,11 +34,11 @@ package loci.formats.in;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Iterator;
-
-import ome.xml.model.primitives.PositiveInteger;
+import javax.xml.parsers.ParserConfigurationException;
 
 import loci.common.DataTools;
 import loci.common.RandomAccessInputStream;
+import loci.common.xml.XMLTools;
 import loci.formats.CoreMetadata;
 import loci.formats.FormatException;
 import loci.formats.FormatReader;
@@ -47,6 +47,13 @@ import loci.formats.MetadataTools;
 import loci.formats.meta.MetadataStore;
 import loci.formats.tiff.IFD;
 import loci.formats.tiff.TiffParser;
+
+import ome.xml.model.primitives.PositiveInteger;
+
+import org.xml.sax.SAXException;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
 
 /**
  * @author Lee Kamentsky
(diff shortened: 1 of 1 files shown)


bramalingam:
./showinf -nopix /Users/bramalingam/Desktop/QA/amnis-isx-cif-sample/HeLasusp\ wt200_17.cif 

Works as expected. No exceptions thrown. Good to merge.

melissalinkert:
Would be good to get confirmation from @zeb as well, since he sent the file that this attempts to fix.

zeb:
Sorry I've let this slide - thanks for the nudge, @melissalinkert. 

Importing (via cli and insight) against a merge build (_5.1.0-m3-1151-5b3cf71-ice35-b411_) looks in good shape so far (due to the very large number of images generated from such samples, the import process takes several hours - will keep you posted after completion).

zeb:
CLI import result for the QA sample:
`==> Summary`
`1 file uploaded, 1 fileset created, 14124 images imported, 0 errors in 7:29:14.006`

melissalinkert:
One more commit pushed which should result in each of the image series having 12 channels now instead of 1.  This is using the ```/AssayDB-Rif/Imaging/ChannelInUseIndicators*``` node, since ```/AssayDB-Rif/Display/Channel*Name``` looks like it can define channels that aren't being used (though in this case all 12 are acquired).

The images do seem more sensible to me, but verification would be much appreciated.

rleigh-dundee:
The code changes look OK and there are 12 channels per series for all series.  There appears to be a bottleneck for OMERO import here, probably simply a factor of the series count.

@zeb Is the channel information and pixel data etc. OK for you with the change?

zeb:
Based on the file metadata, this looks fine by me: another sample with `<ChannelInUseIndicators_0_11>1 1 0 1 0 1 0 0 1 0 0 0 </ChannelInUseIndicators_0_11>` does produce 5 channels. Thanks for this change.

melissalinkert:
Thanks @zeb and @rleigh-dundee.  I'd definitely expect OMERO import to be slow and/or memory intensive, since it's going to be creating 14124 * 12 Channel objects in memory, which as discussed in the Tuesday meeting is not easy to deal with.

I assume it's OK to merge this then?

zeb:
Considering the performance implications, I think I'd be happy for this to be put on hold (or toggled via a feature flag is that makes any sense) for further testing if this is likely to become an import bottleneck/cause performance regressions for the original reader's companion samples/be too risky for inclusion in the 5.1 initial release.

melissalinkert:
The original sample data is unaffected, so I would be inclined to merge this since it does fix the original problem.  Import into OMERO was still slow prior to this PR and isn't something we can really address in the reader.

Happy to hear any strong vetos, otherwise I'll merge tomorrow (Thursday) evening.

melissalinkert:
--no-rebase

PR #1572, opened 2015-02-04
C++: Fixes and additions for OME-TIFF

rleigh-dundee:
Primarily metadata-related fixes, but includes a number of additional assorted fixes and API improvements.  The main change is to the generated model and metadatastore code, to enforce the safe use of shared and weak references.  Where the Java code would have thrown an NPE, here it segfaults the program, so this adds strict checking and will throw an exception if a bad reference is encountered (either null or bad weak refs).  This is what the internal `safe_fetch` function templates do.  They check the validity of the pointer and then return it back to the caller.  This approach was necessary due to the way the templates generate and chain multiple calls together.

--no-rebase

--------

Testing: Should be handled by the unit tests and travis.

When the OME-TIFF reader is added, it will test most of these changes fairly extensively.  You could try applying my cpp-ome-tiffreader branch on top and reading an OME-TIFF with showinf.  However, this isn't yet ready for testing and so will currently not work correctly (but will be completed very soon).

### DIFF
--- a/cpp/lib/ome/bioformats/MetadataTools.cpp
+++ b/cpp/lib/ome/bioformats/MetadataTools.cpp
@@ -179,7 +179,18 @@ namespace ome
     {
       // Parse OME-XML into DOM Document.
       ome::xerces::Platform xmlplat;
-      ome::xerces::dom::Document doc(ome::xerces::dom::createDocument(file));
+      ome::xerces::dom::Document doc;
+      try
+        {
+          doc = ome::xerces::dom::createDocument(file);
+        }
+      catch (const std::runtime_error&)
+        {
+          ome::xerces::dom::ParseParameters params;
+          params.doSchema = false;
+          params.validationSchemaFullChecking = false;
+          doc = ome::xerces::dom::createDocument(file, params);
+        }
       return createOMEXMLMetadata(doc);
     }
 
@@ -188,7 +199,19 @@ namespace ome
     {
       // Parse OME-XML into DOM Document.
       ome::xerces::Platform xmlplat;
-      ome::xerces::dom::Document doc(ome::xerces::dom::createDocument(text));
+      ome::xerces::dom::Document doc;
+      try
+        {
+          doc = ome::xerces::dom::createDocument(text, ome::xerces::dom::ParseParameters(),
+                                                 "OME-XML");
+        }
+      catch (const std::runtime_error&)
+        {
+          ome::xerces::dom::ParseParameters params;
+          params.doSchema = false;
+          params.validationSchemaFullChecking = false;
+          doc = ome::xerces::dom::createDocument(text, params, "Broken OME-XML");
+        }
       return createOMEXMLMetadata(doc);
     }
 
@@ -197,7 +220,19 @@ namespace ome
     {
       // Parse OME-XML into DOM Document.
       ome::xerces::Platform xmlplat;
-      ome::xerces::dom::Document doc(ome::xerces::dom::createDocument(stream));
+      ome::xerces::dom::Document doc;
+      try
+        {
+          doc = ome::xerces::dom::createDocument(stream, ome::xerces::dom::ParseParameters(),
+                                                 "OME-XML");
+        }
+      catch (const std::runtime_error&)
+        {
+          ome::xerces::dom::ParseParameters params;
+          params.doSchema = false;
+          params.validationSchemaFullChecking = false;
+          doc = ome::xerces::dom::createDocument(stream, params, "Broken OME-XML");
+        }
       return createOMEXMLMetadata(doc);
     }
 
--- a/cpp/lib/ome/bioformats/MetadataTools.cpp
+++ b/cpp/lib/ome/bioformats/MetadataTools.cpp
@@ -516,6 +516,37 @@ namespace ome
         }
     }
 
+    void
+    removeBinData(::ome::xml::meta::OMEXMLMetadata& omexml)
+    {
+      omexml.resolveReferences();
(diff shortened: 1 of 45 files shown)


mtbc:
Travis troubles.

rleigh-dundee:
Generated code changes: https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/BUILD_TYPE=Debug,CXXSTD_AUTODETECT=OFF,node=perch/lastSuccessfulBuild/artifact/generated-source-changes/ome-xml-changes.diff

rleigh-dundee:
The last commits fix the CI failure on gretzky35.

cpp/lib/ome/bioformats/MetadataTools.cpp
```diff
@@ -69,7 +76,11 @@ using ome::xml::meta::OMEXMLMetadataRoot;
 
 using ome::xml::model::Image;
 using ome::xml::model::MetadataOnly;
+using ome::xml::model::ModelException;
 using ome::xml::model::Pixels;
+using ome::xml::model::StructuredAnnotations;
+using ome::xml::model::XMLAnnotation;
+using ome::xml::model::OriginalMetadataAnnotation;
```
mtbc: not exactly lexicographically ordered, but maybe that's intended

cpp/lib/ome/bioformats/MetadataTools.cpp
```diff
@@ -494,6 +566,142 @@ namespace ome
...
+                      xerces::dom::Document doc(ome::xerces::dom::createDocument(annotation->getValue()));
+
+                      std::vector<xerces::dom::Element> OriginalMetadataValue_nodeList = ome::xml::model::detail::OMEModelObject::getChildrenByTagName(doc, "OriginalMetadata");
+                      if (OriginalMetadataValue_nodeList.size() > 1)
+                        {
+                          format fmt("Value node list size %1% != 1");
+                          fmt % OriginalMetadataValue_nodeList.size();
+                          throw ModelException(fmt.str());
+                        }
+                      else if (OriginalMetadataValue_nodeList.size() != 0)
```
mtbc: Would `== 1` be clearer in these conditions?

cpp/lib/ome/bioformats/detail/FormatReader.cpp
```diff
@@ -1304,12 +1305,21 @@ namespace ome
...
+      FormatReader::setId(const boost::filesystem::path& id)
       {
         //    LOGGER.debug("{} initializing {}", getFormat(), id);
         if (!currentId || id != currentId.get())
           {
-            initFile(id);
+            path canonicalID(id);
+            try
+              {
+                // Attempt to canonicalise the path.
```
mtbc: -ize

cpp/lib/ome/xml/model/OriginalMetadataAnnotation.cpp
```diff
@@ -0,0 +1,237 @@
...
+            format fmt("Value node list size %1% != 1");
+            fmt % XMLValue_nodeList.size();
+            throw ModelException(fmt.str());
+          }
+        else if (XMLValue_nodeList.size() != 0)
+          {
+            std::vector<xerces::dom::Element> OriginalMetadataValue_nodeList = getChildrenByTagName(XMLValue_nodeList.at(0), "OriginalMetadata");
+            if (OriginalMetadataValue_nodeList.size() > 1)
+              {
+                format fmt("Value node list size %1% != 1");
```
mtbc: For the original metadata exception do we want a different exception message string from the above so they are more easily distinguished?

cpp/lib/ome/xml/model/primitives/ConstrainedNumeric.h
```diff
@@ -406,6 +408,34 @@ namespace ome
...
+          operator++ ()
+          {
+            ++this->value;
+            check();
+            return *this;
+          }
+
+          /**
+           * Decrement the constrained value by one.  If the new value
+           * fails the constraint check, this will cause an error.
```
mtbc: The error-causing is checked by unit tests?

mtbc:
Manipulating XML using Xerces in C++ looks painful.

cpp/lib/ome/xml/model/primitives/ConstrainedNumeric.h
rleigh-dundee: Yes, in cpp/test/ome-xml/* in this PR.

mtbc: Ha, I obviously missed it, thank you!

rleigh-dundee: It's all the tables of inputs, outputs and whether exceptions are thrown or not.

cpp/lib/ome/bioformats/MetadataTools.cpp
rleigh-dundee: Yes, and I'd like to change it at some point.  However, it's a copy of the logic in the xsd-fu templates-java/OMEXMLModelObject.template for a custom derived model object, so I'd rather change it across the board rather than making it inconsistent in one place.  Happy to do that here if @melissalinkert is OK with updating the xsd-fu templates as well.

rleigh-dundee:
@mtbc Your suggestions should all be implemented now except for the one noted above.

mtbc:
Thank you, the latest commits look good.

joshmoore:
Thanks. Merging (also based on https://ci.openmicroscopy.org/pview/?match=.*5.1.*cpp.*&json=%7B%22match%22%3A+%22.*5.1.*cpp.*%22%7D) Following PRs can capture any of the TODOs from here.

PR #1563, opened 2015-01-29
Add version history for 5.0.7

melissalinkert:
As usual, the date probably needs updating once we're closer to tagging.

PR #1574, opened 2015-02-05
Add version history for 5.0.7 (rebased onto develop)

melissalinkert:
This is the same as gh-1563 but rebased onto develop.

----

As usual, the date probably needs updating once we're closer to tagging.


melissalinkert:
--rebased-from #1563

PR #1579, opened 2015-02-06
TillVision: throw a more informative exception when the .vws file is missing

melissalinkert:
This throws an exception indicating that the .vws file could not be found, instead of attempting to parse the .pst file as a .vws file.  .pst files cannot be read alone, as they are just pixels - an accompanying .vws file is expected and required.

/cc @joshmoore, @dietzc
See also https://github.com/scifio/scifio-bf-compat/issues/11


rleigh-dundee:
The change looks fine.  Please merge if green.

melissalinkert:
--no-rebase

PR #1580, opened 2015-02-06
ND2: fix RGB channel splitting

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/12725.  To test, verify that the file from QA 10425 has 10 Z sections and 3 channels, just like in NIS Elements Viewer (version 4.20).  For two of the channels, only one Z section will have non-zero data.

I would expect all builds to remain green.


rleigh-dundee:
Behaves exactly as described and the metadata looks fine.  Please merge if green.

melissalinkert:
--no-rebase

PR #1582, opened 2015-02-07
Fix Python files to be PEP8 1.6.0 compliant

sbesson:
With the latest release of PEP8, `import` needs to be at top-level. The number of Python files is reasonable in the Bio-Formats code base so we can open a quick fix PR. Arguably we could also impose the `pep8` version used by Travis.

To test this PR, check Travis remains green.
--no-rebase


sbesson:
Merging this to prevent Travis failures in subsequent builds.

PR #1577, opened 2015-02-06
Add version history for 5.0.8

sbesson:
Stub version history like we did for 4.4.8

PR #1583, opened 2015-02-09
Add version history for 5.0.8 (rebased onto develop)

sbesson:
This is the same as gh-1577 but rebased onto develop.

----

Stub version history like we did for 4.4.8


sbesson:
--rebased-from #1577

hflynn:
Building on staging now

hflynn:
Good to merge.

PR #1588, opened 2015-02-10
Removing all references to OME FAQs

hflynn:
See https://trello.com/c/3uqvWhlq/240-faq-issues-needing-moved-to-docs-plone - The FAQs are being deleted from the website for 5.1 as part of a general clean-up.

I grepped so hopefully haven't missed any links in the rest of the codebase.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -884,9 +884,8 @@ public abstract class FormatReader extends FormatHandler
       throw new FormatException("Image plane too large. Only 2GB of data can " +
         "be extracted at one time. You can workaround the problem by opening " +
         "the plane in tiles; for further details, see: " +
-        "http://www.openmicroscopy.org/site/support/faq/bio-formats/" +
-        "i-see-an-outofmemory-or-negativearraysize-error-message-when-" +
-        "attempting-to-open-an-svs-or-jpeg-2000-file.-what-does-this-mean", e);
+        "http://www.openmicroscopy.org/site/support/bio-formats/about/" +
+        "bug-reporting.html", e);
     }
     return openBytes(no, newBuffer, x, y, w, h);
   }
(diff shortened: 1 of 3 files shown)


components/formats-api/src/loci/formats/FormatReader.java
```diff
@@ -884,9 +884,8 @@ public boolean isInterleaved(int subC) {
       throw new FormatException("Image plane too large. Only 2GB of data can " +
         "be extracted at one time. You can workaround the problem by opening " +
         "the plane in tiles; for further details, see: " +
-        "http://www.openmicroscopy.org/site/support/faq/bio-formats/" +
-        "i-see-an-outofmemory-or-negativearraysize-error-message-when-" +
-        "attempting-to-open-an-svs-or-jpeg-2000-file.-what-does-this-mean", e);
+        "http://www.openmicroscopy.org/site/support/bio-formats/about/" +
+        "bug-reporting.html", e);
```
melissalinkert: My only worry is that this isn't a bug, and doesn't need to be reported to us - it only happens when the user does something that just isn't going to work.  Is the FAQ text going to be put somewhere else?

hflynn: The new improved bug reporting page has a 'common issues to check' section with the FAQ text on it -https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/about/bug-reporting.html

hflynn: I could add the anchor link for the section if you think that'd be better?

melissalinkert: Ah, OK, that makes sense.  Adding the anchor might be nice to help reduce any confusion.

hflynn:
Done. This will show as a broken anchor in the main Jenkins build until the m4 docs are released though.

ximenesuk:
I'll leave this for @melissalinkert to sign off the final commit.

hflynn:
Debating whether to backport this and the required doc changes. Current plan would be to remove the FAQs from the menu to start with so individual links from old docs would still work anyway...

hflynn:
--no-rebase *IF* we end up with a 5.0.9 I might rebase this after all though.

PR #1592, opened 2015-02-11
Leica SCN: fix pyramids when multiple channels are present

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/12727.

To test, import the file from QA 10469 into OMERO.  There should be 13 pyramids; the first 7 should have C C, 1 Z, 1 T, and 3 resolutions each.  The last 6 should have 2 C, 1 Z, 1 T, and 4 resolutions each.

I would expect all builds to remain green.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
@@ -111,6 +111,7 @@ public class LeicaSCNReader extends BaseTiffReader {
             return true;
           }
           catch (Exception se) {
+            LOGGER.debug("XML parsing failed", se);
           }
         }
       }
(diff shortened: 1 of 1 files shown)


dominikl:
I can open the image with bioformats, import to ImageJ works, also import to Omero works. I just don't know how to verify the pyramids generation.

dominikl:
Could confirm that the last 6 images have 4 zoom levels in Insight fullviewer. But the first images have 12 levels (from 25 to 300% in 25% steps), but I guess it's because Insight doesn't perform a tiled loading on this images and maybe therefore doesn't use the pyramids?
Anyway, I think the PR is good to merge.

melissalinkert:
--no-rebase

PR #1595, opened 2015-02-13
Gatan 32

jburel:
Problem reported by @bramalingam 
Fix issues with metadata reading 
To test: use ```from_skyking/gatan/chip/Nanogold-1.dm3 ```

Metadata to check using either  ```showinf``` or ImageJ

Emission Current (�A): 0.0 before
Emission Current (µA): 0.0 after
Units: �m before
Units: µm after

Using ```showinf``` make sure you do no have displayed 
Populating metadata
Not adjusting for unknown units: µm
Not adjusting for unknown units: µm

To check that the values are correctly read you can use the ImageJ plugin http://rsb.info.nih.gov/ij/plugins/DM3_Reader.html (Thanks @melissalinkert for tip)
This should already be installed under Plugins>Input-Output

Other readers might be affected but we will need to find good testing files

### DIFF
--- a/components/formats-common/src/loci/common/RandomAccessInputStream.java
+++ b/components/formats-common/src/loci/common/RandomAccessInputStream.java
@@ -526,20 +526,9 @@ public class RandomAccessInputStream extends InputStream implements DataInput, C
   public String readString(int n) throws IOException {
     int avail = available();
     if (n > avail) n = avail;
-    byte[] bytes = new byte[n];
-    readFully(bytes);
-    StringBuffer newString = new StringBuffer();
-    for (byte b : bytes) {
-        int v = b & 0xff;
-        if (v > 0x7f) {
-          newString.append(Character.toChars(v));
-        }
-        else {
-          newString.append((char) b);
-        }
-      }
-    String s = newString.toString();
-    return new String(s.getBytes(encoding), encoding);
+    byte[] b = new byte[n];
+    readFully(b);
+    return new String(b, encoding);
   }
 
   /** Read eight input bytes and return a long value. */
--- a/components/formats-common/src/loci/common/RandomAccessInputStream.java
+++ b/components/formats-common/src/loci/common/RandomAccessInputStream.java
@@ -529,21 +529,21 @@ public class RandomAccessInputStream extends InputStream implements DataInput, C
    * @throws IOException Thrown if an error occurred while reading the data.
    */
   public String readByteToString(int n) throws IOException {
-      n = (int) Math.min(available(), n);
-      byte[] bytes = new byte[n];
-      readFully(bytes);
-      StringBuffer newString = new StringBuffer();
-      for (byte b : bytes) {
-          int v = b & 0xff;
-          if (v > 0x7f) {
-              newString.append(Character.toChars(v));
-          }
-          else {
-              newString.append((char) b);
-          }
-      }
-      String s = newString.toString();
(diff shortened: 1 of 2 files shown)


components/formats-gpl/src/loci/formats/in/GatanReader.java
```diff
@@ -378,7 +378,7 @@ else if (n == 3) {
             }
             else {
               if (dataType == 10) in.skipBytes(length);
-              else value = DataTools.stripString(in.readString(length * 2));
+              else value = in.readString(length * 2);
```
melissalinkert: Sorry, maybe I don't follow - if this reader has a readString(int) method that mostly works, could it not be used here instead of changing the RandomAccessInputStream.readString implementation?

jburel: That's the point I wanted to ask you actually since I did not want to affect other reader. 
I will review that section of the code.

melissalinkert:
Excluding pending further changes, as I think this is what's causing the couple of test failures in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-build/687.  Will remove the exclude label after the RandomAccessInputStream changes we discussed this afternoon are finished.

jburel:
Removing label, changes done.

jburel:
Re-adding exclude label. Failure importing gatan files. This needs further investigation.

jburel:
Tests should now pass

jburel:
Removing label, test should pass now

jburel:
Follow a list of readers with possibly similar issues: cc @pwalczysko 
TiffReader, CellSensReader, FV1000Reader, LIFReader, LeicaReader, NativeND2Reader, ZeissLSMReader, ZeissZVIReader

not to be covered by this PR.

jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-test_images_good/411/ is now green

components/formats-common/src/loci/common/RandomAccessInputStream.java
```diff
@@ -522,6 +520,32 @@ public String readCString() throws IOException {
...
+  /**
+   * Reads a byte array of the given length byte by byte. Returns a string
+   * using the set encoding.
+   *
+   * @param n The length of the array.
+   * @return See above
+   * @throws IOException Thrown if an error occurred while reading the data.
+   */
+  public String readByteToString(int n) throws IOException {
+      n = (int) Math.min(available(), n);
```
joshmoore: note the change in the indentation.

pwalczysko:
Testing:
Prior to this PR:
```
Populating metadata
Not adjusting for unknown units: �m
Not adjusting for unknown units: �m
Not adjusting for unknown units: 
...
Units: �m
```

After this PR:
```
...
Not adjusting for unknown units.
...
Units: µm
```

Also in ImageJ the display is as expected with this PR in.

The gatan image looks exactly the same in IJ as with the plugin http://rsb.info.nih.gov/ij/plugins/DM3_Reader.html.

All good here.

joshmoore:
Talking to @jburel this is now blocking. Merging (based on the fact that https://github.com/jburel/bioformats/commit/df74462cd855afea76a8181af3a72580a668c3a5?diff=unified&w=1 is empty). There are still some whitespace issues, but those can be handled in the follow-up.

jburel:
--no-rebase

PR #1596, opened 2015-02-13
J2K image reader memory leak fix from jai-imageio-core

mtbc:
see https://github.com/jai-imageio/jai-imageio-core/pull/2

--no-rebase


melissalinkert:
:+1:.  Do you know if ```JAIImageIOServiceImpl``` needs to be updated to call ```dispose()``` on the readers it creates, or if this method is expected to be called automatically?

mtbc:
It looks to me as though `TIFFJPEGDecompressor.finalize()` calls it, but I don't know if that suffices and I haven't actually tried setting a breakpoint.

melissalinkert:
Looks like ```dispose()``` isn't called, and ```JAIImageIOServiceImpl``` will need a little reworking to make it so.  That doesn't block this PR though - I will open a separate PR later today or tomorrow.

PR #1599, opened 2015-02-16
Fix Python files for PEP8 1.6.2

sbesson:
This should fix all W503 line break before binary operator warnings

--no-rebase


sbesson:
Merging to fix Travis.

PR #1585, opened 2015-02-09
Updating DeltaVision link

hflynn:
See https://github.com/openmicroscopy/bioformats/issues/1581

This is the UK redirect with prettier graphics 

I've also removed the extra link as the main one goes to the equivalent page.

Issue #1581, opened 2015-02-06
DeltaVision website links, no longer api.com, rather redirects/moved to GE website 


chalkie666:
DeltaVision website links, no longer api.com, rather moved to GE website 
currenty
http://www.gelifesciences.com/webapp/wcs/stores/servlet/catalog/en/GELifeSciences-de/products/AlternativeProductStructure_23807/
so bioformats info pages7files about DV file format might like to be updated with that info
api.com now redirects there
cheers
Dan

hflynn:
Thanks for reporting this, I'll update the links this week.

hflynn:
PR opened, whoever has permission can close this now. Thanks again @chalkie666

PR #1602, opened 2015-02-17
Updating DeltaVision link (rebased onto dev_5_0)

hflynn:
This is the same as gh-1585 but rebased onto dev_5_0.

----

See https://github.com/openmicroscopy/bioformats/issues/1581

This is the UK redirect with prettier graphics 

I've also removed the extra link as the main one goes to the equivalent page.


hflynn:
--rebased-from #1585

PR #1609, opened 2015-02-18
cpp: compat: Don't import types into std namespace

rleigh-dundee:
This is a basic search and replace of:

`std::shared_ptr` → `ome::compat::shared_ptr`
`std::tuple` → `ome::compat::tuple`
`std::regex` → `ome::compat::regex`

(plus associated types and functions like `weak_ptr`, `make_shared`, `regex_search` etc.)

The reason for this is that we were abusing the `std` namespace by importing `boost`/`std::tr1` types into it, since this could potentially cause problems given that we are exposing "`std`" types in our interfaces which don't actually exist.  If an end user uses these they will cause problems when they upgrade and the `std` types actually do exist.  With this change, all types are imported into the `ome::compat` namespace, be they in `std`, `std::tr1` or `boost`.  This makes our API independent of any system changes, and will be safe under all circumstances.  Note that the types will still end up in the ABI using their original namespaces, but the API will be safe.  See https://trac.openmicroscopy.org/ome/ticket/12174 for further details.

--no-rebase

--------

Testing: if it builds and is green, it's OK; any mistakes would result in build failure.  Note documentation updates in addition to code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - BUILD=maven
   - BUILD=cpp
   - BUILD=cppwrap
+  - BUILD=flake8
   - BUILD=sphinx
   - BUILD=ant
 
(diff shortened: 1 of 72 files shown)


joshmoore:
Few travis failures were due to maven timeouts, but there's also this:
```
./components/xsd-fu/python/ome/modeltools/property.py:484:80: E501 line too long (84 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:499:80: E501 line too long (84 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:504:80: E501 line too long (84 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:571:80: E501 line too long (81 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:603:80: E501 line too long (84 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:609:80: E501 line too long (80 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:618:80: E501 line too long (84 > 79 characters)
./components/xsd-fu/python/ome/modeltools/property.py:625:80: E501 line too long (84 > 79 characters)
```

rleigh-dundee:
Right, I'll look at that.  I'm unconvinced that this particular check has any real value--given the long type names, splitting it over multiple lines makes things *less* readable and maintainable.

joshmoore:
> *less* readable...

We have a style specification so we don't have to discuss these issues (especially not in arbitrary PRs)

rleigh-dundee:
@joshmoore Where should this be mentioned then?  This has been a time-wasting pain on several of the PRs of the last month.  It adds *zero* value.

rleigh-dundee:
Travis is now green and jenkins is also green: http://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/594/

PR #1606, opened 2015-02-18
Updating ZEN Zeiss link

hflynn:
See http://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/697/warnings3Result/? - ZEN links no longer have 2012 in them.

PR #1611, opened 2015-02-19
Updating ZEN Zeiss link (rebased onto dev_5_0)

hflynn:
This is the same as gh-1606 but rebased onto dev_5_0.

----

See http://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/697/warnings3Result/? - ZEN links no longer have 2012 in them.


hflynn:
--rebased-from #1606

PR #1612, opened 2015-02-19
Base inch on metre

qidane:
Remove the imperial inch as a base, construct it from a metre instead. The conversion code is just more complex and it does not provide an increase in accuracy.

### DIFF
--- a/components/ome-xml/src/ome/units/UNITS.java
+++ b/components/ome-xml/src/ome/units/UNITS.java
@@ -66,7 +66,7 @@ public final class UNITS
   public static final ome.units.unit.Unit<ElectricPotential> VOLT    = ome.units.unit.Unit.<ElectricPotential>CreateBaseUnit("SI.VOLT", "V");
   public static final ome.units.unit.Unit<Frequency>         HERTZ   = ome.units.unit.Unit.<Frequency>CreateBaseUnit("SI.HERTZ", "Hz");
   public static final ome.units.unit.Unit<Length>            METRE   = ome.units.unit.Unit.<Length>CreateBaseUnit("SI.METRE", "m");
-  public static final ome.units.unit.Unit<Length>            INCH    = ome.units.unit.Unit.<Length>CreateBaseUnit("Imperial.INCH", "in");
+  public static final ome.units.unit.Unit<Length>            INCH    = METRE.multiply(0.0254).setSymbol("in");
   public static final ome.units.unit.Unit<Power>             WATT    = ome.units.unit.Unit.<Power>CreateBaseUnit("SI.WATT", "W");
   public static final ome.units.unit.Unit<Pressure>          PASCAL  = ome.units.unit.Unit.<Pressure>CreateBaseUnit("SI.PASCAL", "Pa");
   public static final ome.units.unit.Unit<Pressure>          BAR     = PASCAL.multiply(100000).setSymbol("bar");


components/ome-xml/src/ome/units/UNITS.java
```diff
@@ -66,7 +66,7 @@ public String getName()
   public static final ome.units.unit.Unit<ElectricPotential> VOLT    = ome.units.unit.Unit.<ElectricPotential>CreateBaseUnit("SI.VOLT", "V");
   public static final ome.units.unit.Unit<Frequency>         HERTZ   = ome.units.unit.Unit.<Frequency>CreateBaseUnit("SI.HERTZ", "Hz");
   public static final ome.units.unit.Unit<Length>            METRE   = ome.units.unit.Unit.<Length>CreateBaseUnit("SI.METRE", "m");
-  public static final ome.units.unit.Unit<Length>            INCH    = ome.units.unit.Unit.<Length>CreateBaseUnit("Imperial.INCH", "in");
+  public static final ome.units.unit.Unit<Length>            INCH    = METRE.multiply(0.0254).setSymbol("in");
```
joshmoore: Does this need a non-SI system set?

qidane: We don't have a flag to say whether or not a a unit is SI, might be good to add one, but would be a lot of changes.

melissalinkert: Maybe I'm missing something, but wouldn't adding a ```setMeasurementSystem(String)``` method to ```ome.units.unit.Unit``` (similar to ```setSymbol(String)```) be enough?  Then you could do ```METRE.multiply(0.0254).setMeasurementSystem("Imperial").setSymbol("in")```, and I wouldn't think any other changes would be necessary.

qidane: MeasurementSystem has to be immutable. It is what governs whether units are convertible.  If you did `setMeasurementSystem("Imperial")` the resulting object would have the same problem as the one created by `CreateBaseUnit("Imperial.INCH", "in");` so you would be back to writing custom code to handle metric<->imperial in `Length.java`.

melissalinkert: OK, that makes sense.  I think I'm still misunderstanding something though, as it doesn't seem terribly difficult to make metric/imperial conversion work just for Length.  If ```isConvertible``` is overridden to allow conversion between systems, and ```convertValue``` is overridden to:

1. perform a conversion in the original measurement system, to get either the metre value (if the original system is metric) or the inch value (if the original system is imperial).
2. perform the metre <--> inch conversion as defined in this PR, and change measurement systems
3. perform a conversion within the new measurement system to get the desired output unit

then I would think the inter-system conversion would work with minimal effort.

I guess that's being slightly nitpicky though - what is in the PR works, we just lose the ability to distinguish between the actual measurement systems for length units.  If no one else feels strongly that we need that, happy to merge as-is.

qidane: My objection is not with the complexity of the code but with having the conversion code in more than one place. At the moment all the conversion for this java library (OMERO + XSLT are different) is defined in the single file `UNITS.java`. There is no special casing in any of the quantities. I think this solution neater.

melissalinkert:
OK, fair enough - merging in that case.

qidane:
--no-rebase

PR #1613, opened 2015-02-19
Add generate-sources target

sbesson:
See https://trello.com/c/ZC4Zm0BL/197-autogenerated-code-diffs

This PR is adding a new ant target allowing to commit generated sources. This top-level target copies the source-generated files to a `generated-sources` folder that can be put under version control. The idea is to have this new folder committed and pushed on a branch of @snoopycrimecop as part of latest and merge builds (like `develop/merge/autogen`, `develop/latest/autogen` for the auto-generated documentation). Comparing these branches should allow to compare the generated source changes with/without PRs.

Alternatively, this changes could be pushed to a separate repository if that makes more sense.

/cc @rleigh-dundee @qidane @joshmoore

### DIFF
--- a/ant/toplevel.xml
+++ b/ant/toplevel.xml
@@ -125,6 +125,7 @@ Type "ant -p" for a list of targets.
     description="remove build files for all components">
     <delete dir="${artifact.dir}"/>
     <delete dir="build"/>
+    <delete dir="generated-sources"/>
   </target>
 
   <target name="test"
@@ -957,6 +958,16 @@ Type "ant -p" for a list of targets.
   <target name="clean-xsd-fu"
     description="remove build files for xsd-fu">
     <ant dir="components/xsd-fu" target="xsd-fu.clean"/>
+  </target>
+
+  <target name="generate-sources"
+    description="generate sources using xsd-fu"
+    depends="ome-xml-src, metadata-src">
+    <mkdir dir="generated-sources"/>
+	<copy todir="generated-sources">
+	    <fileset dir="components/ome-xml/target/generated-sources"/>
+	</copy>
+    <ant dir="components/ome-xml" target="generate-source"/>
   </target>
 
   <target name="ome-xml-src"


ant/toplevel.xml
```diff
@@ -959,6 +960,16 @@ Type "ant -p" for a list of targets.
...
   </target>
 
+  <target name="generate-sources"
+    description="generate sources using xsd-fu"
+    depends="ome-xml-src, metadata-src">
+    <mkdir dir="generated-sources"/>
+	<copy todir="generated-sources">
+	    <fileset dir="components/ome-xml/target/generated-sources"/>
+	</copy>
+    <ant dir="components/ome-xml" target="generate-source"/>
```
rleigh-dundee: Isn't this ant call redundant since it's already run by the target dependencies?

ant/toplevel.xml
```diff
@@ -959,6 +960,16 @@ Type "ant -p" for a list of targets.
     <ant dir="components/xsd-fu" target="xsd-fu.clean"/>
   </target>
 
+  <target name="generate-sources"
```
rleigh-dundee: Very minor, but it's a plural while the existing targets are singular "source".

rleigh-dundee:
Overall, looks OK.  Any reason it needs to be at the top-level?  I mean, you could just call the existing targets and then add and commit components/ome-xml/target/generated-sources without needing to make any changes.

We could probably also remove the metadata-src target now it's in ome-xml.  Note that it's a no-op already.

sbesson:
Assuming `metadata-src` is a no-op, then I agree this is totally overkill and unnecessary. I can close this PR and go with your strategy. The reason for not committing `components/ome-xml/target/generated-sources` directly is that `target` is excluded via `.gitignore` but I can easily move this folder elsewhere.
In general, what are your thoughts about this strategy: single commit with generated-sources on top of a branch?

qidane:
I would like to the committed generated sources all in one place not spread about. If it is not going to be used in the regular compilation I think it should be in a folder with a name warning you of the fact.

sbesson:
@qidane: if we want to prevent duplication, we can certainly move towards a complete standalone repository e.g. `openmicroscopy/bio-formats-generated-sources`.

rleigh-dundee:
@sbesson See https://github.com/openmicroscopy/bioformats/pull/1615 for removal of metadata-src

rleigh-dundee:
@sbesson WRT the strategy, this is limited by github's compare view not being able to restrict the files/dirs being shows like git diff can.

I think I would do this:

```
create original branch
generate sources
add generated sources
commit geneated sources
scc merge PR branches
---
create new branch from original branch
git rm generated sources
generate sources
add generated sources
commit generated sources
```

The diff between the original and new branch will now contain only the generated code changes.

PR #1614, opened 2015-02-19
cpp: Add basic logging support

rleigh-dundee:
Initial support for logging of warnings, errors, diagnostics, debug info etc.

Initial coverage is confined to code previously using `std::clog`; additional logging will be added in the future.

Initially, per-class loggers are using only log severity for log filtering.  There is no custom log formatter to log the class name yet, nor a channel name to restrict/select logging domains.  These can be added in future PRs.  The aim of this PR is to get the basic logging infrastructure in place, and the existing debug messages converted to use the logging facilities.

--no-rebase

--------

Testing:

Run

```
./bf cpp/bin/showinf/showinf /path/to/2013-06-image.ome.tiff [--debug]
```

When `--debug` is added, you should get logging of all debug messages to the console.  The OME-TIFF reader should log a few as it parses the TiffData blocks.  If `--debug` is absent, these debug messages should be missing.  You might get warning messages if there are any problems with the OME-XML; these are logged in both cases; they should be suppressed with `--quiet` (but don't bother testing this; you'll need an invalid OME-TIFF and the options are likely to be adjusted anyway to better fit the log levels provided by the logging library when custom log formatters are added).  The main point to test is that `--debug` works correctly, which will validate that all the logging code is working as expected, and that the log severity can be set to filter out unwanted messages.


melissalinkert:
Travis failure appears to be legitimate.

joshmoore:
Minor bike shedding: "log" seems more like `ome::util(s)` to me rather than `ome::compat`

emilroz:
That's the reason for the failure: 
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
Boost_LOG_LIBRARY_RELEASE (ADVANCED)
    linked by target "ome-compat" in directory /home/travis/build/openmicroscopy/bioformats/cpp/lib/ome/compat
-- Configuring incomplete, errors occurred!
```
I will check this PR locally. You will also need to add ```log_setup``` to the boost cmake components otherwise the linker will fail once you'll actually try to call it.

rleigh-dundee:
@joshmoore  Agreed that it's not strictly a compatibility function.  We can split it into a separate library, OTOH we don't really want an explosion of DLLs, so we do also need to constrain splitting every last bit out.  There are probably a few other bits from compat that could be moved into a utils library, but the lines are blurred for some cases.  filesystem.h?  mstream.h?  module.h?  At the moment log.h is a single inline helper function, so there's no content for a shared library.  We could have a base library which covers both the util and compat namespaces given their tiny size.  Either way, we certainly do need to decide where to draw the line between compatibility and utility.  Any thoughts?

WRT the travis failure, this is probably the version in use by travis not having the log library installed.  Might not have been packaged by Ubuntu.  If that's the case then we'll have to add fallbacks (no logging or log to stderr).  I'll look into this.

@emilroz Interesting, log_setup looks to be newish--but it's working without it for me on Linux with 1.55 and FreeBSD with 1.57 without using it!  Might need to add logic in the cmake checks to try to use it if available, if older versions lack it.

joshmoore:
> ...At the moment log.h is a single inline helper function, so there's no content for a shared library. We could have a base library which covers both the util and compat namespaces given their tiny size. Either way, we certainly do need to decide where to draw the line between compatibility and utility. Any thoughts?

I'm certainly happy for them to be in the same lib. What's the current list? (And what's the place to look to find that out?!)

rleigh-dundee:
@joshmoore Run `git grep add_library cpp/lib`.  The current list is (in dependency order): ome-compat ome-xerces ome-xml ome-bioformats ome-test

test is for unit tests only.  xerces could potentially be merged with compat/util since it's really compatibility wrappers and utility functions for using xerces with modern C++.  Currently there's a 1:1 mapping between cpp/lib/ome directory and the shared library name, i.e. cpp/lib/ome/xml is libome-xml.  But that's purely convention rather than requirement.  So compat/util/xerces could all potentially go in some base library.  The only proper (for the end user) libs we have are ome-xml and ome-bioformats.

emilroz:
All works as expected. 

Output:
```
EMil:build-cpp emilrozbicki$ cpp/bin/showinf/showinf /Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif 
Image: /Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif
[2015-02-20 10:40:06.752930] [0x000007fff7bf5730] [warning] Ignoring null UUID object when retrieving filename
[2015-02-20 10:40:06.753271] [0x000007fff7bf5730] [warning] Ignoring null UUID object when retrieving value
[2015-02-20 10:40:06.753438] [0x000007fff7bf5730] [warning] Image ID: Image:0 missing plane #1
Reader setup took 00:00:00.052107

Reading core metadata
Filename = "/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif"
Used files = ["/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif"]
Series count = 1

Series #0:
	Image count = 15
	RGB = false (1) 
	Interleaved = false
	Indexed = false
	Width = 439
	Height = 167
	SizeZ = 5 (effectively 5)
	SizeT = 1 (effectively 1)
	SizeC = 3 (effectively 3)
	Thumbnail size = 128 × 48
	Endianness = little
	DimensionOrder = XYZCT (certain)
	PixelType = uint8
	Bits per Pixel = 8
	MetadataComplete = true
	ThumbnailSeries = false

No global metadata
```

With ```quiet```
```
EMil:build-cpp emilrozbicki$ cpp/bin/showinf/showinf --quiet /Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif 
Image: /Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif
Reader setup took 00:00:00.062663

Reading core metadata
Filename = "/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif"
Used files = ["/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif"]
Series count = 1

Series #0:
	Image count = 15
	RGB = false (1) 
	Interleaved = false
	Indexed = false
	Width = 439
	Height = 167
	SizeZ = 5 (effectively 5)
	SizeT = 1 (effectively 1)
	SizeC = 3 (effectively 3)
	Thumbnail size = 128 × 48
	Endianness = little
	DimensionOrder = XYZCT (certain)
	PixelType = uint8
	Bits per Pixel = 8
	MetadataComplete = true
	ThumbnailSeries = false

No global metadata
```

With ```--debug```:
```
EMil:build-cpp emilrozbicki$ cpp/bin/showinf/showinf --debug /Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif 
Image: /Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif
[2015-02-20 10:52:27.289517] [0x000007fff7bf5730] [debug]   Image[0] {
[2015-02-20 10:52:27.289751] [0x000007fff7bf5730] [debug]     id = Image:0
[2015-02-20 10:52:27.289942] [0x000007fff7bf5730] [debug]     TiffData[0] {
[2015-02-20 10:52:27.290033] [0x000007fff7bf5730] [warning] Ignoring null UUID object when retrieving filename
[2015-02-20 10:52:27.290076] [0x000007fff7bf5730] [warning] Ignoring null UUID object when retrieving value
[2015-02-20 10:52:27.290145] [0x000007fff7bf5730] [debug]       Plane[0]: file=/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif, IFD=0
[2015-02-20 10:52:27.290169] [0x000007fff7bf5730] [debug]     }
[2015-02-20 10:52:27.290197] [0x000007fff7bf5730] [debug]       Plane[1]: CLEARED
[2015-02-20 10:52:27.290214] [0x000007fff7bf5730] [debug]       Plane[2]: CLEARED
[2015-02-20 10:52:27.290232] [0x000007fff7bf5730] [debug]       Plane[3]: CLEARED
[2015-02-20 10:52:27.290248] [0x000007fff7bf5730] [debug]       Plane[4]: CLEARED
[2015-02-20 10:52:27.290264] [0x000007fff7bf5730] [debug]       Plane[5]: CLEARED
[2015-02-20 10:52:27.290281] [0x000007fff7bf5730] [debug]       Plane[6]: CLEARED
[2015-02-20 10:52:27.290298] [0x000007fff7bf5730] [debug]       Plane[7]: CLEARED
[2015-02-20 10:52:27.290314] [0x000007fff7bf5730] [debug]       Plane[8]: CLEARED
[2015-02-20 10:52:27.290330] [0x000007fff7bf5730] [debug]       Plane[9]: CLEARED
[2015-02-20 10:52:27.290346] [0x000007fff7bf5730] [debug]       Plane[10]: CLEARED
[2015-02-20 10:52:27.290363] [0x000007fff7bf5730] [debug]       Plane[11]: CLEARED
[2015-02-20 10:52:27.290379] [0x000007fff7bf5730] [debug]       Plane[12]: CLEARED
[2015-02-20 10:52:27.290396] [0x000007fff7bf5730] [debug]       Plane[13]: CLEARED
[2015-02-20 10:52:27.290414] [0x000007fff7bf5730] [debug]       Plane[14]: CLEARED
[2015-02-20 10:52:27.290433] [0x000007fff7bf5730] [debug]     Verify Plane[0]: file=/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif, IFD=0
[2015-02-20 10:52:27.290453] [0x000007fff7bf5730] [debug]     Verify Plane[1]: file=, IFD=0
[2015-02-20 10:52:27.290470] [0x000007fff7bf5730] [warning] Image ID: Image:0 missing plane #1
[2015-02-20 10:52:27.291083] [0x000007fff7bf5730] [debug]   }
Reader setup took 00:00:00.035311

Reading core metadata
Filename = "/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif"
Used files = ["/Volumes/ome/data_repo/test_images_good/ome-tiff/multi-channel-z-series.ome.tif"]
Series count = 1

Series #0:
	Image count = 15
	RGB = false (1) 
	Interleaved = false
	Indexed = false
	Width = 439
	Height = 167
	SizeZ = 5 (effectively 5)
	SizeT = 1 (effectively 1)
	SizeC = 3 (effectively 3)
	Thumbnail size = 128 × 48
	Endianness = little
	DimensionOrder = XYZCT (certain)
	PixelType = uint8
	Bits per Pixel = 8
	MetadataComplete = true
	ThumbnailSeries = false

No global metadata
```

rleigh-dundee:
So the travis error is due to Boost.Log being added in version 1.54; travis is using 1.48.  Tried building 1.54 and 1.57 versions of Boost.Log on Ubuntu 12.04 but this is missing needed dependencies.  I'll see how feasible it is to remove these bits, so we can embed the needed headers as a fallback for older systems.

joshmoore:
> So compat/util/xerces could all potentially go in some base library. The only proper (for the end user) libs we have are ome-xml and ome-bioformats.

Gotcha. To stay with the Java version, perhaps "common" rather than "util" when the time comes.

rleigh-dundee:
@joshmoore Not sure either way yet, but certainly a possibility.  In the java case we have common/api/bsd/gpl for formats, but this may be still lower-level than what's in the common package for java, so that it might make sense to have a "base" (or similarly named) library in addition to a common-equivalent.  We might also want to split up what's currently in ome-bioformats at some point, but I don't think there's a need for that for the near future.  Would it be OK to defer this for a future PR and keep logging in compat for simplicity here, and then move it and other parts out into a utility library once we've decided on how we want things organised?

joshmoore:
> Would it be OK to defer this for a future PR and keep logging in compat for simplicity here, and then move it and other parts out into a utility library once we've decided on how we want things organised?

Sorry, yes, of course. That's what I meant by "when the time comes".

rleigh-dundee:
Due to travis (Boost 1.48) lacking Boost.Log, we needed to be able to build on this version with some form of logging.  I tried backporting the original Boost.Log from Boost 1.54 in two different ways, but failed despite getting close on the second attempt; the problem is its heavy use of C++11 compatibility code for variadic macros, unicode string types, rvalues, stringrefs and other features which would also require backporting or excising the portability code from newer Boost versions as well.  This proved to be quite challenging, though I did get most of the way there.  As an alternative, I've implemented the minimal needed parts of the Boost.Log API with a couple of trivial classes to duplicate the basic logging functionality.  It can only log to `stderr` (`std::clog`), but it does support basic message severites and per-class loggers.  As we use more features of Boost.Log, we'll need to keep the portable implementation up to date as well.  Hopefully this can be dropped once Travis updates to 14.04 and/or we cease supporting 12.04.

mtbc:
With Boost 1.55 I do indeed get some extra `[debug]` lines among the usual `[warning]` iff I append `--debug`. And, `--quiet` does get rid of the `[warning]` lines.

mtbc:
Is there a machine (or vbox) with C++ prerequisites installed with which I can try Boost <1.54? I can create an Ubuntu 13.10 VM or somesuch, but I found my existing one inadequate for BF C++ in only having 8GB for the root partition. (Things like texlive are pretty enormous, it turns out.) Debian-side, squeeze offers Boost 1.42, so too old.  Unless @emilroz can give it another go ... ?

mtbc:
@rleigh-dundee: Thank you very much for the VM with Boost 1.48! Everything works exactly as in https://github.com/openmicroscopy/bioformats/pull/1614#issuecomment-75747809 so **good to merge**.

melissalinkert:
https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-cpp/604/ is still red - are further fixes needed here?

rleigh-dundee:
Yes, looks like two minor portability issues.  Missing iostream include for gretzky and something else for perch.

rleigh-dundee:
This will likely need some additional commits.  It looks like the use of static loggers in the classes is problematic for some combination of boost version/debug|release/c++98|c++11 mode which is the result of the perch CI failure.  Manual testing on Linux shows the same problem.  This is due to some error during logger destruction at program exit; it's possibly a boost bug but might also be due to how we are using the static globals; it's documented not to use them during cleanup.  It might be necessary to switch to using per-object instances rather than per-class.  There are macros to register global loggers as singletons; it's also possible using the macros might remove the cause of the fault.

rleigh-dundee:
Switched from class to instance variables for loggers.  With class-based loggers some platforms were having segfaults at program exit.  This way is more reliable (I hope) but wastes more memory, particularly for the model objects.  I'd like to improve this in the future, e.g. by having one logger per model object tree; but that requires a little refactoring.

joshmoore:
```
/home/travis/build/openmicroscopy/bioformats/cpp/build/cpp/lib/ome/xml/model/OME.cpp: In constructor ‘ome::xml::model::OME::OME()’:
/home/travis/build/openmicroscopy/bioformats/cpp/build/cpp/lib/ome/xml/model/OME.cpp:270:16: error: ‘class ome::compat::Logger’ has no member named ‘add_attribute’
/home/travis/build/openmicroscopy/bioformats/cpp/build/cpp/lib/ome/xml/model/OME.cpp:270:52: error: ‘ome::logging::attributes’ has not been declared
/home/travis/build/openmicroscopy/bioformats/cpp/build/cpp/lib/ome/xml/model/OME.cpp:270:84: error: expected primary-expression before ‘>’ token
make[2]: *** [cpp/lib/ome/xml/CMakeFiles/ome-xml.dir/model/OME.cpp.o] Error 1
make[1]: *** [cpp/lib/ome/xml/CMakeFiles/ome-xml.dir/all] Error 2
make: *** [all] Error 2
```

rleigh-dundee:
It needs wrapping in an ifdef.

mtbc:
What further testing is needed here? Is CI coverage sufficient or should certain platforms / library versions get a(nother) manual review?

rleigh-dundee:
@mtbc The CI tests should cover the issue of segfaulting at exit.  Other than that, a quick check that `--debug` still gives the same output should be sufficient.

rleigh-dundee:
Latest CI failure on perch was reproducible in a VM before upgrading cmake, so I'll upgrade it and see if it helps.

mtbc:
If I recall correctly, on Debian wheezy I had to bump cmake up to jessie's to get it to pass.

rleigh-dundee:
@mtbc That's most likely a bug, all 2.8.x releases should work, though this version isn't covered by our CI testing (2.8.11 on gretzky35 and 3.1.3 on everything else).  The cmake in wheezy *should* be supported; though most of my testing is on 2.8.12 and 3.1.x.  For the BSD case, it's probably a bug in cmake itself--I updated it from 3.1.1 to 3.1.3; likely a regression in variable expansion/conditionals which was fixed in the point release.  We could look at what's failing if you wanted to downgrade at some point.

mtbc:
@rleigh-dundee: Have downgraded and sent the relevant logs by e-mail.

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/617/

PR #1621, opened 2015-02-23
Add tiling options to bfconvert

melissalinkert:
Fixes http://trac.openmicroscopy.org/ome/ticket/10292.

There are documentation updates as part of the PR that summarize what the new options are.  To further summarize, the tile size used during conversion can be specified explicitly:

```bfconvert -tilex 256 -tiley 256 input-file output-file.tiff```

and each tile can be written to a separate file:

```bfconvert -tilex 256 -tiley 256 input-file output-file_X%x_Y%y.tiff```

where the ```-tilex``` and ```-tiley``` values can be changed as desired.  Note that this only works if the output format is TIFF at the moment, due to known difficulties with writing other formats tile-wise.

To test, pick one or two of your favorite files and run a few ```bfconvert``` commands using varying values for ```-tilex``` and ```-tiley```.  Also verify that writing each tile to a separate file results in readable files that look like the corresponding tile from the original image, and that the documentation page is clear enough.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -113,9 +113,163 @@ public final class ImageConverter {
   private IFormatReader reader;
   private MinMaxCalculator minMax;
 
+  private HashMap<String, Integer> nextOutputIndex = new HashMap<String, Integer>();
+
   // -- Constructor --
 
   private ImageConverter() { }
+
+  private boolean parseArgs(String[] args) {
+    if (args == null) {
+      return true;
+    }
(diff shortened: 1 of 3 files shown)


jburel:
Tested the 2 options with various tile size.
It works fine
When the tile size is not correct e.g. 600 600 for a 480x480 image or negative value
it will be good to indicate what was used in message displayed.
Imported some generated images in OMERO (trout user-2 datasets tools). No issue noticed.

hflynn:
Doc updates read and render fine

melissalinkert:
2f95c7f pushed to log the tile size that is actually used for each series.

jburel:
looks good.

melissalinkert:
--no-rebase

PR #1618, opened 2015-02-20
display OME-XML and rois.

jburel:
Add option to display OME-XML and rois.
See https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7746
This can be tested using fiji (you need to replace the bio-formats_plugins jar and other B-F jars ( if not already using develop jars i.e. jars/bioformats)
and running the following code
Please select a file with ROI, you can easily generate an ome-tiff with rois using BF- exporter
Edit pathToFile
```
from loci.common import Region
from loci.plugins.in import ImporterOptions
from loci.plugins import BF
options = ImporterOptions()
options.setShowROIs(True)
options.setShowOMEXML(True)
options.setColorMode(ImporterOptions.COLOR_MODE_COLORIZED)
options.setId(pathToFile)
imps = BF.openImagePlus(options)
for imp in imps:
   imp.show()
```

PR #1624, opened 2015-02-24
Exporter windowless

jburel:
Allow to save the images w/o passing all parameters
if no file specified,  a temporary file using the name of the image selected is created. We may want to change that and stop if no output specified.
To test: 
 * Open Fiji, update the b-f_plugins jar and all b-f jars if you are not already using develop jar (jars/bioformats)
 * Run the following code (new>script language>python)

```
from loci.plugins.in import ImporterOptions
from loci.plugins import BF
from ij import IJ
options = ImporterOptions()
options.setId(fileToView)
imps = BF.openImagePlus(options)
for imp in imps:
   imp.show()
   image = IJ.runPlugIn("loci.plugins.LociExporter", "outfile=fileToSaveTo.ome.tif windowless=true ")
```
Remove the outfile param to create the tmp file.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -200,7 +200,6 @@ public class Exporter {
                 outfile = new File(p, n).getAbsolutePath();
                 if (Recorder.record) Recorder.recordPath("outputfile", outfile);
                 IJ.log("exporter outputfile "+outfile);
-                
             } catch (Exception e) {
                 //fall back to window mode.
             } finally {
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -140,7 +140,6 @@ public class Exporter {
         String compression = null;
 
         Boolean windowless = Boolean.FALSE;
-        IJ.log(plugin.arg);
         if (plugin.arg != null) {
             outfile = Macro.getValue(plugin.arg, "outfile", null);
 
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works as expected. Not sure if the comment about the ``metadata-extractor`` from https://github.com/openmicroscopy/bioformats/pull/1618#issuecomment-75968229 is valid here as well - I had the jar in my Fiji for this testing, so it certainly works WITH it.

Tested images were 1 ome.tiff and 1 lsm file. In both cases the new image was saved as expected.

melissalinkert:
My only worry is that if ```outfile``` is not specified, the exported file path isn't going to be immediately obvious to the end user.  Creating a temp file is certainly fine, but if the path can be logged or returned somehow that might be nice.

jburel:
@melissalinkert: record the name using the recorder. Other option will be to stop the process if no output file specified

components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
```diff
@@ -153,15 +176,36 @@ public void run() {
...
             }
         }
-
+        //create a temporary file if window less
+        if (windowless && (outfile == null || outfile.length() == 0)) {
+            try {
+                String name = removeExtension(imp.getTitle());
+                String n = name+ ".ome.tif";
+                File tmp = File.createTempFile(name, ".ome.tif");
+                File p = tmp.getParentFile();
```
melissalinkert: This results in an extra empty file being present in /tmp - could this either be deleted upon exit, or replaced with ```System.getProperty("java.io.tmpdir")```?

melissalinkert:
Export works fine barring the minor nitpick above, but I'm still having trouble finding the logged file path if ```outfile``` is not set.  Even with the macro recorder running, I don't see the path being recorded anywhere.  Would using ```IJ.log``` instead of the macro recorder be an option?

Both are minor issues though, so I'm happy to merge if you would rather address them separately.

components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
jburel: oops forgot to add delete. It was in the initial code.

jburel:
Will add path to log, might be easier.
I will do that in this PR

jburel:
@melissalinkert: done

melissalinkert:
Thanks, all looks great now.

jburel:
--no-rebase

PR #1628, opened 2015-02-26
Remove thinkstandards.net link

hflynn:
http://www.thinkstandards.net/benefits.html is broken and when I investigated the entire domain is gone so I've just removed the link assuming it isn't coming back.

PR #1629, opened 2015-02-26
Remove thinkstandards.net link (rebased onto dev_5_0)

hflynn:
This is the same as gh-1628 but rebased onto dev_5_0.

----

http://www.thinkstandards.net/benefits.html is broken and when I investigated the entire domain is gone so I've just removed the link assuming it isn't coming back.


hflynn:
--rebased-from #1628

hflynn:
Building on staging now

hflynn:
http://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.0-merge-docs/1024/ - build green again

PR #1630, opened 2015-02-26
docs: Add support for generating manual pages with Sphinx

rleigh-dundee:
- add `man` target to docs/sphinx `build.xml`
- add manpage generation to toplevel `docs-sphinx` target -- ensures Travis tests them
- manpages are generated in `_build/man`

Known issue: version in the manpages is only major.minor, rather than the full version.

--------

Testing:

- Run `ant man` in `docs/sphinx`
- Run `man -l _build/man/$manpage` and verify it looks OK.

### DIFF
--- a/ant/toplevel.xml
+++ b/ant/toplevel.xml
@@ -189,6 +189,8 @@ Type "ant -p" for a list of targets.
     <ant dir="docs/sphinx" target="html"/>
     <echo>----------=========== Sphinx PDF ===========----------</echo>
     <ant dir="docs/sphinx" target="pdf"/>
+    <echo>----------=========== Sphinx MAN ===========----------</echo>
+    <ant dir="docs/sphinx" target="man"/>
   </target>
 
   <target name="clean-docs-sphinx"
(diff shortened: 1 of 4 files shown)


rleigh-dundee:
@sbesson An issue here is that the version embedded into the manpages is just the major.minor and not the full version as desired.  This is due to using `version` rather than `release`.  If there's a way to get sphinx to use `release`, that would be a nice improvement.

joshmoore:
(For testing on OSX, `man _build/man/bf-cpp-overview.7`)

Few warnings:
```
grotty:<standard input> (<standard input>):3330: character above first line discarded
grotty:<standard input> (<standard input>):3334: character above first line discarded
```

Also, much of the prereq text isn't really appropriate in a man file. Otherwise, quite neat.

rleigh-dundee:
@joshmoore Note this is just making it work and testing it.  I'll be adding the real manual pages for e.g. showinf etc. to follow this up.  These two pages are just examples to show it works, though they may be useful in their own right for developers.

joshmoore:
In that sense, happy once travis is green.

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/617/

rleigh-dundee:
--no-rebase

PR #1631, opened 2015-02-26
Move EXIF parsing to a service

melissalinkert:
This adds an ```EXIFService``` and implementation, which call the ```metadata-extractor``` library.  The service checks for the presence of ```metadata-extractor```, ```xmpcore```, and ```xercesImpl```, and will throw an exception in the constructor if they are missing.

The actual EXIF reading is now wrapped in a try block which catches Throwable, which should prevent an exception from being thrown if the ```xercesImpl``` version precedes 2.8.1.  I couldn't come up with a good way to check the ```xercesImpl``` version in the service constructor.

Tests have also been added to check that the service construction throws an appropriate exception when the ```metadata-extractor``` and ```xercesImpl``` libraries are missing.

To test, verify that all builds are green, and that JPEG files in ```test_images_good/jpeg``` and ```from_skyking/jpeg/exif``` open in showinf/ImageJ and import into OMERO.

### DIFF
--- a/components/formats-bsd/build.properties
+++ b/components/formats-bsd/build.properties
@@ -27,6 +27,7 @@ component.classpath      = ${artifact.dir}/jai_imageio.jar:\
                            ${lib.dir}/perf4j-0.9.13.jar:\
                            ${lib.dir}/metadata-extractor-2.6.2.jar:\
                            ${lib.dir}/xmpcore-5.1.2.jar:\
+                           ${lib.dir}/xercesImpl-2.8.1.jar:\
                            ${lib.dir}/slf4j-api-${slf4j.version}.jar:\
                            ${lib.dir}/guava-${guava.version}.jar:\
                            ${lib.dir}/testng-${testng.version}.jar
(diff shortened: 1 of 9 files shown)


mtbc:
https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/ will take a while yet, of course, but relevant builds seem generally happy.

mtbc:
I *do* generally seem to be getting some EXIF data from images, both through showinf and OMERO, so :+1: for that. For instance,
```
Reading global metadata
Aperture Value: F4
Color Space: sRGB
Components Configuration: YCbCr
Compressed Bits Per Pixel: 2.4 bits/pixel
Date/Time Original: 1998:01:08 08:28:54
Exif Image Height: 1024 pixels
Exif Image Width: 1536 pixels
Exif Version: 2.00
Exposure Bias Value: 0 EV
Exposure Time: 2430/500000 sec
F-Number: F4.7
FlashPix Version: 1.00
Focal Length: 10.4 mm
Max Aperture Value: F2.8
Metering Mode: Center weighted average
Shutter Speed Value: 1/255 sec
Subject Distance: 5.0 metres
White Balance: Unknown
```
 Oddly, importing `4kx4k.jpg` into OMERO 5.1 merge today has left it without any green showing, though the rendering settings look normal:
broken-4kx4k

melissalinkert:
Likely going to need one more commit tomorrow for https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/413

melissalinkert:
97e620a pushed to fix the build.  Similar image issues as above were noted in openmicroscopy/openmicroscopy#3516, but this was thought to be an issue with the colorizer PRs and not the JPEG changes.

rleigh-dundee:
Tested all EXIF samples in OMERO/Insight, with showinf and with ImageJ/loci_tools.  Pixeldata and metadata look good for all samples.  Also tested test_images_good/jpeg with OMERO/Insight and the colour problems above are no longer apparent.  Looks good to merge.

melissalinkert:
--no-rebase

PR #1632, opened 2015-02-27
Adding bar to units

qidane:
This adds the unit 'bar' to pressure.

It is the same 2015-01 but has had its version number bumped to 2. This make it a minor release. 

The xsd change should require a major release but coming so soon on the back of the last release and being in a completely new section for the schema no one should have reader/writer code for this to break.

See version info at: http://www.openmicroscopy.org/site/support/ome-model/schemas/index.html

----

Also some other late schema documentation changes for:
http://trac.openmicroscopy.org/ome/ticket/12693

----

--exclude

### DIFF
--- a/components/specification/released-schema/2015-01/ome.xsd
+++ b/components/specification/released-schema/2015-01/ome.xsd
@@ -760,7 +760,14 @@
 			<xsd:attribute name="NDFilter" use="optional" type="xsd:float">
 				<xsd:annotation>
 					<xsd:documentation>
-						The NDfilter attribute is used to specify the combined effect of any neutral density filters used. [units optical density expressed as a PercentFraction]
+						The NDfilter attribute is used to specify the combined effect of any neutral density filters used.
+						The amount of light the filter transmits at a maximum [units:none]
+						A fraction, as a value from 0.0 to 1.0.
+
+						NOTE: This was formerly described as "units optical density expressed as a PercentFraction".
+						      This was how the field had been described in the schema from the beginning but all
+						      the use of it has been in the opposite direction, i.e. as a amount transmitted,
+						      not the amount blocked. This change has been made to make the model reflect this usage.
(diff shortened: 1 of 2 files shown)


qidane:
This would also require the regeneration of the oxygen docs and a release to the website.

components/specification/released-schema/2015-01/ome.xsd
```diff
@@ -760,7 +760,14 @@
...
 					<xsd:documentation>
-						The NDfilter attribute is used to specify the combined effect of any neutral density filters used. [units optical density expressed as a PercentFraction]
+						The NDfilter attribute is used to specify the combined effect of any neutral density filters used.
+						The amount of light the filter transmits at a maximum [units:none]
+						A fraction, as a value from 0.0 to 1.0.
+
+						NOTE: This was formerly described as "units optical density expressed as a PercentFraction".
+						      This was how the field had been described in the schema from the beginning but all
+						      the use of it has been in the opposite direction, i.e. as a amount transmitted,
+						      not the amount blocked. This change has been made to make the model reflect this usage.
```
qidane: @melissalinkert Is `SlidebookReader.java` populating the value in the same was as DeltaVision, i.e. in the new way described above?

melissalinkert: I think it's the same, but we honestly don't have much information about what the correct NDFilter values are for Slidebook files.

qidane:
I think this is ready to merge, but we will need to work through the release process (on ome-internal) about the same time as merging.

joshmoore:
Talking to @qidane, I'd suggest next steps of:
 * release the copy from /ome/www (with version 2)
 * use that to re-generate the documentation & rsync

Merging this is then independent of either of those. Final vetoes?

sbesson:
Version 2 released today. Thanks @qidane. Merging this.

PR #1635, opened 2015-03-03
C++: Add showinf manual page and update showinf options

rleigh-dundee:
- Reorganise C++ documentation to allow for easier addition of more C++ material (in later PRs)
- Make `cmake` generate manual pages (currently in the cmake run; could defer it with additional work)
- Make `cmake` install manual pages; location is configurable with `CMAKE_INSTALL_[FULL_]MANDIR` as for other install paths.
- Reorganise `showinf` options; this makes it possible to customise all sections of display output
- Document all `showinf` options in new `showinf.1` manual page; currently unimplemented functionality is marked, as are defaults.
- While updating the `showinf` options, I found a bug in the OME-XML model enum error handling; this is corrected here and also adds istream operators to set enum values from streams (test cases added for all changes)

--no-rebase

--------

Testing:

- Builds should all be green
- showinf manual page will also be a section in the regular sphinx manual
- C++ packages should contain a `man/` directory
- View showinf manpage with `man -l man/showinf.1` (package) or `man -l cpp/man/showinf.1` (build tree)
- `showinf --help` output should be the same as the options documented in the manpage (there may be slight wording differences due to the need for brevity in the `--help` output but the meaning and defaults should be the same.


mtbc:
For the double-hyphen options that take an argument, the manpage has, say, `--format=reader` where the command's `-h` help says, `--format arg`. Also, the manpage says things like `series=n` where the help says, `--series arg`.

mtbc:
It's odd that the manpage has arguments that are both unimplemented and default.

docs/sphinx/developers/cpp/showinf.txt
```diff
@@ -0,0 +1,152 @@
...
+  Use the specified sub-resolution (only if not flattened with --flat)
+  (UNIMPLEMENTED).
+
+.. option:: --input-order=XY[ZTC]
+
+  Override the dimension input order (UNIMPLEMENTED).
+
+.. option:: --output-order=XY[ZTC]
+
+  Override dimension output order (UNIMPLEMENTED).
```
mtbc: +"the" to match above?

mtbc:
Docs changes staged at http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/.

mtbc:
This looks generally good to me but I might have easily missed something.

rleigh-dundee:
@mtbc The --help output is autogenerated by the Boost.Program_Options parser, not under our control, hence why it uses the generic "arg" rather than a more meaningful name.  However, it appears that it's now possible to customise this:

http://stackoverflow.com/questions/1249646/when-using-boostprogram-options-how-does-one-set-the-name-of-the-argument

We'll need to make that conditional on having a suitably recent version of Boost.

With respect to unimplemented defaults, it would be the default if it were implemented, but right now it's a no-op.  It's merely a placeholder for the addition of the implementation in the future.  Happy to reword it if there's a clearer alternative!

mtbc:
Ah, both whitespace and `=` separators are okay then?

rleigh-dundee:
@mtbc Both are acceptable forms.

rleigh-dundee:
Corrected inconsistency pointed out by @mtbc and reordered sections as suggested by @joshmoore

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/628/

joshmoore:
Thanks, @rleigh-dundee. @hflynn / @melissalinkert : thoughts on the changes to https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/index.html ?

hflynn:
I wonder if it would make more sense to have the native C++ section *below* the other Java library (from non-Java code) section but I guess the idea is to highlight it as the preferred method and I see you suggested this ordering @joshmoore 
Other than that, looks fine structure-wise to me.

joshmoore:
Would another option be to have all of the non-java-code section as a
single bullet point under "Java"?

hflynn:
That might be the best idea actually yes - add 'Interfacing with Bio-Formats from non-Java code' to the bottom of the 'Using Bio-Formats as a Java Library' list and move all the other pages below it in the TOC.

rleigh-dundee:
Happy to make any changes to the organisation if you're in agreement on what they should be.  Could they be done in a separate PR later today, because I have other branches blocked on this.

hflynn:
Yes, no reason to block this for them. And I'm happy to do that re-arrangement if it won't interfere with what you're doing going forward.

joshmoore:
Merging and leaving the docs proposal in @hflynn's capable hands.

PR #1636, opened 2015-03-03
C++: cmake 3.1 fixes and compiler standard updates

rleigh-dundee:
- Correct conditional logic to work with cmake 3.1.x (variable expansion in conditionals is slightly different)
- Add C++14 to the set of compiler standards to try to enable
- Disable autodetection of compiler standard by default; this makes the default behaviour match the binary builds we will be distributing.

--no-rebase

--------

Testing:

Should be mostly taken care of by CI builds.  The main thing to check for (if using cmake 3.1.x with homebrew/ports) is that the initial compiler checks terminate correctly.  For example on FreeBSD with clang++3.4 (similar to current XCode):

```
-- Using git commit 117b4f5 on 2015-03-03 21:11:34 +0000
-- Performing Test CXX_FLAG_CXX14
-- Performing Test CXX_FLAG_CXX14 - Failed
-- Performing Test CXX_FLAG_CXX11
-- Performing Test CXX_FLAG_CXX11 - Success
-- Performing Test CXX_FLAG_CXX11_CSTDARG
-- Performing Test CXX_FLAG_CXX11_CSTDARG - Success
-- Performing Test CXX_FLAG_CXX11_STDARG
-- Performing Test CXX_FLAG_CXX11_STDARG - Success
-- Performing Test CXX_FLAG_CXX11_ENABLE_IF
-- Performing Test CXX_FLAG_CXX11_ENABLE_IF - Success
-- Performing Test CXX_FLAG-pedantic
...
```

There are no following checks for C++03/C++98/None.  Without this patch, there will be, followed by an warning that the C++ standard could not be determined, i.e. falling back to the compiler default.

If using an older cmake version, i.e. 2.8 or 3.0, the above behaviour should continue to work as before.

The old cmake can be checked using the gretzky35 build logs (ON variant); the new cmake can be checked on perch and mackerel (ON variants).


mtbc:
So are there specific CI jobs we should look at for things like "perch and mackerel (ON variants)" or is it a matter of ssh'ing into them and poking around somewhere?

rleigh-dundee:
@mtbc It's the matrix components of https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/  No need to ssh.  I'll see what's making them fail first.

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/628/

PR #1638, opened 2015-03-04
Ome xml reader

jburel:
Allow to import OME-XML w/o binary data.
This is useful when testing https://github.com/openmicroscopy/samples/pull/3

To test:
 * ```./showinf``` on one of the xml file from samples.
 * Import one of the XML files into OMERO.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
@@ -127,6 +127,7 @@ public class OMEXMLReader extends FormatReader {
   public byte[] openBytes(int no, byte[] buf, int x, int y, int w, int h)
     throws FormatException, IOException
   {
+    if (binDataOffsets.size() == 0) return buf;
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
     int index = no;
@@ -251,10 +252,6 @@ public class OMEXMLReader extends FormatReader {
         lineNumber++;
       }
       binDataOffsets.add(in.getFilePointer() + col - 1);
-    }
-
-    if (binDataOffsets.size() == 0) {
-      throw new FormatException("Pixel data not found");
     }
 
     LOGGER.info("Populating metadata");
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
```diff
@@ -127,6 +127,7 @@ public boolean isThisType(String name, boolean open) {
   public byte[] openBytes(int no, byte[] buf, int x, int y, int w, int h)
     throws FormatException, IOException
   {
+    if (binDataOffsets.size() == 0) return null;
```
melissalinkert: Could this not return ```buf``` instead of ```null```?  That's what we typically do for missing pixel data, and would eliminate the need for special handling of the ```null``` return value.

jburel: Will do that.

jburel:
@melissalinkert: Made changes in reader, do you want me to remove the case handling ```null```?

melissalinkert:
@jburel: Thanks, the reader looks good now.  I'd be in favor of removing the ```null``` handling, since ```openBytes``` is expected to always return non-null.

jburel:
@melissalinkert: ok will do that.

jburel:
@melissalinkert: done

mtbc:
Generally seems to work okay. Only problem I noticed was with `zeiss-lsm/martin/sample files.mdb/sample files.ome.xml` where showinf reports,
```
org.xml.sax.SAXParseException; lineNumber: 1776; columnNumber: 6; The processing instruction target matching "[xX][mM][lL]" is not allowed.
```
but the other two files @will-moore touched open and import just fine.

melissalinkert:
@mtbc: I think that's something that needs to be fixed in the samples PR, as there are two separate blocks of OME-XML contained in ```zeiss-lsm/martin/sample files.mdb/sample files.ome.xml```.

will-moore:
@melissalinkert Removed the duplicate XML block (copy and paste error) in https://github.com/will-moore/samples/commit/e90a59548e233d08bae4d5b48ab88af89e59fc7b

melissalinkert:
Thanks, @will-moore.

jburel:
--no-rebase

PR #1640, opened 2015-03-05
Four float color representation

qidane:
Color stored as RGBA in rangew 0.0 to 1.0

-----

--exclude
--on-hold


qidane:
Closed as not compatible with current code base, but does show what changes are planned.

PR #1644, opened 2015-03-05
Add exclude for oxygen files

qidane:
Exclude files `*.xpr` and `omeOxygenDocConfig.xml`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ testng-results.xml
 testng.css
 tools/*.jar
 *.log
+*.xpr
 
 # For Fiji #
 loci_tools.jar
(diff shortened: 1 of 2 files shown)


rleigh-dundee:
The change itself is OK.  I wonder though if it would be better to add to .gitignore to prevent them being committed into git in the first place.

qidane:
`omeOxygenDocConfig.xml` is used as part of the release process so we want that in git. I can add *.xpr to `gitignore`

qidane:
@rleigh-dundee Ok?

rleigh-dundee:
Does `omeOxygenDocConfig.xml` also need adding?

qidane:
`omeOxygenDocConfig.xml` is used as part of the release process so we want that in git.

qidane:
--no-rebase

PR #1647, opened 2015-03-05
MinimalTiffReader: say which file is invalid

ctrueden:
This is helpful in multi-file datasets with many files.


melissalinkert:
--no-rebase

ctrueden:
Thanks!

PR #1651, opened 2015-03-06
JPEG files have pixels

mtbc:
The JPEG reader should return nothing when asked for companion files.

--no-rebase


melissalinkert:
Looks fine, just waiting to merge until at least some of the merge builds are green this evening.

PR #1652, opened 2015-03-06
Update some obsolete URLs, and shorten others

ctrueden:
TTSIA.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/About.java
+++ b/components/bio-formats-plugins/src/loci/plugins/About.java
@@ -44,11 +44,11 @@ public final class About implements PlugIn {
 
   /** URL of the Bio-Formats web page. */
   public static final String URL_SOFTWARE =
-    "http://www.openmicroscopy.org/site/products/bio-formats";
+    "http://openmicroscopy.org/info/bio-formats";
 
   /** URL of Bio-Formats ImageJ web page. */
   public static final String URL_BIO_FORMATS_IMAGEJ =
-    "http://www.openmicroscopy.org/site/support/bio-formats/users/imagej/features.html";
+    "http://openmicroscopy.org/info/bio-formats/users/imagej/features.html";
 
   /** URL of Data Browser web page. */
   public static final String URL_DATA_BROWSER =
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -312,7 +312,7 @@ public final class FormatTools {
 
   /** URL of OME-TIFF web page. */
   public static final String URL_OME_TIFF =
-    "http://ome-xml.org/wiki/OmeTiff";
+    "http://www.openmicroscopy.org/site/support/ome-model/ome-tiff/";
 
   // -- Constructor --
 
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -304,15 +304,15 @@ public final class FormatTools {
 
   /** URL of Bio-Formats web page. */
   public static final String URL_BIO_FORMATS =
-    "http://www.openmicroscopy.org/site/products/bio-formats";
+    "http://openmicroscopy.org/info/bio-formats";
 
   /** URL of 'Bio-Formats as a Java Library' web page. */
   public static final String URL_BIO_FORMATS_LIBRARIES =
-    "http://www.openmicroscopy.org/site/support/bio-formats/developers/java-library.html";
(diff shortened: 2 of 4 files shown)


snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

components/bio-formats-plugins/src/loci/plugins/About.java
```diff
@@ -44,11 +44,11 @@
 
   /** URL of the Bio-Formats web page. */
   public static final String URL_SOFTWARE =
-    "http://www.openmicroscopy.org/site/products/bio-formats";
+    "http://openmicroscopy.org/info/bio-formats";
```
sbesson: No strong veto to this change that but a mid-term implication is that this info redirect will need to be actively tested/maintained. /cc @hflynn

components/bio-formats-plugins/src/loci/plugins/About.java
```diff
@@ -44,11 +44,11 @@
...
 
   /** URL of Bio-Formats ImageJ web page. */
   public static final String URL_BIO_FORMATS_IMAGEJ =
-    "http://www.openmicroscopy.org/site/support/bio-formats/users/imagej/features.html";
+    "http://openmicroscopy.org/info/bio-formats/users/imagej/features.html";
```
qidane: This is not how the info redirects were meant to work. The initial use was `http://openmicroscopy.org/info/bio-formats` and that plus anything below that should redirect to `http://www.openmicroscopy.org/site/products/bio-formats`.

This means that `http://openmicroscopy.org/info/bio-formats/any-sub-path` is not supported and is probably only working by luck.

qidane:
We have only used the info URLs on leaflets and posters, where we wanted to keep them short as people would be typing them in.

hflynn:
Yeah, the info redirects are not intended for this sort of use. The Data Model doc links are an improvement though, thanks. @ctrueden can you roll back the 2nd commit or would it be easier for me to open a new PR with just that commit for you?

hflynn:
Can someone close this please? I've cherry-picked the first commit to open a new PR. Thanks again @ctrueden

ctrueden:
> This is not how the info redirects were meant to work.

Very sorry to hear that.

> We have only used the info URLs on leaflets and posters

I have linked to `info` subpaths many times, which are now embedded in many list archive messages. Just in 2015 alone:

* https://list.nih.gov/cgi-bin/wa.exe?A2=IMAGEJ;5f97c1d1.1503
* https://groups.google.com/d/msg/icy-software/CBw0QCTyv6Q/nPGUgySUnNsJ
* https://groups.google.com/d/msg/fiji-devel/RnVgeAYfBZY/sOSNpcmM2wwJ
* https://list.nih.gov/cgi-bin/wa.exe?A2=IMAGEJ;c13c0558.1501

And many more before that.

> This means that `http://openmicroscopy.org/info/bio-formats/any-sub-path` is not supported and is probably only working by luck.

I urge you not to "fix" this behavior. It will break all the previous links, for no pragmatic purpose whatsoever.

ctrueden:
@qidane Here is an excerpt from a mail you wrote to `ome-nitpick` on Aug 30, 2013 with subject "URL shortcut redirects":

> Links in the following form will now work:
> http://openmicroscopy.org/info/bio-formats/developers/non-java-code.html (or any other sub folder)
> http://openmicroscopy.org/info/bio-formats/
> http://openmicroscopy.org/info/bio-formats

So it seems that originally, the `info` links _were_ intended to work that way.

qidane:
Nope, no memory of that at all. I guess I must have reworked them for some reason... wonder why. It seems unlikely we would have put the long form on a poster or leaflet.

ctrueden:
@qidane If you read that nitpick thread, you recommended the `info` links as a way of giving the shortest possible links to people in mailing list replies, etc. Do you want me to stop doing that? If so, what is the alternative?

qidane:
I think it is a decision for @hflynn how you want to handle this going forward.

hflynn:
Links in an email can be pasted into a browser or clicked directly so I don't see it matters how long the url is - best just to use the real link. The original point of the info redirects was because people have to type urls from leaflets or posters so shorter/easier to remember has value.

ctrueden:
@hflynn Nonetheless, I hope the `info` links continue to work for the foreseeable future. They are very useful beyond that use case. And there are many uses of them in the wild now.

PR #1655, opened 2015-03-10
C++: Add ome-common component

rleigh-dundee:
Move utility (non-stdlib-compatibility code) from ome-compat and ome-xerces into ome-common.
Note that this increases the minimum CMake version to 2.8.12; travis updated to use a newer version than available in Ubuntu 12.04.  I was going to require 3.0.0 for INTERFACE libraries; instead I've made this optional with a fallback for 2.8.12; earlier versions could possibly still be used, but the previous minimum was 2.8.0 which was unlikely to work and this makes the minimum version what we are actively using and testing; it can be lowered if needed.

Code changes amount to changing namespaces of classes and functions.

- `ome::compat` methods and classes moved to `ome::common`
- `ome::xerces` is now `ome::common::xml` but can be made `ome::common::xerces` if preferred

--------

Testing: all jobs should remain green.


melissalinkert:
All reads fine to me.

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/639/

joshmoore:
@rleigh-dundee: can you explain c4d4530? I assume this PR was not mergeable? But that then means we need a new merge job passing. Further, we've typically not used this style of cherry-pick pieces of commits to fix merge conflicts.

rleigh-dundee:
@joshmoore It's a straightforward merge of origin/develop with a one line change to fix the conflict in `.travis.yml`.  It's nothing unusual, though the commit message isn't the standard git merge message.  There is no cherry-picking here; if you look at the commit graph you'll see origin/develop is the parent.

joshmoore:
Can see that now, but clearly confusing (to at least me). Considering that the last comment from you is "build is green", some next comment stating, "waiting on build" or whatever would be useful. Holding off for now.

rleigh-dundee:
@joshmoore I thought the chronology of the comments and commits showed this, I can delete that if it's confusing.

joshmoore:
No need to delete. I just didn't understand the "build is green" as the last comment and then found a new commit (which I couldn't immediately grok).

rleigh-dundee:
@joshmoore Green again in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/642/

joshmoore:
Very cool. Thanks, @rleigh-dundee

PR #1656, opened 2015-03-10
MRC: add signed bytes support

sbesson:
See https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7757&p=15458#p15458 and http://bio3d.colorado.edu/imod/doc/mrc_format.txt.

According to the MRC specification, mode 0 can correspond to `int8` data depending on the value of `imodStamp`. At the moment MRCReader is imposing `uint8`. This PR adds support for `int8` mrc data by adding an extra check for mode 0 and removing the block which flips the pixel type to unsigned bytes depending on the maximum and minimum pixel values.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MRCReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MRCReader.java
@@ -254,21 +254,6 @@ public class MRCReader extends FormatReader {
     }
     double pixelTypeMax = pixelTypeMin + range;
 
-    if (pixelTypeMax < maxValue || pixelTypeMin > minValue && signed) {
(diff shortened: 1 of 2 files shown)


sbesson:
Marking as exclude since https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/10/ is failing on a couple of sample files.

sbesson:
Closing and re-opening to use the fix from #1657

will-moore:
Image and pixel range look good.

screen shot 2015-03-12 at 11 05 02

sbesson:
--no-rebase

PR #1657, opened 2015-03-11
Impose Sphinx 1.2.3 for Travis build

sbesson:
Sphinx 1.3.0 has been released on March 10 and is installed by default.
However our code currently fails with this version. Fixes will need to
be brought to the documentation. In the meantime, the version is capped
to 1.2.3 to let Travis pass.

PR #1658, opened 2015-03-11
Impose Sphinx 1.2.3 for Travis build (rebased onto dev_5_0)

sbesson:
This is the same as gh-1657 but rebased onto dev_5_0.

----

Sphinx 1.3.0 has been released on March 10 and is installed by default.
However our code currently fails with this version. Fixes will need to
be brought to the documentation. In the meantime, the version is capped
to 1.2.3 to let Travis pass.


sbesson:
--rebased-from #1657

PR #1659, opened 2015-03-11
remove usage of ConcurrentHashMap

mtbc:
ConcurrentHashMap undergoes API changes among Java 6, 7, 8, so to keep code nicely portable it's generally easier to simply avoid using it given that we already have Guava available and that offers a suitable substitute.

(A larger refactoring could use Guava's CacheBuilder so it would handle expiration for us too, but 5.1.0 has more urgent demands!)

Testing: No tests or builds should break, and the Bio-Formats source should no longer contain direct usage of ConcurrentHashMap.

--no-rebase

### DIFF
--- a/components/formats-common/build.properties
+++ b/components/formats-common/build.properties
@@ -15,6 +15,7 @@ component.classpath      = ${lib.dir}/jgoodies-forms-1.7.2.jar:\
                            ${lib.dir}/logback-classic-${logback.version}.jar:\
                            ${lib.dir}/logback-core-${logback.version}.jar:\
                            ${lib.dir}/slf4j-api-${slf4j.version}.jar:\
+                           ${lib.dir}/guava-${guava.version}.jar:\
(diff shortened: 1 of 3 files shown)


melissalinkert:
Reads fine to me, so happy to merge once the merge build picks this up.  Card started for future Guava-related changes: https://trello.com/c/RCIgkCgX/145-evaluate-guava-usage

mtbc:
This PR causes http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-March/005166.html but I don't understand how to adjust it to fix the problem for Fiji.

melissalinkert:
No, this did not cause the problem in Fiji.  See this thread: https://list.nih.gov/cgi-bin/wa.exe?A1=ind1503&L=imagej#12, particularly https://list.nih.gov/cgi-bin/wa.exe?A2=ind1503&L=imagej&F=&S=&P=204603

mtbc:
Ah, great, thank you.

PR #1664, opened 2015-03-12
build: Use long and short git revisions

rleigh-dundee:
The git version used by various git commands in `ant/version.xml` was a short 6-digit form of the commit hash.  This is too short to use without risk of clashes.  This PR adds a long form of the hash and uses that instead; a short revision copy of the hash is retained for user display where needed.

See also: https://github.com/openmicroscopy/openmicroscopy/pull/3612

--no-rebase

/cc @sbesson 

--------

Testing:

Check the content of `ant/gitversion.xml` and `cpp/cmake/GitVersion.cmake`; these should have short and long git versions.

Check `ant version` in the build tree and unpacked zip.

### DIFF
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -264,6 +264,7 @@ your FindBugs installation's lib directory. E.g.:
 
     <replace dir="${dest.dir}" token="$${date}" value="${DATE}"/>
     <replace dir="${dest.dir}" token="$${year}" value="${YEAR}"/>
+    <replace dir="${dest.dir}" token="$${vcs.shortrevision}" value="${vcs.shortrevision}"/>
     <replace dir="${dest.dir}" token="$${vcs.revision}" value="${vcs.revision}"/>
     <replace dir="${dest.dir}" token="$${vcs.date}" value="${vcs.date}"/>
     <replace dir="${dest.dir}" token="$${vcs.date}_unix" value="${vcs.date_unix}"/>
@@ -297,6 +298,7 @@ your FindBugs installation's lib directory. E.g.:
 
   <replace dir="${dest.dir}" token="$${date}" value="${DATE}"/>
   <replace dir="${dest.dir}" token="$${year}" value="${YEAR}"/>
+  <replace dir="${dest.dir}" token="$${vcs.shortrevision}" value="${vcs.shortrevision}"/>
   <replace dir="${dest.dir}" token="$${vcs.revision}" value="${vcs.revision}"/>
   <replace dir="${dest.dir}" token="$${vcs.date}" value="${vcs.date}"/>
   <replace dir="${dest.dir}" token="$${release.version}" value="${release.version}"/>
--- a/ant/toplevel.xml
+++ b/ant/toplevel.xml
@@ -1036,6 +1036,7 @@ Type "ant -p" for a list of targets.
       <arg value="--release=bioformats"/>
       <arg value="--bioformats-shortversion=${release.shortversion}"/>
       <arg value="--bioformats-version=${release.version}"/>
+      <arg value="--bioformats-vcsshortrevision=${vcs.shortrevision}"/>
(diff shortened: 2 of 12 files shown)


sbesson:
Works for me. From the release source zip of [BIOFORMATS-5.1-merge-build](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-build/lastSuccessfulBuild/artifact/artifacts/), I have:

```
sbesson@ls30630:bioformats-5.1.0-DEV $ grep revision ant/gitversion.xml 
    <property name="vcs.shortrevision" value="089ab33"/>
    <property name="vcs.revision" value="089ab33cd635c3ca9ecea9db632dc74366f80858"/>
sbesson@ls30630:bioformats-5.1.0-DEV $ grep REV cpp/cmake/GitVersion.cmake 
set(OME_VCS_SHORTREVISION "089ab33")
set(OME_VCS_REVISION "089ab33cd635c3ca9ecea9db632dc74366f80858")
```

In terms of side-effects, the long format of `VCS_REVISION` is now exposed in the command-line tools logging:

```
sbesson@ls30630:bftools $ ./showinf -version
Version: 5.1.0-DEV
VCS revision: 089ab33cd635c3ca9ecea9db632dc74366f80858
Build date: 16 March 2015
```

Another notable place where `VCS_REVISION` is involved in some logical operations is the version checking of the `Memoizer`. Note however, this option is turned off by default and the Memoizer will be invalidated anyways in `5.1.0` vs `5.0.x` because of the kryo version bump. However, the choice made by this PR will likely be irreversible for `5.1.x`.

Assuming the 2 side effects mentioned above are acceptable, this PR looks good to merge.

melissalinkert:
Can the usage of ```FormatTools.VCS_REVISION``` be replaced with ```FormatTools.VCS_SHORT_REVISION```?  This should be a quick fix, and would preserve existing behavior.

rleigh-dundee:
@melissalinkert Done.

melissalinkert:
There are still several places where ```VCS_REVISION``` is used:

* loci.formats.tools.ImageConverter
* loci.formats.tools.ImageInfo
* loci.formats.Memoizer
* loci.plugins.macro.LociFunctions

(see the output of ```git grep VCS_REVISION```)

rleigh-dundee:
@melissalinkert I left those ones in intentionally, particularly the Memoizer, since they are for debug and/or storing and in the case of the Memoizer the ambiguity of the short form is highly undesirable and for the debug output potentially confusing.  While the chances of a clash is rare, it can clearly happen--this is what prompted this change, and that could screw up the Memoizer.  Happy to revert them if required, but might need to be careful.

melissalinkert:
OK, but the ones changed in 5519ec0 are "just" for debugging as well.  I would argue that this should be consistent across the board - either preserve the existing behavior everywhere, or we decide that it's safe to use the long revision everywhere (I have worries about how this will affect Memoizer, but I won't have sufficient time to thoroughly test until next week).

rleigh-dundee:
@melissalinkert The ones in 5519ec0 were user-visible program messages, not diagnostics.  I can convert the rest to the short form as well if you prefer, but that would mean the debug output would be potentially ambigious.  Or have them all use the long form.  The Memoizer shouldn't be a problem should it?  Any bfmemos created from this point would be using the long form which will naturally not be equal to the earlier short form as required?

Note today's https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-latest-cpp/280/ -- another failure due to a hash clash.

melissalinkert:
In the interest of drawing a line under this, let's assume that the long form is safe everywhere - now that I think on it, OMERO isn't using the hash version checking in Memoizer.  Reverting 5519ec0 should be sufficient then for this to be merged.

rleigh-dundee:
Done

PR #1665, opened 2015-03-12
Correct downgrade transform for map & units

qidane:
Now correctly handles empty maps in imaging environment and wavelength units.

Needed by @jburel OMERO tests.

----

--no-rebase


jburel:
replaced by gh-1668

qidane:
I have some more commits for this that fix the IDs.

jburel:
I have made change so the tests now pass.
Let me know if the xslt is correct. 
Did you find more elements to change?

PR #1668, opened 2015-03-13
Fix transforms

jburel:
Now correctly handles empty maps in imaging environment and wavelength units.
Fixes older transforms not generating valid xml.

Corresponding OMERO tests https://github.com/openmicroscopy/openmicroscopy/pull/3614
cc @qidane 
This PR replaces gh-1665


jburel:
@qidane: what was the reason for adding the ```xslt:fix``` in the first place when downgrading to 2003-FC? I cannot remember

qidane:
I think the problem is/was the regex in the 2003-FC schema for ID always was broken, so if you have a full `urn` form of the LSID then it will not validate. The `xslt.fix:` prefix allowed the urn form to validate and still keep the full LSID in tact. The problem is when the 2003-FC schema is then upgraded again the fixed ID is now broken as it will not validate with the corrected regex.

jburel:
one more commit to come

jburel:
Few more issues noticed when running against files from https://github.com/openmicroscopy/samples
Actions:
 * fix stylesheet
 * Improve mock-objects.

jburel:
@qidane: any reason to have unit-conversion.xsl stylesheet version 2.0 and the rest 1.0.
It will be easier to upgrade the stylesheet.

jburel:
Also I do no think we should strip out wavelength if it is not an int, since in many cases we will have a value like 550.0. I know it is not the rule we have applied in previous transform but I am worried that we lose valuable information
cc @will-moore @rleigh-dundee @pwalczysko @joshmoore @melissalinkert

jburel:
The examples from samples should now validate.

jburel:
Still an issue with microbeam manipulation

jburel:
I think we should re-consider the strategy currently in place when downgrading from float to int
Currently if round(value) != value the value is lost. e.g. 540.1 will be lost instead of having 540 in an older schema.
This generates failures when downgrading to older the schema and also removes some valuable information.
cc @qidane @joshmoore @melissalinkert @will-moore @pwalczysko

jburel:
Ant suite
Total tests run: 76, Failures: 0, Skips: 0
===============================================

jburel:
FYI: @qidane is also making sure that the downgrade paths in ome-transform match the ones suggested in the documentation.

qidane:
There are a few differences:

2008-09 to 2008-02
2009-09 to 2008-02
2010-04 to 2008-02
2009-09 to 2008-09
2010-04 to 2008-09

Just checkin wether docs or ome-transforms have best solution.

qidane:
I think I would prefer to change the ome-transforms file as I the method from the documentation has fewer steps so is likely to produce a margenly better result. @jburel how do you feel about this change? Would you prefer I just updated the docs?

jburel:
can you push a change to the transform so I can run the test locally to access situations?
Tests are not passing on ci-node, due to encoding. investigating issues cc @sbesson

qidane:
https://github.com/jburel/bioformats/pull/3

PR #1669, opened 2015-03-16
Correct the 2008-02 catalog

qidane:
Corrects incorrect entries in the catalog that lead to validation failures of 2008-09 files.

----

--no-rebase


jburel:
@rleigh-dundee: could you have a look? since the catalog is mainly used by C++

rleigh-dundee:
@jburel Looks OK to me.  Sorry about that--looks like an oversight when I was creating all the catalogues.

PR #1671, opened 2015-03-16
Bug Fix (Color Properties): Open OmeroROI's in ImageJ

bramalingam:
Testing Instructions:

1) Open an omero Image in Insight, create/modify ROI shape properties (Specifically stroke width and stroke colour of the ROI's) and save them.
2) Now open the Image on ImageJ using the OMERO-IJ Plugin and check if the ROI's are added to the ROIManager (and the shape properties are replicated in ImageJ).

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/DisplayHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/DisplayHandler.java
@@ -211,8 +211,10 @@ public class DisplayHandler implements StatusListener {
 
   public void displayROIs(ImagePlus[] imps) {
     if (!options.showROIs()) return;
-    ROIHandler.openROIs(process.getOMEMetadata(), imps);
+    ROIHandler.openROIs(process.getOMEMetadata(), imps, options.isOMERO());
+    
   }
+   
 
   // -- StatusListener methods --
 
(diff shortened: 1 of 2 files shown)


pwalczysko:
In ImageJ all works as epxected. Will test Fiji after the meeting.

pwalczysko:
Okay, I have found an edge case which still does not work (both IJ and Fiji).
  - create some regions with changed properties (color and line width) in OMERO
  - open the image in ImageJ (all works fine)
  - create on this image one more region in ImageJ (or Fiji) - use the default line width and color (=yellow)
  - use ``Save to OMERO`` (all is saved as expected and regions still look exactly as created)
  - reopen the same image in ImageJ or Fiji - the region which was created last has slightly thicker line (just like if it would try to adjust to the lines of the previously created regions)  -> not expected, the region should have the default line thickness, just like it has in OMERO and just like it was created

Note that the region is visible though in any case. 
Also note that if you stick to default properties, it does not matter how many round trips you make with the regions, they are always visible as drawn in ImageJ and Fiji.

pwalczysko:
Although I found a bug https://github.com/openmicroscopy/bioformats/pull/1671#issuecomment-82458999, I think this is an edge case. The behaviour here is fine in general for ImageJ and Fiji.
I would suggest to merge this and deal with the problem at a later time.

melissalinkert:
Agreed - this PR only changes colors and likely wouldn't have introduced the line width difference.

pwalczysko:
Thanks. Created a follow-up checklist item on https://trello.com/c/deNMiFvH/88-plugins-for-imagej-implementation

bramalingam:
--no-rebase

PR #1672, opened 2015-03-17
cpp: qtwidgets: Correct camera pan direction in GLView2D

rleigh-dundee:
--no-rebase

--------

Testing:

Run `bf-test view` and open an image.  Without the PR, the "pan" action will be the opposite of the mouse movement i.e. moving the scene rather than the camera.  With the PR the x and y motion is inverted, making panning act like a click-drag operation where the scene being moved will correspond 1:1 with the mouse movement.

/cc @pwalczysko @mtbc


mtbc:
Good to merge.

PR #1677, opened 2015-03-19
Re-use getImageCreationDate logic to populate global DateTime metadata

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-March/005128.html

Under some conditions, retrieving `IFD.DATE_TIME` can return a String array. This is already handled by the `getImageCreationDate()` method which returns the first item of the array. This PR re-uses the existing logic to fill the global `DateTime` metadata with a `String` instead of a `String []`.

To test this PR, run `showinf` on the file mentioned in the forum thread. `DateTime` should be populated in the `Global metadata` section with a string. Alternatively, reproduce the MATLAB steps described in the thread.

--no-rebase


melissalinkert:
:+1:  Completely happy for this be merged if relevant builds are green.

PR #1680, opened 2015-03-23
Tiffsaver

jburel:
Close the output Stream.
Added a close method to simplify.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -343,6 +343,9 @@ public class TiffWriter extends FormatWriter {
     if (in != null) {
       in.close();
     }
+    if (tiffSaver != null) {
+      tiffSaver.close();
+    }
   }
 
   /* @see loci.formats.FormatWriter#getPlaneCount() */
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -85,7 +85,6 @@ public class TiffSaver {
   private CodecOptions options;
 
   // -- Constructors --
-
   /**
    * Constructs a new TIFF saver from the given filename.
    * @param filename Filename of the output stream that we may use to create
@@ -134,6 +133,16 @@ public class TiffSaver {
   }
 
   // -- TiffSaver methods --
+
+  /**
+   * Closes the output stream if not null.
+   * @throws IOException Thrown if an error occurred while closing.
+   */
+  public void close() throws IOException {
+    if (out != null) {
+      out.close();
+    }
+  }
 
   /**
    * Sets whether or not we know that the planes will be written sequentially.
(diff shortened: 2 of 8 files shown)


components/formats-bsd/src/loci/formats/tools/EditTiffG.java
```diff
@@ -158,21 +158,28 @@ public void openFile(File f) {
...
       showError(exc);
     }
     catch (IOException exc) {
       showError(exc);
+    } finally {
+      try {
+        if (in != null) in.close();
+      } catch (Exception e) {}
+      try {
+        if (out != null) in.close();
```
mtbc: `out.close()`

jburel: cut and paste Thanks @mtbc

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -343,6 +343,9 @@ public void close() throws IOException {
     if (in != null) {
       in.close();
     }
+    if (tiffSaver != null) {
+      tiffSaver.close();
```
joshmoore: Moving forward we should probably be going to a:
```
public static boolean closeSafely(Closeable c) {
  if (c == null) return true;
  try {
    c.close();
    return true;
  } exception (Exception|Throwable) {
    log....
    return false;
  }
}
```

jburel: probably in another PR. Since it will require a review of many more readers/writers

mtbc: If Bio-Formats is to be stuck at Java 6 for some time yet, so without `AutoCloseable`, I wonder if it would be useful in such a review to also add a,
```java
@Override
protected void finalize() {
    close();
}
```
Just an idea, though, and outwith this PR.

components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
```diff
@@ -85,7 +85,6 @@
   private CodecOptions options;
 
   // -- Constructors --
-
```
mtbc: this line break nicely fits the pattern in the rest of the class

components/formats-bsd/src/loci/formats/tools/EditTiffG.java
```diff
@@ -158,21 +158,28 @@ public void openFile(File f) {
...
     }
     catch (IOException exc) {
       showError(exc);
+    } finally {
+      try {
+        if (in != null) in.close();
+      } catch (Exception e) {}
+      try {
+        if (out != null) out.close();
+      } catch (Exception e) {}
```
mtbc: may want to log?

mtbc:
Looks good, I don't see any blockers to merging.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
joshmoore: I don't think I would feel comfortable fiddling with finalization at this stage.

mtbc: Oh, sure, I was assuming that @jburel's "another PR" was 5.1.x.

jburel: it is. not for 5.1. 
I was just checking that the streams were closed.

jburel:
--no-rebase

PR #1681, opened 2015-03-23
C++: Update build instructions with viewer dependencies

rleigh-dundee:
- Also align defaults with changed cmake option defaults

--------

Testing: N/A unless you want to try the install commands in the "quick start" section.


joshmoore:
Very cool, thanks @rleigh-dundee. Could I also get the appropriate **runtime** install lists?

rleigh-dundee:
@joshmoore That's still todo.  For homebrew and BSD ports, it's going to be the same list.  For Ubuntu the runtime list is going to be much larger; it might be simpler just to package it up as .debs and have it compute the dependencies automatically with `dpkg-shlibdeps` (since all the Boost and Qt libs are separate packages); probably a similar story for RedHat.  Shouldn't take too long to package; should be able to do it tomorrow if you like, then we'll have homebrew and .debs, leaving just RedHat to sort out (which will likely need the superbuild stuff for Windows to pull in the unpackaged bits like glm and Qt)

joshmoore:
Hmmm..... I'm skeptical. I like the idea of the packages, don't get me wrong. But it seems like we should know the non-dev requirements that we're dealing with regardless.

rleigh-dundee:
@joshmoore It's difficult.  Take most of the -dev packages; they are generic and for the most part *unversioned* (in the package name) e.g. libtiff-dev, libboost-all-dev.  The Debian/Ubuntu package lists we have for building currently work for any version of either distribution dating back many years.  In contrast, the binary packages which the build depends on are tied to a specific distribution version and have very specific versioning.  For example, (from a BSD build I had to hand):

```
% ldd cpp/lib/ome/bioformats/libome-bioformats.so | sort
	libboost_chrono.so.1.55.0 => /usr/local/lib/libboost_chrono.so.1.55.0 (0x806135000)
	libboost_date_time.so.1.55.0 => /usr/local/lib/libboost_date_time.so.1.55.0 (0x804961000)
	libboost_filesystem.so.1.55.0 => /usr/local/lib/libboost_filesystem.so.1.55.0 (0x80336f000)
	libboost_log.so.1.55.0 => /usr/local/lib/libboost_log.so.1.55.0 (0x8040bf000)
	libboost_log_setup.so.1.55.0 => /usr/local/lib/libboost_log_setup.so.1.55.0 (0x803c76000)
	libboost_system.so.1.55.0 => /usr/local/lib/libboost_system.so.1.55.0 (0x803585000)
	libboost_thread.so.1.55.0 => /usr/local/lib/libboost_thread.so.1.55.0 (0x805f1a000)
	libc++.so.1 => /usr/lib/libc++.so.1 (0x804b6e000)
	libc.so.7 => /lib/libc.so.7 (0x80081d000)
	libcxxrt.so.1 => /lib/libcxxrt.so.1 (0x804e29000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x805269000)
	libjbig.so.2 => /usr/local/lib/libjbig.so.2 (0x8058be000)
	libjpeg.so.8 => /usr/local/lib/libjpeg.so.8 (0x805aca000)
	liblzma.so.5 => /usr/lib/liblzma.so.5 (0x80569b000)
	libm.so.5 => /lib/libm.so.5 (0x805043000)
	libome-common.so.5.1.0 => /tmp/b/cpp/lib/ome/common/libome-common.so.5.1.0 (0x8039f6000)
	libome-xml.so.5.1.0 => /tmp/b/cpp/lib/ome/xml/libome-xml.so.5.1.0 (0x801f1f000)
	librt.so.1 => /usr/lib/librt.so.1 (0x80633c000)
	libthr.so.3 => /lib/libthr.so.3 (0x805476000)
	libtiff.so.5 => /usr/local/lib/libtiff.so.5 (0x803788000)
	libxerces-c-3.1.so => /usr/local/lib/libxerces-c-3.1.so (0x8043a5000)
	libz.so.6 => /lib/libz.so.6 (0x805d05000)
```

Each of those lines is a separate package to install in Ubuntu, and from each libraries' shlibs/symbols files, will have a versioned dependency.  This needs computing for every build; `dpkg-shlibdeps` is what does that when building a .deb.  We don't know what these are in advance from installing the -dev package.  I can manually look up what we need for the current Ubuntu 14.04 build, but I can guarantee that it won't be pleasant to maintain or type in!  Using .debs will remove that burden from both us maintaining it and from the end user installing it since it's all automatically generated during the build.

joshmoore:
* dpkg-shlibdeps will generate the runtime list from the dev list?
 * how many of these packages are exposed in our public API? i.e. how many
of the headers and therefore the dev packages will end-users need to
compile anyway?

rleigh-dundee:
@joshmoore No, dpkg-shlibdeps will look at every executable and library in the package, and will obtain from them (using ldd and nm).  I.e. this is all done by introspecting the end product of the build.
- each dependent library
- each exported symbol used from each library
It will then look them up in the .shlibs and .symbols files provided in the library -dev package (`/var/lib/dpkg/info/$pkg.shlibs` etc) and use the symbol versioning and required libs to determine
- the library package(s) to depend upon
- the minimum required version of each package based upon symbol version use

This is pretty much independent of the original -dev package list; that package will usually depend upon these packages and bring them in directly or indirectly via Depends, but we don't know the details until this point--it might depends on a bunch of libraries we don't need at runtime (e.g. libboost-all-dev which we install for convenience but only use a tiny fraction of) or in the case of libtiff-dev/libpng-dev we don't know or care about which version of the library we install, so we don't know what we'll end up linking with until it's done.

WRT public API, that's an irrelevance here.  This concerns only the *ABI*  Whether they are exposed in the public API or not, they are a concrete dependencies we require to be installed to function.  If they aren't available the programs will fail to link at runtime.

joshmoore:
> @joshmoore No, dpkg-shlibdeps will look at every executable and library in the package, and will obtain from them (using ldd and nm). I.e. this is all done by introspecting the end product of the build.

Understood.

> WRT public API, that's an irrelevance here. This concerns only the ABI Whether they are exposed in the public API or not, they are a concrete dependencies we require to be installed to function. If they aren't available the programs will fail to link at runtime.

It's not an irrelevance. What I was getting at is: if someone will need the dev anyway in order to compile, then for the moment, it's less bad that we require it in the docs without us needing to do the deb work immediately. I'd still like to know the answer to my question.

rleigh-dundee:
"how many of these packages are exposed in our public API? i.e. how many of the headers and therefore the dev packages will end-users need to compile anyway?"

At the moment, everything except for libpng and libtiff, which I've taken pains to ensure are not exposed in the headers.  They use pImpl to avoid direct exposure of their headers in our classes.

The Boost usage is primarily header-only by design, so most of that is used in our headers.  Some stuff might be hidden in the implementation e.g. the spatial index/tile cache.  shared_ptr, filesystem paths, timestamps etc. are explicitly used in the public API by design. Xerces-C is exposed; our xml wrapper stuff is just to manage memory usage rather than wrap it entirely, which would be a rather large undertaking.

Right now, we don't really have much in the way of "private" implementation.  Once we get more readers, we could consider making them entirely private and only accessible via a factory.  They could be loadable modules not part of the main library.  This would keep the headers free of what's needed by individual readers, e.g. hdf5, jpeg, whatever else they need.  But on the other side, if we want users to derive readers from our own, that would be precluded without having "private" headers which expose them.

If you're primarily focussing upon users who will be developing against the build, then the current dev package list (or a subset minus the docs dependencies) will be quite fine.  It's only the final end user who they target who would need to have just the runtime libs installed.  With the existing docs, right at the top I've made it quite clear which are needed and which are optional (and for what reason), so it should be easy to omit the stuff you don't need.  I can probably add an extra sentence to explicitly point that out if it's not obvious as it stands.

rleigh-dundee:
@joshmoore Is the updated table and description better for you?

joshmoore:
Wanna build and paste a link?

rleigh-dundee:
@joshmoore Now in https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/cpp/overview.html

joshmoore:
NB: "implementation" as the bullet point for the "overview" page is a bit weird. We may also be approaching a "prereqs" page, but I leave that sort of thing to @hflynn.

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -26,57 +26,100 @@ Prerequisites
...
+  Optional for Bio-Formats build; if used for the Bio-Formats build,
+  headers may be required for client build and libraries and any data
+  files required for deployment
 
 \*
-  Optional
+  Optional, needed to build the OpenGL image viewer and client applications
 †
-  Optional, needed to build the API reference
+  Optional, needed only for building
```
joshmoore: It this still needed? Doesn't it equal the empty dot under `BF Build`?

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -26,57 +26,100 @@ Prerequisites
...
-+----------------+--------------+--------------+
-| Java‡          | 1.7.x        | 1.6.x        |
-+----------------+--------------+--------------+
-| Python Sphinx‡ | 1.2.x        | 1.1.x        |
-+----------------+--------------+--------------+
-| TeX (XeLaTeX)‡ | TeXLive 2014 | TeXLive 2012 |
-+----------------+--------------+--------------+
+but are not supported.  Some packages are required only for building
+Bio-Formats (*BF build*).  A subset of these are required for building
+client applications making use of Bio-Formats (*Client build) For
```
joshmoore: missing `*`.

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -26,57 +26,100 @@ Prerequisites
...
+Quick start
+^^^^^^^^^^^
+
+Install the following packages to build Bio-Formats C++.  A subset of
+these packages (or their dependencies) may be used for deployment,
+where the development package headers and tools for building
+documentation etc. are not required.
+
+BSD Ports
+  run ``pkg install devel/boost-all devel/cmake science/hdf5 graphics/png lang/python textproc/py-genshi graphics/tiff textproc/xerces-c3 devel/git devel/googletest math/glm devel/qt5 graphics/graphviz devel/apache-ant java/openjdk7 textproc/py-sphinx print/texlive-full``
```
joshmoore: These look great, but I find the "run" to be a bit distracting. If it could be removed or be more visually separated from the rest that'd be great.

joshmoore:
General additions are great. Thanks. Few comments on the commits.

rleigh-dundee:
@joshmoore All the above points should now be fixed.

rleigh-dundee:
@joshmoore OK to merge with the above fixes?

joshmoore:
LGTM.

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -613,13 +719,15 @@ C++11
...
-:cpp:class:`ome::compat` namespace, for example as
-:cpp:class:`ome::compat::shared_ptr`.
-
+using a C++11 or C++14 compiler, or when ``-Dcxxstd-autodetect=ON`` is
+used and the compiler can be put into a C++11 or C++14 compatibility
+mode.  When using an older compatbility mode such as C++98, the Boost
+equivalents of C++11 library features will be used as fallbacks to
+provide the same functionality.  In both cases these types are
+imported into the :cpp:class:`ome::compat` namespace, for example as
+:cpp:class:`ome::compat::shared_ptr`, and the types in this namespaces
```
mtbc: "this namespaces"

rleigh-dundee:
--no-rebase

PR #1684, opened 2015-03-23
Improve the units API

ctrueden:
Specifically, this branch makes the `Quantity` superclass nicer to work with, and adds some helpful constructors with default unit specifications, to avoid errors when converting downstream code to the new API.

See [here](https://github.com/scifio/scifio-ome-xml/compare/ome-5.1) for an example where these changes would be helpful.

/cc @qidane

### DIFF
--- a/components/ome-xml/src/ome/units/quantity/Angle.java
+++ b/components/ome-xml/src/ome/units/quantity/Angle.java
@@ -66,6 +66,7 @@ public class Angle extends Quantity implements Comparable<Angle>
     hashCodeValue = SEED2 * hashCodeValue + unit.getSymbol().hashCode();
   }
 
+  @Override
   public Number value()
   {
     return value;
@@ -84,6 +85,7 @@ public class Angle extends Quantity implements Comparable<Angle>
     return null;
   }
 
+  @Override
   public boolean equals(Object other)
   {
     if (other == null)
@@ -138,6 +140,7 @@ public class Angle extends Quantity implements Comparable<Angle>
     return result.toString();
   }
 
+  @Override
   public Unit<ome.units.quantity.Angle> unit()
   {
     return unit;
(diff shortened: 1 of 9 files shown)


ctrueden:
I don't love the `Unit<? extends Quantity> unit()` declaration, but it's the best you can do without introducing a generic parameter to `Quantity`.

ctrueden:
As an aside, it is shame that the OME codebase could not adopt a standard units library. We will need to create an integration layer to bridge OME units to and from SciJava units.

ctrueden:
OK, I see that there are a bunch of new `FormatTools` method signatures that return new quantities with default units. And I guess, if the default unit differs in the schema depending on which field it is, then the new constructors I added above make less sense and are in fact dangerous. Interested to hear others' thoughts.

melissalinkert:
I'd prefer to see 17300c4 reverted, but the other changes read fine.

ctrueden:
@melissalinkert Are there ever two fields in OME-XML that use the same quantity but different units? If so, I will purge the commit. But if not, what's the harm?

If we do eliminate that commit, I would still like to see shorter API calls to create new quantities. As things stand, it is a little verbose to write `meta.setPhysicalSizeX(FormatTools.getPhysicalSizeX(pSizeX)`.

melissalinkert:
```PhysicalSizeX``` on ```Pixels```, ```PositionX``` on ```Plane```, and ```EmissionWavelength``` on ```Channel``` all use the ```Length``` type, but with different default units (microns, reference frame, and nanometers, respectively).

ctrueden:
Thanks @melissalinkert, I purged 17300c4e531670e437f27dbc8190fdecbc1dd303 from the branch.

melissalinkert:
Thanks.  All reads OK now, just waiting on inclusion in the merge builds.

PR #1685, opened 2015-03-23
cpp: cmake: Special-case bf-test shell wrapper rootdir determination for empty path

rleigh-dundee:
- Handle the case where stripping the bindir results in an empty string (set to '.')
- Also handle the case where `dirname` does not have a leading slash
- Use `eval` to strip substring with dynamic string; this makes it clear where the string to strip comes from rather than having cmake substitute it in multiple places

--no-rebase

/cc @joshmoore 

--------

Testing:

Obtain and unpack merge build tar archive.

```
tar xfv bioformats-xxx.tar.xz
cd bioformats-xxx
./bin/bf-test
bin/bf-test
$(pwd)/bin/bf-test
```

All three forms of `bf-test` invocation should work, where previously the middle form was broken.

### DIFF
--- a/cpp/cmake/TemplateShellWrapper.cmake.in
+++ b/cpp/cmake/TemplateShellWrapper.cmake.in
@@ -132,7 +132,11 @@ if [ -d "$INSTALL_PREFIX" ]; then
   fi
 else
   bindir=$(dirname "$prog")
-  rootdir=${bindir%@CMAKE_INSTALL_FULL_BINDIR@}
+  rootdir=$(eval echo "\${bindir%$INSTALL_BINDIR}")
+  rootdir=${rootdir%/}
+  if [ -z "$rootdir" ]; then
+    rootdir=$(pwd)
+  fi
 
   PATH="${rootdir}${INSTALL_FULL_BINDIR}:$PATH"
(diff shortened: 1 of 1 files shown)


mtbc:
I wasn't sure which merge build to look at, but locally from the directory into which I installed,
```
mtbc@ls28101:/tmp/install$ ./bin/bf-test
usage: ./bin/bf-test [--help] [--usage] <cmd> [<args>]

available bf-test commands in '/tmp/install/libexec/ome/bioformats'

	info  view
mtbc@ls28101:/tmp/install$ bin/bf-test
usage: bin/bf-test [--help] [--usage] <cmd> [<args>]

available bf-test commands in '/tmp/install/libexec/ome/bioformats'

	info  view
mtbc@ls28101:/tmp/install$ $(pwd)/bin/bf-test
usage: /tmp/install/bin/bf-test [--help] [--usage] <cmd> [<args>]

available bf-test commands in '/tmp/install/libexec/ome/bioformats'

	info  view
mtbc@ls28101:/tmp/install$ PATH=/tmp/install/bin:$PATH bf-test
usage: /tmp/install/bin/bf-test [--help] [--usage] <cmd> [<args>]

available bf-test commands in '/tmp/install/libexec/ome/bioformats'

	info  view
```
:+1: Good to merge.

PR #1686, opened 2015-03-23
build: Also provide source zip without binary content

rleigh-dundee:
- Add "bioformats-dfsg-$version.zip"; this is a DFSG (Debian Free Software Guidelines) compliant archive which omits binaries (jars and libraries such as .so/.dylib/.dll for which we don't have source in the repository and hence in the source distribution).  This is to make a "clean" source release which is acceptable for public redistribution in compliance with the GPL, for use by upstream homebrew, Debian/Ubuntu/Fedora, BSD ports, etc.

This change is a precursor to being able to provide Bio-Formats (both Java and C++) in distributions directly.  Are there any other generated files imported into git which might also need excising for licence compliance?  How about source files under BSD/GPL-incompatible terms?  Is there a better prefix name than "-dfsg" (which is already used elsewhere for sanitised versions of upstream sources with problematic content stripped out) ?  (I originally used "-clean")

### DIFF
--- a/ant/toplevel.xml
+++ b/ant/toplevel.xml
@@ -1029,11 +1029,21 @@ Type "ant -p" for a list of targets.
   <target name="init"/>
 
   <!-- Source release -->
-  <target name="release" description="zip the git release as bioformats-${release.version}.zip" depends="release-version">
+  <target name="release" description="zip the git release as bioformats-[dfsg-]${release.version}.zip" depends="release-version">
     <mkdir dir="${artifact.dir}"/>
     <exec executable="python" failonerror="true">
       <arg value="${root.dir}/tools/source-archive.py"/>
       <arg value="--release=bioformats"/>
+      <arg value="--bioformats-shortversion=${release.shortversion}"/>
+      <arg value="--bioformats-version=${release.version}"/>
+      <arg value="--bioformats-vcsrevision=${vcs.revision}"/>
+      <arg value="--bioformats-vcsdate=${vcs.date}"/>
+      <arg value="--bioformats-vcsdate-unix=${vcs.date_unix}"/>
+      <arg value="--target=${artifact.dir}"/>
+    </exec>
+    <exec executable="python" failonerror="true">
+      <arg value="${root.dir}/tools/source-archive.py"/>
+      <arg value="--release=bioformats-dfsg"/>
       <arg value="--bioformats-shortversion=${release.shortversion}"/>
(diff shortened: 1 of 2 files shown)


sbesson:
With regard to the hyphen changes:
- this change seems to depart from the policy that was discussed/agreed in https://github.com/openmicroscopy/openmicroscopy/pull/1712 where the source archive is named after the repository. Do we all agree on this? What is the impact on OMERO: should the source archive also be renamed for branding/consistency?
- given the deadline, the impact of this change (all C++ build (merge and latest), all Bio-Formats release builds and ome-release PR) and since we have already tagged m5 which includes all breaking changes, I suggest we schedule this properly for 5.1.1.

I would revert this hyphen change and move forward with the Debian zip.

rleigh-dundee:
@sbesson This was following a bit of discussion with @joshmoore yesterday about the consistency of the naming.  I certainly prefer the existing (no hyphen) scheme myself, since it's easier to type and works in C++ namespaces and other places where hyphens aren't allowed or are already used for other purposes, and I agree it's a bit late for such changes.  I've backed pulled out this change, so this PR is now just the addition of a cleaned up source zip.

joshmoore:
Just to be clear, I was primarily concerned with the **cpp** package name, i.e.
```
bioformats-cpp-5.1.0-m5-Release-Ubuntu14.04-x86_64
```
and I think that will need to be addressed, but yes, in 5.1.1, which is why I only suggested creating a card for discussion.

rleigh-dundee:
Restored use of hyphens, dropped genshi embedding.  gtest could also potentially be dropped with a little tweak to the cmake logic to not fall back to it unconditionally.

joshmoore:
The non-dfsg zip contains everything in the dfsg zip, and the files which are missing are broken down:
```
[jamoore@ome-c6100-3 artifacts]$ for x in $(comm -23 /tmp/bioformats-*sorted); do echo ${x##*.}; done | sort | uniq -c
      3 dll
      3 dylib
      3 html
     36 jar
     38 py
      2 so
      3 txt
```
and all of the py and txt files are from genshi:
```
comm -23 /tmp/bioformats-*sorted | grep -E "(py|txt|html)" | grep -v genshi
```
So in general, this new zip looks fine to me.

melissalinkert:
I assume this is good to merge then?  @sbesson?

rleigh-dundee:
--no-rebase

PR #1688, opened 2015-03-24
JHDFServiceImpl: catch checked exceptions

ctrueden:
This fixes the following build errors on the command line:

    [javac] /Users/curtis/code/ome/bf5/components/formats-gpl/build/src/loci/formats/services/JHDFServiceImpl.java:172: error: unreported exception HDF5JavaException; must be caught or declared to be thrown
    [javac]         return this.hdfReader.string().readArray(path);
    [javac]                                                 ^
    [javac] /Users/curtis/code/ome/bf5/components/formats-gpl/build/src/loci/formats/services/JHDFServiceImpl.java:179: error: unreported exception HDF5JavaException; must be caught or declared to be thrown
    [javac]         return this.hdfReader.readCompoundArray(path, HDF5CompoundDataMap.class);
    [javac]                                                ^

### DIFF
--- a/components/formats-gpl/src/loci/formats/services/JHDFServiceImpl.java
+++ b/components/formats-gpl/src/loci/formats/services/JHDFServiceImpl.java
@@ -33,11 +33,11 @@ import java.util.List;
 import java.util.StringTokenizer;
 import java.util.Vector;
 
+import ncsa.hdf.hdf5lib.exceptions.HDF5JavaException;
 import loci.common.Constants;
 import loci.common.Location;
 import loci.common.services.AbstractService;
 import loci.common.services.ServiceException;
-
 import ch.systemsx.cisd.base.mdarray.MDByteArray;
 import ch.systemsx.cisd.base.mdarray.MDIntArray;
 import ch.systemsx.cisd.base.mdarray.MDShortArray;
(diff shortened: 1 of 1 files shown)


ctrueden:
Both Maven and Ant have the aforementioned compile errors. But the bizarre thing is: the `formats-gpl` project has no such build errors in Eclipse. But if you dig in to the `readArray` and `readCompoundArray` methods of `IHDF5SimpleReader`, it definitely throws `HDF5JavaException`, which is a checked exception.

Also, I don't think it's particularly great to throw `RuntimeException`, but we need to use some sort of unchecked exception there—otherwise the `JHDFService` API contract would need updating to throw a checked exception for the two relevant methods.

melissalinkert:
I would assume this has never been an issue before because ```ncsa.hdf.hdf5lib.exceptions.HDF5JavaException``` extends ```ncsa.hdf.hdf5lib.exceptions.HDF5Exception```, which extends ```RuntimeException```.  I'd guess that it's a compiler bug, given that no other systems show this issue (and you only see it on the command line).

ctrueden:
@melissalinkert I found the culprit:
```
$ ls -l /Library/Java/Extensions/jhdf5.jar
-rwxrwxr-x  1 curtis  admin  49525 Aug 19  2009 /Library/Java/Extensions/jhdf5.jar*
$ javap -cp /Library/Java/Extensions/jhdf5.jar ncsa.hdf.hdf5lib.exceptions.HDF5Exception
Compiled from "HDF5Exception.java"
public class ncsa.hdf.hdf5lib.exceptions.HDF5Exception extends java.lang.Exception {
  protected java.lang.String detailMessage;
  public ncsa.hdf.hdf5lib.exceptions.HDF5Exception();
  public ncsa.hdf.hdf5lib.exceptions.HDF5Exception(java.lang.String);
  public java.lang.String getMessage();
}
```
So apparently, something installed old versions of several libraries—including JHDF5, JOGL and Java3D—on my system. I compared with @hinerm's installation, and he has none of that stuff, so I'm just going to `rm -rf` those things, and close this PR. Very sorry for the noise!

PR #1689, opened 2015-03-25
cmake: Add Bio-Formats-specific paths to cache to allow user configuration

rleigh-dundee:
This allows our paths to be configured/overridden by the end user.  This is needed for Debian/Ubuntu packaging.

--------

Testing: Builds should remain green.

Configure by adding `"-DOME_BIOFORMATS_INSTALL_LIBEXECDIR=testpath -DCMAKE_INSTALL_PREFIX:PATH="` to the cmake command-line.  Check `OME_BIOFORMATS_INSTALL_*LIBEXECDIR` in `CMakeCache.txt`, `cpp/lib/ome-internal/config.h` and `bin/bf-test` (names may differ slightly but will all contain LIBEXECDIR).
After running `make DESTDIR=/tmp/test`,`/tmp/test/bin/bf-test --usage` should report the programs under `testpath`, and `/tmp/test/bin/bf-test info [--help]` should continue to work.
In a new build directory, repeat the cmake run, but omit `-DCMAKE_INSTALL_PREFIX:PATH=` (or add an absolute path such as `/tmp/testpath`), you should see the OME_BIOFORMATS_INSTALL_FULL_LIBEXECDIR set to what you specified, with the install prefix (/usr/local or your own) prepended.

This is testing that OME_BIOFORMATS_INSTALL_LIBEXECDIR (and the OME_BIOFORMATS_INSTALL paths) to be configured when running cmake.


mtbc:
What kind of testpath makes sense for LIBEXECDIR? (I'm not really sure what all this is.)

rleigh-dundee:
@mtbc You can set it to anything you like.  The default is "libexec/ome/bioformats"; for Debian/Ubuntu, it's set to lib/$arch/ome/bioformats/bin due to filesystem requirements there.  Similar to the PostgreSQL binaries.

mtbc:
```
OME_BIOFORMATS_INSTALL_LIBEXECDIR:UNINITIALIZED=testpath
```
```
#define OME_BIOFORMATS_INSTALL_LIBEXECDIR "testpath"
#define OME_BIOFORMATS_INSTALL_FULL_LIBEXECDIR "/testpath"
```
```
BF_INSTALL_LIBEXECDIR="testpath"
BF_INSTALL_FULL_LIBEXECDIR="/testpath"
```

mtbc:
```
mtbc@ls28101:~/src/openmicroscopy$ /tmp/test/bin/bf-test --usage
usage: /tmp/test/bin/bf-test [--help] [--usage] <cmd> [<args>]

available bf-test commands in '/tmp/test/testpath'

	info  view
```
and `info` still works, etc.

mtbc:
`view` is fine too so icon path must be okay.

Good to merge.

joshmoore:
https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-cpp/677/ :thumbsup:

rleigh-dundee:
--no-rebase

PR #1693, opened 2015-03-26
Memoizer docs

sbesson:
This Pull Request adds tests and documentation for the memoization functionalities of Bio-Formats

Using MATLAB as an example consumer where this memoization has been tested, the
description is amended to:
- describe a minimal example of usage of Memoizer
- explain Bio-Formats is not thread safe
- provide a example of Memoizer usage in a multi-threaded environment

The TestMemoizer tests are expanded to cover the examples described in the documentation, including its usage in multi-threaded mode.

To test this PR:
- verify the CI job running the MATLAB tests remains green
- review the MATLAB section of the developer documentation for Bio-Formats

/cc @emilroz @hflynn 

--no-rebase

PR #1694, opened 2015-03-26
Display private fields and methods in Javadocs

melissalinkert:
See gh-1693.  Private fields (e.g. ```minimumElapsed``` in Memoizer) should be present in the Javadocs with this PR included.  Feel free to cherry pick the commit, or close outright if this isn't helpful.

### DIFF
--- a/ant/global.xml
+++ b/ant/global.xml
@@ -11,6 +11,22 @@ Type "ant -p" for a list of targets.
   <property file="${user.home}/.ant-global.properties"/>
   <property file="${root.dir}/ant/global.properties"/>
 
+  <!-- Convenient JDK version properties -->
+
+  <available property="jdk1.6+" classname="java.util.ServiceLoader"/>
+  <available property="jdk1.7+" classname="java.util.Objects"/>
+  <available property="jdk1.8+" classname="java.util.stream.IntStream"/>
+
+  <target name="javadoc-properties">
+    <if>
+      <isset property="jdk1.8+"/>
+      <then>
+        <property name="javadoc.doclint" value="-Xdoclint:none"/>
+      </then>
+    </if>
+    <property name="javadoc.doclint" value=""/>
+  </target>
+
   <!-- Convenient platform properties -->
 
   <condition property="isUnix">
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -187,7 +187,7 @@ your FindBugs installation's lib directory. E.g.:
     </javac>
   </target>
 
-  <target name="docs" depends="compile"
+  <target name="docs" depends="compile, javadoc-properties"
     description="generate javadocs">
     <javadoc sourcepath="${dest.dir}" destdir="${docs.dir}/api"
       classpath="${component.classpath}:${classes.dir}"
@@ -195,7 +195,7 @@ your FindBugs installation's lib directory. E.g.:
       author="true" version="true" use="true"
       access="private"
       nodeprecated="false" windowtitle="Bio-Formats API"
-      additionalparam="-Xdoclint:none">
+      additionalparam="${javadoc.doclint}">
       <packageset dir="${dest.dir}"/>
       <doctitle><![CDATA[<h1>Bio-Formats</h1>]]></doctitle>
       <bottom><![CDATA[${copyright.begin} ${YEAR} ${copyright.end}]]></bottom>
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -193,6 +193,7 @@ your FindBugs installation's lib directory. E.g.:
       classpath="${component.classpath}:${classes.dir}"
       encoding="UTF-8"
       author="true" version="true" use="true"
+      access="private"
       nodeprecated="false" windowtitle="Bio-Formats API">
       <packageset dir="${dest.dir}"/>
       <doctitle><![CDATA[<h1>Bio-Formats</h1>]]></doctitle>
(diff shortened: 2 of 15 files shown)


sbesson:
Because Javadoc links will not be public before we release Bio-Formats 5.1.0 and its API documentation, simultaneously changing gh-1693 to point at a to-be-exposed private field would be untested and potentially result in broken links. That being said, I do not see any strong reason NOT to expose private fields in the Javadocs  and link to this private fields in subsequent patch releases. /cc @hflynn

https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/769/ is yellow because this PR exposes new Javadoc warnings. I will try to open a PR against this branch to fix them.

hflynn:
Yeah, I'm happy to do a patch release once the API docs are live - assuming the links are correct, we can have it live on the web docs on release day even.

melissalinkert:
Javadoc warnings fixes merged in.

sbesson:
`-Xdoclint:none` addition caused https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build-win/592/ to fail and may need to be reverted.

mtbc:
There are docs at http://downloads.openmicroscopy.org/bio-formats/5.1.0-m5/api/ but I don't know where the open PRs are staged so I'll rely on a local build once GitHub permits.

hflynn:
I don't think there is a staging build for these.

mtbc:
Locally (on Linux) if I do,
```
git checkout 7d079dd5
git merge 7e7a7c52
git clean -dfx
ant docs
```
then I get,
```
  [javadoc] Skipping /home/mtbc/src/bioformats/components/stubs/mipav-stubs/build/src since it is no directory.
  [javadoc] /home/mtbc/src/bioformats/components/forks/turbojpeg/build/src contains source files in the default package, you must specify them as source files not packages.
  [javadoc] Generating Javadoc
  [javadoc] Javadoc execution
  [javadoc] usage: javadoc [options] [packagenames] [sourcefiles] [@files]
  [javadoc] javadoc: error - invalid flag: -Xdoclint:none
...
  [javadoc] 1 error

BUILD FAILED
/home/mtbc/src/bioformats/ant/toplevel.xml:172: Javadoc returned 1
```
which looks like the issue noted by @sbesson.

rleigh-dundee:
@sbesson Take a look at

https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/antlib/resources/global.xml#L233
https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/antlib/resources/global.xml#L256
https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/antlib/resources/lifecycle.xml#L381

for how to autodetect and enable this option with jdk 1.8.

sbesson:
@rleigh-dundee: thanks for the pointers. A similar strategy should definitely be adapted here (but I won't have time before 5.1.0 under all likelihood). In the meantime, reverting 37443e3 should be enough to let the CI jobs pass with JDK < 1.8.

rleigh-dundee:
@sbesson No problem.  It's fixed in https://github.com/melissalinkert/bioformats/pull/25

mtbc:
This is now working just fine on Java 1.7. :+1: Two small concerns:

1. There are private method comments that don't seem to have much anticipated being Javadoc'd. They might be worth a (very boring) review someday to better-format todo notes, etc.
1. Few people reading the API docs care about the privately scoped parts, but there are often so many that the other stuff is lost among it. This might actually be worth two different docs builds, depending on if the developer -- who will typically want to focus on the stable API -- even wants to see the private stuff that will often dominate it.

Still, if we are going this way, then I don't see any blockers. Someone should check on Java 1.8 of course. (I don't even have an OpenJDK 8 with Jessie.)

mtbc:
@sbesson: Is the doclint argument required to prevent a failed build in Java 8, so if it seems to work then probably the option is being properly used? (I don't know if over time we plan to fix the docs such that we don't need the option any more.)

mtbc:
I have been testing with "ant docs". Does that suffice, or ought I also test some maven target?

mtbc:
BIOFORMATS-5.1-merge-build-win is now green. :+1:

rleigh-dundee:
@mtbc "ant docs" should be sufficient; the maven build is untouched by this change.

BTW, for Java8 testing on Debian you could just drop the Oracle JDK8 into /opt and temporarily add to your PATH/JAVA_HOME; it worked for me.

manics:
> Few people reading the API docs care about the privately scoped parts, but there are often so many that the other stuff is lost among it. This might actually be worth two different docs builds, depending on if the developer -- who will typically want to focus on the stable API -- even wants to see the private stuff that will often dominate it.

The other advantage of a separate private-docs build target is it reduces the temptation for people to use implementation-dependent functionality in their code.

dominikl:
No problems noticed when built with Java 8 (tested with 1.8.0_25)

jburel:
java version "1.8.0_31" no errors noticed

melissalinkert:
This is good to merge then?

mtbc:
From me, no blockers to merging.

melissalinkert:
OK, merging then.  Reminder card for doing the private Javadoc review: https://trello.com/c/U8HDtN3O/159-review-private-method-javadocs

melissalinkert:
--no-rebase

PR #1699, opened 2015-03-30
MetamorphHandler: Whitelist valid keys in Description

rleigh-dundee:
If a key isn't present in the whitelist, treat it as part of the free text comment.  I have also updated the main reader to use the handler logic to parse the description as well.  There may be a slight behaviour change here--the old code assumed that '=' was a separator for `Exposure`, but I only see `Exposure: <value> <unit>` in the samples I've checked, so not sure of its validity--might be in some particular samples?

--no-rebase
--exclude

/cc @emmenlau

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
@@ -63,6 +63,15 @@ public class MetamorphHandler extends BaseHandler {
   private String stageLabel;
   private Double gain;
 
+  private String[] mDesciptionKeyList = {
+    "Plate Screen", "Acquired from Photometrics", "Barcode:",
+    "Experiment base name:", "Experiment set:", "Exposure:",
+    "Binning:", "Region:", "Subtract:", "Shading:", "Digitizer:",
+    "Gain:", "Camera Shutter:", "Clear Count:", "Clear Mode:",
+    "Frames to Average:", "Trigger Mode:", "Temperature:"
+  };
+
+
   // -- Constructor --
 
   public MetamorphHandler() {
@@ -133,6 +142,8 @@ public class MetamorphHandler extends BaseHandler {
         String k = null, v = null;
 
         if (value.indexOf(delim) != -1) {
+          String vDescription = null;
+
           int currentIndex = -delim.length();
           while (currentIndex != -1) {
             currentIndex += delim.length();
@@ -147,14 +158,46 @@ public class MetamorphHandler extends BaseHandler {
             }
             currentIndex = nextIndex;
 
-            int colon = line.indexOf(":");
-            if (colon != -1) {
-              k = line.substring(0, colon).trim();
-              v = line.substring(colon + 1).trim();
-              if (metadata != null) metadata.put(k, v);
-              checkKey(k, v);
+            if (line.isEmpty()) {
+              continue;
+            }
+
+            // We use a dictionary to identify the standard MetaMorph keys:
+            boolean vLineIsMetaMorphStandard = false;
+            for (String vDescriptionKey : mDesciptionKeyList) {
+              if (line.startsWith(vDescriptionKey)) {
+                vLineIsMetaMorphStandard = true;
+                break;
+              }
+            }
+            if (vLineIsMetaMorphStandard) {
+              // The majority of tags can be split at the colon:
+              int colon = line.indexOf(":");
+              if (colon != -1) {
+                k = line.substring(0, colon).trim();
+                v = line.substring(colon + 1).trim();
+                if (metadata != null) metadata.put(k, v);
+                checkKey(k, v);
+              } else {
+                // There was no colon, so the key is a full line:
+                k = line.trim();
+                v = line.trim();
+                if (metadata != null) metadata.put(k, v);
+                checkKey(k, v);
+              }
+            } else {
+              // This is NOT a standard tag in the description field, so it
+	      // must be a multi-line description from the user (i.e. from
+	      // the MetaXpress free-text "Description" field in the screening
+	      // application).
+              if (vDescription == null) {
+                vDescription = line;
+              } else {
+                vDescription = vDescription + "\n" + line;
+              }
             }
           }
+          if (metadata != null && vDescription != null) metadata.put("Description", vDescription);
         }
         else {
           int colon = value.indexOf(":");
--- a/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
@@ -133,16 +133,18 @@ public class MetamorphHandler extends BaseHandler {
     String k = null, v = null;
 
     String delim = " #13; #10;";
-    if (desc != null && desc.indexOf(delim) < 0) {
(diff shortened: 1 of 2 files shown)


rleigh-dundee:
Some samples aren't parsing correctly due to using an incorrect separator for the reader, which I'll look at tomorrow.

rleigh-dundee:
Parsing fixed, last issue is the fact that something is causing parsing of the metadata to occur multiple times, which will need looking into.  The effect is that the Description key repeats its value multiple times; looks like once per line

`/srv/data/from_skyking/metamorph/dave/Z Series.stk; field is "Multiplication Gain" (missing from mDesciptionKeyList)

sbesson:
This PR is failing some tests in https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-build/787/testReport/

rleigh-dundee:
Should be good to review now; tests pass locally for me.

rleigh-dundee:
@melissalinkert The Description metadata duplication is due to the fact that in `MetamorphReader`, this code:

```
    // parse (mangle) TIFF comment
    String descr = ifds.get(0).getComment();
```

is returning odd data.  It is returning the comment, but the comment is repeated many times over.

```
Exposure: 100 ms
Binning: 1 x 1
Region: 512 x 512, offset at (0, 0)
Acquired from Photometrics
Subtract: Off
Shading: Off
Digitizer: 5 MHz (EM Gain)
Gain: Gain 2 (2x)
Multiplication Gain: 4095
Camera Shutter: Always Open
Clear Count: 2
Clear Mode: CLEAR PRE SEQUENCE
Frames to Average: 1
Trigger Mode: Normal (TIMED)
Temperature: -70.25

Exposure: 100 ms
Binning: 1 x 1
Region: 512 x 512, offset at (0, 0)
Acquired from Photometrics
Subtract: Off
Shading: Off
Digitizer: 5 MHz (EM Gain)
Gain: Gain 2 (2x)
Multiplication Gain: 4095
Camera Shutter: Always Open
Clear Count: 2
Clear Mode: CLEAR PRE SEQUENCE
Frames to Average: 1
Trigger Mode: Normal (TIMED)
Temperature: -70.25

<repeated a further 19 times!>
```

Interestingly, `tools/tiffcomment` shows all the duplicated entries, while `tiffinfo` displays only the first, with a warning about `NUL` causing truncation:

```
% tiffinfo /srv/data/from_skyking/metamorph/dave/Z\ Series.stk 
TIFFReadDirectory: Warning, Unknown field with tag 317 (0x13d) encountered.
TIFFReadDirectory: Warning, Unknown field with tag 33628 (0x835c) encountered.
TIFFReadDirectory: Warning, Unknown field with tag 33629 (0x835d) encountered.
TIFFReadDirectory: Warning, Unknown field with tag 33630 (0x835e) encountered.
TIFFReadDirectory: Warning, Unknown field with tag 33631 (0x835f) encountered.
TIFFFetchNormalTag: Warning, ASCII value for tag "ImageDescription" contains null byte in value; value incorrectly truncated during reading due to implementation limitations.
TIFFFetchNormalTag: Warning, ASCII value for tag "DateTime" contains null byte in value; value incorrectly truncated during reading due to implementation limitations.
TIFF Directory at offset 0xa891d3 (11047379)
```

Looks like this data is actually per-plane with `NUL`-separators...  Not sure why they couldn't have used `RS` for this.  The question here is how best to handle this?  Add each as a list to get numbered per-plane global metadata or add this to the series metadata instead, or just process the first block and ignore the rest?

sbesson:
Quickly tested on an example dataset (`data_repo/from_skyking/metamorph/sebastien-besson/A._WTChlor.2.nd_[Stage1__Position52_]`). Without this PR, some metadata like `Gain` modified by this PR was correctly parsed as key/value pairs in the Series metadata and stored in the DB while importing into OMERO. With this PR, the Gain key/value pairs are now aggregated as part of the `Description` of the Series Metadata.

screen shot 2015-04-21 at 09 19 25

This metadata issue should be fixed before this PR can be merged. Given the size of the changeset, most of the Metamorph samples should likely be reviewed with/without this PR.

rleigh-dundee:
@sbesson Definitely agreed, the whitelist isn't currently comprehensive and needs expanding.  Unfortunately, I wasn't sure what was "custom" free text and what was real metadata to know to add them; it's kind of nasty not knowing due to the nature of the format.

melissalinkert:
Given the late hour, let's leave this out of 5.1.1 - can re-evaluate for 5.1.2.

joshmoore:
@sbesson suggests this was excluded because it was breaking per-series. Post 5.1.2 this will need unexcluding, possibly in the breaking queue.

joshmoore:
@rleigh-dundee : I've put `--exclude` in the description, so once you've pushed commits, you can re-include this yourself.

sbesson:
This is currently conflicting and needs additional work. Closing for now.

emmenlau:
Noo, please don't just close! I would need this feature and most of the code seemed ready?!

All the best Mario




On October 22, 2015 9:23:05 PM GMT+02:00, "Sébastien Besson" <notifications@github.com> wrote:
>Closed #1699.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/openmicroscopy/bioformats/pull/1699#event-443030890

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sbesson:
Hi Mario,

unfortunately, as described [in a previous comment](https://github.com/openmicroscopy/bioformats/pull/1699#issuecomment-94700445), the code changes proposed in http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-March/005162.html and implemented as part this PR seem to work fine for MetaXpress filesets but cause regressions when applied to Metamorph filesets.

More generally, the fact that both Metamorph and MetaXPress filesets use a priori the same specification but have clear variations in the way the metadata is stored makes the `MetamorphReader` one of the most complex readers for us to maintain without breaking our community. In terms of metrics, every code change is tested every day against ~150 different Metamorph/MetaXpress filesets which are our representation of the diversity of implementations of this particular format. 

We have created a [card](https://trello.com/c/efn0T64w/18-metaxpress-multi-line-descriptions) to investigate this issue and try to come up with a solution to be delivered as part of Bio-Formats 5.1.6.

Best,
Sebastien

emmenlau:
Dear Sebastien,

I understand and agree to your comments! But please note that my
underlying assumption is that the code for (at least) MetaXpress
is not correct. It will return sometimes key-value-pairs that should
correctly be part of a multi-line Image Description key. So I do
fully understand that you want to keep the existing code working,
but please also note that the current behaviour is already broken
for me (and for certain files, with the rare multi-line Image
Description :-) :-) ).

My further assumption would be that the code in http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-March/005162.html
would work for all MeaXpress and MetaMorph files if only the whitelist
of keys is complete! Could it be that you are just lacking a few keys
in the whitelist, and this might be easy to solve?

All the best, Mario

sbesson:
Hi Mario,

We definitely agree that the multi-line description is an issue that deserves to be fixed. As said above, we are planning to have this work completed for Bio-Formats 5.1.6 and will notify you as soon as a PR is opened.

Best,
Sebastien

PR #1711, opened 2015-04-03
Roi handler update C Z T

bramalingam:
Testing Instructions:

1) Update the Bioformats_Package.jar in your imageJ plugins folder.
2) Now open imageJ --> open an image, add ROI's to the same (add them to the ROIManager.
3) Export the image as an ome-tiff, using the Plugins-->Bio-Formats-->Bio-Formats Exporter (keep the default options)
4) The image should get exported with the ROI's.

5) Now : Reimport the image using the Bio-Formats Importer : in the import options menu, keep the following checkboxes selected : Display OME-XML metadata and Display ROIs.
6) This should import your image and add the ROI's to the ROIManager.

To Check 
1) Check the OME-XML to see if the newly added ROI's have the correct Z,C and T information.

Also try the following,
1) Open an Image from OMERO using the Omero-IJ plugin.
2) Check the OME-XML to see if the newly added ROI's have the correct Z,C and T information.

ex:

```
<ROI ID="ROI:0:0" xmlns="http://www.openmicroscopy.org/Schemas/ROI/2015-01">
<Union>
<Shape ID="Shape:0:0" StrokeColor="-1" Text="0611-0860" TheC="1" TheT="1" TheZ="1">
<Rectangle Height="390.0" Width="780.0" X="470.0" Y="416.0"/>
<ROI ID="ROI:1:0" xmlns="http://www.openmicroscopy.org/Schemas/ROI/2015-01">
<Union>
<Shape ID="Shape:1:0" Text="1162-2035" TheC="1" TheT="1" TheZ="1">
<Rectangle Height="584.0" Width="250.0" X="1910.0" Y="870.0"/>
```

Without this PR:
The ROI's will not carry any Z,C or T information in the OME-XML.

Note:
If you want ImageJ to display the ROI's according to the slice information, please select the following option in the ROIManger : 
ROIManger--> More >> --> Options --> Associate "Show All" ROIs with slices

This will associate the information while viewing the Image as well.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -308,9 +308,15 @@ public class ROIHandler {
                         roi.setName(shapeObject.getID());
 
                         if (Prefs.showAllSliceOnly){
-                            if(shapeObject.getTheC() != null);c = shapeObject.getTheC().getValue();
-                            if(shapeObject.getTheZ() != null);z = shapeObject.getTheZ().getValue();
-                            if(shapeObject.getTheT() != null);t = shapeObject.getTheT().getValue();
+                            if(shapeObject.getTheC() != null){
+                                c = shapeObject.getTheC().getValue();
+                            }
+                            if(shapeObject.getTheZ() != null){
+                                z = shapeObject.getTheZ().getValue();
+                            }
+                            if(shapeObject.getTheT() != null){
+                                t = shapeObject.getTheT().getValue();
+                            }
                             roi.setPosition(c, z, t);
                         }
                         roi.setImage(images[imageNum]);
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -308,9 +308,9 @@ public class ROIHandler {
                         roi.setName(shapeObject.getID());
 
                         if (Prefs.showAllSliceOnly){
-                            if(shapeObject.getTheC() != null)c = shapeObject.getTheC().getValue();
-                            if(shapeObject.getTheZ() != null)z = shapeObject.getTheZ().getValue();
-                            if(shapeObject.getTheT() != null)t = shapeObject.getTheT().getValue();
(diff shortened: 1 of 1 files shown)


jburel:
can ```roi.getCPosition()``` be null?
might be good to add a check in the unwrap method

pwalczysko:
Behaves as expected. Ready to merge after the question of @jburel is sorted.

bramalingam:
roi.getCPosition is never null. Its either 0 or some value. I tested this with a null value (without setting roi.setPosition) and figured out that the default was always 0 (which means that the ROI is propagated through the whole stack/hyperstack.

jburel:
```if(shapeObject.getTheC() != null)c = shapeObject.getTheC().getValue();```
formatting

File: components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java [hunk 1]
```diff
@@ -301,17 +306,13 @@ else if (shapeObject instanceof ome.xml.model.Rectangle) {
...
-                          int c = shapeObject.getTheC().getValue();
-                          int z = shapeObject.getTheZ().getValue();
-                          int t = shapeObject.getTheT().getValue();
-                          roi.setPosition(c, z, t);
-                        }
 
+                        if (Prefs.showAllSliceOnly){
+                            if(shapeObject.getTheC() != null);c = shapeObject.getTheC().getValue();
+                            if(shapeObject.getTheZ() != null);z = shapeObject.getTheZ().getValue();
+                            if(shapeObject.getTheT() != null);t = shapeObject.getTheT().getValue();
```
melissalinkert: The semi-colons after the ```if(...)``` statements are probably not what you want here (see https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-build/782/findbugsResult/new/HIGH/).  Something like:

```
if (shapeObject.getTheC() != null) {
  c = shapeObject.getTheC().getValue();
}
```

addresses the formatting comment from @jburel and preserves the original logic.

bramalingam:
@melissalinkert , @jburel : I've updated the formatting on this. Does it look good?

melissalinkert:
Looks much better now, thanks.

PR #1712, opened 2015-04-03
FileStitcher: fix file indexing for RGB images

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/11095.

```showinf -stitch Series001_t0_z0_ch00.tif``` or opening in ImageJ with ```Group files with similar names``` checked using QA 7328 should now work correctly.  There are 9 files (3 Z sections x 3 channels), each of which contains a single RGB image, so with stitching/grouping enabled there should be 9 planes.  Without this change, all 9 planes would have been cyan (only the files corresponding to the first channel were used); with this change, there should be 3 distinct channels (cyan, red, and green).

### DIFF
--- a/components/formats-bsd/src/loci/formats/FileStitcher.java
+++ b/components/formats-bsd/src/loci/formats/FileStitcher.java
@@ -1093,7 +1093,7 @@ public class FileStitcher extends ReaderWrapper {
     lenC[sno] = new int[numC + 1];
     lenT[sno] = new int[numT + 1];
     lenZ[sno][0] = sizeZ[sno];
-    lenC[sno][0] = sizeC[sno];
+    lenC[sno][0] = sizeC[sno] / reader.getRGBChannelCount();
     lenT[sno][0] = sizeT[sno];
 
     for (int i=0, z=1, c=1, t=1; i<count.length; i++) {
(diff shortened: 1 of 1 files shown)


melissalinkert:
May need one more commit to fix issues noted by BIOFORMATS-5.1-merge-full-repository.

melissalinkert:
Turns out a commit was not necessary, but a few configuration updates were (in ```leica-lif/andrew/New XZY LIF Upload```, ```simplepci/martin```, ```zeiss-zvi/rolf```, and ```zeiss-zvi/cecilia```).  All should be sorted out now.

mtbc:
Needs relisting?

bramalingam:
"Warning: Data has too many channels for Default colour mode", from IJ.log.
And the number of channels seems to be 9, which again is not right.. Am I doing something wrong while importing?

screen shot 2015-04-13 at 14 45 31

screen shot 2015-04-13 at 14 45 40

screen shot 2015-04-13 at 14 45 07

bramalingam:
screen shot 2015-04-13 at 14 49 15

melissalinkert:
```showinf``` should have shown the correct number of channels, but it is worrying how this case is handled in the ImageJ plugin.  Happy to hold off on merging this PR if that's a blocker, otherwise I will try to fix the plugin behavior for 5.1.2.

melissalinkert:
Leaving this out of 5.1.1; I'll update this later in the week and we can re-evaluate for 5.1.2.

melissalinkert:
This does seem to be working in ImageJ - 9 channels is correct, since 3 channels are described in the file names and each of the 9 files (3 Z x 3 C) contains an RGB image.  Removing exclude so this is at least considered for 5.1.2.

bramalingam:

PR #1713, opened 2015-04-03
Imaris HDF: fix SizeC if channels are stored in reverse order

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12170.  Without this change, the file from QA 8019 had 3 Z sections and 1 channel.  With this change, the same file should have 3 Z sections and 3 channels when opened in ImageJ or imported into OMERO.  Note that all planes are fairly dark, so you may need to adjust rendering settings/brightness to see the artifact in the images.


pwalczysko:
Tested both ImageJ (with updated ``bioformats_package.jar`` and OMERO Insight. Behaves as expected. Ready to merge.

melissalinkert:
--no-rebase

PR #1714, opened 2015-04-06
Zeiss CZI: ignore null FilterSet references

melissalinkert:
This should prevent warnings similar to ```ome.xml.model.Channel@579e8e6e reference to null missing from object hierarchy.``` when working with some .czi files.  See https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=7669.

To test, use the file from QA 10974.  Verify that ```showinf -nopix -omexml``` shows the above warning without this change; with this change, the warning should not be present.


sbesson:
Tested using the Bio-FOrmats merge build with the command above and imported into an OMERO merge server. In both cases, the warnings disappear with this PR. Once the repository jobs have validated its impact on the CZI sample set, this looks good to merge.

melissalinkert:
--no-rebase

PR #1716, opened 2015-04-07
Use the unit from imagej comment

dominikl:
Fixes [Ticket 11399](https://trac.openmicroscopy.org/ome/ticket/11399)

Test: Import the images mentioned in the ticket in OMERO and make sure the unit 'nm' is used.

(Note: As I'm not yet familiar with the bioformats code, I have no idea if this is the way things like this should be done.)


melissalinkert:
This all seems fine, thanks.  Just waiting one more day to merge though, as this exposes a few issues in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/450/ for which I'll need to update a few of the configuration files (which I'll do first thing Thursday).

PR #1718, opened 2015-04-08
Leica lif reader

jburel:
Fix the sizeX and sizeY if not set.
Also fix some encoding issue noticed while testing the file
to test
 * run ```./showinf  PATH_TO/QA/10480/example.lif ```
 * Check that the image opened w/ error
 * Check that metadata display does not contain any undecoded character i.e. ```ATLConfocalSettingDefinition|ClimateControl|ClimateUnit|DimensionUnit: °C``` should not be displayed correctly.


dominikl:
Looks fine, image can be opened without problems, didn't notice any encoding issues. Good to merge.

PR #1720, opened 2015-04-09
Zeiss CZI: avoid reordering series

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12673.

To test, use the file from QA 10379 and verify that the images in series 1 and 2 (the middle two) are switched with each other when compared with the same series without this change.


mtbc:
:+1: Now the series number matches the timepoint number written on the image. Other CZIs still seem fine. Once BF 5.1 jobs turn back green, good to merge.

melissalinkert:
--no-rebase

PR #1721, opened 2015-04-10
Preserve the pixel type for floating-point thumbnails

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/6833.

To test, run:

```
showinf -normalize test_images_good/khoros/head.xv
showinf -normalize -thumbs test_images_good/khoros/head.xv
```

and verify that both produce an image with no exception, and that the latter is a reasonable-looking thumbnail of the former.


mtbc:
The head does indeed look fine:
heads

The kitten from the same directory looks less happy, but perhaps that's outwith this PR:
cat

mtbc:
The issue with the kitten is indeed preexisting, so no regression there, and for the head it's a definite improvement. :+1:

melissalinkert:
--no-rebase

PR #1722, opened 2015-04-10
Flex reader

jburel:
Remove trailing character.

To test: 
 * Import an image from QA : 9456

see https://trac.openmicroscopy.org/ome/ticket/12529#comment:3

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FlexReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FlexReader.java
@@ -813,9 +813,13 @@ public class FlexReader extends FormatReader {
       ifd = file.ifds.get(0);
     }
     else {
-      TiffParser parser = new TiffParser(file.file);
-      ifd = parser.getFirstIFD();
-      parser.getStream().close();
+      RandomAccessInputStream ras = new RandomAccessInputStream(file.file);
+      try {
+        TiffParser parser = new TiffParser(ras);
+        ifd = parser.getFirstIFD();
+      } finally {
+        ras.close();
+      }
     }
     String xml = XMLTools.sanitizeXML(ifd.getIFDStringValue(FLEX));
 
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/FlexReader.java
```diff
@@ -814,9 +813,13 @@ private void parseFlexFile(int currentWell, int wellRow, int wellCol,
...
     else {
-      TiffParser parser = new TiffParser(file.file);
-      ifd = parser.getFirstIFD();
-      parser.getStream().close();
+        RandomAccessInputStream ras = new RandomAccessInputStream(file.file);
+        try {
+            TiffParser parser = new TiffParser(ras);
+            ifd = parser.getFirstIFD();
+        } finally {
+            if (ras != null) ras.close();
```
joshmoore: Few things:
 * It's impossible for `ras` to be `null` here.
 * Whitespace is off (4 rather than 2 spaces)
 * Commit message may be misleading. Why was this change made?

jburel: * The null check is more a cut and
* The method  ```getFirstIFD``` can throw an exception => the stream might never be closed.
* intend: this is really a paint when switching between BF and omero. We probably need to sort that one out.

dominikl:
Looks fine to me.  Import of `imagebad.flex` still fails, but that's expected, right? Import of `imagebad2.flex` passes. 
@jburel I created two separate eclipse workspaces for handling the different settings for BF and omero, works quite well.

jburel:
@dominikl: let me check ```imagebad.flex```

jburel:
@dominikl: That's an option just wanted to avoid to do that constantly.

jburel:
@dominikl: no error reported when I do ```./showinf```, further check required. Does not seem to be a B-F issue

dominikl:
That's true, no problem with `imagebad.flex` when using `./showinf`, so it's an Insight problem, i. e. not a matter of this PR and can be merged I guess.

PR #1724, opened 2015-04-10
LIF reader empty sring

jburel:
Handle empty string before converting into number.

To test this PR
 * Try to import or use showinf to open image from QA 10983
 * Check that 
```
: empty String
at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:992)
at java.lang.Double.valueOf(Double.java:475)
at java.lang.Double.<init>(Double.java:567)
at loci.formats.in.LIFReader.translateFilterSettings(LIFReader.java:1495)
```
No longer appears

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -1256,9 +1256,13 @@ public class LIFReader extends FormatReader {
       roi.scaleY = parseDouble(roiNode.getAttribute("transScalingY"));
       roi.rotation = parseDouble(roiNode.getAttribute("transRotation"));
       String linewidth = roiNode.getAttribute("linewidth");
-      if (linewidth != null && !linewidth.trim().isEmpty()) {
-        roi.linewidth = Integer.parseInt(linewidth.trim());
+      try {
+        if (linewidth != null && !linewidth.trim().isEmpty()) {
+          roi.linewidth = Integer.parseInt(linewidth.trim());
+        }
       }
+      catch (NumberFormatException e) { }
+
       roi.text = roiNode.getAttribute("text");
 
       NodeList vertices = getNodes(roiNode, "Vertex");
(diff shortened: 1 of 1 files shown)


mtbc:
Imports fine, and I don't see any obvious problems among the metadata.

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -1255,11 +1256,8 @@ private void translateROIs(Element imageNode, int image)
       roi.scaleY = parseDouble(roiNode.getAttribute("transScalingY"));
       roi.rotation = parseDouble(roiNode.getAttribute("transRotation"));
       String linewidth = roiNode.getAttribute("linewidth");
-      if (linewidth != null) {
-        try {
-          roi.linewidth = Integer.parseInt(linewidth);
-        }
-        catch (NumberFormatException e) { }
+      if (linewidth != null && !linewidth.trim().isEmpty()) {
+        roi.linewidth = Integer.parseInt(linewidth.trim());
```
melissalinkert: Removing the NFE catch here worries me a little.

jburel: could we have something other than int?

melissalinkert: I think it's possible to have a double - maybe makes sense to use (int) Double.parseDouble and catch NFE just to be safe (since we were doing that anyway).

melissalinkert:
Nitpicky comment aside, I think this is fine.  At some later date, it might make sense to have convenience methods to do the ```foo == null || foo.isEmpty() ? 0 : Integer.parseInt(foo.trim())``` checks, but that's very low priority and certainly fine for post-5.1.1.

jburel:
@melissalinkert: I was thinking the same thing, since it might be used in other readers. Will revisit in post 5.1.1

PR #1725, opened 2015-04-10
Zeiss CZI: fix series count for fused images

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12789.

To test, use the two files referenced in the ticket (both have been uploaded to ```zeiss-czi/sebastian/fused```).  The ```*_fused.czi``` file would have shown 6 series of strange images without this change; with this change, there should be a single large image that more or less matches what is shown in Zen.  The behavior when opening the ```*_stitched.czi``` file should be unchanged.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -645,9 +645,26 @@ public class ZeissCZIReader extends FormatReader {
       seriesCount == (planes.size() / getImageCount()) &&
       prestitched != null && prestitched)
     {
-      prestitched = false;
-      ms0.sizeX = planes.get(planes.size() - 1).x;
-      ms0.sizeY = planes.get(planes.size() - 1).y;
+      boolean equalTiles = true;
+      for (SubBlock plane : planes) {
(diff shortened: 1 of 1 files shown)


sbesson:
Import of the `_fused.czi` file into OMERO with this PR went fine resulting in one image as opposed to 6 without this PR. So the PR at least fixes the series mismatch issue. On the flip side, pyramid generation for the CZI file tested in the context of #1733 hangs and blocks pyramid generation for this file as well. This may be either an issue at the reader level or caused by the upstream https://github.com/openmicroscopy/openmicroscopy/pull/3713 PR touching pyramid logic. Still under investigation.

sbesson:
After completion of the pyramid for  file mentioned in #1733, pyramid generation for the fused CZI, the pyramid for the fused CZI file was generated in 10 min

```
2015-04-16 18:37:04,597 INFO  [ ome.services.pixeldata.PixelDataHandler] (2-thread-2) HANDLED EventLog:3907902(entityId=28526) [680826 ms.]
```

and the data visualization looks good:

screen shot 2015-04-16 at 19 01 32
screen shot 2015-04-16 at 19 01 51

Barring concerns about #1733, this PR looks ready to merge.

melissalinkert:
--no-rebase

PR #1727, opened 2015-04-13
Bf importer performance

bramalingam:
Testing instructions : 

1) Download and update the Bioformats_Package.jar in your ImageJ plugin's folder.
2) Import without ROI's : Import the following image using the Bioformats_Importer: data_repo/from_skyking/cellh5/samples
with the following checkboxes selected in the Importer Options,
             - Display OME-XML metadata.
             - Use virtual stack.
3) Import with ROI's : Import the following image using the Bioformats_Importer: data_repo/from_skyking/cellh5/samples
with the following checkboxes selected in the Importer Options,
           - Display OME-XML metadata.
           - Display ROIs.
           - Use virtual stack.

4) Compare performance between steps 2 and 3. The performance of "Import without ROIs" should be better than the latter. And the OME-XML metadata should not have the ROI information in step 2.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
@@ -493,7 +493,17 @@ public class ImportProcess implements StatusReporter {
     BF.status(options.isQuiet(), "Analyzing " + getIdName());
     baseReader.setMetadataFiltered(true);
     baseReader.setGroupFiles(!options.isUngroupFiles());
+    if(options != null && !options.showROIs()){
+        MetadataOptions mo = baseReader.getMetadataOptions();
+        if(mo == null){
+            mo = new DefaultMetadataOptions();
+        }else{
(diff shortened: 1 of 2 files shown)


melissalinkert:
This is definitely progress, but with "Display OME-XML metadata" checked and "Display ROIs" unchecked, the file from ```cellh5/samples/reduced``` still logs that the ROIs are being parsed.

I think ImportProcess needs one more change, around line 495.  If MetadataOptions is set on baseReader similar to what's being done in lines 536-544, then I think that will do the trick.

bramalingam:
Definitely makes a huge difference in performance. @melissalinkert : Please check this, when you get some time.

melissalinkert:
Performance is much better now, and leaving "Display ROIs" unchecked no longer shows ROI logging from the CellH5 reader.  This is fine to merge as soon as the CellH5 change is reverted, as discussed earlier this evening (that shouldn't affect the behavior of this PR, but it does affect what happens when the MetadataLevel is set to MINIMUM).

melissalinkert:
:+1:

PR #1729, opened 2015-04-13
Documentation for manual upgrade of Bio-Formats in FIJI

bramalingam:
Testing Instructions : 
Replicate steps given in the documentation to check if the upgrade works.

### DIFF
--- a/docs/sphinx/users/fiji/index.txt
+++ b/docs/sphinx/users/fiji/index.txt
@@ -23,3 +23,43 @@ checks for updates every time it is launched, so you will always be
 notified when new versions of Bio-Formats (or any other bundled plugin)
 are available.
 
+
+Manual Upgrade
+--------------
+
+1) Fiji must first be fully updated
+2) Close Fiji
+3) Open the Fiji installation folder (typically named 'Fiji.app')
+4) Remove bio-formats_plugins-5.0.7.jar from the 'plugins' sub-folder
+5) Remove all of the following from the 'jars/bio-formats' sub-folder:
+jai_imageio-5.0.7.jar
+formats-gpl-5.0.7.jar
+formats-common-5.0.7.jar
+turbojpeg-5.0.7.jar
+ome-xml-5.0.7.jar
+formats-bsd-5.0.7.jar
+ome-poi-5.0.7.j.ar
+specification-5.0.7.jar
+mdbtools-java-5.0.7.jar
+metakit-5.0.7.jar
+formats-api-5.0.7.jar
+6) Download bio-formats_plugins.jar (from our latest stable release http://downloads.openmicroscopy.org/bio-formats/) and place it in the 'plugins' sub-folder
+7) Download each of the following (from our latest stable release http://downloads.openmicroscopy.org/bio-formats/) and place them in the 'jars/bio-formats' sub-folder:
+​jai_imageio.jar
+formats-gpl.jar
+formats-common.jar
+turbojpeg.jar
+ome-xml.jar
+formats-bsd.jar
+ome-poi.jar
+specification.jar
+mdbtools-java.jar
+metakit.jar
+formats-api.jar
+8) To Check Version of Bio-Formats : Select Help > About Plugins > Bio-Formats Plugins...
+Check that the version of Bio-Formats matches the freshly downloaded version.
+9) Start Fiji and open any Image file using "Plugins > Bio-Formats > Bio-Formats Importer".
+
+Note : It is vital to perform all of those steps in order; omitting even one will cause a problem. In particular, make sure that the old files are fully removed; it is not sufficient to add the new files to any sub-directory without removing the old files first.
+
+


hadim:
Just in case it can helps someone : https://github.com/hadim/fiji_tools/blob/master/update_bfplugin.sh

You may want to add this script into the doc.

hflynn:
Sphinx build failed, I suspect it's your formatting

docs/sphinx/users/fiji/index.txt
```diff
@@ -23,3 +23,43 @@ checks for updates every time it is launched, so you will always be
...
+3) Open the Fiji installation folder (typically named 'Fiji.app')
+4) Remove bio-formats_plugins-5.0.7.jar from the 'plugins' sub-folder
+5) Remove all of the following from the 'jars/bio-formats' sub-folder:
+jai_imageio-5.0.7.jar
+formats-gpl-5.0.7.jar
+formats-common-5.0.7.jar
+turbojpeg-5.0.7.jar
+ome-xml-5.0.7.jar
+formats-bsd-5.0.7.jar
+ome-poi-5.0.7.j.ar
```
melissalinkert: ```.jar``` instead of ```.j.ar``` (given the line number, this might be what was confusing the build)

melissalinkert:
@hadim: thanks for the suggestion.  Note though that line 22 of that script downloads ```bioformats_package.jar```, which we definitely don't recommend using with Fiji - if that line were changed to use ```bio-formats_plugins.jar``` (as in the instructions here), then we'd be happy to add a link to the script here or in a follow-up pull request.

hadim:
Thank you @melissalinkert . It's now fixed !

hadim:
I suggest you to add directly the script itself (instead of a link to the script) because the persistance of such script in my repo is far from being stable :-)

hflynn:
Travis is failing this on the Maven & cpp builds, can someone kick it please?

joshmoore:
Kicked, but some of these may persist if there are travis upgrade issues:
```
Exception in thread "main" java.lang.UnsupportedClassVersionError: org/apache/maven/cli/MavenCli : Unsupported major.minor version 51.0
```

bramalingam:
@melissalinkert : Any more changes in here?

hinerm:
@bramalingam From a Fiji perspective, Fiji always wants to include the latest Bio-Formats stable release. It [now ships Bio-Formats 5.1](http://imagej.net/2015-04-15_-_ImageJ_2.0.0-rc-29,_plus_Bio-Formats_5.1.0_in_Fiji). @ctrueden typically updates within a few days (at worst) of each new release - so instructions for upgrading to a stable release should be unnecessary.

If the goal is to document how to use Bio-Formats development builds in Fiji, why not link to [instructions for the Bio-Formats update site](http://imagej.net/Bio-Formats#Daily_builds)?

If the goal is to document how to replace Bio-Formats with an arbitrary build - as @ctrueden mentioned Tuesday, this is a perfect opportunity for automation. Let's generalize the [upload-bf.sh](https://github.com/fiji/fiji/blob/master/bin/upload-bf.sh) script to fulfill this use case, instead of manual instructions.

@ctrueden and I will follow up soon with succinct instructions on how to use such a script.

melissalinkert:
Thanks, @hinerm.  We definitely want to recommend a script as the preferred way of upgrading to a non-stable build outside of the Bio-Formats update site, and can update this page once you and @ctrueden have instructions for using said script (@hadim: this is likely to be a more general version of what you have linked).

However, I think it makes sense to keep the manual instructions for doing this as a fall-back.  Some users may be uncomfortable with the idea of running a script, or may have wound up with an installation that the script cannot fully handle (e.g. partial upgrade between stable versions, incorrectly adding ```bioformats_package.jar```, etc.).  We already have the text written up, it's known to work (per ImageJ mailing list threads), and so it would be unfortunate to lose it entirely.

I would propose that we merge this, and open a separate PR to place the script instructions above the manual instructions (as soon as they are available).

hinerm:
@melissalinkert Those reasons for keeping manual instructions sound good to me.

What originally caught my attention was that the instructions, as-is, say: update Fiji, then delete versions `5.0.7`, then download `"our latest stable release"` - but the act of udpating Fiji will now give you `5.1.0` jars which invalidate the subsequent steps.

Perhaps steps `4` and `5` should be generalized to instructions on deleting any bio-formats jars from `/jars`, `/jars/bio-formats` and `/plugins`? This might be more future proof, and would then cover some of the scenarios you mentioned.

docs/sphinx/users/fiji/index.txt
```diff
@@ -23,3 +23,81 @@ checks for updates every time it is launched, so you will always be
...
+
+Fiji currently shipping with the 5.1.x release versions of Bio-Formats.
+However, if you have encountered a bug which has been fixed by the Bio-Formats
+team but not yet released, you can use the Bio-Formats update site to
+access the daily build as described in the
+`Fiji documentation <http://fiji.sc/Bio-Formats#Daily_builds>`_.
+
+.. warning:: These builds are **not yet released** and should be considered
+    **beta** in quality. In particular, you should **avoid exporting data
+    using the Bio-Formats Exporter** in case you write incompatible files
```
joshmoore: I'm not sure this holds. This was for the break between 5.0 and 5.1. @melissalinkert?

melissalinkert: The recommendation to avoid exporting is there because daily builds may not be using a released schema version (see https://trac.openmicroscopy.org/ome/ticket/10168).

joshmoore: But never for a stable version, eh? Couldn't this only happen during the period of a DEV schema?

melissalinkert: Correct, that only applies to daily builds and not stable releases.

hflynn: Is it not clear enough that this warning only applies to the daily builds Josh? I thought we wanted to make clear to users that daily builds are risky not start detailing the caveats on those warnings - whether or not the schema version is in flux isn't ever going to be obvious to the average Fiji user.

joshmoore: The only bit here that I'm particularly worried about is the bit about incompatible files:
> In particular, you should **avoid exporting data
> using the Bio-Formats Exporter** in case you write incompatible files

 since that only holds for DEV releases:

hflynn: Ok, so from chatting to Josh - this export issue doesn't hold currently and we don't know when it will be true again so warning about it seems unnecessary. I can remove it here but in that case we should also ask that the warning on the Fiji docs is updated too so we are all giving users to the same advice.

ctrueden: > we should also ask that the warning on the Fiji docs is updated too so we are all giving users the same advice.

Fortunately, the ImageJ web site is a wiki, so you can edit it as you like. No asking needed—[be bold](https://en.wikipedia.org/wiki/Wikipedia:Be_bold)!

hflynn: Ok, final decision then @joshmoore ? Shall I remove that part of the warning here and on the Fiji wiki before this is merged?

joshmoore: @hflynn : @melissalinkert has convinced me (...to be quiet). :smile: Roughly, this needs some more planning, since though what I'm saying is true *now*, what's she saying will be true *eventually*. This will take some planning, so let's go with the status quo for 5.1.1.

hflynn: :+1:

hflynn: Good to merge from my POV then.

hflynn:
--no-rebase

PR #1730, opened 2015-04-13
clean up older Java code

mtbc:
This PR is another batch of making the code more idiomatically Java 6, with a focus on switching away from synchronized collections where access is single-threaded, also some attempts to avoid unnecessary heap allocations and other small efficiency tweaks. It should *not* change any reader behaviors.

--no-rebase

### DIFF
--- a/components/formats-bsd/src/loci/formats/tools/PrintDomains.java
+++ b/components/formats-bsd/src/loci/formats/tools/PrintDomains.java
@@ -33,12 +33,12 @@
 package loci.formats.tools;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Hashtable;
-import java.util.Vector;
+import java.util.Collection;
+import java.util.Map;
+
+import com.google.common.collect.TreeMultimap;
 
 import loci.formats.FormatException;
-import loci.formats.FormatTools;
 import loci.formats.IFormatReader;
 import loci.formats.ImageReader;
 
@@ -52,18 +52,13 @@ public class PrintDomains {
     // get a list of all available readers
     IFormatReader[] readers = new ImageReader().getReaders();
 
-    Hashtable<String, Vector<IFormatReader>> domains =
-      new Hashtable<String, Vector<IFormatReader>>();
-
-    for (String domain : FormatTools.ALL_DOMAINS) {
-      domains.put(domain, new Vector<IFormatReader>());
-    }
+    final TreeMultimap<String, String> domains = TreeMultimap.create();
 
     for (IFormatReader reader : readers) {
       try {
         String[] readerDomains = reader.getPossibleDomains("");
         for (String domain : readerDomains) {
-          domains.get(domain).add(reader);
+          domains.put(domain, reader.getFormat());
         }
       }
       catch (FormatException e) {
@@ -74,14 +69,11 @@ public class PrintDomains {
       }
     }
 
-    String[] domainKeys = domains.keySet().toArray(new String[domains.size()]);
-    Arrays.sort(domainKeys);
-
-    for (String domain : domainKeys) {
-      System.out.println(domain + ":");
-      Vector<IFormatReader> r = domains.get(domain);
-      for (IFormatReader reader : r) {
-        System.out.println("  " + reader.getFormat());
+    for (final Map.Entry<String, Collection<String>> domain :
+             domains.asMap().entrySet()) {
+      System.out.println(domain.getKey() + ":");
+      for (final String readerFormat : domain.getValue()) {
+        System.out.println("  " + readerFormat);
       }
     }
   }
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,3 +10,4 @@
 *.so binary
 *.dylib binary
 *.h5 binary
+*.mdb binary
--- a/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
@@ -43,6 +43,7 @@ import java.awt.Panel;
 import java.awt.event.ItemEvent;
 import java.io.File;
 import java.math.BigInteger;
+import java.util.ArrayList;
 import java.util.Vector;
 
 import javax.swing.Box;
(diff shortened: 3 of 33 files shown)


mtbc:
A breaking label, please: this touches a fair few readers and it'd be nice to get wrinkles out without affecting others' work.

melissalinkert:
Could you please revert the change to ```test.mdb```?  That seems to be failing the build, and I don't see a reason for it being modified.

mtbc:
Gah, yes, sorry. That file gives me no end of trouble, git keeps thinking it was modified and no amount of reset or checkout or whatever makes it think differently. Messing with `.gitattributes` can help but then *that* file looks modified instead.

melissalinkert:
No worries - feel free to add ```*.mdb binary``` to the repository ```.gitattributes``` if that helps (for some reason I thought that had been done already, but now I don't see it).

mtbc:
https://ci.openmicroscopy.org/view/Breaking/job/BIOFORMATS-5.1-breaking-test_images_good/ looks good so remove breaking label?

components/formats-bsd/src/loci/formats/tools/PrintDomains.java
```diff
@@ -52,18 +52,13 @@ public static void main(String[] args) {
     // get a list of all available readers
     IFormatReader[] readers = new ImageReader().getReaders();
 
-    Hashtable<String, Vector<IFormatReader>> domains =
-      new Hashtable<String, Vector<IFormatReader>>();
-
-    for (String domain : FormatTools.ALL_DOMAINS) {
-      domains.put(domain, new Vector<IFormatReader>());
-    }
+    final TreeMultimap<String, String> domains = TreeMultimap.create();
```
rleigh-dundee: Why are we not filling the `domains` map with the values in `ALL_DOMAINS` as before?

rleigh-dundee:
Other than the question about domains, all the changes look very sensible.  All the builds are green except for full-repo-test which is still running; please merge if green and the domains change is correct.

components/formats-bsd/src/loci/formats/tools/PrintDomains.java
mtbc: I figured we were just putting a bunch of empty lists in so that we can then add a format to a list without checking if a list is waiting to receive it. (Multimaps handle this chore for us.) I suppose it means that for a domain that had no readers we wouldn't now bother printing the heading with nothing below it before the next heading, but in practice all the domains have a good few readers associated, at least these days. (This can be tried out with `tools/domainlist`.)

rleigh-dundee: Ah, OK.  So long as that's not breaking anything (if anything assumes they are present by default), then it's fine.

mtbc:
Thanks! (Ha, it's still running, I'll look again tomorrow...)

melissalinkert:
The changes to PCIReader seem to be causing:

```
[testng] [2015-04-16 14:48:20,620] [pool-1-thread-9] Initializing /ome/data_repo/from_skyking/simplepci/jcb/Test Run, 60 Sec PassesNEW.cxd: 
   [testng] [2015-04-16 14:48:26,404] [pool-1-thread-9] memo file size for /ome/data_repo/from_skyking/simplepci/jcb/Test Run, 60 Sec PassesNEW.cxd = 342339 bytes
   [testng] [2015-04-16 14:48:26,887] [pool-1-thread-9] 	SizeT: FAILED (Series 0 (expected 35, actual 30))
   [testng] [2015-04-16 14:48:26,922] [pool-1-thread-9] 	SizeZ: FAILED (Series 0 (expected 24, actual 28))
```

though it's not immediately obvious to me why that is.  Excluding temporarily.

mtbc:
Thank you for pinning down the problem images, I'll give them a try locally.

mtbc:
To my surprise, the difference between passing and failing is:
```diff
diff --git a/components/formats-gpl/src/loci/formats/services/POIServiceImpl.java b/components/formats-gpl/src/loci/formats/services/POIServiceImpl.java
index 4a86548..7d0569f 100644
--- a/components/formats-gpl/src/loci/formats/services/POIServiceImpl.java
+++ b/components/formats-gpl/src/loci/formats/services/POIServiceImpl.java
@@ -29,9 +29,11 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Vector;
 
 import loci.common.RandomAccessInputStream;
@@ -55,7 +57,7 @@ public class POIServiceImpl extends AbstractService implements POIService {
   private RandomAccessInputStream stream;
 
   private List<String> filePath;
-  private Hashtable<String, Integer> fileSizes;
+  private Map<String, Integer> fileSizes;
   private Hashtable<String, DocumentEntry> files;
 
   // -- POIService API methods --
@@ -90,7 +92,7 @@ public class POIServiceImpl extends AbstractService implements POIService {
 
     // build the list of files in the file system
     filePath = new ArrayList<String>();
-    fileSizes = new Hashtable<String, Integer>();
+    fileSizes = new HashMap<String, Integer>();
     files = new Hashtable<String, DocumentEntry>();
 
     parseFile(root);
```

mtbc:
While a more minimal fix satisfies the tests, I restored all the synchronized collections to `POIServiceImpl` for safety. I'd love to know why https://github.com/mtbc/bioformats/commit/cf35aff361f6c6afc1400f81564589a1da598550 was required. Remove `exclude` label?

melissalinkert:
I kind of wonder if the ```fileSizes.keySet()``` call on line143 of POIServiceImpl is the issue, but I'm not completely sure.  Removing the exclude label.

mtbc:
I think I broke gatan too. Feel free to add exclude label again while I investigate.

sbesson:
Given the nature of the error (incorrect `PixelType`), my suspicion is that https://github.com/openmicroscopy/bioformats/pull/1745 is the PR responsible for the Gatan failures.

mtbc:
I think you're right it's not mine, actually: thank you. I tried running the failing Gatan tests with and without my extra commits and it made no difference.

mtbc:
(Just for interest regarding the reverted change: `LinkedHashMap` doesn't help either.)

melissalinkert:
Builds are all green now - merging for 5.1.1.

PR #1731, opened 2015-04-13
Prevent test failures when PhysicalSize* units are in pixels

melissalinkert:
This should fix the 8 test failures that are not logged in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository-win/3/ and https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/454/.

This issue was exposed by gh-1716; if the physical unit is set to 'pixel', then it won't be convertible to ```Units.MICROM```, which results in hidden NPEs when the tests are run.  No testing needed other than to verify that the failure count is reduce by 8 with this PR merged.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -468,19 +468,19 @@ public class Configuration {
 
       Length physicalX = retrieve.getPixelsPhysicalSizeX(series);
       if (physicalX != null) {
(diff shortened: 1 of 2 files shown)


dominikl:
The build from Apr 12 has 30 failed tests, yesterday's just 4. Unfortunately I'm struggling to get any details which tests failed, why and if they didn't fail again because of the PR. But the code changes look fine, so I think good to merge.

melissalinkert:
--no-rebase

Issue #1719, opened 2015-04-08
Error on opening CZI file -- bfGetPlane 


gvoysey:
I'm attempting to open a 760 MB CZI file acquired by another research group with, i believe, Zen Lite.   I have installed the matlab toolbox, added it to my path, and set my java maximum heap memory to 1024 MB.  I'm running r2012a (64 bit) on Windows 7. 

Executing 
``` matlab
 data = bfopen('IP-Stitching-01-Change Scaling-01.czi');
```

results in 
``` java
ZeissCZIReader initializing C:\Projects\LMR-Acquisition\IP-Stitching-01-Change Scaling-01.czi

Unknown IlluminationType value 'Fluorescence' will be stored as "Other"

ome.xml.model.Channel@66156d8 reference to null missing from object hierarchy.

ome.xml.model.Channel@28b689e0 reference to null missing from object hierarchy.

ome.xml.model.Channel@7b0064e3 reference to null missing from object hierarchy.

ome.xml.model.Channel@26b72884 reference to null missing from object hierarchy.

ome.xml.model.Channel@232a32bf reference to null missing from object hierarchy.

ome.xml.model.Channel@150abd60 reference to null missing from object hierarchy.

ome.xml.model.Channel@1a15deb6 reference to null missing from object hierarchy.

ome.xml.model.Channel@a756b37 reference to null missing from object hierarchy.

ome.xml.model.Channel@45b6867 reference to null missing from object hierarchy.

ome.xml.model.Channel@75af8109 reference to null missing from object hierarchy.

ome.xml.model.Channel@782a519b reference to null missing from object hierarchy.

ome.xml.model.Channel@76792357 reference to null missing from object hierarchy.

ome.xml.model.Channel@129e49c0 reference to null missing from object hierarchy.

ome.xml.model.Channel@2db45934 reference to null missing from object hierarchy.

ome.xml.model.Channel@3a78cbab reference to null missing from object hierarchy.

ome.xml.model.Channel@f0896b1 reference to null missing from object hierarchy.

ome.xml.model.Channel@1a0d8377 reference to null missing from object hierarchy.

ome.xml.model.Channel@cabe02e reference to null missing from object hierarchy.

ome.xml.model.Channel@a832ce5 reference to null missing from object hierarchy.

ome.xml.model.Channel@627f7051 reference to null missing from object hierarchy.

Reading series #1
    .Error using loci.formats.ChannelSeparator/openBytes
Java exception occurred:
[...]

PR #1733, opened 2015-04-13
Zeiss CZI: do not set indexed for RGB images

melissalinkert:
See QA 10987, fixes gh-1719.

To test, verify that opening the file from QA 10987 in Matlab (as described in the issue) or via ```showinf -separate``` no longer throws an exception.  The image is quite large, so you may need to increase the heap size or crop the image to prevent an OOM.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -754,7 +754,7 @@ public class ZeissCZIReader extends FormatReader {
       indexIntoPlanes.put(c, indices);
     }
 
-    if (channels.size() > 0 && channels.get(0).color != null) {
+    if (channels.size() > 0 && channels.get(0).color != null && !isRGB()) {
       for (int i=0; i<seriesCount; i++) {
         core.get(i).indexed = true;
       }


sbesson:
Tested directly from the branch (note this PR will need to go through the merge builds tonight). The `ChannelSeparator` reader is correctly initialized on the file from QA 10987 and all planes are read without any exception thrown for all series (tested using `bfopen`).
Data quality is inconsistent between series e.g. series 9 renders fine but series 0 looks buggy. Not sure whether this arises from the original file and/or is covered by other CZI bug fixes.

melissalinkert:
This should probably be retested in conjunction with gh-1725 or as part of the merge build - with gh-1725 included, there should only be 1 series that represents the full image, and I'd expect it to more or less look like it does in Zen.

mtbc:
It's much faster to `showinf` this image from a local filesystem than over CIFS! Things seem good with the merge build. Opening had,
```
Initializing reader
ZeissCZIReader initializing /tmp/IP-Stitching-01-Change Scaling-01.czi
Unknown IlluminationType value 'Fluorescence' will be stored as "Other"
Initialization took 0.335s

Reading core metadata
Filename = /tmp/IP-Stitching-01-Change Scaling-01.czi
Series count = 1
Series #0 :
	Image count = 3
	RGB = false (1) (separated)
	Interleaved = true
	Indexed = false (false color)
	Width = 12393
	Height = 9369
	SizeZ = 1
	SizeT = 1
	SizeC = 3
	Thumbnail size = 128 x 96
	Endianness = intel (little)
	Pixel type = uint16
	Valid bits per pixel = 14
	Metadata complete = false
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0
	Plane #1 <=> Z 0, C 1, T 0
	Plane #2 <=> Z 0, C 2, T 0


Reading pixel data (0-2)
	Read 1/3 planes (33%)
	Read 2/3 planes (66%)
	Read 3/3 planes (100%)
[done]
217.266s elapsed (72422.0ms per plane)

Launching image viewer
```
only one series and no exceptions. Other CZI files still seem fine.

Trying it with `-thumbs` doesn't look too good, perhaps that's expected?
With `-separate`:
with-separate
Without `-separate`:
without-separate

sbesson:
Note pyramid completion time for the file above was:

```
2015-04-16 18:25:43,555 INFO  [ ome.services.pixeldata.PixelDataHandler] (2-thread-1) HANDLED EventLog:3907867(entityId=28314) [24851493 ms.]
```

mtbc:
The resulting image looks good, though. The thumbnail from Insight is 
thumbnail

melissalinkert:
Any objections to merging this?  The ```-thumbs``` behavior is outside the scope of this PR.

mtbc:
No objections, go for it! :+1:

bramalingam:
Tested with Matlab and bftools (showinf -nopix) and in both cases there were no exceptions. Good to merge..

melissalinkert:
--no-rebase

PR #1734, opened 2015-04-14
Fix SSS timestamp parsing

sbesson:
Fixes https://trac.openmicroscopy.org/ome/ticket/12810

This commit is a follow-up of https://github.com/openmicroscopy/bioformats/pull/1508 and
- refactors the special milliseconds handling logic implemented in `LeicaReader` as another `DateTools.getDate()` helper method
- adds some basic unit tests to cover both implementations of `DateTools.getDate()` with/without millliseconds handling
- goes through all the readers parsing date using `SSS` format and uses the new `DateTools.getDate()` helper

To test this PR, check timestamps/dates from the various readers modified by this PR are correctly read. All other readers mentioned in the ticket above should also be checked. As far as I could tell `SSS` is used for formatting not parsing purposes so they should not need any special handling.


melissalinkert:
This all seems fine - I don't see any problem with the new timestamp values.  Only slight issue is that IPWReader, QuesantReader, and TillVisionReader will still need to be updated (due to use of DateTools.formatDate), but that's fine to happen post-5.1.1.

PR #1736, opened 2015-04-14
ImageJ: use series state in open*ImagePlus(String) methods

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/11145.

To test, use a macro similar to the following in ImageJ:

```
run("Bio-Formats Macro Extensions");
path = "/path/to/multi-series-file";
Ext.setSeries(1);
Ext.openImagePlus(path);
Ext.setSeries(0);
Ext.openImagePlus(path);
```

Verify that without this change, the images for series 0 are opened twice.  With this change, the images for series 1 should be opened first, followed by the images for series 0.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/macro/LociFunctions.java
+++ b/components/bio-formats-plugins/src/loci/plugins/macro/LociFunctions.java
@@ -80,6 +80,7 @@ public class LociFunctions extends MacroFunctions {
   // -- Fields --
 
   private ImageProcessorReader r;
+  private int series = 0;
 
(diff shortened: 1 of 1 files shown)


jburel:
it works as expected. It might be worth adding check for null or negative value when setting the series.

melissalinkert:
Pushed commit with null/negative series check.

jburel:
restarted travis

jburel:
Thanks for adding check. Ready to be merged

melissalinkert:
--no-rebase

PR #1737, opened 2015-04-15
VSI: prevent infinite loop when parsing extended tags

melissalinkert:
The files that demonstrate the problem are unfortunately private, so the only testing necessary is to verify that the builds are/remain green.

/cc @emilroz


mtbc:
The code change looks plausible, at least.

emilroz:
Check the latest build:
```
EMil:showinf emilrozbicki$ ./showinf -version
Version: 5.1.0-DEV
VCS revision: e2a3ec2754110bad44eeccd1ebe30fef65a84cf8
Build date: 16 April 2015
```
on the private files -  infinite loops is gone.
Looks good.

melissalinkert:
--no-rebase

PR #1738, opened 2015-04-15
ICS writer: fix saving of > 2GB of pixel data

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12602.

To test, use an input file with more than 2GB of pixel data - I used ```large-file&sizeX=2048&sizeY=2048&sizeZ=1000.fake```.  Use bfconvert to convert the input file to an .ics file, and verify that the conversion succeeds without an exception, the output file is larger than 2GB, and the last plane from the output file can be read.  Without this change, the same test would have resulted in an exception during bfconvert.


rleigh-dundee:
The change looks fine and I've tested writing a large 5GB image and reading it back in and it's all working correctly.  Please merge.

melissalinkert:
--no-rebase

PR #1739, opened 2015-04-15
Improvision TIFF: reset image count if ZCT dimensions are incorrect

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12184.

To test, place the TIFF files from QA 8025 and 9163 in the same directory (unzip as necessary and remove extra zip/log files).  Run ```showinf``` on both files, and import both into OMERO - with this change, each of the files should import as a separate fileset, with one plane and no exception.  Without this change, both would have thrown an exception during import as noted in the ticket.


mtbc:
Everything exactly as promised in the description. :+1: CI willing, good to merge.

melissalinkert:
--no-rebase

PR #1740, opened 2015-04-15
ND2: make series names the same length

melissalinkert:
This adds 0-padding to make the index lengths equal, which should make
the series list easier to read when sorted alphanumerically.

Fixes https://trac.openmicroscopy.org/ome/ticket/12809.  To test, check the output of ```showinf -nopix -omexml``` with the .nd2 file from QA 10945.  The Image names should all be the same length with this change.


dominikl:
Works like expected, checked with bftools and Insight (images are now nicely sorted). Interesting usage of the `String.format()` method, I should use this method more often, too.
Good to merge.

melissalinkert:
--no-rebase

PR #1741, opened 2015-04-16
Fix deprecated UpgradeCheck parsing logic

sbesson:
The UpgradeCheck logic assumed the upgrade check query returns a string
containing the latest version. This commit reproduces the UpgradeChecker logic
implemented in OMERO which only tests for the length of the returned string.

PR #1742, opened 2015-04-16
Fix deprecated UpgradeCheck parsing logic (rebased onto develop)

sbesson:
This is the same as gh-1741 but rebased onto develop.

----

The UpgradeCheck logic assumed the upgrade check query returns a string
containing the latest version. This commit reproduces the UpgradeChecker logic
implemented in OMERO which only tests for the length of the returned string.


sbesson:
--rebased-from #1741

sbesson:
To test this PR, run any of the command-line tools (`showinf`) with this PR included. Contrary to #1741, there should be no warning message  indicating a new stable version is available.

melissalinkert:
Seems fine to me, but probably good to have one more pair of eyes on this given how important it is.

PR #1743, opened 2015-04-16
C++: Add TIFF writer

rleigh-dundee:
This PR adds a TIFF writer.  It also adds infrastructure for future work for TIFF and OME-TIFF, including:

- TIFF codec querying
- Reading a single subchannel from a TIFF
- Writing a single subchannel to a TIFF (interface only)
- Minor refactoring/cleanup of FormatWriter implementation, plus addition of helper methods
- Addition of significant bit sizes for PixelProperties
- Adjustment of CopySubchannelVisitor to use planar layout by default
- Addition of MergeSubchannelVisitor
- Helper functions to set up an OMEXMLMetadataStore from CoreMetadata
- Helper functions to compute total per-series and per-file pixel storage sizes
- Addition of a few TIFF/IFD helper methods
- Refactoring of some of the TIFF test logic for reuse by the writers

What it does *not* add:

- TIFF compression (codec stuff is present, but needs exposing and using in the writer)
- Extensive TIFFWriter testcases (currently 8-bit only)
- OME-TIFF writing

These will need adding in following PRs.

-no-rebase

--------

Testing: There is not as yet a `bfconvert` tool to write image, so testing is limited to the unit tests.  By default the test files are deleted, so you will need to:

- edit `cpp/test/ome-bioformats/minimaltiffwriter.cpp`; delete or comment the two lines in `TearDown` which delete the images.
- Make this change to test BigTIFF:
```
-  ASSERT_NO_THROW(tiffwriter.setId(testfile));
+  ASSERT_NO_THROW(tiffwriter.setId(testfile.replace_extension(".btf")));
```
  You can also edit `cpp/lib/ome/internal/config.h` and change the value of `TIFF_HAVE_BIGTIFF` from 1 to 0 to test the ifdefs and messages in combination with a bigTIFF extension.
- build normally, then run `make test` or `bf-test cpp/test/ome-bioformats/minimaltiffwriter` to run just the writing tests
- TIFF files written by the TIFF writer are in `cpp/test/ome-bioformats/data/minimaltiffwriter-*.tiff`; they should all be 64×64 8-bit greyscale with 1 row/strip and three IFDs (for RGB planes), i.e. similar to:

```
TIFF Directory at offset 0x1008 (4104)
  Image Width: 64 Image Length: 64
  Bits/Sample: 8
  Compression Scheme: None
  Photometric Interpretation: min-is-black
  Samples/Pixel: 1
  Rows/Strip: 1
  Planar Configuration: separate image planes
```

### DIFF
--- a/cpp/lib/ome/bioformats/detail/FormatWriter.h
+++ b/cpp/lib/ome/bioformats/detail/FormatWriter.h
@@ -222,6 +222,82 @@ namespace ome
         ome::compat::shared_ptr< ::ome::xml::meta::MetadataRetrieve>&
         getMetadataRetrieve();
 
+        /**
+         * Determine the number of image planes in the current series.
+         *
+         * @returns the number of image planes.
+         */
+        virtual
+        dimension_size_type
+        getImageCount() const;
+
+        /**
+         * Get the size of the X dimension.
+         *
+         * @returns the X dimension size.
+         */
+        virtual
+        dimension_size_type
+        getSizeX() const;
+
+        /**
+         * Get the size of the Y dimension.
+         *
+         * @returns the Y dimension size.
+         */
+        virtual
+        dimension_size_type
+        getSizeY() const;
+
+        /**
+         * Get the size of the Z dimension.
+         *
+         * @returns the Z dimension size.
+         */
+        virtual
+        dimension_size_type
+        getSizeZ() const;
+
+        /**
+         * Get the size of the T dimension.
+         *
+         * @returns the T dimension size.
+         */
+        virtual
+        dimension_size_type
+        getSizeT() const;
+
+        /**
+         * Get the size of the C dimension.
+         *
+         * @returns the C dimension size.
+         */
+        virtual
+        dimension_size_type
+        getSizeC() const;
+
+        /**
+         * Get the pixel type.
+         *
+         * @returns the pixel type.
+         */
+        virtual
+        ome::xml::model::enums::PixelType
+        getPixelType() const;
+
+        /**
+         * Get the number of valid bits per pixel.
+         *
+         * The number of valid bits per pixel is always less than or
+         * equal to the number of bits per pixel that correspond to
+         * getPixelType().
+         *
+         * @returns the number of valid bits per pixel.
+         */
+        virtual
+        pixel_size_type
+        getBitsPerPixel() const;
+
         // Documented in superclass.
         void
         setFramesPerSecond(frame_rate_type rate);
--- a/cpp/lib/ome/bioformats/detail/FormatWriter.h
+++ b/cpp/lib/ome/bioformats/detail/FormatWriter.h
@@ -121,6 +121,9 @@ namespace ome
         /// Current series.
         mutable dimension_size_type series;
 
+        /// Current plane.
+        mutable dimension_size_type plane;
+
         /// The compression type to use.
         boost::optional<std::string> compression;
 
@@ -183,6 +186,25 @@ namespace ome
         // Documented in superclass.
         dimension_size_type
         getSeries() const;
+
+        /**
+         * Set the active plane.
+         *
+         * @param no the plane to activate.
+         *
+         * @todo Remove use of stateful API which requires use of
+         * plane switching in const methods.
+         */
+        virtual void
+        setPlane(dimension_size_type no) const;
+
+        /**
+         * Get the active plane.
+         *
+         * @returns the active plane.
+         */
+        virtual dimension_size_type
+        getPlane() const;
 
         // Documented in superclass.
         bool
--- a/cpp/lib/ome/bioformats/out/MinimalTIFFWriter.cpp
+++ b/cpp/lib/ome/bioformats/out/MinimalTIFFWriter.cpp
@@ -47,6 +47,8 @@
 #include <ome/bioformats/tiff/IFD.h>
 #include <ome/bioformats/tiff/TIFF.h>
 
+#include <ome/internal/config.h>
+
 using ome::bioformats::detail::WriterProperties;
 using ome::bioformats::tiff::TIFF;
 using ome::bioformats::tiff::IFD;
@@ -102,7 +104,7 @@ namespace ome
         ifd(),
         ifdIndex(0),
         seriesIFDRange(),
-        bigTIFF(false)
+        bigTIFF(boost::none)
       {
       }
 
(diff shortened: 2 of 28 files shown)


rleigh-dundee:
The IFD readImage/writeImage methods added here were based upon a misunderstanding of the reader and writer interface plane mapping to subchannels.  Given that they can be implemented by the caller trivially, they can probably be dropped.

cpp/lib/ome/bioformats/out/MinimalTIFFWriter.cpp
```diff
@@ -0,0 +1,308 @@
...
+            if (!bigTIFF) // Not set manually
+              boost::format fmt
+                ("Pixel data size is %1%, but TIFF without BigTIFF "
+                 "support enabled has a maximum size of %2%; "
+                 "automatically enabling BigTIFF support to prevent potential failure");
+            fmt % pixelSize % std::numeric_limits<uint32_t>::max();
+
+            BOOST_LOG_SEV(logger, ome::logging::trivial::warning) << fmt.str();
+          }
+        else (!getBigTIFF && needBig)
```
melissalinkert: Will the contents of this ```else``` ever be executed?

cpp/lib/ome/bioformats/detail/FormatWriter.h
```diff
@@ -121,6 +121,9 @@ namespace ome
         /// Current series.
         mutable dimension_size_type series;
 
+        /// Current plane.
+        mutable dimension_size_type plane;
```
mtbc: `dimension_size_type` makes sense for plane number too? (I don't object, it's just not what I'd have guessed.)

cpp/lib/ome/bioformats/tiff/Util.cpp
```diff
@@ -0,0 +1,446 @@
...
+
+        /// @todo Image orientation (storage order and direction) from
+        /// ORIENTATION; fix up width and length from orientation.
+
+        setMetadata(ifd, core, "SamplesPerPixel", SAMPLESPERPIXEL);
+        setMetadata(ifd, core, "Software", SOFTWARE);
+        setMetadata(ifd, core, "Instrument Make", MAKE);
+        setMetadata(ifd, core, "Instrument Model", MODEL);
+        setMetadata(ifd, core, "Make", MAKE);
+        setMetadata(ifd, core, "Model", MODEL);
```
mtbc: I assume it's okay to use the same make/model for both pairs.

mtbc:
All looks fine to me. The various `minimaltiffwriter-*.tiff` files all have exactly the same content.
```
Series count = 1
Series #0 :
	Image count = 3
	RGB = false (1) 
	Interleaved = false
	Indexed = false (false color)
	Width = 64
	Height = 64
	SizeZ = 1
	SizeT = 3
	SizeC = 1
	Thumbnail size = 64 x 64
	Endianness = intel (little)
	Dimension order = XYCZT (uncertain)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0
	Plane #1 <=> Z 0, C 0, T 1
	Plane #2 <=> Z 0, C 0, T 2


Reading global metadata
BitsPerSample: 8
Compression: Uncompressed
ImageLength: 64
ImageWidth: 64
MetaDataPhotometricInterpretation: Monochrome
MetaMorph: no
NumberOfChannels: 1
PhotometricInterpretation: BlackIsZero
PlanarConfiguration: Planar
SampleFormat: unsigned integer
SamplesPerPixel: 1
```

cpp/lib/ome/bioformats/tiff/Util.cpp
rleigh-dundee: It's a direct copy of the `BaseTiffReader.java` behaviour; it's unchanged in this PR (just moved from IFD.cpp)

cpp/lib/ome/bioformats/out/MinimalTIFFWriter.cpp
rleigh-dundee: Yes, it should be possible:

If BigTIFF has been enabled explicitly or implicitly then the first block runs.
If BigTIFF has been disabled explicitly but is implicitly needed (too big) then the second block runs to issue a warning.  The combinations:

```
BigTIFF  FileSize Action
-------- -------- ------------------------
Enabled  >4g      Enable BigTIFF
Enabled  <4g      Enable BigTIFF
Disabled >4g      Enable BigTIFF, warn (1)
Disabled <4g      None
Unspec   >4g      Enable BigTIFF, warn (2)
Unspec   <4g      None
```

There is a bug here--getBigTIFF is not being called but is being used as a pointer by accident making it a not-null pointer check, so I've pushed a fix for that.

cpp/lib/ome/bioformats/out/MinimalTIFFWriter.cpp
```diff
@@ -0,0 +1,308 @@
...
+            if (!bigTIFF) // Not set manually
+              boost::format fmt
+                ("Pixel data size is %1%, but TIFF without BigTIFF "
+                 "support enabled has a maximum size of %2%; "
+                 "automatically enabling BigTIFF support to prevent potential failure");
+            fmt % pixelSize % std::numeric_limits<uint32_t>::max();
+
+            BOOST_LOG_SEV(logger, ome::logging::trivial::warning) << fmt.str();
+          }
+        else (!getBigTIFF() && needBig)
```
melissalinkert: I understand the possible combinations, but I still don't completely understand the conditionals here.

If ```needBig``` is true, the ```if``` block will be executed, correct?  So if it gets to the point where the ```else``` is evaluated, ```needBig``` must be false, which means the ```else``` block will never be executed?  Or am I missing something really obvious?

rleigh-dundee: Ah right, sorry I didn't see what you meant there.  Yes, that's definitely wrong.  Should be making use of the fact that bigTIFF is optional, so the first should be "(bigTIFF && *bigTIFF) || (!bigTIFF && needBig)" while the second should be "bigTIFF && !*bigTIFF && needBig".  I'll fix this shortly.

rleigh-dundee:
@melissalinkert The BigTIFF warnings etc. should all be fixed now.  I also made it use BigTIFF extensions as for your PR for the Java TIFF writer.

rleigh-dundee:
Note instructions for testing BigTIFF were added to the PR description.

melissalinkert:
BigTIFF changes make more sense now - no real objections to merging at this point.

PR #1744, opened 2015-04-16
TIFF writer: automatically switch to BigTIFF when possible

melissalinkert:
If a BigTIFF extension is used, or we know that more than 4GB of data
will be written, then BigTIFF is used automatically.  This does not
attempt to switch to BigTIFF when compression is used and a BigTIFF
extension is not present, as we can't easily predict the actual size
of the output file in that case.

Fixes https://trac.openmicroscopy.org/ome/ticket/6589.

To test, take a file with 4GB or more of pixel data, e.g. ```large-file&sizeX=2048&sizeY=2048&sizeZ=1200.fake```  Run bfconvert with an output file name that has ```.btf```, ```.tf8```, or ```.tf2``` as the extension and no other arguments; the output file should be in the BigTIFF format, and there should be a ```Switching to BigTIFF``` message printed.  The same test with an output file name that ends in ```.tif``` or ```.tiff``` should print a slightly different ```Switching to BigTIFF``` message and also result in a BigTIFF output file.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -73,6 +73,15 @@ public class TiffWriter extends FormatWriter {
   public static final String COMPRESSION_JPEG =
     CompressionType.JPEG.getCompression();
 
+  private static final String[] BIG_TIFF_SUFFIXES = {"tf2", "tf8", "btf"};
+
+  /**
+   * Number of bytes at which to automatically switch to BigTIFF
+   * This is approximately 3.9 GB instead of 4 GB,
+   * to allow space for the IFDs.
+   */
+  private static final long BIG_TIFF_CUTOFF = 1024 * 1024 * 3990;
+
   // -- Fields --
 
   /** Whether or not the output file is a BigTIFF file. */
@@ -117,7 +126,8 @@ public class TiffWriter extends FormatWriter {
   // -- Constructors --
 
   public TiffWriter() {
-    this("Tagged Image File Format", new String[] {"tif", "tiff"});
+    this("Tagged Image File Format",
+      new String[] {"tif", "tiff", "tf2", "tf8", "btf"});
   }
 
   public TiffWriter(String format, String[] exts) {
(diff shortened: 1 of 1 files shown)


ehrenfeu:
That's great news, thanks!

Just a thought on the side: I would think it was ok to simply switch over to BigTIFF when compression is used if the original (uncompressed) dataset is larger than 4GB - then you'd be on the save side. But personally I'm fine with "actively" switching by simply selecting an appropriate file name suffix.

melissalinkert:
My worry with doing the same thing when compression is enabled is that depending on the data and compression type, > 4GB of raw pixels could easily become < 1GB on disk (e.g. with JPEG-2000 compression), in which case switching to BigTIFF isn't strictly necessary.  It doesn't really matter from the perspective of reading the output file with Bio-Formats/OMERO, but it could easily be an issue if the output file is to be read with software that doesn't yet support BigTIFF.

ehrenfeu:
BTW, as there is no such thing as an "official" BigTIFF standard (at least to my knowledge), the question came up what the Bio-Formats BigTIFF is based upon? Am I right in guessing that you follow the proposals from http://bigtiff.org/ ?

ehrenfeu:
Plus, another minor question, what's the suggested "complete" suffix for a "BigOME-TIFF" file? As we're already having one microscope that stores its datasets in OME-TIFF format, and they're using `.tf8` when writing BigTIFF style, my personal preference is to use this variant for consistency. Still, will there be a difference between
- `.ome.tf8`
- simply `.tf8` without the `.ome` interjection

I guess it won't, no?

dominikl:
Works like expected, good to merge.

melissalinkert:
@ehrenfeu: yes, we are using the format defined by http://bigtiff.org/ and http://www.awaresystems.be/imaging/tiff/bigtiff.html, as that seems to be the closest thing to a standard.

At the moment, the only supported extensions for OME-TIFF are ```.ome.tif``` and ```.ome.tiff```.  The BigTIFF extensions are not currently supported in place of ```.tif/.tiff``` when writing OME-TIFF; using either ```.tf8``` or ```.ome.tf8``` will produce a BigTIFF file without OME-XML metadata.

ehrenfeu:
Oh, that's very surprising to me. So there is no OME'ish way of storing data in *one* file and preserving the metadata?

melissalinkert:
You can write a single > 4GB OME-TIFF file by setting ```-bigtiff``` when running ```bfconvert``` (or calling the corresponding API method), we just don't yet support distinguishing between standard-TIFF OME-TIFF and BigTIFF OME-TIFF via extension.  It's certainly something we can consider, it just hasn't really been discussed up until now.

ehrenfeu:
Ah, my question was more how to deal with this situation from within Fiji: if I now tell an average user to export his data >4GB by putting the `.tf8` suffix, they'll loose the OME-XML metadata - or did I get this wrong?

melissalinkert:
@ehrenfeu: no, you're correct - this pull request won't change things if you are exporting OME-TIFF from Fiji.  We'll discuss whether the OME-TIFF spec can be extended to allow the ```.ome.tf8```, ```.ome.tf2```, and ```.ome.btf``` extensions, and open a new pull request if so.

melissalinkert:
--no-rebase

ehrenfeu:
@melissalinkert that would be awesome!

We're facing this problem actually quite often, imagine a long timelapse where each channel is already larger than 4GB - the only choice to store it in OME-TIFF is to split it by timepoints then, but depending on the setup, this can already create thousands of files with just like 5MB each or so. And small files are a pure nightmare regarding disk I/O, there's not much to be done about this (except of storing the data in a larger container type, of course).

ehrenfeu:
Sorry, I'm still a little confused. I can see openmicroscopy/ome-documentation#1205 has been merged into 5.1.2 - does this mean that using the `.ome.tf8` or any of the other mentioned suffixes will now trigger BigTIFF OME-TIFF writing from the Fiji plugin?

joshmoore:
```
$ bftools/bfconvert niko.fake niko.ome.tf8
niko.fake
FakeReader initializing niko.fake
[Simulated data] -> niko.ome.tf8 [OME-TIFF]
Switching to BigTIFF (by file extension)
        Converted 1/1 planes (100%)
[done]
0.687s elapsed (35.0+233.0ms per plane, 325ms overhead)
```

Yes. :smile:

ehrenfeu:
How did you know I am fake today? :flushed:
However, according to @melissalinkert's comment above, this did already work for `bfconvert` before - but not for the Fiji-plugin. Therefore I'm curious...

joshmoore:
Aha. Then I'll just be quiet. @melissalinkert will certainly set all straight shortly...

melissalinkert:
Yes, BigTIFF OME-TIFF writing should work from Fiji with 5.1.2.  This isn't the PR that introduces that fix though - the relevant code change is gh-1790.

ehrenfeu:
Great, I'll test and let you know about my feelings - latest next week in Paris! :smile:

PR #1745, opened 2015-04-16
DM3: use the DataType tag to set the pixel type

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12559.

To test, use the file from QA 9526 and verify that the pixel type with this change is int32.  Without this change, it would have been uint32.  Note that a number of other sample .dm3 files (in the ```gatan``` directory) changed sign; while I did check each one in ImageJ with the DM3 reader plugin (http://rsb.info.nih.gov/ij/plugins/DM3_Reader.html), it might also be worth spot-checking a few to make sure that the DM3 reader plugin concurs as to the pixel values.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -204,7 +204,9 @@ public class GatanReader extends FormatReader {
     }
     int bytes = numPixelBytes / (getSizeX() * getSizeY());
 
-    m.pixelType = FormatTools.pixelTypeFromBytes(bytes, signed, false);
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
The change is int32 as expected.  I've compared with the DM32 reader plugin and the BF Importer.  For each image I've done a whole-plane measurement for min/max/total, done a difference between the two images and measured that.  The measurements for both images are identical, and the difference is 0/0/0 min/max/total so it looks like the pixel data being read in both cases is identical.

Looks good to merge.

melissalinkert:
--no-rebase

PR #1749, opened 2015-04-20
C++: Reader and writer channel-related fixes

rleigh-dundee:
Both the existing readers and writers were making some mistaken assuptions with SizeC/EffectiveSizeC/ImageCount/RGBChannelCount.  This PR corrects these assumptions to use these data correctly, and it also adds some fixes to the API to handle subchannel data and calculations correctly in all circumstances.  I was going to wait on doing this, as discussed, but the impact of getting things wrong is quite severe.  Unlike the Java situation, where using incorrect dimension sizes/indices will at worst give an ArrayOutOfBoundsException, here we will at best immediately segfault or worse, silently corrupt memory and write out junk data.  I thought it better to err on the side of caution rather than give the user incorrect data with potentially catastrophic results.

- CoreMetadata: Make `sizeC` a vector of integers to accurately represent SamplesPerPixel subchannels on a per-channel basis; remove the now redundant `rgb` field.  SizeC is the sum of all subchannels; EffectiveSizeC is the size of the vector.  `imageCount` is retained as a sanity check but otherwise redundant.
- FormatReader: `isRGB`, `getRGBChannelCount`, `getOptimalTileWidth` and `getOptimalTileHeight` have a `channel` parameter to specify the channel to use, and make use of the per-channel subchannel count.  `getSizeC` and `getEffectiveSizeC` updated to use the new calculations.
- FormatWriter already provided a subset of these methods, which are extended to provide an equivalent set (using MetadataStore rather than CoreMetadata)
- MetadataTools updated to use updated channel count calculations
- All readers can now read RGB data with correct plane indexing and buffer sizing
- MinimalTIFFWriter can now write RGB data
- OMETIFFReader initFile logic updated to use subchannels; this has simplified it a bit
- ImageInfo updated to display subchannels
- Testcases updated

Additional minor API fixes:

- Parameter names and API docs updated to replace the generic `no` by `series`, `plane` etc. to make the API more self-describing; similar for sample counts.
- LookupTable parameters swapped to make the API consistent for methods using plane index and buffer arguments; fixed assertions and unit test for default LUT behaviour
- FormatWriter has added interleaving methods, used by the TIFF writer

--------

Question: The addition of the `channel` argument works, but an alternative would be to use a `plane` argument to make the method calls align open/saveBytes where the information is most likely to be used (these would internally call getZCTCoords with the provided plane number to get the channel).  Would that be better than channel, or is channel OK?  From the POV of display/rendering, channel is what's needed, but for the user of openBytes/saveBytes, it saves them doing the getZCTCoords themselves to do the lookup before calling methods with channel arguments.

--------

Testing: All the dimension-related changes are tested by the unit tests.  The TIFF writing changes may be tested by editing `cpp/test/ome-bioformats/minimaltiffwriter.cpp` and changing:

```
diff --git a/cpp/test/ome-bioformats/minimaltiffwriter.cpp b/cpp/test/ome-bioformats/minimaltiffwriter.cpp
index b042a89..9734bc4 100644
--- a/cpp/test/ome-bioformats/minimaltiffwriter.cpp
+++ b/cpp/test/ome-bioformats/minimaltiffwriter.cpp
@@ -124,8 +124,8 @@ public:
   TearDown()
   {
     // Delete file (if any)
-    if (boost::filesystem::exists(testfile))
-      boost::filesystem::remove(testfile);
+    // if (boost::filesystem::exists(testfile))
+    //   boost::filesystem::remove(testfile);
   }
 };
 
@@ -148,7 +148,7 @@ TEST_P(TIFFWriterTest, setId)
 
   tiffwriter.setMetadataRetrieve(retrieve);
 
-  bool interleaved = true;
+  bool interleaved = false;
 
   tiffwriter.setInterleaved(interleaved);
 ```

After running the tests, look at the files in `cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-*`.  These should be RGB with one IFD (previously greyscale with 3 IFDs), for example:

```
% tiffinfo cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-planar-tiles-48x16.tiff
TIFF Directory at offset 0x3008 (12296)
  Image Width: 64 Image Length: 64
  Bits/Sample: 8
  Compression Scheme: None
  Photometric Interpretation: RGB color
  Samples/Pixel: 3
  Rows/Strip: 1
  Software: OME Bio-Formats (C++) 5.1.0
```

Also, run `./bf-test ./cpp/libexec/info/info /path/to/ome-tiff` with a 2013-06 OME-TIFF; ImageInfo will show the constituent subchannels in the RGB output.

### DIFF
--- a/cpp/lib/ome/bioformats/FormatReader.h
+++ b/cpp/lib/ome/bioformats/FormatReader.h
@@ -212,15 +212,16 @@ namespace ome
       getImageCount() const = 0;
 
       /**
-       * Check if the image planes in the file have more than one channel per
-       * openBytes() call.
+       * Check if the image planes for a channel have more than one
+       * subchannel per openBytes() call.
        *
+       * @param channel the channel to use.
        * @returns @c true if and only if getRGBChannelCount() returns
        * a value greater than 1, @c false otherwise.
        */
       virtual
       bool
-      isRGB() const = 0;
+      isRGB(dimension_size_type channel = 0U) const = 0;
 
       /**
        * Get the size of the X dimension.
@@ -305,18 +306,19 @@ namespace ome
       getEffectiveSizeC() const = 0;
 
       /**
-       * Get the number of channels returned with each call to openBytes().
+       * Get the number of channels returned for a call to openBytes().
        *
        * The most common case where this value is greater than 1 is for interleaved
        * RGB data, such as a 24-bit color image plane. However, it is possible for
        * this value to be greater than 1 for non-interleaved data, such as an RGB
        * TIFF with Planar rather than Chunky configuration.
        *
+       * @param channel the channel to use.
        * @returns the number of channels.
        */
       virtual
       dimension_size_type
-      getRGBChannelCount() const = 0;
+      getRGBChannelCount(dimension_size_type channel = 0U) const = 0;
 
       /**
        * Get whether the image planes are indexed color.
@@ -1070,21 +1072,23 @@ namespace ome
        * Get the optimal sub-image width.
        * This is intended for use with openBytes().
        *
+       * @param channel the channel to use.
        * @returns the optimal width.
        **/
       virtual
       dimension_size_type
-      getOptimalTileWidth() const = 0;
+      getOptimalTileWidth(dimension_size_type channel = 0U) const = 0;
 
       /**
        * Get the optimal sub-image height.
        * This is intended for use with openBytes().
        *
+       * @param channel the channel to use.
        * @returns the optimal height.
        **/
       virtual
       dimension_size_type
-      getOptimalTileHeight() const = 0;
+      getOptimalTileHeight(dimension_size_type channel = 0U) const = 0;
 
       // Sub-resolution API methods
 
--- a/cpp/lib/ome/bioformats/FormatReader.h
+++ b/cpp/lib/ome/bioformats/FormatReader.h
@@ -221,7 +221,7 @@ namespace ome
        */
       virtual
       bool
-      isRGB(dimension_size_type channel = 0U) const = 0;
+      isRGB(dimension_size_type channel) const = 0;
 
       /**
        * Get the size of the X dimension.
@@ -318,7 +318,7 @@ namespace ome
        */
       virtual
       dimension_size_type
-      getRGBChannelCount(dimension_size_type channel = 0U) const = 0;
+      getRGBChannelCount(dimension_size_type channel) const = 0;
 
       /**
        * Get whether the image planes are indexed color.
@@ -365,7 +365,7 @@ namespace ome
       virtual
       void
       getLookupTable(VariantPixelBuffer& buf,
-                     dimension_size_type plane = 0U) const = 0;
+                     dimension_size_type plane) const = 0;
 
       /**
        * Get the Modulo subdivision of the Z dimension.
@@ -1077,7 +1077,7 @@ namespace ome
        **/
       virtual
       dimension_size_type
-      getOptimalTileWidth(dimension_size_type channel = 0U) const = 0;
+      getOptimalTileWidth(dimension_size_type channel) const = 0;
 
       /**
        * Get the optimal sub-image height.
@@ -1088,7 +1088,7 @@ namespace ome
        **/
       virtual
       dimension_size_type
-      getOptimalTileHeight(dimension_size_type channel = 0U) const = 0;
+      getOptimalTileHeight(dimension_size_type channel) const = 0;
 
       // Sub-resolution API methods
 
--- a/cpp/lib/ome/bioformats/FormatReader.h
+++ b/cpp/lib/ome/bioformats/FormatReader.h
@@ -359,13 +359,13 @@ namespace ome
        * but other sizes are possible.
        *
        * @param buf the destination pixel buffer.
-       * @param no the image index within the file.
+       * @param plane the plane index within the series.
        * @throws FormatException if a lookup table could not be obtained.
        */
       virtual
       void
       getLookupTable(VariantPixelBuffer& buf,
-                     dimension_size_type no = 0U) const = 0;
+                     dimension_size_type plane = 0U) const = 0;
 
       /**
        * Get the Modulo subdivision of the Z dimension.
@@ -531,34 +531,34 @@ namespace ome
       /**
        * Obtain an image plane.
        *
-       * Obtain and copy the image plane from the current file into a
-       * VariantPixelBuffer of size
+       * Obtain and copy the image plane from the current series into
+       * a VariantPixelBuffer of size
        *
        * \code{.cpp}
-       * getSizeX * getSizeY * bytesPerPixel * getRGBChannelCount()
+       * getSizeX * getSizeY * bytesPerPixel * getRGBChannelCount(channel)
        * \endcode
        *
-       * @param no the image index within the file.
+       * @param plane the plane index within the series.
        * @param buf the destination pixel buffer.
        * @throws FormatException if there was a problem parsing the
        *   metadata of the file.
        */
       virtual
       void
-      openBytes(dimension_size_type no,
+      openBytes(dimension_size_type plane,
                 VariantPixelBuffer& buf) const = 0;
 
       /**
        * Obtain a sub-image of an image plane.
        *
        * Obtain and copy the sub-image of an image plane from the
-       * current file into a VariantPixelBuffer of size
+       * current series into a VariantPixelBuffer of size
        *
        * \code{.cpp}
-       * w * h * bytesPerPixel * getRGBChannelCount()
+       * w * h * bytesPerPixel * getRGBChannelCount(channel)
        * \endcode
        *
-       * @param no the image index within the file.
+       * @param plane the plane index within the series.
        * @param buf the destination pixel buffer.
        * @param x the @c X coordinate of the upper-left corner of the sub-image.
        * @param y the @c Y coordinate of the upper-left corner of the sub-image.
@@ -569,7 +569,7 @@ namespace ome
        */
       virtual
       void
-      openBytes(dimension_size_type no,
+      openBytes(dimension_size_type plane,
                 VariantPixelBuffer& buf,
                 dimension_size_type x,
                 dimension_size_type y,
@@ -580,14 +580,14 @@ namespace ome
        * Obtain a thumbnail of an image plane.
        *
        * Obtail and copy the thumbnail for the specified image plane
-       * from the current file into a VariantPixelBuffer.
+       * from the current series into a VariantPixelBuffer.
        *
-       * @param no the image index within the file.
+       * @param plane the plane index within the series.
        * @param buf the destination pixel buffer.
        */
       virtual
       void
-      openThumbBytes(dimension_size_type no,
+      openThumbBytes(dimension_size_type plane,
                      VariantPixelBuffer& buf) const = 0;
 
       /**
@@ -607,7 +607,7 @@ namespace ome
        *
        * @note This also resets the resolution to 0.
        *
-       * @param no the series to activate.
+       * @param series the series to activate.
        *
        * @todo Remove use of stateful API which requires use of
        * series switching in const methods.
@@ -616,7 +616,7 @@ namespace ome
        */
       virtual
       void
-      setSeries(dimension_size_type no) const = 0;
+      setSeries(dimension_size_type series) const = 0;
 
       /**
        * Get the active series.
--- a/cpp/lib/ome/bioformats/FormatReader.h
+++ b/cpp/lib/ome/bioformats/FormatReader.h
@@ -213,9 +213,9 @@ namespace ome
 
       /**
        * Check if the image planes for a channel have more than one
-       * subchannel per openBytes() call.
+       * sub-channel per openBytes() call.
        *
-       * @param channel the channel to use.
+       * @param channel the channel to use, range [0, EffectiveSizeC).
        * @returns @c true if and only if getRGBChannelCount() returns
        * a value greater than 1, @c false otherwise.
        */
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

**Comment by mtbc** [comment]
```
The following tests FAILED:
	 36 - ome-bioformats/minimaltiffwriter (Failed)
Errors while running CTest
```
though I don't know if that's caused by,
```diff
diff --git a/cpp/test/ome-bioformats/minimaltiffwriter.cpp b/cpp/test/ome-bioformats/minimaltiffwriter.cpp
index 0155f90..0355a44 100644
--- a/cpp/test/ome-bioformats/minimaltiffwriter.cpp
+++ b/cpp/test/ome-bioformats/minimaltiffwriter.cpp
@@ -124,8 +124,8 @@ public:
   TearDown()
   {
     // Delete file (if any)
-    if (boost::filesystem::exists(testfile))
-      boost::filesystem::remove(testfile);
+//    if (boost::filesystem::exists(testfile))
+//      boost::filesystem::remove(testfile);
   }
 };
 
@@ -147,7 +147,7 @@ TEST_P(TIFFWriterTest, setId)
   ome::compat::shared_ptr< ::ome::xml::meta::MetadataRetrieve> retrieve(ome::compat::static_pointer_cast< ::ome::xml::meta::MetadataRetrieve>(meta));
 
   tiffwriter.setMetadataRetrieve(retrieve);
-  tiffwriter.setInterleaved(true);
+  tiffwriter.setInterleaved(false);
 
   ASSERT_NO_THROW(tiffwriter.setId(testfile));
```

mtbc:
```
Series count = 1
Series #0 :
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (false color)
	Width = 64
	Height = 64
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Thumbnail size = 64 x 64
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0
```
seems okay. As with #1743 all the `minimaltiffwriter-data-layout-64x64-*.tiff` have the same content.

rleigh-dundee:
@mtbc This was the diff in the testing instructions being wrong.  I've pushed a fix to make the interleaving customisable, and updated the diff.  It now works with `interleaving=true|false` and tiffinfo will report `single` or `separate` image planes, respectively.

mtbc:
```
100% tests passed, 0 tests failed out of 48
```
```
Series count = 1
Series #0 :
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (false color)
	Width = 64
	Height = 64
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Thumbnail size = 64 x 64
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata
BitsPerSample: 8
Compression: Uncompressed
ImageLength: 64
ImageWidth: 64
MetaDataPhotometricInterpretation: RGB
MetaMorph: no
NumberOfChannels: 3
PhotometricInterpretation: RGB
PlanarConfiguration: Planar
SampleFormat: unsigned integer
SamplesPerPixel: 3
```

mtbc:
`./bf-test` is showing me,
```
RGB = [false, false, false] ([1, 1, 1]) 
```

rleigh-dundee:
@mtbc Hmm, it should be a single channel if the input is a single RGB plane:

Note: `      reader = ome::compat::make_shared<in::TIFFReader>();` in `cpp/libexec/info/ImageInfo.cpp` to force it to use the TIFF reader.

```
% ./bf-test ./cpp/libexec/info/info cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff
Image: cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff
Using reader: TIFF (Tagged Image File Format)
getFieldInfo: tag 50838 unknown
Reader setup took 00:00:00.002679

Filename = "/tmp/b/cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff"
Used files = ["/tmp/b/cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff"]

Reading core metadata
Series count = 1

Series #0:
        Image count = 1
        RGB = [true] ([3]) 
        Interleaved = false
        Indexed = false
        Width = 64
        Height = 64
        SizeZ = 1 (effectively 1)
        SizeT = 1 (effectively 1)
        SizeC = 3 (effectively 1)
        Thumbnail size = 64 × 64
        Endianness = little
        DimensionOrder = XYCZT (certain)
        PixelType = uint8
        Bits per Pixel = 8
        MetadataComplete = true
        ThumbnailSeries = false

No global metadata

Series #0 metadata:
        BitsPerSample: 8
        Compression: 1
        ImageLength: 64
        ImageWidth: 64
        NumberOfChannels: 3
        PhotometricInterpretation: 2
        Planar Configuration: 2
        SampleFormat: 1
        SamplesPerPixel: 3
        Software: OME Bio-Formats (C++) 5.1.0
```

mtbc:
Huh, with those I get,
```
mtbc@ls28101:/tmp/build$ ./bf-test ./cpp/libexec/info/info cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff 
Image: cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff
E: No TIFF ImageDescription found
```

rleigh-dundee:
@mtbc You need to make the suggested correction to ImageInfo for it to use the TIFF reader, otherwise it tries to get the OME-XML comment (which isn't present).

Also pushed a small fix for the interleaved reader property which was previously fixed but now reflects the interleaved property when writing, via PlanarConfiguration.

mtbc:
Sorry, didn't realize it was a suggested correction! :smiley:

mtbc:
```diff
diff --git a/cpp/libexec/info/ImageInfo.cpp b/cpp/libexec/info/ImageInfo.cpp
index 96319a1..5ab403f 100644
--- a/cpp/libexec/info/ImageInfo.cpp
+++ b/cpp/libexec/info/ImageInfo.cpp
@@ -78,7 +78,7 @@ namespace info
   void
   ImageInfo::setReader(ome::compat::shared_ptr<FormatReader>& reader)
   {
-    this->reader = reader;
+    this->reader = ome::compat::make_shared<in::TIFFReader>();
   }
 
   void
diff --git a/cpp/test/ome-bioformats/minimaltiffwriter.cpp b/cpp/test/ome-bioformats/minimaltiffwriter.cpp
index b042a89..3301f3c 100644
--- a/cpp/test/ome-bioformats/minimaltiffwriter.cpp
+++ b/cpp/test/ome-bioformats/minimaltiffwriter.cpp
@@ -124,8 +124,8 @@ public:
   TearDown()
   {
     // Delete file (if any)
-    if (boost::filesystem::exists(testfile))
-      boost::filesystem::remove(testfile);
+    //if (boost::filesystem::exists(testfile))
+    //  boost::filesystem::remove(testfile);
   }
 };
 
@@ -148,7 +148,7 @@ TEST_P(TIFFWriterTest, setId)
 
   tiffwriter.setMetadataRetrieve(retrieve);
 
-  bool interleaved = true;
+  bool interleaved = false;
 
   tiffwriter.setInterleaved(interleaved);
```
gives,
```
mtbc@ls28101:/tmp/build$ ./bf-test ./cpp/libexec/info/info cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff
Image: cpp/test/ome-bioformats/data/minimaltiffwriter-data-layout-64x64-chunky-strips-01.tiff
E: No TIFF ImageDescription found
```
I am currently at commit `be89539f`.

Maybe push this PR for another round of testing tomorrow?

rleigh-dundee:
@mtbc The line

```
```

would indicate you're not using the changes from your diff--it's still using OMETIFFReader, not TIFFReader.  Did you rerun make?

mtbc:
I wiped out the build directory entirely and rebuilt.

rleigh-dundee:
@mtbc Ah, your change is in the wrong place; the method you used isn't called.  You want the line in `testRead()`, where OMETIFFReader is changed to TIFFReader:

```
  void
  ImageInfo::testRead(std::ostream& stream)
  {
    if (!reader)
      reader = ome::compat::make_shared<in::TIFFReader>();
```

cpp/lib/ome/bioformats/FormatReader.h
```diff
@@ -1070,21 +1072,23 @@ namespace ome
...
        * Get the optimal sub-image height.
        * This is intended for use with openBytes().
        *
+       * @param channel the channel to use.
        * @returns the optimal height.
        **/
       virtual
       dimension_size_type
-      getOptimalTileHeight() const = 0;
+      getOptimalTileHeight(dimension_size_type channel) const = 0;
```
melissalinkert: Not sure I follow why ```getOptimalTile*``` are per-channel now.  Tile dimensions shouldn't be affected by channel count or layout.

rleigh-dundee: @melissalinkert It's from https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-api/src/loci/formats/FormatReader.java#L1242 due to the call to getRGBChannelCount.

melissalinkert: OK, but I'm not thoroughly convinced we want the tile dimensions to be variable like that.  Maybe use the maximum value of ```getRGBChannelCount``` instead?

melissalinkert:
It might be helpful to update the instances of ```@param channel``` to explain what the maximum value of that parameter should be.

I assume the new way of recording channels is flexible enough to work with e.g. ModuloC data without requiring further API changes?

cpp/lib/ome/bioformats/FormatReader.h
rleigh-dundee: I agree it's not ideal, it does expose what should really be an implementation detail.  But is this an implementation detail in all circumstances?  If the model is using different sample counts over different channels, and the tiling does alter as a result, then this would be an appropriate way to expose that.

If some channels have different tile/strip sizes due to different sample counts, then removing the argument could lead to suboptimal performance for some channels/planeIndexes, while having the argument means you can pass through the correct sizes in e.g. TiffReader where we can get the values directly from the IFD.  I guess that's one argument in favour of using the dimension number rather than channel--the reader can map the values directly onto the specific IFD for formats which allow that.  Using the highest value of getRGBChannelCount would certainly mitigate the problem though.

I'm not currently decided which way is better.  Both have their points.

rleigh-dundee:
I'll update the channel description; unless you'd prefer it switching to use plane indexes instead as mentioned in the PR description?

The modulo question is a good one.  Since the ModuloC is subdividing the existing C, but using EffectiveSizeC, it won't be affected as far as I can see.  The Modulo stuff doesn't consider SamplesPerPixel varying across its span, but this seems like it is certainly allowed-but-quite-evil (but not changed by this PR).  The FormatTools unit tests make sure that the modulo indexing isn't broken by these changes, and I updated the tests to check it still works correctly with varying subchannels, so I *think* this is taken care of for now, but Modulo is still a bit of a nasty hack which would be great to replace with nD down the line to clean up the complexity!

mtbc:
```
Series #0:
	Image count = 1
	RGB = [true] ([3]) 
	Interleaved = false
	Indexed = false
	Width = 64
	Height = 64
	SizeZ = 1 (effectively 1)
	SizeT = 1 (effectively 1)
	SizeC = 3 (effectively 1)
	Thumbnail size = 64 × 64
	Endianness = little
	DimensionOrder = XYCZT (certain)
	PixelType = uint8
	Bits per Pixel = 8
	MetadataComplete = true
	ThumbnailSeries = false

No global metadata

Series #0 metadata:
	BitsPerSample: 8
	Compression: 1
	ImageLength: 64
	ImageWidth: 64
	NumberOfChannels: 3
	PhotometricInterpretation: 2
	Planar Configuration: 2
	SampleFormat: 1
	SamplesPerPixel: 3
	Software: OME Bio-Formats (C++) 5.1.0
```
and,
```
Series #0 :
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (false color)
	Width = 64
	Height = 64
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Thumbnail size = 64 x 64
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata
BitsPerSample: 8
Compression: Uncompressed
ImageLength: 64
ImageWidth: 64
MetaDataPhotometricInterpretation: RGB
MetaMorph: no
NumberOfChannels: 3
PhotometricInterpretation: RGB
PlanarConfiguration: Planar
SampleFormat: unsigned integer
SamplesPerPixel: 3
Software: OME Bio-Formats (C++) 5.1.0
```
So, looks good functionally.

cpp/lib/ome/bioformats/FormatReader.h
rleigh-dundee: One counter to using planeIndex rather than channel is that isInterleaved is already using channel rather than planeIndex.

rleigh-dundee:
Channel range is now documented.

rleigh-dundee:
I think the summary of the current state would be that everything is working as intended and tests correctly.  The outstanding question is the signature of the `getOptimalTile*` methods: should they include zero arguments, channel or plane?  Or should we continue to use channel but have a zero-arg overload which uses the heuristic suggested by @melissalinkert?  This is a bit more convienient, but at the cost of being strictly correct (may result in suboptimal performance if not correct).  But maybe that's acceptable?  Happy to chat about this if needed.

melissalinkert:
Current state is fine I think - merging for 5.1.1.

rleigh-dundee:
Thanks.  I'll follow this up with the OMETIFFWriter very soon.

PR #1750, opened 2015-04-21
matlab: restore state of the warnings instead of turning all of them 'on'

carandraug:
The existing code was modifying the state of the warnings but was turning them all back on instead of restoring to the previous state.

Ideally, this would also turn off only the specific warning that was causing issues but I can't figure out what warning was that. This was done in 664274bb but the commit message does not give me any hints.

### DIFF
--- a/components/formats-gpl/matlab/bfopen.m
+++ b/components/formats-gpl/matlab/bfopen.m
@@ -153,14 +153,14 @@ for s = 1:numSeries
         else
             colorMaps{s, i} = r.get16BitLookupTable()';
         end
-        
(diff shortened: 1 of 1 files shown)


sbesson:
CI jobs are green and this commit definitely makes sense. Ready to merge for 5.1.1. I will add a card to 1- extend the bfopen tests to cover colormaps (indexed) and 2- investigate the warnings requirement (may have been related to an old version of MATLAB).

melissalinkert:
Merging - thanks, @carandraug.

PR #1752, opened 2015-04-21
Performance updates for working with large pattern filesets

melissalinkert:
To test, run ```echo "test_Z<00000-60000>.fake" > fake.pattern```.  ```time showinf -nopix fake.pattern``` should result in an OOM without this change.  The same test with 4b223c4 should run to completion, but be fairly slow (1m50s locally); the same test with all commits should run to completion and be substantially faster (16s locally).

### DIFF
--- a/components/formats-bsd/src/loci/formats/FileStitcher.java
+++ b/components/formats-bsd/src/loci/formats/FileStitcher.java
@@ -38,7 +38,7 @@ import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Set;
@@ -667,7 +667,7 @@ public class FileStitcher extends ReaderWrapper {
     // have to call initFile on each constituent file; but we can only do so
     // when each constituent file does not itself have multiple used files
 
-    Set<String> files = new HashSet<String>();
+    Set<String> files = new LinkedHashSet<String>();
     for (ExternalSeries s : externals) {
       String[] f = s.getFiles();
       for (String file : f) {
--- a/components/formats-bsd/src/loci/formats/FileStitcher.java
+++ b/components/formats-bsd/src/loci/formats/FileStitcher.java
@@ -65,6 +65,8 @@ public class FileStitcher extends ReaderWrapper {
   private static final Logger LOGGER =
     LoggerFactory.getLogger(FileStitcher.class);
 
+  private static final int MAX_READERS = 1000;
+
   // -- Fields --
 
   /**
@@ -181,7 +183,7 @@ public class FileStitcher extends ReaderWrapper {
    */
   public DimensionSwapper getReader(int series, int no) {
     if (noStitch) return (DimensionSwapper) reader;
-    DimensionSwapper r = externals[getExternalSeries(series)].getReaders()[no];
+    DimensionSwapper r = externals[getExternalSeries(series)].getReader(no);
     initReader(series, no);
     return r;
   }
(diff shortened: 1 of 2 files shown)


melissalinkert:
Taking this back out of breaking now that 5.1.1 is out.

melissalinkert:
Introduces a couple of failures in the quicktime directory, see https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/478/consoleText.  Excluding temporarily.

melissalinkert:
Commit pushed to fix the quicktime directory failures.  Removing the exclude label.

joshmoore:
on trout/merge today:
```
Series #0 :
        Image count = 60001
        RGB = false (1)
        Interleaved = false
        Indexed = false (true color)
        Width = 512
        Height = 512
        SizeZ = 60001
        SizeT = 1
        SizeC = 1
        Thumbnail size = 128 x 128
        Endianness = intel (little)
        Pixel type = uint8
        Valid bits per pixel = 8
        Metadata complete = true
        Thumbnail series = false
        -----
        Plane #0 <=> Z 0, C 0, T 0
        Plane #29998 <=> Z 29998, C 0, T 0
        Plane #29999 <=> Z 29999, C 0, T 0
        Plane #30000 <=> Z 30000, C 0, T 0
        Plane #30001 <=> Z 30001, C 0, T 0
        Plane #30002 <=> Z 30002, C 0, T 0
        Plane #60000 <=> Z 60000, C 0, T 0


Reading global metadata

Reading metadata

real    0m7.594s
user    0m9.015s
sys     0m13.619s
```

joshmoore:
Questions:
 * to understand the implications, the impact of 4b223c4c82272700231641442c4974216ff90c6a is that there is more context switching between different calls to openBytes? What would be the easiest way to gauge this impact?
 * what's the basis for the 1000 `MAX_READERS` setting? Looking at `900 < x < 1000` files certainly seems sensible for fakes (see below). Just wondering for other formats if there could be an impact?

991 files:
```
Series #0 :
        Image count = 991
        RGB = false (1)
        Interleaved = false
        Indexed = false (true color)
        Width = 512
        Height = 512
        SizeZ = 991
        SizeT = 1
        SizeC = 1
        Thumbnail size = 128 x 128
        Endianness = intel (little)
        Pixel type = uint8
        Valid bits per pixel = 8
        Metadata complete = true
        Thumbnail series = false
        -----
        Plane #0 <=> Z 0, C 0, T 0
        Plane #493 <=> Z 493, C 0, T 0
        Plane #494 <=> Z 494, C 0, T 0
        Plane #495 <=> Z 495, C 0, T 0
        Plane #496 <=> Z 496, C 0, T 0
        Plane #497 <=> Z 497, C 0, T 0
        Plane #990 <=> Z 990, C 0, T 0


Reading global metadata

Reading metadata

real    0m3.204s
user    0m5.356s
sys     0m8.139s
```

melissalinkert:
The impact of 4b223c4 is that when more than 1000 files are present, the same reader will be used for all files.  Reading planes strictly in order shouldn't be any slower (so things like conversion shouldn't be impacted), but e.g. reading from file 0, then 1, then 0 results in 3 setId calls where previously there would have been 2.

1000 is a semi-arbitrary choice, and can certainly be changed.  Mostly I wanted something large enough to not impact the many small datasets in the repo that require FileStitcher, but small enough to still prevent memory problems.

melissalinkert:
Anything further that needs to happen before this is OK to merge?

joshmoore:
Sorry for letting this linger. Been debating whether or not to try to get a file set into regular testing. That can be done independently though.

melissalinkert:
Merging then, since this at least hasn't had a negative impact on builds.

PR #1755, opened 2015-04-21
SlideBook 6 reader

RichardMyers:
First version of new SlideBook6 reader plugin.
- this reader is based on a JNI DLL wrapper that calls SlideBook 6 file format reader DLL
- reader includes simple meta data support
- DLLS are for Windows and x64 only

To work, the following .dll files must be in the ImageJ bin path (eg. jre/bin):
  SlideBook6Reader.dll  [x32](https://drive.google.com/open?id=0B5XYovwfrhQDfnRlTHBJUURJbHdBREJuZGJlYUVHdnN3Z1NzSUthMm0zWWZ5WEpPQl93a1E&authuser=0) [x64](https://drive.google.com/open?id=0B5XYovwfrhQDflN2QU1INUwzRC1pMVA3WHlyWExiUjQ2bWptcGIyLTZTXy1jZ250Uzc2RDA&authuser=0)
  SBReadFile.dll  [x32](https://drive.google.com/open?id=0B5XYovwfrhQDfnRlTHBJUURJbHdBREJuZGJlYUVHdnN3Z1NzSUthMm0zWWZ5WEpPQl93a1E&authuser=0)  [x64](https://drive.google.com/open?id=0B5XYovwfrhQDflN2QU1INUwzRC1pMVA3WHlyWExiUjQ2bWptcGIyLTZTXy1jZ250Uzc2RDA&authuser=0) 

Should I put the two required DLLs in the bioformats source tree? 

They will also be distributed as part of our free [SlideBook Reader](https://www.intelligent-imaging.com/reader.php) application.

Questions? comments?

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
@@ -1,0 +1,402 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+ * %%
+ * Copyright (C) 2005 - 2015 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/gpl-2.0.html>.
+ * #L%
+ */
+
+package loci.formats.in;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.Hashtable;
+import java.util.Vector;
+
+import loci.common.Constants;
+import loci.common.DataTools;
+import loci.common.RandomAccessInputStream;
+import loci.formats.CoreMetadata;
+import loci.formats.FormatException;
+import loci.formats.FormatReader;
+import loci.formats.FormatTools;
+import loci.formats.MetadataTools;
+import loci.formats.meta.MetadataStore;
+
+import ome.xml.model.primitives.PositiveFloat;
+import ome.units.quantity.Length;
+import ome.units.quantity.Time;
+import ome.units.UNITS;
+
+import loci.formats.MissingLibraryException;
+
+/**
+ * SlideBook6Reader is a file format reader for 3i SlideBook SLD files that uses
+ * the SlideBook SBReadFile SDK via the (Windows only) dlls: 
+ *   SBReadFile.dll and SlideBook6Reader.dll
+ *
+ * @author Richard Myers, richard at intelligent-imaging.com
+ *
+ */
+public class SlideBook6Reader  extends FormatReader {
+
+	// -- Constants --
+
+	public static final int SLD_MAGIC_BYTES_1_0 = 0x006c;
+	public static final int SLD_MAGIC_BYTES_1_1 = 0x0100;
+	public static final int SLD_MAGIC_BYTES_1_2 = 0x0200;
+	public static final int SLD_MAGIC_BYTES_2_0 = 0x01f5;
+	public static final int SLD_MAGIC_BYTES_2_1 = 0x0102;
+
+	public static final long SLD_MAGIC_BYTES_3 = 0xf6010101L;
+
+	private static final String URL_3I_SLD =
+			"http://www.openmicroscopy.org/site/support/bio-formats/formats/3i-slidebook6-sld.html";
+	private static final String NO_3I_MSG = "3i SlideBook SBReadFile library not found. " +
+			"Please see " + URL_3I_SLD + " for details.";
+	private static final String GENERAL_3I_MSG = "3i SlideBook SBReadFile library problem. " +
+			"Please see " + URL_3I_SLD + " for details.";
+
+	// -- Static initializers --
+
+	private static boolean libraryFound = false;
+
+	static {
+		String library_path = System.getProperty("java.library.path");
+
+		try {
+			// load JNI wrapper of SBReadFile.dll
+			System.loadLibrary("SlideBook6Reader");
+			libraryFound = true;
+		}
+		catch (UnsatisfiedLinkError e) {
+			LOGGER.warn(NO_3I_MSG, e);
+			libraryFound = false;
+		}
+		catch (SecurityException e) {
+			LOGGER.warn("Insufficient permission to load native library", e);
+			libraryFound = false;
+		}
+	}
+
+	// -- Constructor --
+
+	public SlideBook6Reader() {
+		super("SlideBook 6 SLD (native)", new String[] {"sld"});
+		domains = new String[] {FormatTools.LM_DOMAIN};
+		suffixSufficient = false;
+	}
+
+	// -- IFormatReader API methods --
+
+	/* @see loci.formats.IFormatReader#isThisType(RandomAccessInputStream) */
+	public boolean isThisType(RandomAccessInputStream stream) throws IOException {
+		final int blockLen = 8;
+		stream.seek(4);
+		boolean littleEndian = stream.readString(2).equals("II");
+		if (!FormatTools.validStream(stream, blockLen, littleEndian)) return false;
+		int magicBytes1 = stream.readShort();
+		int magicBytes2 = stream.readShort();
+
+		boolean isMatch = ((magicBytes2 & 0xff00) == SLD_MAGIC_BYTES_1_1 ||
+				(magicBytes2 & 0xff00) == SLD_MAGIC_BYTES_1_2) &&
+				(magicBytes1 == SLD_MAGIC_BYTES_1_0 ||
+				magicBytes1 == SLD_MAGIC_BYTES_2_0);
+
+		return isMatch;
+	}
+
+	/**
+	 * @see IFormatReader#isThisType(String, boolean)
+	 */
+	public boolean isThisType(String file, boolean open) {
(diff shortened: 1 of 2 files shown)


melissalinkert:
Thank you for this, @RichardMyers.

I'll test this more thoroughly on Windows tomorrow, but for the moment could you please remove the comments that are directly copied from the reader guide documentation?

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,402 @@
...
+			Length physicalSizeZ = FormatTools.getPhysicalSizeZ(stepSize);
+			if (physicalSizeZ != null) {
+				store.setPixelsPhysicalSizeZ(physicalSizeZ, capture);
+			}
+
+			// link other meta data to images
+			int numPositions = getNumPositions(capture);
+			int numTimepoints = getNumTimepoints(capture);
+			int numChannels = getNumChannels(capture);
+			int numZPlanes = getNumZPlanes(capture);
```
melissalinkert: It might be a little faster to use the ```sizeZ```, ```sizeT```, ```sizeC``` values from CoreMetadata (as set above) instead of making additional native method calls here.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,402 @@
...
+		// super.initFile(String). You will also need to set up the stream for reading the file, as well as initializing
+		// any dimension information and metadata. Most of this logic is up to you; however, you should populate the 'core'
+		// variable (see loci.formats.CoreMetadata).
+
+		// Note that each variable is initialized to 0 or null when super.initFile(String) is called. Also,
+		// super.initFile(String) constructs a Hashtable called "metadata" where you should store any relevant metadata.
+
+		try {
+			openFile(id);
+			int numSeries = getNumCaptures();
```
melissalinkert: Is it possible to have multiple positions or montage tiles per capture?  If so, ```numSeries``` should likely reflect the sum of all positions and montage tiles across all captures - otherwise only the first position from each tile will be accessible.

RichardMyers: I see what you mean; is it possible to have dimensionOrder="XYCZPT" ?  When we capture a montage we capture all channels and planes for each montage xy stage position and may perform that loop for multiple time points.

joshmoore:
Closing & re-opening to adjust travis status.

RichardMyers:
I have updated the dll links and plugin to handle montages properly.   Comments are cleaned now too; let me know if you think it needs anything else.  I think the next step is to get some sample data for testing.

Montages positions are handled as time points in a single series, but each position of a multipoint capture that is not a montage is a new series.  Does that sound correct to you?  This matches how we treat these situations in SlideBook.

melissalinkert:
That looks much better, thanks.  There is one issue with the close() method, which I'll note separately, but this reads OK now.

We have plenty of test data already - as soon as I have access to the latest .dlls, I'll test here and verify that the images and generated OME-XML look fine.  The handling of montage positions should be fine, particularly if it matches what SlideBook users are used to seeing.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,393 @@
...
+				}
+			}
+		}
+		return buf;
+	}
+
+	// -- Internal FormatReader API methods --
+	public void close(boolean fileOnly) throws IOException {
+		super.close(fileOnly);
+		closeFile();
```
melissalinkert: The call to ```closeFile()``` needs to be in an ```if (libraryFound)``` block.  ```close(fileOnly)``` can be called on an uninitialized reader, which in this case means that an ```UnsatisfiedLinkError``` will be thrown when the .dlls are not present even if this reader isn't being used.

RichardMyers:
Thanks @melissalinkert  I've updated the sharing on the links so anyone can access them.  I also fixed close(fileOnly) - good catch.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,395 @@
...
+				ms.littleEndian = true;
+				ms.dimensionOrder = "XYCZT";
+				ms.indexed = false;
+				ms.falseColor = false;
+			}
+			setSeries(0);
+
+			// fill in meta data
+			MetadataStore store = makeFilterMetadata();
+			MetadataTools.populatePixels(store, this);
```
melissalinkert: This should be ```MetadataTools.populatePixels(store, this, true);```, since attributes on Plane are being populated later on.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,395 @@
...
+						if (magnification > 0) {
+							store.setObjectiveNominalMagnification(
+									magnification, 0, objectiveIndex);
+						}
+					}
+
+					// link Objective to Image
+					String objectiveID =
+							MetadataTools.createLSID("Objective", 0, objectiveIndex);
+					store.setObjectiveID(objectiveID, 0, objectiveIndex);
```
melissalinkert: ```setObjectiveID``` should be called before the other ```setObjective*``` methods.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,395 @@
...
+		try {
+			openFile(id);
+
+			// read basic meta data
+			int numCaptures = getNumCaptures();
+			int[] numPositions = new int[numCaptures];
+			int[] numTimepoints = new int[numCaptures];
+			int[] numZPlanes = new int[numCaptures];
+			int[] numChannels = new int[numCaptures];
+			int[] numImages = new int[numCaptures];
```
melissalinkert: ```numImages``` can be removed, as it's no longer used.

melissalinkert:
Aside from the few minor comments above, the only worry I have is that there seems to be a lot (~4000 lines) of debugging output during setId, of the form:

```
CMaskDataSharedPtr()
operator=()
Uncompress
CMaskDataSharedPtr(const CMaskDataSharedPtr& inCopy)
```

I assume that's coming from SlideBook6Reader.dll, but if there's any way to remove that or turn it off, that would be helpful.

Otherwise, this looks good - once the comments above are addressed, this should be fine to include for the CI builds.

RichardMyers:
I found the offending cout << messages filling up the debugging output log.  I removed them and updated the links to new dlls.  I also compiled x32 dlls but haven't had a chance to test them yet.
I also fixed the setObjective  call order so hopefully everything is in good shape for wider testing.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
melissalinkert: This line will still need to be updated before we can include the pull request - otherwise, all OME-XML generated via this reader will be invalid.

RichardMyers:
@melissalinkert thanks for the heads up; I've added that missing parameter to populatePixels now.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
RichardMyers: Thanks for the heads up; I've added the missing parameter.

components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
```diff
@@ -0,0 +1,388 @@
...
+		boolean isMatch = ((magicBytes2 & 0xff00) == SLD_MAGIC_BYTES_1_1 ||
+				(magicBytes2 & 0xff00) == SLD_MAGIC_BYTES_1_2) &&
+				(magicBytes1 == SLD_MAGIC_BYTES_1_0 ||
+				magicBytes1 == SLD_MAGIC_BYTES_2_0);
+
+		return isMatch;
+	}
+
+	/**
+	 * @see IFormatReader#isThisType(String, boolean)
```
sbesson: This `@see` reference is causing a warning in the Javadocs (see https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/861/warnings39Result/). You may to use

```
@see loci.formats.IFormatReader#isThisType(String, boolean)
```

as above

melissalinkert:
Thanks, @RichardMyers.  Merging this so that it's sure to be in 5.1.2.

RichardMyers:
Thanks you @melissalinkert for walking me through the process.  I'm sure once this starts to get used I will have more pull requests based on the feedback we get.

ehrenfeu:
Does this mean only Windows x64 will be able to use Bio-Formats for reading SlideBook6 files?

RichardMyers:
I have now created both sets of dlls so both x32 and x64 versions should
work fine.

ehrenfeu:
Thanks, @RichardMyers ! I'm actually less worried about x32 personally, but more about Mac and Linux - my question was rather if the behaviour of Bio-Formats will be different on these platforms...?

joshmoore:
On any platform that the shared library ([`System.loadLibrary("SlideBook6Reader")`](https://github.com/openmicroscopy/bioformats/blob/4d332dd5f71d7ce56b35e16967668eaad1dd4f2b/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java#L88)) can be successfully loaded, then the logic here will be used. Otherwise, Bio-Formats will fall back to using the previous reader ([`loci.formats.in.SlidebookReader       # sld`](https://github.com/openmicroscopy/bioformats/blob/4d332dd5f71d7ce56b35e16967668eaad1dd4f2b/components/formats-api/src/loci/formats/readers.txt#L39)).

ehrenfeu:
Thanks, Josh! So the `System.loadLibaray()` call will only succeed on Windows, as that's the only platform the DLLs are provided for, right?

joshmoore:
...will only succeed where you've correctly installed those libraries. (They don't ship with Bio-Formats itself)

RichardMyers:
That's right, we only have Windows DLLs built up now.  We have done some
work on an OS X version and so that is our next goal.  We don't think Linux
DLLs will be too different from OS X.

Thanks, Josh! So the System.loadLibaray() call will only succeed on
Windows, as that's the only platform the DLLs are provided for, right?

—
Reply to this email directly or view it on GitHub
<https://github.com/openmicroscopy/bioformats/pull/1755#issuecomment-106306014>
.

ehrenfeu:
Thanks all for the clarifications!

PR #1756, opened 2015-04-22
ImageJ: prevent NPE if one of the sub-channel types is null

melissalinkert:
See
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005289.html

I was able to reproduce the original exception by copying the ```ics/anke/``` dataset and replacing the ```ch``` in line 5 of the .ics file with ```p```.  With this change, the same dataset should open in ImageJ without an exception; the labels at the top of the image window should show that there are 4 phases.


pwalczysko:
Tried with Fiji. Without this PR, the error repeated as expected with ``anke/...ics``. 
Replaced all the files in jars/bio-formats and also the bio-formats_plugins.jar for the ones in https://ci.openmicroscopy.org/view/5.1/job/BIOFORMATS-5.1-merge-build/lastSuccessfulBuild/artifact/artifacts/
With this, the image opened without errors and showed the 4 Phases on the top of the image window and these were selectable.

All looks good here.

melissalinkert:
--no-rebase

PR #1757, opened 2015-04-23
Changes for Octave compatibility

carandraug:
These two commits are required for the code to work with GNU Octave.

The first commit only explicitly ends each of the function blocks (and indents them). This is not really required for Octave compatibility but I believe it to be good form, specially when the file has subfunctions (which bioformats does). It is best viewed locally with the `--ignore-all-space` option.

The second commit is the one that actually changes the code but I believe it to still work in Matlab. More details are on the commit message.

Note that even after this changes, the code will not run on the latest release of Octave (3.8.2). However, I have been working on the Octave side to fix the other incompatibilities (which could be treated as Octave bugs) so that bioformats can work on the upcoming Octave release (which will be 4.0.0 and is currently on rc3).

### DIFF
--- a/components/formats-gpl/matlab/bfCheckJavaMemory.m
+++ b/components/formats-gpl/matlab/bfCheckJavaMemory.m
@@ -33,7 +33,7 @@ function [] = bfCheckJavaMemory(varargin)
 % with this program; if not, write to the Free Software Foundation, Inc.,
 % 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
-runtime = java.lang.Runtime.getRuntime();
+runtime = javaMethod('getRuntime', 'java.lang.Runtime');
 maxMemory = runtime.maxMemory() / (1024 * 1024);
 
 ip = inputParser;
--- a/components/formats-gpl/matlab/bfCheckJavaPath.m
+++ b/components/formats-gpl/matlab/bfCheckJavaPath.m
@@ -68,7 +68,7 @@ if ~status && ip.Results.autoloadBioFormats,
     jarPath = getJarPath(bfJarFiles);
     assert(~isempty(jarPath), 'bf:jarNotFound',...
         'Cannot automatically locate a Bio-Formats JAR file');
-    
+
     % Add the Bio-Formats JAR file to dynamic Java class path
     javaaddpath(jarPath);
     status = true;
@@ -76,7 +76,11 @@ end
 
 if status
     % Read Bio-Formats version
-    version = char(loci.formats.FormatTools.VERSION);
+    if is_octave()
+        version = char(java_get('loci.formats.FormatTools', 'VERSION'));
+    else
+        version = char(loci.formats.FormatTools.VERSION);
+    end
 else
     version = '';
 end
--- a/components/formats-gpl/matlab/bfGetFileExtensions.m
+++ b/components/formats-gpl/matlab/bfGetFileExtensions.m
@@ -34,12 +34,20 @@ function fileExt = bfGetFileExtensions
 % 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 % Get all readers and create cell array with suffixes and names
-imageReader = loci.formats.ImageReader();
-readers = imageReader.getReaders();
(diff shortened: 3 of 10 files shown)


melissalinkert:
Thanks for this, @carandraug.

Could you please revert the whitespace changes from cadbfbe?  The original indenting conforms to standard Matlab formatting, so unless it's strictly necessary for Octave it would be best not to change it.

Otherwise, we're currently discussing how best to test this across both Matlab and Octave, and will follow up early next week.

carandraug:
No, the whitespace changes are not necessary, you could even write everything in a single line. I can revert them. What about the explicitly end of function blocks? Should I leave them? I never quite understood why Matlab programmers don't end function blocks. I was not aware of a coding guidelines for Matlab which is why I done this.

There are a few other things which I would like to do for this. I would like to make a proper octave package for bioformats. This would allow users to simply use `pkg load bioformats` on their code (and also use `pkg` to install the package). There is also [Octave Forge](http://octave.sourceforge.net/packages.php), which provides a kind of "repository" where this packages can be released. I am quite experienced with this and can make it happen easily. I am just not familiar with your build and release system.

sbesson:
@carandraug: regarding coding guidelines, the existing code follows the guidelines used by MATLAB core functions i.e. classic indentation and line wrapping (http://uk.mathworks.com/help/matlab/matlab_prog/improve-code-readability.html).

Regarding terminating functions blocks with `end`, this has been introduced recently to support nested functions but is not mandatory (http://uk.mathworks.com/help/matlab/matlab_prog/nested-functions.html). Functions terminated with `end` may not be backwards compatible below a certain version. I will double check what this minimum requirement is.

Also, related to Melissa's last comment, is there any authoritative framework for writing test suites in Octave?

carandraug:
The section of the manual on [tests](https://www.gnu.org/software/octave/doc/interpreter/Test-Functions.html#Test-Functions) would be the most authoritative but I think it's a bit hard to follow. The [wiki's page on the subject](http://wiki.octave.org/Tests) is easier to follow but I'm biased.

Basically, tests are written at the bottom of each function file in `%!test` blocks (they are ignore by Matlab as comments). If the test block causes an error then it failed, otherwise it passed. The function `assert ()` is typically used for throwing the error since it creates a nice message and allows to specify precision.

    octave:1> function x = foo (x)
    x += 4;
    endfunction
    octave:2> assert (foo (3), 5)
    error: ASSERT errors for:  assert (foo (3),5)

      Location  |  Observed  |  Expected  |  Reason
         ()           7            5         Abs err 2 exceeds tol 0
    octave:2> assert (foo (3), 5, 3)

`%!assert` one-liners are syntactic sugar for `%!test` with a single assert. To test for errors, use `%!error` blocks which will use a regular expression to compare against an error message.

[Here's an example](http://hg.savannah.gnu.org/hgweb/octave/file/1784bf79ed43/scripts/strings/strsplit.m#l218) from Octave's own sources that uses test, assert, and error blocks.

`%shared` and `%!function` are used to create variables and functions to be shared between all the tests in that file. To run all tests of an Octave package, run `runtests (path_to_the_directory)` .

carandraug:
I force pushed so that the whitespace changes won't mess up history. I also changed the whitespace of the new `is_octave` private function to match the format guidelines you are using.

sbesson:
Thanks for the last push @carandraug. Going through the code, the changes introduced by this PR look good functionally. On our side, we will use this opportunity to extend our MATLAB test coverage and make sure all Java calls modified here are tested on our CI infrastructure /cc @bramalingam.
Once the MATLAB tests have been extended and merged, we will include this PR in the CI build for a final review. I would expect this improvement to be included in the 5.1.2 release of Bio-Formats.

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-matlab/761/console is green with this PR included. Additionally, a quick smoke test of the various MATLAB functionalities using the Bio-Formats merge build does not return any unexpected behavior. This PR should be ready to merge and we should make sure we have a proper round of testing of Bio-Formats MATLAB prior to 5.1.2 release /cc @bramalingam.

sbesson:
Tested under MATLAB R2012a for backwards compatibility. Works as expected. Merging for upcoming 5.1.2.

PR #1759, opened 2015-04-27
travis: Switch from saucy to trusty

sbesson:
saucy has been removed from the mirrors and was no longer supported.

See https://github.com/openmicroscopy/ome-documentation/pull/1197

PR #1760, opened 2015-04-27
travis: Switch from saucy to trusty (rebased onto dev_5_0)

sbesson:
This is the same as gh-1759 but rebased onto dev_5_0.

----

saucy has been removed from the mirrors and was no longer supported.

See https://github.com/openmicroscopy/ome-documentation/pull/1197


sbesson:
--rebased-from #1759

PR #1763, opened 2015-04-27
JPEG-2000: fix redundant file parsing

melissalinkert:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005300.html.

To test, use the .jp2 file from ```jpeg2000/mario```.  With both commits, ```showinf -debug``` should indicate that 88 comments were found.  The same test with d742be2 reverted should show that 176 comments were found, but the output should otherwise be the same.


mtbc:
```diff
10c12
< Found 176 comments
---
> Found 88 comments
```
looks as hoped. And, the image still looks fine. :+1:

CI willing, good to merge.

melissalinkert:
--no-rebase

PR #1766, opened 2015-04-28
Zeiss LSM: fix lookup tables for 16-bit data

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12383 and https://trac.openmicroscopy.org/ome/ticket/12384.

To test, use ```test_images_good/zeiss-lsm/QA images/10052011_R1_GR1_B1_L49.lsm```.  Without this change, ```showinf``` should report that the images are indexed, but a lookup table should not be present - both images will appear gray/black.  With this change, the same test should show that the images are indexed and that a lookup table with 65536 entries is present.  The first image should be colored gray/white, and the second image should be colored red.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
@@ -165,7 +165,6 @@ public class ZeissLSMReader extends FormatReader {
   private double pixelSizeX, pixelSizeY, pixelSizeZ;
   private byte[][][] lut = null;
   private List<Double> timestamps;
-  private int validChannels;
 
   private String[] lsmFilenames;
   private List<IFDList> ifdsList;
@@ -195,6 +194,7 @@ public class ZeissLSMReader extends FormatReader {
 
   private Map<Integer, String> acquiredDate =
     new HashMap<Integer, String>();
+  private Color[] channelColor;
 
   // -- Constructor --
 
@@ -250,7 +250,6 @@ public class ZeissLSMReader extends FormatReader {
       pixelSizeX = pixelSizeY = pixelSizeZ = 0;
       lut = null;
       timestamps = null;
-      validChannels = 0;
       lsmFilenames = null;
       ifdsList = null;
       tiffParser = null;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
```diff
@@ -343,15 +343,16 @@ public int fileGroupOption(String id) throws FormatException, IOException {
...
     short[][] s = new short[3][65536];
-    for (int i=2; i>=3-validChannels; i--) {
-      for (int j=0; j<s[i].length; j++) {
-        s[i][j] = (short) j;
-      }
+    Color color = channelColor[prevChannel];
+    for (int j=0; j<s[0].length; j++) {
+      s[0][j] = (short) ((color.getRed() / 255.0) * j);
+      s[1][j] = (short) ((color.getGreen() / 255.0) * j);
+      s[2][j] = (short) ((color.getBlue() / 255.0) * j);
```
mtbc: Not sure why the `* j` makes sense but it seems to work!

mtbc:
```
Indexed = true (false color, 16-bit LUT: 3 x 65536)
```
looks pretty good. :+1:

In Insight this imports as a two-channel image, 
channels
I'm guessing that's expected, it all looks good image-wise.

components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
melissalinkert: The idea is to make a gradient, so that relative pixel intensities are preserved when the lookup table is applied.  Not multiplying by ```j``` would set all lookup table entries to be the same, which would result in a blank image.

melissalinkert:
Yes, that's the expected result.

mtbc:
:+1:

melissalinkert:
--no-rebase

components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
mtbc: Thank you, that makes sense.

PR #1767, opened 2015-04-28
Adding addition BF related mailing lists/groups

hflynn:
List of places people look for help with BF issues wasn't complete.

Will be staged on https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/about/index.html#help

Also, rearranged order so OME, ImageJ and Fiji items are grouped together.

### DIFF
--- a/docs/sphinx/about/index.txt
+++ b/docs/sphinx/about/index.txt
@@ -109,16 +109,23 @@ Other places where questions are commonly asked and/or bugs are reported
 include:
 
 -  `OME Trac <http://trac.openmicroscopy.org.uk/ome>`_
--  `Fiji Bugzilla (for ImageJ/Fiji
-   issues) <http://fiji.sc/cgi-bin/bugzilla/index.cgi>`_
 -  `ome-devel mailing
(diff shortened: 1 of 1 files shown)


hflynn:
The docs build doesn't like the Fiji google group anchor link but it does work.

joshmoore:
The list makes sense to me. I'll follow up on the card re: internal usage. The only addition that might be worth it if we have text around what we *expect/want* users/developers to do with these lists.

hflynn:
Elsewhere in the BF docs our sole recommendation to mail the ome-users list with problems/bugs. I'm not sure it's worth trying to suggest an alternative policy here given we already know a lot of users get referred to us from elsewhere without ever having read this documentation.

hflynn:
N.B. I'm going to add the Fiji google group link to https://github.com/openmicroscopy/sphinx-ignore-links/blob/master/broken_links.txt so it stops making the build yellow.

melissalinkert:
This is fine to merge then?

hflynn:
From my POV yes. @joshmoore ?

joshmoore:
Yup.

PR #1768, opened 2015-04-28
AIM: return blank planes for truncated files

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12785.  To test, verify that opening the file from QA 10907 does not result in an exception.

Note though that the images are likely to look weird/wrong.  The dimensional metadata matches the corresponding unprocessed data from QA 10913, so I think everything is fine there.  The file is much too small for the recorded dimensions, though, which leads me to believe that the file is somehow corrupted.  A request for further information (http://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7762) has gone unanswered so far.


mtbc:
Opens fine in `showinf` and OMERO. Indeed, looks weird, but no obvious exceptions. Thus, seems good to merge.

melissalinkert:
--no-rebase

PR #1771, opened 2015-04-29
C++: Various writer corrections

rleigh-dundee:
Major changes include:

- `setId` methods for both FormatReader and FormatWriter always canonicalise the path, so that comparisons always use the same path; previously there was potential for miscomparison where different paths for the same file could be used (relative vs. absolute vs. canonical absolute)
- BigTIFF checking factored out of MinimalTIFFWriter for use by OMETIFFWriter
- detail::FormatWriter methods using channel calculations corrected, and getSeriesCount() method added
- Addition of MetadataTools `validateModel` function, which can optionally correct some channel inconsistencies.

The model validation and correction currently deals solely with ambiguity between *SizeC*, channel element count and SamplesPerPixel (*SPP*) on channels.  The correction ensures that *SizeC* == sum(*SPP* on each channel).  It aims to handle the common cases (no channel elements, channel elements without *SPP*, *SizeC* mismatch), plus all the corner cases.  The following rules are used:

- If SamplesPerPixel is present on a channel, this is always treated as being more correct than *SizeC* (since it's more specific) and will never be changed
- If no channel elements are defined, then *SizeC* channel elements are created, each with *SPP*=1
- If all channels specify *SPP* but it mismatches with *SizeC*, *SizeC* is set to sum(*SPP* for each channel)
- If channel elements are present but do not specify the *SPP*, the number of channels in *SizeC* which are unaccounted for are divided between these channel elements.  This matches the correction the Java code does, but it also handles the corner case where there may be a mixture of channel elements with and without *SPP* set
- If correction is impossible, then we throw an exception; this is basically when we can't evenly divide the unallocated channels between channel elements with no defined *SPP*.

The testcases detail the input and output state for all the different possible corrections.

--------

Testing: unit tests should continue to pass.  Model validation and correction has a comprehensive set of unit tests in metadatatools.cpp and datafiles in `cpp/test/ome-bioformats/data/*.ome`; the main point for review here is whether the implemented behaviour is acceptable.  The other changes don't all include unit tests--they either already exist or will be tested in the OME-TIFF writer PR to follow.

### DIFF
--- a/cpp/lib/ome/bioformats/MetadataTools.cpp
+++ b/cpp/lib/ome/bioformats/MetadataTools.cpp
@@ -57,6 +57,7 @@
 #include <ome/common/xml/dom/NodeList.h>
 
 #include <ome/xml/meta/Convert.h>
+#include <ome/xml/meta/MetadataException.h>
 #include <ome/xml/meta/OMEXMLMetadataRoot.h>
 
 #include <ome/xml/model/Annotation.h>
@@ -80,6 +81,7 @@
 using boost::format;
 
 using ome::xml::meta::Metadata;
+using ome::xml::meta::MetadataException;
 using ome::xml::meta::MetadataStore;
 using ome::xml::meta::MetadataRoot;
 using ome::xml::meta::OMEXMLMetadata;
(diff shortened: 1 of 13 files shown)


melissalinkert:
This reads fine to me, and builds seem to be stable.

PR #1772, opened 2015-04-29
TIFF export: use long arithmetic for BigTIFF check

melissalinkert:
This should prevent ```BIG_TIFF_CUTOFF``` from being negative, which caused all TIFF/OME-TIFF files to be written as BigTIFF.

To test, pick any small file (e.g. ```test_images_good/al3d/Iron Plate.al3d```).  Verify that using ```bfconvert``` to convert the file to TIFF without this change produces a message to the effect that BigTIFF is being used; with this change, the same test should not produce a BigTIFF message.  Running ```file``` on the output files should also indicate BigTIFF without this change and plain TIFF with this change.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -80,7 +80,7 @@ public class TiffWriter extends FormatWriter {
    * This is approximately 3.9 GB instead of 4 GB,
    * to allow space for the IFDs.
    */
-  private static final long BIG_TIFF_CUTOFF = 1024 * 1024 * 3990;
+  private static final long BIG_TIFF_CUTOFF = (long) 1024 * 1024 * 3990;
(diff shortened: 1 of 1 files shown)


melissalinkert:
@joshmoore mentioned trying to catch this via Findbugs: http://findbugs.sourceforge.net/bugDescriptions.html#ICAST_INTEGER_MULTIPLY_CAST_TO_LONG

That error class unfortunately does not handle integer multiplication where the result is assigned to a long variable - it only handles the case where the cast to long occurs when returning a value or calling a method that takes a long parameter.  We can certainly consider changing the Findbugs reporting level to medium or low in a separate PR, but that wouldn't have caught this issue.

mtbc:
Works exactly as described in `bfconvert` and `file` output between BF-latest and BF-merge. Good to merge.

melissalinkert:
--no-rebase

PR #1773, opened 2015-04-29
Race condition during writing of TIFF leading to an exception

Strachu:
There is a race condition occurring during multithreaded writing of a (OME-)TIFF file resulting in an exception.

It occurs fairly rare but has occurred to me several times when writing huge OME-TIFF file. It ends with a stack trace similar to this one:

    Caused by: loci.formats.FormatException: X:184320 must be < 13206
    at loci.formats.FormatWriter.checkParams(FormatWriter.java:363)
    at loci.formats.out.TiffWriter.saveBytes(TiffWriter.java:211)
    at loci.formats.out.OMETiffWriter.saveBytes(OMETiffWriter.java:196)

This exception occurs when one thread enters FormatWriter.checkParams() while different one is in [TiffWriter.prepareToWriteImage()](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/out/TiffWriter.java#L368).

The cause of the problem is that linked block of code in prepareToWriteImage() changes temporarily the current series. While the above function is in synchronized block, checkParams() is not, leading to an exception when another thread changes current series in the middle of params checking.

I have fixed it by modifying the prepareToWriteImage() function so that it does not change current series anymore.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -315,12 +315,9 @@ public abstract class FormatWriter extends FormatHandler
 
     MetadataRetrieve r = getMetadataRetrieve();
     initialized = new boolean[r.getImageCount()][];
-    int oldSeries = series;
     for (int i=0; i<r.getImageCount(); i++) {
-      setSeries(i);
-      initialized[i] = new boolean[getPlaneCount()];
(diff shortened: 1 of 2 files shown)


melissalinkert:
This reads fine to me - thank you, @Strachu.  Adding the "include" label so that the CI builds pick this up tomorrow.

PR #1774, opened 2015-04-29
ImageJ: fix display of composite virtual stacks

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12267.

To test, use the .czi file from ```/ome/team/rleigh/lmaier-ij-composite.zip``` and run the following macro in ImageJ:

```
run("Bio-Formats Importer", "open=[/path/to/changing colors in composite mode.czi] color_mode=Composite view=Hyperstack stack_order=XYCZT use_virtual_stack");
```

Without this change, the resulting image window should look OK when initially displayed and each time the Z slider is changed, but the image will change when the C slider is changed (which is not correct for Composite mode).

With this change, adjusting the C slider should not cause the image to change.  This should only affect virtual stacks; it might be worth spot checking a few other multi-channel files with Composite mode and virtual stacks enabled.

In testing this, I noticed that the committed ImageJ jar was 1.45s, while the POM specifies 1.48s; c9bb5b4 updates the jar to 1.48s.


melissalinkert:
Discussed with @bramalingam; there was one more bug where the initial coloring was incorrect.  It took me a couple of tries, but this seemed to reliably reproduce the issue:

 * run the macro
 * adjust the C slider to anything other than C=1
 * close the image window
 * without restarting ImageJ, run the macro again and observe that the coloring is different from before

and 819a947 should resolve it.

bramalingam:
Everything looks good now. Please merge.

PR #1776, opened 2015-04-30
Add instructions to the MATLAB developers page for PermGenSpace errors

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005331.html

This PR should document the issue with `PermGenSpace` errors on MATLAB.

### DIFF
--- a/docs/sphinx/developers/matlab-dev.txt
+++ b/docs/sphinx/developers/matlab-dev.txt
@@ -26,9 +26,10 @@ using ``-Xmx512m`` in your :file:`java.opts` file. Calling::
 will also throw a warning if the runtime memory is lower than the recommended
 value.
 
-If errors of type ``java.lang.OutOfMemoryError: Java heap space`` are thrown
-while using Bio-Formats, you may try to increase the default values of
-``-XX:MaxPermSize`` and ``-XX:PermSize`` via the :file:`java.opts` file.
+If errors of type ``java.lang.OutOfMemoryError: PermGen space`` are thrown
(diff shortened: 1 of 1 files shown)


docs/sphinx/developers/matlab-dev.txt
```diff
@@ -19,7 +19,16 @@ MATLAB can be retrieved using::
...
+using ``-Xmx512m`` in your :file:`java.opts` file. Calling::
+
+  bfCheckJavaMemory()
+
+will also throw a warning if the runtime memory is lower than the recommended
+value.
+
+If errors of type ``java.lang.OutOfMemoryError: Java heap space`` are thrown
+while using Bio-Formats, you may try to increase the default values of
+``-XX:MaxPermSize`` and ``-XX:PermSize`` via the :file:`java.opts` file.
```
mtbc: Worth a note that this is for Java 6,7? (Is it?)

sbesson: I do not think it is necessary since the JVM bundled with MATLAB is only Java 6 or 7 so far (http://en.wikipedia.org/wiki/MATLAB#Release_history)

mtbc: Okay, makes sense.

rleigh-dundee: Is it possible to use a non-bundled system install of Java?  If that's possible, then using a current Java8 would be a possibility, which doesn't require a static size for PermGen space.

bramalingam:
This change looks good. Please merge.

jburel:
Looks good.

PR #1777, opened 2015-05-01
Memoizer relocation improvements

sbesson:
This PR fixes a bug with the memoization that can be reproduced following two scenarios:

- at the Bio-Formats level, using a memoizer-decorated reader, call `setId` on an image file in a directory and create an in-place memo file. Close the reader, rename the directory containing both the original file and the memo file. Using the same memoized reader, call `setId()` on the new file location.
- at the OMERO level, set up a server and import and image. Then stop the server, rename the `omero.data.dir` folder and modify the `omero.data.dir/BioFormatsCache` hierarchy to match the new naming scheme. After fixing the server configuration, restart the server and try to access the image.

In both cases, the `setId()` call should fail with `FileNotFoundException` leaving both the reader and the memo file in a stale state.

This PR fixes this bug by isolating the `reopenFile()` call and encapsulating it in a `try/catch` statement. `FileNotFoundException` will now effectively invalidate the memo file which will be safely recreated in the follow-up block. A Memoizer unit test has been added to cover the scenario described above.

Within the scope to this PR, we might want to be conservatice and catch all `Throwable` exceptions like we do in `loadMemo()` to prevent stale states. Are there edge cases where we do not want memo invalidation if `reopenFile()` fails?

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -655,6 +655,21 @@ public class Memoizer extends ReaderWrapper {
       loadedFromMemo = false;
       savedToMemo = false;
 
+      if (memo != null) {
+        // loadMemo has already called handleMetadataStore with non-null
+        try {
+          loadedFromMemo = true;
+          reader = memo;
+          reader.reopenFile();
+        } catch (FileNotFoundException e) {
+          LOGGER.warn("could not reopen file - deleting invalid memo file: {}", memoFile);
+          deleteQuietly(memoFile);
+          memo = null;
+          reader.close();
+          loadedFromMemo = false;
+        }
+      }
+
       if (memo == null) {
         OMEXMLService service = getService();
         super.setMetadataStore(service.createOMEXMLMetadata());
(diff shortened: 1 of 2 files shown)


joshmoore:
Initial (local) testing seems sane on the OMERO side. I'll try to find more ways to break it tomorrow.

joshmoore:
Finished all of my OMERO side testing. Couldn't find anything in particular. Comments:

 * we're still at `DEBUG` in OMERO, which we should likely change (**OMERO PR**);
 * the reopen failure may should even go lower than `WARN` at least for `FileNotFoundException` (**follow-up PR?**);
 * catching other exceptions may make sense (**follow-up PR?**) but that really depends a bit on the contract of `reopenFile`. Perhaps we can specify in `IFormatReader.reopenFile`?
 * there was some sense that it might be possible to protect against this in OMERO -- we already have `2015-05-05 09:44:03,650 WARN  [      o.s.blitz.repo.AbstractRepositoryI] (2-thread-1) Data directory moved: /tmp/beta updated to /tmp/gamma` and perhaps we could use a symlink to keep things functioning? Or would this be a case for Location mapping, @melissalinkert?

melissalinkert:
Using Location mapping would be possible if a single mapping from the old top-level directory name to the new top-level directory name correctly updated any paths that referenced the old directory.  I'm pretty sure that doesn't work right now, though.

joshmoore:
LGTM. We'll likely need to come back to the exception possibilities in `reopenFile` and see if we can't address some of the other points in OMERO.

PR #1778, opened 2015-05-01
Search for bioformats_package.jar in BF_JAR_DIR

tp81:
A small change to fix the fact that it was not checking the BF directory, but just the current dir for bioformats_package.jar. Hope this helps!


bramalingam:
Thank you. This change looks good. @melissalinkert : Please merge.

We will have to come up with a subsequent PR to fix the sequence of loading jars in bf.sh and probably bf.bat.

PR #1779, opened 2015-05-05
Fix broken anchor links

melissalinkert:
This should fix warnings noted by https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-docs/.  Should be sufficient to verify that the build is green, and that the four changed links still point to the corresponding named method.


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/843/ is indeed green now and all links work as expected - good to merge, thanks.

melissalinkert:
--no-rebase

PR #1780, opened 2015-05-05
Fix getPlanePosition* macro commands in ImageJ

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12865.

To test, use the macro added in f074113.  Without ccccf80, running the macro in ImageJ should result in an error - see http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-April/005339.html.

With ccccf80, the macro should run successfully and output the X, Y, and Z positions for each plane.  I was using the .nd2 files from test_images_good to test, but almost any file with populated positions should work.


bramalingam:
Everything works as expected. Please merge.

Issue #1782, opened 2015-05-06
Metadata skew in Bioformats for LightsheetZ1 CZI data


StephanPreibisch:
Hi,

I realized that different versions of Bioformats store the individual stack size for the views under different tags. I implemented a workaround here https://github.com/bigdataviewer/SPIM_Registration/commit/d10db40f989c20ec07b3d615f4a2965e20c573d2, but it seems that this is not right.

In the current Fiji there is Bioformats-5.1.1. and it stores the z-stack size metadata under the key:
SizeZ|View|V|Image|Information #1
SizeZ|View|V|Image|Information #2
SizeZ|View|V|Image|Information #3
...

In Eclipse I still use Bioformats-5.1.0 where the key is
Information|Image|V|View|SizeZ #1
Information|Image|V|View|SizeZ #2
Information|Image|V|View|SizeZ #3
...

Or is this change on purpose?

Thanks a lot, Stephan

melissalinkert:
Thanks for point this out, @StephanPreibisch.  This was not intentional; the proposed fix is in gh-1784.

melissalinkert:
Closing, as the corresponding fix has been tested and merged for 5.1.2.

PR #1723, opened 2015-04-10
LIF reader wavelength

jburel:
This PR fixes issue https://trac.openmicroscopy.org/ome/ticket/12808

To test:
 * Import ```leica-lif/niko/6-2-15MF2-Pos3-S1.lif```
 * Check that the excitation wavelength is correctly set.

PR #1784, opened 2015-05-06
Zeiss CZI: fix ordering of metadata tags

melissalinkert:
Switching from a Stack to a Deque reversed the order in which the tags
were built.  This fixes the tags by using the Deque's descending
iterator.

See gh-1782.

To test, use any .czi file.  Verify that the original metadata table entries changed between 5.1.0 and 5.1.1; with this PR included, the 5.1.0 behavior should be restored.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -33,6 +33,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.Iterator;
 import javax.xml.parsers.DocumentBuilder;
 
 import loci.common.ByteArrayHandle;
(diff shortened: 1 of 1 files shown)


mtbc:
I'll look over my previous PR in case I might have introduced any similar bugs.

rleigh-dundee:
I can confirm this restores the 5.1.0 behaviour.  Please merge.

mtbc:
Outwith this PR: It looks to me like #1730 may have introduced a similar issue in `LeicaHandler.startElement` and `LIFReader.populateOriginalMetadata` where both iterate over the `nameStack`. Do you want to review, or should I open a PR that does the same as this?

melissalinkert:
@mtbc: definitely feel free to fix those two methods, though you may wish to wait a bit and/or speak with @jburel to avoid conflicts with ongoing work for gh-1723.

mtbc:
Opened #1793 accordingly.

PR #1787, opened 2015-05-07
Zeiss CZI: fix assembly of overlapping tiles

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12843.

To test, import the file from ```zeiss-czi/sebastian/12843``` into OMERO; wait for the pyramid to be generated, then compare the resulting image with what is shown in Zeiss' Zen software.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -298,9 +298,6 @@ public class ZeissCZIReader extends FormatReader {
 
     Region image = new Region(x, y, w, h);
 
-    int currentX = 0;
-    int currentY = 0;
-
     int bpp = FormatTools.getBytesPerPixel(getPixelType());
     int pixel = getRGBChannelCount() * bpp;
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
Testing on trout; still generating pyramids after 6 hours.

melissalinkert:
Pushed one more commit to fix test failures noted in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/485/

mtbc:
We have a Travis failure.

melissalinkert:
Restarted.

rleigh-dundee:
Pyramids are now being generated and they do match what I see in Zen Blue.  Incidentally, do we actually need to generate pyramids or does the CZI already contain them?  Over a network drive, it loaded the zoomed-out view and smoothly zoomed in and out across the whole image, making it look like it already had the data to hand.

Looks good to merge if the jobs are all green now.

melissalinkert:
Using the pyramids may be possible in the future, but is outside the scope of this PR.

Issue #1785, opened 2015-05-06
DeltavisionReader trying to open a FEI Tiff file.


avandecreme:
Hello,

When running ImageInfo on a particular FEI Tiff file, it crashes in the DeltavisionReader class:
```
Exception in thread "main" java.lang.NegativeArraySizeException
	at loci.formats.in.DeltavisionReader.initPixels(DeltavisionReader.java:387)
	at loci.formats.in.DeltavisionReader.initFile(DeltavisionReader.java:267)
	at loci.formats.FormatReader.setId(FormatReader.java:1395)
	at loci.formats.ImageReader.setId(ImageReader.java:835)
	at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:651)
	at ImageInfo.testRead(ImageInfo.java:989)
	at ImageInfo.main(ImageInfo.java:1071)
```

It appears that the `isThisType(RandomAccessInputStream stream)` method returns true because DV_MAGIC_BYTES_1 is stored at offset 96.
Could this method be made more restrictive by checking other bytes as well?
For example in my case, checking that sizeX, sizeY, imageCount > 0 or that pixelType is between 0 and 6 would have detected that my file is not a Deltavision one.

melissalinkert:
Thank you for reporting this, @avandecreme.

Would you be willing to upload the FEI TIFF to https://www.openmicroscopy.org/qa2/qa/upload/?  That will help us to test the fix and ensure that future versions of Bio-Formats do not break this file again.

avandecreme:
I tried to upload a sample file. I am not sure whether it worked or not.
Here is the feedback link:  https://www.openmicroscopy.org/qa2/qa2/qa/feedback/11061/?token=ae2a7c48b3d4e2038f1f00541f929411

Note that I had to randomize most of the pixels because I don't think I am allowed to share the original image.

melissalinkert:
Thank you for uploading a sample file.  gh-1788 should resolve the problem; I would expect this to be included in the upcoming 5.1.2 release.

melissalinkert:
Closing, as the corresponding fix has been tested and merged for 5.1.2.

avandecreme:
Yes it works great. Thanks.

PR #1788, opened 2015-05-08
Deltavision: check for valid dimensions in isThisType

melissalinkert:
See gh-1785.

To test, check the file from QA 11061 with ```showinf```.  Without this change, a NegativeArraySizeException should be thrown in DeltavisionReader.  With this change, the format should be listed as ```FEI TIFF``` and no exception should be thrown.


mtbc:
The image opens fine without errors.
```
FEITiffReader initializing /home/mtbc/ome/apache_repo/11061/Tile_020-009-000_0.tif
```
It looks like random noise. Here's a section of it,
pr-1788
Perhaps that's expected though. Otherwise seems good to merge.

avandecreme:
Yes, I had to remove the actual data of the image except for the bits concerned.
Thanks for tackling this.

PR #1789, opened 2015-05-10
NDPI: fix tile offsets for files between 2 and 4 GB

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12839.

To test, import the file from ```hamamatsu/qi/``` into OMERO.  Scroll through the zoom levels for the first imported image, pan to the lower right corner, and verify that nothing looks obviously wrong; the tiles should load normally, and there should be no exceptions in the Blitz log.

We have a few other files that may or may not be affected by this change, so it's probably a good idea to import them as well:

```
hamamatsu/tony/2010-02-26 17.39.12.ndpi
hamamatsu/paul/2009-10-29 17.04.12.ndpi
hamamatsu/philippe/fluorescence/PHAS 01.ndpi
```

and also verify that viewing/zooming/panning works without missing tiles or Blitz exceptions.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPIReader.java
@@ -172,6 +172,11 @@ public class NDPIReader extends BaseTiffReader {
       try {
         service.close();
         long[] markers = ifd.getIFDLongArray(MARKER_TAG);
+        if (!use64Bit) {
+          for (int i=0; i<markers.length; i++) {
+            markers[i] = markers[i] & 0xffffffffL;
+          }
+        }
         if (markers != null) {
           service.setRestartMarkers(markers);
         }


mtbc:
There is something very odd going on here:

tissue40x-8b

mtbc:
tissue40x-8b

melissalinkert:
The difference between the thumbnail/bird's eye view and the full image is known, and outwith this PR.

mtbc:
Okay, I'll re-test today then, suitably limiting "obviously wrong",

mtbc:
No obvious exceptions, but this looks like a problem:
bottom-right
bottom-right

joshmoore:
@melissalinkert mentioned looking into this. Status to be double-checked tomorrow morning.

melissalinkert:
```green redZ.ndpis``` isn't one of the ones that would have been affected.  In both cases, I can duplicate the behavior of a small strip of gray at the bottom of the image for some resolutions.  For ```green redZ.ndpis```, the same thing happened in 5.1.1; ```tissue40x-8B.ndpi``` doesn't import without this PR, so also not a regression.

In this context, the gray pixels usually indicate that the corresponding 8x8 JPEG block(s) either contain no data or could not be fully decoded - it doesn't indicate that the stream of pixel data was truncated (as that would result in an exception or black pixels).  It's very unclear to me whether this is something we can or should solve - looking at the files in NDP.view2 (http://www.hamamatsu.com/eu/en/product/category/5002/5007/U12388-01/index.html#1328463611394) is inconclusive as both seem to have small artifacts at the bottom edge of the image.

I'd be inclined to merge this PR for 5.1.2, since it does allow ```tissue40x-8B.ndpi``` to import and doesn't introduce behavior that wasn't present in 5.1.1.

mtbc:
I didn't notice anything else obviously wrong; I am happy to bow to @melissalinkert's expert judgment after her own investigations and see this PR merged.

sbesson:
Merging for 5.1.2-rc1

PR #1793, opened 2015-05-12
fix iteration order for nameStack in Leica readers

mtbc:
https://github.com/openmicroscopy/bioformats/commit/7345a479d072953d6669c9d8cde140311d136afa introduced a possible bug wherein the natural iteration order for `nameStack` is reversed. This PR reverts that change, following the pattern of #1784.

--no-rebase


melissalinkert:
The keys do now match what was present in 5.1.0 :+1:

PR #1797, opened 2015-05-13
CellSens VSI: fix channel count when multiple pyramids exist

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12848.

To test, import the following .vsi files into OMERO:

```
cellsens/qa-11039/Image_01.vsi
cellsens/olivier/small-tiles/*.vsi
cellsens/olivier/14-09-2012/Image_07.vsi
```

Verify that each results in multiple pyramids, some of which will have a single channel and some of which will have multiple channels.  Open each pyramid in the full viewer and zoom/pan to verify that there is nothing obviously wrong with the image.

When reading through the code, please pay special attention to 76b37a2, as this is a very subtle change to how ```FormatReader.setCoreIndex(int)``` works.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1362,8 +1362,8 @@ public abstract class FormatReader extends FormatHandler
     if (no < 0 || no >= core.size()) {
       throw new IllegalArgumentException("Invalid series: " + no);
     }
-    coreIndex = no;
     series = coreIndexToSeries(no);
+    coreIndex = no;
   }
 
   // -- IFormatHandler API methods --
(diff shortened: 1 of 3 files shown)


rleigh-dundee:
Agreed with the setCoreIndex change, that is subtle but the change is correct.  Please could you also make the corresponding change to cpp/lib/ome/bioformats/detail/FormatReader.cpp to keep the implementations in sync?  Also note that the Java implementation is missing

```
        resolution = index - seriesToCoreIndex(series);
```

after the coreIndex and series are set, which can result in the current resolution being invalidated.  Not sure why it's missing.

Tested with the suggested files and all pyramids for all images look correct.  The only artifacts I saw were with qa-11039, but this looks like missing tiles which were deliberately not acquired:

cellsens-qa11039

melissalinkert:
Fixes pushed for both FormatReader implementations.

rleigh-codelibre:
Thanks, both changes look good.  Please merge once it's green.

PR #1798, opened 2015-05-15
ND2: better timestamp and plane position population

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12880.

To test, use each of the .nd2 files from QA 11064.  Compare the DeltaT, PositionX, PositionY, and PositionZ values from ```showinf -nopix -omexml``` with the Time, X, Y, and Z values in the ```Recorded Data``` tab of NIS Elements Viewer's ```Image Properties``` window.  Also verify that the Z, C, T, and series counts match the Z, C, T, and XY position counts in NIS Elements.

This does also add DeltaT and Position* automated tests, so that going forward we'll have an easier time of catching and fixing these sorts of problems.  I have not gone through all of the .nd2 samples and updated configuration files to include the correct DeltaT and Position* values (it's only been done for ```nd2/qa-11064```).  Follow-up Trello card is https://trello.com/c/ZTCDiSU9/181-deltat-position-configuration-updates, likely for 5.1.3.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Handler.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Handler.java
@@ -111,6 +111,7 @@ public class ND2Handler extends BaseHandler {
   private ArrayList<Boolean> validLoopState = new ArrayList<Boolean>();
 
   private boolean canAdjustDimensions = true;
+  private boolean firstTimeLoop = true;
 
   // -- Constructor --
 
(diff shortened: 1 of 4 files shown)


bramalingam:
Works as expected. Please Merge.

sbesson:
From the automated tests standpoint, new individual tests are now properly executed for each file as expected (and pass for the modified configuration files).

PR #1799, opened 2015-05-16
cpp: ome-xml: Correct intermittent incorrect Timestamp timezone offset parsing

rleigh-dundee:
- Add sentry (not for this bug, just for general correctness)
- Use a string rather than a character array for the TZ offset;
  the use of a character array led to inconsistent behaviour since
  it decays to a char* and hence requires NUL termination, which
  may or may not occur depending on whether the memory has been
  used before, which is unpredictable.

Note: whitespace-only diff: https://github.com/openmicroscopy/bioformats/pull/1799/files?w=1

--------

Testing:

I've used Ubuntu 14.04 to reproduce and test.  Do a build like normal, then run

```
while ./bf-test ./cpp/test/ome-xml/timestamp ; do : ; done
```

Without this PR, you should see the tests fail intermittently if you run this a few times.  With the PR, no failures should be apparent.

Note that you can test on other platforms, but I can't reproduce on e.g. current MacOS or FreeBSD--the compiler and libc obviously use different memory allocation patterns which don't cause the parsing code to trip up, so the above is recommended.

### DIFF
--- a/cpp/lib/ome/xml/model/primitives/Timestamp.h
+++ b/cpp/lib/ome/xml/model/primitives/Timestamp.h
@@ -131,84 +131,90 @@ namespace ome
         operator>> (std::basic_istream<charT,traits>& is,
                     Timestamp& timestamp)
         {
-          Timestamp::value_type value;
-
-          // Save locale.
-          std::locale savedlocale = is.getloc();
-
-          try
+          std::istream::sentry s(is);
+          if (s)
             {
-              boost::posix_time::time_input_facet *input_facet =
-                new boost::posix_time::time_input_facet();
-              input_facet->set_iso_extended_format();
-              std::locale iso8601_loc(std::locale::classic(), input_facet);
+              Timestamp::value_type value;
 
-              is.imbue(iso8601_loc);
-              is >> value;
+              // Save locale.
+              std::locale savedlocale = is.getloc();
 
-              if (is)
+              try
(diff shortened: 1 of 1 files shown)


mtbc:
Yeah, passes just fine without this PR on my jessie system, so probably do need Ubuntu 14.04. Interesting.

joshmoore:
Reproduced using:
```
time sh -c "while ./bf-test ./cpp/test/ome-xml/timestamp ; do : ; done"
```
I typically saw failures in under 2 seconds. With this PR, none in 10s of seconds. :thumbsup:

(NB: Interestingly, the following never errored:
```
13327 jamoore@blue:/opt/bioformats/build$ ./bf-test ./cpp/test/ome-xml/timestamp --gtest_repeat=10000
...
[----------] Global test environment tear-down
[==========] 64 tests from 1 test case ran. (11 ms total)
[  PASSED  ] 64 tests.
```)

rleigh-dundee:
@joshmoore Yes, it does appear to be the case that if it's going to work it will do so consistently, and if it fails it will do so consistently or alternatively passing/failing.  Likely some glibc/libstdc++ initialisation isn't completely deterministic and leaves dirty memory on the stack which we don't reinitialise.

And newer compilers have stack and memory protection enabled by default which likely sets a guard byte to prevent such problems with strings/char arrays.

PR #1801, opened 2015-05-18
Use native kHz units for DeltavisionReader read-out rate

sbesson:
This should use the native kHz unit included in the log file. Additionally this fixes a bug introduced in 5.1.0 with an incorrect conversion.


rleigh-dundee:
This looks fine and testing shows correct model units in the ome-xml.  Please merge.

melissalinkert:
Also checked in OMERO - import works, and the unit is correctly shown in Insight :+1:

PR #1802, opened 2015-05-19
C++: Document and correct use of qtwidgets namespace

rleigh-dundee:
- Document the namespace
- Make TexelProperties use the namespace (it was using the bioformats namespace incorrectly)

--------

Testing: None; check job is green.


mtbc:
Changes look fine, cpp builds are green. :+1:

PR #1803, opened 2015-05-20
Fix handling of Leica LIF files with GSD data

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12585 (GSD refers to http://www.leica-microsystems.com/products/super-resolution-microscopes/details/product/leica-sr-gsd-3d/)

To test, use the file in ```leica-lif/ian/```.  Import into OMERO or view in ImageJ, and compare with the what Leica's LAS AF Lite viewer shows.  The images should look the same; LAS AF will show several additional event list entries (i.e. non-image data), which won't be present in ImageJ/OMERO.

This shouldn't affect other files, and builds should remain green.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -1020,7 +1020,9 @@ public class LIFReader extends FormatReader {
       if (!grandparent.getNodeName().equals("ProcessingHistory")) {
         // image is being referenced from an event list
         imageNodes.add(image);
-        offsets.add(oldOffsets[nextOffset]);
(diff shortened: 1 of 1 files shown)


jburel:
Compare results with and w/o this PR
 * w/o 12 images imported in OMERO (Wrong)
 * w 8 images. Viewing the image in LAS AF. 8 images + 2 files (non images files)
 * Metadata reading: not much metadata in the file. The relevant ones are correctly read.
 * Code: due to the limited number of examples of this type, probably good to add some null check when checking grandparent.

melissalinkert:
Commit pushed to add some null checks.

jburel:
Thanks.
Ready to be merged

PR #1806, opened 2015-05-20
Java: Unify plane state handling

rleigh-dundee:
The current `FormatReader` interface and implementation contain some suboptimal aspects:
- `openBytes` takes a plane index, while `get*LookupTable` rely on the *previous* plane rather than taking a plane index as well
- The plane index must be cached by the reader; some readers do this, but the state can be easily invalidated by either calling `get*LookupTable` before `openBytes`, or changing the current series/resolution (most readers don't reset the cached plane index)

This PR corrects these defects by:
- Adding a plane index to `FormatReader` as for the coreIndex/series, and keeping it updated
- Adding `getPlane` and `setPlane` methods to `FormatReader` to expose the plane state to the user
- Converting the readers to use the cached plane state in place of each reader implementing this logic independently and potentially inconsistently
- Adding `get*LookupTable` methods which take a plane index explicitly (readers which rely on `openBytes` calling first are updated accordingly so that all `openBytes`/`get*LookupTable` method calls are idempotent)
- All use of `get*LookupTable` updated to use explicit plane indexes; this clearly documents exactly which plane is needed; this potentially unclear/ambigious in several places, where it could behave inconsistently depending upon which methods the user previously called
- Deprecating `getLookupTable` methods which don't use a plane index
- C++ interfaces also provide the same methods, and the sentry saves and restores the plane state along with the coreIndex/series

--breaking

Note that the unit tests don't currently pass--there's an indexing bounds failure testing LUTs; marking breaking.  I'm not sure why this is happening--I suspect I'm not doing something right in the `ReaderWrapper` and/or other wrapper classes in getting/setting/caching the plane state correctly, but not sure exactly what.  It might also be that I've exposed a latent bug which was previously not encountered.  Or it might be the unit tests themselves at fault.  Any thoughts here would be much appreciated.

--------

Testing: All tests should remain green.  Lookup table functionality should not have changed; the behaviour of the existing get*LookupTable functions should remain unchanged (this should not be a breaking change).

### DIFF
--- a/components/formats-api/src/loci/formats/IFormatReader.java
+++ b/components/formats-api/src/loci/formats/IFormatReader.java
@@ -175,7 +175,7 @@ public interface IFormatReader extends IFormatHandler, IMetadataConfigurable {
    * anything other than {@link FormatTools#INT8} or {@link FormatTools#UINT8},
    * this method will return null.
    *
-   * @deprecated Superseded by stateless get8BitLookupTable(int).
+   * @deprecated Superseded by stateless {@link #get8BitLookupTable(int)}.
    */
   byte[][] get8BitLookupTable() throws FormatException, IOException;
 
@@ -187,7 +187,7 @@ public interface IFormatReader extends IFormatHandler, IMetadataConfigurable {
    * anything other than {@link FormatTools#INT16} or {@link
    * FormatTools#UINT16}, this method will return null.
    *
-   * @deprecated Superseded by stateless get16BitLookupTable(int).
+   * @deprecated Superseded by stateless {@link #get16BitLookupTable(int)}.
    */
   short[][] get16BitLookupTable() throws FormatException, IOException;
 
--- a/components/formats-api/src/loci/formats/IFormatReader.java
+++ b/components/formats-api/src/loci/formats/IFormatReader.java
@@ -302,11 +302,21 @@ public interface IFormatReader extends IFormatHandler, IMetadataConfigurable {
   /** Gets the number of series in this file. */
   int getSeriesCount();
 
-  /** Activates the specified series. This also resets the resolution to 0. */
+  /** Activates the specified series. This also resets the resolution
+   * and current plane to 0.
+   */
   void setSeries(int no);
 
   /** Gets the currently active series. */
   int getSeries();
+
+  /** Activates the specified plane. This also resets the resolution
+   * to 0.
+   */
+  void setPlane(int no);
+
+  /** Gets the currently active plane. */
+  int getPlane();
 
   /** Specifies whether or not to normalize float data. */
   void setNormalized(boolean normalize);
--- a/components/formats-api/src/loci/formats/IFormatReader.java
+++ b/components/formats-api/src/loci/formats/IFormatReader.java
@@ -143,11 +143,39 @@ public interface IFormatReader extends IFormatHandler, IMetadataConfigurable {
 
   /**
    * Gets the 8-bit color lookup table associated with
+   * the most recently opened image plane.
+   * If {@link #isIndexed()} returns false, then this may return
+   * null. Also, if {@link #getPixelType()} returns anything other
+   * than {@link FormatTools#INT8} or {@link FormatTools#UINT8}, this
+   * method will return null.
+   *
+   * @param no the image index within the file.
+   * @return the 8-bit LUT.
+   */
+  byte[][] get8BitLookupTable(int no) throws FormatException, IOException;
+
+  /**
+   * Gets the 16-bit color lookup table associated with
+   * the most recently opened image plane.
+   * If {@link #isIndexed()} returns false, then this may return
+   * null. Also, if {@link #getPixelType()} returns anything other
+   * than {@link FormatTools#INT16} or {@link FormatTools#UINT16},
+   * this method will return null.
+   *
+   * @param no the image index within the file.
+   * @return the 16-bit LUT.
+   */
+  short[][] get16BitLookupTable(int no) throws FormatException, IOException;
+
+  /**
+   * Gets the 8-bit color lookup table associated with
    * the most recently opened image.
    * If no image planes have been opened, or if {@link #isIndexed()} returns
    * false, then this may return null. Also, if {@link #getPixelType()} returns
    * anything other than {@link FormatTools#INT8} or {@link FormatTools#UINT8},
    * this method will return null.
+   *
+   * @deprecated Superseded by stateless get8BitLookupTable(int).
    */
   byte[][] get8BitLookupTable() throws FormatException, IOException;
 
@@ -158,6 +186,8 @@ public interface IFormatReader extends IFormatHandler, IMetadataConfigurable {
    * false, then this may return null. Also, if {@link #getPixelType()} returns
    * anything other than {@link FormatTools#INT16} or {@link
    * FormatTools#UINT16}, this method will return null.
+   *
+   * @deprecated Superseded by stateless get16BitLookupTable(int).
    */
   short[][] get16BitLookupTable() throws FormatException, IOException;
 
--- a/components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
@@ -64,8 +64,6 @@ public class ImprovisionTiffReader extends BaseTiffReader {
   private String[] files;
   private MinimalTiffReader[] readers;
 
-  private int lastFile = 0;
-
   // -- Constructor --
 
   public ImprovisionTiffReader() {
@@ -100,7 +98,6 @@ public class ImprovisionTiffReader extends BaseTiffReader {
       }
       readers = null;
       files = null;
-      lastFile = 0;
     }
   }
 
@@ -111,28 +108,38 @@ public class ImprovisionTiffReader extends BaseTiffReader {
     return noPixels ? null : files;
   }
 
-  /* @see loci.formats.IFormatReader#get8BitLookupTable() */
+  /* @see loci.formats.IFormatReader#get8BitLookupTable(int) */
   @Override
-  public byte[][] get8BitLookupTable() throws FormatException, IOException {
+  public byte[][] get8BitLookupTable(int no) throws FormatException, IOException {
     FormatTools.assertId(currentId, true, 1);
-    if (readers == null || lastFile < 0 || lastFile >= readers.length ||
+
+    int[] zct = getZCTCoords(no);
+    int lastFile = FormatTools.getIndex("XYZCT", getSizeZ(), getEffectiveSizeC(),
+      getSizeT(), getImageCount(), zct[0], zct[1], zct[2]) % files.length;
+
+    if (readers == null || lastFile >= readers.length ||
       readers[lastFile] == null)
     {
-      return super.get8BitLookupTable();
+      return super.get8BitLookupTable(0);
     }
-    return readers[lastFile].get8BitLookupTable();
+    return readers[lastFile].get8BitLookupTable(0);
   }
 
-  /* @see loci.formats.IFormatReader#get16BitLookupTable() */
+  /* @see loci.formats.IFormatReader#get16BitLookupTable(int) */
   @Override
-  public short[][] get16BitLookupTable() throws FormatException, IOException {
+  public short[][] get16BitLookupTable(int no) throws FormatException, IOException {
     FormatTools.assertId(currentId, true, 1);
-    if (readers == null || lastFile < 0 || lastFile >= readers.length ||
+
+    int[] zct = getZCTCoords(no);
+    int lastFile = FormatTools.getIndex("XYZCT", getSizeZ(), getEffectiveSizeC(),
+      getSizeT(), getImageCount(), zct[0], zct[1], zct[2]) % files.length;
+
+    if (readers == null || lastFile >= readers.length ||
       readers[lastFile] == null)
     {
-      return super.get16BitLookupTable();
+      return super.get16BitLookupTable(0);
     }
-    return readers[lastFile].get16BitLookupTable();
+    return readers[lastFile].get16BitLookupTable(0);
   }
 
   /**
@@ -148,7 +155,6 @@ public class ImprovisionTiffReader extends BaseTiffReader {
     int file = FormatTools.getIndex("XYZCT", getSizeZ(), getEffectiveSizeC(),
       getSizeT(), getImageCount(), zct[0], zct[1], zct[2]) % files.length;
     int plane = no / files.length;
-    lastFile = file;
 
     return readers[file].openBytes(plane, buf, x, y, w, h);
   }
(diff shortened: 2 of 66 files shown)


melissalinkert:
The Java changes here really shouldn't go into 5.1.x - these will need to wait until a dev_5_1 branch is present before the PR can be included in merge builds.  If you want the C++ changes included in 5.1.x, then it's probably best to split this into two PRs.

rleigh-dundee:
OK, I can certainly split it.  Note that it was written explicitly not to break backward compatibility on the Java side, and only contains interface additions.

rleigh-dundee:
@melissalinkert Split into two PRs now.  If it's not mergeable into 5.1, that's fine.  If you had any idea why it's failing that would still be good to know if you have a chance to have a look.

mtbc:
Anything further to do here or waiting for `dev_5_1` branch?

rleigh-dundee:
@mtbc Waiting for dev_5_1 but there is a failing unit test which still needs fixing.

joshmoore:
Generally moving to `breaking` for 5.2-targeted PRs. Labels updated.

sbesson:
Heads-up `develop` has been split from `dev_5_1`. Some conflicts to be solved on this PR.

rleigh-dundee:
I have merged in develop to resolve the conflicts.  Note that there's still an issue with the unit tests for which I'm not sure about the cause, and would appreciate it if anyone had any suggestions.

melissalinkert:
The get*LookupTable changes all make sense.  I'm not completely clear on why ```setPlane(int)``` and ```int getPlane()``` are necessary for this, though, since the plane index is being passed into both get*LookupTable and openBytes now.

rleigh-dundee:
@melissalinkert This was done for backward compatibility with code relying upon the stateful behaviour of openBytes.  If we wanted to make this completely stateless then getPlane/setPlane and the cached state could certainly be removed entirely.  My thought was to do this as a second change after a period of deprecation of the old API, but if making this a breaking change is acceptable then it could be done in this PR.

rleigh-dundee:
I still get test failure.  Here on MacOS X:

```
    [junit] Testcase: testDatasetGroupFiles(loci.plugins.in.ImporterTest):      Caused an ERROR
    [junit] Invalid image index: 7/7
    [junit] java.lang.IllegalArgumentException: Invalid image index: 7/7
    [junit]     at loci.formats.FormatTools.getZCTCoords(FormatTools.java:578)
    [junit]     at loci.formats.FormatTools.getZCTCoords(FormatTools.java:497)
    [junit]     at loci.formats.FormatReader.getZCTCoords(FormatReader.java:1162)
    [junit]     at loci.formats.in.FakeReader.get8BitLookupTable(FakeReader.java:172)
    [junit]     at loci.formats.ImageReader.get8BitLookupTable(ImageReader.java:351)
    [junit]     at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
    [junit]     at loci.formats.FileStitcher.get8BitLookupTable(FileStitcher.java:379)
    [junit]     at loci.formats.ChannelFiller.get8BitLookupTable(ChannelFiller.java:116)
    [junit]     at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
    [junit]     at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
    [junit]     at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
    [junit]     at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
    [junit]     at loci.plugins.util.ImageProcessorReader.createColorModel(ImageProcessorReader.java:289)
    [junit]     at loci.plugins.util.ImageProcessorReader.openProcessors(ImageProcessorReader.java:199)
    [junit]     at loci.plugins.in.ImagePlusReader.readProcessors(ImagePlusReader.java:416)
    [junit]     at loci.plugins.in.ImagePlusReader.readPlanes(ImagePlusReader.java:380)
    [junit]     at loci.plugins.in.ImagePlusReader.readImage(ImagePlusReader.java:277)
    [junit]     at loci.plugins.in.ImagePlusReader.readImages(ImagePlusReader.java:238)
    [junit]     at loci.plugins.in.ImagePlusReader.readImages(ImagePlusReader.java:216)
    [junit]     at loci.plugins.in.ImagePlusReader.openImagePlus(ImagePlusReader.java:111)
    [junit]     at loci.plugins.BF.openImagePlus(BF.java:98)
    [junit]     at loci.plugins.in.ImporterTest.datasetGroupFilesTester(ImporterTest.java:1035)
    [junit]     at loci.plugins.in.ImporterTest.testDatasetGroupFiles(ImporterTest.java:2019)
```

melissalinkert:
I'd prefer not to be adding new methods just for backwards compatibility.  For this PR, I think it makes most sense to remove getPlane/setPlane, but leave the plane state caching from the readers (adding ```TODO``` or ```@deprecated``` as appropriate).  That can then be cleaned up in 5.3 when the deprecated methods are removed.

rleigh-dundee:
If I remove getPlane/setPlane from IFormatReader, is it OK to retain in FormatReader so the various reader implementations can cache the state using the semi-private calls here?

melissalinkert:
Please don't remove the ```get*LookupTable``` methods without deprecating first.  I'd be fine with having them just return ```get*LookupTable(0)``` (though as before, it would be nicer to just revert the removal of the appropriate reader-specific caching), but removing them outright seems like asking for trouble.

Note that there are also compile errors in the last commit, as detected by Travis.

sbesson:
Is there any planned work on this PR for 5.2.0 notably to address https://github.com/openmicroscopy/bioformats/pull/1806#issuecomment-157429718? Or should we close for now and reopen when we come back to this?

rleigh-dundee:
I've pushed a rebased copy of this branch to https://github.com/rleigh-dundee/bioformats/tree/plane-state-wip -- it's building but failing the same test as before.  The `loci.plugins.in.ImporterTest` test is failing due to the plane count being unexpectedly clipped.

```
testDatasetGroupFiles
"java.lang.IllegalArgumentException: Invalid plane: 7
	at loci.formats.FormatReader.setPlane(FormatReader.java:1005)
	at loci.formats.FormatTools.checkPlaneParameters(FormatTools.java:936)
	at loci.formats.in.FakeReader.openBytes(FakeReader.java:310)
	at loci.formats.FormatReader.openBytes(FormatReader.java:938)
	at loci.formats.FormatReader.openBytes(FormatReader.java:907)
	at loci.formats.in.FakeReader.get8BitLookupTable(FakeReader.java:287)
	at loci.formats.ImageReader.get8BitLookupTable(ImageReader.java:351)
	at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
	at loci.formats.FileStitcher.get8BitLookupTable(FileStitcher.java:379)
	at loci.formats.ChannelFiller.get8BitLookupTable(ChannelFiller.java:116)
	at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
	at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
	at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
	at loci.formats.ReaderWrapper.get8BitLookupTable(ReaderWrapper.java:263)
	at loci.plugins.util.ImageProcessorReader.createColorModel(ImageProcessorReader.java:290)
	at loci.plugins.util.ImageProcessorReader.openProcessors(ImageProcessorReader.java:199)
	at loci.plugins.in.ImagePlusReader.readProcessors(ImagePlusReader.java:416)
	at loci.plugins.in.ImagePlusReader.readPlanes(ImagePlusReader.java:380)
	at loci.plugins.in.ImagePlusReader.readImage(ImagePlusReader.java:277)
	at loci.plugins.in.ImagePlusReader.readImages(ImagePlusReader.java:238)
	at loci.plugins.in.ImagePlusReader.readImages(ImagePlusReader.java:216)
	at loci.plugins.in.ImagePlusReader.openImagePlus(ImagePlusReader.java:111)
	at loci.plugins.BF.openImagePlus(BF.java:98)
	at loci.plugins.in.ImporterTest.datasetGroupFilesTester(ImporterTest.java:1035)
	at loci.plugins.in.ImporterTest.testDatasetGroupFiles(ImporterTest.java:2019)
... Removed 23 stack frames
```

With a few debugging prints:

```
   [testng] IPR: GETTING PLANE 7 TP=105
   [testng] IPR: Z=7 T=5 C=3 E=3
   [testng] RR: GETTING PLANE 7 TP=105
   [testng] RR: Z=7 T=5 C=3 E=3
   [testng] RR: GETTING PLANE 7 TP=105
   [testng] RR: Z=7 T=5 C=3 E=3
   [testng] RR: GETTING PLANE 7 TP=105
   [testng] RR: Z=7 T=5 C=3 E=3
   [testng] RR: GETTING PLANE 7 TP=105
   [testng] RR: Z=7 T=5 C=3 E=3
   [testng] CF: GETTING PLANE 7 TP=105
   [testng] CF: Z=7 T=5 C=3 E=3
   [testng] FS: GETTING PLANE 7 TP=105
   [testng] FS: Z=7 T=5 C=3 E=3
   [testng] RR: GETTING PLANE 0 TP=7
   [testng] RR: Z=7 T=1 C=1 E=1
   [testng] ImR: GETTING PLANE 0 TP=7
   [testng] ImR: Z=7 T=1 C=1 E=1
   [testng] FR: OpenBytes plane=7
   [testng] FR: OpenBytes Z=7 T=1 C=1 E=1
   [testng] FR: OpenBytes FAIL
```

You can see the plane count drops from 105 to 7 with the last nested reader wrapper for some reason, which makes the plane number out of range.

This might be a latent bug I've tripped up on, or something triggered by the refactoring.  I've not yet pinned down the cause.

sbesson:
After discussion with @rleigh-dundee, the remaining errors above should be solved by converting the plane index using `getAdjustedIndex()` in `FileStitcher`.

rleigh-dundee:
Note the C++ parts of the original PR were already committed a long time ago.  Dropped from this patchset.

I *think* @melissalinkert's comments were addressed, but worth double-checking in case I missed or misunderstood anything since it's been nearly 13 months since this was originally opened.

components/formats-api/src/loci/formats/IFormatReader.java
```diff
@@ -143,11 +143,39 @@
...
+
+  /**
+   * Gets the 8-bit color lookup table associated with
    * the most recently opened image.
    * If no image planes have been opened, or if {@link #isIndexed()} returns
    * false, then this may return null. Also, if {@link #getPixelType()} returns
    * anything other than {@link FormatTools#INT8} or {@link FormatTools#UINT8},
    * this method will return null.
+   *
+   * @deprecated Superseded by stateless get8BitLookupTable(int).
```
mtbc: `@link` can be nice here

components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
```diff
@@ -111,28 +108,38 @@ public void close(boolean fileOnly) throws IOException {
...
+  /* @see loci.formats.IFormatReader#get16BitLookupTable(int) */
   @Override
-  public short[][] get16BitLookupTable() throws FormatException, IOException {
+  public short[][] get16BitLookupTable(int no) throws FormatException, IOException {
     FormatTools.assertId(currentId, true, 1);
-    if (readers == null || lastFile < 0 || lastFile >= readers.length ||
+
+    int[] zct = getZCTCoords(no);
+    int lastFile = FormatTools.getIndex("XYZCT", getSizeZ(), getEffectiveSizeC(),
+      getSizeT(), getImageCount(), zct[0], zct[1], zct[2]) % files.length;
```
mtbc: So this `lastFile` calculation will go okay even if readers is null?

rleigh-dundee: Good point, I'm not certain.  It can be moved below to be safe.

sbesson:
Note the current state of this PR breaks the automated tests for some files - see https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-test_images_good/462/consoleText

components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
rleigh-dundee: Actually, no, it' can't be moved since lastFile is used in both cases.  But I don't spot any problems--both files and readers are initialised in initStandardMetadata, and the LookupTable methods aren't valid to call until after this is completed.  I might be missing something subtle, or not so subtle, though.

mtbc: Thank you for looking!

rleigh-dundee: The logic here is in theory correct, but what's wrong is that it's being triggered by this:

```
    ms0.indexed = photo == PhotoInterp.RGB_PALETTE &&
      (get8BitLookupTable(0) != null || get16BitLookupTable(0) != null);
```

in `MinimalTiffReader.initFile`.  At this point in time, the reader is still not even minimally configured which means this will fail.  I've adjusted it to cope better, but I'm not sure that what initFile is doing is particularly safe since it requires hacks in every derived reader.  It could perhaps check for the presence of a colourmap directly.

rleigh-dundee:
Green in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-build/506/ and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-test_images_good/464/

dgault:
The code looks good to me and the recent commits appear to resolve the previously reported feedback.
Builds and tests are all green and my testing shows no changes in expected functionality.

PR looks good to me.

sbesson:
Note the latest state of the PR has been passing the full repository build - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/205/. Reading at the history of this PR:

- what is the next action in terms of the `setPlane/getPlane` API which was discussed in https://github.com/openmicroscopy/bioformats/pull/1806#issuecomment-157061703? 
- as I do not hink we have any coverage of the lookup table API in the automated tests, is there a reference set of sample files that can be used for testing the modified readers?

rleigh-dundee:
@sbesson Regarding getPlane/setPlane, my `plane-state2` branch has these removed from IFormatReader.  However, it's failing some ImageProcessorReader tests since the plane state is now not propagated through the reader wrappers.  If that's the way we want to go, that will need more investigation--still looking into it.

Regarding the LUT API, if it's not explicitly tested, it's certainly being used in a number of places in e.g. the `createColorModel()` method of ImageProcessorReader, so it's definitely being exercised to some degree, including the new stuff.

joshmoore:
> However, it's failing some ImageProcessorReader tests since the plane state is now not propagated through the reader wrappers.

As an aside, this is something I've struggled through as well (passing of options down the stack). If there's any potential for defining stricter semantics on what gets passed when, I have use cases to contribute.

sbesson:
As we passed the API freeze milestone for 5.2.0, closing this PR. While reinvestigating this unification effort in a future set of API changes, we might need to assess and fix the issues raised in the last few comments i.e. passing information through the stack. The flexible `MetadataOptions` key/value pairs introduced in the `metadata` branch might be a good candidate for this. If so we should prioritize it with the corresponding set of unit tests.

PR #1807, opened 2015-05-20
C++: Add plane state introspection and manipulation to reader and writer interfaces

rleigh-dundee:
- Add getPlane and setPlane to FormatReader and FormatWriter, to match getSeries and setSeries.
- FormatReader sentry saves and restores plane state in addition to series state.
- Current plane is set to zero on series change.

This is the C++ part of #1806 

--------

Testing: Jobs should remain green.  The only code currently exercising the plane state is the writers, and these were already using the methods (the change for FormatWriter is simply moving them up to the public interface to expose them)

### DIFF
--- a/cpp/lib/ome/bioformats/FormatReader.h
+++ b/cpp/lib/ome/bioformats/FormatReader.h
@@ -99,8 +99,10 @@ namespace ome
       private:
         /// Reader for which the state will be saved and restored.
         const FormatReader& reader;
-        /// Saved state.
+        /// Saved core index.
         dimension_size_type coreIndex;
+        /// Saved plane index.
+        dimension_size_type plane;
 
       public:
         /**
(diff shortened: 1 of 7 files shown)


jburel:
Changes look sensible.

joshmoore:
Excluding for 5.1.2.

joshmoore:
Re-including now post-5.1.2

cpp/lib/ome/bioformats/FormatReader.h
```diff
@@ -124,6 +127,8 @@ namespace ome
             {
               if (coreIndex != reader.getCoreIndex())
                 reader.setCoreIndex(coreIndex);
+              if (plane != reader.getPlane())
+                reader.setCoreIndex(plane);
```
melissalinkert: Should this be ```reader.setPlane(plane)```?

rleigh-dundee: Oops, yes, it should, thanks.  Fix pushed.

mtbc:
Looks good to me. :+1:

joshmoore:
As commented on gh-1806, this is not something likely to hit the Java API until 5.2. Working under the assumption that this is still the direction we want to go, though, I'm merging.

PR #1808, opened 2015-05-20
Add memoize functionality to ImageInfo

sbesson:
This should also ease the testing of Memoizer functionalities which can be invoked in the command-line tools via:

```
showinf -cache /path/to/file
showinf -cache-dir /tmp /path/to/file
```

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -258,7 +258,7 @@ public class ImageInfo {
         }
         else if (args[i].equals("-map")) map = args[++i];
         else if (args[i].equals("-format")) format = args[++i];
-        else if (args[i].equals("-cachedir")) {
+        else if (args[i].equals("-cache-dir")) {
             cache = true;
             cachedir = args[++i];
         }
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
This all reads fine and works for me.  The only minor nit is that "-memoize" could be shortened to "-memo" to reduce typing and make things a bit simpler.

sbesson:
Agreed on the terminology, I had been using the same convention as other wrappers which are exposed via a verbal form in the command-line tool (`-expand/-merge/-separate`) but happy to switch to `-memo` if this makes things simpler

rleigh-codelibre:
I don't mind either way really, I was really just thinking about making it short and simple rather than being a verb/noun.  (Whether memoise is a valid verb is another discussion... :)

joshmoore:
Thoughts on adding in support for a directory? Something like `-memo` and `-memo-dir` perhaps. This would facilitate, e.g., testing the memoizer file on the server.

joshmoore:
Re: naming: perhaps `--cache` would be simpler?

sbesson:
Will push a commit with `-cache/-cache-dir` for discussion.

sbesson:
Pushed.

rleigh-dundee:
All looks fine here.  The -cache and -cache-dir options all work just fine.

One minor nit: If the cache-dir does not exist, it does not get created:

```
skipping memo: directory not writeable - /tmp/c2
```

Would it be a bit friendlier for showinf to create the directory if it doesn't exist; the behaviour at the Memoizer level is likely correct though.

sbesson:
Thanks. Given the fact we are very close to Bio-Formats code freeze, is it okay to have this RFE (`mkdir`) implemented for 5.1.3?

rleigh-dundee:
@sbesson Definitely fine with me to for 5.1.3.  What's here is good to merge.

PR #1810, opened 2015-05-21
Reduce missing library log level

melissalinkert:
This should prevent exceptions from being logged at WARN when the SlideBook .dlls are missing.

To test, pick any .sld file from ```test_images_good/slidebook``` or ```from_skyking/slidebook```; verify that ```showinf -nopix``` shows a logged exception without this change, and no exception with this change.

PR #1811, opened 2015-05-21
Fix exceptions when parsing SimplePCI TIFF metadata

melissalinkert:
See QA 11088 and http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-May/005386.html.

To test, verify that the file from QA 11088 throws an exception without this change (as noted in the thread).  With this change, the file should open without any error via showinf/ImageJ.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SimplePCITiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SimplePCITiffReader.java
@@ -160,11 +160,14 @@ public class SimplePCITiffReader extends BaseTiffReader {
 
     IniList ini = parser.parseINI(new BufferedReader(new StringReader(data)));
 
-    String objective = ini.getTable(" MICROSCOPE ").get("Objective");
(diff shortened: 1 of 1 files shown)


mtbc:
Without this change there is an NPE from `SimplePCITiffReader.initStandardMetadata`. With this change I can see the image, and others, just fine. Good to merge, assuming that we may ignore,
```
3i SlideBook SBReadFile library not found. Please see http://www.openmicroscopy.org/site/support/bio-formats/formats/3i-slidebook6-sld.html for details.
java.lang.UnsatisfiedLinkError: no SlideBook6Reader in java.library.path
...
```

sbesson:
@mtbc: did you build from this branch or did you use the merge build? The message above should have been reduced to DEBUG level in https://github.com/openmicroscopy/bioformats/pull/1810.

mtbc:
I built with this branch merged into latest.

PR #1813, opened 2015-05-22
Command-line tools: 5.1.2 docs

sbesson:
Following https://github.com/openmicroscopy/bioformats/pull/1808, this PR:
- adds basic documentation for the `-cache`/`-cache-dir` options of `showinf`
- adds basic documentation fro the `-fast`/`-autoscale` options /cc @mtbc 

The last commit implements a proposal to migrate this page to using the `program/option` markup (similarly to what was done for the CLI). The benefit would be 1- to register these options in the generated index and 2- to allow cross-references across the documentation using :option:`showinf -nopix`. On the flip side, this does not fully matches the way this page was originally written. This proposal was implemented in an isolated commit and can be reverted /cc @hflynn

### DIFF
--- a/docs/sphinx/users/comlinetools/display.txt
+++ b/docs/sphinx/users/comlinetools/display.txt
@@ -96,3 +96,24 @@ Debugging output can also be enabled if more information is needed:
 ::
 
   showinf -debug /path/to/file
+
+Image display can be done as quickly as possible using::
+
+  showinf -fast /path/to/file
+
+Note that this command results in a loss of precision.
+
+In combination with the :option:`-fast` option, the display range can be
+adjusted to the minimum and maximum pixel values::
+
+  showinf -fast -autoscale /path/to/file
+
+After initialization, the reader can be cached under the same directory as the
+input file::
+
+  showinf -cache /path/to/file
+
+It is also possible to specify the base directory under which the reader
+should be cached::
+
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
The changes all look and read fine to me.  Two minor comments:

The `-fast` option is still vague from the user's perspective.  What does it mean to be fast?  Does it change what the user sees, or the rendering quality?  Why isn't it always fast by default--why would I want to use this option?

The `-autoscale` option requires `-fast`.  If this is the case, why is `-fast` not implicit?  Why does the user need to specify it by hand?

sbesson:
@rleigh-dundee: agreed. Looking back at the commit history, this was originally implemented in https://github.com/openmicroscopy/bioformats/commit/434cba76cbb3eb9962f44c40612a8e7953619941. I can file a follow-up card to revisit this in 5.1.3, and allow `-autoscale` to be used without the `-fast` option.

jburel:
Card on Formats board?

sbesson:
https://trello.com/c/1x2IXY5X/188-allow-autoscale-to-be-used-without-fast-in-the-cl-tools

rleigh-dundee:
Good to merge from me.

hflynn:
Late to the party but I agree the use of option markup is an improvement so we're all good!

PR #1814, opened 2015-05-22
cpp: common: Add units of measurement support

rleigh-dundee:
Base unit definition types, constants and quantity types for units of measurement support.  Note: this is not model units support; this is completely generic for use in any code, and will be used by the model units support in the future.

Uses Boost.Units to provide all the base unit system and unit definitions, from which we create a complete set of types for each unit prefix needed.  While these aren't strictly necessary for basic unit conversion, providing these allows the programmer to control conversions between units and unit prefixes so that no implicit conversions can occur.  That is to say, a metre_quantity and micrometre_quantity can both be assigned a value in micrometres upon construction, but the metre_quantity will force a conversion to metres while the micrometre_quantity does not.  The separate types can also be used by the model units code and individual readers/writers as needed for their requirements.

Minor issues:

- pound-force per square inch (psi) has not been defined in terms of lbf/in^2 but uses a simple conversion from pascal; this means you can't multiply by in^2 to get the lbf, which didn't seem too important
- mmHg conversion in Boost.Units uses a low precision 6dp constant which isn't as correct as it could be; see https://svn.boost.org/trac/boost/ticket/11335
- angle conversion constants also have lower precision than expected; https://svn.boost.org/trac/boost/ticket/6893

--------

Testing: covered by unit tests.  Run `./bf-test ./cpp/test/ome-common/units` by hand to see them run.

The unit tests:

- Instantiate every type defined in the headers; since this is templated code, verifying it builds is sufficient to check that the types are instantiable
- Default constructs every instantiated type
- Checks conversion to and from every SI prefixed type to the unprefixed form to ensure the static_rational powers are correct
- Checks conversion of every non-SI type to and from the default (SI) unit system types; testing every combination would be quite burdensome, so we test that we can convert each to and from a prefixed or unprefixed SI type
- Checks the unit name; if the name is invalid it won't compile and if incorrect it won't find the associated test data and will fail the test
- Checks the unit symbol and stream output of numerical values.  We output to 4 s.f. and compare with expected output; note accuracy is tested by the conversion code; this is a sanity check that we're using the correct unit type

--------

Future work:

- Raise test precision threshold after conversion constant precision is improved (longer term); This is fairly simple, so I'll probably push a few one-liners upstream to use more precise constants


joshmoore:
Excluded from 5.1.2

rleigh-dundee:
Minor portability fix: reduce symbol count per object to <65535 by splitting unit tests; also split headers for performance and to reduce clutter.

joshmoore:
Re-including now post-5.1.2

cpp/lib/ome/common/units/length.h
```diff
@@ -0,0 +1,604 @@
...
+      static std::string symbol() { return "px"; }
+    };
+
+    /// Unit information for reference frame.
+    template<> struct base_unit_info<ome::common::units::reference_frame_base_unit>
+    {
+      /// Unit name.
+      static std::string name()   { return "reference frame"; }
+      /// Unit symbol.
+      static std::string symbol() { return "r.f."; }
```
joshmoore: Where does this value get used?

cpp/lib/ome/common/units/pressure.h
```diff
@@ -0,0 +1,403 @@
...
+namespace boost
+{
+  namespace units
+  {
+
+    /// Unit information for pound-force per square inch (psi).
+    template<> struct base_unit_info<ome::common::units::psi_base_unit>
+    {
+      /// Unit name.
+      static std::string name()   { return "pound-force per square inch"; }
```
joshmoore: I'd think here again we'll need to make sure that they symbol(ic) representations between Java & C++ should match.

joshmoore:
Just going to give the unit tests a local try. Otherwise, primary concern is matching unit representation and symbol values with https://github.com/openmicroscopy/bioformats/blob/develop/components/ome-xml/src/ome/units/UNITS.java

joshmoore:
Otherwise the tests look sensible:
```
...
[----------] 3 tests from NonstandardTimeTest/UnitConv/5, where TypeParam = UnitConversion<boost::units::quantity<boost::units::unit<boost::units::list<boost::units::dim<boost::units::time_base_dimension, boost::units::static_rational<1l, 1l> >, boost::units::dimensionless_type>, boost::units::heterogeneous_system<boost::units::heterogeneous_system_impl<boost::units::list<boost::units::heterogeneous_system_dim<boost::units::si::second_base_unit, boost::units::static_rational<1l, 1l> >, boost::units::dimensionless_type>, boost::units::list<boost::units::dim<boost::units::time_base_dimension, boost::units::static_rational<1l, 1l> >, boost::units::dimensionless_type>, boost::units::list<boost::units::scale_list_dim<boost::units::scale<10l, boost::units::static_rational<-2l, 1l> > >, boost::units::dimensionless_type> > >, void>, double>, boost::units::quantity<boost::units::unit<boost::units::list<boost::units::dim<boost::units::time_base_dimension, boost::units::static_rational<1l, 1l> >, boost::units::dimensionless_type>, boost::units::heterogeneous_system<boost::units::heterogeneous_system_impl<boost::units::list<boost::units::heterogeneous_system_dim<boost::units::scaled_base_unit<boost::units::si::second_base_unit, boost::units::scale<86400l, boost::units::static_rational<1l, 1l> > >, boost::units::static_rational<1l, 1l> >, boost::units::dimensionless_type>, boost::units::list<boost::units::dim<boost::units::time_base_dimension, boost::units::static_rational<1l, 1l> >, boost::units::dimensionless_type>, boost::units::dimensionless_type> >, void>, double> >
[ RUN      ] NonstandardTimeTest/UnitConv/5.DefaultConstruct
[       OK ] NonstandardTimeTest/UnitConv/5.DefaultConstruct (0 ms)
[ RUN      ] NonstandardTimeTest/UnitConv/5.Conversion
Testing unit conversion from centisecond to day (1 tests)
[       OK ] NonstandardTimeTest/UnitConv/5.Conversion (0 ms)
[ RUN      ] NonstandardTimeTest/UnitConv/5.StreamOutput
Testing unit stream output from centisecond to day (1 tests)
[       OK ] NonstandardTimeTest/UnitConv/5.StreamOutput (0 ms)
[----------] 3 tests from NonstandardTimeTest/UnitConv/5 (0 ms total)

[----------] Global test environment tear-down
[==========] 1185 tests from 395 test cases ran. (432 ms total)
[  PASSED  ] 1185 tests.
```
with the one caveat that the templated names are gnarly.

cpp/lib/ome/common/units/length.h
rleigh-dundee: Stream output.

```
reference_frame_quantity q(4.0 * reference_frame);
std::cout << q << '\n';
```

Note that these are *not* the unit enumerations defined in the OME data model; they are abbreviations required by the underlying units library.  The actual model units support will match the model exactly.  This unit symbol can be changed if needed, though it's not indended to be user-visible except perhaps in debug logging.

cpp/lib/ome/common/units/pressure.h
rleigh-dundee: They will do, in the model units code.  But not at this level--this is the low-level unit conversion; the Java code doesn't do this.  This will mostly be hidden entirely unless you need to do explicit unit conversion; the following PR should make this a bit clearer; see my `xsdfu-cpp-2015` branch (incomplete work in progress https://github.com/rleigh-dundee/bioformats/tree/xsdfu-cpp-2015).

However, note I found some fundamental problems with the Java units code which weren't picked up in the code review at the time; I certainly don't want to blindly copy something which is defective.  If we could have a chat about this that would be great.

rleigh-dundee:
This is an outline of my thoughts so far looking through the Java code.  They may at this stage be incomplete.

----------

Java:

- `UnitsFoo` enumerations from xsd-fu
- Unit: `Unit<Quantity>`
- Quantity: `Quantity`; contains number, `Unit<Q>`
- Instance: by symbol, are `Unit<Q>` *even though they are quantities!*

Why (WHY?) is `Unit` specialised by `Quantity` when it is *not* a quantity and *not related to a quantity*.  A `Quantity` should be specialised by a `Unit`, not the other way around.  All the actual units, like `Length`, `Power` are `Quantities`.  **No**, they are actually `Units`...  This looks like a design flaw.

Also, in C++, the short symbol names are inevitably going to clash with preprocessor macros, either in the system headers, or the application headers of the application using our headers.  We already
saw this for one case, and it's likely there will be more in real use.

----------

C++, ome-common basic unit types (this PR):

- Unit: base_unit::unit_type
- Quantity: quantity<unit_type>
- Constant: base_unit::unit_type()
- Creation of a quantity with unit: number * Constant

Clear and clean separation of concerns, matching Boost.Units' design.  The only part of the existing Java implementation this provides is the conversion constants.  Everything else here is unique to C++ and not part of the model units classes provided by the Java implementation.

----------

C++, ome-xml model unit types (proposed design):

- `UnitsFoo` enumerations from xsd-fu
- Unit: `Unit<UnitFoo>` - specialised for a given dimension
- Quantity: `Quantity<Unit>`; contains value and unit enum value of the specified dimension, and is not capable of any unit conversions
- Creation of a quantity with unit: quantity constructor from value+enum (model serialisation) or direct from base quantity.

`Quantity<Unit>` constructor can allow direct construction from a base level `quantity<unit_type>`; the type will map directly to the corresponding enum.  Conversely, it can also be converted back to a base type either via a templated class method or templated free function.

`Quantity<Unit>` does not intrinsically do conversions; it's simply a container for the model value and associated unit.  Free functions or helper methods can do conversions via use of the base types.

The Java code is mixing up these separate concerns as well as getting some of them completely backward.

joshmoore:
After talking to @rleigh-dundee, I'm adding the current state of unit/quantity classes from OMERO. The Hibernate level will typically not be seen by users. `*` represents each of the unit types: Length, Time, etc.

Value | BF-java | BF-cpp | Hibernate | Blitz
------------ | ------------- | --------- | ------------ | -----------
Common library | ome.units.UNITS.java | ome::common | Genshi-generated | Genshi -generated
Discovery | ~ome.units.UNITS.java~ | TBD | ome.model.units.UNITS is an enum of enums. | N/A
Quantities | ome.units.quantity.* | TBD | ome.model.units.*.java | omero.model.*I.java (src/ and resources/omero/model/*.ice)
Base quantity interface | ome.units.quantity.Quantity | TBD | ome.model.units.Unit | N/A
Enums | ome.units.unit.Unit<*> | TBD | ome.model.units.Units* | omero.model.enums.Units* (Units.ice)
Base enum interface | ome.units.unit.Unit | TBD | ome.model.units.UnitEnum | N/A

That being said, we should likely migrate this discussion to cards.

The state of this PR is:
 * the fields which I pointed out are only part of the "common library" here, and as such we'll consider them "internal-only" for the moment, though we may want to standardize them across the columns in this table one day.
 * due to the various *skews* presented in this comment, we will likely want to start adapting the various implementations to have them more similar. This will likely require some breaking changes. This, in general, will simplify usage, we may consider specific changes even in the 5.1 line. Minimally, we will need to agree on the target so that @rleigh-dundee can work toward that in C++.

If that's all agreed, then this can go in. :smile:

cc: @sbesson @jburel @melissalinkert

rleigh-dundee:
See the "Unit defects" column on https://trello.com/b/Uoadzs9k/formats for a summary of known defects.

rleigh-dundee:
The last three commits are minor fixes--adding unit tests for some non-SI units and fixing the unit definitions.  They just got missed out.

joshmoore:
https://github.com/snoopycrimecop/bioformats/commit/43b3de7cff2f0eca3a117205fa7b00d7bf992108 successfully built by https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/805/

With the previous caveats, **LGTM**

PR #1818, opened 2015-05-26
C++: Minor writer improvements and writer example with documentation

rleigh-dundee:
- Minor cleanups (indent, error message detail increased)
- Make close on destruction reliable (readers and writers)
- Minor fix to existing writer example
- Add OME-TIFF writing example
- Add OME-TIFF writing documentation to tutorial

--------

Testing:

- Take a look at the tutorial https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/cpp/tutorial.html
- Run the example and check you get a 512×512 16-bit image (RGB, planar, 12 signif. bits/pixel):

```
% ./bf-test ./cpp/examples/metadata-formatwriter test.ome.tiff
Image count: 1
        Plane count: 1
PixelBuffer PixelType is uint16
VariantPixelBuffer PixelType is uint16
Wrote 786432 uint16 pixels

% tiffinfo test.ome.tiff 
TIFF Directory at offset 0x180008 (1572872)
  Image Width: 512 Image Length: 512
  Bits/Sample: 16
  Compression Scheme: None
  Photometric Interpretation: RGB color
  Samples/Pixel: 3
  Rows/Strip: 1
  Planar Configuration: separate image planes
  ImageDescription: <?xml version="1.0" encoding="UTF-8" standalone="no" ?><OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2013-06" UUID="urn:uuid:8fc8ecfc-3d06-4672-bfbb-82ecb87607c4" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2013-06 http://www.openmicroscopy.org/Schemas/OME/2013-06/ome.xsd"><!-- Warning: this comment is within an OME-XML metadata block, which contains crucial dimensional parameters and other important metadata. Please edit cautiously (if at all), and back up the original data before doing so. For more information, see the OME-TIFF web site: http://www.openmicroscopy.org/site/support/ome-model/ome-tiff/ --><Image ID="Image:0"><Pixels BigEndian="true" DimensionOrder="XYZTC" ID="Pixels:0" Interleaved="false" SignificantBits="12" SizeC="3" SizeT="1" SizeX="512" SizeY="512" SizeZ="1" Type="uint16"><Channel ID="Channel:0:0" SamplesPerPixel="3"/><TiffData FirstC="0" FirstT="0" FirstZ="0" IFD="0" PlaneCount="1"><UUID FileName="test.ome.tiff">urn:uuid:8fc8ecfc-3d06-4672-bfbb-82ecb87607c4</UUID></TiffData></Pixels></Image></OME>
  Software: OME Bio-Formats (C++) 5.1.1
```

- View the example with the Java showinf:

```
./tools/showinf -fast -autoscale /path/to/test.ome.tiff
Checking file format [OME-TIFF]
Initializing reader
OMETiffReader initializing /tmp/b/test.ome.tiff
Reading IFDs
Populating metadata
Initialization took 0.085s

Reading core metadata
filename = /tmp/b/test.ome.tiff
Series count = 1
Series #0 :
        Image count = 1
        RGB = true (3) 
        Interleaved = false
        Indexed = false (false color)
        Width = 512
        Height = 512
        SizeZ = 1
        SizeT = 1
        SizeC = 3 (effectively 1)
        Thumbnail size = 128 x 128
        Endianness = intel (little)
        Dimension order = XYZTC (certain)
        Pixel type = uint16
        Valid bits per pixel = 12
        Metadata complete = true
        Thumbnail series = false
        -----
        Plane #0 <=> Z 0, C 0, T 0
```

### DIFF
--- a/cpp/lib/ome/bioformats/tiff/IFD.cpp
+++ b/cpp/lib/ome/bioformats/tiff/IFD.cpp
@@ -42,6 +42,7 @@
 
 #include <ome/internal/config.h>
 
+#include <boost/format.hpp>
 #include <boost/thread.hpp>
 
 #include <ome/bioformats/PlaneRegion.h>
@@ -749,10 +750,18 @@ namespace ome
         va_start(ap, tag);
 
         if (!tag)
-          throw Exception("Error getting field: Tag is not valid");
+          {
+            boost::format fmt("Error getting field: Tag %1% is not valid");
+            fmt % tag;
+            throw Exception(fmt.str());
+          }
 
         if (!TIFFVGetField(tiffraw, tag, ap))
-          sentry.error("Error getting field: Tag was not found");
+          {
+            boost::format fmt("Error getting field: Tag %1% was not found");
+            fmt % tag;
+            sentry.error();
+          }
       }
 
       void
@@ -770,7 +779,11 @@ namespace ome
         va_start(ap, tag);
 
         if (!tag)
-          throw Exception("Error getting field: Tag is not valid");
+          {
+            boost::format fmt("Error getting field: Tag %1% is not valid");
+            fmt % tag;
+            throw Exception(fmt.str());
+          }
 
         if (!TIFFVSetField(tiffraw, tag, ap))
           sentry.error();
@@ -957,20 +970,20 @@ namespace ome
           }
         else
           {
-            SampleFormat fmt;
+            SampleFormat sampleformat;
             try
               {
-                getField(SAMPLEFORMAT).get(fmt);
+                getField(SAMPLEFORMAT).get(sampleformat);
               }
             catch(const Exception& e)
               {
                 // Default to unsigned integer.
-                fmt = UNSIGNED_INT;
+                sampleformat = UNSIGNED_INT;
               }
 
             uint16_t bits = getBitsPerSample();
 
-            switch(fmt)
+            switch(sampleformat)
               {
               case UNSIGNED_INT:
                 {
@@ -983,7 +996,11 @@ namespace ome
                   else if (bits == 32)
                     pt = PixelType::UINT32;
                   else
-                    throw Exception("Unsupported bit depth for unsigned integer pixel type");
+                    {
+                      boost::format fmt("Bit depth %1% unsupported for unsigned integer pixel type");
+                      fmt % bits;
+                      throw Exception(fmt.str());
+                    }
                 }
                 break;
               case SIGNED_INT:
@@ -995,7 +1012,11 @@ namespace ome
                   else if (bits == 32)
                 pt = PixelType::INT32;
                   else
-                    throw Exception("Unsupported bit depth for signed integer pixel type");
+                    {
+                      boost::format fmt("Bit depth %1% unsupported for signed integer pixel type");
+                      fmt % bits;
+                      throw Exception(fmt.str());
+                    }
                 }
                 break;
               case FLOAT:
@@ -1005,7 +1026,11 @@ namespace ome
                   else if (bits == 64)
                     pt = PixelType::DOUBLE;
                   else
-                    throw Exception("Unsupported bit depth for floating point pixel type");
+                    {
+                      boost::format fmt("Bit depth %1% unsupported for floating point pixel type");
+                      fmt % bits;
+                      throw Exception(fmt.str());
+                    }
                 }
                 break;
               case COMPLEX_FLOAT:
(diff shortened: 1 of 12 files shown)


joshmoore:
Re-including now post-5.1.2

cpp/lib/ome/bioformats/tiff/IFD.cpp
```diff
@@ -1057,7 +1090,11 @@ namespace ome
             break;
 
           default:
-            throw Exception("Unsupported OME data model PixelType");
+            {
+              boost::format fmt("Unsupported OME data model PixelType %1%");
```
mtbc: should there be a "by"?

docs/sphinx/developers/cpp/tutorial.txt
```diff
@@ -50,6 +50,11 @@ certain modifications to the data model may only be made via direct
 access to the model objects, otherwise the higher-level metadata store
 interface should be preferred.
 
+The header :file:`ome/bioformats/MetadataTools.h` provides several
```
mtbc: Some kind of `:source:` link won't work here?

docs/sphinx/developers/cpp/tutorial.txt
```diff
@@ -66,7 +71,26 @@ themselves are more of an implementation detail at present.
    :start-after: read-example-start
    :end-before: read-example-end
 
-Full example source: :download:`metadata-formatreader.cpp <examples/metadata-formatreader.cpp>`
+If implementing a reader, it is fairly typical to set the basic image
```
mtbc: Does this imply "recommended"?

docs/sphinx/developers/cpp/tutorial.txt
```diff
@@ -493,3 +525,59 @@ Full example source: :download:`metadata-formatreader.cpp <examples/metadata-for
...
+OME-TIFF (TIFF with OME-XML metadata).
+
+Using a writer involves these steps:
+
+#. Create a writer instance
+#. Set metadata store to use
+#. Set options to control writer behavior
+#. Call :cpp:func:`setId` to specify the image file to write
+#. Store pixel data for each plane of each image in the specified
+   dimension order
```
mtbc: should each item end with a period?

mtbc:
Functionally it seems fine and accords with the PR description. Quite pretty, really:
test-image

docs/sphinx/developers/cpp/tutorial.txt
rleigh-dundee: We do use it ourselves, but there is no obligation to do so.  While it is certainly convenient, a reader implementer is free to do their own thing.  However, this section isn't about that, it's about conversion between metadata forms, so I was trying to avoid going into specifics at this point in the tutorial; this is just an example of where it could be used.

docs/sphinx/developers/cpp/tutorial.txt
rleigh-dundee: Changed for doxygen links.

cpp/lib/ome/bioformats/tiff/IFD.cpp
rleigh-dundee: I don't think so here.  It's the same as most of the messages above: the TIFF IFD does not support a given PixelType; it's also never going to be seen since we support every pixel type in the data model (at present); it's here only to catch any future model changes.

docs/sphinx/developers/cpp/tutorial.txt
rleigh-dundee: I don't know; I thought we omitted them for bulleted and numbered lists.  @hflynn what's the policy here?
This is matching the list in the reader section, above, but I can change both if it's wrong.

cpp/lib/ome/bioformats/tiff/IFD.cpp
mtbc: There's a "by" in the previous very-similar message; maybe the difference makes sense though. I'm happy with whatever looks correct to you!

docs/sphinx/developers/cpp/tutorial.txt
mtbc: My understanding is that if they're sentences they get initial capital and full stop; if they're not they get neither.

rleigh-dundee: OK, fixed now.

hflynn: Yes, @mtbc is correct

cpp/lib/ome/bioformats/tiff/IFD.cpp
rleigh-dundee: Looked through a bit more carefully.  They are subtly different.  This one maps pixeltype to sampleformat/bit depth.  It's constrained to the pixeltype enum.  The other one is the reverse mapping and is not constrained; a TIFF could potentially have sampleformats and bit depth combinations we simply can't handle, which is why this is "unsupported by the data model", rather than "unsupported pixel type"--there isn't a pixel type we can map to.

mtbc: Aha, that makes sense, thank you.

mtbc:
Looks good to merge!

PR #1823, opened 2015-05-28
Update POM version numbers to 5.1.3-SNAPSHOT

melissalinkert:
This should be the first PR merged now that 5.1.2 is tagged.


joshmoore:
I assume we should get this in ASAP cc: @sbesson

PR #1825, opened 2015-06-01
specification: Correct psi documentation description to pound-force

rleigh-dundee:
Nothing to test; it's just a comment correction.  We might want to regenerate the published Oxygen documentation if anyone knows how to do that.


hflynn:
I think Andrew showed @mtbc how to generate those docs.

mtbc:
:+1: Fine with me, good to merge.

Anything I ever knew got noted at https://www.openmicroscopy.org/private/ome-internal/instructions/publish-schema.html#schema-documentation but we have oXygen just for Mac so I didn't get to try it myself, it's probably fairly easy though or I would have noted more about how. (I think there's a different way to do it with whatever the usual free tool is but @qidane explained that oXygen does a better job.)

PR #1827, opened 2015-06-03
octave: create minimum files required for an Octave package.

carandraug:
Still needs to be done (but I would need someone that knows ant and maven):

* copy the LICENSE.txt at the root of repository to root of octave package
* replace the tokens in DESCRIPTION for date and release
* copy the matlab/ directory and rename as inst/
* create a bioformats-VERSION.tar.gz out of this directory


sbesson:
For packaging, the easiest way is to extend the `ant/toplevel.xml` to add a `dist-octave` target similar to the [dist-matlab](https://github.com/openmicroscopy/bioformats/blob/v5.1.2/ant/toplevel.xml#L1008) one.

If the `tar.gz` is mandatory, it is possible to use the [Tar task](https://ant.apache.org/manual/Tasks/tar.html) instead of the Zip task. Each file/component can be added with a nested `tarfileset` element with the `prefix` key/value specifying the subdirectory.

For the release/data version logic, there are multiple tasks available, I would try out the [filterset task](https://ant.apache.org/manual/Types/filterset.html) as you may be able to combine it with the tar target. Note that to be able to use the version number in the target (for substitution and naming purposes), you will need to add the dependency `depends="release-version" to the target definition.

carandraug:
I couldn't make use of filterset with the Tar task. I also tried [replace](https://ant.apache.org/manual/Tasks/replace.html) but got the same error, that tar doesn't support nested filterset or replace elements. So I create a temporary file, call filterset on it, add it to the released tarball, and then delete it.

sbesson:
Thanks @carandraug, including for a first review via the CI builds.

carandraug:
Another option, would be to rename `DESCRIPTION` as `DESCRIPTION.in`, and have ant copy it to `DESCRIPTION` and then replace the date and version tokens in build dir before packaging. This has the advantage of making it obvious that the file in the repository still needs to be processed, and the `.in` extension will be immediately familiar to anyone used to autotools.

It would need to add an extra line to gitignore though.

sbesson:
The current option is certainly fine for now but no veto to using the  `DESCRIPTION.in` logic you described above. With regard to naming, the current target will create a `bioformats-$RELEASE.tar.gz` which is confusing since `bioformats-$RELEASE.zip` relates to the source code zip. Can we modify the bundle name to minimally include Octave in it?

carandraug:
Yes, the tarball can be renamed without issues. But should the version number be stripped completely from the tarball filename? Or should I just prepend the word octave ?

Also, do you have any `NEWS` or `Changelog` file that could be copied into the tarball?

sbesson:
Sorry for the delay. The versioning in the tarball name is an interesting question as we currently have a mixture of both (the source zip is versioned while the command-line tools and Bio-Formats bundle are not versioned in the name). Deferring to @melissalinkert on this, although my feeling it that embedding the version number does not hurt.

As for a changelog file, the closest we have is the whats-new.txt file maintained under https://github.com/openmicroscopy/bioformats/blob/develop/docs/sphinx/about/whats-new.txt.

carandraug:
I have added the "whats-new.txt" file as the NEWS file for the octave package. I will await for more feedback about the filename for the tarball.

melissalinkert:
bioformats-octave-${RELEASE}.tar.gz is fine with me - no need to remove the version number as far as I'm concerned.

carandraug:
I have changed the tarball filename as requested.

sbesson:
Thanks @carandraug. Will have a last round of review tomorrow.

sbesson:
Looks good. Merging.

PR #1830, opened 2015-06-10
cmake: Add toplevel SuperBuild two-phase processing

rleigh-dundee:
Add initial cmake "superbuild" infrastructure.

- Will download and build zlib and libpng as a proof of concept (more will follow); this is to verify that the inter-module dependencies and build logic are functional
- Not enabled by default; existing builds will be unaffected
- When enabled, all builds are relocatable; it's not possible to build with a fixed install prefix.  This is likely what is always wanted on Windows.  This is due to the complications of building with a fixed prefix and then linking against the staged libraries and its interactions with rpath
- Currently hard-coded to use `make` with `DESTDIR` staging, so won't work on Windows yet; I'll get the superbuild setup fully functional on Unix and then tweak it to work on Windows once all the library support is in place (this avoids the additional complexity of battling with superbuild setup and Windows at the same time)

When enabled, the superbuild will add `zlib`, `png` and `bioformats` as subsidiary "external" projects, with the toplevel project being `bioformats-superbuild`.  This means the main build will recurse into these subprojects, which depend upon each other (you can still parallel-build them all).  Our code provides both the top-level `bioformats-superbuild` project (`cpp/cmake/BioFormatsSuperBuild.cmake`) and the subsidiary `bioformats` project (`CMakeLists.txt`); the `bioformats` project is the only project used if the superbuild feature is disabled.

--------

Testing:

- Build by adding `-Dbioformats-superbuild:BOOL=ON` to the cmake command-line
- You can also set `-DBIOFORMATS_EP_SOURCE_CACHE:PATH=/path/to/cache` to avoid repeatedly re-downloaded sources each time
- You can configure an install prefix with `-DCMAKE_INSTALL_PREFIX=/prefix`, and this will be used by `make install` with or without `DESTDIR` set; but it won't be used for the subsidiary builds
- After running `make install` you should see that the `include` and `lib` directories now contain the zlib and png headers and libraries in addition to the bioformats headers and libraries.

You can verify that the superbuild is working by:

- zlib and libpng will be downloaded, built and installed when running `make`
- cmake output in the subsidiary bioformats build will find zlib and libpng in the superbuild staging directory:

```
-- Found ZLIB: /tmp/b/superbuild-install/lib/x86_64-linux-gnu/libz.so (found version "1.2.8") 
-- Found PNG: /tmp/b/superbuild-install/lib/x86_64-linux-gnu/libpng.so (found version "1.6.17") 
```

- bioformats libraries and programs will be linked against the libraries in the superbuild staging directory.  On Linux:

```
% ldd bioformats-build/cpp/test/ome-bioformats/tiff | egrep "lib(png|z)"
        libpng16.so.16 => /tmp/b/superbuild-install/lib/x86_64-linux-gnu/libpng16.so.16 (0x00007f43a39ab000)
        libz.so.1 => /tmp/b/superbuild-install/lib/x86_64-linux-gnu/libz.so.1 (0x00007f439fe99000)
```

- If you build with `-Dbioformats-superbuild:BOOL=OFF` (or omitted, since it's the default), you'll get a normal build (the superbuild will be ignored)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,114 +53,131 @@ if (POLICY CMP0054)
   cmake_policy(SET CMP0054 NEW)
 endif(POLICY CMP0054)
 
-project(bioformats)
-include(cpp/cmake/Version.cmake)
-
-if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
-  message(FATAL_ERROR "In-tree builds are not supported; please run cmake from a separate build directory.")
-endif("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
-
+# For MSVC
 enable_language(CXX)
 
-list(APPEND CMAKE_MODULE_PATH
-     "${PROJECT_SOURCE_DIR}/cpp/cmake")
-
-include(CheckIncludeFileCXX)
-include(CheckCXXCompilerFlag)
-include(CheckCXXSourceCompiles)
-
-include(GNUInstallDirs)
-
-if (NOT DEFINED OME_BIOFORMATS_INSTALL_DATADIR)
-  set(OME_BIOFORMATS_INSTALL_DATADIR "${CMAKE_INSTALL_DATADIR}/ome/bioformats"
-      CACHE PATH "Bio-Formats-specific datadir")
+set(sb_default OFF)
+if(MSVC)
+  set(sb_default ON)
 endif()
+option(bioformats-superbuild "Build Bio-Formats and the projects it depends upon using a Super-Build setup" ${sb_default})
+mark_as_advanced(bioformats-superbuild)
+unset(sb_default)
 
-# Use standard path if using a prefix.
-if(CMAKE_INSTALL_PREFIX)
-  if (NOT DEFINED OME_BIOFORMATS_INSTALL_LIBEXECDIR)
-    set(OME_BIOFORMATS_INSTALL_LIBEXECDIR "${CMAKE_INSTALL_LIBEXECDIR}/ome/bioformats"
-        CACHE PATH "Bio-Formats-specific libexecdir")
-  endif()
-  if (NOT DEFINED OME_BIOFORMATS_INSTALL_ICONDIR)
-    set(OME_BIOFORMATS_INSTALL_ICONDIR "${CMAKE_INSTALL_DATADIR}/icons/hicolor/scalable"
-        CACHE PATH "Icon directory")
-  endif()
+include("${CMAKE_CURRENT_LIST_DIR}/cpp/cmake/BioFormatsCommon.cmake")
+
+if(bioformats-superbuild)
+  include("${CMAKE_CURRENT_LIST_DIR}/cpp/cmake/BioFormatsSuperBuild.cmake")
+  return()
 else()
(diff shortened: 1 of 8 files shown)


rleigh-dundee:
Note there was a file missing from the original commit which I've just added.

joshmoore:
In the process of testing via...
 * Installed visual studio express 
 * Installed choco (https://github.com/chocolatey/choco/wiki/Installation)
 * choco install git
 * choco install cmake
 * git clone bioformats && cd bioformats && mkdir cppbuild && mkdir cache
 * `c:\Program Files\CMake\bin\cmake.exe -D... ..`

joshmoore:
Haven't made it very far yet and I'll keep playing, but a bit of feedback:

 * in the Visual Studio shell, I'm currently getting this error:

screen shot 2015-06-11 at 16 37 20

 * in a regular shell:

screen shot 2015-06-11 at 16 38 10

rleigh-dundee:
@joshmoore Why are you testing this on Windows?  I stated in the PR description that this won't work on Windows at this stage, and that it will be made to work on Unix initially, then made to work on Windows once we know it works fully in an environment we already know works.  This is just adding the superbuild infrastructure and a couple of packages to demonstrate that it works; it needs additional packages adding to complete the collection (which I'm working on), and then it'll need tweaking to build on Windows.  I would recommend testing this on Ubuntu for the moment.

If we made it work on Windows right now, we'd also need to battle with fixing up all our cmake logic to work with MSVC, fixing all our code to work with MSVC, making all the tests pass, and the complication of setting up the superbuild setup on a platform we've never ever used before with this project.  The approach here is to do just one thing at once rather than four; we'll get to that point quite soon I hope, but in separate steps we can verify work correctly along the way.

joshmoore:
Sorry, I misread (& too quickly) your comment thinking that it was the DESTDIR logic which wasn't yet supported on Windows. Nevertheless, I have a VM set up for testing if this does becomes Windows compatible in < 90 days. :smile:

joshmoore:
OSX build certainly passes:
```
ioformats/cppbuild$ cmake .. -Dbioformats-superbuild:BOOL=ON -DBIOFORMATS_EP_SOURCE_CACHE:PATH=/tmp/cppcache/
...
bioformats/cppbuild$ du
475M    ./bioformats-build
 42M    ./bioformats-install
 16K    ./bioformats-prefix
304K    ./CMakeFiles
2.4M    ./png-build
 32K    ./png-prefix
5.9M    ./png-source
1.7M    ./superbuild-install
1.1M    ./zlib-build
 32K    ./zlib-prefix
3.0M    ./zlib-source
532M    .
bioformats/cppbuild$ du /tmp/cppcache/
1.3M    /tmp/cppcache/
```

rleigh-dundee:
The DESTDIR logic will certainly need factoring out into separate Unix and Windows commands once I've figured out how to do the equivalent staging install on Windows.

rleigh-dundee:
Note: Green in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/ (tests regular build only).  I'll switch the gretzky35 build to use the superbuild logic once it builds boost (since gretzky35 is the build using the custom boost build; this will get us a redistributable build).

joshmoore:
Thanks, @rleigh-dundee. @melissalinkert : merge at your discretion.

PR #1831, opened 2015-06-10
bfconvert: mention BigTiff flag and extension

sbesson:
See https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7830

This PR moves to using the `program/option` markup across the command-line tools pages and also documents the usage of the `-bigtiff` flag and/or the BigTiff extension in `bfconvert`.

### DIFF
--- a/docs/sphinx/users/comlinetools/conversion.txt
+++ b/docs/sphinx/users/comlinetools/conversion.txt
@@ -118,7 +118,9 @@ name pattern, then the other must be included too.  The only exception is if
 
       bfconvert -bigtiff /path/to/input output.ome.tiff
 
-    Note that this option is not necessary if a BigTiff extension is used for
-    the output file, e.g.::
+    .. versionadded:: 5.1.2
 
-      bfconvert -bigtiff /path/to/input output.ome.btf
+      The :option:`-bigtiff` option is not necessary if a BigTiff extension is
+      used for the output file, e.g.::
+
+        bfconvert /path/to/input output.ome.btf
--- a/docs/sphinx/users/comlinetools/conversion.txt
+++ b/docs/sphinx/users/comlinetools/conversion.txt
@@ -20,60 +20,48 @@ The output file format is determined by the extension of the output file, e.g.
 .. option:: -series SERIES
 
     All images in the input file are converted by default.  To convert only one
-    series:
-
-    ::
+    series::
 
       bfconvert -series 0 /path/to/input output-first-series.tiff
 
 .. option:: -timepoint TIMEPOINT
 
-    To convert only one timepoint:
-
-    ::
+    To convert only one timepoint::
 
       bfconvert -timepoint 0 /path/to/input output-first-timepoint.tiff
 
 .. option:: -channel CHANNEL
 
-    To convert only one channel:
-
-    ::
+    To convert only one channel::
 
       bfconvert -channel 0 /path/to/input output-first-channel.tiff
 
 .. option:: -z Z
 
-    To convert only one Z section:
-
-    ::
+    To convert only one Z section::
 
       bfconvert -z 0 /path/to/input output-first-z.tiff
 
 .. option:: -range START END
 
-    To convert images between certain indices (inclusive):
-
-    ::
+    To convert images between certain indices (inclusive)::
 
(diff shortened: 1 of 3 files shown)


docs/sphinx/users/comlinetools/conversion.txt
```diff
@@ -15,61 +17,60 @@ To convert a file to single output file (e.g. TIFF):
...
 
-  bfconvert -tilex 512 -tiley 512 /path/to/input output-512x512-tiles.tiff
+      bfconvert -tilex 512 -tiley 512 /path/to/input output-512x512-tiles.tiff
 
-``-tilex`` is the width in pixel of each tile; ``-tiley`` is the height in
-pixels of each tile.  The last row and column of tiles may be slightly smaller
-if the image width and height are not multiples of the specified tile width
-and height.  Note that specifying ``-tilex`` and ``-tiley`` will cause tiles
-to be written even if the image is smaller than 4096x4096.
+    :option:`-tilex` is the width in pixel of each tile; :option:`-tiley` is
```
mtbc: pixels

docs/sphinx/users/comlinetools/conversion.txt
```diff
@@ -15,61 +17,60 @@ To convert a file to single output file (e.g. TIFF):
...
-::
+    To convert images between certain indices (inclusive)::
 
-  bfconvert -range 0 2 /path/to/input output-first-3-images.tiff
+      bfconvert -range 0 2 /path/to/input output-first-3-images.tiff
 
-All images larger than 4096x4096 will be saved as a set of tiles if the output
-format supports doing so.  The default tile size is determined by the input
-format, and can be overridden like this:
+.. option:: -tilex TILEX, -tiley TILEY
```
mtbc: In Firefox when I click a `-tiley` and come back to this line I don't get the nice yellow background highlight for this line.

docs/sphinx/users/comlinetools/formatlist.txt
```diff
@@ -2,34 +2,35 @@ List supported file formats
...
+    argument::
 
-    formatlist
+        formatlist -txt
 
-This can also be accomplished using by specifying the ``-txt`` argument:
+.. option:: -html
 
-::
+    This option allows to print the list of formats as HTML::
```
mtbc: Simply "prints" rather than "allows to print"?

Though other pages seem to use the future tense for some reason, so "This option will print the ..." might also be an option.

docs/sphinx/users/comlinetools/formatlist.txt
```diff
@@ -2,34 +2,35 @@ List supported file formats
...
 
-    formatlist -help
+    formatlist
 
-The default behavior is to print a plain-text list of formats:
+.. option:: -txt
 
-::
+    This can also be accomplished using by specifying the :option:`-txt`
+    argument::
```
mtbc: We're too distant from the previous sentence for the "This" to comfortably make immediate sense. Perhaps switch to follow the form of the other options, except with some additional brief note that it's the default? Not this PR's fault though.

docs/sphinx/users/comlinetools/formatlist.txt
```diff
@@ -2,34 +2,35 @@ List supported file formats
...
+    This option allows to print the list of formats as XML::
 
-    formatlist -html
+        formatlist -xml
 
-or XML:
+.. option:: -help
 
-::
+    Current usage information can be shown by running::
```
mtbc: "Current" seems a bit odd, unless there's a risk of my thinking that it will show me usage information from the past or the future, but I don't strongly object if you like it.

docs/sphinx/users/comlinetools/conversion.txt
sbesson: Correct, that may be a Sphinx limitation though which I cannot work around. The only alternate proposal is to have something like

```
.. option:: -tilex TILEX
.. option:: -tiley TILEY
...
```
Leaving the decision up to you before pushing this.

mtbc: It's not a big deal: happy to go with whichever option you think is best within what Sphinx allows.

mtbc:
Good to merge.

PR #1832, opened 2015-06-11
AVI: read frame index tables

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/11264.

To test, use the files from QA 7450 and 11011.  When opened in ImageJ (```Use virtual stack``` may be necessary), each file should have a frame count that matches what is indicated by ```mplayer```.

The only difference between Bio-Formats and mplayer should be with QA 11011 - in this file, there is a conflict between the plane height stored in the AVI container (408) and the plane height stored in each of the MJPEG compressed streams (204).  Bio-Formats is using the MJPEG height instead of scaling the images (which I think is what mplayer does).


bramalingam:

PR #1835, opened 2015-06-16
Metamorph: better detection of time series

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12503.

To test, use the files linked from http://fiji.sc/bugzilla/show_bug.cgi?id=868.  ```file1.stk``` and ```file2.stk``` should each be a single time series with this change (i.e. SizeZ = 1).  Without this change, SizeZ would have been > 1.  Builds should remain green; other datasets should not be affected.


hadim:
Thanks @melissalinkert. I confirm this PR fixes my issue with some Metamorph files.

bramalingam:
Works as expected. Please merge

PR #1836, opened 2015-06-16
Also include ome-transforms.xml in Maven specification JAR

sbesson:
While trying to consume Maven specification.jar for the OMERO exporter
integration tests, the transformation catalog could not be found as it
currently not bundled.

/cc @jburel


sbesson:
NB: the deployment of this PR will depend on https://github.com/openmicroscopy/bioformats/pull/1838

sbesson:
Closing in favor of a `develop` PR.

PR #1837, opened 2015-06-16
cope with file format version 2

bjoernthiel:


bjoernthiel:
Can we have this delivered as a patch?

PR #1834, opened 2015-06-11
Add a merge build profile to POMs

melissalinkert:
This allows the snapshot repository to be changed; by default, the
latest-build profile (== ome.snapshots repository) will be used.  If
'mvn -P merge-build' is run instead, then the merge-build profile (==
ome.unstable repository) is used.

This is the prerequisite for a BIOFORMATS-5.1-merge-maven job.  See https://trello.com/c/5S2dBuxW/227-bf-decoupling.

/cc @sbesson, @bramalingam

PR #1838, opened 2015-06-16
Add a merge build profile to POMs

melissalinkert:
This allows the snapshot repository to be changed; by default, the
latest-build profile (== ome.snapshots repository) will be used.  If
'mvn -P merge-build' is run instead, then the merge-build profile (==
ome.unstable repository) is used.

This was previously gh-1834.


sbesson:
Tested artifacts deployment via https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-merge-maven/4/console, passing the profile option to the `mvn` command is enough to get the artifacts deployed on the `ome.unstable` repository.
Will be investigating how to consume these artifacts from the OMERO/Ivy standpoint next..

sbesson:
Discussed today with @melissalinkert @joshmoore @jburel @bramalingam. The strategy above did not work as initially expected since there is no repository classifier that can be used by Ivy while retrieving Maven artifacts. An alternative to be explored is to use the profile to set a different groupId for merge artifacts.

sbesson:
Depending on the testing result of https://github.com/openmicroscopy/openmicroscopy/pull/3935, we may want to resurrect this PR/strategy for uploading merge SNAPSHOTS.

melissalinkert:
Replaced by gh-1867 (on develop).

PR #1839, opened 2015-06-16
bfconvert: fix plane indexes for multi-series exports

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12926.  This prevents the plane index from being too large for series beyond the first.

To test, choose any MIF (the ticket indicates ```test_images_good/leica-lif/Beta Canenin.lif```), and use ```bfconvert``` to convert to a single OME-TIFF.  With this change, the conversion should be successful, and ```showinf -nopixels``` on both files should show matching metadata.  Without this change, the conversion would have thrown an exception.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -552,6 +552,9 @@ public final class ImageConverter {
         write += e - m;
 
         nextOutputIndex.put(outputName, outputIndex + 1);
+        if (i == endPlane - 1) {
+          nextOutputIndex.remove(outputName);
+        }
 
         // log number of planes processed every second or so
         if (count == numImages - 1 || (e - timeLastLogged) / 1000 > 0) {


mtbc:
In comparing the metadata with `showinf`, `Beta Catenin.lif` has no global metadata, only series #0 metadata. The generated OME-TIFF appears to have only global metadata prefixed with Image000, Image001, Image002, Image003, Series004, where perhaps the Image000 data corresponds to that listed under series #0 for the other. This may be okay?

No other issues noted, so may be good to merge.

melissalinkert:
Yes, the metadata is expected - if you change the series passed to showinf for ```Beta Catenin.lif```, the metadata for that series should be present.  What is saved to the OME-TIFF is the union of all series hashtables, with the series/Image name prepended to the keys.

mtbc:
Aha, yes, thank you: this works exactly as one would expect, comparing `showinf` on the OME-TIFF with the original file with various `-series` options and looking at the `Image name:`s too. Good to merge.

PR #1841, opened 2015-06-18
Flex: add support for multiple plate acquisitions

melissalinkert:
If a directory contains multiple Meas_* directories, each with .flex
files, then each Meas_ directory will be automatically detected as an
acquisition of the same plate.

See https://trello.com/c/dyEq0xsZ/187-flex-runs.  I haven't yet thoroughly tested this in OMERO - that's the next step.  We haven't had real exposure to the multiple acquisitions use case before, so there may be some things that need changing/fixing.

The small dataset I was testing with was ```flex/karsten/Columbus_ExtendedData/Images/AFo_071214_measurement3...```.  This has 8 directories, each with 10 .flex files (one per well).  Picking any one of the .flex files results in 80 series, so 80 Images, 80 WellSamples, 1 Plate, 8 PlateAcquisitions.  I think all of the linkages are correct, but carefully checking the output of ```showinf -nopix -omexml``` is important.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FlexReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FlexReader.java
@@ -30,7 +30,9 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Replaced by gh-1845

PR #1843, opened 2015-06-18
Metamorph: fix Z and T settings when the .nd file is inconsistent

melissalinkert:
Back (forward?) ported from a6fb748 and 2e27ca2.

See QA 11128 and http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-June/005429.html.

To test, use the dataset in QA 11128; the .nd file is the one to open/import.  Without this change, choose the second or third series and verify that the first channel appears to be from a different stage position relative to the other two channels.  With this change, all three channels in the same series should clearly be from the same stage position.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -499,6 +499,12 @@ public class MetamorphReader extends BaseTiffReader {
         currentValue.append(line.substring(comma + 1).trim());
       }
 
+      if (!globalDoZ) {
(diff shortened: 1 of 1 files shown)


pwalczysko:
Sorry, did not manage tday
Please relist for tomorrow with my name.

pwalczysko:
Tested in Fiji, series 2 and 3 work as expected.

pwalczysko:
Tested in ImageJ, series 2 and 3 work as expected. The original mismatch (without this PR) was reproduced successfully.

Ready to merge.

PR #1844, opened 2015-06-19
Improve performance of displaying roi's and enable show all labels

bramalingam:
Update the bioformats_package.jar in the imageJ/plugins folder. (with the merge build)

Testing Instructions : 

1) Open an ome-tiff, which has more than 1000 roi's in it, using the Bio-Formats Importer in ImageJ.
and compare performance (with and without this PR).

2)Can also open an existing image within OMERO (trout/latest, user-4, image_id: 29753), and use the omero-ij plugin to open the image. 

In both cases, the performance should be much better with this PR included.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -330,11 +330,12 @@ public class ROIHandler {
                         }
                         if (sc != null) roi.setStrokeColor(sc);
                         manager.add(images[imageNum], roi, nextRoi++);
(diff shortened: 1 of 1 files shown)


pwalczysko:
Opening the image from OMERO without this PR in IJ takes more than a minute.

pwalczysko:
Will finish testing tomorrow, please relist with my name.

pwalczysko:
With this PR, the opening of user-4, image_id: 29753 takes ca ~10 sec.

pwalczysko:
Also from local disk, the performance is in the ~10 sec range. 

Ready to merge.

PR #1845, opened 2015-06-19
Flex: add support for multiple plate acquisitions

melissalinkert:
If a directory contains multiple Meas_* directories, each with .flex
files, then each Meas_ directory will be automatically detected as an
acquisition of the same plate.

See https://trello.com/c/dyEq0xsZ/187-flex-runs. I haven't yet thoroughly tested this in OMERO - that's the next step. We haven't had real exposure to the multiple acquisitions use case before, so there may be some things that need changing/fixing.

The small dataset I was testing with was ```flex/karsten/Columbus_ExtendedData/Images/AFo_071214_measurement3...```. This has 8 directories, each with 10 .flex files (one per well). Picking any one of the .flex files results in 80 series, so 80 Images, 80 WellSamples, 1 Plate, 8 PlateAcquisitions. I think all of the linkages are correct, but carefully checking the output of showinf -nopix -omexml is important.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FlexReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FlexReader.java
@@ -565,7 +565,18 @@ public class FlexReader extends FormatReader {
     for (int run=0; run<runCount; run++) {
       String plateAcqID = MetadataTools.createLSID("PlateAcquisition", 0, run);
       store.setPlateAcquisitionID(plateAcqID, 0, run);
-      store.setPlateAcquisitionName(runDirs.get(run).getName(), 0, run);
+
+      String acqName = runDirs.get(run).getName();
+      store.setPlateAcquisitionName(acqName, 0, run);
+
+      int timeStart = acqName.indexOf("(");
+      if (timeStart > 0) {
+        String time = acqName.substring(timeStart);
+        time = DateTools.formatDate(time, "(yyyy-MM-dd_HH-mm-ss)");
+
+        store.setPlateAcquisitionStartTime(new Timestamp(time), 0, run);
+        plateAcqStartTime = null;
+      }
 
       PositiveInteger maxFieldCount = FormatTools.getMaxFieldCount(fieldCount);
       if (maxFieldCount != null) {
--- a/components/formats-gpl/src/loci/formats/in/FlexReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FlexReader.java
@@ -30,7 +30,9 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import loci.common.Constants;
 import loci.common.DataTools;
@@ -103,6 +105,7 @@ public class FlexReader extends FormatReader {
   private int plateCount;
   private int wellCount;
   private int fieldCount;
+  private int runCount;
 
   private int wellRows, wellColumns;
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Combining of runs into a single plate looks good: http://dogfish.openmicroscopy.org/webclient/?show=plate-101

Is there any chance of having the run name filled out based on the `Meas` directories? Or alternatively, is there a better way to guarantee the mapping between directory and run?

melissalinkert:
Commit pushed to remove the run name from the Plate name, and add names to the PlateAcquisitions.  The Plate/PlateAcquisition names should now match the grandparent/parent directory names, but it's easy enough to change if a different format would be more useful.

joshmoore:
http://dogfish.openmicroscopy.org/webclient/?show=plate-151 certainly shows properly named measurements which was critical since the ordering was *not* as expected. A few questions:

 * Is there any potential for enforcing the ordering based on directory listing?
 * At least in this particular screen, the tabular information is linked via the short-name "Meas_01", etc. Can we assume that will often be the case? Does it make sense to drop the date info?
 * If so, can we capture that as start?

cc: @eleanorwilliams @manics @antalbalint

melissalinkert:
> Is there any potential for enforcing the ordering based on directory listing?

The run order should always match the directory listing order, per https://github.com/openmicroscopy/bioformats/pull/1845/files#diff-e11bfe5bdc0cab83e3eb911cd46a2dbcR511.  Or do I misunderstand the question?

> At least in this particular screen, the tabular information is linked via the short-name "Meas_01", 
> etc. Can we assume that will often be the case? Does it make sense to drop the date info?

Completely up to you all whether the short or long name is more sensible.  Switching to the short name is very easy.

> If so, can we capture that as start?

Will do in a forthcoming commit, since this is independent of the PlateAcquisition name.

joshmoore:
> The run order should always match the directory listing order, per https://github.com/openmicroscopy/bioformats/pull/1845/files#diff-e11bfe5bdc0cab83e3eb911cd46a2dbcR511.  Or do I misunderstand the question?

Something at the DB level then since `plate -> runs` is not ordered. The listing in the UI is based on the ID rather than the NAME, so for 5.1, we'll have to handle it at that level. cc: @jburel @will-moore 

> Completely up to you all whether the short or long name is more sensible.  Switching to the short name is very easy.

K. We'll discuss. If that turns out to be just a "nice to have" then this would be good to go.

> Will do in a forthcoming commit, since this is independent of the PlateAcquisition name.

Check.

melissalinkert:
PlateAcquisition StartTime commit pushed.

melissalinkert:
Anything further needed here?

manics:
AFAIK the last commit wasn't tested- I'll start an overnight import.

manics:
Looks good:
```
In [18]: [(datetime.datetime.fromtimestamp(qs.get('PlateAcquisition',i).getStartTime().val/1000), qs.get('PlateAcquisition',i).getName().val) for i in [251,252,253,253,255]]
Out[18]:
[(datetime.datetime(2011, 2, 22, 20, 6, 15), 'Meas_05(2011-02-22_20-06-15)'),
 (datetime.datetime(2011, 2, 22, 18, 0, 38), 'Meas_02(2011-02-22_18-00-38)'),
 (datetime.datetime(2011, 2, 22, 18, 42, 33), 'Meas_03(2011-02-22_18-42-33)'),
 (datetime.datetime(2011, 2, 22, 18, 42, 33), 'Meas_03(2011-02-22_18-42-33)'),
 (datetime.datetime(2011, 2, 22, 17, 17, 8), 'Meas_01(2011-02-22_17-17-08)')]
```
(Note web currently displays `Creation Date: 2015-07-07 14:06:26` for the plate and all acquisitions)

PR #1847, opened 2015-06-22
C++: Add cmake superbuild logic for bzip2, tiff and boost

rleigh-dundee:
Note that because Windows doesn't have a `patch` command, I've had to include full files in `cpp/cmake/External_foo_files` rather than patches to apply.  There's a CMake ticket to add a built-in patch command, but it doesn't exist yet.  Diff the -source directory against the original downloaded source to see the true differences (only needed for `tiff` source files `configure.ac`, `libtiff/tiffiop.h`, `libport.h`, `test/rewrite_tag.c` and `tools/tiffgt.c`).

- bzip2: Add cmake build files; build with cmake;
- tiff: Add cmake build files and patched sources; build with cmake
- boost: Configure with msvc/clang/gcc toolset; build with cmake-configured compiler

These are by no means final.  Tested on Linux and BSD but will need further Windows work.  Their integration into the wider project will be completed with the addition of further libraries (xerces-c, libjbig, libjpeg, liblzma2, libicu).

--------

Testing:

The aim of the testing here is to demonstrate that the `bzip2`, `tiff` and `boost` targets invoke the external build of the corresponding source and complete without error.

Configure with something similar to

```
cmake -Dbioformats-superbuild:BOOL=ON -DBIOFORMATS_EP_SOURCE_CACHE:PATH=~/sources -DCMAKE_INSTALL_PREFIX=/tmp/bf ~/code/bioformats
```

This will enable the superbuild and source cache.  Then, to test each of the new targets:

```
make bzip2
make tiff
make boost
```

- tiff depends on zlib
- boost depends on zlib and bzip2

so if you build the targets without first having built their prerequisites, these will be triggered automatically.

If you run `make` you'll get a bioformats-cpp built against the new libraries as for #1830. **Note:** this needs LD_LIBRARY_PATH or DYLD_LIBRARY_PATH setting to $builddir/superbuild-install/$libdir or else it will fail; this will need addressing in a followup PR to make the process fully transparent.  Part of the complexity of the superbuild is setting the job environment so that this works; it doesn't intrinsically support setting environment variables as part of an external project, so you need to invoke a script to set the environment and *then* do the build.


joshmoore:
I'm having issues getting this to pass on 10.10:

```
  539  rm -rf cppbuild/
  540  mkdir cppbuild
  541  cd cppbuild/
  542  cmake -Dbioformats-superbuild:BOOL=ON -DBIOFORMATS_EP_SOURCE_CACHE:PATH=/tmp/cpp/cache -DCMAKE_INSTALL_PREFIX=/tmp/cpp/bf ..
  543  export DYLD_LIBRARY_PATH=`pwd`/superbuild-install/lib
  544  make
```
produces:
```
-- Installing: /opt/ome9/components/bioformats/cppbuild/superbuild-install/share/doc/bioformats-superbuild/html/man/tiffsv.1.html
dyld: Symbol not found: __cg_TIFFClientOpen
  Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
  Expected in: /opt/ome9/components/bioformats/cppbuild/superbuild-install/lib/libTIFF.dylib
 in /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
/bin/sh: line 1: 34896 Trace/BPT trap: 5       /usr/local/Cellar/cmake/3.2.2/bin/cmake -E touch /opt/ome9/components/bioformats/cppbuild/tiff-prefix/src/tiff-stamp/tiff-install
make[2]: *** [tiff-prefix/src/tiff-stamp/tiff-install] Error 133
make[1]: *** [CMakeFiles/tiff.dir/all] Error 2
make: *** [all] Error 2
```

rleigh-dundee:
@joshmoore Maybe unset DYLD_LIBRARY_PATH and try DYLD_FALLBACK_LIBRARY_PATH instead?

Also, please do the build in the order detailed in the testing notes.  i.e. build tiff *first* and then export the library path for the last step only.  MacOSX is doing odd things here.  I already dropped zlib and libpng from the superbuild when running on MacOSX, since the base system provides them and they conflicted in a similar way.  If there's a better way to configure the linker to avoid conflicts, I'm all ears.  Likewise, if we can use a system imageio library rather than building libtiff ourselves, we can also do that.

joshmoore:
Trying FALLBACK now. I've tested the other commands (in order) and failed at the same location, so I cleaned and then tried a full new clean build.

joshmoore:
FALLBACK seems to be running further, I'll keep you posted.

rleigh-dundee:
@joshmoore Note you'll need to test the other commands with DYLD_LIBRARY_PATH unset.  If you look at your error, it's actually unable to run `cmake -E`.

joshmoore:
Full build passed:

```
-- Installing: /opt/ome9/components/bioformats/cppbuild/bioformats-install/share/icons/actions/ome-zoom2d.svg
[100%] Completed 'bioformats'
[100%] Built target bioformats
bioformats/cppbuild$ echo $DYLD_FALLBACK_LIBRARY_PATH
bioformats/cppbuild/superbuild-install/lib
bioformats/cppbuild$ echo $DYLD_LIBRARY_PATH

```

The individual commands also seem to work fine (with or without FALLBACK set), though it did seem as if `make tiff` didn't build `zlib` on 10.10 and instead picked up the system version:

```
bioformats/cppbuild$ otool -L superbuild-install/lib/libtiffxx.dylib
superbuild-install/lib/libtiffxx.dylib:
        @rpath/libtiffxx.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libtiff.5.dylib (compatibility version 5.0.0, current version 5.2.1)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
        /usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
        /usr/local/lib/libjpeg.8.dylib (compatibility version 13.0.0, current version 13.0.0)
        /usr/local/lib/liblzma.5.dylib (compatibility version 8.0.0, current version 8.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
```

rleigh-dundee:
@joshmoore That's good to know.  I did make a commit to force the staging path to be the final install path on OS X but I'll leave it out if it's not needed.  The reason the separate commands work without the FALLBACK is that none of these targets actually *use* the libraries after installation.  Bio-Formats, however, does need to use the installed boost library when it does its initial checks, and this is what fails.

WRT zlib, as mentioned above we currently avoid building it on MacOS and use the system copy due to the linker errors.  If using DYLD_FALLBACK_LIBRARY_PATH is a solution to that then we can revisit that decision.

So it all seems acceptable with DYLD_FALLBACK_LIBRARY_PATH.

joshmoore:
Agreed. Thanks, @rleigh-codelibre / @rleigh-dundee

cpp/cmake/External_tiff_files/test/rewrite_tag.c
```diff
@@ -0,0 +1,344 @@
...
+    if (!TIFFSetField(tif, TIFFTAG_BITSPERSAMPLE, 8)) {
+        fprintf (stderr, "Can't set BitsPerSample tag.\n");
+        goto failure;
+    }
+    if (!TIFFSetField(tif, TIFFTAG_SAMPLESPERPIXEL, 1)) {
+        fprintf (stderr, "Can't set SamplesPerPixel tag.\n");
+        goto failure;
+    }
+    if (!TIFFSetField(tif, TIFFTAG_ROWSPERSTRIP, rows_per_strip)) {
+        fprintf (stderr, "Can't set SamplesPerPixel tag.\n");
```
mtbc: maybe RowsPerStrip tag?

rleigh-dundee: I'm not sure what you mean here.  This is the verbatim upstream source; the only change in this file is removing the void cast

```
(void) argc
(void) argv
```

from `main()`, which was breaking compilation with visual studio.

rleigh-dundee: Sorry, I wasn't reading properly, you were referring to the incorrect string.  I'll push a fix to the upstream developers, but leave it untouched here until then since this is intended to be a verbatim copy.

mtbc: :+1:

PR #1848, opened 2015-06-22
TIFF: reduce number of seeks in OnDemandLongArray

melissalinkert:
This should reduce the seek count to O(1) instead of O(n) when calling
toArray().

Fixes https://trac.openmicroscopy.org/ome/ticket/12930.  To test, use the files from QA 11133 and QA 11125.  Without this change, ```showinf``` on each file should take many minutes to display images.  With this change, the same command should have minimally an order of magnitude improvement in time.  I'd suggest repeating the test several times with each branch, and comparing the average and minimum times (not just the first time).

For reference, on my Linux VM, QA 11133 takes 1m21s and QA 11125 takes 1m6s with this change.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/OnDemandLongArray.java
+++ b/components/formats-bsd/src/loci/formats/tiff/OnDemandLongArray.java
@@ -33,6 +33,7 @@
 package loci.formats.tiff;
 
 import java.io.IOException;
+import loci.common.DataTools;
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Vast improvement.  ~10s on Windows now with the image from QA#11125.

rleigh-dundee:
All jobs are green (except for full-repo-win, and that looks like jenkins issue).  Please merge.

PR #1850, opened 2015-06-23
Remove dead MRC specification link

hflynn:
Link has been dead since 16th June and the domain doesn't seem to exist any more so I'm assuming it's not coming back (also we have another specification link anyway).


hflynn:
Docs merge build is green today and the format page looks fine on staging -http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/formats/mrc.html so this should be good to merge.

Issue #1851, opened 2015-06-23
ImageJ plugin import remote with https


dpwrussell:
When trying to import a a remote url which is https, it will fail. Presumably this is because it does not recognise the protocol and defaults to trying to open it as a file:

```
java.io.FileNotFoundException: https:/s3.amazonaws.com/dpwr/s3test/bus.png (No such file or directory)
	at java.io.RandomAccessFile.open0(Native Method)
	at java.io.RandomAccessFile.open(RandomAccessFile.java:316)
	at java.io.RandomAccessFile.<init>(RandomAccessFile.java:243)
	at loci.common.NIOFileHandle.<init>(NIOFileHandle.java:123)
	at loci.common.NIOFileHandle.<init>(NIOFileHandle.java:139)
	at loci.common.NIOFileHandle.<init>(NIOFileHandle.java:148)
	at loci.common.Location.getHandle(Location.java:321)
	at loci.common.Location.getHandle(Location.java:291)
	at loci.common.Location.getHandle(Location.java:281)
	at loci.common.Location.getHandle(Location.java:271)
	at loci.common.Location.checkValidId(Location.java:346)
	at loci.formats.ImageReader.getReader(ImageReader.java:173)
	at loci.plugins.in.ImportProcess.createBaseReader(ImportProcess.java:628)
	at loci.plugins.in.ImportProcess.initializeReader(ImportProcess.java:486)
	at loci.plugins.in.ImportProcess.execute(ImportProcess.java:139)
	at loci.plugins.in.Importer.showDialogs(Importer.java:137)
	at loci.plugins.in.Importer.run(Importer.java:75)
	at loci.plugins.LociImporter.run(LociImporter.java:78)
	at ij.IJ.runUserPlugIn(IJ.java:199)
	at ij.IJ.runPlugIn(IJ.java:163)
	at ij.Executer.runCommand(Executer.java:132)
	at ij.Executer.run(Executer.java:65)
	at java.lang.Thread.run(Thread.java:745)
```

ImageJ's `File -> Import -> URL` does manage to open from https URLs.

bramalingam:
Hi Douglas, 

Thank you for reporting the bug. The https case was not handled in the importer, and the following PR should fix the same,
https://github.com/openmicroscopy/bioformats/pull/1854

Best,
Balaji

PR #1854, opened 2015-06-24
Bug Fix : Read https urls

bramalingam:
Ref: https://github.com/openmicroscopy/bioformats/issues/1851

To test:
1) Plugins--> Bio-Formats--> Bio-Formats Remote Importer
and paste this url : https://s3.amazonaws.com/dpwr/s3test/bus.png

This should open the image in ImageJ.

Without this PR you should be getting a "File not found" exception.

### DIFF
--- a/components/formats-common/src/loci/common/Location.java
+++ b/components/formats-common/src/loci/common/Location.java
@@ -300,7 +300,7 @@ public class Location {
       LOGGER.trace("no handle was mapped for this ID");
       String mapId = getMappedId(id);
 
-      if (id.startsWith("http://")) {
+      if (id.startsWith("http://") || id.startsWith("https://")) {
         handle = new URLHandle(mapId);
       }
       else if (allowArchiveHandles && ZipHandle.isZipFile(mapId)) {


components/formats-common/src/loci/common/URLHandle.java
```diff
@@ -64,8 +64,10 @@
    * Constructs a new URLHandle using the given URL.
    */
   public URLHandle(String url) throws IOException {
-    if (!url.startsWith("http") && !url.startsWith("file:")) {
-      url = "http://" + url;
+    if (!url.startsWith("https") && !url.startsWith("file:")) {
+        url = "https://" + url;
+    }else if (!url.startsWith("http") && !url.startsWith("file:")) {
+        url = "http://" + url;
```
mtbc: This is a strange change. Why not leave it as it was? (Note the cunning lack of a colon suffixing the first "http" originally.)

bramalingam: True, its actually the prefix added to the url(url = ...) that was concentrated upon.

mtbc: Could you give an example of an initial value for `url` that your `else` will do something useful for?

bramalingam: Something like ?
url = "http://" + url;
if (!url.startsWith("https") && !url.startsWith("file:")) {
 url = "https://" + url;
}

mtbc: My guess was that this code is to guess a valid prefix if one is not already supplied but I may be missing some important intent here.

bramalingam: True. I had updated my previous comment , please check if that makes looks good?

pwalczysko:
In Fiji, I replaced all the jars and the bioformats_plugins.jar with the jars from https://ci.openmicroscopy.org/view/5.1/job/BIOFORMATS-5.1-merge-build/lastSuccessfulBuild/artifact/artifacts/. But even after thsi change, I still get the ``File not found`` exception.

pwalczysko:
And the same in ImageJ after I have replaced the bioformats_package.jar.
Does not work.

bramalingam:
Checking if this is included in the merge build.

pwalczysko:
Seems to be included as per https://github.com/snoopycrimecop/openmicroscopy/commit/3077218c60f98244e6c32159e970a80b208bbe65.

screen shot 2015-06-25 at 10 28 23

components/formats-common/src/loci/common/URLHandle.java
mtbc: I confess that I don't understand how it improves anything over the original but I haven't worked with this code so wiser heads may see things that I don't.

bramalingam:
I have updated the link. Please recheck.

pwalczysko:
Sorry, even with https://s3.amazonaws.com/dpwr/s3test/bus.png I am still getting the same error in IJ.

pwalczysko:
The bioformats_package.jar from https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/lastSuccessfulBuild/artifact/artifacts/ does not give the expected results. 

The bioformats_package.jar from @bramalingam 's local build does give the expected result.

```
-rwx------@   1 petr  staff  28626497 25 Jun 10:55 bioformats_package.jar*
-rw-r--r--@   1 petr  staff  28627675 25 Jun 11:04 bioformats_package_Balaji.jar
```
Note the slight difference in the file size as noted above (``bioformats_package_Balaji.jar`` is the one @bramalingam has provided from his local build)

components/formats-common/src/loci/common/URLHandle.java
melissalinkert: @mtbc has it right - the original code was meant to add a valid URL scheme (```http://```) if none is present.

Assuming URLHandle can accept a value of ```url``` in one of the following formats:

1. ```abc.com/xyz.png```
2. ```file://a/b/c/xyz.png```
3. ```http://abc.com/xyz.png```
4. ```https://abc.com/xyz.png```

the original code would only have changed the value of ```url``` in case 1 (it would become ```http://abc.com/xyz.png```).

With this change:

1. becomes ```https://abc.com/xyz.png```
2. is not changed
3. becomes ```https://http://abc.com/xyz.png```
4. is not changed

joshmoore: Can I suggest some unit tests? :smile:

bramalingam:
@melissalinkert & @mtbc : ahhhh yes, Thank you for the comments (Some muddled thinking on my part!), have fixed the bug now. Please check.

bramalingam:
The change in URL handle was rhetorical, when compared to the original version of the file. So I have reverted to the original version.

joshmoore:
Kicked travis.

pwalczysko:
Please relist for tomorrwo, did not manage today.

pwalczysko:
Tested in ImageJ. Was able to reproduce the pre-this-PR behaviour. With this PR it works as expected. As long as @joshmoore @melissalinkert are happy with the code changes, ready to merge.

joshmoore:
I'd suggest we rebase away the commits to a single clean one and look into unit tests if at all possible.

melissalinkert:
@bramalingam: you're still working on this?

bramalingam:
I can close and reopen a clean PR with just one commit, if that helps.

joshmoore:
```
$ git rebase -i f8213e88cc0b8e44c46c59952a95f98d277e2f76^
$ git push -f gh Import_https
```
should do it. Assuming you're remote is named `gh`

bramalingam:
@joshmoore @melissalinkert  : Can you please check now?

mtbc:
The code change looks reasonable.

joshmoore:
Yes, exactly. Thanks, @bramalingam

melissalinkert:
Merging so we can have this in 5.1.3.  Card opened (https://trello.com/c/YUSaP4Qz/270-update-urlhandle-unit-tests-for-https) to update the unit tests in the near future.

PR #1855, opened 2015-06-24
PNG: fix interlaced images that have a width or height that is not a multiple of 8

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12850.

See also
http://www.libpng.org/pub/png/spec/1.2/PNG-DataRep.html#DR.Interlaced-data-order.

To test, use the file from QA 11038.  Without this change, ```showinf``` on the file should throw an exception; with this change, ```showinf``` should produce the same image as ImageMagick (the ```display``` command) and ```File > Open``` in ImageJ.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/APNGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/APNGReader.java
@@ -405,6 +405,16 @@ public class APNGReader extends FormatReader {
       int nRowBlocks = height / 8;
       int nColBlocks = width / 8;
 
+      // row and column counts have to be adjusted when the image
+      // dimensions are not multiples of 8
+
+      if (8 * nRowBlocks != height) {
+        nRowBlocks++;
+      }
+      if (8 * nColBlocks != width) {
+        nColBlocks++;
+      }
+
       if (nRowBlocks <= 0) {
         nRowBlocks = 1;
       }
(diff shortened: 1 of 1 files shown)


melissalinkert:
Fix pushed for https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/535/

components/formats-bsd/src/loci/formats/in/APNGReader.java
```diff
@@ -420,12 +437,168 @@ else if (compression == 0) {
...
+                break;
+              case 5:
+                if (i == 3) {
+                  passWidth -= (PASS_WIDTHS[i] - 1);
+                }
+                if (i == 5) {
+                  passWidth -= (PASS_WIDTHS[i] - 2);
+                }
+                if (i == 4) {
+                  passWidth -= (PASS_WIDTHS[i] =- 3);
```
mtbc: is this second `=` desired? (also could make these "else if"'s but doesn't much matter really)

mtbc:
It's a strange looking image but other viewers agree that's what it looks like! So, this PR looks good functionally. :+1:

components/formats-bsd/src/loci/formats/in/APNGReader.java
melissalinkert: No, that's just a typo - fix pushed.

PR #1856, opened 2015-06-24
TIFF export: save dimensions in ImageJ-style comment

melissalinkert:
See https://trello.com/c/w49BvpYX/11-save-imagej-metadata-in-tiff.

To test, use ```bfconvert``` to convert any file with SizeZ and/or SizeC > 1 to TIFF and check the output of ```showinf -nopix``` on the converted file.  I used ```test_images_good/andor/FURA-2CH-T.tif```, but anything will do.

The test without this change should result in the converted file having SizeZ == SizeC == 1, with all planes stored in T.  With this change, the Z, C, and T sizes should match between the original and converted files.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -366,7 +366,8 @@ public class TiffWriter extends FormatWriter {
     int z = retrieve.getPixelsSizeZ(series).getValue().intValue();
     int t = retrieve.getPixelsSizeT(series).getValue().intValue();
     ifd.putIFDValue(IFD.IMAGE_DESCRIPTION,
-      "ImageJ=\nchannels=" + channels + "\nslices=" + z + "\nframes=" + t);
+      "ImageJ=\nhyperstack=true\nimages=" + (channels * z * t) + "\nchannels=" +
+      channels + "\nslices=" + z + "\nframes=" + t);
 
     int index = no;
     for (int i=0; i<getSeries(); i++) {
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -317,14 +317,14 @@ public class TiffWriter extends FormatWriter {
     ifd.put(new Integer(IFD.IMAGE_LENGTH), new Long(height));
 
     Length px = retrieve.getPixelsPhysicalSizeX(series);
-    Double physicalSizeX = px == null ? null : px.value(UNITS.MICROM).doubleValue();
+    Double physicalSizeX = px == null || px.value() == null ? null : px.value(UNITS.MICROM).doubleValue();
     if (physicalSizeX == null || physicalSizeX.doubleValue() == 0) {
       physicalSizeX = 0d;
     }
     else physicalSizeX = 1d / physicalSizeX;
 
     Length py = retrieve.getPixelsPhysicalSizeY(series);
-    Double physicalSizeY = py == null ? null : py.value(UNITS.MICROM).doubleValue();
+    Double physicalSizeY = py == null || py.value() == null ? null : py.value(UNITS.MICROM).doubleValue();
     if (physicalSizeY == null || physicalSizeY.doubleValue() == 0) {
       physicalSizeY = 0d;
     }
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
With your PR:

```
ImageJ=
channels=2
slices=144
frames=1
```

Native from ImageJ save as tiff:

```
ImageJ=1.48v
images=288
channels=2
slices=144
hyperstack=true
mode=composite
unit=pixel
loop=false
min=1.0
max=4095.0
```

While Bio-Formats can read the former at get the correct dimensionality, it isn't used properly by imagej using "file->open".  This just results in a basic image series with no higher dimensions (though opening the channels tools and changing the display mode switches it on).  It works if you open it as a virtual stack though.  But the imagej metadata opens it with the correct dimensionality.  Should we change to match the default imagej behaviour here?

I'm wondering if it should simply be setting the other options as well, at least adding

```
hyperstack=true
mode=composite
```

And we could probably also set `unit` to the physical size X unit, if available, otherwise default to pixel.

Not for this PR, but could we also write out TIFF tags 50838 and 50839 (ImageJMetadataByteCounts and ImageJMetadata)?  Is there a proper spec for these fields?
Also, should I update the C++ TIFFWriter to write out the ImageJ header as well (we already handle reading)?

It looks good to merge as is, but I think adding the additional hyperstack/mode keys would be a useful improvement for imagej interoperability.

melissalinkert:
Fix pushed which should result in ```File > Open``` picking up the correct dimensions.  Do keep in mind that ImageJ's TIFF reader makes many assumptions based upon the first IFD, so while the dimensions should be correct, any images beyond the first may not be read correctly.

The point here was not to fully support ImageJ's metadata model, but rather to reduce the number of reports that converting to plain TIFF removes the dimensions (since this is popularly thought of as a bug, even though it isn't really).  There is no formal specification for any of the ImageJ-specific tags, unless you count http://rsb.info.nih.gov/ij/developer/source/ij/io/TiffDecoder.java.html.

rleigh-dundee:
To reproduce the problems: Files are in `/ome/team/rleigh/bf-ij-tiff`.  I have reproduced on Windows and MacOS X.

```
tools/bfconvert vessels.ome.tiff /tmp/test.tiff
```

You will see the corruption in the above screenshot.

```
bfconvert test2.tif testfail.tiff
```

This will result in the exception above.  This file was created by loading `vessels.ome.tiff` into ImageJ with the bioformats importer, and then saving as tiff (to get the ImageJ metadata for comparing with the bioformats output).  This file is readable by both imagej and bioformats showinf; but it fails with bfconvert.

rleigh-dundee:
Note that the exception occurs without this PR merged in.  The corrupt pixel data does not occur unless the PR is merged in, though it may merely be exposing a bug in the tiff reader when this metadata is present.

melissalinkert:
Commit pushed to address the exception.

The incorrect images are, as noted above (https://github.com/openmicroscopy/bioformats/pull/1856#issuecomment-118373122), a result of an assumption made by ImageJ's TIFF reader.  If an ImageJ-style comment is present, ImageJ's reader assumes that all planes are written sequentially with no gaps, so the offsets stored in every IFD beyond the first are ignored.

See the the ```getTiffInfo``` method in http://rsb.info.nih.gov/ij/developer/source/ij/io/TiffDecoder.java.html, and in particular the ```ignore extra IFDs``` comment.

rleigh-dundee:
@melissalinkert I'm afraid I still get the same NPE with your latest commit.

melissalinkert:
Should be fixed now.

rleigh-dundee:
@melissalinkert Thanks, this fixes the NPE for me.  Looks good to merge.

On the ImageJ front, should a ticket be opened about it being unable to read our TIFFs?

melissalinkert:
The ImageJ behavior is intentional, I believe - it's meant as a performance optimization.

I guess the question is which "bug" is preferable: dimension information being lost on export to plain TIFF, or ```File > Open``` in ImageJ not showing correct images for all exported plain TIFFs with multiple planes.  I'm not sure at this point which is going to cause more problems in the long run.  Expanding CC in case anyone else has strong feelings.

/cc @pwalczysko, @bramalingam, @joshmoore, @jburel

pwalczysko:
After discussion with @rleigh-dundee I think that the ``File > Open`` in ImageJ bug is fairly minor from our point of view. 
Is the workflow of the people who asked for the additional metadata dependent on ``File > Open`` in ImageJ ? If yes, reconsider. If not, I think merge  this and submit a reasonably formulated bug report for ImageJ list.

Possibly wait till other cc-ed persons express their opinion as well.

mtbc:
cc'd people ought all now to be back.

jburel:
I will agree that the issue of ```File > Open``` in ImageJ is minor. we could certainly add in the doc that for  metadata reading it is preferable to use the BF importer

rleigh-dundee:
I think that it's fairly clear after looking that the ImageJ behaviour is non-standard and that it shouldn't block this being merged.  The user can always use the Bio-Formats importer.

melissalinkert:
Merging as discussed in standup.

PR #1858, opened 2015-06-29
C++: Add cmake superbuild logic for icu and xerces-c

rleigh-dundee:
This completes the minimum necessary external projects to build bioformats-cpp.  Others will be needed in the future e.g. jpeg/jbig to support more TIFF codecs, but this gives it what we need for working on initial Windows support.

In addition to adding icu and xerces, this cleans up and generalises some of the logic to make it easier to add more projects in the future.

--------

Testing: as in https://github.com/openmicroscopy/bioformats/pull/1847

This time, try

```
make icu
make xerces
```

as well as plain `make`.  This should now build to completion on a system without any of the library dependencies installed.

MacOS X note: you need to set DYLD_FALLBACK_LIBRARY_PATH=$(builddir)/superbuild-install/lib for the bioformats build to work.  We'll make this automatic in a future update, but this requires factoring all the configure/build/install/test steps into separate reusable scripts.  You can see this done for Boost, where it sources the ExternalProjectEnvironment script to configure the environment.


mtbc:
> without any of the library dependencies installed

which ones? Is there a VM I ought to start from?

rleigh-dundee:
The library dependencies means without boost, tiff, png, xerces, icu, bzip2, zlib installed (or, at least the development packages since some will be impossible to remove).  This is simply to verify that it is linking against the versions built as part of the build, rather than silently using the system copy--it should in fact work just fine if the system copies are around.

joshmoore:
@rleigh-dundee : I think the question is the inverse, what *do* you want installed? (Which gets us back to the test & its definition of a node, etc)

rleigh-dundee:
At the moment, it's all the non-library dependencies from http://www.openmicroscopy.org/site/support/bio-formats5.1/developers/cpp/overview.html#prerequisites
i.e. cmake, python are the only strictly required dependencies; the optional non-library dependencies may or may not be present

joshmoore:
And Visual Studio, right?

rleigh-dundee:
Yes, the compiler is a given.

mtbc:
Argh, this needs Windows too? (Sorry, just glanced at the PR quickly while trying to run standup, and guessed I could use a local chroot if all else failed.) I haven't even figured out my latest OMERO issues there and I'll certainly have some of the library dependencies already scattered around from previous Windows/C++/BF stuff, but even how to find out what I'm not immediately sure. The Windows VM I used under Linux died for licensing reasons and (as usual) I never heard back from CLS-IT about my attempt to get a new image.

Is there a suitable VM image available, or anyone else who might be in a better position to review this PR? I could start poking around with Windows now but it moves like molasses (I'm considering wholly wiping it and installing from fresh media without CLS "enhancements") so even in the best case it would be many hours before I could get to somewhere useful.

rleigh-dundee:
@mtbc No, it does not need windows at all.

mtbc:
Ah, ok! Well that helps. :smiley:

mtbc:
Builds to completion in a very minimal chroot under Debian. :+1:

PR #1861, opened 2015-07-02
POM: add merge build profile and classifier

melissalinkert:
This replaces gh-1838.  A ```merge-build``` profile is still added, which is activated with the ```-P merge-build``` command line option.  Instead of changing the deployment repository, though, it adds the ```merge``` classifier to the generated artifacts.  The default behavior of ```mvn``` should be unchanged.

See https://maven.apache.org/pom.html#Dependencies.  Consuming merge build artifacts from Ivy would likely look like: http://stackoverflow.com/questions/16344151/ivy-retrieve-with-classifiers.  I'm not totally sure of the potential stumbling blocks here, but it looks like classifiers are already being used when packaging some OMERO components, so hopefully this isn't too far off base.

I did try to get groupId switching working as discussed, but that can't be done within a profile, and I haven't found a way to override the groupId on the command line.  If that's still the preferred approach, we may need separate merge build POMs or a pre-processing step to update the groupId (similar to how the version number is changed).  Certainly happy to discuss further or try other things; the current commit just seemed like the "Maven way" from everything I've read so far.

/cc @sbesson, @bramalingam, @joshmoore, @jburel

### DIFF
--- a/components/autogen/pom.xml
+++ b/components/autogen/pom.xml
@@ -67,6 +67,7 @@
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
         <configuration>
+          <classifier>${envClassifier}</classifier>
           <archive>
             <manifest>
               <mainClass>MetaSupportAutogen</mainClass>
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -106,6 +106,7 @@
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
         <configuration>
+          <classifier>${envClassifier}</classifier>
           <archive>
             <manifest>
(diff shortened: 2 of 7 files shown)


sbesson:
Tested this PR by:
- deploying the `merge` classified artifacts to `ome.snapshots` alongside the regular latest build (with `null` classifiers)
- trying to consume these artifacts from the OMERO build using the `m:classifier` attribute in the Ivy dependency. 

My major issue while experimenting this is that the latest SNAPSHOT is always resolved before the classifier. Depending on the classifier of the last SNAPSHOT and the classifier used in the Ivy dependency, the resolution may or may not work. Unless someone can think of a workaround, this does not seem to solve our issue. Closing this PR.

PR #1864, opened 2015-07-06
Metamorph TIFF: better check for dimension consistency

melissalinkert:
This should do a better job of ensuring that Z * C * T is equal to the
number of planes.

Fixes https://trac.openmicroscopy.org/ome/ticket/12940.

To test, use the files from QA 11135.  Verify that there are 61 planes in each file using ```tiffinfo filename.tif | grep -c "TIFF Directory"```.  Without this change, ```showinf``` should show that both files have SizeZ * SizeC * SizeT != 61.  With this change, SizeT should be 61 for both files.


rleigh-dundee:
The change looks reasonable and the testing shows the behaviour described above.  Please merge.

PR #1865, opened 2015-07-06
ScreenReader: implement reopenFile

joshmoore:
Memoizing a ScreenReader with OpenlabTiff files led to a
NullPointerException. Now each reader in the ScreenReader
has its reopenFile method called.

Testing with OMERO:
 * activate the `ScreenReader` in both lib/client and lib/server
 * import the Rad52 screen
 * view the plate in OMERO.web

Before this PR, no thumbnails were loaded. With this PR, thumbnails load normally. Note: if you try the above with only one import of the screen, it will be necessary to delete the memo file.


melissalinkert:
Works for me with a minimal memoizing test outside of OMERO.  Just waiting on Travis/green merge build (not blocked by full-repo jobs though).

PR #1866, opened 2015-07-06
Also include ome-transforms.xml in Maven specification JAR

sbesson:
While trying to consume Maven specification.jar for the OMERO exporter integration tests, the transformation catalog could not be found as it currently not bundled.


melissalinkert:
Any reason not to merge this now?

PR #1867, opened 2015-07-06
Add a merge build profile to POMs

melissalinkert:
This allows the snapshot repository to be changed; by default, the
latest-build profile (== ome.snapshots repository) will be used.  If
'mvn -P merge-build' is run instead, then the merge-build profile (==
ome.unstable repository) is used.

Replaces gh-1838.


sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-maven/11/. :+1:

PR #1868, opened 2015-07-06
Metamorph: skip TIFF files with no IFDs

melissalinkert:
See QA 11163 and http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-July/005488.html.

Without this change, ```showinf``` on the .nd file from QA 11163 should result in an exception, as noted in the thread.  With this change, ```showinf``` should initialize the .nd file without error.  The series count should match the ```NStagePositions``` value in the .nd file, and every series should open without error (this is easiest to test in ImageJ with the ```Open all series``` option).  Only the first 4 series will contain valid image data - everything else should be completely black.


sbesson:
Tested via merge build of OMERO with this PR included. All TIFF files included in the dataset are properly detected and imported into the server with the four first series having non-zero data:

screen shot 2015-07-10 at 17 17 18

Good to merge once the repository jobs are green.

PR #1869, opened 2015-07-06
PSD: fix bug in pixel data offset calculation

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12629.

To test, use the file from QA 10347.  Without this change, ```showinf``` or ImageJ should show an image that looks obviously wrong.  With this change, ```showinf``` and ImageJ should show the same thing as Image/GraphicsMagick - a single RGB image with three text labels in the upper-left corner.


dominikl:
Works like expected, good to merge!

PR #1870, opened 2015-07-07
Metamorph: fix multi-series with mismatching Z

sbesson:
Fixes http://trac.openmicroscopy.org.uk/ome/ticket/12951

With this PR included, Bio-Formats should properly include all TIFF files of the dataset linked in the ticket with the even series being single-Z brightfield images and the odd series being multi-Z fluorescent images e.g. using command-line tools:

```
showinf -series 0 -fast -autoscale /path/to/file.nd
showinf -series 1 -fast -autoscale /path/to/file.nd
```


melissalinkert:
Code changes look reasonable.  Still need to double-check the configuration changes (it's not 100% clear why the names would have changed), but expect to merge Friday PM.

PR #1872, opened 2015-07-09
Imspector: better detection of Z stacks

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/11722.

To test, use the files from QA 7702 and QA 9468.

The file from QA 9468 should have two channels with this change, as indicated by http://lists.openmicroscopy.org.uk/pipermail/ome-users/2014-August/004673.html.  The .msr files in QA 7702 should match dimensions/images with the corresponding OME-TIFF file(s) in the same directory.  Without these changes, the .msr files from both QA issues will have thrown an exception upon initialization.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ImspectorReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImspectorReader.java
@@ -102,22 +102,14 @@ public class ImspectorReader extends FormatReader {
     }
     index += no;
 
-    int block = 0;
-    int plane = 0;
-    while (index >= 0) {
-      int planeCount = planesPerBlock.get(block);
-      if (planeCount <= index) {
-        index -= planeCount;
-        block++;
-      }
-      else {
-        plane = index;
-        index = -1;
-      }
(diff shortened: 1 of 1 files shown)


mtbc:
Looks good. The actual image in `02 -  500x500x105_1046x1046x57_1.875um_steps.msr` is pretty dim, not sure if minmax is working great there, but all the images import and seem usable. Using this PR description as guide, the only odd thing I notice is that, in Insight, `Arf_FLIM_z-stack_1.msr` is described,

msr

but `02 - Arf_z-stack_FLIM_TDC_C0.ome.tif` instead has just,

ome tif

Still, this PR is certainly an improvement! :+1:

melissalinkert:
The difference in OMERO is expected - the original .msr file contains FLIM data, so the time points are stored in "t" (lifetime) instead of T via a ModuloT annotation.  The corresponding OME-TIFF uses the 2008-02 schema, so a ModuloT annotation is not present.

PR #1876, opened 2015-07-10
cmake: TemplateShellWrapper uses DYLD_FALLBACK_LIBRARY_PATH

rleigh-dundee:
Josh found last week while testing the superbuild code that DYLD_FALLBACK_LIBRARY_PATH should be used in place of DYLD_LIBRARY_PATH on MacOS X.  This PR changes this for the `bf-test` script template so that it can work in a superbuild context.

--------

Testing: `bf-test` will continue to work.  You can test with the zip build from any of the Mac build nodes. e.g. https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/BUILD_TYPE=Debug,CXXSTD_AUTODETECT=ON,node=mackerel/


joshmoore:
Built version runs fine on 10.10.  :thumbsup:

PR #1877, opened 2015-07-10
SDT: read MCS_TA data

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12918.

To test, use the files from QA 11109.  Both .sdt files should have 2 series, the first of which has 256 timepoints and the second of which has 1024 timepoints.  As noted in the ticket, the SPCM software can be used to double-check.

This shouldn't affect other .sdt files (i.e. builds should remain green), as the MCS_TA metadata blocks were introduced in newer versions of the .sdt specification.  The current SPC handbook and data file header have been uploaded to the ```specs/sdt``` directory for future reference.


imunro:
Number of timepoints confirmed to be 256 for series 1 and 1024 for series 1 for both files.
The images in  FLIMfit look similar to the screenshot provided and the decays look plausible.
Looks good to merge to me.

PR #1879, opened 2015-07-14
c++ docs: add `boost-devel` as a centos dependency

joshmoore:
Currently, all boost dependencies are missing from the
centos Quick Start. Without any boost, or with only
`boost` or `boost boost-regex` the following error
occurs:

```
centos6_1 | -- Performing Test OME_HAVE_BOOST_REGEX
centos6_1 | -- Performing Test OME_HAVE_BOOST_REGEX - Failed
centos6_1 | CMake Error at cpp/cmake/RegexChecks.cmake:84 (message):
centos6_1 |   No working regular expression implementation found
centos6_1 | Call Stack (most recent call first):
centos6_1 |   CMakeLists.txt:91 (include)
```

cc: @rleigh-dundee 

see: http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/cpp/overview.html#quick-start

### DIFF
--- a/docs/sphinx/developers/cpp/overview.txt
+++ b/docs/sphinx/developers/cpp/overview.txt
@@ -133,8 +133,10 @@ Debian/Ubuntu
 Partial quick starts
 ^^^^^^^^^^^^^^^^^^^^
 
-Homebrew and RedHat/CentOS do not provide packages for everything you need. The lists below will install
-*most* of the dependencies, but you will need to see the details below to install further dependencies.
+Homebrew and RedHat/CentOS do not provide packages for everything that is
+needed. The commands listed will install *most* of the dependencies, but
+further dependencies will need to be installed as described in various
+sections below.
 
 Homebrew
   ``brew install boost cmake hdf5 libpng python libtiff xerces-c git glm qt5 graphviz ant``
--- a/docs/sphinx/developers/cpp/overview.txt
+++ b/docs/sphinx/developers/cpp/overview.txt
@@ -132,7 +132,7 @@ Debian/Ubuntu
 Homebrew
   ``brew install boost cmake hdf5 libpng python libtiff xerces-c git glm qt5 graphviz ant``
 RedHat/CentOS
-  ``yum install libhdf5-devel libpng-devel python python-genshi libtiff-devel xerces-c-devel git gtest-devel graphviz java-1.7.0-openjdk``
+  ``yum install boost-devel libhdf5-devel libpng-devel python python-genshi libtiff-devel xerces-c-devel git gtest-devel graphviz java-1.7.0-openjdk``
 
 Note that Homebrew and RedHat/CentOS do not provide packages for everything you need; see below for details.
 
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
Note that the `boost-devel` in CentOS 6 is too old--it's less than our minimum requirement.  You need to build by hand or use our superbuild.  The `RedHat/CentOS do not provide packages for everything you need; see below for details.` sentence below was supposed to cover this part.

Note that the superbuild job is now working, so e.g.
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp-superbuild/BUILD_TYPE=Release,CXXSTD_AUTODETECT=OFF,node=gretzky35/lastSuccessfulBuild/artifact/install/bioformats-cpp-superbuild-5.1.2-242-g375acc2-Release-CentOS6.6-x86_64.tar.xz should be runnable on any CentOS (or indeed any later Linux distribution) such as Ubuntu 14.04.  I'd like to make these our downloadable release builds once they are tested, replacing what we currently have.

joshmoore:
@rleigh-dundee: all noted, but that doesn't change that the quick start can/should have `boost-devel` listed, right?

rleigh-dundee:
`boost-devel` is too old, and is unusable, on CentOS 6.  It may be different for CentOS 7.

joshmoore:
I'm testing on centos:7 now and will let you know. For centos:6, though, there should be *some* statement re: boost-devel in the quick start, because otherwise it's not a quick start; it's just confusing.

rleigh-dundee:
The quick start lists the packages you should install, and for CentOS it explicitly states that this isn't sufficient.  If we want to add some further detail here, that's fine, but that information is listed directly above and below this section.

joshmoore:
Perhaps adding links to the specifics would help. I'll look into that here.

At the moment, though, this feels pretty unusable as a consumer. I can't get centos6 or centos7 docker images to work due to the dependencies that have been added. That seems like a pretty bad place to be. I'd almost prefer to just say "unsupported" on centos until we have more time to invest. (Windows being the priority at the moment)

cc: @manics @sbesson

rleigh-dundee:
@joshmoore CentOS certainly isn't going to work "out of the box" like for Ubuntu.  Part of the consideration for getting the superbuild working on Unix as well as Windows was so that we could build a working build which would deploy there without any additional work, similar to how the ZeroC Ice builds will deploy on any Linux platform (within reason).  This will give us a "generic Linux" build which we can offer.

joshmoore:
Pushed. cc: @hflynn getting into wording / section issues.

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -129,12 +129,19 @@ BSD Ports
   ``pkg install devel/boost-all devel/cmake science/hdf5 graphics/png lang/python textproc/py-genshi graphics/tiff textproc/xerces-c3 devel/git devel/googletest math/glm devel/qt5 graphics/graphviz devel/apache-ant java/openjdk7 textproc/py-sphinx print/texlive-full``
 Debian/Ubuntu
   ``apt-get install build-essential libboost-all-dev cmake libhdf5-dev libpng12-dev python python-genshi libtiff5-dev libxerces-c-dev git libgtest-dev libglm-dev qt5-default libqt5-opengl5-dev libqt5-svg5-dev graphviz ant ant-contrib ant-optional openjdk-7-jdk openjdk-7-jre python-sphinx texlive-full``
+
+Partial quick starts
+^^^^^^^^^^^^^^^^^^^^
+
+Homebrew and RedHat/CentOS do not provide packages for everything you need. The lists below will install
```
hflynn: This 'lists below' followed by 'the details below' is somewhat confusing. How about:
"The commands listed will install *most* of the dependencies, but…"

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -129,12 +129,19 @@ BSD Ports
...
+*most* of the dependencies, but you will need to see the details below to install further dependencies.
```
rleigh-dundee: I like the intent of all the changes, but I do dislike the unnecessary use of personal pronouns such as `you` in documentation.

hflynn:
Rebuilding on staging now to look at sectioning/formatting.

docs/sphinx/developers/cpp/overview.txt
joshmoore: I was copying the use of "your" from https://github.com/openmicroscopy/bioformats/pull/1879/files#diff-001920d55adf9e9dc69e9c467845094dR126 -- happy to use whatever style, really.

hflynn:
Travis needs a kick

mtbc:
Travis now seems happy.

joshmoore:
Pushed.

rleigh-dundee:
Looks fine to me.

PR #1880, opened 2015-07-14
FluoviewReader: TimeStamp - Accommodate 24 hour time format as well

bramalingam:
ref: https://trac.openmicroscopy.org/ome/ticket/12964

Testing instructions : 

Please import the 2 files in the following thread in ImageJ using the Bio-Formats Importer,
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-July/005492.html

with the "display-metadata" option checked.

And check if you can see the TimeStamp information in the metadata table in both cases,

ex: Timestamp for Z= *s, C=*s, T=*s

Without this PR : The Old file (FURA.tif) will display the Timestamp but the new file will not show the TimeStamp, in the metadata table.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
@@ -799,7 +799,7 @@ public class FluoviewReader extends BaseTiffReader {
       }
       if (date != null) {
         date = DateTools.formatDate(date.trim(),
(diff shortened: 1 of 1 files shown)


melissalinkert:
Excluding as discussed to avoid merge build confusion; fine to re-include as soon as 5.1.3(-rc1) is tagged.

JimboMahoney:
Many thanks for the speedy fix! (I'm the original reporter of this issue).

melissalinkert:
Looks good; merging.

PR #1885, opened 2015-07-16
docs: Additional C++ superbuild documentation

rleigh-dundee:
- Add explanatory paragraph to explain the difference between a normal and super build
- Add to table of dependencies
- Add source cache description


docs/sphinx/developers/cpp/overview.txt
```diff
@@ -48,52 +48,63 @@ In order to build the C++ library and its documentation, a number of
...
-+-----------------+--------------+--------------+----------+--------------+--------+
-| Graphviz†       | 2.x          | 1.8.10       |    ◦     |              |        |
-+-----------------+--------------+--------------+----------+--------------+--------+
-| Python Sphinx‡§ | 1.2.x        | 1.1.x        |    ◦     |              |        |
-+-----------------+--------------+--------------+----------+--------------+--------+
-| TeX (XeLaTeX)§  | TeXLive 2014 | TeXLive 2012 |    ◦     |              |        |
-+-----------------+--------------+--------------+----------+--------------+--------+
+Bio-Formats (*BF [super]build*).  A subset of these are required for
+building client applications making use of Bio-Formats (*Client
+build*) For end-user deployment (*Deploy*), the library packages
```
mtbc: needs period before "For"?

docs/sphinx/developers/cpp/overview.txt
```diff
@@ -679,7 +690,7 @@ instead, if desired.
 
 Start by creating a temporary build directory.  This directory may be
 in any location inside or outside the Bio-Formats source tree.
-However, you may not use the source directory as the build directory.
+However, the source directory can not be used as the build directory.
```
mtbc: cannot

mtbc:
Staged at http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/cpp/overview.html.

joshmoore:
Various things at the readability level:

 * first, I like the new prereq table. Thank you!
 * took me a second to understand the italics markup of the column names. Don't know if quotes would help. *shrugs*
 * Could the symbols `*`, `†`, `‡`, and `§` be inlined along with the dots? Does that work semantically?

Still in the process of testing the actual settings via docker.

rleigh-dundee:
WRT the italic markup, that's purely presentational; the table markup is correct Sphinx markup.  If we want to improve the table appearance then I *think* that's down to the "plonematch" CSS we use (which is already annoyingly broken in several respects).

I'm not sure what you mean by inlining the symbols; Sphinx RST does not support table footnotes, so these are a best-effort approximation using a description list.

joshmoore:
> WRT the italic markup, that's purely presentational...

Well, sure. I'm just telling you as a reader/user/consumer, that it took me a bit to understand what you were getting at.

> I'm not sure what you mean by inlining the symbols ...

I mean have those symbols replace the dots. I.e. why the two strategies: some in the column; some by the terms?

hflynn:
@joshmoore is right Roger, that table is really confusing. Would it not be clearer to just write out 'Required for…' or 'Optional for…' in one final column rather than having 4 columns with symbols that still don't cover all the info?

rleigh-dundee:
Ah, I see what you're getting at now.  I can change that.

rleigh-dundee:
The table should be simpler now.

hflynn:
Yep, table looks better

rleigh-dundee:
Could this be merged now, please.

PR #1888, opened 2015-07-17
CellomicsReader:Fix for sparsely populated fields

bramalingam:
ref: https://trac.openmicroscopy.org/ome/ticket/12967

Testing instructions : 
See fileset uploaded in ​https://www.openmicroscopy.org/qa2/qa2/qa/feedback/11308/
Run:
showinf -nopix -omexml MFGTMP_150418130002_A01f111d0.C01 

do the same for files under: /Volumes/ome/data_repo/from_skyking/cellomics/qa-7352/test/
Run:
showinf -nopix -omexml /Volumes/ome/data_repo/from_skyking/cellomics/qa-7352/test/MFGTMP-PC_130509120002_B02f00d0.C01 

This should not throw any exceptions.
Without this PR: throws an IndexOutOfBoundsException.

Extra Check:
And Try the files on ImageJ as well! Opening just one series during the import.
Check if both the filesets open without any exceptions.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
@@ -373,21 +373,17 @@ public class CellomicsReader extends FormatReader {
       if (row < realRows && col < realCols) {
         wellIndex = row * realCols + col;
 
-        if (i==0){
-            wellIndexPrev = wellIndex;
-        }
-
-        if (wellIndexPrev != wellIndex){
+        if ((wellIndexPrev != wellIndex) || i==0){
+          if(i>0){
             wellCntr++;
-        }
-
-        if (wellIndexPrev != wellIndex){
+            fieldCntr = 0;
+          }else{
+            wellIndexPrev = wellIndex;
+          }
 
           store.setWellID(MetadataTools.createLSID("Well", 0, wellIndex), 0, wellCntr);
           store.setWellRow(new NonNegativeInteger(row), 0, wellCntr);
           store.setWellColumn(new NonNegativeInteger(col), 0, wellCntr);
-          fieldCntr = 0;
-
         }
 
         if (files.length == 1) {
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/CellomicsReader.java
```diff
@@ -215,6 +215,36 @@ protected void initFile(String id) throws FormatException, IOException {
...
+            String field2 = String.valueOf(field2_2);
+            if (field2_2<10) {
+                field2 = "00" + field2;
+            }else if (field2_2<100) {
+                field2 = "0" + field2;
+            }
+
+            int fileId1 = Integer.valueOf(String.valueOf((wellRow1 * wellColumnsFinal) + wellCol1) + field1);
+            int fileId2 = Integer.valueOf(String.valueOf((wellRow2 * wellColumnsFinal) + wellCol2) + field2);
+            return fileId1 - fileId2;
```
melissalinkert: It shouldn't be necessary to convert the well and field to a new string for comparison.  The values of ```wellRow*```, ```wellCol*``` and ```field*_*``` can be compared directly, similar to https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/ScanrReader.java#L460.  This would have the added advantage of not needing to be updated if we encounter a dataset with more than 1000 fields.

sbesson:
See also FindBugs warnings related to this PR in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/955/findbugsResult/ and Cellomics automated test failures in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/551/.

mtbc:
This works well. There just seems to be stuff in well A2, right? Insight won't allow selection of the other wells, but valid OME-XML can be generated for the various `C01` files including that named in the PR description and the plate can now be imported into OMERO (it failed before). Seems to me to be good to merge.

bramalingam:
@mtbc : Yes this plate had just one well (A2 alone).

melissalinkert:
Just holding off until we have confirmation that https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/ is green again - build 553 (the last one finished as of writing) still showed some failures on existing Cellomics datasets.

sbesson:
See failing Cellomics datasets on https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/554/ as well.

melissalinkert:
Excluding for now, so that builds can go back to a green state.  @bramalingam, happy to discuss if needed, otherwise just re-list when you think it's resolved and we'll remove the exclude label.

bramalingam:
The last commit should fix the errors noticed in the builds. @jburel : Can you please remove the exclude label on this PR?

sbesson:
All 11 failures in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/568/ are related to cellomics datasets. Excluding.

melissalinkert:
XML population changes look good to me.

The 8 MD5 failures are a legitimate issue - there is one more part of the filename that needs to be compared around line 241, which is easier to see (and test) if you add print out the file name in openBytes.  Once that fix is pushed, should be fine to remove the breaking label.

melissalinkert:
Breaking label removed, see https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-breaking-repository-subset/64/.

PR #1890, opened 2015-07-17
cmake: External boost: Use local mirror

rleigh-dundee:
SourceForge seem to be breaking their download URLs this week;
switch to a local mirror instead.

--------

Testing: https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp-superbuild/ should go green again, since this broken due to SourceForge breaking their URLs.

### DIFF
--- a/cpp/cmake/External_boost.cmake
+++ b/cpp/cmake/External_boost.cmake
@@ -41,7 +41,7 @@ if(NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
   ExternalProject_Add(${proj}
     ${BIOFORMATS_EP_COMMON_ARGS}
-    URL "http://sourceforge.net/projects/boost/files/boost/1.58.0/boost_1_58_0.tar.bz2"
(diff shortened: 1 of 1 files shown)


joshmoore:
Testing the URL it seems fine, but personally, this seems like an odd long-term change to me. Homebrew uses "https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2" as well.

rleigh-dundee:
I don't mind either way.  It was a reaction to the SF breakage and downtime breaking FLIMfit and bfcpp last week.  I think it's a little more sociable to use a mirror rather than the primary source, and my trust in sourceforge isn't high since they started bundling malware and making the download pages a sea of clickbait links to adware and other nasties.  We were already using non-SF links for zlib and bzip2 (heanet.ie; they could also be switched to mirrorservice)

joshmoore:
Talking to @rleigh-dundee, things should now be working. There's some merit in thinking about dropping SF completely, but not using their CDN seems like a single-point-of-failure. Closing for the moment.

PR #1891, opened 2015-07-19
CMake 3.2 minimum version

rleigh-dundee:
For 5.1.4.  To address https://trello.com/c/kJGhA6QU/21-cmake-minimum-version-3-2 and https://trello.com/c/GIgTdQXh/12-travis-cmake-2-8 (if I guessed what it is correctly).

As the card states, the primary purpose is not so we can use 3.2-specific features, though there are a number we would be able to make use of as followup work.  It's more to reduce the combinatorial complexity of our testing matrix to something manageable.  Since we aren't actively testing with older CMake versions, the chances of regressions or accidental use of 3.0+ features breaking 2.8 is high if we aren't testing properly.  While Travis was using 2.8, this only tests a single and limited configuration and is not going to be representative of wider use.

In one place in the source we currently have separate codepaths for 2.8 and 3.0+ (interface libraries for ome-compat).  This is one part I'd like to simplify.

--------

Testing: jobs should remain green.  It will no longer build with CMake versions older than 3.2.


joshmoore:
NB: I'd like to see what this does to https://github.com/joshmoore/cppspace before merging.

mtbc:
Note that Debian 8 (jessie) comes with cmake 3.0.

joshmoore:
Thanks for the heads up, @mtbc. How substantially different would the content of this PR be if the min. was set to 3.0, @rleigh-dundee ?

rleigh-dundee:
@joshmoore To set the minimum to 3.0, we would need to change:

- Minimum version and policy version to 3.0.0
- Revert some of the policy CMPXXX removals to keep the ones added after 3.0.0 and retain the existing behaviour
- Update the versions in the documentation

*However*, do note that
- the codebase hasn't been tested with 3.0 for a good while (at least three months)
- this applies to the normal build only; the superbuild has never been tested with anything below 3.2.
- while 3.0 is an improvement over 2.8.12 as a baseline, we do not have *any* Travis or Jenkins CI testing of this version (or of 3.1)

None of the CI nodes provide either 3.0 or 3.1, and in terms of support by distributions:

- mac nodes are all 3.2 (3.3 in homebrew already, so will be 3.3 on next update)
- bsd node is 3.2 (soon to be 3.3 also)
- cod (Ubuntu 14.04, provides 2.8.12; currently using 3.2 from /opt)
- gretzky35 (CentOS 6, provides 2.8.12--was previously very old; currently using 3.2 from /opt, was using 2.8.12 from /opt)
- [CentOS 7 provides 2.8.11 (older than CentOS6 !)]
- maxquant  Windows node is 3.2

As you can see, 3.0 and 3.1 don't actually feature in any LTS releases and on all other systems 3.2+ is already in use.  So there isn't really much downside to making 3.2 the minimum here--in practice if we're dropping 2.8.x support, 3.2+ is what will be in the next set of LTS releases.  3.0 and 3.1 just haven't seen that much use, being released as the previous set of LTS releases came out it missed being included.

mtbc:
As another random data point, I just checked on my OpenBSD 5.7 system which cmake binary build is offered for that: it's 3.1.3. Still, it must mean something that all our CI already has 3.2!

(Obviously when I've tested @rleigh-dundee's C++ PRs, including superbuild, in recent months it's been with cmake 3.0.)

joshmoore:
The change needed for https://github.com/openmicroscopy/cppspace looks like:

```
diff --git a/centos6/Dockerfile b/centos6/Dockerfile
index cc6bfe3..c37c83b 100644
--- a/centos6/Dockerfile
+++ b/centos6/Dockerfile
@@ -20,6 +20,15 @@ RUN yum -y install tar && \
     cd .. && \
     rm -rf boost_1_57_0*

+RUN yum -y install wget && \
+    yum -y remove cmake && \
+    wget http://www.cmake.org/files/v3.2/cmake-3.2.3.tar.gz && \
+    tar -zxvf cmake-3.2.3.tar.gz && \
+    cd cmake-3.2.3 && \
+    ./bootstrap && \
+    gmake && \
+    gmake install
+
 RUN adduser build
 USER build
 WORKDIR /home/build
diff --git a/centos7/Dockerfile b/centos7/Dockerfile
index a4c6406..f57dc8b 100644
--- a/centos7/Dockerfile
+++ b/centos7/Dockerfile
@@ -4,9 +4,9 @@ RUN yum -y install gcc gcc-c++ make cmake git
 RUN yum -y install libhdf5-devel libpng-devel python python-genshi libtiff-devel xerces-c-devel git gtest-devel graphviz java-1.7.0-openjdk
 RUN yum -y install boost-devel
 RUN yum -y install wget && \
-    wget http://www.cmake.org/files/v3.0/cmake-3.0.0.tar.gz && \
-    tar -zxvf cmake-3.0.0.tar.gz && \
-    cd cmake-3.0.0 && \
+    wget http://www.cmake.org/files/v3.2/cmake-3.2.3.tar.gz && \
+    tar -zxvf cmake-3.2.3.tar.gz && \
+    cd cmake-3.2.3 && \
     ./bootstrap && \
     gmake && \
     gmake install
```

and now I'm in the process of adding the same thing to the image I found for testing gh-1896: https://registry.hub.docker.com/u/library/gcc/ (wheezy)

Personally, I fine the fact that I can't fine any image that works with this PR out of the box annoying and likely indicative of the fact that others are going to be put through the same paces. If there's no blocker with CMake 3.0, I'd argue for at least dropping back to it. If our CI systems aren't testing it, then that's a problem we can handle differently without placing the burden on users.

rleigh-dundee:
Note that we do require 3.2 for the superbuild which is why that's what is on the CI nodes.

mtbc:
Ah, yes, now I notice that I did some superbuild testing in a chroot that actually has cmake 3.2 installed. (Is 3.2 really needed?)

rleigh-dundee:
@mtbc Unknown. It's possible this is the case but without getting cmake 3.0 and downgrading the policies to test, we simply don't know.  And that's really the point behind this--to reduce the supported major versions from (2.8 3.0 3.1 3.2 3.3) to (3.2 3.3) to keep the supported set manageable.

I should also make the distinction between
- regular builds
- superbuilds

Regular builds currently have 2.8 as the baseline; this can be upgraded to 3.0 or 3.2, and while this PR deliberately doesn't actually enable use of new features like interface libraries that's something I'd do as a followup

Superbuilds have only every been tested on 3.2.  We can certainly downgrade and test on 3.0, but I'm not sure quite who this would benefit.  It's certainly not applicable to the platforms already shipping 3.2 or the LTS Linux distributions we officially support.

mtbc:
What do we officially support? For instance, our version requirements page lists Debian 8 as supported for 5.2 but that comes with cmake 3.0.2.

rleigh-dundee:
@mtbc This is somewhat unclear.  I thought that our primary supported platforms were the CentOS and Ubuntu LTS releases, with everything else being on a best-effort basis.  Certainly on the CI side, we're limited to these.  The others I listed on the version requirements page were more informational; while we say certain versions are recommended or supported, quite how that compares to the others is not clear; this is probably something we should clarify.

mtbc:
I wonder if Docker would help to minimize the coverage effort here.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-5.1-merge-push#79](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-push/79/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-push/79/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - .travis.yml

joshmoore:
Closing for the moment; see https://trello.com/c/kJGhA6QU/345-cmake-minimum-version-3-2

PR #1892, opened 2015-07-20
Submodule cleanup

sbesson:
Following the release of Bio-Formats 5.1.3 and https://github.com/openmicroscopy/openmicroscopy/pull/393, this PR cleans up many items of the Bio-Formats source code repository related to the OMERO/Bio-Formats submodule compatibility.

This PR should not modify any build target and all CI jobs should remain green.


joshmoore:
For the XML bits, this certainly likes a clean deletion of what I added previously to make the submodule work. We also have jobs which test the source zips, no? If so, I would think this looks good.

sbesson:
@rleigh-dundee may have more insight on it but I thought the cpp jobs were consuming the source zips.

rleigh-dundee:
@sbesson This is certainly the case for bioformats, but looking at the merge and latest cpp jobs for OMERO, they are all currently using git.  I'd suggest switching one of them such as merge-cpp-cmake to use the source zip.

sbesson:
Agreed. But this is outside the scope of this PR which is impacting the Bio-Formats source code zips right?

rleigh-dundee:
@sbesson Yes.  For this job it should get tested by the cpp jobs.  I don't think we have any java jobs using the source zips, but ideally we should be testing that as well.

melissalinkert:
This is good to merge then, yes?

rleigh-dundee:
Fine with me.

PR #1894, opened 2015-07-22
Fix config directory path substitution on Windows

melissalinkert:
```replaceAll``` on Windows path names can quickly become kind of weird due to backslashes.  This replaces ```replaceAll``` with a less platform-dependent method of replacing the directory names.

Locally, I ran ```ant "-Dtestng.configDirectory=C:\Users\melissa\data_repo_config\test_images_good" "-Dtestng.directory=E:\test_images_good" test-automated``` (pointing to the current master branch of data_repo_config and a current copy of test_images_good respectively).  Without this change, every file in test_images_good is logged as "not configured".  With this change, only those files whose parent directory is symlinked (so no corresponding file is in data_repo_config) are logged as "not configured".

/cc @sbesson
See also https://trello.com/c/tbUfkJiE/6-external-configuration-directory-follow-up

### DIFF
--- a/components/test-suite/src/loci/tests/testng/ConfigurationTree.java
+++ b/components/test-suite/src/loci/tests/testng/ConfigurationTree.java
@@ -40,6 +40,7 @@ import loci.common.Constants;
 import loci.common.IniList;
 import loci.common.IniParser;
 import loci.common.IniTable;
+import loci.common.Location;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sbesson** [comment]
Generally looks sensible and I'll have the CI jobs running using this PR soon hopefully. In terms of refactoring, as I was also modifying the relocation logic to resolve symlink resolution, I wondered if it would make sense to migrate this logic into the `Configuration` class and use something like:

```
config.convertRepoToConfig(repoRoot);
config.convertConfigToRepo(configRoot);
```

IF that makes sense, I am happy to open a follow-up PR to achive this once we validate this Window fix.

melissalinkert:
Definitely sounds fine - I was focusing on just getting something working, not making it pretty :-)

sbesson:
https://ci.openmicroscopy.org/view/Breaking/job/BIOFORMATS-5.1-breaking-test_images_good-win/8/ being green, I propose we get this merged and work on refactoring as part of the next improvement (symlink support).

PR #1895, opened 2015-07-23
DicomReader/OBFReader: use native units for storing the physical size

sbesson:
See https://trello.com/c/nO35TKXK/4-units-and-readers

This PR:
- modifies the `DICOMReader` and `OBFReader` to use the native units while storing the physical size
- implements `FormatTools.getPhysicalSizeZ(Double, String)` to restore symmetry
- fixes the automated tests to account for rounding errors during unit conversions

/cc @bjoernthiel


components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1430,9 +1430,20 @@ public static Length getPhysicalSizeY(Double value, String unit) {
   }
 
   public static Length getPhysicalSizeZ(Double value) {
+    return getPhysicalSizeZ(value, null);
+  }
+
+  public static Length getPhysicalSizeZ(Double value, String unit) {
```
joshmoore: Could you add here and above javadoc that if `unit == nulll`, `MICROM` will be used?

components/formats-bsd/src/loci/formats/in/DicomReader.java
```diff
@@ -763,24 +763,22 @@ else if (isJPEG || isJP2K) {
...
           if (x != null) {
             store.setPixelsPhysicalSizeX(x, i);
           }
         }
         if (pixelSizeY != null) {
-          Double sizeY = new Double(pixelSizeY);
-          sizeY *= 1000;
-          Length y = FormatTools.getPhysicalSizeY(sizeY);
+          Double physicalSizeY = new Double(pixelSizeY);
+          Length y = FormatTools.getPhysicalSizeY(physicalSizeY, "mm");
```
joshmoore: Using the enums throughout the code (rather than string representations like `"mm"`) will make later searching/refactoring easier if it doesn't make the code too unwieldy now.

sbesson: So would that mean implementing helper methods a la `FormatTools.getPhysicalSize{X,Y,Z}(String, Unit);`? Or use `new Length(Double, Unit)`?

components/formats-api/src/loci/formats/FormatTools.java
sbesson: Definitely. Will add some javadocs to these methods as part of this PR.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1387,53 +1387,166 @@ public static void convert(IFormatReader input, IFormatWriter output,
...
+   * @param value  the value of the physical size in Y
+   * @param unit   the unit of the physical size in Y
+   *
+   * @return       the physical size formatted as a {@link Length}
+   */
+  public static Length getPhysicalSizeY(Double value, ome.units.unit.Unit<ome.units.quantity.Length> unit) {
+    if (value != null && value - Constants.EPSILON > 0 &&
+      value < Double.POSITIVE_INFINITY)
+    {
+      createLength(value, unit);
```
joshmoore: This line is missing a `return`! Maybe time to add unit tests for these methods?

sbesson:
Last set of commits should implement some proper unit tests for all variations of the `FormatTools.getPhysicalSize{X,Y,Z}` methods and unify them with the tests previously living under formats-bsd.

components/test-suite/src/loci/tests/testng/FormatReaderTest.java
```diff
@@ -963,10 +965,11 @@ public void testPhysicalSizeZ() {
...
       }
       Length realSize = retrieve.getPixelsPhysicalSizeZ(i);
-      Number size = realSize == null ? null : realSize.value(UNITS.MICROM);
+      Double size = realSize == null ? null : realSize.value(UNITS.MICROM).doubleValue();
 
       if (!(expectedSize == null && size == null) &&
-        (expectedSize == null || !expectedSize.equals(size)))
+        (expectedSize == null || (
+          Math.abs(size - expectedSize) > Constants.EPSILON)))
       {
```
melissalinkert: I think there might need to be one additional set of null checks for each of these three cases.  ```realSize.value(UNITS.MICROM)``` can be null even if ```realSize``` is non-null; I suspect this is what's behind the 10 unlogged failures in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/558/ (2 in the ```imspector``` directory, not sure yet which directory the other 8 are from).

sbesson:
https://github.com/openmicroscopy/bioformats/commit/6c1cb892aeb869e3d2e23e57ea4416ec74d17e20 should fix the FAILURE that were not displayed anymore in the automated-test output.

The reason of the Imspector files failure is that the `Math.abs(size - expectedSize) > Constants.EPSILON))` condition is not satisfied for one dataset which corresponds to the special case where the length in meters was converted into microns. For now, I would  propose to revert to the initial behavior and multiple the original length.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1387,56 +1387,174 @@ public static void convert(IFormatReader input, IFormatWriter output,
...
+      LOGGER.debug("Expected positive value for PhysicalSizeX; got {}", value);
+      return null;
     }
-    LOGGER.debug("Expected positive value for PhysicalSizeX; got {}", value);
-    return null;
   }
-  
+
+  /**
+   * Formats the input value for the physical size in Y into a length of the
```
mtbc: X?

melissalinkert:
This reads fine to me now - just waiting on https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/562/.

sbesson:
See https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/566/

PR #1896, opened 2015-07-26
cmake: Check for broken GCC regex implementation

rleigh-dundee:
This comes from https://github.com/codelibre-net/schroot/pull/2/files and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=778112 leading to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67015

There's a subly broken regex implementation in GCC, which we weren't detecting.  This PR detects this and will fall back to using boost::regex in this case.

--------

Testing: Travis and the CI builds won't detect this since we aren't using any of the affected compiler versions.  Ideally we would have builds on more recent GCC versions, but we don't at present.  If you build with GCC 4.9.x, 5.0 or 5.1, you will find that where before the `OME_HAVE_REGEX` cmake test would succeed, it will now fail:

- run `cmake /path/to/bioformats` and note OME_HAVE_REGEX and OME_HAVE_BOOST_REGEX results
- merge this branch and repeat.  In the latter test you should see OME_HAVE_REGEX fail and OME_HAVE_BOOST_REGEX succeed, meaning it will fall back to using Boost's regex implementation.

### DIFF
--- a/cpp/cmake/RegexChecks.cmake
+++ b/cpp/cmake/RegexChecks.cmake
@@ -65,6 +65,16 @@ int main() {
   if (${namespace}_match(fail, bar)) return 11;
   if (${namespace}_match(fail, chk)) return 12;
 
+  // Checks for broken support in GCC 4.9 and 5.1
+  ${namespace} range1(\"^[a-z0-9][a-z0-9-]*\$\", ${namespace}::extended);
+  ${namespace} range2(\"^[a-z0-9][-a-z0-9]*\$\", ${namespace}::extended);
+  if (!${namespace}_match(test, range1)) return 13;
+  if (!${namespace}_match(test, range2)) return 14;
+  if (!${namespace}_match(\"a-\", range1)) return 15;
+  if (!${namespace}_match(\"a-\", range2)) return 16;
+  if (${namespace}_match(\"-a\", range1)) return 17;
+  if (${namespace}_match(\"-a\", range2)) return 18;
+
   return 0;
 }"
 ${outvar})


mtbc:
With gcc 4.9.2, with or without this PR,
```
OME_HAVE_BOOST_REGEX:INTERNAL=1
OME_HAVE_BOOST_REGEX_COMPILED:INTERNAL=TRUE
OME_HAVE_BOOST_REGEX_EXITCODE:INTERNAL=0
OME_HAVE_REGEX:INTERNAL=
OME_HAVE_REGEX_COMPILED:INTERNAL=FALSE
OME_HAVE_TR1_REGEX:INTERNAL=
OME_HAVE_TR1_REGEX_COMPILED:INTERNAL=FALSE
```
This PR doesn't make any difference to `CMakeCache.txt` for me.

rleigh-dundee:
@mtbc This is because it failed in both cases, i.e. the compiler you are using was already found to be broken.  That's the `OME_HAVE_REGEX:INTERNAL=` line having a blank value.  If you look at the cmake output you'll see `-- Performing Test OME_HAVE_REGED - Failed`.

Basically you'll need GCC 5.0/5.1/4.9.3 to test the PR's failure mode.  But if you look at https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/BUILD_TYPE=Release,CXXSTD_AUTODETECT=ON,node=perch/886/consoleText which uses clang you'll see the tests pass correctly here since the compiler and standard library aren't broken.

joshmoore:
(Note: don't have a working testing environment. Happy to keep trying)

rleigh-dundee:
To be honest, it doesn't really matter to test the failure.  It's just two trivial regexes, and it's passing everywhere it's not broken, so it's obviously not detrimental.  And I have tested it myself with those specific compilers to provide it does fail here.

joshmoore:
Works fine with https://github.com/openmicroscopy/cppspace/commit/f65ae5edcc7a41f8bcd91c36020057d36503cf00 :

```
wheezygcc493_1 | -- Performing Test OME_HAVE_REGEX
wheezygcc493_1 | -- Performing Test OME_HAVE_REGEX - Failed
wheezygcc493_1 | -- Performing Test OME_HAVE_TR1_REGEX
wheezygcc493_1 | -- Performing Test OME_HAVE_TR1_REGEX - Failed
wheezygcc493_1 | -- Performing Test OME_HAVE_BOOST_REGEX
wheezygcc493_1 | -- Performing Test OME_HAVE_BOOST_REGEX - Success
```

PR #1897, opened 2015-07-27
cpp: ome-xml: Use pImpl idiom in OMEXMLModelObject classes

rleigh-dundee:
Converts the OME-XML model objects to use the pImpl data-hiding idiom to avoid the exposure of internal implementation details in our public headers.  This is for current and future Windows porting efforts to reduce the number of exported templates in DLLs.

Note: this increases the number of memory allocations and may have a performance impact.  Check the performance of the merge build tests with the previous builds to make sure it hasn't regressed significantly.  It uses `shared_ptr` to own the `Impl` object, but could use a bare pointer with a `delete` in the destructor.  This isn't done since this way it's easily switched for `unique_ptr` down the line which will have other bigger performance benefits such as being able to emplace objects and avoid a lot of overhead in copying and allocating model objects.

Marking breaking in case this causes problems, though seems OK locally.

--------

Testing: All jobs should continue to be green; code changes should be covered by unit tests.  The main thing to review here are the generated code changes to make sure it's not changed any behaviour.

### DIFF
--- a/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelObject.template
@@ -96,8 +96,6 @@ namespace ome
       class ${f};
 {% end for %}\
 
-{% end header %}\
-{% if fu.SOURCE_TYPE == "header" %}\
       /**
        * ${klass.name} model object.
        */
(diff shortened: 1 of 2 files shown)


mtbc:
Travis failed.

joshmoore:
Travis exceeded 50 minutes; restarted.

joshmoore:
The overall build times don't look significantly longer. Is there somewhere you look for a summary of the unit test times?

joshmoore:
OME-XML diffs are empty.

joshmoore:
Code changes seem to make sense and I've compiled on several different platforms. I don't guess the performance hit will be significant, but there's also the added complexity. Has this helped with the symbol export issue on Windows?

rleigh-dundee:
With using static libs for the moment, this won't have a significant effect upon Windows build errors, it's more pre-emptive for making DLLs down the line.

https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/877/BUILD_TYPE=Release,CXXSTD_AUTODETECT=OFF,node=cod/consoleText
vs
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/BUILD_TYPE=Release,CXXSTD_AUTODETECT=OFF,node=cod/889/consoleText

24: [----------] 108 tests from ModelVariants/ModelTest (5913 ms total)
24: [----------] 108 tests from ModelVariants/ModelTest (5954 ms total)

Not a particularly huge difference and without looking at multiple jobs to see what the variance is it may be completely insignificant.

Regarding the diff this is most likely a problem with the merge job which will need looking at.  If you want to diff it yourself you just need to diff the content of cpp/lib/ome/xml in the build dir before and after the merge, or I can do it and attach it here.

jburel:
@rleigh-dundee: removed "breaking" from description so it will be included in tomorrow's build

PR #1898, opened 2015-07-27
cpp: ome-xml: Correct return in Timestamp ostream output

rleigh-dundee:
Move return to the outermost block, to fix a compiler warning.

--------

Testing: builds should remain green, no manual testing required.


mtbc:
Good to merge.

PR #1899, opened 2015-07-28
Fix .dll packaging in formats-gpl

melissalinkert:
See https://trello.com/c/ukrm7pbD/25-make-sure-dlls-end-up-in-formats-gpl-jar.

To test, check the output of ```jar tf artifacts/formats-gpl.jar``` and ```jar tf artifacts/bioformats_package.jar``` without this change.  Verify that no .dll files are present in formats-gpl.jar, and the only .dll files present in bioformats_package.jar correspond to Turbojpeg.

The same commands with this change should show two .dll files in the META-INF directory of formats-gpl.jar; the same two files should be present in bioformats_package.jar.

### DIFF
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -257,7 +257,9 @@ your FindBugs installation's lib directory. E.g.:
         include (none) to prevent problems if component.resources-bin is empty
       -->
       <fileset dir="${src.dir}" includes="(none) ${component.resources-bin}"/>
-      <fileset dir="${src.dir}/.." includes="(none) ${component.resources-bin}"/>
+    </copy>
+    <copy todir="${dest.dir}/META-INF" preservelastmodified="true">
+      <fileset dir="${src.dir}/.." includes="(non) ${component.resources-bin}"/>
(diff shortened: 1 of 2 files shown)


sbesson:
Looks sensible. As discussed with @melissalinkert, this causes https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-test_images_good-win/ to fail. We will likely need to manage multiple configurations for the Slidebook reader depending on the platform. Once https://trello.com/c/tbUfkJiE/6-external-configuration-directory-follow-up is properly resolved, this will be easier to set up. Excluding for now.

joshmoore:
Or alternatively not ship the DLLs with Bio-Formats?

sbesson:
As https://github.com/openmicroscopy/bioformats/pull/1951 seems to be working as expected, unexcluding this PR.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/599/

PR #1820, opened 2015-05-27
xsd-fu: Add javadocs to Java metadata templates

rleigh-dundee:
Generated MetadataStore, MetadataRetrieve and implementing classes now have a complete set of Javadoc comments.  These have been copied and adapted from doxygen comments in the C++ templates.

--------

Testing:

See 
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/javadoc/ome/xml/meta/package-summary.html
Interfaces: MetadataStore, MetadataRetrieve, OMEXMLMetadata
Classes: AggregateMetadata, DummyMetadata, FilterMetadata, OMEXMLMetadataImpl

They should all have a complete set of javadocs.

PR #1900, opened 2015-07-29
xsd-fu: MetadataRetrieve: Drop 'unused' from used parameters

rleigh-dundee:
This is purely a docstring change, though you can verify the generated code to check that the parameters are actually used (both for Java and C++).

A perhaps bigger question is why for example `setImageAnnotationRef` and `getImageAnnotationRef` do not have symmetric signatures, or rather do but in the former annotationRefIndex is unused and in the latter it is used, so the behaviour is not the same.  This applies to all `(get|set).*AnnotationRef` methods.  What makes these different than every other method in the model using indexes?  Even if this is valid, it would be helpful to know why and document it.

### DIFF
--- a/components/xsd-fu/templates-cpp/MetadataRetrieve.template
+++ b/components/xsd-fu/templates-cpp/MetadataRetrieve.template
@@ -17,7 +17,7 @@
 {% for param in indexes %}\
          * @param ${param['argname']} the ${param['type']} index.
 {% end for %}\
-         * @param ${lang.index_signature(prop.name, 0, 0)['argname']} ${prop.name} index (unused).
+         * @param ${lang.index_signature(prop.name, 0, 0)['argname']} ${prop.name} index.
          * @returns the ${prop.name} property.
          */
         virtual ${prop.metadataStoreRetType}
--- a/components/xsd-fu/templates-java/MetadataRetrieve.template
+++ b/components/xsd-fu/templates-java/MetadataRetrieve.template
@@ -19,7 +19,7 @@
 {% for param in indexes %}\
          * @param ${param['argname']} the ${param['type']} index.
 {% end for %}\
-         * @param ${lang.index_signature(prop.name, 0, 0)['argname']} ${prop.name} index (unused).
+         * @param ${lang.index_signature(prop.name, 0, 0)['argname']} ${prop.name} index.
          * @return the ${prop.name} property.
          */
 	${prop.metadataStoreRetType} get{% if is_multi_path[o.name] %}${parent}{% end %}${obj.name}${prop.name}(${indexes_string(indexes)}, ${index_string(prop.name)});\


mtbc:
I wonder who would know the answer to that question. @joshmoore or @chris-allan perhaps?

chris-allan:
I'd need some more contextual examples to comment further. The whole XsdFu stack as it stands bears very little resemblance to previous implementations that I worked on.

This single word reference to something being "unused" could easily have been a typo, copy and paste error, refer to the fact that Bio-Formats readers never called those methods (which they still don't) or that `MetadataConverter` methods had not been written to account for them.

The line in question was added by @rleigh-dundee in 0cc119d1f6cfa964211583e8a5246f8041d08384 when he was adding Javadoc as part of #1820. I'd expect him to be the authority as to the source of "unused".

mtbc:
For instance, in OMERO `OMEROMetadataStoreClient` ignores them, as in
https://github.com/openmicroscopy/openmicroscopy/blob/5dd4367b/components/blitz/src/ome/formats/OMEROMetadataStoreClient.java#L4191, but `OmeroMetadata` uses them, as in https://github.com/openmicroscopy/openmicroscopy/blob/5dd4367b/components/blitz/src/ome/services/blitz/impl/OmeroMetadata.java#L428.

chris-allan:
The importer does not guarantee the orderedness of the references under those conditions. Everything is implicitly ordered in XML. Not every collection in OMERO is ordered and references supplied at import time are not guaranteed to be ordered. Hence those indexes are effectively ignored by the importer.

These semantics are identical across all reference handling methods in `OMEROMetadataStoreClient` irrespective of if they apply to annotations are not. `setImageMicrobeamManipulationRef()` and `setImageROIRef()` are some examples of such methods.

A getter such as `getImageAnnotationRef()` in the context of `OmeroMetadata` which is to return singular primitives and has to interact with an interface whose ancestry conforms to the inherent implementation semantics of XML has to be able to operate on these collections somehow. Therefore, it does so based on collection indexing as it has no other choice. The order is implicit, not guaranteed to be consistent across executions and fundamentally designed to satisfy the sole requirement of iteration across all the references that apply at that level:

```
for (int i = 0; i < getImageAnnotationRefCount(); i++) {
    String annotationRef = getImageAnnotationRef(imageIndex, i);
    ...
}
```

These are all fundamental restrictions of the "primitives only" design of the `MetadataStore` and `MetadataRetrieve` interfaces. The utility and fitness for purpose of that design can certainly be debated but that is the design. Were we to allow collections as part of these interfaces the prototypes would be similar to the following:

```
    void setImageAnnotationRefs(int imageIndex, Collection<AnnotationRef>, annotationRefs)
    Collection<AnnotationRef> getImageAnnoationRefs(int imageIndex)
```

The fact is however that we don't and doing so opens a whole different can of worms with respect to what `AnnotationRef` would be, whether the interface as a whole makes sense at all now, etc.

melissalinkert:
Was this intended to be changed in MetadataStore as well?

rleigh-dundee:
@melissalinkert No, and this was the reason for my question.  It's actually unused in MetadataStore, hence the discrepancy in behaviour between the two.

mtbc:
```java
        /**
         * Get the AnnotationRef property of Image.
         *
         * @param imageIndex the Image index.
         * @param annotationRefIndex AnnotationRef index.
         * @return the AnnotationRef property.
         */
	String getImageAnnotationRef(int imageIndex, int annotationRefIndex);
```

```java
	/**
         * Set the AnnotationRef property of Image.
         *
         * @param annotation AnnotationRef to set.
         * @param imageIndex the Image index.
         * @param annotationRefIndex AnnotationRef index (unused).
         */
	void setImageAnnotationRef(String annotation, int imageIndex, int annotationRefIndex);
```

:+1: Looks to me to be good to merge.

PR #1901, opened 2015-07-29
use OME XML meta data in file format version 2

bjoernthiel:
please review

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -192,7 +192,7 @@ public class OBFReader extends FormatReader
     }
 
     MetadataStore ome = makeFilterMetadata();
-    if (fileVersion <= 1 || ome_xml.isEmpty())
+    if (fileVersion <= 1)
     {
       MetadataTools.populatePixels(ome, this);
 
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -192,7 +192,7 @@ public class OBFReader extends FormatReader
     }
 
     MetadataStore ome = makeFilterMetadata();
-    if (fileVersion == 1)
+    if (fileVersion == 1 || ome_xml.isEmpty())
     {
       MetadataTools.populatePixels(ome, this);
 
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -192,7 +192,7 @@ public class OBFReader extends FormatReader
     }
 
     MetadataStore ome = makeFilterMetadata();
-    if (fileVersion == 1 || ome_xml.isEmpty())
+    if (fileVersion <= 1 || ome_xml.isEmpty())
     {
       MetadataTools.populatePixels(ome, this);
 


components/formats-bsd/src/loci/formats/in/OBFReader.java
```diff
@@ -168,56 +191,76 @@ protected void initFile(String id) throws FormatException, IOException
...
-        }
-        if (lengthZ > 0)
-        {
-          final Length physicalSizeZ = FormatTools.createLength( lengthZ / obf.sizeZ  , UNITS.MICROM);
-          ome.setPixelsPhysicalSizeZ(physicalSizeZ, series);
-        }
+        ServiceFactory factory = new ServiceFactory() ;
+        OMEXMLService service = factory.getInstance( OMEXMLService.class ) ;
+
+        service.convertMetadata( ome_xml, ome ) ;
```
melissalinkert: My only real worry here is what happens if the OME-XML stored in the file is invalid or has dimensions that don't match what was stored in the file (e.g. if the file were modified post-acquisition).  Would it make sense to try parsing and converting the OME-XML first, and if that doesn't work - ServiceException is thrown or the dimensions are incorrect - then fall back to the version 1 way of populating the MetadataStore?

components/formats-bsd/src/loci/formats/in/OBFReader.java
```diff
@@ -152,9 +156,28 @@ protected void initFile(String id) throws FormatException, IOException
...
+      for (String key = readString() ; key.length() > 0 ; key = readString())
+      {
+        if (key.equals( "ome_xml" ))
+        {
+          ome_xml = readString() ;
+          break ;
+        }
+        else
+        {
+          readString() ;
```
melissalinkert: Are keys other than ```ome_xml``` likely to occur?  If so, it might be useful to store those key/value pairs in the original metadata table, by calling ```addGlobalMeta(key, readString())``` here.

bjoernthiel:
I have to put this aside for the next three weeks as I'm on vacation.

melissalinkert:
@bjoernthiel - for when you're back, the ```origin/develop``` branch will need to be merged into this one, and any conflicts will need to be resolved.  We'll can review again as soon as that is done.

bjoernthiel:
@melissalinkert conflicts resolved

melissalinkert:
Conflict resolution looks fine, but it looks like my previous comment about falling back to the version 1 behavior in case of invalid OME-XML hasn't been addressed?

bjoernthiel:
Right. 
I don't know how to validate the OME-XML meta data other than using them. 
In case the meta data are invalid there will be an exception and we will need some hand work anyway to find out what is right and what is wrong with the particular dataset. 
Are the other readers aware of possibly invalid OME-XML meta data and do they have recovery code inside? Please give an example.

components/formats-bsd/src/loci/formats/in/OBFReader.java
sbesson: See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-test_images_good/648/console. This line generated a `NullPointerException` for some `msr` data. Excluding for a rebuild.

bjoernthiel:
@sbesson can you please pass me the test case causing the error:
/ome/data_repo/from_skyking/imspector/zoltan/3D-CONFOCAL-PSF-235nm.msr

sbesson:
@bjoernthiel http://downloads.openmicroscopy.org/images/ should contain all the OME sample filesets which have been made publicly available. This implies all the other sample filesets used for the development and testing of Bio-Formats cannot be shared with the community including the Imspector OBF file you are mentioning.
To make sure we include your fix in the upcoming version of Bio-Formats, we will likely cherry-pick your commits and open another PR with the additional fix allowing it to work with all of our repository samples.

bjoernthiel:
@sbesson can you please debug the test case yourself and pass me the XML string causing the error.

sbesson:
@bjoernthiel in that case the `ome_xml` is the empty string, i.e the default value you set in your PR.

bjoernthiel:
@sbesson 
I'm not aware of an Imspector Application writing version 2 data with an empty OME-XML string. 
Do you know where that test data comes from? 
I fixed the reader.

sbesson:
@bjoernthiel: looking at the data, the `fileVersion` is `0`. Should the initial condition be `if (fileVersion <= 1)` instead of `if (fileVersion == 1)`?

bjoernthiel:
@sbesson Thanks for making that clear to me. I've not been aware of a file version 0.

sbesson:
You are welcome. I'll run the automated tests from the latest version of this PR against our `imspector` samples once Travis has validated your last commits.

sbesson:
Thanks. https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/60/ looks good with this PR included. Reincluding into the week-end builds. If everything passes, we will be able to merge this and include it in the upcoming Bio-Format s5.1.5.

sbesson:
Merging for Bio-Formats 5.1.5

PR #1902, opened 2015-07-30
Fix new exceptions when parsing SimplePCI TIFF

sbesson:
See  http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-July/005511.html and QA 11327

Recent version of HCImage seems to have modified the header of the TIFF file. This commit should fix NumberFormatException when parsing the bit depth and fix a following NullPointerException due to the absence of the Capture table.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SimplePCITiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SimplePCITiffReader.java
@@ -181,8 +181,12 @@ public class SimplePCITiffReader extends BaseTiffReader {
       m.bitsPerPixel = Integer.parseInt(displayDepth);
     } else {
       String bitDepth = cameraTable.get("Bit Depth");
-      bitDepth = bitDepth.substring(0, bitDepth.length() - "-bit".length());
-      m.bitsPerPixel = Integer.parseInt(bitDepth);
+      if (bitDepth != null && bitDepth.length() > "-bit".length()) {
+        bitDepth = bitDepth.substring(0, bitDepth.length() - "-bit".length());
+        m.bitsPerPixel = Integer.parseInt(bitDepth);
+      } else {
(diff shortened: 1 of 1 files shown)


jburel:
```displayDepth != null```: i will also check that the string is not empty to avoid NPE during parsing

**Comment by sbesson** [comment]
@jburel the main problem is that if bitDepth is also null, we have a priori no information to fill the `bitsPerPixel` core metadata which means the reader initialization would fail anyways. Do we have a generic intialization exception to throw instead of the NPE?

**Comment by jburel** [comment]
you could throw a ```FormatException```

**Comment by jburel** [comment]
```
if (displayDepth != null) {
      m.bitsPerPixel = Integer.parseInt(displayDepth);
    }
```
if the string is not ```null``` but empty it will throw a NPE, so I will add check in if statement
```displayDepth.trim().getLength() > 0``

**Comment by jburel** [comment]
withpr

**Comment by jburel** [comment]
The file will need to be added to the testing suite if not already

**Comment by sbesson** [comment]
@jburel: definitely, the QA file will be added together with the configuration to the test repository. Regarding exception, adding the extra check you suggested would effectively replace the `NumberFormatException` by a generic `FormatException`.

**Comment by sbesson** [comment]
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/576/

```
   [testng] [2015-08-11 15:02:42,587] [pool-1-thread-11] Initializing /ome/data_repo/from_skyking/simplepci/qa-11327/bone_tmp1981_00011.tif: 
   [testng] [2015-08-11 15:02:43,749] [pool-1-thread-11] memo file size for /ome/data_repo/from_skyking/simplepci/qa-11327/bone_tmp1981_00011.tif = 34249 bytes

```

PR #1903, opened 2015-07-30
cpp: common: Correct boolean copy constructor to copy raw value

rleigh-dundee:
Found while fixing an optimisation warning for MSVC.

We were copy constructing using the boolean value rather than the raw value.  No change to the unit tests, though I've added a number of extra tests to check both the logical Boolean value and the raw bit pattern in memory to make sure this is caught properly.

--------

Testing: CI jobs should remain green.

If you want to verify by hand that the change is needed, then apply only the unit test changes without the boolean.h changes, and you'll get:

```
[ RUN      ] Boolean.CopyConstruct
/Users/rleigh/code/bioformats/cpp/test/ome-common/boolean.cpp:50: Failure
Value of: raw
  Actual: '\x1' (1)
Expected: std::numeric_limits<uint8_t>::max()
Which is: '\xFF' (255)
[  FAILED  ] Boolean.CopyConstruct (30 ms)
```

### DIFF
--- a/cpp/lib/ome/common/boolean.h
+++ b/cpp/lib/ome/common/boolean.h
@@ -100,7 +100,7 @@ namespace ome
        * @param value the value to copy.
        */
       boolean(const boolean& value):
-        value(value)
+        value(value.value)
       {}
 
       /**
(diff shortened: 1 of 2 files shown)


joshmoore:
 * Changes look sensible.

PR #1904, opened 2015-07-30
cpp: common: Correct Boost.Endian C++11 noexcept macro

rleigh-dundee:
Fixes a compilation failure on Windows.   This failure isn't triggered on any of our other platforms.  Essentially, we should have been using OME_HAVE_NOEXCEPT (see `cpp/cmake/CompilerChecks.cmake` and `cpp/lib/ome/common/config.h.in`).

Testing: CI jobs should remain green.  There's not much in the way of practical testing at this point.


joshmoore:
 * Changes look sensible.

PR #1905, opened 2015-07-30
cpp: common: Update compatibility snprintf for VS2013 and VS2012

rleigh-dundee:
Another Windows compatibility tweak.  We've had a Windows compatible snprintf/vsnprintf implementation for ages, but obviously not used until now.  This change fixes the detection of snprintf and also the fallback implementation.

Note: See cpp/cmake/External_tiff_files/port/snprintf.c which is where this implementation derives from.

Testing: All CI jobs should remain green--they don't use this codepath.

When running cmake, the OME_HAVE_SNPRINTF check should be true for all CI jobs since all the existing platforms have it--just check one of the CI node logs.  This will make sure the fallback will be used if this check fails on Windows.  The actual fallback isn't testable at this point.  But note it has been tested in the libtiff code I contributed to the tiff upstream, so it's not untested.


joshmoore:
 * Changes look sensible.

PR #1906, opened 2015-07-30
cmake: Rename BIOFORMATS_EP_SOURCE_CACHE to source-cache

rleigh-dundee:
No functional changes; this is simply to make the cmake interface more friendly and accessible by replacing the obscure and originally internal-only variable with a shorter and more memorable option name.

--------

Testing: You can use `-Dsource-cache=/path/to/source-cache` rather than `-DBIOFORMATS_EP_SOURCE_CACHE=/path/to/source-cache` when running cmake superbuild with `-Dbioformats-superbuild=ON`.


joshmoore:
* Local build on OS X passes (and caches!)
* Changes look sensible.

PR #1908, opened 2015-08-03
adjust XML serialization of unit property values

mtbc:
The values of dimensioned quantities were being serialized as floating-point numbers regardless of their definition in the OME-XML model schema. This PR adjusts the serialization so that dimensioned integers are not serialized with a decimal point in their value. Check that the new unit tests make sense and that they pass.


rleigh-dundee:
Looks fine, as do the unit tests.  Tests pass with the first commit and fail without.  Please merge.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-generated-sources/ for impact on generated model sources.

PR #1910, opened 2015-08-04
xsd-fu: Correct static initialisation order for enum key-value mappings

rleigh-dundee:
The string_map and value_map static maps were not protected with a static method, required to prevent use in other translation units ahead of construction in the program's static initalisation phase. Wrap these and also the lowercased string map (for case-insensitive lookups) in static methods, and use these methods where the plain maps were originally used.  We did have two static methods, but they just returned a reference to the static maps; they now implicitly initialise and fill the maps on their first invocation.

This is needed for Windows, though it's actually buggy on all platforms.  The reason it worked on Unix is that when using shared libraries, the static initialisation of each shared library is done in dependency order, but when using static libs this implicit ordering no longer occurs and the program will crash when it attempts to use a map prior to it being constructed.

--------

Testing: Check jobs are green.  The actual model enum API is unchanged, this being an implementation detail of the enum mappings, so unit tests are sufficient.

### DIFF
--- a/components/xsd-fu/templates-cpp/OMEXMLModelEnum.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelEnum.template
@@ -68,8 +68,52 @@
 
 using boost::trim;
 using boost::format;
-{% end %}\
 
+namespace
+{
+
+  typedef std::pair<std::string,${klass.langTypeNS}::enum_value> smap;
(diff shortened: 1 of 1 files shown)


mtbc:
The generated code in `cpp/lib/ome/xml/model/enums/` looks plausible and builds fine and bf-test's "info" and "view" still work. The CPP DEV jobs look good. Looks to me to be good to merge.

PR #1913, opened 2015-08-05
Formats-bsd: native time units

sbesson:
See https://trello.com/c/nO35TKXK/4-units-and-readers

This PR modifies the storage of time metadata in the BSD formats to use the native time unit when appropriate. The `pixelSizes` variable in the `ICSReader` is also renamed as `scales` to prevent confusion as it applies both to the pixels physical size and the pixels time increment.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -336,7 +336,7 @@ public class MicromanagerReader extends FormatReader {
             nextStamp < p.timestamps.length &&
             p.timestamps[nextStamp] != null)
           {
-            store.setPlaneDeltaT(new Time(p.timestamps[nextStamp++], UNITS.S), i, q);
+            store.setPlaneDeltaT(new Time(p.timestamps[nextStamp++], UNITS.MS), i, q);
           }
         }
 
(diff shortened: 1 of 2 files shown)


components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
```diff
@@ -626,11 +626,11 @@ else if (token.trim().startsWith("}")) {
 
           if (key.equals("Exposure-ms")) {
             double t = Double.parseDouble(value);
-            p.exposureTime = new Time(new Double(t / 1000), UNITS.S);
+            p.exposureTime = new Time(new Double(t), UNITS.MS);
           }
           else if (key.equals("ElapsedTime-ms")) {
             double t = Double.parseDouble(value);
-            stamps.add(new Double(t / 1000));
+            stamps.add(new Double(t));
```
melissalinkert: This (and exposureTime above) could probably be simplified to just one line e.g. ```stamps.add(Double.valueOf(value))```.

melissalinkert:
Minor nitpick aside, reads fine to me.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/576

melissalinkert:
Minor nitpick still kind of stands, but this works, so merging - can clean up in a follow-up PR.

sbesson:
@melissalinkert: did 8ee7d60 not fix the nitpick mentioned in https://github.com/openmicroscopy/bioformats/pull/1913#discussion_r36363096?

melissalinkert:
Not quite, because it's doing ```Double.valueOf``` on a ```double```, when calling ```Double.valueOf(value)``` (and removing the ```t = Double.parseDouble(value)``` lines) should give the same result.

PR #1914, opened 2015-08-05
minor code adjustments to a few GPL readers

mtbc:
This PR rewrites some GPL reader code in a minor way to increase readability and/or efficiency:
* FV1000Reader
* InCellReader
* MIASReader
* ScanrReader

CI does the testing.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
@@ -629,8 +629,8 @@ public class FV1000Reader extends FormatReader {
     final List<String> baseKeys = new ArrayList<String>();
 
     for (int i=0, ii=0; ii<getImageCount(); i++, ii++) {
-      String file = filenames.get(new Integer(i));
-      while (file == null) file = filenames.get(new Integer(++i));
+      String file = filenames.get(i);
+      while (file == null) file = filenames.get(++i);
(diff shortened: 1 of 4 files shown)


mtbc:
Is further work needed here? (Causing problems with some CI job perhaps?)

melissalinkert:
I guess my only real concern here (and in the other cleanup PRs) is the widespread impact on memoization.  If we're already assuming that all .bfmemo files are invalidated on every release, then that's fine - if the expectation is that most memo files will be usable between 5.1.3 -> 5.1.4, then it might be better to hold off until after the dev_5_1 split (I honestly can't remember what we've said about this in the past).  @sbesson, @pwalczysko?

mtbc:
https://www.openmicroscopy.org/site/support/omero5.1/sysadmins/server-upgrade.html#memoization-files-invalidation says that the memo files won't be usable; the only "not minor version" caveat I've noticed is the warning box about database upgrades. So, it's already warned about. Having said that,

1. I don't mind if we choose to hold off anyway, though I'd worry about making future rebases harder.
1. If you want, I could try to split up future cleanup PRs to separate out the commits that change a field type.

melissalinkert:
OK, that's fine then.  Discussed in standup earlier today, and we'll plan to get these in pre-5.1.4.  Haven't reviewed all the diffs yet, but this one is good to go, so merging.

PR #1919, opened 2015-08-06
cpp: bioformats: Correct unused variable warnings in catch statements

rleigh-dundee:
Fixes warnings about unused variables with MSVC.  GCC and clang don't seem to treat unused exceptions as being problematic.  Comment out as for unused function arguments.

Also mark a couple of exceptions as `const` or else they won't be caught (since all exceptions are const).

--------

Testing: Just check the jobs are green.

### DIFF
--- a/cpp/lib/ome/bioformats/MetadataMap.h
+++ b/cpp/lib/ome/bioformats/MetadataMap.h
@@ -240,7 +240,7 @@ namespace ome
             list_type& list(get<list_type>(key));
             list.push_back(value);
           }
-        catch (const boost::bad_get& /* e */)
+        catch (const boost::bad_get&)
           {
             list_type new_list;
             new_list.push_back(value);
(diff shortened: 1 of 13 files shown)


jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/898/ is green

any reason not to remove instead of commented out?

jburel:
in OMETiffReader: ```catch (const std::exception& e)```

jburel:
Ignore comment about remaining catch in Tiff.cpp and Tiffreader. The exception was actually used. I missed it

rleigh-dundee:
@jburel "any reason not to remove instead of commented out?"  Sorry, I don't understand the question or the context.  What is this referring to?

jburel:
in some files we have cases w/o argument i.e. no ```e``` present

rleigh-dundee:
Pushed a change to make this consistent.

jburel:
Thanks @rleigh-dundee

PR #1920, opened 2015-08-07
cmake: Windows portability and efficiency fixes

rleigh-dundee:
- Detect debug libraries; required for supporting debug builds of bioformats, and obviously essential to debug bioformats properly.  Previously the cmake macros for TIFF and Xerces-C only detected release libraries
- Remove duplicate xsd-fu targets and dependencies; probably just cosmetic, but could potentially confuse some generators
- Remove hard-coded assumptions that the "Visual Studio" generator is used on Windows.  This will not affect this generator, but it adds the ability to use other more powerful generators, for example Ninja and GNU make which can fully parallelise the build.  msbuild/VS can only parallelise building of projects, but the projects themselves are build in a serial manner which means that in practice it's only using a single CPU for most of the build.  Testing with Ninja, it can peg all the system's CPUs at 100% while building bioformats, so should cut the build time down by a factor of 20 or more on the CI node, and will be of obvious benefit to developers by dramatically reducing the wait between builds.

--------

Testing:

The regular CI builds will test everything (https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/901/,BIOFORMATS-5.1-merge-cpp-superbuild-win).  I'll also add a job to build with Ninja so we can compare the build times with the existing job, and convert them if it proves useful.


rleigh-dundee:
Where would we include it?  It can't be in the files themselves; they are to be applied to the original source code for each project.  Or can this wait until we move the whole superbuild logic into a separate git repo?  This is on the todo list once the windows build is working.

jburel:
@rleigh-dundee: as long as it is on the TODO list

rleigh-dundee:
@jburel Added to https://trello.com/b/552WPdoZ/bio-formats-5-1-4

jburel:
@rleigh-dundee: Thanks
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/901/ is green

PR #1921, opened 2015-08-07
xsd-fu: Correct static initialisation order issues in the C++ templates

rleigh-dundee:
While GCC and Clang provide certain guarantees with regard to the initialisation order (or at least, do so in practice), this is not the case with MSVC.  This PR removes all static variables in the unnamed namespace in the enum template, as well as the XML namespace in the model object template.  It also fixes a potential problem with iterators which MSVC seemed to object to but which the other compilers were happy with.

--------

Testing: Check jobs are green (https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp).  Double-check the code generation changes look sane.

### DIFF
--- a/components/xsd-fu/templates-cpp/OMEXMLModelEnum.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelEnum.template
@@ -75,42 +75,6 @@ namespace
   typedef std::pair<std::string,${klass.langTypeNS}::enum_value> smap;
   typedef std::pair<${klass.langTypeNS}::enum_value,std::string> vmap;
 
-  smap init_string_values[] =
-    {
-{% for value in klass.possibleValues %}\
-{% if value == klass.possibleValues[-1] %}\
-      smap("${value}", ${klass.langTypeNS}::${enum_value_name(value, False).upper()}),
-{% end %}\
-{% if value != klass.possibleValues[-1] %}\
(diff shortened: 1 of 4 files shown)


jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/ is green 
waiting on super build

rleigh-dundee:
@jburel Done.

jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp-superbuild-win/21/ is green

rleigh-dundee:
Note procedure for viewing the generated source changes:

```
git checkout origin/develop
mkdir old && cd old
cmake .. && make gensrc
cd ..
git merge gh/cpp-static-init-xsdfu
mkdir new && cd new
cmake .. && make gensrc
cd ..
diff -urN old/cpp/lib/ome/xml/model new/cpp/lib/ome/xml/model
```

rleigh-dundee:
https://gist.github.com/rleigh-dundee/7ef5a93217773596b0e7 is the diff

jburel:
ran the diff
namespace and enumeration correctly modified.
unfortunately, I cannot attach it to the PR
@rleigh-dundee is looking at fixing the job so the diff will be available when the job is done

jburel:
diff now available
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-generated-sources/

PR #1922, opened 2015-08-08
cpp: common: Correct string iterator use in mstream and formatreader tests

rleigh-dundee:
This change fixes the nonstandard use of iterators in these unit tests.  This doesn't work on Windows--dereferencing an end iterator is not really allowed but works fine with GCC and Clang.  Use a pointer to the start and size instead, or pointer to the start and pointer to one-past-the-end as pointer arithmetic without using iterators.

--------

Testing: Covered by the mstream unit test; check jobs are green.


rleigh-dundee:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/901/

jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/902/ is green

PR #1923, opened 2015-08-08
cpp: common: Correct units test for MSVC stream output

rleigh-dundee:
The output of floating point exponents is slightly different for MSVC--it uses three digits rather than two when there are only two significant digits.  This is not specified by the C++ standard and so implementations can differ slightly.  Adjust the output to match what is expected by removing the leading zero with a regular expression.

See http://stackoverflow.com/questions/9226400/portable-printing-of-exponent-of-a-double-to-c-iostreams for some background.

--------

Testing: Check jobs are green (https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/).  Should be a no-op for the existing Unix CI jobs.  Not tested by the windows jobs yet but will be once the other pending PRs are merged and we can turn on a full Windows build of the whole bioformats library set.

### DIFF
--- a/cpp/lib/ome/compat/regex.h
+++ b/cpp/lib/ome/compat/regex.h
@@ -59,6 +59,7 @@ namespace ome
     using std::regex;
     using std::regex_error;
     using std::regex_match;
+    using std::regex_replace;
     using std::regex_search;
     using std::cmatch;
     using std::smatch;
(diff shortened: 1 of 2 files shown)


jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp/902/ is green

PR #1924, opened 2015-08-10
Roi

jburel:
Do not store the z, c, t on the shape if no position set.
To test:

test 1:
 * Open an image with several z/t using B-F importer e.g. ```dv/IAGFP-Noc01_R3D.dv```
 * Draw a shape and add it to the ROI manager
 * Check that the position is not set (ROIManager>properties)
 * Export using B-F exporter
 * Open the generated OME-TIFF and check the OME-XML section of it. You should not have the following attributes ```TheC```, ```TheZ``` ```TheT```

test 2:
 * Repeat the same steps but this time, set a position e.g. 1, 3, 4 (depends on the image selected)
 * Export using B-F exporter
 * Open the generated OME-TIFF and check the OME-XML section of it. You should have the following attributes ```TheC=0```, ```TheZ=2``` ```TheT=3```

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -355,15 +355,21 @@ public class ROIHandler {
 
     }
 
-
+    /**
+     * Returns the roi if any stored in the ROI manager.
+     * @return
+     */
     public static Roi[] readFromRoiManager(){
 
         RoiManager manager = RoiManager.getInstance();
         if (manager == null) return null;
-        Roi[] rois = manager.getRoisAsArray();
-        return rois;
+        return manager.getRoisAsArray();
     }
 
+    /**
+     * Returns rois if any from the overlay.
+     * @return See above
+     */
     public static Roi[] readFromOverlays(){
 
         ImagePlus image = IJ.getImage();
@@ -377,7 +383,7 @@ public class ROIHandler {
     public static void saveROIs(MetadataStore store) {
 
         Roi[] rois = readFromOverlays();
-        if (rois == null) {
+        if (rois == null || rois.length == 0) {
             rois = readFromRoiManager();
         }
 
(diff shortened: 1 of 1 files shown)


components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
```diff
@@ -355,15 +355,21 @@ else if (shapeObject instanceof ome.xml.model.Rectangle) {
 
     }
 
-
+    /**
+     * Returns the roi if any stored in the ROI manager.
+     * @return
```
sbesson: See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-build/978/warnings39Result/

pwalczysko:
test 1 passes (but you have to manually set the Properties in ROI Manager to ``0, 0, 0`` or ``none`` for it to work (probably an IJ bug).

pwalczysko:
test 2 : I drawn a region on t = 7, z =2, c = 1. Checked the Properties and it said ``1, 2, 7``. Then confirmed the Properties with OK. Then Exported the image. Reimported it and got in the XML: 
```
<Shape ID="Shape:0:0" StrokeColor="-65281" Text="0020-0167-0176" TheC="0" TheT="7" TheZ="1">
```
This seems wrong.

pwalczysko:
The https://github.com/openmicroscopy/bioformats/pull/1924#issuecomment-130250293 was done with IJ 1.49o java 1.6

pwalczysko:
With IJ 1.50 java 1.8 I have with the same image as in https://github.com/openmicroscopy/bioformats/pull/1924#issuecomment-130250806 and the  same setup 
```
<Shape ID="Shape:0:0" StrokeColor="-65281" Text="0020-0152-0112" TheC="0" TheT="6" TheZ="1">
```

pwalczysko:
Further investigation showed that the problem with the wrong t index reported https://github.com/openmicroscopy/bioformats/pull/1924#issuecomment-130250293 is associated to the ROI type (ellipse), not the ImageJ version. 

All works here as expected.

pwalczysko:
Created https://trac.openmicroscopy.org/ome/ticket/12987#comment:1 
cc @bramalingam

pwalczysko:
Works as expected. Ready to merge.

melissalinkert:
For a future PR, might make sense to add a method to go from OME-XML Color to ImageJ Color to reduce duplicate code.

PR #1927, opened 2015-08-12
TEST: Force generation of diffs for metadata

rleigh-dundee:
Not for merging.

Check that https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-generated-sources/ has metadata diffs for java and cpp, and that both match the template change for MetadataStore (one comment addition per file).

### DIFF
--- a/components/xsd-fu/templates-cpp/MetadataStore.template
+++ b/components/xsd-fu/templates-cpp/MetadataStore.template
@@ -105,6 +105,8 @@
  *─────────────────────────────────────────────────────────────────────────────
  */
 
+// Test diff generation (C++).
+
 {% if fu.SOURCE_TYPE == "header" %}\
 #ifndef ${fu.GUARD}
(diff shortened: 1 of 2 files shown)


rleigh-dundee:
ome-xml-metadata-cpp.diff is empty.

jburel:
@rleigh-dundee: just looked at it and noticed the same problem
Java works fine.
cc @sbesson

jburel:
Marking it as onhold so it does not get listed.

sbesson:
Thanks for the review. The generated `meta` C++ component was not previously committed. After modification of the jobs, see https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-generated-sources/253/.

jburel:
Thanks @sbesson.
Closing the PR, now that we have tested the job

rleigh-dundee:
Looks good, thanks.

PR #1929, opened 2015-08-13
Add new public Imaris HDF documentation to the format page

melissalinkert:
http://open.bitplane.com/Default.aspx?tabid=268 was announced today, see http://www.bitplane.com/news/bitplane-makes-the-ims-file-format-open-source-130815.

As the documentation is now publicly available, the openness rating is also increased accordingly.


hflynn:
Changes look good on staging

hflynn:
Good to merge

PR #1932, opened 2015-08-13
MicromanagerReader: do not return intermediate primitive double

sbesson:
Instead use Double.valueOf() directory to create a Double instance for both timestamps and exposure times.

Follow-up of https://github.com/openmicroscopy/bioformats/pull/1913#issuecomment-130858035

Tests should remain green with this change.


melissalinkert:
:+1: :smiley:

PR #1935, opened 2015-08-15
Handle NPDI files with macro slide image and mask image

sbesson:
In addition to the pyramid images and the macro slide image, some NDPI files contain a trailing image associate to a mask or activation map. Previously, this was handled by assuming the last series only was not part of the pyramid for data with sizeZ=1. This commit uses the tag 65421 instead to determine the macro slide image (-1) and the mask (-2) if applicable.

To test this PR:
- check the automated tests remain green
- import `npdi` samples into OMERO with this PR included into Bio-Formats. Check either 2 or 3 images are created depending on the file and that the pyramid image and the macro slide image are always imported as separate images.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPIReader.java
@@ -55,6 +55,7 @@ public class NDPIReader extends BaseTiffReader {
   // -- Constants --
 
   private static final int MAX_SIZE = 2048;
+  private static final int SOURCE_LENS = 65421;
   private static final int MARKER_TAG = 65426;
   private static final int THUMB_TAG_2 = 65439;
   private static final int METADATA_TAG = 65449;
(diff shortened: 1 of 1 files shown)


sbesson:
This change should not be breaking when series are flattened and thus not automated NDPI tests should not be affected as suggested by https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-breaking-repository-subset/65/console.

A contrario, it should affect unflattened series (i.e. pyramids) in a breaking manner. Under OMERO, NDPI images containing 1 multi-resolution whole slide image, 1 macro-slide and 1 mask should now be imported as 3 images instead of 2 images without with PR (with the macro-slide image imported as part of the pyramid image) - for examples of such files see for instance `from_skyking/hamamatsu/marc/test.ndpi` /cc @pwalczysko @aleksandra-tarkowska

bramalingam:

melissalinkert:
Potential NPE discussed in devteam, and will be resolved in a follow-up PR.  Merging so we have this for 5.1.4.

PR #1939, opened 2015-08-19
Prevent potential NullPointerException while checking the pyramid height

sbesson:
If the tag defining the macro/map images is not defined, fallback to the legacy behavior assuming the last IFD is the macro image for non multi-Z NDPI images.

Follow-up of https://github.com/openmicroscopy/bioformats/pull/1935


rleigh-dundee:
The logic looks fine to me.  Does this need any manual testing, or is this taken care of by the full-repo jobs?

sbesson:
Imported various NDPI samples into a development OMERO server with this PR included to check for any regression of the fix described in #1935. Everything seems fine so I would expect the validation by the full-repository job to be sufficient.

melissalinkert:
:+1: https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/589/

PR #1941, opened 2015-08-19
specification: Correct incorrect catalogue entry

rleigh-dundee:
Found while working on catalogue processing and the duplicate/inconsistent filename for the same system ID caused it to throw an exception.

--------

Testing: none.  There's nothing using this particular catalogue at the moment, so just confirm it's matching the `2008-02` schema.


jburel:
looks good

PR #1942, opened 2015-08-20
FEI TIFF: parse metadata from the Titan system tag

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12743.

To test, use the .tif file from QA 10514.  Compare the output of ```showinf -nopix`` with and without this change; there should be noticeably more metadata shown with this change.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FEITiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FEITiffReader.java
@@ -28,7 +28,9 @@ package loci.formats.in;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.HashMap;
 
 import loci.common.DateTools;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/FEITiffReader.java
```diff
@@ -350,6 +393,12 @@ else if (key.equals("User")) {
         else if (key.equals("Magnification")) {
           magnification = new Double(value);
         }
+        else if (key.equals("X") && "PixelsSize".equals(parent)) {
+          sizeX = new Double(value);
```
sbesson: For the QA file described in this PR, the main issue is that the physical pixel size seems to be expressed in meters. Since the format mostly deals with TEM data, the value is smaller than `Constants.EPSILON` and thus is registered as a zero value length. Thoughts about actively convert this value to microns (for Titan data at least)?

sbesson:
Apart from the comment about the physical pixels size, this TITAN TIff file opens fine with this PR with substantial metadata parsed from the file. All previous versions of the FEI TIFF still open. We might want to add QA 10514 to the automated test suite with a relevant configuration file.

melissalinkert:
Commit pushed to update the physical pixel sizes - we were already multiplying by 1000000 for Helios data, so doing the same for Titan data is consistent (and the corresponding data_repo_config PR has been updated now too).

sbesson:
Retested on the sample file with the latest version of this PR. Works as expected. Ready to merge once it has been validated by the automated tests.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/599/. Ready to merge.

PR #1943, opened 2015-08-20
cpp: common: Rework XML EntityResolver catalogue registration

rleigh-dundee:
The aim is to make the registration mechanism more flexible so it can be used to manually register the catalogues on Windows.  It was previously quite rudimentary because it was automatic and never needed to be used manually.

- Simplify EntityResolver.  It's now a simple class with methods to add entities or catalogues.  No global state.  Any user may freely add additional entities or catalogues.  Removal is not possible.
- Add OMEEntityResolver; this is simply an EntityResolver with the OME catalogues registered by default.
- Refactor createDocument functions to add an EntityResolver as a parameter; add additional createDocument functions in ome-xml which use OMEEntityResolver.  These are for use by any code which wants to read and validate documents using OME schemas.
- Code calling ome::common::xml::dom::createDocument now uses ome::xml::createDocument, including unit tests.  This means they all use local copies of the schemas via the OMEEntityResolver.
- The stricter checking of system ID/filename mismatches means that it will now abort if the OME-XML catalogues contain invalid duplicate system IDs as fixed in https://github.com/openmicroscopy/bioformats/pull/1941.  When adding an entity which was already registered, this will only work if the entity being added has the same filename, otherwise an exception is thrown due to being inconsistent.  This has the handy side-effect of validating the correctness of the catalogue files in the unit tests.
- Also address a static initialisation issue in module.cpp for internal path data.

--------

Testing: The cpp jobs should continue to be green, but https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp-superbuild-win/ should reduce from 2 hours to something more reasonable.  The changes made here should be completely covered by the existing unit tests.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
@@ -629,8 +629,8 @@ public class FV1000Reader extends FormatReader {
     final List<String> baseKeys = new ArrayList<String>();
 
     for (int i=0, ii=0; ii<getImageCount(); i++, ii++) {
-      String file = filenames.get(new Integer(i));
-      while (file == null) file = filenames.get(new Integer(++i));
+      String file = filenames.get(i);
+      while (file == null) file = filenames.get(++i);
       file = sanitizeFile(file, path);
 
       if (file.indexOf(File.separator) != -1) {
@@ -1168,7 +1168,7 @@ public class FV1000Reader extends FormatReader {
       // populate ROI data - there is one ROI file per plane
       for (int i=0; i<roiFilenames.size(); i++) {
         if (i >= getImageCount()) break;
-        String filename = roiFilenames.get(new Integer(i));
+        String filename = roiFilenames.get(i);
         filename = sanitizeFile(filename, path);
         nextROI = parseROIFile(filename, store, nextROI, i);
       }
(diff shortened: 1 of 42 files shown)


rleigh-dundee:
Testing this evening shows a 115 min to 55 min change, so a ~50% reduction in build time isn't too shabby!

rleigh-dundee:
See also https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-generated-sources/lastSuccessfulBuild/artifact/ome-xml-model-cpp.diff/*view*/

jburel:
we may want at some point unify the usage of the annotation i.e. ```@returns``` and ```@return```.

jburel:
Superbuild: is certainly much faster 1h5 https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-cpp-superbuild-win/ 12 only, 11 currently broken

jburel:
minor: header new file should have 2015 not a date range

rleigh-dundee:
The copyright is fixed now, as are three minor doxygen warnings.

jburel:
@rleigh-dundee: thanks looking good.

joshmoore:
https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-cpp-superbuild-win/65/ is green with times for Release and Debug down from just under 2 hours to just over 1 hour. Merging.

PR #1945, opened 2015-08-20
docs: Document JACE JVM restrictions

rleigh-dundee:
Document that Java 6 or 7 is needed, and that Java 8 is not supported.


joshmoore:
Works for me.

PR #1946, opened 2015-08-20
QT: use the previous plane if there is no data for the current plane

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12888.

To test, use the file from QA 11074.  Without this change, ```showinf``` on the file will throw an exception as noted in the ticket.  With this change, ```showinf``` should display the entire time series, which can be compared against ```mplayer``` (or some other video player).


dominikl:
Good to merge. Just had to increase the max memory in bf.sh to be able to open then image with showinf.

PR #1947, opened 2015-08-20
CZI: fix channel colors when alpha is not recorded

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12996.

To test, use the file from QA 11361.  Without this change, the first channel's color should be black; with this change, the channels should be red and green.  This is easiest to see in ImageJ; you'll probably need to open the brightness/contrast window.


dominikl:
Good to merge.

PR #1948, opened 2015-08-20
Leica SCN: do not report more resolutions than there are IFDs

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12864.

To test, use the file from QA 11049.  Without this change, ```showinf``` should throw an exception as in the ticket; with this change ```showinf``` should report 3 series (to match ```tiffinfo *.scn | grep -c "TIFF Directory"```), all of which should be viewable with ```showinf```/ImageJ.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
@@ -311,6 +311,10 @@ public class LeicaSCNReader extends BaseTiffReader {
 
     ifds = tiffParser.getIFDs();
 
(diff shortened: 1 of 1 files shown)


sbesson:
Tested with https://github.com/snoopycrimecop/bioformats/commit/339ce51ff2d7293c89d5f11c55a46641acf1bdf9. The SCN file from QA 11049 is correctly initialized with 3 series corresponding to the different resolutions. Additionally this change does not destabilize SCN files as shown in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/590/. Ready to merge.

PR #1949, opened 2015-08-20
CZI: fix coordinates of pre-stitched tiles

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12993.

To test, use the file attached to the thread in the ticket.  Without this change, all 8 series had all 0 pixel values.  With this change, every plane in each series should match what is shown in Zen.  This is probably easiest to test in ImageJ, optionally with the ```Open all series``` and ```Stitch tiles``` options selected.


melissalinkert:
Commit pushed to fix https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/588.

bramalingam:
works as expected. Please merge.

PR #1950, opened 2015-08-23
SDT: add support for Zip-compressed data

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12893.

To test, use the files from QA 11079.  Without this change, the compressed file should have obviously wrong image data when compared with the uncompressed file.  With this change, the images for both files should match.

/cc @imunro

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SDTReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SDTReader.java
@@ -28,6 +28,7 @@ package loci.formats.in;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.zip.ZipInputStream;
 
 import loci.common.DataTools;
 import loci.common.RandomAccessInputStream;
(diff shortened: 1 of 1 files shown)


imunro:
Thanks for the heads-up.

imunro:
Do you think that the "if (codec == null) {"   inside the loops is optimised out?  If not, in the interests of performance it might be good to take this outside the loop unfortunately I can't see a way to do so that doesn't involve rather messy code duplication.

imunro:
Would it be better to have a class that inherits from RandomAccessInputStream  with a read method 
read(byte[] rowBuf)  {
 int nread = 0;
        while (nread < rowBuf.length) {
           int n = codec.read(rowBuf, 0, rowBuf.length - nread);
           nread += n;
            if (n <= 0) {
             break;
            }
         }
then just set it to use this class instead in the case of compressed data?

imunro:
Tested using FLIMfit and the data from QA 11079. Image appearance, max value and timebase all match for both files as expected so could be merged "as is" although  earlier comments still apply.

imunro:
@melissalinkert do you need anything else from me on this ?

melissalinkert:
Don't think so, thanks @imunro.  What I will likely do for now is just add a helper method in SDTReader to reduce the duplicate buffer reading - there's definitely room for a new class(es) at the loci.common level to help with this, but I want to avoid adding a lot of new API given that this is a patch release and we have I/O re-work coming up later in the year.

imunro:
Whatever you think best. It certainly works exactly as is which is great.

melissalinkert:
Commit pushed to reduce duplicate logic.

rleigh-dundee:
The change reads fine and looks good to merge.  I presume the full-repo-test will be using the new configuration files?

PR #1951, opened 2015-08-24
Automated tests: remove SlideBook6Reader class from ImageReader

sbesson:
SlideBook6Reader is bundled with Bio-Formats on Windows only and is maintained externally. Until we have a proper API in place allowing reader discovery, this commit should allows to remove it from the list of readers when running automated tests against the data repository.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -1557,7 +1557,7 @@ public class FormatReaderTest {
 
       LOGGER.debug("newFile = {}", newFile);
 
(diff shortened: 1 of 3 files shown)


melissalinkert:
Safe to unexclude https://github.com/openmicroscopy/bioformats/pull/1899 then?

sbesson:
I would expect so. I will have a go at running automated tests for Slidebook on Windows tomorrow with both PR included.

melissalinkert:
PR opened against this branch: https://github.com/sbesson/bioformats/pull/5

melissalinkert:
Merging as builds are now green.

PR #1952, opened 2015-08-24
Slidebook: fix offsets for newer files

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12919.

To test, use the file from QA 11119.  Without this change, attempting to open it should throw an exception (as in the ticket).  With this change, the file should have 3 T x 2 C (as indicated by the file name) and the images should look reasonable (though 16 bit, so you may need to open in ImageJ).

Putting this in the breaking queue to start, so that the subset job can be used to verify that builds won't break.


melissalinkert:
This apparently really breaks other things, per https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-breaking-repository-subset/66/console.

Closing for now - now that we have a plan for gh-1951/gh-1899, I am thinking it might make sense to close remaining Slidebook tickets as wont-fix, but let's discuss at the next standup.

PR #1953, opened 2015-08-25
docs: Use urllib conditionally

rleigh-dundee:
While this shouldn't be encountered in practice (it's a default python2 module?), it was mentioned as a problem by a user on IRC.  This moves the urllib import into the existing try block so that if the import fails, it's not a fatal error.  May also ease future python3 porting, since the docs are likely already buildable directly with a python3 version of sphinx-build.

--------

Testing: Check the jobs are green.


sbesson:
Same comment as https://github.com/openmicroscopy/ome-documentation/pull/1272#issuecomment-134955370. Merging.

PR #1934, opened 2015-08-14
Added CZT positions computation for non-hyperstacks

stelfrich:
This fix is related to https://github.com/openmicroscopy/openmicroscopy/pull/4045

It improves the heuristic for finding out, whether to call `roi.setPosition(int)` or `roi.setPosition(int, int, int)` based on the dimensionality of the image that the ROI is associated with.

PR #1954, opened 2015-08-25
Add CZT positions computation for non-hyperstacks

jburel:
replace gh-1934

This fix is related to https://github.com/openmicroscopy/openmicroscopy/pull/4045

It improves the heuristic for finding out, whether to call `roi.setPosition(int)` or `roi.setPosition(int, int, int)` based on the dimensionality of the image that the ROI is associated with.

cc @stelfrich

Make sure that this change does not affect the other scenarios

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -358,17 +358,25 @@ public class ROIHandler {
                 t = shapeObject.getTheT().getValue();
               }
               // ImageJ expects 1-based indexing, opposed to 
-              // 0-based indexing in OME  
-              if (images[imageNum].getNChannels() == 1 &&
-              images[imageNum].getNSlices() == 1) {
-                roi.setPosition(t+1);
-              }
-              else {
+              // 0-based indexing in OME
+              // Roi positions differ between hyperstacks and normal stacks
+              ImagePlus imp = images[imageNum];
+              if (imp.isHyperStack()) {
+                // Set position in CZT fashion for hyperstack
                 roi.setPosition(c+1, z+1, t+1);
+              } else {
+                // Number of channels in imp
+                int imageC = images[imageNum].getNChannels();
+                // Number of frames in imp
+                int imageT = images[imageNum].getNSlices();
+                // Determine which dimension is used as position
+                if (imageT > 1) {
+                  roi.setPosition(t+1);
+                } else if (imageC > 1) {
+                  roi.setPosition(t+1);
+                }
               }
-
             }
-            roi.setImage(images[imageNum]);
 
             if (sw == null) {
               roi.setStrokeWidth((float) 1);
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -373,13 +373,14 @@ public class ROIHandler {
               if (c == 0) c = 1;
               if (t == 0) t = 1;
               if (z == 0) z = 1;
-              if (c == 1 && z == 1 && t > 1) {
-                roi.setPosition(0, 0, t);
-              } else if (c == 1 && z > 1 && t == 1) {
-                roi.setPosition(0, z, 0);
-              } else if (c > 1 && z == 1 && t == 1) {
-                roi.setPosition(c, 0, 0);
-              } else {
+              if (imp.getNChannels() == 1 && imp.getNSlices() == 1 && t > 1) {
+                roi.setPosition(t);
+              } else if (imp.getNChannels() == 1 && z > 1 && imp.getNFrames() == 1) {
+                roi.setPosition(z);
+              } else if (c > 1 && imp.getNSlices() == 1 && imp.getNFrames() == 1) {
+                roi.setPosition(c);
+              } else if (imp.getNChannels() > 1 && imp.getNSlices() > 1
+                    && imp.getNFrames() > 1) {
                 roi.setPosition(c, z, t);
               }
             }
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -379,8 +379,7 @@ public class ROIHandler {
                 roi.setPosition(z);
               } else if (imp.getNSlices() == 1 && imp.getNFrames() == 1) {
                 roi.setPosition(c);
-              } else if (imp.getNChannels() > 1 && imp.getNSlices() > 1
-                    && imp.getNFrames() > 1) {
+              } else if (imp.isHyperStack()) {
                 roi.setPosition(c, z, t);
               }
             }
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -374,7 +374,15 @@ public class ROIHandler {
               if (c == 0) c = 1;
               if (t == 0) t = 1;
               if (z == 0) z = 1;
-              roi.setPosition(c, z, t);
+              if (c == 1 && z == 1 && t > 1) {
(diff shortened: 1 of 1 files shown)


pwalczysko:
Found a bug which is present also in 5.1.3 release BF. See http://trac.openmicroscopy.org/ome/ticket/13005

pwalczysko:
@jburel : Stopping here - the PR was listed in trout paragraph in Standup - is this expected ? (the header says "develop" here ? Please relist

pwalczysko:
Thinking and checking CI, I think I was doing the right thing after all. There is no BF-develop-merge build running, so this must be the 
``BIOFORMATS-5.1-merge-build #994`` which I am using.

pwalczysko:
The multi-z multi-t workflow works fine, but it works just like if https://github.com/openmicroscopy/openmicroscopy/pull/4045 was not in... -> Stefans images do not work (the tiff images of ourselves do not work either, the Z/T info gets lost when going through Stefans scenario from https://github.com/openmicroscopy/openmicroscopy/pull/4045. Is this expected ?

pwalczysko:
This PR is possibly not even in a build as see Standup

```
IMPORTANT: BIOFORMATS-5.1-merge-daily was re-run at 6am. The merge CI servers (eel & trout) are bundled with yesterday’s merge build of Bio-Formats

```

pwalczysko:
Stopping the second time

pwalczysko:
Please relist

pwalczysko:
@jburel : What extent of testing is actually needed here ? Test_all_IJ_workflows ?

jburel:
The export as ome-tiff workflow
I will check the cellH5

jburel:
@pwalczysko @bramalingam: modified the way position is set.

bramalingam:
Significant improvement in performance, but there seems to be a bug in the Z,C,T positions of the ROI's. Except for the first plane all other planes, have a single coordinate (which points at the slice number in the stack) and this bug seems to populate the ROI's in all the planes(irrespective of the position properties).

jburel:
@bramalingam: I wonder if it is not the case prior to my change. Could you verify?

bramalingam:
Compared the Develop branch and the current commit.! Both have a 3 coordinate system for the ROI's(for CZT) and the display of ROI's is still a problem (but is not a regression). But on closer observation, it showed that the current commit does not have continuous ROI coordinates (0,0,1 ; 0,0,2.....etc) , when compared to the develop branch and this definitely looks like a regression.!

bramalingam:
Tested the latest commit locally and the cellh5 image (when opened as a hyper stack), sets the positions of the ROI's correctly. Will re-test the same tomorrow, from the merge-build.

bramalingam:
Was able to reproduce results from yesterday. Good to merge.

stelfrich:
Oh.. I was a little late to the party, it seems.. There is still a problem with saving of the positional information, should I open a new PR?

jburel:
@stelfrich: what is the problem now?

stelfrich:
`TheC`, `TheZ`, and `TheT` are just not set for ROIs when saving a non-hyperstack image to OME-TIFF. It's the same problem with the `roi.getPosition()` or `roi.getCPosition()` et al, just vice versa..

To re-produce, open any image with 3 dimensions (I have tried one with several timepoints) and add ROIs to the overlay. When done, try to save that to OME-TIFF. The ROIs will not have `TheC`, `TheZ`, and `TheT` set and can hence not be loaded properly..

bramalingam:
@stelfrich: This is an ImageJ bug if am right, because getC,Z,TPosition on Overlay's returns a null in all cases. Could you please try using the ROIManager in your workflow and confirm if you could reproduce the bug?

stelfrich:
@bramalingam: I can confirm that the problem also exists with the ROIManager. It is again the general issue that ImageJ uses the property of a ROI that is returned by `getPosition()` instead of `getC,Z,TPosition()` for 3D images.

pwalczysko:
@jburel @bramalingam : I can confirm that with a
  1. bioformats_package.jar from Bioformats merge build  and
  2. insight-ij plugin from OMERO-5.1-merge-build

The problem as Stefan suggests above is fully present using ImageJ, the exported OME-TIFF has no ``TheZ``, ``TheT`` and ``TheC`` in it (see below).

```
<ROI ID="ROI:0:0" xmlns="http://www.openmicroscopy.org/Schemas/ROI/2015-01">
<Union>
<Shape ID="Shape:0:0" Text="0001-0051-0042">
<Rectangle Height="36.0" Width="34.0" X="25.0" Y="33.0"/>
<ROI ID="ROI:1:0" xmlns="http://www.openmicroscopy.org/Schemas/ROI/2015-01">
<Union>
<Shape ID="Shape:1:0" Text="0002-0025-0052">
<Rectangle Height="25.0" Width="37.0" X="34.0" Y="13.0"/>
<ROI ID="ROI:2:0" xmlns="http://www.openmicroscopy.org/Schemas/ROI/2015-01">
<Union>
<Shape ID="Shape:2:0" Text="0003-0046-0076">
<Rectangle Height="19.0" Width="21.0" X="66.0" Y="37.0"/>`
```

pwalczysko:
With the same setup as https://github.com/openmicroscopy/bioformats/pull/1954#issuecomment-139589989 I can confirm the bug when import to OMERO as described by Stefan in https://github.com/openmicroscopy/openmicroscopy/pull/4045#issuecomment-130983820 -> the ROIs have no positional info on them in OMERO (I have used ROI Manager)

jburel:
@pwalczysko: I will have a look at it to see if we can do something on our side

pwalczysko:
When I merged Stefan's branch ``stelfrich/fixRoiPositions`` and built the ``bioformats_package.jar`` locally, I did not get a remedy on the bug https://github.com/openmicroscopy/bioformats/pull/1954#issuecomment-139589989. 

Neither did the OMERO bug https://github.com/openmicroscopy/openmicroscopy/pull/4045#issuecomment-130983820 got solved when I merged ``stelfrich/fixRoiStoring`` and ``stelfrich/fixCompositeRoiImport`` to my openmicroscopy branch locally and built insight-ij.  In fact there was a regression (no ROIs were found on the image in OMERO) with this setup. 

In total, it seems that we simply do not have the fix, neither pending, unmerged or otherwise....

stelfrich:
> When I merged Stefan's branch stelfrich/fixRoiPositions and built the bioformats_package.jar locally, I did not get a remedy on the bug #1954 (comment).

That seems odd. I have built that branch just now and it properly writes TheC, TheZ, TheT for the ROIs. How did you do the import: via File > Import > Bio-Formats or just File > Open? The former works perfectly fine for me..

> Neither did the OMERO bug openmicroscopy/openmicroscopy#4045 (comment) got solved when I merged stelfrich/fixRoiStoring and stelfrich/fixCompositeRoiImport to my openmicroscopy branch locally and built insight-ij. In fact there was a regression (no ROIs were found on the image in OMERO) with this setup.

We will have to migrate the same logic that is developed here to OMERO. I just wanted to rule out individual problems with OMERO and stick to Bio-Formats first..

stelfrich:
> When I merged Stefan's branch stelfrich/fixRoiPositions [...]

Into which branch did you merge it?

pwalczysko:
@stelfrich :  No, I do not think the discrepancy between what you see and what I see is caused by different workflow (for me, also when everything is done via Bio-formats importer, no TheC.... is present).
Rather, the discrepancy will be in the branching, just as you say. What I did was:
```
git checkout develop
git fetch 
git rebase
```
this brings me to the status of the last commit being (as checked by git log) 
```
Date:   Fri Sep 11 15:08:54 2015 +0100

    Merge pull request #1972 from nylocx/fix_sphinx_babel
```
Then I created a new branch and merged
``git merge stelfrich/fixRoiPositions ``
Then ``ant clean jars tools``

pwalczysko:
@stelfrich : I am doing some further checks on my side, there might be a bit of misunderstanding.

pwalczysko:
We have found the source of the problem and we will open a fix later on tonight.

stelfrich:
Great! I will gladly test the fix :)

PR #1958, opened 2015-08-31
Add autogeneration logic to allow multiple readers per format

sbesson:
Certain formats like Olympus Slidebook have multiple readers. The current logic of FormatPageAutogen simply sums the number of readers defined in format-pages to count the total number of formats. This PR adds an extra field allowing to reference alternate readers in format-pages.txt and adds minimal logic to subtract duplicates in FormatPageAutogen.

To test this PR, check the autogenerated commit with this PR included lists 142 supported formats instead of 143 previously.


sbesson:
See https://github.com/snoopycrimecop/bioformats/commit/f2cdd1fd1ccf900bfdc7c6ab6f0b4a9f7c0b0c4b

melissalinkert:
Format count certainly looks fine.  It will still generate a separate page for each reader, though, correct?  Would it make more sense to have a single page, with each reader listed on it instead?

sbesson:
Last commit should now unify the slidebook pages and add support to point at multiple metadata pages (matching the readers source code links).

sbesson:
See https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/formats/3i-slidebook.html. The format page should link to the source code as well as the metadata pages for both slidebook readers.

hflynn:
The links on the metadata fields table are still not right:
SlideBook6Reader link is still going to /formats/simplepci-hcimage-tiff-metadata.html and there are two links (SlidebookReader and SlidebookTiffReader) both going to  https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/formats/3i-slidebook-metadata.html

The two metadata pages and the 3i SlideBook page all look fine though.

sbesson:
As discussed with @hflynn, we can use this PR to fix the broken links in the metadata list page. Excluding for 5.1.4.

PR #1959, opened 2015-09-01
Updated version history for 5.1.4

melissalinkert:
Usual disclaimers apply with respect to date and verbosity.

/cc @hflynn, @mtbc for general review
/cc @rleigh-dundee for C++ review/additions

### DIFF
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -1,6 +1,57 @@
 Version history
 ===============
 
+5.1.4 (2015 September 7)
+------------------------
+
+* Bug fixes, including:
+   - Command line tools
+       - fixed display of usage information
+   - Automated testing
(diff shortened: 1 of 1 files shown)


hflynn:
Building on staging now, Travis obviously wasn't green in time for it to be included in last merge build.

hflynn:
All formats correctly and reads fine to me

hflynn:
If the plan is to release the C++ stuff separately (later), the C++ point on here should probably be removed, otherwise good to go AFAIK

rleigh-dundee:
The C++ stuff isn't being released separately--it's in the same release as the java code.  The super build stuff is entirely separate and is not the C++ release.

joshmoore:
@hflynn : the code fixes to make it work on Windows are included, but @rleigh-dundee, we will have to be careful that no on thinks this means that there are necessarily c++ builds ready.

PR #1964, opened 2015-09-07
Support for .spe files and bug fixes

dgault:
Commits for 3 separate issues, introduction of a new reader Princeton Instruments .spe files and 2 bug fixes.

New features:
A new reader to add support for Princeton Instruments .spe files
Specification and sample files can be found on the related ticket:
https://trac.openmicroscopy.org/ome/ticket/12852

Bug Fixes:
ImageJ plugin - Exception when both 'group files' and 'open individually' options selected
https://trac.openmicroscopy.org/ome/ticket/12941

VoxelSize reading in MetaMorph 
https://trac.openmicroscopy.org/ome/ticket/12977

PR #1966, opened 2015-09-08
Incorrect VoxelSize in MetaMorph Reader

dgault:
To fix the physical Z size, the difference between the absolute Z
positions will need to be used.
MetaMorph handler already parses the z positions. The physical z size is
then calculated by subtracting the top and bottom values and dividing by
the number of steps.

Ticket details below:
https://trac.openmicroscopy.org/ome/ticket/12977

Splitting from https://github.com/openmicroscopy/bioformats/pull/1964

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -27,6 +27,7 @@ package loci.formats.in;
 
 import java.io.File;
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.text.DecimalFormatSymbols;
 import java.util.Arrays;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/MetamorphReader.java
```diff
@@ -778,6 +778,32 @@ else if (differentZs) {
       if (zDistances != null) {
         stepSize = zDistances[0];
       }
+      else {
+        Vector<Double> zPositions = new Vector<Double>();
+        Vector<Double> uniqueZ = new Vector<Double>();
```
mtbc: An `ArrayList` would suffice unless you need thread synchronization.

sbesson:
Note this PR is causing regressions on the Metamorph files of `test_images_good` - see https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-test_images_good/

mtbc:
Note that if you configure the build parameters you can use https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/ (presumably someday DEV too) to do quick tests of these -- it does something similar to what is described on http://www.openmicroscopy.org/site/support/bio-formats5.1/developers/commit-testing.html. Often the problem is just a little bug in the code but one may sometimes legitimately need to generate adjustments for the `.bioformats` files that these tests check against.

dgault:
Testing Steps:

- retrieve the following files mentioned in https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7864

nom1.nd
nom1_w1561 Em 600-50.TIF
nom1_w2488 Em 525-50.TIF

- Import the image series using default settings
- Check that PhysicalSizeZ in OME metadata matches expected value

Expected physical z size for:
1.5 µm

bramalingam:
Tested using ImageJ. Everything works as expected. Please merge.

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/615 is also green.

PR #1967, opened 2015-09-08
ImageJ plugin - Group files Exception 

dgault:
ImageJ plugin - Exception when both 'group files' and 'open individually' options selected

A test case and steps to reproduce are linked from the ticket.
https://trac.openmicroscopy.org/ome/ticket/12941

Splitting from https://github.com/openmicroscopy/bioformats/pull/1964

Issues included:
- IndexOutOfBoundsException when both options selected in Main Dialog
- For certain file formats the regular expression and 'file contains'
file patterns were not being implied
- Entering incorrect values into the Dimensions setting of the
FilePatternDialog resulted in a 'File is not of a supported format'
error

Solution:
- IndexOutOfBoundsException resolved in ImportProcess by reading the
group files option. When both options are selected then group files is
not automatically the inverse of ungroup files
- For file formats that have a fileGroupOption of FormatTools.MUST_GROUP
the second FilePatternDialog is skipped and replaced with an ImageJ
message box informing the user why file pattern options are not
available, stating "Image specifications state that files of this given
format cannot be handled separately and must be grouped according to
supporting metadata". A new importer option flag was added to test and
store the must group condition

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
@@ -492,7 +492,7 @@ public class ImportProcess implements StatusReporter {
 
     BF.status(options.isQuiet(), "Analyzing " + getIdName());
     baseReader.setMetadataFiltered(true);
-    baseReader.setGroupFiles(options.isGroupFiles());
+    baseReader.setGroupFiles(!options.isUngroupFiles() || options.isGroupFiles());
     if(options != null && !options.showROIs()){
         MetadataOptions mo = baseReader.getMetadataOptions();
         if(mo == null){
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
@@ -492,7 +492,7 @@ public class ImportProcess implements StatusReporter {
 
     BF.status(options.isQuiet(), "Analyzing " + getIdName());
     baseReader.setMetadataFiltered(true);
-    baseReader.setGroupFiles(!options.isUngroupFiles());
+    baseReader.setGroupFiles(options.isGroupFiles());
     if(options != null && !options.showROIs()){
         MetadataOptions mo = baseReader.getMetadataOptions();
         if(mo == null){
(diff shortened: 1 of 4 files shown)


jburel:
* The files attached to the e-mail sent by  Joaquim now import correctly. The 2 files are considered and the listed as 2 series.
A dialog pops up when the "Group files by similar names" is selected. Is that still expected?
This is in the only case it happens,
* using ```showinf```
Reading core metadata
filename = /Users/jmarie/Desktop/002002001.flex
Used files:
	/Users/jmarie/Desktop/002002001.flex
	/Users/jmarie/Desktop/002002002.flex


@dgault:
in the description, we also usually indicate steps required to test the PR
e.g. 
* Select image A
* Check the following options in dialog

dgault:
@jburel 
Firstly, my apologies for not fully including testing steps. 

- The first point regarding the dialog is I guess more of a user experience discussion point. The popup when "Group files by similar names" is selected is the expected outcome in this case (when the files used have been marked as 'Must Group'). 

The question is whether or not it is better to have that popup informing the user why the file stitching options are not being presented as normal, or if the popup is too intrusive then perhaps it would be best to leave it out all together and go straight to the series options screen. I've attached an image below to clearer demonstrate the 2 possible options. If there is a preference for the second option then the PR can be easily updated.

pr1967

- The second point regarding the filenames used I will look into and update if required

jburel:
@dgault: no problem at all, The description of the PR was already well documented

jburel:
Having an extra window is probably not a bad thing and does not break the "usual" ImageJ workflow (several windows can pop up when performing a task e.g. export as OME-TIFF)
The display of the text should be improved.

dgault:
Updated the text being displayed to try and make it clearer to the user that the file stitching options are not available for the selected files.

joshmoore:
Does this need to be relisted?

jburel:
I was planning to do it yesterday afternoon
I will take care of it either today or tomorrow for a final check

jburel:
* Typo: ```Stitiching``` => ```Stitching```
* ```\n``` should be after ```format```

jburel:
Text display is improved. typo fixed. Just noticed a missing dot at the end of the second sentence.
When this is fixed and travis is happy, it will be ready to be merged.

jburel:
@dgault thanks. Good to merge

PR #1965, opened 2015-09-08
Adding support for Princeton Instruments .spe data

dgault:
This is a new reader to add support for Princeton Instruments .spe files
Specification and sample files can be found on the related ticket:
https://trac.openmicroscopy.org/ome/ticket/12852

Splitting from https://github.com/openmicroscopy/bioformats/pull/1964

PR #1968, opened 2015-09-09
Container travis with findbugs

joshmoore:
gh-1965 was merged into the daily build and had priority HIGH
findbugs warnings. If findbugs had been activated in travis,
the build status wouldn't have allowed merging.

A possible next step would be to activate findbugs on every
compile.

**Note:** in order to speed up travis (after slowing it down with
findbugs), the `cpp` and and the `sphinx (pdf)` targets have been
removed. This permitted a move to the container-based travis
builds. These removed components are currently being tested by
jenkins.

See: https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-build/1009/findbugsResult/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ jdk:
 env:
   - BUILD=maven_findbugs
   - BUILD=cppwrap
-  - BUILD=sphinx
+  - BUILD=sphinx_html
   - BUILD=ant
 
 matrix:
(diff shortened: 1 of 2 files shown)


sbesson:
No conceptual veto to adding findbugs to Travis. However our organization currently struggles with the size of the matrix and the fact we cannot switch to container-based infrastructure (`cpp` + `sphinx` components).

A couple of orthogonal proposals:
- including `findbugs` as part of `mvn` ?
- dropping `cpp` and `sphinx` and moving to container-based infrastructure

.travis.yml
```diff
@@ -43,18 +45,10 @@ script:
...
-      env: BUILD=sphinx
-    - jdk: oraclejdk7
-      env: BUILD=sphinx
-    - jdk: openjdk6
-      env: BUILD=cpp
+      env: BUILD=sphinx_html
     - jdk: oraclejdk7
-      env: BUILD=cpp
-    - jdk: oraclejdk8
       env: BUILD=cppwrap
```
sbesson: This should be switched back to excluding `oraclejdk8` for `cppwrap`

sbesson:
Given the recent history of Travis and the fact the Bio-Formats builds are the current bottlenecks for the entire organization, I do not see a reason not to merge this PR (and revert it later if needed). The consequences of the removed components will have to be reviewed but the impact of this change should overall be an improvement in terms of individual PR validation and more generally in terms of the organization workflow. Thoughts @melissalinkert ?

melissalinkert:
Definitely fine with me.

PR #1971, opened 2015-09-10
Change the install path of cmake specific config files

nylocx:
to follow the general practice

I'm currently packaging the bioformats c++ libraries for my arch linux system and came across this. joshmoore (from the #ome IRC channel) said creating a pull request to get comments on this should be ok.
Some projects like to add a version number to the directory in the cmake dir. I hope that just using ome is ok.

### DIFF
--- a/cpp/lib/ome/bioformats/CMakeLists.txt
+++ b/cpp/lib/ome/bioformats/CMakeLists.txt
@@ -189,7 +189,7 @@ set(LIBRARY_HEADER ome/bioformats/CoreMetadata.h)
 configure_file(${PROJECT_SOURCE_DIR}/cpp/cmake/TemplateConfig.cmake.in
                ${CMAKE_CURRENT_BINARY_DIR}/ome-bioformats-config.cmake)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/ome-bioformats-config.cmake
-        DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/cmake)
+        DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/cmake/ome-bioformats)
 
 # Dump header list for testing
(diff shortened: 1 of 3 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-5.1-merge-push#80](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-push/80/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-push/80/consoleText) for more details.
Possible conflicts:
  - Conflict detection failed
    - None

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-5.1-merge-push#81](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-push/81/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-push/81/consoleText) for more details.
Possible conflicts:
  - Conflict detection failed
    - None

nylocx:
After doing a bit more research I think this PR needs some changes. From the cmake docs I got that the search path for config files is <prefix>/(lib/<arch>|lib|share)/cmake/<name>*/.
See: http://www.cmake.org/cmake/help/git-master/command/find_package.html#command:find_package
So if all the config files would be but into the same directory I still could not use
    find_package(ome-bioformats REQUIRED)
I would have to use
    find_package(ome-bioformats REQUIRED PATHS /usr/lib/cmake/ome)
which is really not a nice thing.
So I think it would be best to put the config files into separate subdirectories to …/cmake/ome-bioformats, …/cmake/ome-qtwidgets and …/cmake/ome-xml.
I will adjust the PR and wait for comments.

rleigh-dundee:
Thanks for the PR.  The install locations look sensible, and this looks good to merge.

[Not for this PR, just some general thoughts.]  We should probably switch to using `configure_package_config_file` from `CMakePackageConfigHelpers` to generate these files.  I wasn't aware of this module when I originally wrote these templates.  We should probably also add an `ome` module so that

```
find_package(ome COMPONENTS common xml bioformats)
```

will automatically find the ome-common, ome-xml and ome-bioformats packages.  We can copy what the `Qt5` package does I think.

Thanks,
Roger

PR #1972, opened 2015-09-10
Fix babel include for upcomming sphinx 1.3.2

nylocx:
With the current sphinx version 1.3.1 the pdf creation fails with a
undefined control sequence error. This bug is allready fixed upstream
but is not released yet. See: https://github.com/sphinx-doc/sphinx/issues/1923

The fix will require to define babel as an empty string, this is what
this commit is for.


sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-docs/1063/ remains green with this change included. Thanks @nylocx.

PR #1974, opened 2015-09-14
Roi position

jburel:
This PR fixes the problem reported by @pwalczysko see comment
https://github.com/openmicroscopy/bioformats/pull/1954#issuecomment-139589989

Follow the testing instruction described in the comment above.
cc @pwalczysko @stelfrich 

Note that the OMERO part has not yet been updated.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -475,6 +475,10 @@ public class ROIHandler {
         } else if (imp.getNSlices() == 1 && imp.getNFrames() == 1) {
           c = pos-1;
         }
+        if (t > imp.getNFrames()-1 || c > imp.getNChannels() -1 ||
+           z > imp.getNSlices()-1) {
+          continue;//
+        }
       }
       if (ijRoi.isDrawingTool()) {//Checks if the given roi is a Text box/Arrow/Rounded Rectangle
         if (ijRoi.getTypeAsString().matches("Text")) {
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -453,6 +453,7 @@ public class ROIHandler {
     int roicount = root.sizeOfROIList();
     int cntr = roicount;
 
+    ImagePlus imp = WindowManager.getCurrentImage();
     for (int i = 0; i < rois.length; i++) {
 
       String polylineID = MetadataTools.createLSID("Shape", cntr, 0);
@@ -461,16 +462,21 @@ public class ROIHandler {
       int c = ijRoi.getCPosition()-1;
       int z = ijRoi.getZPosition()-1;
       int t = ijRoi.getTPosition()-1;
-      ImagePlus imp = WindowManager.getImage(ijRoi.getImageID());
-      int pos = ijRoi.getPosition();
-      if (imp.getNChannels() == 1 && imp.getNSlices() == 1) {
-        t = pos-1;
-      } else if (imp.getNChannels() == 1 && imp.getNFrames() == 1) {
-        z = pos-1;
-      } else if (imp.getNSlices() == 1 && imp.getNFrames() == 1) {
-        c = pos-1;
+      ImagePlus image = WindowManager.getImage(ijRoi.getImageID());
+      if (image == null) {
+        image = imp; //pick the current image in that case
       }
-      if (ijRoi.isDrawingTool()){//Checks if the given roi is a Text box/Arrow/Rounded Rectangle
+      int pos = ijRoi.getPosition();
+      if (imp != null) {
+        if (imp.getNChannels() == 1 && imp.getNSlices() == 1) {
+          t = pos-1;
+        } else if (imp.getNChannels() == 1 && imp.getNFrames() == 1) {
+          z = pos-1;
+        } else if (imp.getNSlices() == 1 && imp.getNFrames() == 1) {
+          c = pos-1;
+        }
+      }
+      if (ijRoi.isDrawingTool()) {//Checks if the given roi is a Text box/Arrow/Rounded Rectangle
         if (ijRoi.getTypeAsString().matches("Text")) {
           if (ijRoi instanceof TextRoi){
             store.setLabelID(polylineID, cntr, 0);
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -30,6 +30,7 @@ package loci.plugins.util;
 import ij.IJ;
 import ij.ImagePlus;
 import ij.Prefs;
+import ij.WindowManager;
 import ij.gui.EllipseRoi;
 import ij.gui.Line;
 import ij.gui.OvalRoi;
@@ -460,6 +461,15 @@ public class ROIHandler {
       int c = ijRoi.getCPosition()-1;
       int z = ijRoi.getZPosition()-1;
       int t = ijRoi.getTPosition()-1;
+      ImagePlus imp = WindowManager.getImage(ijRoi.getImageID());
+      int pos = ijRoi.getPosition();
+      if (imp.getNChannels() == 1 && imp.getNSlices() == 1) {
+        t = pos-1;
+      } else if (imp.getNChannels() == 1 && imp.getNFrames() == 1) {
+        z = pos-1;
+      } else if (imp.getNSlices() == 1 && imp.getNFrames() == 1) {
+        c = pos-1;
+      }
       if (ijRoi.isDrawingTool()){//Checks if the given roi is a Text box/Arrow/Rounded Rectangle
         if (ijRoi.getTypeAsString().matches("Text")) {
           if (ijRoi instanceof TextRoi){
--- a/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
@@ -112,20 +112,20 @@ public class ROIHandler {
         manager = new RoiManager();
       }
 
-      for (int roiNum=0; roiNum<roiCount; roiNum++) {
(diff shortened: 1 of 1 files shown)


stelfrich:
This fixes the storing if ROIs have been added after loading of an image. When loading ROIs from an image and trying to save it, however, a NPE is thrown because the images are not set properly in the ROIs. See jburel/bioformats#5 for fix.

pwalczysko:
Agree with @stelfrich on the first part - the fix looks good with the workflow described in the header. 

@stelfrich : Cannot get to the NPE you are reporting, I must be missing something.
  - are you using Overlays to add ROIs ?
  - are you using some special image ?
  - are your ROIs on the image coming from another software, or OMERO or are they drawn in ImageJ in previous round and exported using B-F ?
  - at which point exactly are you experiencing the NPE, when saving the image ? Or the ROIs in ImageJ ? 
  - if the NPE demonstrates when saving the image, how are you saving it ? (File > Save As or B-F exporter or....)

Thanks a lot for your efforts here.

pwalczysko:
@stelfrich : I think I got the NPE:
  - open an image which you have previously drawn ROIs on and exported it using B-F exporter
  - the ROI Manager will open and list the ROIs on the image
  - close the image
  - open another image
  - use B-F Exporter to save it again (expecting the ROIs from the Manager from the other image to be saved with it)
  - observe NPE

```
	at loci.plugins.util.ROIHandler.saveROIs(ROIHandler.java:466)
	at loci.plugins.out.Exporter.run(Exporter.java:648)
	at ij.plugin.filter.PlugInFilterRunner.processOneImage(PlugInFilterRunner.java:263)
	at ij.plugin.filter.PlugInFilterRunner.<init>(PlugInFilterRunner.java:112)
	at ij.IJ.runUserPlugIn(IJ.java:201)
	at ij.IJ.runPlugIn(IJ.java:163)
	at ij.Executer.runCommand(Executer.java:131)
	at ij.Executer.run(Executer.java:64)
	at java.lang.Thread.run(Thread.java:695)
```

stelfrich:
@pwalczysko: Exactly..

pwalczysko:
@stelfrich : Yes, agreed, could reproduce as well, this simplifies the workflow, thanks a lot.

jburel:
@stelfrich @pwalczysko: if no image set in roi, I will use the active image. (in the spirit of https://github.com/jburel/bioformats/pull/5)
This will not be correct for one of the cases described by @pwalczysko.
I am not sure we can do much more if the rois don't have ref to the image

stelfrich:
@jburel: imho, there's not much that can be done in the one case where the RoiManager contains ROIs with lost references for different images..

jburel:
@stelfrich: the only thing I can think of is adding an extra control in case the position is greater than the number of slices for example.

stelfrich:
@jburel: But what do you do with the ROIs for which that check fails (that have position < number of slices). I think it would make sense to add them either? So you'd have to check all ROIs and then decided, based on the outcome, if the ROIs should be stored or not.. that sounds reaaally odd..

jburel:
so I will leave it as it is

pwalczysko:
Both workflows mentioned above https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-140335456 and https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-140335527 work fine.

pwalczysko:
Note that when I use File > Open when opening the image the second time in workflow  https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-140335456, I still get

screen shot 2015-09-16 at 12 21 29

and 
```
ImageJ 1.49o; Java 1.6.0_65 [64-bit]; Mac OS X 10.10.3; 32MB of 6000MB (<1%)
 
	at loci.plugins.out.Exporter.run(Exporter.java:359)
	at ij.plugin.filter.PlugInFilterRunner.processOneImage(PlugInFilterRunner.java:263)
	at ij.plugin.filter.PlugInFilterRunner.<init>(PlugInFilterRunner.java:112)
	at ij.IJ.runUserPlugIn(IJ.java:201)
	at ij.IJ.runPlugIn(IJ.java:163)
	at ij.Executer.runCommand(Executer.java:131)
	at ij.Executer.run(Executer.java:64)
	at java.lang.Thread.run(Thread.java:695)
```

But this is expected, isn't it ? (I have opened an image via File > Open while there were some ROIs in ROI Manager from last time (from another image) and then used B-F export

pwalczysko:
Unfortunately, I have the problems described in https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-140712870 also if I do
   - start IJ
   - File > Open > <some-multi-t-tiff-image>
   - Draw couple of ROIs on different planes and add each to the ROI manager
   - Use B-F exporter to export the image with the ROIs
   - Observe the warning and NPE as in https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-140712870

This does not seem acceptable -> we have to keep the File > Open workflow sound I think.

pwalczysko:
Correction: The File > Open problem https://github.com/openmicroscopy/bioformats/pull/1974#issuecomment-140728699 is specific to ome-tiff. Also this problem is present in 5.1.0 release BF as well. 
So nothing what this PR would do wrongly. 

All major workflows are passing here. The only question is the handling of the medley of regions from another images hanging about in ROI Manager. At the moment with this PR, the regions will be saved onto the image, even if they do not fit.

pwalczysko:
Works as expected. Tested ROI Manager and Overlays, always at least 2 rounds of B-F Export (with ROIs). 
Tested the image submitted by @stelfrich , with a newly created tiff as described by @stelfrich , with the ``test_images_good/tiff/QA images/....tiff``, with 4D ome-tiff, with 3D (multi-t and multi-z) dv, with multi-t dv, with multi-z dv.

stelfrich:
Also works fine for me for B-F-only cases.

However, TheC, TheZ, and TheT of shapes are not consistent when loading from OMERO and from OME-TIFF. It seems to me that the assumption in [ROIHandler.java:L361-363](https://github.com/jburel/bioformats/blob/roi-position/components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java#L361-L363) is not violated when loading shapes from OMERO. For my tests, they seem to be 1-based..

jburel:
@stelfrich: could you clarify?
Could it be that the OME-TIFF was created with the wrong z, t or c?

stelfrich:
@jburel: The workflow that results in the different indices:

1. Open image with multiple slices
2. Add ROIs to overlay
3. Plugins > OMERO > Save Image(s) to OMERO
5. File > Save as > OME-TIFF from the same ImageJ instance
6. Import image from file via File > Import Bio-Formats -> ROIs have correct position with TheC/Z/T 1-based
7. Import image from OMERO via Plugins > OMERO > Connect to OMERO and View in ImageJ -> ROIs have shifted position TheC/Z/T are 0-based

Where I am running this PR for Bio-Formats and https://github.com/openmicroscopy/openmicroscopy/pull/4170 for OMERO-insight.ij

jburel:
@stelfrich: thanks. I will have a look at it tonight

pwalczysko:
@stelfrich : Thank you very much for the workflow. According to my tests I cannot find any discrepancy between ROI positioning in OMERO and IJ (when opened from OMERO).  Could you please provide following information as well:
  - Please submit the image which you are using via QA
  - How do you get the OMERO-insight-ij plugin, do you checkout the branch https://github.com/openmicroscopy/openmicroscopy/pull/4170 and build locally ? Actually, that is not what we do, we have a merge build - would you try your workflow with https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/lastSuccessfulBuild/artifact/artifacts/ please ? 
  - is the image you are using stack or hyperstack
  - on what plane did you draw the ROI(s), was it z = ?, t = ? , c = ?
  - how big was your image ?
  - how did you create it ?
> ROIs have shifted position TheC/Z/T are 0-based

In my testing here, both via B-F only and via OMERO -> IJ (or IJ -> OMERO -> IJ) I always get the ``TheZ/T/C`` in the ome-xml as 0-based, in all cases. But the ``real`` Z, C, and T info is always right and matching between OMERO and IJ.

pwalczysko:
@stelfrich : One more important thing: The Overlay information in ImageJ is not stored properly on hyperstacks, so if you draw the ROI in IJ and add to overlay on a multi C, T, Z image and then import to OMERO, the T info will be lost in OMERO (because it was not properly created in IJ in the first place)  and all the T will be at 0). This is a known issue (but possibly not what you mean here, hopefully).

pwalczysko:
@stelfrich : Sorry, gave you just the B-F link, the merge build we use for OMERO is https://ci.openmicroscopy.org/job/OMERO-5.1-merge-build/ICE=3.5,jdk=8_LATEST,label=trout/

stelfrich:
@pwalczysko: Everything's fine with the latest CI builds of OMERO and Bio-Formats. Sorry for the confusion! And thanks for sticking with me :+1: 

> so if you draw the ROI in IJ and add to overlay on a multi C, T, Z image and then import to OMERO, the T info will be lost in OMERO

I can't confirm that one, btw.. I have just tested that case with a multi-T, multi-C image and the ROIs have proper TheC, Z, T in OMERO.

pwalczysko:
@stelfrich : Great to hear, thanks for all the efforts, reports, testing and PRs. 
Re the Overlay -> T problem in OMERO issue - yes, we are on the same page as well, I can confirm that a multi-T, multi-C image will work fine (e.g. your .tff you submitted will work, also a C = 2, T > 2 and Z = 1 dv image will work) but a C > 1 T > 1 Z > 1 will NOT work (and especially in T dimansion).  This is a ``known problem`` which is in the workflow since the beginning, and starts with IJ. If you go Image > Overlay > List Elements, you can see in the end of the line which is belonging to your region you just added to overlay output like:
```
0	null	Straight Line	140	197	271	130	yellow	none	0	3	0	0	0
```
In IJ, this output is sometimes right (for first two cases named above) and sometimes wrong (for the last case named above) in the sense that T is ``zero`` even if you know and see that it should be something else.

stelfrich:
@pwalczysko: I was aware of that edge-case.. but no one is complaining about that peculiarity on ImageJ side, so I wouldn't bother too much for now.

PR #1976, opened 2015-09-15
Flex NPE

joshmoore:
See: https://trello.com/c/1Ik5PRc6/300-idr-npe-in-flexreader


joshmoore:
Assuming this is at least partially useful even if further fixes may be needed.

PR #1977, opened 2015-09-15
Disable travis on regions

joshmoore:
Due to the requirement for having schema files and
similar deployed for testing, travis is being disabled
on this branch.


joshmoore:
The goal is to have travis **working** on the spaces. Closing.

PR #1978, opened 2015-09-15
Add PlaneRegion.h to CMakeLists.txt

nylocx:
The PlaneRegion.h header file was missing in the CMakeList.txt which
caused the file not to be installed via make install. Wich in turn
resulted in "No such file or directory" compiler errors if one tried to
use the installed libraries.


joshmoore:
Thanks, @nylocx

rleigh-dundee:
This looks good to merge.  Thanks.

PR #1979, opened 2015-09-15
Gatan reader

jburel:
Fix issue when parsing the metadata
to test:
* Try to import file from QA 11392

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -484,7 +484,7 @@ public class GatanReader extends FormatReader {
 
         if (labelString.equals("Scale")) {
           if (value.indexOf(",") == -1) {
-            pixelSizes.add(new Double(value));
+            pixelSizes.add(f.parse(value).doubleValue());
           }
         }
         else if (labelString.equals("Units")) {


sbesson:
From a functional point of view this PR works as expected since the QA file fails to open with Bio-Formats 5.1.3 but opens properly with 5.1.4-SNAPSHOT including this PR. Looking at the metadata read

```
      <Pixels BigEndian="false" DimensionOrder="XYZTC" ID="Pixels:0" Interleaved="false" PhysicalSizeX="0.01255588699132204" PhysicalSizeXUnit="µm" PhysicalSizeY="0.05" PhysicalSizeYUnit="µm" PhysicalSizeZ="5.0" PhysicalSizeZUnit="µm" SignificantBits="32" SizeC="1" SizeT="1" SizeX="4096" SizeY="4096" SizeZ="1" Type="uint32">
```

having such a difference between the physical sizes in X and Y is surprising but may relate to our official lack of support for `dm4` format even if the `GatanReader` opens them /cc @melissalinkert

sbesson:
Since additional to-dos are captured in https://trello.com/c/2y71bKRs and both full-repository jobs have been green with this PR included, I would merge this PR.

martlj:
Thanks for the quick responses to this. The physical sizes of the pixels are definitely not intended to be as they are read in the metadata snippet (it should be 0.01255um in X and Y and 0.05um in Z). I checked the original metadata via the bioformats import and it doesn't look as though the x and y resolution are explicitly given, are they calculated via the FOV and pixel numbers?

PR #1980, opened 2015-09-15
Use netcdf.version variable

sbesson:
See https://trello.com/c/srr1vUFa/22-update-netcdf-dependency

This PR bumps the netcdf dependency to 4.3.22 (should still be Java 6 compatible) and uses a version parameter to simplify further updates.

### DIFF
--- a/ant/global.properties
+++ b/ant/global.properties
@@ -26,3 +26,4 @@ objenesis.version = 2.1
 minlog.version = 1.2
 testng.version = 6.8
 guava.version = 17.0
+netcdf.version = 4.3.19
--- a/ant/toplevel.properties
+++ b/ant/toplevel.properties
@@ -18,7 +18,7 @@ merged-docs.classpath = ${lib.dir}/jgoodies-forms-1.7.2.jar:\
                         ${lib.dir}/logback-classic-${logback.version}.jar:\
                         ${lib.dir}/logback-core-${logback.version}.jar:\
                         ${lib.dir}/native-lib-loader-2.0.2.jar:\
-                        ${lib.dir}/netcdf-4.3.19.jar:\
+                        ${lib.dir}/netcdf-${netcdf.version}.jar:\
                         ${lib.dir}/jhdf5-13.06.2.jar:\
                         ${lib.dir}/omero_client.jar:\
                         ${lib.dir}/slf4j-api-${slf4j.version}.jar:\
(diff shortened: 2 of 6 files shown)


sbesson:
Excluding for now, as I need to update the OMERO Ivy logic to consume edu.ucar artifacts from the Unidata Maven repository.

sbesson:
Unexcluding for review.

sbesson:
Discussed with @melissalinkert, I am not sure this PR actually brings much overall. Despite `netcdf 4.3.22` being in Maven Central, its parent POM `thredds-parent 4.3.22` is not. Removing the `unidata.releases` repositories from all `pom.xml`, cleaning up `~/.m2/repositories/edu*` and running `$ mvn` results in:

```
[ERROR] Failed to execute goal on project formats-gpl: Could not resolve dependencies for project ome:formats-gpl:jar:5.1.5-SNAPSHOT: Failed to collect dependencies at edu.ucar:netcdf:jar:4.3.22: Failed to read artifact descriptor for edu.ucar:netcdf:jar:4.3.22: Could not find artifact edu.ucar:thredds-parent:pom:4.3.22 in ome.snapshots (http://artifacts.openmicroscopy.org/artifactory/ome.snapshots) -> [Help 
```

I'll reduce this PR to a simple cleanup to use `${netcdf.version}`.

hflynn:
Assuming https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/1084/warnings3Result/ is related to this, can you make sure that is fixed at the same time if it will otherwise remain an issue, please? Thanks.

sbesson:
Last commit should also fix the merge docs warnings since we are not changing the version of the JAR anymore

hflynn:
:+1:

jburel:
* in components/formats-gpl/src/loci/formats/services/NetCDFServiceImpl.java version variable is not used.
* Do we also want to improve the handling of the version in java-library.txt?

sbesson:
Propagating build properties inside the Java classes themselves or the documentation requires some infrastructure. In this first case, I do not even think we have a proper solution (cf version number bump in the UpgradeChecker). I would be inclined to drive this maintenance via the documentation as the linkcheck will warn us if the linked JAR are not modified. Concrete proposal would be to remove the line https://github.com/openmicroscopy/bioformats/blob/v5.1.4/components/formats-gpl/src/loci/formats/services/NetCDFServiceImpl.java#L66. Thoughts?

sbesson:
Discussed with @melissalinkert, closing this ticket in favor of a broader discussion about the HDF5 library to use in Bio-Formats 5.2.0 - see https://trello.com/c/IsKUcIPR.

PR #1981, opened 2015-09-16
Adding RLE compression support for BMP's

dgault:
Adding support for RLE_8 and RLE_4 compression formats

Bug: BMP Reader failed to read a specific image
https://trac.openmicroscopy.org/ome/ticket/12971

This was due to the compression encoding RLE 8, which is not currently supported
OpenBytes method has been modified to deal with RLE 8 and RLE 4 compression encodings
Raw data is handled as before and is unaffected

Test Steps - 
1. Download the bmp (foo.bmp) attached to the trac ticket https://trac.openmicroscopy.org/ome/ticket/12971
2. Import using bio formats bmp reader
3. Compare the resulting image with the correct and incorrect expected images in the ticket

Other useful links -
To read more about RLE encoding the following links may be helpful:
https://msdn.microsoft.com/en-us/library/windows/desktop/dd183383(v=vs.85).aspx
http://www.binaryessence.com/dct/en000073.htm

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/BMPReader.java
+++ b/components/formats-bsd/src/loci/formats/in/BMPReader.java
@@ -149,28 +149,97 @@ public class BMPReader extends FormatReader {
     in.skipBytes(rowsToSkip * pad);
 
     int effectiveC = palette != null && palette[0].length > 0 ? 1 : getSizeC();
-    for (int row=h-1; row>=0; row--) {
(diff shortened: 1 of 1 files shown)


melissalinkert:
Looks fine to me, and has been in green builds.  PR opened to add the configuration: https://github.com/openmicroscopy/data_repo_config/pull/22 (the file has already been copied).

PR #1984, opened 2015-09-17
Allow fake plates to have multiple screens, including none.

ximenesuk:
This PR adds a flag to the fake filename, `screens=n` to create multiple screens, in particular `screens=0` will allow for no containing Screen to be created when a Plate is created. The omission of the flag maintains the current behaviour, to create a containing Screen, ie it is equivalent to `screens=1`.

Is this sufficient or are there other areas of the codebase that need to accommodate this change? If this is reasonable then there will also need to be a minor change to the docs.

To test, via OMERO,
```
$ touch "SPW&plates=1&plateRows=1&plateCols=1&fields=1&plateAcqs=1&screens=0.fake"
$ omero import "SPW&plates=1&plateRows=1&plateCols=1&fields=1&plateAcqs=1&screens=0.fake"
```
should import a Plate with no Screen. Importing files
```
$ touch "SPW&plates=1&plateRows=1&plateCols=1&fields=1&plateAcqs=1.fake"
$ touch "SPW&plates=1&plateRows=1&plateCols=1&fields=1&plateAcqs=1&screens=1.fake"
```
should each import single Plates each within a Screen, while
```
$ touch "SPW&screens=2&plates=2&plateRows=1&plateCols=1&fields=1&plateAcqs=1.fake"
```
should import two Screens each with two Plates.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -901,7 +901,7 @@ public class FakeReader extends FormatReader {
     if (screens==0) {
       ome = xml.createPopulatedPlate(plates, rows, cols, fields, acqs);
     } else {
-      ome = xml.createPopulatedScreen(plates, rows, cols, fields, acqs);
+      ome = xml.createPopulatedScreen(screens, plates, rows, cols, fields, acqs);
     }
     getOmeXmlMetadata().setRoot(new OMEXMLMetadataRoot(ome));
     // copy populated SPW metadata into destination MetadataStore
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -93,7 +93,7 @@ import ome.units.UNITS;
  *  <li>showinf '32bit-floating&amp;pixelType=float&amp;sizeZ=3&amp;sizeC=5&amp;sizeT=7&amp;sizeY=50.fake'</li>
  *  <li>showinf '64bit-floating&amp;pixelType=double&amp;sizeZ=3&amp;sizeC=5&amp;sizeT=7&amp;sizeY=50.fake'</li>
  *  <li>showinf 'SPW&amp;plates=2&amp;plateRows=3&amp;plateCols=3&amp;fields=8&amp;plateAcqs=5.fake'</li>
- *  <li>showinf 'Plate&amp;screen=0&amp;plates=1&amp;plateRows=3&amp;plateCols=3&amp;fields=8&amp;plateAcqs=5.fake'</li>
+ *  <li>showinf 'Plate&amp;screens=0&amp;plates=1&amp;plateRows=3&amp;plateCols=3&amp;fields=8&amp;plateAcqs=5.fake'</li>
  * </ul></p>
  */
 public class FakeReader extends FormatReader {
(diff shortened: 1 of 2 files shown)


melissalinkert:
Looks sensible to me on initial reading.  Would it make sense to have e.g. ```screen=5``` create 5 screens with the specified plate dimensions instead of just one?

joshmoore:
I'd be for 5 screens each with new plates. Also, in the spirit of https://twitter.com/sempf/status/514473420277694465, will `plate=0` do what @pwalczysko would expect?

ximenesuk:
@joshmoore if @pwalczysko expects a lizard he will be disappointed. From local testing `plates=0` will produce separate Images if I remember correctly as it would not be recognised as being SPW. I'll switch `screen` to `screens` and allow multiple Screens. Thanks both.

ximenesuk:
@melissalinkert @joshmoore Quick question... Is it possible for a single file(set) to contain multiple Screens? Here there is just one file being uploaded which effectively ties the multiple Screens together in a sort of MSF (Multi-Screen File) In fact the underlying Images form a MIF as I found out when deleting individual Screens (Images were orphaned). Given that the indexing (to create unique IDs) needs to use both the Screen index and the Plate Index I thought it worth checking I wasn't creating some impossible FrankenScreen.

melissalinkert:
It should definitely be possible from the model perspective, though I don't think we've ever really tested it in earnest, so there could well be some surprises.

ximenesuk:
I'm closing this temporarily while I'm away. I push further commits and rebase if necessary once I'm back.

melissalinkert:
@ximenesuk: do you want this re-opened for 5.1.5 consideration?

ximenesuk:
@melissalinkert I have some work not yet pushed that allows for multiple screens. But, before I pushed it I wanted to check a couple of behaviours of the existing reader, ie without my PR. I noticed that is `plateAcqs=0` or is absent then the result is a single orphaned image. Is that the expected/intended behavious? If so there is a lot of code setting up the plates, wells, etc for that fairly trivial result.

The second question is if `plateAcqs=2`, or any other number greater than one, a series of runs are created. In those runs the corresponding wells have the same ids. That makes sense to me - it's the same well - but I wanted to double check that fitted with expectations.

ximenesuk:
This final commit allows for multiple screens without chnaging existing behaviour as queried in my comment above. It doesn't fix any object names, ie should the screen id be part of other object names?

jburel:
@ximenesuk: I will add the id as part of the name. This might make things easier for testing/querying.

jburel:
I set the value to ```screen=-1``` 
The import goes ahead without error but nothing in the display.

ximenesuk:
I didn't check for negative numbers as that reflects the other settings, except `plateAcqs`, but I can check that and then force to 1 (or 0).

jburel:
I will force it to 0 in that case.

sbesson:
We might want to update the unit tests under `FakeReaderTest` to cover the new semantics. That can be done in a separate PR.

ximenesuk:
@sbesson The tests are fairly minimal but I can easily add one to test for two screens, which parallels that for two plates. I will look at adding one for no screen.

ximenesuk:
@sbesson re https://github.com/openmicroscopy/bioformats/pull/1984#issuecomment-145765492 these tests will need to be added later. The unit tests use `generateScreen` from https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/tools/FakeImage.java and so this class would need some work too. If this PR is merged than that work can be done later.

sbesson:
Retested last commit. Works as expected. Merging this and we can have RFEs in  follow-up PRs.

PR #1985, opened 2015-09-18
Schema publication refactoring

sbesson:
This PR starts streamlining the schema release process achieved via the `publish` script to allow to compute Git diffs of the public schema content

Major features of this PR are:
- remove the `XMLschemas` generation - we are not regenerating legacy schema files
- remove legacy schemas from being listed on the main index pages
- remove `web-extras`
- remove the `Schemas` namespace in the target directory to simplify diffs
- allow to specify an output directory - defaulting to `published`

To test this PR
- clone https://github.com/sbesson/schemas which is a Git snapshot of the current live schema
- run `./publish /path/to/schemas`
- compute the `git diff` locally to see the difference

### DIFF
--- a/components/specification/publish
+++ b/components/specification/publish
@@ -293,8 +293,6 @@ done
 # $2=year (may be blank)
 # $3=filter (current or legacy)
 # $4=verbose (true or false)
-# $5=schema set (current or legacy)
-# $6=schema location (current=Schemas, legacy=XMLschemas)
 createlink() {
     name=$(basename "$1")
     release="$2"
@@ -302,39 +300,19 @@ createlink() {
     if [ "$shortname" = "ome" ]; then
         shortname="OME"
     fi
-    if [ "$5" = "legacy" ]; then
-        reldate="2003 ($2)"
+    if [ -z "$2" ]; then
+        release=$(mostrecentversion "$name")
+        reldate=$(parsedate "$release")
     else
-        if [ -z "$2" ]; then
-            release=$(mostrecentversion "$name")
-            reldate=$(parsedate "$release")
-        else
-            reldate=$(parsedate "$2")
-        fi
+        reldate=$(parsedate "$2")
     fi
-    if [ "$5" = "current" ]; then
-        relversion=$(schemaversion "$name" "$release")
-    else
-        relversion=$(schemaversion "$name" "2003-$release")
-    fi
+    relversion=$(schemaversion "$name" "$release")
     reldesc=" The $(schemadesc "$name")"
     if [ "$4" = "false" ]; then
         reldesc=""
     fi
-    if [ "$5" = "current" ]; then
-        link=$(find $root -name "$name" | grep "$release" | head -n1)
-        link=$(echo "$link" | sed -e "s;$root/;;")
-    else
-        link=$(find published/XMLschemas -name "$name" | grep "$release" | head -n1)
-        link=$(echo "$link" | sed -e 's;published/XMLschemas/;;')
-        if [ "$6" = "current" ]; then
-            if [ -z "$2" ]; then
-                link="../XMLschemas/$link"
-            else
-                link="../../XMLschemas/$link"
-            fi
-        fi
-    fi
+    link=$(find $root -name "$name" | grep "$release" | head -n1)
+    link=$(echo "$link" | sed -e "s;$root/;;")
     if [ -n "$2" ]; then
         link=$(echo "$link" | sed -e "s;^${shortname}/;;")
     fi
@@ -420,7 +398,7 @@ EOF
 EOF
 
     for schema in $(find $root -name '*xsd' | grep -v "Samples" | xargs $XARGSOPTS -n1 basename | sort | uniq ); do
-        createlink "$schema" "" "current" "true" "current" "$1"
+        createlink "$schema" "" "current" "true"
     done
 
     cat <<EOF
@@ -430,7 +408,7 @@ EOF
 EOF
 
     for schema in $(find $root -name '*xsd' | grep -v "Samples" | xargs $XARGSOPTS -n1 basename | sort | uniq); do
-        createlink "$schema" "" "legacy" "true" "current" "$1"
+        createlink "$schema" "" "legacy" "true"
     done
     cat <<EOF
 </ul>
@@ -509,7 +487,7 @@ EOF
 EOF
 
     for year in $(find $root -name "$schema" | grep -v "Samples" | xargs $XARGSOPTS -n1 dirname | xargs $XARGSOPTS -n1 basename | grep -v $devmarker | grep -v html | sort -r); do
-        createlink "$schema" "$year" "current" "false" "current" "$2"
+        createlink "$schema" "$year" "current" "false"
     done
     cat <<EOF
 </ul>
@@ -518,7 +496,7 @@ EOF
 EOF
 
     for year in $(find $root -name "$schema" | grep -v "Samples" | xargs $XARGSOPTS -n1 dirname | xargs $XARGSOPTS -n1 basename | grep -v $devmarker | grep -v html | sort -r); do
-        createlink "$schema" "$year" "legacy" "false" "current" "$2"
+        createlink "$schema" "$year" "legacy" "false"
     done
 
     if [ "$1" = "OME" ]; then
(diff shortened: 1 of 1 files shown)


mtbc:
It adds a `Transforms/units-conversion.xsl` which is untracked. Other than that, it adjusts the transforms, presumably bringing in @jburel's recent unpublished fixes.

mtbc:
```
...
‘samples/libxml2’ -> ‘/home/mtbc/src/schemas//Samples/libxml2’
‘samples/libxml2/Makefile’ -> ‘/home/mtbc/src/schemas//Samples/libxml2/Makefile’
‘samples/libxml2/README’ -> ‘/home/mtbc/src/schemas//Samples/libxml2/README’
‘samples/libxml2/example.c’ -> ‘/home/mtbc/src/schemas//Samples/libxml2/example.c’
HTML [2] -> /home/mtbc/src/schemas///home/mtbc/src/schemas/BinaryFile/index.html
./publish: 534: ./publish: cannot create /home/mtbc/src/schemas///home/mtbc/src/schemas/BinaryFile/index.html: Directory nonexistent
```

components/specification/publish
```diff
@@ -560,27 +528,12 @@ EOF
 EOF
 }
 
-for type in $(find "${root}/Schemas" -mindepth 1 -maxdepth 1 -type d | sed -e 's;^published/Schemas/;;'); do
+for type in $(find ${root} -mindepth 1 -maxdepth 1 -type d | sed -e "s;^${root}/;;"); do
```
mtbc: maybe make that extra `/` optional

mtbc:
Okay, not having the trailing space on the path worked better, though we should certainly tolerate it.

There are interesting changes in the diff like,
```diff
--- a/OME/index.html
+++ b/OME/index.html
@@ -7,7 +7,7 @@
 <body>
 <h1>Open Microscopy Environment OME Schemas</h1>
 <div class="head">
-<p>March 2015</p>
+<p>September 2015</p>
 </div>
 <div id="toc">
 <h2>Table of contents</h2>
@@ -37,12 +37,6 @@
 <li>September 2008 - namespace /2008-09/</li>
 <li>February 2008 - namespace /2008-02/</li>
 <li>June 2007 - namespace /2007-06/</li>
-<li>2003 (FC) - namespace /FC/ome.xsd</li>
-<li>2003 (RC6) - namespace /RC6/ome.xsd</li>
-<li>2003 (RC5) - namespace /RC5/ome.xsd</li>
-<li>2003 (RC4) - namespace /RC4/ome.xsd</li>
-<li>2003 (RC3) - namespace /RC3/ome.xsd</li>
-<li>2003 (RC2) - namespace /RC2/ome.xsd</li>
 </ul>
 <p>Information regarding schemas released before June 2007 is available at <a href="http://www.openmicroscopy.org/XMLschemas/">http://www.openmicroscopy.org/XMLschemas/</a>. This includes the schemas used by the original OME server.
 </p>
@@ -67,12 +61,6 @@
 <li>ome.xsd - September 2008, version 1 [<a href="2008-09/ome.xsd">2008-09</a>]</li>
 <li>ome.xsd - February 2008, version 2 [<a href="2008-02/ome.xsd">2008-02</a>]</li>
 <li>ome.xsd - June 2007, version 2 [<a href="2007-06/ome.xsd">2007-06</a>]</li>
-<li>ome.xsd - 2003 (FC), version 1 [<a href="../../XMLschemas/OME/FC/ome.xsd">FC</a>]</li>
-<li>ome.xsd - 2003 (RC6), version 1 [<a href="../../XMLschemas/OME/RC6/ome.xsd">RC6</a>]</li>
-<li>ome.xsd - 2003 (RC5), version 1 [<a href="../../XMLschemas/OME/RC5/ome.xsd">RC5</a>]</li>
-<li>ome.xsd - 2003 (RC4), version 1 [<a href="../../XMLschemas/OME/RC4/ome.xsd">RC4</a>]</li>
-<li>ome.xsd - 2003 (RC3), version 1 [<a href="../../XMLschemas/OME/RC3/ome.xsd">RC3</a>]</li>
-<li>ome.xsd - 2003 (RC2), version 1 [<a href="../../XMLschemas/OME/RC2/ome.xsd">RC2</a>]</li>
 </ul>
 </div>
 <hr/>
```
which may be reasonable and intended. We *do* use 2003-FC to transform from 2010-04 to 2007-06.

mtbc:
Incidentally, in the above diff,
```html
<p>Information regarding schemas released before June 2007 is available at <a href="http://www.openmicroscopy.org/XMLschemas/">http://www.openmicroscopy.org/XMLschemas/</a>. This includes the schemas used by the original OME server.
```
may bear on our recent discussion. :smiley:

sbesson:
@mtbc: the last commit should strip the trailing slash. With it I would expect `./publish /path/to/Schemas/` to work without issue. Re XMLschemas, as we are only generating the new set of schemas, removing the legacy hyperlinks while preserving the generic hyperlink to legacy schemas is the expected behavior.

mtbc:
The latest commits work to fix the trailing slash issue: the script works with and without it. My remaining question is just that of if it's okay to remove the link to the 2003-FC schema given how it's part of the transformation paths for later schemas. (If I go to XMLschemas/index.html that page doesn't list it either and includes a link to itself!)

sbesson:
@mtbc: I leave that decision up to you. If it makes sense, happy to add a special rule for the FC schema linking to the absolute URL instead of a relative symlink when approriate.

mtbc:
That would be most reassuring, thank you. (Feel free to add a comment doubting if we want it long-term!) It's fine that you get rid of all those RC ones of course.

sbesson:
Pushed

components/specification/publish
```diff
@@ -153,18 +153,23 @@ contains() {
...
 devmarker="dev"
-root=published
-
+root=${1:-published}
+root=${root%/}
 
 # Clean up
 
-rm -rf "$root"
+rm -rf "$root/*"
```
rleigh-dundee: Why not clean up the root directory properly?  The glob can leave junk around which it doesn't match.  Better to remove entirely and recreate as on the next line.  Why was this changed?

rleigh-dundee:
Not sure I see the reason behing removing the legacy `XMLschemas` generation.  The intent of this script is to generate the public structure for every schema we support, which includes these schemas.  What's the benefit of removing it?

components/specification/publish
sbesson: As mentioned in the PR description, the idea is to generate the output of this script into a Git repository to have diffs. In this case, we do not want to remove the `.git` subfolder minimally - maybe the existing `Documentation` subfolder in the mid-term. Arguably, this cleanup step could be moved outside the `publish` script which is really a build target and people could go with the following:

```
rm -rf target_path & ./publish target_path
```

sbesson:
@rleigh-dundee: as far as I am know, these schemas are clearly marked as legacy and nothing is appended to  them since all new schemas go to `Schemas/YEAR-DATE`. In the wider context of the refactoring of the schema release process, the goal is to make this `XMLschemas` folder read-only like old downloads pages. Removing legacy schemas regeneration also largely simplifies the Git worflow discussed with @mtbc for reviewing new schema deployment.

mtbc:
The diff looks much better to me now. I am not sure if it's correct but it certainly looks plausibly so. :+1: No other issues from my point of view.

components/specification/publish
rleigh-dundee: Maybe use a subdirectory in the git repository?  I don't think such logic belongs here.

rleigh-dundee:
@sbesson Regarding XMLschemas, do we really want to remove our capability of being able regenerate this structure?  What if the webserver data is ever lost (for any reason)?  We're then completely reliant upon backups (assuming we have them, and assuming they won't be overwritten before a problem is noticed); being able to regenerate the whole structure every time does have value and it does give us the safety and security of being able to regenerate and redeploy the entire lot should any unfortunate problems occur.

components/specification/publish
sbesson: I will update the description to create an intermediate repo when modifying the Git files.

sbesson:
@rleigh-dundee: there are actually many flaws in the current schema deployment process and this PR is one of the first steps trying to address it. Trying to move forward, my current thought would beto have a separate `publish_legacy` script or even `./publish legacy` if this exists that would allow us to regenerate the legacy schemas if needed. Thoughts?

rleigh-dundee:
@sbesson I definitely appreciate how horrible this workflow is; just wanted to be sure we weren't dropping something we might regret if we needed it.  Certainly nice to see the rsync removed.

If at the end of this script (or external in the jenkins job) we tar/zip the whole thing; that can then be used by a separate deployment job for the website and this can be part of the release workflow as an actual artifact.  This would contain the complete generated content in a single archive, and the latest release can then be consumed by the deployment job.

While we could certainly separate the process into legacy/non-legacy, I'm not sure why we want to make this distinction.  Why the cutoff between the two.  Why for example are we not considering e.g. 2009 schemas legacy?  (I know there's a split in the path naming; the question is if we are regenerating not only the latest, but all of them rather than half of them, then to avoid regressions it makes sense to continue to regenerate everything.)

sbesson:
Closing this for now. Can be revisited when we tackle https://trello.com/c/zkbHhnNO/31-schema-release-process.

PR #1987, opened 2015-09-22
xsd-fu: Remove OMERO templates and targets

rleigh-dundee:
Remove:

- OMERO templates (java only)
- OMERO-specific language/configuration properties
- OMERO-specific xsd-fu targets and functions

None of this code has been actively used in at least the last four years, and given the extensive refactoring which has taken place over this period, it is most unlikely that any of the OMERO source generation would work without a large amount of work.  Since there is no expectation of this being resurrected any time in the short or medium term, remove to reduce the clutter and ease maintenance of the code which is actively used.

Testing: None of this code is used in the Java or C++ builds, so it is not expected that any specific testing is required.  Check that there are no generated source changes since this change should not affect any of the bioformats metadata or model source generation for either C++ or Java.


sbesson:
Trying to run the `omero` targets low-level, it seems indeed that some work would be needed to make them usable. This change  cleans up unmaintained `xsd-fu` targets. Since all Bio-Formats daily jobs are green, barring any comment from @joshmoore, this change looks ready to merge to me.

Regarding mid-term plans, if we wanted to reinstantiate these targets and templates, we might consider resurrecting them directly to `openmicroscopy` and add a `xsd-fu` dependency to the OMERO build so that it could generate these templates. This might go along with (or follow) the general decoupling epic of the `xsd-fu/model` component.

joshmoore:
As discussed in Dundee and elsewhere, all makes sense. Merging.

PR #1989, opened 2015-09-24
Dicom max pixel value bug

dgault:
Bug: DICOM incorrect maxPixelValue
DICOM incorrect maxPixelValue
For DICOM images using photometric interpretation MONOCHROME1,
bioformats needs to invert the pixel the data. For 16 bit data we had
been using the WINDOW_WIDTH property as a max pixel value, instead this
actually provides us with the pixel range and is not clipped to 0.

Code has been updated to try and read the WINDOW_CENTER value (the
central value in the range) in order to work out an appropriate scale.
If this is not available or not a sensible value then we will use the
full max pixel range.

OME-Users message:
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-September/00
5629.html

QA: https://www.openmicroscopy.org/qa2/qa/feedback/11401/

For info on the WINDOW_WIDTH and WINDOW_CENTER properties:
https://www.dabsoft.ch/dicom/3/C.11.2.1.2/

For info on the photometric interpretations:
https://www.medicalconnections.co.uk/kb/Photometric_Interpretations

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -128,8 +128,9 @@ public class DicomReader extends FormatReader {
   private byte[][] lut;
   private short[][] shortLut;
   private long[] offsets;
-  private int maxPixelValue;
-
+  private int maxPixelRange;
+  private int centerPixelValue;
+  
   private double rescaleSlope = 1.0, rescaleIntercept = 0.0;
 
   private boolean isJP2K = false;
@@ -406,7 +407,8 @@ public class DicomReader extends FormatReader {
         }
       }
       else if (bpp == 2) {
-        if (maxPixelValue == -1) maxPixelValue = 65535;
+        int maxPixelValue = maxPixelRange + (centerPixelValue/2);
+        if (maxPixelRange == -1 || centerPixelValue == -1 || centerPixelValue < (maxPixelRange/2)) maxPixelValue = 65535;
         boolean little = isLittleEndian();
         for (int i=0; i<buf.length; i+=2) {
           short s = DataTools.bytesToShort(buf, i, 2, little);
@@ -432,7 +434,8 @@ public class DicomReader extends FormatReader {
       lut = null;
       offsets = null;
       shortLut = null;
-      maxPixelValue = 0;
+      maxPixelRange = 0;
+      centerPixelValue = 0;
       rescaleSlope = 1.0;
       rescaleIntercept = 0.0;
       pixelSizeX = pixelSizeY = null;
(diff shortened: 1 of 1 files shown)


sbesson:
@dgault: note this change is modifying the hash of the first plane for the Dicom samples under `test_images_good` - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-test_images_good/642/. Is that expected and would involve a configuration PR or is it a regression?

components/formats-bsd/src/loci/formats/in/DicomReader.java
```diff
@@ -563,6 +566,18 @@ else if (s.indexOf("1.2.4") > -1 || s.indexOf("1.2.5") > -1) {
         case SLICE_SPACING:
         case RESCALE_INTERCEPT:
         case WINDOW_CENTER:
+            String winCenter = in.readString(elementLength);
+            if (winCenter.trim().length() == 0) centerPixelValue = -1;
```
mtbc: `StringUtils.isBlank` can be useful

components/formats-bsd/src/loci/formats/in/DicomReader.java
```diff
@@ -563,6 +566,18 @@ else if (s.indexOf("1.2.4") > -1 || s.indexOf("1.2.5") > -1) {
...
+            else {
+              try {
+                centerPixelValue = new Double(winCenter.trim()).intValue();
```
mtbc: You needn't trim it, the constructor does that for you.

components/formats-bsd/src/loci/formats/in/DicomReader.java
```diff
@@ -406,7 +407,8 @@ else if (isDeflate) {
         }
       }
       else if (bpp == 2) {
-        if (maxPixelValue == -1) maxPixelValue = 65535;
+        int maxPixelValue = maxPixelRange + (centerPixelValue/2);
+        if (maxPixelRange == -1 || centerPixelValue == -1 || centerPixelValue < (maxPixelRange/2)) maxPixelValue = 65535;
```
mtbc: Generally try to keep BF code lines within 80 chars.

dgault:
Relevant configuration PR can be found here: https://github.com/openmicroscopy/data_repo_config/pull/21

jburel:
It might be better to use the method ```defaultMinMax``` from FormatTools to determine the maximum value corresponding to the pixels type instead of hard-coded value.

jburel:
3 images in the images folder are correct. Checked with OsiriX
Note that it is not possible to open the "directory" image, this is possible in OsiriX
Is it something we want to look into?

dgault:
The directory image is something that isn't currently supported but is certainly something that we could look into adding in future

jburel:
@dgault: comment was not for this PR.
I will create a trello card

jburel:
Card created see https://trello.com/c/ciKINMvW/310-dicom-support-for-directory-image

PR works as expected, ready to go.

sbesson:
Unfortunately, it seems the last commit introduced new failures - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-repository-subset/59/console. Do the configuration files need to be updated again?

dgault:
Retested and the new value is expected.
I've updated the config accordingly:
https://github.com/openmicroscopy/data_repo_config/pull/21

melissalinkert:
Double-checked all changed files with latest commits - all looks fine.  Merging for 5.1.5.

PR #1991, opened 2015-09-25
cpp: bioformats: Add n-dimensional array lookup calculations

rleigh-dundee:
This adds a DimensionSet class containing a list of Dimension objects.  It may be used to calculate offsets in arbitrary n-dimentional spaces:

- calculate logical offsets (logical index ↔ logical coordinate)
- calculate storage offsets (logical coordinate ↔ storage index)
- separate logical dimension order from storage order and direction
- optionally work on a relative window into an n-dimensional space (subrange)

This implements the basic logic used in Boost.MultiArray.  However, the number of dimensions is not fixed at compile-time, so this can handle calculations with an arbitrary number of dimensions.  The intent is that this will allow the replacement of the PixelBuffer class (which uses Boost.MultiArray) with a custom n-dimensional array type using this class as a basis for the index calculations.  This will make the PixelBuffer class handle arbitrary dimensions, and will also simplify the interface, making it more usable (it currently exposes the guts of the Boost.MultiArray implementation).  This also allows for this to be re-implemented in Java fairly trivially, so Java could potentially use an nDim pixel buffer in place of raw byte arrays at some future point, making the two implementations the same in this respect; the Boost.MultiArray logic is C++-specific and will not port to Java due to the heavy template usage.

In addition to PixelBuffer, this may also be used to replace the hardcoded getIndex/getCZTCoords methods and functions; this can be used directly in CoreMetadata to supplement/replace sizes and (eventually) modulo since it has an extent list of all dimensions.  This can be used to implement getIndex/getCZTCoords or it could be given directly to the user for their use, obsoleting these methods entirely.  It can also be used in any reader/writer needing to do format-specific calculations relating to dimensions.

Everything is documented, and the internal logic is extensively commented.

CI testing: Unit tests are provided for most of the functionality; the jobs should remain green

Review:
- Run `./bf-test ./cpp/test/ome-bioformats/dimension` and pipe to less or a file for review.
- For each test you'll see something like this:

```
DimensionSet with storage order:
Logical dimensions:      (X,Y)
Logical extents:         (3,4)
Logical ranges:          ([0,3),[0,4))
Logical sizes:           (3,4)
Logical strides:         (1,3)
Size:                    2
Elements:                12
Storage dimensions:      (X,Y)
Storage extents:         (+3,+4)
Base:                    0
Strides:                 (1,3)
Descending offsets:      (0,0)

DimensionSet storage layout:

    X
    ┏━━┯━━┯━━┓
    ┃ 0│ 1│ 2┃
Y ┏━╋━━┿━━┿━━┫
  ┃0┃ 0│ 1│ 2┃
  ┠─╂──┼──┼──┨
  ┃1┃ 3│ 4│ 5┃
  ┠─╂──┼──┼──┨
  ┃2┃ 6│ 7│ 8┃
  ┠─╂──┼──┼──┨
  ┃3┃ 9│10│11┃
  ┗━┻━━┷━━┷━━┛

0,0 →  0 → 0,0
1,0 →  1 → 1,0
2,0 →  2 → 2,0
0,1 →  3 → 0,1
1,1 →  4 → 1,1
2,1 →  5 → 2,1
0,2 →  6 → 0,2
1,2 →  7 → 1,2
2,2 →  8 → 2,2
0,3 →  9 → 0,3
1,3 → 10 → 1,3
2,3 → 11 → 2,3
```

- The first section is a description of the `DimensionSet`.  The first part is the user-specified details such as dimension names and sizes, and subranges.  The rest are the computed strides and offsets to handle the mapping from logical to storage coordinates.
- The second section is is a visual representation of the mapping.  The row and column numbers are the logical coordinates; when there are more than two dimensions, higher dimensions are shown as multiple planes.  The grid values are the storage indices corresponding to these logical coordinates.
- The last section is a linear list of the mappings, including round-trip back to verify the logic works both ways; this is checked by the unit tests so doesn't require manual validation.

What to verify:
- The logic in the `logical_index` and `storage_index` methods; also the `subrange` methods and stride/offset methods.
- Could any of the performance-critical methods be improved; this is essentially `logical_index` and `storage_index`; the others such as subranges and stride/offset calculations only happen once so are not as critical.  The critical ones and their dependent bits should all be inlined.
- That the storage indices are valid.  While this is easy for the simple 1D case (it's a 1:1 mapping) we also:
  - use multiple dimensions
  - vary dimension direction which will use negative strides and nonzero offsets
  - use subranges to provide a window into the nD array; these can be verified by adding the begin value from the range to the logical index, and looking it up in the original layout (the tests display the original and subrange layouts for this purpose)--the number of combinations here is very large; verifying a small selection should be sufficient to verify the correctness of the algorithm

### DIFF
--- a/cpp/lib/ome/bioformats/CMakeLists.txt
+++ b/cpp/lib/ome/bioformats/CMakeLists.txt
@@ -41,6 +41,7 @@ include_directories(${OME_TOPLEVEL_INCLUDES}
 
 set(OME_BIOFORMATS_SOURCES
     CoreMetadata.cpp
+    Dimension.cpp
     FormatException.cpp
     FormatTools.cpp
     MetadataConfigurable.cpp
@@ -60,6 +61,7 @@ set(OME_BIOFORMATS_SOURCES
 
 set(OME_BIOFORMATS_HEADERS
     CoreMetadata.h
+    Dimension.h
     FileInfo.h
     FormatException.h
     MetadataMap.h
--- a/cpp/lib/ome/bioformats/Dimension.cpp
+++ b/cpp/lib/ome/bioformats/Dimension.cpp
@@ -1,0 +1,346 @@
+/*
+ * #%L
+ * OME-BIOFORMATS C++ library for image IO.
+ * Copyright © 2015 Open Microscopy Environment:
+ *   - Massachusetts Institute of Technology
+ *   - National Institutes of Health
+ *   - University of Dundee
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The views and conclusions contained in the software and documentation are
+ * those of the authors and should not be interpreted as representing official
+ * policies, either expressed or implied, of any organization.
+ * #L%
+ */
+
+#include <set>
+
+#include <ome/bioformats/Dimension.h>
+
+namespace ome
+{
+  namespace bioformats
+  {
+
+    void
+    Dimension::check(size_type previous_begin,
+                     size_type previous_end)
+    {
+      if (extent == 0)
+        {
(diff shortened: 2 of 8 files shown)


joshmoore:
Sorry for not getting around to this yesterday, @rleigh-dundee. Obviously, it's quite a bit to think about. A few meta concerns:

 * It would be good to have much of the content of the PR in either the commit message or in the docs folder.
 * We don't have any formal process of API/design proposals, especially not across Java & C++, but it certainly feels like that's what we need for this.

Do you have any PRs planned which are dependent on this?

rleigh-dundee:
On the docs side, this should already be in the inline doxygen markup and hence in the API reference documentation.  That should be where the developer-visible documentation is, other than the higher-level tutorials/examples stuff?  I mean, the commit message and sphinx docs are all fine too, but neither are the primary place for API documentation details.  Certainly happy to extend any or all of these as needed though.  Are there any places in the inline docs, particularly the main class description, where it's not clear or where additional explanation/description is needed?

WRT API/design review, that's certainly something needed here.  This is really the initial groundwork for building other things upon (in the 5.2+ timeframe).  It's something I've been working at on the side for a few months more to satisfy our longer-term nDim needs than any current short-term goal; it's not urgently needed for 5.1.5.  In terms of dependent PRs these would potentially include:

- Java reimplementation of this PR once it's reviewed and finalised
- Replacement of getCZTCoords/getIndex in the C++ and Java code (either direct replacement or added alongside with deprecation of the old methods; the old methods could internally use this class)
- Replacement of the 9D C++ PixelBuffer currently using Boost.MultiArray with a new nD implementation containing this class for the indexing
- Replacement of the C++ TIFF pixel read/write tile/strip copying logic using the new PixelBuffer and/or this class (in IFD.cpp)
- On the Java side there are many readers which implement this sort of logic many times over for all sorts of purposes from subchannel/plane calculations and more; they could potentially be simplified and made potentially more robust by using a common class to do these calculations.
- Addition of an nDim PixelBuffer in Java to match the C++ one; part of the reason for changing the C++ class above is to allow the two languages to share a common PixelBuffer API
- Addition of nDim PixelBuffer openBytes/saveBytes methods to C++ and Java to replace the existing plane-based methods

joshmoore:
Based on the above and discussions, marking as `breaking` since this isn't targeted for 5.1.5 and needs further design discussions.

sbesson:
Note this PR is currently breaking https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-CPP-5.1-breaking/label=gretzky35/

rleigh-dundee:
The broken build on gretzky35 should be fixed now.

mtbc:
gretzky35 now looks fine, https://ci.openmicroscopy.org/view/job/BIOFORMATS-CPP-5.1-breaking/ in general will probably be finished later this afternoon.

rleigh-dundee:
Added design document as requested including rationale, implementation details and algorithms, and potential applications.

mtbc:
It will be interesting to see all this work feed back into Java Bio-Formats someday!

rleigh-dundee:
Thanks.  I've fixed these and some additional typos.  http://trac.openmicroscopy.org/ome/ticket/12081 was the original driver for this on the C++ side, i.e. simplification and usability improvement for the pixel buffer class, also noted by end users.

jburel:
Discussed today during B-F today:
round table to discuss design to be scheduled post BF 5.1.6 
This PR does not block @rleigh-dundee .

mtbc:
Hoping to schedule design discussion before break for holidays?

jburel:
@mtbc: very unlikely, deployment IDR amongst others things to sort out before holidays

sbesson:
Discussed with @rleigh-dundee @mtbc @joshmoore, excluding for now to prevent collisions with the work on #2160.

sbesson:
Closing as the C++ reference implementation is getting decoupled as of https://github.com/openmicroscopy/bioformats/pull/2258.

PR #1992, opened 2015-09-27
Becker hickl spc

imunro:
Add a new reader for Becker & Hickl .spc FIFO files. See https://trac.openmicroscopy.org/ome/ticket/13000.
This format is documented in the TCSPC Handbook available from http://www.becker-hickl.com/handbookphp.htm see Page 675 FIFO Data Files (SPC-134, SPC-144, SPC-154, SPC-830).
Note that this code assumes the presence of pixel clock events and therefore the MacroTime information is not used.


sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/1028/findbugsResult/. Can you fix the findbugs warnings?

imunro:
Done.

sbesson:
Moving to `breaking` like the other new reader.

mtbc:
Doing okay outside breaking? Maybe wait until 5.1.7 release to review?

imunro:
Fine by me.

sbesson:
Discussed with @imunro, the last state of this PR is passing the minimal `showinf` test but fails while wrapping the `ImageReader` with a `Memoizer` and calling `reopenFile` (see https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/42/console).

imunro:
Please note that I can reproduce the problem locally by running 

BF_MAX_MEM=2g ./tools/showinf -cache  file.set     twice. 
However, there is no problem if I do the same with the .spc file.
I suspect therefore that this arises because of the changes I've added to allow opening via the .set companion file.

imunro:
A little further debugging and the issue appears to be that when OpenBytes is called after caching   "in" is the .set companion file not the .spc file containing the data (presumably the caching skips InitFile) @melissalinkert Whats' the best way to address this.

imunro:
Anyone ^^ ? @sbesson

melissalinkert:
Ah, OK, I didn't realize you were testing with caching.  Your options then are:

1. make sure ```currentId``` is set to the .spc file in initFile (so reopenFile works by default)
2. override reopenFile to explicitly find .spc file and set it to ```in```.  another maybe more relevant example is https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/ICSReader.java#L744

imunro:
Thanks Melissa. Option !) looks a lot simpler to implement is there any advantage to option 2) ?

imunro:
Failed competely to implement option 1) so went with option 2) in the end

components/formats-gpl/src/loci/formats/in/SPCReader.java
```diff
@@ -0,0 +1,792 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+
+ * %%
+ * Copyright (C) 2005 - 2015 Open Microscopy Environment:
```
sbesson: Minor: this should be only Copyright (C) 2015

melissalinkert:
Builds are green, and I don't see anything obviously worrying here.  Merging.

PR #1993, opened 2015-09-29
Fixed SlideBook6Reader montage tile stage position reading

RichardMyers:
This change allows the Grid/Collection stitching plug-in to properly read montage tile stage position information from SlideBook6 SLD files.


melissalinkert:
Were changes needed in the Java reader here, or is it just an update of the .dll files?

RichardMyers:
The bug was in the dll, the java interface didn't change.  It only affected xy position of montage tiles so I didn't notice before.

melissalinkert:
Closing in favor of gh-2217.

PR #1994, opened 2015-09-29
DICOM: use the last row and column tags instead of the first

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13023.

To test, use the file from QA 11381.  Without this change, the image size was 1280x2048 and the images should have diagonal stripes.  With this change, the image size should be larger in both dimensions, and there should be an obvious artifact (which can be compared against almost any other DICOM viewer).

This may conflict with gh-1989; I'll exclude this one if that turns out to be the case.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -550,12 +550,16 @@ public class DicomReader extends FormatReader {
           break;
         case ROWS:
           int y = in.readShort();
-          m.sizeY = y;
+          if (y > getSizeY()) {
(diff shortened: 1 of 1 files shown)


sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/55/console. This does not conflict with gh-1989 but leads to a couple of failures when testing the sample dicom files.

melissalinkert:
Fix pushed and re-run: https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-repository-subset/56/

bramalingam:

PR #1998, opened 2015-09-30
ROIs v2

mtbc:
* introduce 2016-DEV0 schema
* remove Shape.Visible from model
* allow validation of unpublished schemas


mtbc:
Went deeply weird, abandoning.

PR #1999, opened 2015-09-30
ROIs v2

mtbc:
* introduce 2016-DEV0 schema
* remove Shape.Visible from model
* allow validation of unpublished schemas

### DIFF
--- a/ant/xsd-fu.xml
+++ b/ant/xsd-fu.xml
@@ -7,7 +7,7 @@ Include this in components generating sources using xsd-fu.
 
 <project>
 
-  <property name="xsdfu.schemaver" value="2015-01"/>
+  <property name="xsdfu.schemaver" value="2016-DEV0"/>
   <property name="xsdfu.schemapath" value="${root.dir}/components/specification/released-schema/${xsdfu.schemaver}"/>
   <property name="xsdfu.templatepath" value="${root.dir}/components/xsd-fu"/>
   <property name="xsdfu.schema.ome" value="${xsdfu.schemapath}/ome.xsd"/>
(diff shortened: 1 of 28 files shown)


rleigh-dundee:
The model changes all look fine; I've reviewed the diff against the 2015-01 model to check the changes.

`components/specification/transforms/ome-transforms.xml` also looks fine, though it's fairly horrible; could we consider replacing this with one line per transform and then create a graph of the transforms?  (As a future change, not in this PR.) This would describe the same transformations but in a tiny fraction of the size and would scale better when adding new model versions.

The code changes also look fine including the version change and the addition of the schema resolving from the jar.

I'll do some testing this afternoon to double-check, but the changes themselves all look good.

mtbc:
The double-checking went okay? Thanks for the scrutiny. :smiley:

rleigh-dundee:
I did some basic testing to make sure schema validation was working; no problems found, so looks good to merge from me.

mtbc:
@rleigh-dundee: I too was horrified by `ome-transforms.xml` and its scalability. This PR reordered some entries to make the whole layout more consistent from older to newer, but it sounds like you have some further ideas. In the end, partly because I should probably learn more Python anyway, I wrote a script to generate new versions of `ome-transforms.xml` in the hope of saving others from having to adjust it manually. Should I perhaps push that somewhere? Though, I don't know where it would go, somewhere in the specifications module? Then if you wanted, I could open a PR for it and you could improve it later, or you could just cherry-pick my commit into one of your PRs and then rename and edit the script yourself however you please.

rleigh-dundee:
A standalone script in `specification` would be appropriate I think; the other alternative would be either a standalone script in `xsd-fu` or a target in the `xsd-fu` program itself.  `specification` might be better since that's where the transforms are, but there's no build logic in this component, so `xsd-fu` might be a better fit in practice?  Either would be fine as a PR I think.

mtbc:
Thank you; opened #2024. Do feel free to majorly move/rework it later as you see fit.

PR #2000, opened 2015-09-30
source-archive: Make tar source release

rleigh-dundee:
This script creates both a zip and tar source release.  The zip file is unfortunately incompatible with some extractors, making it break the superbuild on a number of platforms using libarchive to extract.  This includes Ubuntu and BSD systems using the current release of libarchive when using "cmake -E tar xfv" to extract the zips (done in the superbuild).  Since fixing these platforms won't happen in the short-term, making an archive they can extract is a simpler solution.  The content is exactly the same as the zip.

Using a current CMake downloaded from cmake.org for e.g. Windows or Linux works fine since they use a newer (unreleased) copy of libarchive git, where the bug is fixed.  Unfortunately for Ubuntu and BSD systems (and likely others) which provide cmake linked against their stable libarchive, they use a version of libarchive with the bug and this breaks the superbuild.  Ideally, I'd like to switch to using the tar source release for the 5.1.5 superbuild.

--------

Testing:

```
# Make release
ant release
cd artifacts

# Unpack all archives
unzip bioformats-5.1.4-DEV.zip
unzip bioformats-dfsg-5.1.4-DEV.zip
mv bioformats-5.1.4-DEV bioformats-5.1.4-DEV-zip
mv bioformats-dfsg-5.1.4-DEV bioformats-dfsg-5.1.4-DEV-zip
tar xfv bioformats-5.1.4-DEV.tar
tar xfv bioformats-dfsg-5.1.4-DEV.tar

# Compare
diff -urN bioformats-5.1.4-DEV-zip bioformats-5.1.4-DEV
diff -urN bioformats-dfsg-5.1.4-DEV-zip bioformats-dfsg-5.1.4-DEV

# Check that files which are executable in git have execute permissions

# Check appended generated content
# This will match the content in the -zip directories, checked above
cat bioformats-5.1.4-DEV/ant/gitversion.xml
cat bioformats-5.1.4-DEV/cpp/cmake/GitVersion.cmake
cat bioformats-dfsg-5.1.4-DEV/ant/gitversion.xml
cat bioformats-dfsg-5.1.4-DEV/cpp/cmake/GitVersion.cmake
```

### DIFF
--- a/tools/source-archive.py
+++ b/tools/source-archive.py
@@ -7,6 +7,8 @@ import os
 from subprocess import call
 import sys
 import zipfile
+import tarfile
+import StringIO
 
 # This script archives the base tree and repacks it into a single zip which is
 # the source release, taking care to preserve timestamps and exectute
@@ -86,7 +88,7 @@ if __name__ == "__main__":
         '--output', "%s/%s-base.zip" % (options.target, prefix),
         'HEAD'])
     if base_archive_status != 0:
-        raise Exception('Failed to create git base archive')
+        raise Exception('Failed to create git zip base archive')
 
     zips = list(["%s/%s-base.zip" % (options.target, prefix)])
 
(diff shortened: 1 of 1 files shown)


mtbc:
Good to merge.

sbesson:
Can we create a follow-up card from this to list all the CI tasks to would need to be carried out (like archiving the tar artifacts alongside the other ones, adding it to the release jobs, downloads page?). I suppose creating a tar.gz would not solve the issue that this PR fixes?

rleigh-dundee:
We need to:

- Compress and archive the tar files
- Add to release job if needed
- Add to download page (will need source download button with "tar" in place of "zip")

WRT "creating a tar.gz", I'm not sure exactly what you're referring to since this is exactly what this PR is doing.

melissalinkert:
Merging for 5.1.5.

PR #2002, opened 2015-10-01
BMP: fix cropping of compressed images

melissalinkert:
This should fix the error in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/628/console.


sbesson:
Both full repository jobs turned green this week-end. Merging.

PR #1996, opened 2015-09-30
Add documentation page on the autogenerated documentation

sbesson:
See https://trello.com/c/bGwmDei8/8-document-how-to-add-update-format-pages

This PR adds a first version of a developer documentation page explaining how the formats and readers pages are generated and can be modified.

/cc @hflynn

PR #1988, opened 2015-09-22
Reader metadata pages refactoring 

sbesson:
See https://trello.com/c/kaw2jhNL/20-metadata-pages

This PR:
- renames all reader metadata pages as `metadata/XXXReader.txt` instead of reusing the formats name as previously
- uses the `reader` list in `formats-pages.txt` to link both to the source code on Github and to the reader metadata pages
- removes the deprecated `metadataPage` field in `formats-pages.txt`
- the old metadata pages `formats/xxxx-metadata.txt` are preserved to prevent breaking URLs and a top-level deprecation warning is added with a link to the formats page
- removes the duplicate Slidebook format page in favor of linking to both metadata pages for the `SlidebookReader` and `Slidebook6Reader`
- adds the missing metadata page for the base readers e.g. `BaseTiffReader`

To review this PR, the main entry points are:
- the [main metadata page](https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/metadata-summary.html): all links in the reader list should point to the correct reader metadata page named `metadata/XXXReader`
- the [supported formats page](https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/supported-formats.html). Formats should now link to the source code and the metadata page. Check formats with multiple readers, like Bitplane Imaris and Slidebook, are behaving as expected.
- the old metadata pages e.g. https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/formats/aperio-afi-metadata.html should now have a top-level deprecation warning redirecting to the formats page.

PR #2004, opened 2015-10-01
Generated format pages: cleanup

sbesson:
Follow-up of gh-1988 and gh-1996, this PR does some additional cleanup for the format pages auto-generation

- removes the `export` flag in `src/format-pages.txt`
- use the `writer` field instead to determine whether Bio-Formats can write the data
- moves the `Writer: ` block under the `Reader: ` for clarity in individual format pages
- removes the  `Notes:` heading and make the `Additional information` header conditional to the existence of notes
- registers the `CellH5Writer` in `src/format-pages.txt`

All format pages have been regenerated. Apart from CellH5 which should mention export, there should be no notable difference apart from a general tidying up.

### DIFF
--- a/components/autogen/src/doc/FormatPage.vm
+++ b/components/autogen/src/doc/FormatPage.vm
@@ -42,6 +42,9 @@ Readers:
 Reader: ${item} (:$readerextlink:`Source Code <${item}.java>`, :doc:`Supported Metadata Fields </metadata/${item}>`)
 #end
 
+#if ($writer)
+Writer: ${writer} (:$writerextlink:`Source Code <${writer}.java>`)
+#end
 
 #if ($software)
 Freely Available Software:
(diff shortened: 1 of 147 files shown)


sbesson:
See pages under https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/supported-formats.html

hflynn:
Staging pages for formats all look good. Only minor quibble is whether you also want to update the text for 'notes' on https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/format-documentation.html ?

sbesson:
Is there anything to update on this? the behavior of the notes field should be unchanged - only its display has one less header.

hflynn:
Ah ok, I wasn't sure. Good to merge in that case.

PR #2005, opened 2015-10-01
cpp: Remove additional warnings

rleigh-dundee:
Ignore these warnings:

-  -Wvariadic-macros
-  -Wunused-local-typedef
-  -Wlanguage-extension-token
-  -Wold-style-cast

These had an exeedingly large number of false positives, particularly in external library headers, and these masked real warnings by the sheer volume.

With these warnings ignored, a number of real warnings were exposed which are fixed here.  This should significantly reduce the size of the build log.

--------

Testing: no special testing required; check CPP jobs are green.

### DIFF
--- a/components/xsd-fu/templates-java/OMEXMLModelObject.template
+++ b/components/xsd-fu/templates-java/OMEXMLModelObject.template
@@ -129,7 +129,6 @@ public class ${klass.name} extends ${klass.parentName}
 	public ${klass.name}(Element element, OMEModel model)
 	    throws EnumerationException
 	{
-                // comment
 		update(element, model);
 	}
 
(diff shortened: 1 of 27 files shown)


components/xsd-fu/templates-java/OMEXMLModelObject.template
```diff
@@ -129,6 +129,7 @@ public class ${klass.name} extends ${klass.parentName}
 	public ${klass.name}(Element element, OMEModel model)
 	    throws EnumerationException
 	{
+                // comment
```
sbesson: Was this addition meant to be included in this PR - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-generated-sources/lastSuccessfulBuild/artifact/ome-xml-model-java.diff ?

rleigh-dundee: No, that's an oversight which I'll remove.

rleigh-dundee:
@sbesson I've dropped that comment now.

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-5.1-merge/ is green and the build logs are consistently 11M per matrix component with this PR included.

PR #2010, opened 2015-10-06
docs: Document C++ super-build library dependency caveat

rleigh-dundee:
Noted during release testing, documenting as requested.


mtbc:
Good to merge.

PR #2006, opened 2015-10-01
Temporarily disable DeltavisionReader for IDR

joshmoore:
In order to import tara.

PR #1969, opened 2015-09-10
ScreenReader: the next generation

melissalinkert:
Rewrite of ScreenReader to only accept a *.screen file, which describes how the plate/well/field-to-file mapping works.  ScreenReader will no longer attempt to guess the plate layout based upon the data file names, and so is now always included.

Examples of what a ```.screen``` file should look like are in /ome/team/mlinkert/idr_screen_files; happy to make other more generally useful examples and docs once we know that those actually work as desired.

I have not yet tested this in OMERO, so there may yet be one or two small issues (possibly related to memoization).

See https://trello.com/c/vpc8NKAC/286-metamorph-stk-per-channel, https://trello.com/c/3D964ahX/299-plate-grouping-of-zeiss-axiovision-zvi-files.

PR #2012, opened 2015-10-07
screen-reader-2.0 perf

joshmoore:
This follows on from gh-1969 and replaces gh-2006. The intent is to speed up the performance of the new `ScreenReader` class for cases where there are many (4K+) files each of which takes a significant amount of time to open.

Especially the introduction of the `Memoizer` *inside* of `ScreenReader` may not be a general solution, but until we can inject some form of context, this may be the only type of solution possible.

NB: one further idea would be to write `Reader` subclasses and inject them via `ClassList` but that has its own issues.


joshmoore:
Wrong branch.

PR #2011, opened 2015-10-06
Fix activation of test-jar profile

ctrueden:
The conditional for deciding when to build the tests JAR was wrong, because it was keyed to `src/test/java/` instead of `test/`.

Consequently, no `*-tests.jar` files were being produced.

To test, perform a clean Maven build, and then verify that the `*-tests` artifacts are now present. I.e.:

```bash
$ mvn clean
...
$ mvn
...

$ for t in components/*/test/; do echo $t; done
components/bio-formats-plugins/test/
components/formats-api/test/
components/formats-bsd/test/
components/formats-common/test/
components/formats-gpl/test/
components/metakit/test/
components/ome-jxr/test/
components/ome-xml/test/

$ ls -l **/target/*-tests.jar
-rw-r--r--  1 curtis  staff   32732 Oct  6 10:45 components/bio-formats-plugins/target/bio-formats_plugins-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff    6621 Oct  6 10:42 components/formats-api/target/formats-api-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff  155070 Oct  6 10:45 components/formats-bsd/target/formats-bsd-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff   83597 Oct  6 10:42 components/formats-common/target/formats-common-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff  135476 Oct  6 10:45 components/formats-gpl/target/formats-gpl-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff   22528 Oct  6 10:45 components/metakit/target/metakit-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff   24056 Oct  6 10:46 components/ome-jxr/target/ome-jxr-5.1.5-SNAPSHOT-tests.jar
-rw-r--r--  1 curtis  staff   16959 Oct  6 10:42 components/ome-xml/target/ome-xml-5.1.5-SNAPSHOT-tests.jar
```

PR #2014, opened 2015-10-07
Remove activation profile for building test JARs

sbesson:
Follow-up of https://github.com/openmicroscopy/bioformats/pull/2011.

The PR inadvertently broke https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-latest-maven/507/. The reason is that the activation of the `test-jar` profile deactivated the profile marked as "active by default" (see http://maven.apache.org/guides/introduction/introduction-to-profiles.html) which defines the default deployment configuration. Thus the `mvn clean deploy` command is currently broken whereas `mvn clean deploy -Platest-build` will deploy artifacts as expected.

This commit proposes to remove the requirement for a conditional activation profile in favor of building test JARs unconditionally for every component. Components like xsd-fu with no `test` directory will simply create dummy test JARs (like the `xsd-fu` JAR itself).

/cc @melissalinkert @ctrueden

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -362,6 +362,13 @@
               <exclude>**/*.h</exclude>
             </excludes>
           </configuration>
+          <executions>
+            <execution>
+              <goals>
+                <goal>test-jar</goal>
+              </goals>
+            </execution>
+          </executions>
         </plugin>
 
         <plugin>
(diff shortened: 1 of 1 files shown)


ctrueden:
Personally I never use `activeByDefault`, but instead always activate the profile on some appropriate criteria. Even if it's just something like "such-and-such system property is unset"—which effectively becomes "active by default, unless you explicitly disable it".

I think a strategy like that would be superior, to avoid the creation of the dummy test JARs.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-maven/108/consoleText. With the last commit only non-empty JAR tests have been uploaded to the artifactory.

ctrueden:
Awesome, @sbesson. I forgot about the `skipIfEmpty` flag. We actually simplified the pom-scijava parent to use that as well last year (scijava/pom-scijava@dc15f56a9fc0f8bd8ffc2d6ce19a2f08683d8180), but it had been so long it slipped my mind.

If you would indulge a very short ideological rant: this sort of issue would often be avoided if the Bio-Formats base POM simply extended pom-scijava like [other SciJava projects do](http://imagej.net/Architecture#Maven_component_structure). It seems like an obvious synergy point to me, and I don't understand the technical obstacles.

PR #2017, opened 2015-10-09
octave - drop unecessary String::toString and char()

carandraug:
This has two commits.  The first is an issue reported on the Octave mailing list by Philip Nienhuis (no github account)

> bfGetFileExtensions.m:
> ----------------------
> ...when run in Octave chokes on
>    fileExt{i, 2} = char(readers(i).getFormat().toString)
> 
> error: sq_string cannot be indexed with .
> error: evaluating argument list element number 1
> 
> ....probably because readers(i).getFormat() is already (un)boxed to a char string.
> See attached patch1 for a fix.

According to [Matlab documentation](http://uk.mathworks.com/help/matlab/matlab_external/handling-data-returned-from-a-java-method.html#f61197), `java.lang.String` objects should be automatically converted into char arrays:

> When a method call returns data of type java.lang.Object, MATLAB converts its value, depending on its actual type, according to the following table.

Since a char array is a Matlab type, it does not have a `toString()` method. I am surprised that it works so I may be  misunderstanding Matlab documentation.  However, Octave does it as stated which causes the issue.

Since these String are automatically converted to char array, calls to `char()` are unnecessary. The second commit removes those.

### DIFF
--- a/components/formats-gpl/matlab/bfGetFileExtensions.m
+++ b/components/formats-gpl/matlab/bfGetFileExtensions.m
@@ -48,7 +48,7 @@ for i = 1:numel(readers)
     else
         fileExt{i, 1} = arrayfun(@char, suffixes, 'Unif', false);
     end
-    fileExt{i, 2} = char(readers(i).getFormat().toString);
+    fileExt{i, 2} = char(readers(i).getFormat());
 end
 
 % Concatenate all unique formats


components/formats-gpl/matlab/bfCheckJavaPath.m
```diff
@@ -77,9 +77,9 @@
 if status
     % Read Bio-Formats version
     if is_octave()
-        version = char(java_get('loci.formats.FormatTools', 'VERSION'));
+        version = java_get('loci.formats.FormatTools', 'VERSION');
     else
-        version = char(loci.formats.FormatTools.VERSION);
+        version = loci.formats.FormatTools.VERSION;
```
sbesson: Unfortunately on MATLAB R2014b at least, the conversion does not seem to happen by default

```
>> version = loci.formats.FormatTools.VERSION;
>> class(version)

ans =

java.lang.String
```

which is why the casting was there in the first place.

carandraug: hmmm... that's interesting. I can't test this on Matlab myself but seems to go against their documentation. Maybe what they meant to say is that Matlab does an implicit conversion to char when required and not as soon as the java method returns? (their documentation does state "When a method call returns data of type java.lang.Object, MATLAB converts its value").

Anyway, this commit is not really important, I just thought of removing unnecessary function calls. The other does cause compatibility issues for Octave though.

carandraug:
The second commit does not really change much, it only removes functions call that in my view change nothing (apparently, this may not be true on certain Matlab versions).

However, the first commit is necessary for Octave compatibility and should work in Matlab too. (it drops a `java.lang.String::toString()`). Can that one be merged?

sbesson:
@carandraug: quickly checked the first commit dropping the `.toString()` method and everything seemed to worked fine on MATLAB R2014b. Can you force push away the unnecessary commits/changes from this PR?

carandraug:
> Can you force push away the unnecessary commits/changes from this PR?

Done.

sbesson:
Tests are green. Tested using MATLAB R2014b and `bfopen`. Works as expected.

PR #2018, opened 2015-10-09
matlab and octave: add flexibility to the filename of bioformats jar.

carandraug:
This issue was reported on the Octave mailing list by Philip Nienhuis (no github account). I used his patch to prepare the commit.

> bfCheckJavaPath.m:
> ------------------
> ...rigidly checks for a jar strictly named bioformats_package.jar or loci_tools.jar.
> Several Linux distros rename the jars by appending a version number. On my boxes (Linux and Windows) I usually do the same (to keep track of versions) and found that bioformats_package-5.1.4.jar isn't detected. So I attached a patch for that (patch2) that just checks for any jar with bioformats_package of loci_tools in the name.

Please note that the Octave package does not includes the jar file. It must be downloaded separately (including the jar file in the Octave package would not make much sense since Octave requires the jar file to be on the java static classpath).

To my knowledge, bioformats is not packaged by any Linux distro.  What I have found is that Linux distros do add a version number but then also have an unversioned file which is a symlink to one of the versioned files (this is what I do on my own system).

> An often better approach to this sort of check is to try one of the methods in the jar in question with a try-catch around it (e.g., the version number call). That way the jar names don't matter, just the presence of the jar, and you'll know if it is the right version that is first in the javaclasspath.
> Anyway, some lines down, this function expects bioformats_package.jar (or loci_tools.jar) to live in some predefined location. All well, but if it isn't there the javaaddpath call crashes. So, my advice is to put a try-catch around it with a sensible message.

I also prefer this approach and indeed, that's how I wrote [Octave's PKG_ADD for the bioformats package](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/octave/PKG_ADD#L12).  This script is executed when the bioformats package is loaded by Octave and errors if it can't create a FakeReader (maybe there's a better things to check?).

As such, an alternative fix for this would be to replace `bfCheckJavaPath` with a modified `PKG_ADD`.

### DIFF
--- a/components/formats-gpl/matlab/bfCheckJavaPath.m
+++ b/components/formats-gpl/matlab/bfCheckJavaPath.m
@@ -50,11 +50,12 @@ ip.parse(varargin{:});
 % Check if a Bio-Formats JAR file is in the Java class path
 % Can be in either static or dynamic Java class path
 jPath = javaclasspath('-all');
-bfJarFiles = {'bioformats_package.jar', 'loci_tools.jar'};
+jars = cell2mat (cell2mat (regexp (jPath, '[\\/]{1}([^/\\]+)\.jar$', 'tokens')));
+bfJarFiles = {'bioformats_package', 'loci_tools'};
 hasBFJar =  false(numel(bfJarFiles), 1);
 for i = 1: numel(bfJarFiles);
-    isBFJar =  @(x) ~isempty(regexp(x, ['.*' bfJarFiles{i} '$'], 'once'));
-    hasBFJar(i) = any(cellfun(isBFJar, jPath));
+    isBFJar =  @(x) ~isempty(regexp(x, bfJarFiles{i}, 'once'));
+    hasBFJar(i) = any(cellfun(isBFJar, jars));
 end
 
 % Check conflicting JARs are not loaded


components/formats-gpl/matlab/bfCheckJavaPath.m
```diff
@@ -50,11 +50,12 @@
 % Check if a Bio-Formats JAR file is in the Java class path
 % Can be in either static or dynamic Java class path
 jPath = javaclasspath('-all');
-bfJarFiles = {'bioformats_package.jar', 'loci_tools.jar'};
+jars = cell2mat (cell2mat (regexp (jPath, '[\\/]{1}([^/\\]+)\.jar$', 'tokens')));
```
sbesson: Unfortunately this change fails to run in MATLAB - see error messages in https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-matlab/924/

carandraug: This flattens a cell array of cell arrays. It works in Octave, I assumed that Matlab also supported it:

    octave:1> bfJarFiles = {'bioformats_package.jar', 'loci_tools.jar'};
    octave:2> jPath = javaclasspath('-all');
    octave:3> cell2mat (cell2mat (regexp (jPath, '[\\/]{1}([^/\\]+)\.jar$', 'tokens')))
    ans = 
    {
      [1,1] = ij
      [1,2] = bioformats_package
    }

Would you please check if the alternative syntax is supported in Matlab?

    octave:8> names = regexp (jPath, '[\\/]{1}([^/\\]+)\.jar$', 'tokens');
    octave:9> [[names{:}]{:}]
    ans = 
    {
      [1,1] = ij
      [1,2] = bioformats_package
    }

carandraug:
Whether this patch applies or not, the original bug report suggested an alternative which I agree would be better.

Instead of checking for the presence of a jar file with a specific name (with or without a version suffix), the function should check that specific classes and methods, the ones required for the package to work, are available. This would also massively simplify this function.

What do you think?

sbesson:
@carandraug: as you may have seen from the history of the file, it was originally introduced in 2012 and has been barely modified since. Now is definitely a good time to review and improve its implementation. I cannot see any obvious reason why the approach proposed in the description would not work.

Given the semantics of the function, encapsulating the `loci.formats.FormatTools.VERSION` call in a `try/catch` block and handling `ClassNotFoundException` exception instead of parsing the Java class path certainly makes sense to me. Do you want to have a go at this proposal as part of this PR (or a separate one)? I will not be able to get back to this before next week.

carandraug:
Yes, I will have a go implementing the alternative fix to this issue.

carandraug:
I have added 3 commits that have pretty much rewrote this function.

Would you please check my [rewrite-bfCheckJavaPath](https://github.com/carandraug/bioformats/commits/rewrite-bfCheckJavaPath) branch as an alternative fix to this?

sbesson:
Looking at https://github.com/openmicroscopy/bioformats/compare/develop...carandraug:rewrite-bfCheckJavaPath, this diff certainly seems sensible to me. Feel free to close this PR and open a new one from your branch.

carandraug:
Done. Discussion continues on issue #2027

But I think it should be possible to do `curl --user "sbesson" --request POST --data '{"issue":"2018","head":"carandraug:rewrite-bfCheckJavaPath","base":"develop"}' https://api.github.com/repos/openmicroscopy/bioformats/pulls` to avoid having the discussion over multiple places (it works for turning an issue into a pull request but not sure about replacing a pull request)

sbesson:
Sadly, I am pretty sure the GH API does not allow the modification of the `head/base` of a `PullRequest` once it has been created.

PR #2019, opened 2015-10-10
Add version history for 5.1.5

melissalinkert:
See https://trello.com/c/z0ZIeT8e/22-version-history-prs

I didn't add C++ changes, on the assumption that those will be written up separately due to the difference in release dates.  Happy to add text in this PR though if that's preferable.

### DIFF
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -23,6 +23,8 @@ Version history
          are present
    - Fluoview and Andor TIFF
        - fixed physical Z size calculation
+   - Imspector OBF
+       - updated to parse OME-XML metadata (thanks to Bjoern Thiel)
 * Documentation updates, including:
    - substantial updates to the format pages
        - improved linking of reader/writer classes to each format page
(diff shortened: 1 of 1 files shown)


sbesson:
Overall works for me. We might want to mention the Imspector OBF improvements from https://github.co /openmicroscopy/bioformats/pull/1901.

melissalinkert:
Thanks - missed that one since it was in the middle of the ones merged for 5.1.4.  Updated now.

rleigh-dundee:
@melissalinkert I've opened a PR to add the C++ changes; if this isn't formatted appropriately please feel free to change as needed.

Regarding writing up separately and the separate release dates, we are tagging and releasing both at the same time in the same repo, so I think it's appropriate to put the changes together here.  The separate superbuild repo doesn't alter this--it will need a separate set of changes documenting such as bumping the provided component versions, but won't itself include any code changes.

sbesson:
After second thought, merging this. I will open @rleigh-dundee's branch as a separate PR to allow validation.

PR #2023, opened 2015-10-14
XML reading

jburel:
In this PR, we do not record the light settings if the light source is not present in the file
This can be the case with 3rd party generated ome-tiff see https://trac.openmicroscopy.org/ome/ticket/13038
To test this PR:

test 1
 * Open the image from ​https://www.openmicroscopy.org/qa2/qa/feedback/11402/, using BF importer
 * Check the "Display OME-XML metadata" option
 * Check that there is no ligthsettings under channel and no light source under instrument. W/o this PR, light settings were present under channels

test 2:
 * Follow the same steps but this time opened the sample image https://github.com/openmicroscopy/bioformats/blob/develop/components/specification/samples/2015-01/6x4y1z1t1c8b-swatch-instrument.ome. Check that light settings are present under channel.

The image should now import in OMERO.

### DIFF
--- a/components/formats-api/src/loci/formats/meta/MetadataConverter.java
+++ b/components/formats-api/src/loci/formats/meta/MetadataConverter.java
@@ -705,40 +705,40 @@ public final class MetadataConverter {
     }
     catch (NullPointerException e) { }
     for (int i = 0; i < instrumentCount; i++) {
-        int lightSourceCount = 0;
-        try {
-          lightSourceCount = src.getLightSourceCount(i);
+      int lightSourceCount = 0;
+      try {
+        lightSourceCount = src.getLightSourceCount(i);
+      }
+      catch (NullPointerException e) { }
+      for (int j = 0; j < lightSourceCount; j++) {
+        String type = src.getLightSourceType(i, j);
+        if (type.equals("Arc")) {
+          String id = src.getArcID(i, j);
+          if (id != null && id.trim().length() > 0) {
+            lightSourceIds.add(id);
+          }
+        } else if (type.equals("Filament")) {
+          String id = src.getFilamentID(i, j);
+          if (id != null && id.trim().length() > 0) {
+            lightSourceIds.add(id);
+          }
+        } else if (type.equals("GenericExcitationSource")) {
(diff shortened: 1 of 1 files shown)


components/formats-api/src/loci/formats/meta/MetadataConverter.java
```diff
@@ -698,6 +698,48 @@ private static void convertFileAnnotations(MetadataRetrieve src,
...
+        int lightSourceCount = 0;
+        try {
+          lightSourceCount = src.getLightSourceCount(i);
+        }
+        catch (NullPointerException e) { }
+        for (int j = 0; j < lightSourceCount; j++) {
+            String type = src.getLightSourceType(i, j);
+            if (type.equals("Arc")) {
+                String id = src.getArcID(i, j);
+                if (id != null && id.trim().length() > 0) {
```
mtbc: It's fine as it is, but FYI `StringUtils.isNotBlank` can be useful for this.

jburel: will need to check if we have the dependency in BF

jburel: also need to reformat it

mtbc: ah yes, two-space indent :smiley:

melissalinkert:
All makes sense in principle - would it make more sense to move the ID array building to ```convertLightSources``` though?  That should be called before ```convertImages```, and already has light source type case logic.

bramalingam:
Since we are not consuming the latest BF merge on eel, was not able to test the import to OMERO.
Everything else works as expected. Good to merge.

joshmoore:
@bramalingam : and that tested the new commit (https://github.com/jburel/bioformats/commit/0b51f3c2f45df9ede02cb4ba283195a679b78c7f)?

sbesson:
Note with https://github.com/openmicroscopy/openmicroscopy/pull/4278, this PR will now be included in the merge build of eel.

bramalingam:
checked the import as well, works as expected.! Good to merge.

PR #2024, opened 2015-10-14
add Python utility script for generating ome-transforms.xml

mtbc:
Following discussion at https://github.com/openmicroscopy/bioformats/pull/1999#issuecomment-147967612 this gets a local utility script of mine into version control. `cd components/specification/transforms/` then run `python util/write-transforms.py` to see a suggested replacement for `ome-transforms.xml` in largely its current style, thus avoiding the manual maintenance headache. Followup PRs from @rleigh-dundee or others can then improve the script or even have the transforms file autogenerated, but right now this PR's better than nothing.

--no-rebase

### DIFF
--- a/components/specification/transforms/util/write-transforms.py
+++ b/components/specification/transforms/util/write-transforms.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python
+
+# Copyright (C) 2015 University of Dundee & Open Microscopy Environment.
+# All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+# author: m.t.b.carroll@dundee.ac.uk
+
+from collections import defaultdict
+from os import listdir
+from re import compile
+
+
+# note the schemas (in reverse order) and the transforms among them
+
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
This looks great.  I've opened a PR (https://github.com/mtbc/bioformats/pull/1) to add the XML header etc.  If that could be added in before this is merged, that would be great.

In terms of correctness, all the transforms in the original file are present and unchanged; the only bit which differs is the `info` attribute for 2003-FC which is looks unimportant, and some elements with no content.

Good to merge as-is or with my PR added.

mtbc:
@rleigh-dundee: I cherry-picked the generator part of your PR's commit and pushed it to this branch, thank you for that!

components/specification/transforms/util/write-transforms.py
```diff
@@ -0,0 +1,189 @@
...
+    print '\t\t\t</target>'
+
+
+# print in XML all the transforms among the schemas
+
+print """<?xml version = "1.0" encoding = "UTF-8"?>
+<!--
+	#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+	#
+	# Copyright (C) 2012-2014 Open Microscopy Environment
```
sbesson: Should this range be extended to cover `2015` ?

mtbc: I figured that we're not really adding anything new to the transforms file outside the regions space so on this branch there's probably not much to protect. @rleigh-dundee: Once this PR is merged I could open a PR merging BF into regions and updating this to 2015, and once OMERO 5.2.0 is tagged I could open a PR merging that into regions, if you like? Basically, I'm happy to help, but also happy to keep clear of conflicts with your work.

rleigh-dundee: All sounds fine to me.

sbesson:
Looks good to merge. Since we are still maintaining `ome-transforms.xml` under version control for now, should we regenerate the latest version  of this file using the latest state of this PR and have a baseline for future modifications? Or is your plan to get rid of it in the mid-term?

mtbc:
Unfortunately if we do that it'll probably conflict horribly when we merge the regions branch back in.

Simply having the file generated as part of the build is an interesting idea!

PR #2027, opened 2015-10-15
matlab and octave: add flexibility to the filename of bioformats jar

carandraug:
See all discussion on issue #2018

### DIFF
--- a/components/formats-gpl/matlab/bfCheckJavaPath.m
+++ b/components/formats-gpl/matlab/bfCheckJavaPath.m
@@ -51,16 +51,14 @@ ip.parse(varargin{:});
 [status, version] = has_working_bioformats();
 
 if ~status && ip.Results.autoloadBioFormats,
-    jarPaths = getJarPaths();
-    % Try each of the found jar files until we find something that works
-    for i = 1:numel(jarPaths)
-      javaaddpath(jarPaths{i});
-      [status, version] = has_working_bioformats();
-      if (status)
-          break
-      else
-          javarmpath(jarPaths{i});
-      end
+    jarPath = fullfile(fileparts(mfilename('fullpath')), ...
+                       'bioformats_package.jar');
+    if (exist(jarPath) == 2)
+        javaaddpath(jarPath);
+        [status, version] = has_working_bioformats();
+        if (~status)
+            javarmpath(jarPath);
+        end
     end
 end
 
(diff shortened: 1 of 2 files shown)


sbesson:
From the tests standpoint, https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-matlab/927/ is green. Tested manually MATLAB R2014b: 
- the MATLAB zip bundle loads the included Bio-Formats JAR as expected and returns the correct version
- subsequent calls to `bfCheckJavaPath` are returning quickly as the Java path is not returned and no regexp is performed :+1:

The functionality to include `bioformats_package.jar` and `loci_tools.jar` from the MATLAB path is now lost in favor of a hard-coded list of paths including the function directory. This functionality was used primarily for development purposes. For instance the following workflow is now invalid:
- build Bio-Formats using `ant` or `Maven`
- add `artifacts` or `components/bundles/bioformats_package/target` to the MATLAB path
- in MATLAB, go to `component/formats-gpl/matlab` and run `bfCheckJavaPath`

A couple of breaking changes are also included as part of this PR:
- conflicting JAR files in the classpath are not detected anymore
- with the `autoloadBioFormats` option turned on, `bfCheckJavaPath()` does not throw an exception anymore when no suitable JAR file is found but the return value of the function should be checked.

Are the breaking changes listed above be here be design? If so, we could review and include them for Bio-Formats 5.2.0 while backwards compatible changes could be included in Bio-Formats 5.1.x.

carandraug:
> The functionality to include bioformats_package.jar and loci_tools.jar from the MATLAB path is now lost in favor of a hard-coded list of paths including the function directory.

I would argue that they are not completely hard coded.  I'm following the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) which allows for setting of environment variables.  The only hardcoded thing is that the jar files will be inside a `java` directory which is a widespread practice (unfortunately, I am not aware of any actual specs about default place for java classes).

I have checked that in Debian, Ubuntu, Fedora, CentOS, and ArchLinux.  They all follow this (note that Debian and Fedora are the basis for the vast majority of Linux distributions and its a good guess their parents follow the same organization).  In FreeBSD it will fail since jar files are inside `java/classes`.  For Apple and Microsoft systems I did nothing as I'm not familiar with those systems and do not know where system wide installations of java shared libraries are typically placed there (I did left an empty place for them).

However, this is all about trying to guess where jar files are (either installed by the system or the user).  There's only so much guessing one can do.  If the user placed them in an atypical place, then they can always set the java classpath manually.

> This functionality was used primarily for development purposes. For instance the following workflow is now invalid:
>
> * build Bio-Formats using ant or Maven
> * add artifacts or components/bundles/bioformats_package/target to the MATLAB path
> * in MATLAB, go to component/formats-gpl/matlab and run bfCheckJavaPath

But is the Matlab path the best place to add this?  That path is the place where Matlab should look for Matlab functions, or scripts.  Not libraries meant to be shared between applications.  If one is adding `artifacts` only so that `bioformats_package.jar` can be found there, why not add `artifacts/bioformats_package.jar` to the java classpath instead?

> A couple of breaking changes are also included as part of this PR:
>
> * conflicting JAR files in the classpath are not detected anymore
> * with the autoloadBioFormats option turned on, bfCheckJavaPath() does not throw an exception anymore when no suitable JAR file is found but the return value of the function should be checked.
>
> Are the breaking changes listed above be here be design? If so, we could review and include them for Bio-Formats 5.2.0 while backwards compatible changes could be included in Bio-Formats 5.1.x.

Yes, I made these changes by design.

1. What's the issue with conflicting jar files?  If bioformats is working, why would we bother the user?  I couldn't find any details about why this was done in f8888afa1e3fec6666339de9827d5492c848fe1b If there's a specific thing that breaks and requires such check, this could be added as another check done by `has_working_bioformats()`.

2. Previously, with `autoloadBioFormats` false or true, one would have to check for return value or catch an exception respectively.  This two patterns within the same function seemed odd to me.  
It was also possible that no exception was throw, the function returned true, and there was no working bioformats (it would require an unsuitable `bioformats_package.jar` which I agree is unlikely).  That the function returned true in such case is a bug which I fixed with the [first commit in this pull request](https://github.com/carandraug/bioformats/commit/351879a0a052649628684a2f089dce3857ea84a3#diff-3eafdd860da0b2af6a7d99cfb8b36999L72). It remains that the exception would mean no jar file found, while the return code would mean working bioformats found.  Therefore, even with `autoloadBioFormats` set to true, one would always have to check the return code, in addition to catch any possible exception.

sbesson:
Thanks for the thorough answer. Likely I should have asked for the scope of this PR first: do you see this as a 5.1.x or is it a 5.2.x feature? For most points above, my main concern here is mostly the breaking nature of some of the changes.

Answering specifically the three points mentioned above:

- on the return code logic, dropping the boolean vs exception ambiguity makes sense. 
- on the MATLAB vs Java path front, I agree with you that the MATLAB path is not the best place to add the JAR detection logic. Dropping this JAR detection "feature" on 5.2.x seems reasonable to me. Thinking more about it, this set of functions is currently shipped in 2 ways: 
 - in Octave, the Bio-Formats JAR needs to be set in the static Java classpath (unless this has evolved?)
 - in MATLAB, the JAR file is shipped under the same folder as the functions.

 Thus I would argue the latter case is the only one we truly need to support and maintain in `bfCheckJavaPath`. For any other location, either a system wide installation or a development JAR, I would make it the user responsibility to set the Java classpath accordingly to include the correct JAR.
- on the conflicting JARs, this warning was introduced together with `bioformats_package.jar` in 5.0.x (see https://github.com/openmicroscopy/bioformats/pull/888). The only notable difference between both JAR are the included logging JARs (`log4j` in `loci_tools` vs `logback` on `bioformats_package`). While it made sense at that time, dropping it makes sense since this new implementation no longer retrieves the Java class path.

carandraug:
> Thanks for the thorough answer. Likely I should have asked for the scope of this PR first: do you see this as a 5.1.x or is it a 5.2.x feature?

I am used that patch releases only fix regressions.  This is not the case for bioformats and I don't know how you decide what goes in a patch or in a minor release.

If we agree that these backwards incompatible changes are acceptable for 5.2, then I'm happy with it.  That seems better than making a different set of changes for 5.1 only.

> [...]
>   * in Octave, the Bio-Formats JAR needs to be set in the static Java classpath (unless this has evolved?)

No, this has not evolved.  But bioformats' `PKG_ADD` will check this. It runs a snippet that fails unless it's on the static classpath (checking parent classes of java objects with `isa()`). Once Octave fixes it, it will automatically start working in bioformats.

> [...]
> * Thus I would argue the latter case [JAR file is shipped under the same folder as the functions] is the only one we truly need to support and maintain in bfCheckJavaPath.

This patch still supports it and has it as the preferred directory to look for jars.

> For any other location, either a system wide installation or a development JAR, I would make it the user responsibility to set the Java classpath accordingly to include the correct JAR.

So should I remove the extra directories being searched?  They are not standards, there is none for java, but they are very widespread conventions.

sbesson:
:+1: So my vote would be for removing the extra directories search logic and keeping the rest of the changes as they are for inclusion into Bio-Formats 5.2.x unless @melissalinkert has other recommendations.

carandraug:
Should I go ahead and remove the search for the installed jar files? Or should I wait for feedback from @melissalinkert ?

sbesson:
@carandraug: I would say go ahead and remove the search part. Ping me when this is pushed. I will review the PR when ready.

carandraug:
ping

sbesson:
https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-merge-matlab/ being green. Local tests are passing with the various scenarios described in the PR thread. Merging. Thanks @carandraug.

sbesson:
--no-rebase

PR #2030, opened 2015-10-16
cope with pixel unit length

bjoernthiel:
please review

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/Calibrator.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/Calibrator.java
@@ -68,11 +68,14 @@ public class Calibrator {
     double zcal = Double.NaN, tcal = Double.NaN;
 
     Length xd = meta.getPixelsPhysicalSizeX(series);
-    if (xd != null) xcal = xd.value(UNITS.MICROM).doubleValue();
+    if (xd != null && xd.unit().isConvertible(UNITS.MICROM)) 
+        xcal = xd.value(UNITS.MICROM).doubleValue();
     Length yd = meta.getPixelsPhysicalSizeY(series);
-    if (yd != null) ycal = yd.value(UNITS.MICROM).doubleValue();
+    if (yd != null && yd.unit().isConvertible(UNITS.MICROM)) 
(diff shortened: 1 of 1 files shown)


sbesson:
@bjoernthiel: could you give us more information on either the scenario and/or a sample image to test this change?

melissalinkert:
Including in builds, as @sbesson was able to reproduce the problem that this solves by using UNITS.PIXEL and/or UNITS.REFERENCE_FRAME in calls to setPixelsPhysicalSize* (from a reader).  I think this should be fine to merge assuming all builds green tomorrow (and then we can rebase for dev_5_1).

For future PRs, @bjoernthiel, please do include a description of the problem that is being solved, as this helps us to understand the proposed change and thus reduce the time needed for review.

sbesson:
For reference, a minimal scenario to reproduce the bug is to create an OME-TIFF file as follows:

```
$ bfconvert test\&physicalSizeX=1.fake test.ome.tiff
$ tiffcomment test.ome.tiff > test.ome
$ sed -i '' -e 's/µm/pixel/g' test.ome
$ tiffcomment -set test.ome test.ome.tiff
```

Importing this file under ImageJ/Fiji using the Bio-Formats Importer results in a `NullPointerException` and the code change proposed here should correct this. To be tested tomorrow.

sbesson:
Tested using the scenario described above and ImageJ. Without this PR, the image opening gives a `NullPointerException`. With this PR included, the image opens. Thanks @bjoernthiel. MErging and will rebase it on dev_5_1

sbesson:
--rebased-to #2041

bjoernthiel:
@melissalinkert @sbesson sorry for any inconvenience. 
That issue came on my desk half an hour before I left for one week to stay on an island...
Can we have this delivered as a patch so that people using FIJI can import Imspector stacks with unit length 'pixel' in z direction?

sbesson:
Hi @bjoernthiel, no worries at all.
We independently figured out how to reproduce the bug and your use case matches our findings. As we recently split the development lines of Bio-Formats, I rebased your changes against `dev_5_1` - see https://github.com/openmicroscopy/bioformats/pull/2041 - so that this patch can be included in the next 5.1.6 release of Bio-Formats.

PR #2036, opened 2015-10-22
Jpeg physical dims

dgault:
Reading Physical Size X and Y from Exif Tags along with units.

Note the alternative Exif tags Exif.Image.YResolution and
 Exif.Image.ResolutionUnit are not supported by the Exif library we are
 using.

Alternative tags Exif.Image.FocalPlaneXResolution,
 Exif.Image.FocalPlaneYResolution and Exif.Image.FocalPlaneResolutionUnit
 are used instead.

https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7877&p=16128#p16128

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/JPEGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/JPEGReader.java
@@ -54,7 +54,11 @@ import java.util.HashMap;
 import org.joda.time.DateTime;
 import loci.formats.MetadataTools;
 import loci.formats.meta.MetadataStore;
+import ome.units.UNITS;
+import ome.units.quantity.Length;
 import ome.xml.model.primitives.Timestamp;
+
+import com.drew.metadata.exif.ExifSubIFDDirectory;
 
 /**
  * JPEGReader is the file format reader for JPEG images.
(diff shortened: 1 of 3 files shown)


components/formats-bsd/src/loci/formats/in/JPEGReader.java
```diff
@@ -207,8 +211,38 @@ protected void initFile(String id) throws FormatException, IOException {
...
+                sizeY = FormatTools.getPhysicalSizeY(Double.parseDouble(tags.get(tagName)), UNITS.INCH);
+              }
+              else if (Integer.parseInt(units) == 3) { 
+                sizeY = FormatTools.getPhysicalSizeY(Double.parseDouble(tags.get(tagName)), UNITS.CM);
+              }
+            }
+            store.setPixelsPhysicalSizeX(sizeY, 0);
+          }
+          else {
+            addGlobalMeta(tagName, tags.get(tagName));
```
rleigh-dundee: Should every tag be added to the global metadata?  i.e. should this be removed from the else block to it's used unconditionally?

dgault:
Upon further testing the alternative tags used don't appear to be suitable for the purposes we are using them here:
Exif.Image.FocalPlaneXResolution,
Exif.Image.FocalPlaneYResolution
Exif.Image.FocalPlaneResolutionUnit

PR #2037, opened 2015-10-22
Add a "Multiple Images" column to the supported formats table

melissalinkert:
See https://trello.com/c/9yZdHX52/15-mif-checkbox-in-pff-list.  The checkbox seemed to make more sense on the supported formats table, but I can move it to the dataset structure table if anyone feels strongly.


melissalinkert:
*sigh* Wrong branch...

PR #2038, opened 2015-10-22
Add a "Multiple Images" column to the supported formats table

melissalinkert:
See https://trello.com/c/9yZdHX52/15-mif-checkbox-in-pff-list. The checkbox seemed to make more sense on the supported formats table, but I can move it to the dataset structure table if anyone feels strongly.

### DIFF
--- a/components/autogen/src/FormatPageAutogen.java
+++ b/components/autogen/src/FormatPageAutogen.java
@@ -115,6 +115,7 @@ public class FormatPageAutogen {
       context.put("utilityRating", table.get("utilityRating"));
       context.put("reader", table.get("reader"));
       context.put("writer", table.get("writer"));
+      context.put("mif", table.get("mif"));
       context.put("notes", table.get("notes"));
       context.put("privateSpecification", table.get("privateSpecification"));
       context.put("readerextlink",
--- a/components/autogen/src/doc/FormatTable.vm
+++ b/components/autogen/src/doc/FormatTable.vm
@@ -7,12 +7,12 @@ Supported Formats
 
     You can sort this table by clicking on any of the headings.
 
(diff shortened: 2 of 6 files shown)


sbesson:
Quickly reviewed this PR. Looks good to me although I didn't check the mapping between the MIF column and the dataset structure table into details. I wonder if this PR does not imply the deprecation of the dataset structure table for two reasons:
- the main column of this table (i.e. the fileset structure) is now partly covered in the supported formats table
- the table has a reader/format ambiguity (it is really per reader and not per format)
The expected fileset structure stored in the reader could arguably be listed in the reader metadata page instead. Thoughts?

melissalinkert:
No objections to deprecating the dataset structure table, or at least putting the same information on each format page.  I'll have to think about how best to do it, since the dataset structure is encoded in the readers, not ```format-pages.txt``` like everything else in the reader/supported formats pages.  OK to do as a follow-up PR?

sbesson:
Sure. The other changes in this PR (`ndpis`, `LegacyND2Reader`) also look good.

joshmoore:
I have a strange rendering in Chrome on OSX:

screen shot 2015-10-24 at 00 41 35

sbesson:
I could not reproduce this on latest Chrome on OSX 10.10.5. Does it still happen?

joshmoore:
Just checked, yes. Version 46.0.2490.71 (64-bit) on 10.10.5 as well.

hflynn:
Looks fine to me on Chrome, Firefox and Safari but if there is a problem with some versions of Chrome it'll be because the vertical table headings are images not text.

melissalinkert:
--rebased-to #2086

PR #2039, opened 2015-10-22
IMODReader: add support for native length units

sbesson:

This PR adjusts the IMODReader to store the physical pixel size read from `imod` files using the native units. The specification for the units can be found at http://bio3d.colorado.edu/imod/doc/binspec.html.

Since one of the samples IMOD file uses the pixel unit (`0`) which that cannot be converted to micrometers, the format reader automated tests are also adjusted to handle this case (using `isConvertible` similarly to what is done in https://github.com/openmicroscopy/bioformats/pull/2030). The size comparison logic used across all physical size tests is also refactored into a single helper method to simplify future maintenance.

For automated tests, this means we do not differentiate the case where the physical size is 1 micron or 1 pixel. In the future we may want to introduce a `physicalSize{X,Y,Z}Unit` optional configuration to store non-convertible length unit.

To test this PR, check `imod` files are read correctly and automated tests are passing.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/IMODReader.java
+++ b/components/formats-gpl/src/loci/formats/in/IMODReader.java
@@ -457,7 +457,7 @@ public class IMODReader extends FormatReader {
         }
       }
       if (physicalZ > 0) {
-        Length z = FormatTools.getPhysicalSizeX(physicalZ, physicalSizeUnit);
+        Length z = FormatTools.getPhysicalSizeZ(physicalZ, physicalSizeUnit);
         if (z != null) {
           store.setPixelsPhysicalSizeZ(z, 0);
         }
--- a/components/formats-gpl/src/loci/formats/in/IMODReader.java
+++ b/components/formats-gpl/src/loci/formats/in/IMODReader.java
@@ -36,7 +36,9 @@ import loci.formats.FormatReader;
 import loci.formats.FormatTools;
 import loci.formats.MetadataTools;
 import loci.formats.meta.MetadataStore;
+
 import ome.units.UNITS;
+import ome.units.unit.Unit;
 import ome.units.quantity.Length;
 import ome.xml.model.primitives.Color;
 import ome.xml.model.primitives.NonNegativeInteger;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/IMODReader.java
```diff
@@ -441,45 +443,52 @@ else if (chunkType.equals("MINX")) {
...
-          FormatTools.createLength(adjustForUnits(pixSizeUnits, physicalY), UNITS.MICROM), 0);
+        Length y = FormatTools.getPhysicalSizeY(physicalY, physicalSizeUnit);
+        if (y != null) {
+          store.setPixelsPhysicalSizeY(y, 0);
+        }
       }
       if (physicalZ > 0) {
-        store.setPixelsPhysicalSizeZ(
-          FormatTools.createLength(adjustForUnits(pixSizeUnits, physicalZ), UNITS.MICROM), 0);
+        Length z = FormatTools.getPhysicalSizeX(physicalZ, physicalSizeUnit);
```
mtbc: not getPhysicalSizeZ?

components/test-suite/src/loci/tests/testng/FormatReaderTest.java
```diff
@@ -897,6 +897,34 @@ public void testLittleEndian() {
     result(testName, true);
   }
 
+  public void compareSizes(Double expectedSize, Length realSize, String testName, int i) {
+    if (expectedSize == null || expectedSize == 0d) {
```
mtbc: doesn't ` if (expectedSize == 0d) {` suffice?

components/formats-gpl/src/loci/formats/in/IMODReader.java
sbesson: Thanks

sbesson:
Discussed this morning with @dgault who will cherry-pick the test commits and work on improving the physical size test logic to compare Length quantiites instead of converting units to micrometer.

sbesson:
--exclude to test https://github.com/openmicroscopy/bioformats/pull/2050 first

dgault:
Changes are in line with the specification listed.
Files can be read correctly and appropriate units are selected.
Automated tests are also passing.

PR looks good to me and is ready to be merged

sbesson:
--rebased-to #2093

PR #2040, opened 2015-10-22
Video spot tracker

sbesson:
This PR:

- adds the Video Spot Tracker software to the list of projects using the Jace C++ bindings - see http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-October/005690.html
- removes the BF-CPP abbreviation and uses Jace C++ bindings in the documentation instead to prevent any confusion with Bio-Formats C++. /cc @rleigh-dundee

### DIFF
--- a/docs/sphinx/developers/jace/overview.txt
+++ b/docs/sphinx/developers/jace/overview.txt
@@ -21,12 +21,12 @@ Other projects using the JACE C++ bindings include:
 -  :doc:`/users/xuvtools/index` which uses an adapted version of the JACE C++
    bindings called
    `BlitzBioFormats <http://www.xuvtools.org/devel:libblitzbioformats>`_.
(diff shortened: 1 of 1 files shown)


docs/sphinx/developers/jace/overview.txt
```diff
@@ -2,29 +2,34 @@ JACE C++ bindings for the Java API
 ==================================
 
 To make Bio-Formats accessible to software written in C++, we have
```
mtbc: To make the Bio-Formats Java classes accessible ... ? Or maybe something else mentioning Java BF?

Just wondering if there's some way we can make it clearer what this is about: after all, the native implementation is also to make BF accessible to software written in C++, albeit with rather fewer readers at present!

mtbc:
Good to merge as-is.

mtbc:
Their source download is actually a git repository.

sbesson:
--rebased-to #2083

PR #2042, opened 2015-10-23
bump model doc reference to new 5.2 OMERO

mtbc:
Effects https://trello.com/c/lUQDSthL/81-update-model-docs-to-point-at-omero-5-2-docs. Staged at http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/users/index.html#omero.

### DIFF
--- a/docs/sphinx/conf.py.in
+++ b/docs/sphinx/conf.py.in
@@ -135,7 +135,7 @@ oo_root = 'http://www.openmicroscopy.org'
 oo_site_root = oo_root + '/site'
 lists_root = 'http://lists.openmicroscopy.org.uk'
 downloads_root = 'http://downloads.openmicroscopy.org'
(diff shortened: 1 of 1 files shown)


sbesson:
This likely does not apply to the `develop` branch but to the `dev_5_1` branch?

mtbc:
Ah, probably it does apply to 5.1, you're right. It does apply here too? Removed --no-rebase.

mtbc:
> javac: invalid target release: 1.7

Something seems to need fixing in our BF Travis setup.

sbesson:
No it does not apply to `develop` which will point to a different version of OMERO.

mtbc:
Well, yes, but it surely applies no less than the current line and could cause conflicts in the meantime if trying to rebase PRs that change nearby lines.

mtbc:
Replaced by #2043.

PR #2043, opened 2015-10-23
bump model doc reference to new 5.2 OMERO

mtbc:
Effects https://trello.com/c/lUQDSthL/81-update-model-docs-to-point-at-omero-5-2-docs. Staged at http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/users/index.html#omero.

--no-rebase

### DIFF
--- a/docs/sphinx/conf.py.in
+++ b/docs/sphinx/conf.py.in
@@ -135,7 +135,7 @@ oo_root = 'http://www.openmicroscopy.org'
 oo_site_root = oo_root + '/site'
 lists_root = 'http://lists.openmicroscopy.org.uk'
 downloads_root = 'http://downloads.openmicroscopy.org'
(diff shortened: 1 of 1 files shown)


hflynn:
Hi @mtbc great of you to be helpful, but this isn't actually the change I had in mind - it is the formats folder of the ome-documentation repo that I meant not Bio-Formats.

mtbc:
Ha. Is this change useful or not?

hflynn:
Probably not just now. I need to check where OMERO docs links occur but now we've decoupled, changing it to a version numberless reference so it just always points to the latest release is probably best.

mtbc:
Okay, closing then. Happy to also close https://github.com/openmicroscopy/ome-documentation/pull/1302 if that's also not what's required.

PR #2044, opened 2015-10-23
AliconaReader - Using native units

dgault:
Using metres rather then micrometers

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/AliconaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/AliconaReader.java
@@ -273,11 +273,11 @@ public class AliconaReader extends FormatReader {
       // populate Dimensions data
 
       if (pntX != null && pntY != null) {
-        double pixelSizeX = Double.parseDouble(pntX) * 1000000;
-        double pixelSizeY = Double.parseDouble(pntY) * 1000000;
+        double pixelSizeX = Double.parseDouble(pntX);
+        double pixelSizeY = Double.parseDouble(pntY);
 
-        Length sizeX = FormatTools.getPhysicalSizeX(pixelSizeX);
-        Length sizeY = FormatTools.getPhysicalSizeY(pixelSizeY);
+        Length sizeX = FormatTools.getPhysicalSizeX(pixelSizeX, UNITS.METRE);
+        Length sizeY = FormatTools.getPhysicalSizeY(pixelSizeY, UNITS.METRE);
 
         if (sizeX != null) {
           store.setPixelsPhysicalSizeX(sizeX, 0);


sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/80. It seems we are hitting the same issue as in https://github.com/openmicroscopy/bioformats/pull/1895#issuecomment-125122293, i.e. the stored value in meters is lower than `0.000001` and does not passes the check in `FormatTools.isPositiveValue()`:

```
[hudson@ome-c6100-3 BIOFORMATS-5.1-merge-build]$ ./tools/showinf -nopix -debug /ome/data_repo/from_skyking/al3d/jens/Abformung\ 1.al3d | grep Physical
Expected positive value for PhysicalSizeX; got 4.37917E-7
Expected positive value for PhysicalSizeY; got 4.37917E-7
```

I am not sure about the proper way to deal with these native units with very low values. Options that come to my mind are:
- leave the reader as it is i.e. preserve the statu quo and always convert to microns
- test the size value is larger than `Constants.EPSILON` in the reader itself and choose the appropriate value/unit, converting if necessary
- relax the condition in `FormatTools.getPhysicalSize{XYZ}()` to allow lower values

Thoughts @melissalinkert ?

melissalinkert:
I think we'll need to leave the check for very small values, otherwise we're just deferring the problem.

Maybe it makes sense to keep this change (and other reader changes like it), but update ```FormatTools.getPhysicalSize*``` and similar to:

1. check if the value is greater than ```Constants.EPSILON```
2. if so, use whatever unit was supplied
3. if not, convert to the smallest order of magnitude unit such that (1) passes (or the default unit?)
4. if the value really is 0, or is less than ```Constants.EPSILON``` for even the smallest defined unit, return null

dgault:
Created https://github.com/openmicroscopy/bioformats/pull/2071 which would implement the above logic into FormatTools. The PR only covers Length values so would need to be increased for other Units if approved.

melissalinkert:
Does this need to be updated for inclusion in 5.1.6?

dgault:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/173/consoleFull

dgault:
Related config PR - https://github.com/openmicroscopy/data_repo_config/pull/46

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/182/console looks good/

dgault:
Note: This PR requires https://github.com/openmicroscopy/bioformats/pull/2084

sbesson:

dgault:
--rebased-to #2141

PR #2047, opened 2015-10-23
Travis: Only test openjdk7

sbesson:
Since Oracle JDK 7 should be tested via the OME CI infrastructure, testing it on the Travis matrix is redundant with openjdk7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ addons:
 
 jdk:
   - oraclejdk8
-  - oraclejdk7
   - openjdk7
 
 env:
@@ -46,8 +45,6 @@ script:
 
 matrix:
   exclude:
-    - jdk: oraclejdk7
-      env: BUILD=sphinx_html
     - jdk: openjdk7
       env: BUILD=sphinx_html
     - jdk: oraclejdk8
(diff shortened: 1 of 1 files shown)


joshmoore:
That looks significant -- 2h20 to 1h15:
screen shot 2015-10-24 at 01 37 47

mtbc:
If I log into trout and eel I have openjdk7 in my default path. Is there something in the job configs that I'm missing, or maybe in hudson's profile, that has it use oraclejdk7 for jobs?

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/16/injectedEnvVars/. The `JAVA_HOME` is using Jenkins installed JDK which should be oracledjdk8 in that case.

mtbc:
:+1: Yes, that looks like Oracle on both trout and eel, and these changes look fine. So we can burn a little less needless CPU and save the environment that little bit more.

Good to merge.

mtbc:
i.e. shouldn't we be removing oraclejdk8 from the build matrix instead of oraclejdk7?

sbesson:
@mtbc: the current statu quo is to keep testing all changes with either JDK 7 or JDK 8 on Travis. The goal here was to remove the `openjdk7/oracldejdk7` duplication. Removing `oraclejdk8` could discussed separately though.

mtbc:
Okay, thank you. Good to merge.

PR #2049, opened 2015-10-25
C++: Enable minimal C++11 features

rleigh-dundee:
Library features:
- `std::shared_ptr` and related classes from `<memory>`.

Language features:
- `<::` is not  a trigraph, but a template opening followed by an absolute namespace
- `>>` is two templates closing not a right-shift operator

There are no changes to the program logic in this PR.  The change to use `std::shared_ptr` is simply a change from using `ome::compat::shared_ptr` (which is effectively `std::shared_ptr`), and the language changes are simply removing extra spaces where they are no longer needed to avoid parsing ambiguity.  E.g. `std::vector<ome::compat::shared_ptr< ::ome::xml::model::OMEModelObject> >` will now be `std::vector<std::shared_ptr<::ome::xml::model::OMEModelObject>>`.  The main point of this PR is not to change to code to take advantage of any significant new C++ functionality, but to verify that the CI build nodes can put the compiler into C++11 mode and successfully build with some very trivial changes.  More useful changes such as `unique_ptr` and rvalue references can come later.

--breaking

Marking breaking since it will require job changes and an additional commit to fix noexcept for VS2013 (it's missing).  Not for review until it's working on all platforms.

--------

Testing: Covered entirely by the CI jobs since if it's broken it won't build, or the tests will fail.

### DIFF
--- a/components/xsd-fu/python/ome/modeltools/property.py
+++ b/components/xsd-fu/python/ome/modeltools/property.py
@@ -455,7 +455,7 @@ class OMEModelProperty(OMEModelEntity):
                 else:
                     itype = "const %s&" % self.langTypeNS
             elif self.maxOccurs > 1 and not self.parent.isAbstractProprietary:
-                itype = "std::vector<std::shared_ptr<%s> >&" % ns_sep
+                itype = "std::vector<std::shared_ptr<%s>>&" % ns_sep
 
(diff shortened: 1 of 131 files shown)


rleigh-dundee:
Closing - I've redone this after the split.

PR #2050, opened 2015-10-26
Test suite refactor native units

dgault:
Adding support for the following configuration options:
- TimeIncrementUnits
- ExposureTimeUnits
- PhysicalSizeZUnits
- PhysicalSizeYUnits
- PhysicalSizeXUnits
- EmissionWavelengthUnits
- ExcitationWavelengthUnits

When read the string units types will be converted into an appropriate
Length or Time object. If not present the default units will be used instead (mircometres for
sizes, nm for wavelengths and seconds for time).

FormatTestReader now compares Length objects rather than double values.

Extra utility methods added to FormatTools to easily convert from String to appropriate time and wavelength values.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1398,7 +1398,7 @@ public final class FormatTools {
    * given unit.
    *
    * @param value  the value of the wavelength
-   * @param unit   the unit of the wavelength
+   * @param unit   the unit of the wavelength. If null will default to Nanometre
    *
    * @return       the wavelength formatted as a {@link Length}
 
@@ -1419,7 +1419,7 @@ public final class FormatTools {
    * given unit.
    *
    * @param value  the value of the time
-   * @param unit   the unit of the time
+   * @param unit   the unit of the time. If null will default to Seconds
    *
    * @return       the wavelength formatted as a {@link Length}
 
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -54,6 +54,7 @@ import loci.formats.services.OMEXMLServiceImpl;
 
 import ome.xml.model.enums.EnumerationException;
 import ome.xml.model.enums.UnitsLength;
+import ome.xml.model.enums.UnitsTime;
 import ome.xml.model.primitives.PrimitiveNumber;
 import ome.xml.model.primitives.PositiveFloat;
 import ome.xml.model.primitives.PositiveInteger;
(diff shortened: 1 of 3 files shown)


components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -104,7 +104,14 @@
   private static final String X_POSITION = "PositionX_";
   private static final String Y_POSITION = "PositionY_";
   private static final String Z_POSITION = "PositionZ_";
-
+  private static final String TIME_INCREMENT_UNITS = "TimeIncrementUnits";
```
sbesson: Is there a reason to pluralize these keys since there is only one unit per metadata type?

components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -240,39 +247,48 @@ public String getTileAlternateMD5() {
...
   }
 
-  public Double getPhysicalSizeX() {
+  public Length getPhysicalSizeX() {
     String physicalSize = currentTable.get(PHYSICAL_SIZE_X);
     if (physicalSize == null) return null;
+    String sizeXUnits = currentTable.get(PHYSICAL_SIZE_X_UNITS);
     try {
-      return new Double(physicalSize);
+      return physicalSize == null ? null : FormatTools.getPhysicalSize(new Double(physicalSize), sizeXUnits);
```
sbesson: Many configuration files include

```
PhysicalSizeX = 0.0
PhysicalSizeY = 0.0
```

which then lead to test errors or type

```
  [testng] [2015-10-26 15:27:02,594] [pool-1-thread-4] 	PhysicalSizeZ: FAILED (Series 0 (expected ome.units.quantity.Length: value[0.0], unit[µm] stored as java.lang.Double, actual null))
```

see https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/83/console

For now we might want to handle `physicalSize == 0d` as well in `Configuration.getPhysicalSize` and have a cleanup later.

components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -240,39 +247,48 @@ public String getTileAlternateMD5() {
...
   }
 
-  public Double getPhysicalSizeX() {
+  public Length getPhysicalSizeX() {
     String physicalSize = currentTable.get(PHYSICAL_SIZE_X);
     if (physicalSize == null) return null;
+    String sizeXUnits = currentTable.get(PHYSICAL_SIZE_X_UNIT);
     try {
-      return new Double(physicalSize);
+      return physicalSize == null ? null : FormatTools.getPhysicalSize(new Double(physicalSize), sizeXUnits);
```
sbesson: See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-build/1059/findbugsResult/

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1392,6 +1393,48 @@ public static boolean isPositiveValue(Double value) {
...
+  /**
+   * Formats the input value for the time into a length of the
+   * given unit.
+   *
+   * @param value  the value of the time
+   * @param unit   the unit of the time
+   *
+   * @return       the wavelength formatted as a {@link Length}
+
+   */
```
melissalinkert: Maybe note the default units in the javadocs for each of these methods?

components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -316,14 +337,26 @@ public Double getPositionZ(int plane) {
...
+    catch (NumberFormatException e) { 
+      return null;
+    } 
   }
 
-  public Double getExcitationWavelength(int channel) {
+  public Length getExcitationWavelength(int channel) {
     String wavelength = currentTable.get(EXCITATION_WAVELENGTH + channel);
-    return wavelength == null ? null : new Double(wavelength);
+    String excitationUnits = currentTable.get(EXCITATION_WAVELENGTH_UNIT);
```
melissalinkert: Worth allowing each wavelength to have its own units?

melissalinkert:
Generally reads fine to me - comments are mostly nitpicky and can be addressed separately if desired. As mentioned to @sbesson, I'd assume this is fine to merge if builds are passing.

sbesson:
Looking at https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/98/console and the current full repository run, I am afraid we are creating unnecessary double precision maintenance for the configuration files by using directly `Length.equals()` for comparing quantities. Additionally, I just realized this method has some logic to convert the value which is not what we want to in these automated tests.

I would propose we implement an internal method a la `FormatReaderTest.compareQuantities(q1,q2)` which would minimally return the output of `q1.unit.equals(q2.unit()) && Math.abs(q1.value().doubleValue() - q2.value().doubleValue()) <= Constants.EPSILON`. This should preserve the current statu quo in terms of double precision and force us to update only the configuration files with incorrect native units. Thoughts?

sbesson:
See https://github.com/dgault/bioformats/pull/1

dgault:
That makes sense to me Sebastien, merged your changes into the PR

sbesson:
Looks good to me. https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/661/ is green with the relevant configuration changes. Merging and we can have follow-up PRs (one of the RFE is to modify the configuration files generation to handle units).

sbesson:
--rebased-to #2059

PR #2051, opened 2015-10-26
specification: Consolidate schema into single file

rleigh-dundee:
- Move BinaryFile, SA, SPW and ROI schemas into OME schema
- Upgrade and downgrade transforms should handle the namespace migration (but I'm no XSLT expert and am not 100% sure this is correct--needs testing)
- xsdfu is invoked with a single schema
- delete unused schemas and drop from catalogue
- drop/update namespace prefixes
- change tabs to spaces in the schema

xsd-fu can still process multiple schemas--needed for C++/2013-06 until we can upgrade to 2015-01.

--------

Outstanding questions:

- [ ] Is the change to OMEXMLWriter appropriate; does it need to care about multiple  model versions here?
- [x] Why is this not now generating proper BinData methods in the metadata code?  The generator bug *should* have been fixed, unless we're running into additional problems (ordering within the schema?)  That it would work immediately is probably wishful thinking on my part.  What else needs changing?
- [ ] What should the version number on this be?

--------

Testing:

- check generated source diff
- check changes to ome.xsd, ignoring whitespace changes
- test upgrade and downgrade, in particular the migration to the OME namespace on upgrade and splitting back into separate namespaces on downgrade

--------

/cc @mtbc @dgault @joshmoore


joshmoore:
Added a question about whether this is DEV0 or DEV1? cc: @mtbc

joshmoore:
Added a question about whether this is DEV0 or DEV1? cc: @mtbc

mtbc:
I don't see a problem with just taking it as expanding DEV0: until we actually release a milestone version people won't already have DEV0-schema XML lying around? Then we can keep functions like `OMEXMLServiceImpl.transformToLatestVersion` simpler (though perhaps that should be using `ome-transforms.xml` anyway!). Would want to bump versions after we tag a developer preview though.

rleigh-dundee:
The following change is needed to make the BinData metadata methods get generated, though it does not yet produce valid generated code (needs additional template/generator tweaks):

```
diff --git a/components/xsd-fu/python/ome/modeltools/config.py b/components/xsd-fu/python/ome/modeltools/config.py
index c1919c8..c6ee732 100644
--- a/components/xsd-fu/python/ome/modeltools/config.py
+++ b/components/xsd-fu/python/ome/modeltools/config.py
@@ -8,7 +8,7 @@ import re
 # Types which should be ignored from metadata store, retrieve, etc. code
 # generation due either to their incompatibility or complexity as it applies
 # to these interfaces and implementations.
-METADATA_OBJECT_IGNORE = ('BinData', 'External', 'MapPairs', 'M')
+METADATA_OBJECT_IGNORE = ('MapPairs', 'M')

 # Type counts which should be ignored from metadata store, retrieve, etc. code
 # generation due either to their incompatibility or complexity as it applies
```

rleigh-dundee:
@mtbc https://trello.com/c/AAmXgsAQ/61-make-shape-a-complextype is a card for followup changes to allow proper changes to the code generation to add the missing `Mask` `BinData` methods.  This should probably be done as a separate change.

dgault:
Diffs of generated source files and changes to ome.xsd are as expected and are correct.
Upon testing the transforms, the upgrade transform appears to run as expected.
Wehen running the downgrade transform the following compilation errors are shown:
compilation error: file /Users/dgault/Documents/GitHub/bioformats-breaking/components/specification/transforms/2016-DEV0-to-2015-01.xsl line 40 element stylesheet
xsl:exclude-result-prefixes : undefined namespace Bin
compilation error: file /Users/dgault/Documents/GitHub/bioformats-breaking/components/specification/transforms/2016-DEV0-to-2015-01.xsl line 40 element stylesheet
xsl:exclude-result-prefixes : undefined namespace SPW
compilation error: file /Users/dgault/Documents/GitHub/bioformats-breaking/components/specification/transforms/2016-DEV0-to-2015-01.xsl line 40 element stylesheet
xsl:exclude-result-prefixes : undefined namespace SA
compilation error: file /Users/dgault/Documents/GitHub/bioformats-breaking/components/specification/transforms/2016-DEV0-to-2015-01.xsl line 40 element stylesheet
xsl:exclude-result-prefixes : undefined namespace ROI

The transform still runs successfully despite the errors. Adding in the namespaces at the top of the stylesheet removes the error completely

components/specification/transforms/2016-DEV0-to-2015-01.xsl
```diff
@@ -33,10 +33,6 @@
 
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
     xmlns:OME="http://www.openmicroscopy.org/Schemas/OME/2016-DEV0"
-    xmlns:Bin="http://www.openmicroscopy.org/Schemas/BinaryFile/2016-DEV0"
-    xmlns:SPW="http://www.openmicroscopy.org/Schemas/SPW/2016-DEV0"
-    xmlns:SA="http://www.openmicroscopy.org/Schemas/SA/2016-DEV0"
-    xmlns:ROI="http://www.openmicroscopy.org/Schemas/ROI/2016-DEV0"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
dgault: Removing the namespaces here results in an error in exclude-result-prefixes

rleigh-dundee: Removing the elements from the exclusion list is the real fix, it seems.  I've pushed a new commit which makes the namespace prefixes get unified to use `OME:` on upgrade and restored to their separate prefixes on downgrade.

dgault:
Retested the up and downgrade transforms by generating a sample xml from the new xsd, downgrading to 2015 and then upgrading the 2015 xml back to 2016.

Results all seem correct to me, the 2015 file after downgrade has the namespaces in the correct location and after upgrade they are all correctly stripped out.

Im happy to merge these changes.

rleigh-dundee:
@melissalinkert Is the OMEXMLWriter change here OK with you?  Am I correct that it can only ever write out the current model version, or is there some way this can be made to write out older versions?

joshmoore:
General sign-off during standup. Merging.

PR #2053, opened 2015-10-27
C++: Path introspection on Windows, and support for relocatable installation

rleigh-dundee:
- Add `relocatable-install` option (see https://github.com/ome/ome-cmake-superbuild/pull/10)
- Support path introspection on Windows, to match the existing Unix support
- Drop unused superbuild options (see again https://github.com/ome/ome-cmake-superbuild/pull/10)

----------

Testing: Covered by unit tests on Unix.  On Windows, apply this patch:

```
diff --git a/cpp/lib/ome/common/module.cpp b/cpp/lib/ome/common/module.cpp
index 0cc5580..42ae249 100644
--- a/cpp/lib/ome/common/module.cpp
+++ b/cpp/lib/ome/common/module.cpp
@@ -47,6 +47,7 @@
 #include <cstring>
 #include <map>
 #include <stdexcept>
+#include <iostream>
 
 namespace fs = boost::filesystem;
 
@@ -306,6 +307,7 @@ namespace ome
             {
               fs::path moduledir(module.parent_path());
               bool match = true;
+              std::cerr << "MODULEDIR: " << moduledir << '\n';
 
 #ifdef _MSC_VER
               fs::path libdir(INSTALL_BINDIR);
```

Then build and run:

```
<buildpath>>bf-test.bat cpp\libexec\info\Debug\info.exe --help
MODULEDIR: "<buildpath>\cpp\libexec\info\Debug"
E: Could not determine Bio-Formats runtime path for "doc" directory```

This is telling you that the runtime path was successfully introspected.  Since ome-common is static, it's giving you the path to the executable.  When made into a DLL, this will then work correctly.

### DIFF
--- a/cpp/cmake/Options.cmake
+++ b/cpp/cmake/Options.cmake
@@ -73,16 +73,3 @@ if(SPHINX_DEFAULT AND XELATEX AND MAKEINDEX)
   set(SPHINX_PDF_DEFAULT ON)
 endif()
 option(sphinx-pdf "Enable sphinx PDF documentation" ${SPHINX_PDF_DEFAULT})
-
-set(SUPERBUILD_OPTIONS
-    "-Dbioformats-superbuild:BOOL=OFF"
-    "-Dcxxstd-autodetect:BOOL=${cxxstd-autodetect}"
-    "-Dextra-warnings:BOOL=${extra-warnings}"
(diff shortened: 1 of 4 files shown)


rleigh-dundee:
--rebased-to #2054

PR #2055, opened 2015-10-27
C++: Path introspection on Windows, and support for relocatable installation (5.1)

rleigh-dundee:
- Add `relocatable-install` option (see https://github.com/ome/ome-cmake-superbuild/pull/11)
- Support path introspection on Windows, to match the existing Unix support
- Drop unused superbuild options (see again https://github.com/ome/ome-cmake-superbuild/pull/11)

----------

Testing: Covered by unit tests on Unix.  On Windows, apply this patch:

```
diff --git a/cpp/lib/ome/common/module.cpp b/cpp/lib/ome/common/module.cpp
index 0cc5580..42ae249 100644
--- a/cpp/lib/ome/common/module.cpp
+++ b/cpp/lib/ome/common/module.cpp
@@ -47,6 +47,7 @@
 #include <cstring>
 #include <map>
 #include <stdexcept>
+#include <iostream>
 
 namespace fs = boost::filesystem;
 
@@ -306,6 +307,7 @@ namespace ome
             {
               fs::path moduledir(module.parent_path());
               bool match = true;
+              std::cerr << "MODULEDIR: " << moduledir << '\n';
 
 #ifdef _MSC_VER
               fs::path libdir(INSTALL_BINDIR);
```

Then build and run:

```
<buildpath>>bf-test.bat cpp\libexec\info\Debug\info.exe --help
MODULEDIR: "<buildpath>\cpp\libexec\info\Debug"
E: Could not determine Bio-Formats runtime path for "doc" directory```

This is telling you that the runtime path was successfully introspected.  Since ome-common is static, it's giving you the path to the executable.  When made into a DLL, this will then work correctly.
```
--rebased-from #2053

### DIFF
--- a/cpp/cmake/Options.cmake
+++ b/cpp/cmake/Options.cmake
@@ -73,16 +73,3 @@ if(SPHINX_DEFAULT AND XELATEX AND MAKEINDEX)
   set(SPHINX_PDF_DEFAULT ON)
 endif()
 option(sphinx-pdf "Enable sphinx PDF documentation" ${SPHINX_PDF_DEFAULT})
-
-set(SUPERBUILD_OPTIONS
-    "-Dbioformats-superbuild:BOOL=OFF"
-    "-Dcxxstd-autodetect:BOOL=${cxxstd-autodetect}"
-    "-Dextra-warnings:BOOL=${extra-warnings}"
-    "-Dfatal-warnings:BOOL=${fatal-warnings}"
-    "-Dtest:BOOL=${test}"
-    "-Dextended-tests:BOOL=${extended-tests}"
-    "-Dembedded-gtest:BOOL=${embedded-gtest}"
-    "-Dqtgui:BOOL=${qtgui}"
-    "-Ddoxygen:BOOL=${doxygen}"
-    "-Dsphinx:BOOL=${sphinx}"
-    "-Dsphinx-pdf:BOOL=${sphinx-pdf}")
--- a/cpp/cmake/Options.cmake
+++ b/cpp/cmake/Options.cmake
@@ -16,6 +16,14 @@ option(test "Enable unit tests (requires gtest)" ON)
 option(extended-tests "Enable extended tests (more comprehensive, longer run time)" ON)
 option(embedded-gtest "Use embedded gtest rather than an external build" OFF)
 
+# The installation is relocatable; this affects path lookups (if OFF,
+# paths are assumed to be their configured absolute install location;
+# paths will still be introspected as a fallback); if ON paths will be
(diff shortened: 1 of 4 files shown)


rleigh-dundee:
@dgault To test the patch on Windows, I suggest:

- check out ome-cmake-superbuild.git, dev_5_1 branch
- check out bioformats.git, this PR branch plus the patch (make sure to commit it)

In a temporary build directory, run:

```
cmake -G "Visual Studio 12 2013 Win64" -Dsource-cache=\path\to\sources -Dbf-git-url=/path/to/bioformats.git -Dgf-git-branch=branch-name
cmake --build .
set "PATH=superbuild-install\bin;%PATH%"
bf-test.bat cpp\libexec\info\Debug\info.exe --help
```

Basically, we're doing a superbuild, but for the bioformats component, we configure it to use our local test branch rather than the default.

melissalinkert:
I assume this should be listed for 5.1.6 review?

rleigh-dundee:
@melissalinkert Yes.  I thought it had been a couple of times.  David did take a look but had some problem with building on his Windows system; if anyone who can build on Windows wants to take a look, that would be great.

dgault:
Sorry should have followed up on this earlier in the week. Ive tried testing this a few times but have always run into unrelated build issues. The major problems I was seeing was due to the version of cmake i had installed (3.3), within its findboost module it did not list nd therefore recognise boost 1.59. As such it was picking up an earlier 1.58 installation from elsewhere on my machine. This was leading to the boost part of the superbuild failing. However I was able to build the projects individually within Visual Studio.

rleigh-dundee:
@dgault Good to know you found the cause.  However, you shouldn't need to patch.  The superbuild passes the appropriate option to tell FindBoost about version 1.59, so as long as it's not finding Boost 1.58 then it should then work.  If it doesn't, please let us know what the logs show so that we can diagnose it further.  Thanks.

dgault:
I have now been able to test this successfully and view the expected behaviour

joshmoore:
thanks, @dgault ! Any tips that need sharing/documenting?

dgault:
At the end of the day the real issue for me was building on a machine that many previous builds and versions on it already. Cleaning up and removing any previous releases solved my issues with older versions of boost etc being picked up on my path. 

Using the cache while building also makes the process a lot easier once you have all the prerequisites downloaded, I had seen a number of failed builds due to being timed out while trying to retrieve some of these libraries.

joshmoore:
Makes sense. Thanks, guys.

rleigh-dundee:
--rebased-to #2146

rleigh-dundee:
--rebased-to #2146

Issue #2057, opened 2015-10-29
Split Channels setting is persistent when it shouldn't be


nashv:
Hello, 
I’ve encountered  the following behavior difference in calling Bio-formats manually versus from a macro. If someone runs Bio-Formats from the plugins menu and checks ‘split_channels’, that option is not cleared if subsequently Bio-formats is called from a macro that is not intended to split channels.   

for example, run from macro to split channels: 
 run(“Bio-Formats Importer”,”open=[bunchastuff] color_moded=Grayscale split_channels view=Hyperstack stack_order=default”); -> splits channels 

then  run from macro to NOT split channels: 
 run(“Bio-Formats Importer”,”open=[bunchastuff] color_moded=Grayscale  view=Hyperstack stack_order=default”); -> does not split channels 

Now, open Bio-Formats from plugins menu, manually select ‘split channels’ -> splits channels 

then run from macro to NOT split channels: 
 run(“Bio-Formats Importer”,”open=[bunchastuff] color_moded=Grayscale  view=Hyperstack stack_order=default”); ->  still splits channels 

Is there a way around this?

melissalinkert:
https://github.com/openmicroscopy/bioformats/pull/2061, which was included in 5.1.6, should have resolved this problem.  Please feel free to reopen if the problem persists.

PR #2061, opened 2015-11-02
Fiji preferences not persistent for macros

dgault:
https://github.com/openmicroscopy/bioformats/issues/2057

When the importer has been run from an IJ macro, the options list should
not be loaded from preferences but instead read from the macro arguments
only.

In order to test the PR:
Open Fiji and record 2 macros (Plugins -> Macros -> Record)
Now import a multi channel image using bioformats
For the first macro select the split channels option during import, for the second make sure it is unselected

Now import an image again using bioformats and ensure the split channel option is selected. This will save the option to your Image J preferences.
Now rerun your 2 macros and ensure that the image channels are split in one and combined in the other

Now import an image again using bioformats and ensure the split channel option is not selected.
Now rerun your 2 macros and ensure that they still both run correctly.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/Importer.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/Importer.java
@@ -29,6 +29,7 @@ package loci.plugins.in;
 
 import ij.ImageJ;
 import ij.ImagePlus;
+import ij.Macro;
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
All looks good.  Merging.

sbesson:
--rebased-to #2104

PR #2063, opened 2015-11-02
Slidebook: fix physical Z size calculation

melissalinkert:
See lists.openmicroscopy.org.uk/pipermail/ome-users/2015-October/005703.html.

Tested with ```slidebook/emma/WT BY4742 Nem1 mcherry genome.sld``` on Windows (with appropriate .dlls).  Without this change, the physical Z size was not set at all; with this change, it is set to the absolute difference between two adjacent Z sections.


dgault:
Change makes sense, have tested against multiple slidebook files and the calculated physical z sizes now appear correct.

melissalinkert:
--rebased-to #2087

PR #2065, opened 2015-11-03
Use native units for the pixels physical size of UnisokuReader

sbesson:

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/UnisokuReader.java
+++ b/components/formats-gpl/src/loci/formats/in/UnisokuReader.java
@@ -37,9 +37,12 @@ import loci.formats.FormatReader;
 import loci.formats.FormatTools;
 import loci.formats.MetadataTools;
 import loci.formats.meta.MetadataStore;
+
+import ome.units.quantity.Length;
+import ome.units.unit.Unit;
+import ome.units.UNITS;
 import ome.xml.model.primitives.PositiveFloat;
 import ome.xml.model.primitives.Timestamp;
-import ome.units.quantity.Length;
 
 /**
  * UnisokuReader is the file format reader for Unisoku STM files.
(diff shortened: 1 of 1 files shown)


dgault:
The Unisoku changes look good to me. Seeing some unrelated failures for gatan files but not caused by changes to Unisoku.

components/formats-gpl/src/loci/formats/in/UnisokuReader.java
```diff
@@ -206,7 +211,7 @@ else if (key.startsWith(":y_data ->")) {
             pixelSizeY = Double.parseDouble(v[2]) - Double.parseDouble(v[1]);
             pixelSizeY /= getSizeY();
             if (unit.equals("nm")) {
-              pixelSizeY /= 1000;
+              pixelSizeYUnit = UNITS.NM;
```
joshmoore: NB: If there's not a way to do `Unit<Length> unit = UNITS.find("nm")` it might be worth it.

dgault: Well you can pass the String unit to FormatTools also (FormatTools.getPhysicalSizeX(value, unit)) and it will do UnitsLength.fromString(unit), which is essentially the same thing. The UNITS class isn't an enum so there isn't an easy way of converting the String without a large if else.

sbesson: Happy to use `FormatTools.getPhysicalSizeX(value, unit)` as in https://github.com/dgault/bioformats/commit/5019659229e5e496b77b67cae53f8a7ed6ce28f3 if is sensible.

melissalinkert: Removing the ```unit.equals("nm")``` checks and just using ```FormatTools.getPhysicalSizeX(pixelSizeX, unit)``` seems to make sense, since that will default to UNITS.MICROM anyway.

dgault:
Updates and changes look good to me

sbesson:
--rebased-to #2094

PR #2068, opened 2015-11-04
OBF: add fallback case for unparseable OME-XML

melissalinkert:
See https://trello.com/c/ZNENhFMs/5-obf-reader-improvements

We don't have any example OBF files with OME-XML metadata, so code/job stability review is probably the best that can be done.  The idea is to allow image data to be read if the metadata is invalid or otherwise cannot be parsed.

### DIFF
--- a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
+++ b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
@@ -60,6 +60,8 @@ import loci.formats.meta.ModuloAnnotation;
 import loci.formats.meta.OriginalMetadataAnnotation;
 import loci.formats.ome.OMEXMLMetadata;
 import loci.formats.ome.OMEXMLMetadataImpl;
+
+import ome.units.quantity.Length;
 import ome.xml.meta.OMEXMLMetadataRoot;
 import ome.xml.model.BinData;
 import ome.xml.model.Channel;
@@ -807,7 +809,6 @@ public class OMEXMLServiceImpl extends AbstractService implements OMEXMLService
         throw new IllegalArgumentException(
             "Expecting OMEXMLMetadata instance.");
       }
-
       dest.setRoot(ome);
     }
     else {
@@ -815,6 +816,25 @@ public class OMEXMLServiceImpl extends AbstractService implements OMEXMLService
       // metadata object and copy it into the destination
       IMetadata src = createOMEXMLMetadata(xml);
       convertMetadata(src, dest);
+
+      // make sure that physical sizes are corrected
+      for (int image=0; image<src.getImageCount(); image++) {
+        Length physicalSizeX = src.getPixelsPhysicalSizeX(image);
+        if (physicalSizeX != null && physicalSizeX.value() != null) {
+          physicalSizeX = FormatTools.getPhysicalSize(physicalSizeX.value().doubleValue(), physicalSizeX.unit().getSymbol());
+          dest.setPixelsPhysicalSizeX(physicalSizeX, image);
+        }
+        Length physicalSizeY = src.getPixelsPhysicalSizeY(image);
+        if (physicalSizeY != null && physicalSizeY.value() != null) {
+          physicalSizeY = FormatTools.getPhysicalSize(physicalSizeY.value().doubleValue(), physicalSizeY.unit().getSymbol());
+          dest.setPixelsPhysicalSizeY(physicalSizeY, image);
+        }
+        Length physicalSizeZ = src.getPixelsPhysicalSizeZ(image);
+        if (physicalSizeZ != null && physicalSizeZ.value() != null) {
+          physicalSizeZ = FormatTools.getPhysicalSize(physicalSizeZ.value().doubleValue(), physicalSizeZ.unit().getSymbol());
+          dest.setPixelsPhysicalSizeZ(physicalSizeZ, image);
+        }
+      }
     }
   }
 
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -203,6 +203,7 @@ public class OBFReader extends FormatReader
         if (service.validateOMEXML(ome_xml)) {
           service.convertMetadata( ome_xml, ome ) ;
           parsedXML = true;
+          MetadataTools.populatePixels(ome, this, false, false);
         }
       }
       catch (DependencyException exception) 
@@ -221,7 +222,6 @@ public class OBFReader extends FormatReader
     if (fileVersion <= 1 || !parsedXML)
     {
       MetadataTools.populatePixels(ome, this);
-
       for (int series = 0; series != core.size(); ++ series)
       {
         CoreMetadata obf = core.get(series);
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -270,6 +270,10 @@ public class OBFReader extends FormatReader
       {
         throw new FormatException( exception ) ;
       }
+      catch (Exception e) {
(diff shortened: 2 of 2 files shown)


sbesson:
Sounds good to me and https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/668/ was green with this PR included. @bjoernthiel: do you have any OBF sample files of version 2 or more that could be shared with us to check for regressions ?

bjoernthiel:
@sbesson I can send a simple test image - please say where to.

melissalinkert:
Files can be uploaded to https://www.openmicroscopy.org/qa2/qa/upload/

bjoernthiel:
@melissalinkert I've uploaded a test image named Measurement1.msr.

melissalinkert:
Updated to handle the uploaded sample; this will reset the units for physical sizes if what is stored in the file's OME-XML would cause a validation error.  https://github.com/openmicroscopy/data_repo_config/issues/42 will need to be updated accordingly.

bjoernthiel:
@melissalinkert @sbesson @ctrueden 
I'm pretty much at a loss: The test image is intended to work, not to fail! 
Your validateOMEXML method is buggy. It won't even validate your minimum specification at 
http://www.openmicroscopy.org/site/support/ome-model/specifications/minimum.html
Maybe we (I myself?) should go for a patch of the startElement method from ValidationSAXHandler 
before this change set gets merged.

bjoernthiel:
@sbesson Is Imspector writing ill formed ome-xml?
If so there is more to fix:
http://www.openmicroscopy.org/site/support/ome-model/specifications/index.html
http://www.openmicroscopy.org/site/support/ome-model/specifications/compliant-hcs.html

bjoernthiel:
@melissalinkert @sbesson @joshmoore 
I have a patch fixing this by letting ValidationSAXHandler find a non default OME namespace as well: 
#2080

melissalinkert:
We're aware of two issues with the sample file.  The first, as you've
rightly pointed out, is that XMLTools' validateXML method does not
correctly validate files that are missing an 'xmlns' attribute on the
root element.  We're working to clarifying the existing examples (see discussion on
https://github.com/openmicroscopy/ome-documentation/pull/1329), and we'll review the PR that you opened this morning (though we likely won't be able to include this in 5.1.6).  I would
not expect that Imspector's behavior will need to change as a result of
this issue.

The second issue is that the PhysicalSizeX and PhysicalSizeY attributes on Pixels
are small enough (3.333333387E-9) to be rounded to 0.0, which is invalid
as the type of both attributes is PositiveFloat.  We're handling this in
other formats by changing the units such that the value is no longer
considered a rounding error (see https://github.com/openmicroscopy/bioformats/pull/2071/files).
Ideally, this is something that would be changed when Imspector writes
OME-XML into an .msr file, but for the moment it is handled by this PR.

I'm not sure that either issue needs to block this PR, as validation
errors should be caught no matter what.  Certainly happy to discuss
further if any of this is unclear.

bjoernthiel:
@melissalinkert @sbesson @joshmoore 
For the first issue I think we should first repair the validation method before we start using it here. 
Merging this would otherwise just stop the reader from importing Imspector's OME-XML meta data. 

The second issue unfortunately I can't see. Loading the test image into ImageJ via the BioFormats plug in works fine. All OME-XML meta data are present and the code you mention (#2071) is not used at all.

melissalinkert:
@bjoernthiel: could you please confirm which commit was built and installed in ImageJ during your tests?  ```Help > About Plugins > Bio-Formats Plugins``` in ImageJ should show the exact commit hash.  Please keep in mind that ImageJ does not perform a validation step on the OME-XML that is displayed; it's only during validation (and thus automated testing) that the small physical sizes become an issue.

Given that there are some open questions here, we'll keep this PR out of next week's 5.1.6 release.  This PR in addition to https://github.com/openmicroscopy/bioformats/pull/2080 will be the target of a 5.1.7 release later in November.

bjoernthiel:
@melissalinkert Please find the 'About Box' attached.
about_bio-formats

bjoernthiel:
The plug in has been built incorporating service.validateOMEXML( ome_xml ) and #2080.

melissalinkert:
Removing on hold label now that 5.1.6 is out.

melissalinkert:
Corresponding XMLTools change rebased to gh-2106 to be reviewed for 5.1.7.

mtbc:
Now #2109 is merged do we need to review this PR for 5.1.7?

sbesson:
Thanks for the reminder @mtbc. At this stage all remaining open `dev_5_1` PRs should be listed and minimally discussed as part of Monday BF standup. I will add them to the agenda.

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/176/console tested the state of this PR together with https://github.com/openmicroscopy/bioformats/pull/2084 against the Imspector sample file uploaded. The tests currently generates 5 failures:
- some of them are due to the physical size values (already reported in https://github.com/openmicroscopy/bioformats/pull/2068#issuecomment-155826949)
- some others are failures thrown while testing the OME-XML. Not sure whether these failures are meant to be caught by `validateOMEXML`.

melissalinkert:
Couple of commits pushed, which together with https://github.com/openmicroscopy/bioformats/pull/2126 should resolve the 5 failures noted above.

sbesson:
With the latest state of this PR, the [automated tests for the Imspector OBF format](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/193/) are passing and reading the OME-XML from the sample file uploaded by @bjoernthiel, I get:

```
$ ./tools/showinf -nopix -omexml-only /ome/data_repo/from_skyking/imspector/qa-16895/Measurement1.msr 
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2015-01" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" UUID="urn:uuid:44ce7b12-284f-4a39-acbd-756b6936d46b" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2015-01 http://www.openmicroscopy.org/Schemas/OME/2015-01/ome.xsd">
   <Image ID="Image:0" Name="SimCam #0 {0}">
      <AcquisitionDate>2015-11-06T15:01:38</AcquisitionDate>
      <Pixels BigEndian="false" DimensionOrder="XYZCT" ID="Pixels:0" Interleaved="false" PhysicalSizeX="3.3333333869999996E-6" PhysicalSizeXUnit="mm" PhysicalSizeY="3.3333333869999996E-6" PhysicalSizeYUnit="mm" PhysicalSizeZ="1.0" PhysicalSizeZUnit="pixel" SignificantBits="16" SizeC="1" SizeT="1" SizeX="128" SizeY="128" SizeZ="1" Type="uint16">
            <LightPath/>
         </Channel>
         <MetadataOnly/>
         <Plane PositionX="-2.133333368E-7" PositionXUnit="m" PositionY="-2.133333368E-7" PositionYUnit="m" TheC="0" TheT="0" TheZ="0"/>
      </Pixels>
   </Image>
...
```

i.e. the OME-XML header is parsed and the physical size value/units are properly converted internally (thanks to https://github.com/openmicroscopy/bioformats/pull/2084).

bjoernthiel:
@melissalinkert @sbesson Can you point me to the lines of code where the rounding to zero used to happen?

sbesson:
@bjoernthiel: the [FormatTools.getPhysicalSizeX(Double, String)](https://github.com/openmicroscopy/bioformats/blob/v5.1.6/components/formats-api/src/loci/formats/FormatTools.java#L1489) method is testing whether the passed value is strictly positive before creating the `Length` object.

sbesson:

melissalinkert:
--rebased-to #2148

PR #2046, opened 2015-10-23
FEITiffReader - Using native units

dgault:
Using metres for physical size rather than micrometres

PR #2071, opened 2015-11-05
Format tools units automatic conversion

dgault:
For some readers we have native units that are not logical for the purpose resulting in very low values.
See https://github.com/openmicroscopy/bioformats/pull/2044 for an example.

This proposed change to FormatsTools getPhysicalSize would implement the following logic as suggested in the PR mentioned above:
1.check if the value is greater than  Constants.EPSILON  
2.if so, use whatever unit was supplied
3.if not, convert to the smallest order of magnitude unit such that (1) passes (or the default unit?)
4.if the value really is 0, or is less than  Constants.EPSILON  for even the smallest defined unit, return null

I have also updated getPhysicalSizeX and Y methods to use the same getPhysicalSize function as we had 2 separate workflows depending on if the units were provided as String or UNITS.MM for example.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1439,7 +1439,17 @@ public final class FormatTools {
     if (unit != null) {
       try {
         UnitsLength ul = UnitsLength.fromString(unit);
-        return UnitsLength.create(value, ul);
+        int ordinal = ul.ordinal();
+        Length returnLength = UnitsLength.create(value, ul);
+
+        // If the requested unit produces a value less than Constants.EPSILON then we switch to the next smallest unit possible
+        // Using UnitsLength.values().length - 2 as a boundary so as not to include Pixel and Reference Frame as convertible units
+        while (returnLength.value().doubleValue() < Constants.EPSILON && ordinal < (UnitsLength.values().length - 2)) { 
+          ordinal++;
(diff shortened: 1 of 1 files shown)


sbesson:
https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-full-repository/669/ and https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-full-repository/670 have been failing with 14 failures and no obvious reasons in the logs since this PR was included. This usually hints at some NPE. Excluding for tomorrow's full repository build.

melissalinkert:
Sorry, that's my fault for not getting https://github.com/openmicroscopy/data_repo_config/pull/40 open sooner.  Removing exclude from this PR.

sbesson:
I also missed the `not configured` line at the top of the repository job. Sorry @dgault for the wrong diagnostics :)

melissalinkert:
All makes sense, and https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-full-repository/673/ looking good.

dgault:
Upon further testing related to https://github.com/openmicroscopy/bioformats/pull/2046 this section of code is never actually being used anywhere and cannot be reached. Any values which are less than Constants.EPSILON are treated as non positive values and a null is returned instantly before this method is called.

dgault:
--no-rebase

PR #2076, opened 2015-11-10
CZI: update line scan detection for airyscan data

melissalinkert:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-October/005701.html

To test, use the .czi file from the thread (uploaded to ```team/community_images/```).  It's easiest to see the problem by using the ```Histo``` tab in Zen, and comparing the histogram table values for each plane against the output of something like:

```
    reader.setId(args[0]);

    for (int t=0; t<100; t++) {
      for (int c=0; c<reader.getSizeC(); c++) {
        byte[] b = reader.openBytes(reader.getIndex(0, c, t));
        int[] counter = new int[256];
        for (int pixel=0; pixel<b.length; pixel++) {
          int v = b[pixel] & 0xff;
          counter[v]++;
        }
        System.out.println("t = " + t + ", c = " + c);
        for (int pixel=0; pixel<counter.length; pixel++) {
          if (counter[pixel] != 0) {
            System.out.println("  * value = " + pixel + ", count = " + counter[pixel]);
          }
        }
      }
    }

    reader.close();
```

or any equivalent in ImageJ or MATLAB.  Without this change, the above code shows that all pixel values are 0 for t > 0.  With this change, the pixel value counts should match what is shown in Zen.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -306,7 +306,7 @@ public class ZeissCZIReader extends FormatReader {
     int outputRow = 0, outputCol = 0;
 
     boolean validScanDim =
-      scanDim == (getImageCount() / getSizeC()) && scanDim > 1;
+      scanDim == (getImageCount() / (getSizeC() * phases)) && scanDim > 1;
     if (planes.size() == getImageCount()) {
       validScanDim = false;
     }


sbesson:
Finally managed to test this PR using the CZI file shared via the email thread. Without this PR (i..e from Bio-Formats 5.1.5), I can reproduce the fact that `r.openBytes` returns 0 values for every timepoint greater than 0. With this PR included, `r.openBytes(r.getIndex(z,c,t))` returns sensible values independently of the (`z,c,t`) input. I did not have a working Zen software available to compare the returned values to the expectations. Barring this, this PR looks ready to merge.

sbesson:
Tested the histogram values for `time 1 phase 1`, `time 2 phase 1`, `time 12436 phase 1` and `time 14365 phase 13` with the help of @jburel and @melissalinkert and the code snippet in the description of this PR. Worked as expected. Merging

melissalinkert:
--rebased-to #2092

PR #2077, opened 2015-11-10
cpp: PixelBuffer: Disable debug bounds checking

rleigh-dundee:
This is extremely expensive for some compilers, such as clang++, so disable unconditionally.  It was always disabled for Release builds; this ensures it's also disabled for Debug builds.

Possibly too late for the freeze, but note it's a trivial one-liner which should shave *several hours* off the build time of the slowest builds.

--------

Testing: check that the build times for Debug builds are reduced for both https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-5.1-merge-superbuild/ and https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-5.1-merge-win-superbuild/

### DIFF
--- a/cpp/lib/ome/bioformats/PixelBuffer.h
+++ b/cpp/lib/ome/bioformats/PixelBuffer.h
@@ -44,6 +44,8 @@
 #include <stdexcept>
 #include <string>
 
+// Disable expensive bounds checking
+#define BOOST_DISABLE_ASSERTS 1
 #include <boost/multi_array.hpp>
 
 #include <ome/bioformats/PixelProperties.h>
--- a/cpp/test/ome-bioformats/pixelbuffer.h
+++ b/cpp/test/ome-bioformats/pixelbuffer.h
@@ -429,7 +429,7 @@ TYPED_TEST_P(PixelBufferType, SetIndex)
 
 TYPED_TEST_P(PixelBufferType, SetIndexDeathTest)
 {
-#ifndef NDEBUG
+#if !defined(NDEBUG) && !defined(BOOST_DISABLE_ASSERTS)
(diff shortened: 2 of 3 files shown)


rleigh-dundee:
--rebased-from #2078

rleigh-dundee:
Needs another commit--this makes the unit test for out of bounds behaviour change; we need to adjust them to check for whether bounds checking is enabled or disabled.

rleigh-dundee:
See https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-5.1-merge-superbuild/BUILD_TYPE=Release,CXXSTD_AUTODETECT=OFF,node=marlin/buildTimeTrend and https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-5.1-merge-superbuild/BUILD_TYPE=Release,CXXSTD_AUTODETECT=ON,node=perch/buildTimeTrend for the dramatic effect this had--it reduces the build time by more than 2/3.

joshmoore:
Impressive! Not for this PR since we really want to get things green & merged, but is the assertion something that someone else would want to be able optionally turn on?

rleigh-dundee:
@joshmoore It's possible but unlikely.  Adding an option for it is doable if required.  If someone did want to, they can comment out the single line in the header; the unit tests will cope.  The main purpose I can see would be if we did any work on the pixel buffer or tiff code, we could comment it out and run the tests once to be sure we aren't failing any bounds checks.  However, I don't see this code lasting that long--I'd like to replace it using the logic from my recent nD calculations PR in the next few months.

joshmoore:
Fair enough. Leaving the more general TODO then in your backlog.

PR #2078, opened 2015-11-10
cpp: PixelBuffer: Disable debug bounds checking

rleigh-dundee:
This is extremely expensive for some compilers, such as clang++, so disable unconditionally.  It was always disabled for Release builds; this ensures it's also disabled for Debug builds.

--------

Testing: check that https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-DEV-merge-superbuild/ and https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-DEV-merge-win-superbuild/ are green; the build time reduction was verified in #2077.

--rebased-to #2077

### DIFF
--- a/cpp/lib/ome/bioformats/PixelBuffer.h
+++ b/cpp/lib/ome/bioformats/PixelBuffer.h
@@ -44,6 +44,8 @@
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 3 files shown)


sbesson:
Builds are green. Merging

PR #2083, opened 2015-11-11
Video spot tracker (rebased onto develop)

sbesson:
This is the same as gh-2040 but rebased onto develop.

----

This PR:

- adds the Video Spot Tracker software to the list of projects using the Jace C++ bindings - see http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-October/005690.html
- removes the BF-CPP abbreviation and uses Jace C++ bindings in the documentation instead to prevent any confusion with Bio-Formats C++. /cc @rleigh-dundee


sbesson:
--rebased-from #2040

dgault:
Updates to documentation look good to me, ready to merge to develop

PR #2084, opened 2015-11-12
Bug fix - Convert values

dgault:
When converting the units the value must also be converted.
Have to use a workaround of using a temporary created Length in order to
get the new Unit<Length> as opposed to the UnitsLength

This a a bug fix for the merged PR https://github.com/openmicroscopy/bioformats/pull/2071

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1436,7 +1436,7 @@ public final class FormatTools {
   }
   
   public static Length getPhysicalSize(Double value, String unit) {
-    if (value != null && value != 0) {
+    if (value != null && value != 0 && value < Double.POSITIVE_INFINITY) {
       if (unit != null) {
         try {
           UnitsLength ul = UnitsLength.fromString(unit);
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1449,7 +1449,7 @@ public final class FormatTools {
           
           // If the requested unit produces a value less than Constants.EPSILON then we switch to the next smallest unit possible
           // Using UnitsLength.values().length - 2 as a boundary so as not to include Pixel and Reference Frame as convertible units
-          while (returnLength.value().doubleValue() < Constants.EPSILON && ordinal < (UnitsLength.values().length - 2)) { 
+          while (returnLength.value().doubleValue() < Constants.EPSILON && ordinal < (UnitsLength.values().length - 3)) { 
             ordinal++;
             ul = UnitsLength.values()[ordinal];
             Length tempLength = UnitsLength.create(0, ul);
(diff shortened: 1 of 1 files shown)


mtbc:
I wonder if we're missing something in the units API that would make this more convenient.

dgault:
Yeah, the use of ordinals to find the next size of unit is also not ideal. Hopefully in 5.2 we can address the units API a bit and clean it up. In the mean time, if the previous PR (https://github.com/openmicroscopy/bioformats/pull/2071) goes in then the bug fix is probably needed. Leaving both out would also be perfectly valid.

rleigh-dundee:
The logic used here looks correct to me.  Is there any testing required, or is this taken care of by the full-repo-test jobs?

The ordinals is one place I am unsure of.  I'm unsure of the ordering guarantees in the generator, and what happens for the non-metric cases.  It might be worth terminating the loop if you hit a non-metric unit?  The reference frame/pixel types are skipped at least, which is good.

I wonder if we need these methods in the Unit enums and handlers at all.  Could these not be static methods on the Quantity classes?  Certainly it would be good to move all these pieces into a common place.

dgault:
Added a new commit which will enable the automatic unit conversion to be used.

The null check is now handled in getPhysicalSize

The checks for > Constants.EPSILON and < Double.POSITIVE_INFINITY are
handled internally within getPhysicalSize, but will attempt to convert
values before assuming null.

sbesson:
@dgault some unit test failures - see https://travis-ci.org/openmicroscopy/bioformats/jobs/93403456

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1436,31 +1436,39 @@ public static Time getTime(Double value, String unit) {
...
+          int ordinal = ul.ordinal();
+          Length returnLength = UnitsLength.create(value, ul);
+  
+          if (returnLength.value().doubleValue() > Constants.EPSILON && returnLength.value().doubleValue() < Double.POSITIVE_INFINITY) {
+            return returnLength;
+          }
+          
+          // If the requested unit produces a value less than Constants.EPSILON then we switch to the next smallest unit possible
+          // Using UnitsLength.values().length - 2 as a boundary so as not to include Pixel and Reference Frame as convertible units
+          while (returnLength.value().doubleValue() < Constants.EPSILON && ordinal < (UnitsLength.values().length - 2)) {
```
sbesson: Several `NullPointerException` seem to be thrown by this line -see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-5.1-merge-openbytes-performance/687/testngreports/loci.tests.testng/OpenBytesPerformanceTest/setId_101/ or ttps://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-full-repository/

dgault:
Updated with a fix for the CI build failures.

From testing it is worth noting that the automatic units conversion will only be effective for metric units as the imperial units are ordered differently (smallest to largest).

dgault:
Lots of config changes can be found here - https://github.com/openmicroscopy/data_repo_config/pull/53

sbesson:

dgault:
--rebased-to #2143

PR #2086, opened 2015-11-13
Add a "Multiple Images" column to the supported formats table (rebased onto develop)

melissalinkert:
This is the same as gh-2038 but rebased onto develop.

----

See https://trello.com/c/9yZdHX52/15-mif-checkbox-in-pff-list. The checkbox seemed to make more sense on the supported formats table, but I can move it to the dataset structure table if anyone feels strongly.


melissalinkert:
--rebased-from #2038

sbesson:
Staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/supported-formats.html. Works as expected.

PR #2087, opened 2015-11-13
Slidebook: fix physical Z size calculation (rebased onto develop)

melissalinkert:
This is the same as gh-2063 but rebased onto develop.

----

See lists.openmicroscopy.org.uk/pipermail/ome-users/2015-October/005703.html.

Tested with ```slidebook/emma/WT BY4742 Nem1 mcherry genome.sld``` on Windows (with appropriate .dlls).  Without this change, the physical Z size was not set at all; with this change, it is set to the absolute difference between two adjacent Z sections.


melissalinkert:
--rebased-from #2063

dgault:
Retested with slidebook files and everything looks good

PR #2069, opened 2015-11-04
Metamorph TIFF: fix parsing of multi-line freeform descriptions

melissalinkert:
See https://trello.com/c/efn0T64w/18-metaxpress-multi-line-descriptions

Based upon analysis of various Metamorph TIFF samples, the user-entered
freeform description appears to be stored contiguously, and always
before the "Exposure" key.  The "Exposure" key seems to consistently be
the first official standard Metamorph key, with "Binning", "Region",
etc. following in a predictable order.  This commit takes all lines from
the XML description field up to "Exposure" and assembles them into a
single "User Description" entry in the metadata table.

Non-freeform lines with no value are also now stored in the metadata table, with the key and value being the same.

This should match the behavior described in http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-March/005172.html, using the file linked from that thread.  This seems to be safer and less invasive than trying to maintain a whitelist of official vs. freeform keys, but feedback would be appreciated.

/cc @emmenlau

PR #2088, opened 2015-11-13
Metamorph TIFF: fix parsing of multi-line freeform descriptions (rebased onto develop)

melissalinkert:
This is the same as gh-2069 but rebased onto develop.

----

See https://trello.com/c/efn0T64w/18-metaxpress-multi-line-descriptions

Based upon analysis of various Metamorph TIFF samples, the user-entered
freeform description appears to be stored contiguously, and always
before the "Exposure" key.  The "Exposure" key seems to consistently be
the first official standard Metamorph key, with "Binning", "Region",
etc. following in a predictable order.  This commit takes all lines from
the XML description field up to "Exposure" and assembles them into a
single "User Description" entry in the metadata table.

Non-freeform lines with no value are also now stored in the metadata table, with the key and value being the same.

This should match the behavior described in http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-March/005172.html, using the file linked from that thread.  This seems to be safer and less invasive than trying to maintain a whitelist of official vs. freeform keys, but feedback would be appreciated.

/cc @emmenlau

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
@@ -134,6 +134,8 @@ public class MetamorphHandler extends BaseHandler {
         if (metadata != null) metadata.remove("Comment");
 
         String k = null, v = null;
+        boolean freeform = true;
+        String freeformDescription = "";
 
         if (value.indexOf(delim) != -1) {
           int currentIndex = -delim.length();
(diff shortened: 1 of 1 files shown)


melissalinkert:
--rebased-from #2069

sbesson:
```
sbesson@ls30630:bftools $ ./showinf -version
Version: 5.1.5-DEV
VCS revision: 99f4f3a45a986c66fb0c5cd0d90cd9c14031bb36
Build date: 16 November 2015
sbesson@ls30630:bftools $ ./showinf -nopix -omexml-only bBZIX-021_wD13_s3_z0_t1_cCy5_u001.tif
...
      <XMLAnnotation ID="Annotation:85" Namespace="openmicroscopy.org/OriginalMetadata">
         <Value>
            <OriginalMetadata>
               <Key>User Description</Key>
               <Value>Experiment base name:AC20-2-1--TetR-GFPforEntryAssay
Experiment set:AC20-2-1--TetR-GFPforEntryAssay
siRNA transfection of HeLa cells 
infection with ACBr165 (TetR-GFP and const. dsRED
Entry 4h
Induction of TetR-GFP for 4 h together with Gentamycin killing of extracellular bacteria
extracellular: red
intracellular: red and green
DAPI staining (1:1000)
Dy-647-phalloidin (1:100)</Value>
            </OriginalMetadata>
         </Value>
      </XMLAnnotation>
...
```

and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/12/ is green. Merging.

PR #2073, opened 2015-11-09
Fake: warn on invalid physical sizes

melissalinkert:
This warns instead of throwing an exception, and ensures that physical
sizes of 0.0 are not allowed.

To test, compare the behavior of ```showinf -nopix -omexml fake/samples/emptySizes&physicalSizeX=0.0.fake```.  Without this change, XML validation should fail; with this change, PhysicalSizeX should not be populated at all.

PR #2090, opened 2015-11-13
Fake: warn on invalid physical sizes (rebased onto develop)

melissalinkert:
This is the same as gh-2073 but rebased onto develop.

----

This warns instead of throwing an exception, and ensures that physical
sizes of 0.0 are not allowed.

To test, compare the behavior of ```showinf -nopix -omexml fake/samples/emptySizes&physicalSizeX=0.0.fake```.  Without this change, XML validation should fail; with this change, PhysicalSizeX should not be populated at all.


melissalinkert:
--rebased-from #2073

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/12/ is green and the command above worked as expected. Merging.

PR #2074, opened 2015-11-09
Metamorph TIFF: add dual camera support

melissalinkert:
See QA 16887.  Dual camera acquisitions store two channels side-by-side;
this splits the image and records two separate channels.  The same
feature is already present in MetamorphReader.

To test, verify that the file from QA 16887 is 512x512, 2 channels, 200 timepoints, and that the channels are distinct.  Builds should remain green.

PR #2091, opened 2015-11-13
Metamorph TIFF: add dual camera support (rebased onto develop)

melissalinkert:
This is the same as gh-2074 but rebased onto develop.

----

See QA 16887.  Dual camera acquisitions store two channels side-by-side;
this splits the image and records two separate channels.  The same
feature is already present in MetamorphReader.

To test, verify that the file from QA 16887 is 512x512, 2 channels, 200 timepoints, and that the channels are distinct.  Builds should remain green.


melissalinkert:
--rebased-from #2074

sbesson:
Retested on the QA file lined in the description. Dimensionality is correct and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/12/ is green. MErging.

PR #2093, opened 2015-11-13
IMODReader: add support for native length units (rebased onto develop)

sbesson:
This is the same as gh-2039 but rebased onto develop.

----

This PR adjusts the IMODReader to store the physical pixel size read from `imod` files using the native units. The specification for the units can be found at http://bio3d.colorado.edu/imod/doc/binspec.html.

Since one of the samples IMOD file uses the pixel unit (`0`) which that cannot be converted to micrometers, the format reader automated tests are also adjusted to handle this case (using `isConvertible` similarly to what is done in https://github.com/openmicroscopy/bioformats/pull/2030). The size comparison logic used across all physical size tests is also refactored into a single helper method to simplify future maintenance.

For automated tests, this means we do not differentiate the case where the physical size is 1 micron or 1 pixel. In the future we may want to introduce a `physicalSize{X,Y,Z}Unit` optional configuration to store non-convertible length unit.

To test this PR, check `imod` files are read correctly and automated tests are passing.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/IMODReader.java
+++ b/components/formats-gpl/src/loci/formats/in/IMODReader.java
@@ -457,7 +457,7 @@ public class IMODReader extends FormatReader {
         }
       }
       if (physicalZ > 0) {
-        Length z = FormatTools.getPhysicalSizeX(physicalZ, physicalSizeUnit);
+        Length z = FormatTools.getPhysicalSizeZ(physicalZ, physicalSizeUnit);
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #2039

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/13

dgault:
Changes look good to me and builds are green

PR #2094, opened 2015-11-13
Use native units for the pixels physical size of UnisokuReader (rebased onto develop)

sbesson:
This is the same as gh-2065 but rebased onto develop.

----


sbesson:
--rebased-from #2065

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/13

dgault:
Changes look good and jobs are green, good to merge to develop

PR #2097, opened 2015-11-16
Add a "Pyramid" column to the supported formats table

melissalinkert:
See https://trello.com/c/WL8qzFDe/44-list-has-pyramids-on-the-formats-list

To review, check the staged supported formats table page.  Verify that there is a ```Pyramid``` column on the far right, and verify that none of the values are obviously incorrect.

This does also make the column widths for ```Multiple Images``` and ```Pyramid``` a little larger, to hopefully prevent rendering issues noted in the last formats table PR.  Happy to revert those changes if   they don't help.


mtbc:
Should this be listed for review?

sbesson:
Rendered at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/supported-formats.html. Looks and renders good to me. Thoughts @hflynn @joshmoore?

hflynn:
I'm increasingly not a fan of the way we render the middle items with the image headers* but as far as this PR goes, yes, it looks fine. 

*Although I'm not sure what to do about it - being able to collapse and expand that section might be useful for people on small screens maybe.

joshmoore:
Would this be a possible place to attempt SVG? See http://tutorials.jenkov.com/svg/text-element.html#vertical-text

hflynn:
That might be an idea yes, although we'd have to check what the browser support is like now too (although given I have Firefox 42 and that says version 22 didn't support it at time of writing, I guess that may be a rather old problem!). I will try to find time to look at that but I don't think it blocks merging this PR.

hflynn:
https://trello.com/c/4rodQhTp/388-try-vertical-text-for-bf-supported-formats-table for reference.

PR #2098, opened 2015-11-16
Showinf: preload format improvements

sbesson:
This PR has been used to test https://github.com/openmicroscopy/bioformats/pull/2089 using the command-line tools. It allows to call `FormatReader.isThisType(RandomAccessInputStream)` using `showinf`.

To test this PR, run

```
./showinf -preload -format FORMAT /path/to/file
```

and verify the format type is also checked before caching the bytes.


melissalinkert:
All looks good.

PR #2100, opened 2015-11-17
EM format type checking fixes

melissalinkert:
Resolves https://trac.openmicroscopy.org/ome/ticket/13058 and https://trac.openmicroscopy.org/ome/ticket/13059.

To test, use the files from QA 16850, ignoring ```cvim006036.0.img``` as it is raw data without the corresponding header file.  Without this change, ```showinf``` on the three remaining files should report ```UnknownFormatException```.  With this change, the same test should show ```15_movie_gc.mrcs``` being detected as MRC, and both .dat files being detected as Spider.  EMAN2 should be able to read all three (per http://blake.bcm.edu/emanwiki/EMAN2ImageFormats), so hopefully can be used to validate the images themselves.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MRCReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MRCReader.java
@@ -56,7 +56,7 @@ public class MRCReader extends FormatReader {
   // there, according to: http://bio3d.colorado.edu/imod/doc/mrc_format.txt
 
   private static final String[] MRC_SUFFIXES =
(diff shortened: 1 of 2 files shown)


mtbc:
Should this be listed for review?

dgault:
Tested and files are read and displayed as described above. Actual image data appears to show some fixed pattern noise around the edges, but the image is displayed exactly the same when opened with EMAN2, so this is likely to be expected and not caused by the reader.

PR #2085, opened 2015-11-13
Add version history for 5.1.6

melissalinkert:
A couple of yet-unmerged PRs are referenced here, so depending on how Friday goes this may need one more commit.

PR #2101, opened 2015-11-17
Add version history for 5.1.6 (rebased onto develop)

melissalinkert:
This is the same as gh-2085 but rebased onto develop.

----

A couple of yet-unmerged PRs are referenced here, so depending on how Friday goes this may need one more commit.


melissalinkert:
--rebased-from #2085

sbesson:
https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/about/whats-new.html :+1:

PR #2080, opened 2015-11-11
find non default namespace ome

bjoernthiel:
please review

PR #2106, opened 2015-11-18
find non default namespace ome (rebased onto dev_5_1)

melissalinkert:
This is the same as gh-2080 but rebased onto dev_5_1.

----

please review


melissalinkert:
--rebased-from #2080

melissalinkert:
Replaced by gh-2109 (due to branch naming conflicts).

PR #2111, opened 2015-11-21
Fix current year in POM file

melissalinkert:
To test, build bioformats_package.jar using Maven and copy to your ImageJ plugins folder.  Without this change, ```Help > About Plugins > Bio-Formats Plugins``` in ImageJ should show the end copyright year as 2013.  With this change, the same test should show the end copyright year as 2015.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,7 @@
     <release.version>5.1.7-SNAPSHOT</release.version>
     <vcs.shortrevision>${shortrevision}</vcs.shortrevision>
     <date>${maven.build.timestamp}</date>
-    <year>2013</year>
+    <year>2015</year>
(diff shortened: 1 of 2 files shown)


sbesson:
Worth adding to https://github.com/openmicroscopy/bioformats/blob/develop/tools/update_copyright (and maybe renaming this file) so that we don't forget in 2016?

melissalinkert:
Updated update_copyright to include the POM change.  I don't think we need to rename that script though, as the POM year is being used to set copyright end dates.

sbesson:
Tested with bioformats_package-5.1.7-20151124.023449-8.jar on ImageJ as described as well as `./tools/update_copyright 2017`. All worked as expected. Merging.

melissalinkert:
--rebased-to #2128

PR #2113, opened 2015-11-23
Making Shape and LightSource complex type

dgault:
What this means:
- Shape and LightSource become abstract classes and are no longer elements
- There is now a separate index for each Shape and LightSource type (as for other abstract implementations eg Annotations)

Changes to xsd:
- Shape and LightSource become complexTypes
- Remove abstractproprietary in favour of abstract

Changes to CodeGen:
- Remove abstractproprietary in favour of only abstract
- Enable code gen for mask bindata 

Changes to templates:
- Remove abstractproprietary in favour of only abstract
- Add in AnnotationCount methods for each of the Shape and LightSource types (same as annotations are currently handled)
- Remove hardcoded PixelsBinData and getType methods for Shape and LightSource

Other changes:
- Clean-up existing unit tests etc to use the new indexing for Shape and LightSource

To Do:
- Clean up all abstract types to use single indexing for all (removing the need for much of the hardcoded template functions)


sbesson:
Closing this since the regions PR is now integrated, can you rebase against `develop`?

rleigh-dundee:
Great stuff for removing AbstractProprietary!  I don't think it goes quite far enough though--there are two fairly major things I can see which need further work:

- Your patch implements separate methods for each abstract type; it would be nicer if these worked with the generic types.  Example, `instrument.addLightSource(laser);` becoming `instrument.addLaser(laser);`.  I know this is how `Abstract` is currently implemented, but the existing implementation in the generator is currently very suboptimal, and needs changing to generate a single list of the abstract type, rather than a separate list for each concrete type.  This will mean most of the user-facing API can remain as it is, which will make the needed changes vastly smaller.

- There are a lot of hardcoded `getFooCount(idx ...)` methods being added to the templates.  These are all bugs in the generator.  Ideally, we update the template with a *single* instance which is expanded for each `Count` method needed, i.e. making the template truly generic.  All the existing `Count` methods are equally buggy and should be also removed once they are also getting generated.

PR #2102, opened 2015-11-18
Big decimal exception

dgault:
Adding a limit to the decimal precision of BigDecimal divide.

This is required to prevent the following exception: Non-terminating
decimal expansion; no exact representable decimal result.

Issue was reported below:
https://www.openmicroscopy.org/community/viewtopic.php?f=4&p=16439#p16439

Note: This behaviour is the same as we already use in the Fluoview reader. There are no other instances of BigDecimal division that require correcting.

PR #2116, opened 2015-11-24
Using MathContext with Big Decimal (rebased onto develop)

dgault:
This is the same as https://github.com/openmicroscopy/bioformats/pull/2102 but rebased onto develop.

Adding a limit to the decimal precision of BigDecimal divide.
This is required to prevent the following exception: Non-terminating
decimal expansion; no exact representable decimal result.


dgault:
--rebased-from https://github.com/openmicroscopy/bioformats/pull/2102

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/19/console is green and `showinf` is passing locally against the sample file. Merging.

PR #1997, opened 2015-09-30
Physical Z Size for Andor Tiff

dgault:
User reported bug:

http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-September/005658.html

Using the stamps data (position data for dimensions other than X and Y)
to retrieve a list of physical z values. By identifying the unique
values and ensuring we have a correct number of values, we can average
the correct physical z size.

PR #2117, opened 2015-11-25
Andor - Plane Position Z

dgault:
Reported Issue: http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-November/005745.html

This is a follow up to : https://github.com/openmicroscopy/bioformats/pull/1997

To reproduce: 
- Using uploaded from file QA 16928
- Import with bioformats and view metadata
- Verify that the PhysicalSizeZ and TheZ are set correctly and PositionZ is 0 for each image

To test the PR:
- Follow the same steps above but verify that PositionZ is now set for each image, PositionZUnit should also now be µm rather than reference frame

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
@@ -326,10 +326,12 @@ public class FluoviewReader extends BaseTiffReader {
         ArrayList<Double> uniqueZ = new ArrayList<Double>();
         if (i > 1) {
           zPositions = stamps[i - 2];
-          for (Double z : zPositions) {
-            BigDecimal bd = new BigDecimal(z);
-            bd = bd.setScale(10, RoundingMode.HALF_UP);
-            if (!uniqueZ.contains(bd.doubleValue())) uniqueZ.add(bd.doubleValue());
+          if (zPositions != null) {
+            for (Double z : zPositions) {
+              BigDecimal bd = new BigDecimal(z);
+              bd = bd.setScale(10, RoundingMode.HALF_UP);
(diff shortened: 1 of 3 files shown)


sbesson:
Tested by generating the OME-XML with and without this PR on the sample file

```
$ showinf -nopix -omexml-only 16928/2Ch_Z10_T10.tif > /tmp/2Ch_Z10_T10.ome
$ /opt/ome/bioformats/tools/showinf -omexml-only 16928/2Ch_Z10_T10.tif  > /tmp/2Ch_Z10_T10_2117.ome
```

and computing the diff between headers. As described in the PR, the only difference is the `PositionZ` which is now properly filled for all the planes. Do you have the bandwidth to add a matching configuration file to test the QA file and prevent future regressions?

dgault:
I have generated a matching config for the QA file, however it has thrown up a small bug in the test-suite for zPosition units. I have a fix for the test suite bug, but should it be opened in a separate PR?

sbesson:
If the test-suite changes have been tested and fixed in the context of this PR, feel free to add the commit here.

dgault:
Matching config PR for new sample file can be found - https://github.com/openmicroscopy/data_repo_config/pull/51

dgault:
Added a final commit to this PR to fix the units for Y and Z Positions in test-suite configuration

dgault:
Updated the test suite further as reference frame was hardcoded as the expected unit type for plane positions.

https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/191/

sbesson:

dgault:
--rebased-to https://github.com/openmicroscopy/bioformats/pull/2136

PR #2124, opened 2015-11-30
bfCheckJavaPath.m: Only check MATLAB static javaclasspath once

mkitti:
In bfCheckJavaPath.m, only check MATLAB static javaclasspath once by using a persistent variable. 

The static javaclasspath can only be changed at startup by adding a javaclasspath.txt file in either the MATLAB startup directory or in `prefdir`. Therefore, we only need to check the static classpath once for the jars. On subsequent calls to `bfCheckJavaPath.m`, just recheck the dynamic path and merge the results.

As the static classpath is quite long and the dynamic classpath is quite short in MATLAB the efficiency gains by not constantly rechecking the static classpath are substantial.

This change is against 5.1.x only and thus is against dev_5_1. It would need be reconciled with the version in develop ( 549b8673a619dce95e482234fad54367a8975b97 ) which does no analysis of the classpath directly. However, unlike the version in develop, it should retain the previous behavior and should not be breaking as in #2027. See also #2018.

### DIFF
--- a/components/formats-gpl/matlab/bfCheckJavaPath.m
+++ b/components/formats-gpl/matlab/bfCheckJavaPath.m
@@ -42,6 +42,8 @@ function [status, version] = bfCheckJavaPath(varargin)
 % with this program; if not, write to the Free Software Foundation, Inc.,
 % 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+persistent hasBFJarStatic;
+
 % Input check
 ip = inputParser;
 ip.addOptional('autoloadBioFormats', true, @isscalar);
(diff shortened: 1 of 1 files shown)


sbesson:
@mkitti: looking quickly, the code changes look sensible and https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-matlab/ is green with this PR included. On `develop` https://github.com/openmicroscopy/bioformats/pull/2027 implements a different (breaking) strategy  to improve the performance of this function and I do not think there is any need to reconcile both. @bramalingam did you have the time to review this PR today or should we relist for tomorrow?

bramalingam:
Couldn't get to this today, please re-list for tomorrow.

bramalingam:
Tested using the following code snippet,

```
for i = 1:100; tic ; status = bfCheckJavaPath(1); t(i)=toc; end; plot(t)
```

And obtained the following plot:
100_iterations_bfcheckjavapath

The first 2 iterations seem to be taking longer than the rest of the iterations, shouldnt it be just the first iteration that should be taking longer than the rest of the iterations?

bramalingam:
Sorry, The plot pasted above was for BF 5.1.6, and here's the plot with the changes

100_iterations_bfcheckjavapath_current_change

Definitely a performance enhancement seen. Good to merge.

sbesson:
Thanks @bramalingam, as a quick follow-up, could you check that the HEAD of `develop` of Bio-Formats gives similar metrics for the iterated `bfCheckJavaPath` call but also resolves the overhead in the first 2 iterations due to the removal of the `javaclasspath` call?

sbesson:
--no-rebase

mkitti:
@sbesson Here's what I see

testbfcheckjavapath

The mean time for the second trial on are as follows:

5.56e-02 (5.1.6)
9.77e-04 (dev_5_1)
5.82e-04 (develop)

Develop is faster by less than 2X than dev_5_1. Overall, the 2 orders of magnitude improvement is fantastic. I think it might possible to optimize both versions further given that this is run on every `bfOpen` call.

Areas for Improvement

* Get rid of inputParser (input is simple)
* Only process output if requested (no need to cast version to char)
* Only check for Octave / matlab once, use try / catch or persistent, or use a common approach
* Inline the regular expression (dev_5_1)
* Inline the any(cellfun) call, break once the first one has been found (dev_5_1)
* Consider a parameter about whether to cache the status using persistent rather than rechecking every single time

I would be happy to pursue this at a later time if no one else gets to it first.

sbesson:
Thanks for the detailed report and your contribution @mkitti . At this point, it seems we are in a good place in terms of performance on both 5.1.x and 5.2.x lines.

PR #2126, opened 2015-11-30
Update automated tests to read configured plane position units

melissalinkert:
This should prevent silent failures of the plane position test, as we're now using reference frame units only if an X/Y/Z position unit is not present in the configuration file.

In combination with today's commits to https://github.com/openmicroscopy/bioformats/pull/2068, this should help the repository jobs to become green.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -340,6 +340,18 @@ public class Configuration {
     return pos == null ? null : new Double(pos);
   }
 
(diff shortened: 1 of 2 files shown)


melissalinkert:
Excluding for now, to avoid conflicts with https://github.com/openmicroscopy/bioformats/pull/2117.

sbesson:
I suppose this is no longer necessary now the jobs are passing ?

melissalinkert:
No, gh-2117 covers what this was trying to accomplish.

PR #2114, opened 2015-11-23
MINC2 - Endianess 

dgault:
For short, int, float and double data in MINC2 format, the endianess of
the pixel format must be set before unpacking the data.

Issue was reported here: https://github.com/openmicroscopy/bioformats/issues/2110

To reproduce:
- Using a sample 16bit minc2 image open with bioformats
- Verify that the outline of the data is visible but the pixel values appear to noise

To verify:
- Reopen the image with the PR
- Verify that the data is no longer appearing as noise and instead as sensible values

Issue #2110, opened 2015-11-20
Invalid pixel values for some MINC 2.0 files?


ctrueden:
See [this thread on the ImageJ list](http://imagej.1557.x6.nabble.com/mnc-files-tt5014995.html).

As linked from the thread, sample data illustrating the issue can be [found here](http://www.mouseimaging.ca/mnc/mouse_embryo_atlas/final.model.mnc).

dgault:
I was able to reproduce the issue. It appears that the data was being unpacking incorrectly thus leading to noise like values. I have opened a PR to correct the order in which the endianess of the data is set and the pixels unpacked.

https://github.com/openmicroscopy/bioformats/pull/2114

melissalinkert:
Closing in favor of https://github.com/openmicroscopy/bioformats/pull/2114.

PR #2127, opened 2015-12-01
Rebased: MINC2 - Endianess

dgault:
--rebased-from #2114

For short, int, float and double data in MINC2 format, the endianess of
 the pixel format must be set before unpacking the data.

Issue was reported here: #2110

To reproduce:
•Using a sample 16bit minc2 image open with bioformats
•Verify that the outline of the data is visible but the pixel values appear to noise

To verify:
•Reopen the image with the PR
•Verify that the data is no longer appearing as noise and instead as sensible values


sbesson:
Tested with the daily merge 5.2.0-SNAPSHOT

screen shot 2015-12-02 at 11 07 19

Merging.

sbesson:
Also https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/22/console is green.

PR #2128, opened 2015-12-01
Fix current year in POM file (rebased onto develop)

melissalinkert:
This is the same as gh-2111 but rebased onto develop.

----

To test, build bioformats_package.jar using Maven and copy to your ImageJ plugins folder.  Without this change, ```Help > About Plugins > Bio-Formats Plugins``` in ImageJ should show the end copyright year as 2013.  With this change, the same test should show the end copyright year as 2015.


melissalinkert:
--rebased-from #2111

sbesson:
screen shot 2015-12-02 at 09 07 41

:+1:

PR #2129, opened 2015-12-01
Gatan: explicitly set the locale used for parsing physical sizes

melissalinkert:
See https://trello.com/c/dx5t3btt/14-gatan-locale-issues

There is already a check for the "," separator not being present, so
using Locale.ENGLISH for parsing should be safe across the board.
This is thankfully the only place where we're using NumberFormat and Collator.

To test, pick any Gatan .dm3 file.  Verify that without this change, ```export BF_FLAGS="" && showinf -nopix -omexml``` shows valid values for PhysicalSizeX and PhysicalSizeY, and that ```export BF_FLAGS="-Duser.language=fr -Duser.country=FR" && showinf -nopix -omexml``` does not.  With this change, both tests should show the same valid values for PhysicalSizeX and PhysicalSizeY (and jobs running with fr_FR should pass).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -31,6 +31,7 @@ import java.text.NumberFormat;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 import loci.common.Constants;
 import loci.common.RandomAccessInputStream;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Build running now: https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-5.1-merge-repository-subset/196/console

melissalinkert:
...and it's now green.

melissalinkert:
--rebased-to #2149

PR #2133, opened 2015-12-03
Whitelist .ome.xml as an official extension of the OME-XML format

sbesson:
This extension is currently supported by the permissive nature of the reader but unsupported by the writer. There are a couple of reasons why we want to consider it as official including our own usage, the symmetry with the OME-TIFF format extension and the usability with other software.

To test this PR, try to convert an OME-XML into another one with the `.ome.xml` extension

```
./tools/bfconvert test.ome.xml test2.ome.xml
```

/cc @rleigh-dundee @mtbc @hflynn @joshmoore @melissalinkert

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
@@ -85,7 +85,7 @@ public class OMEXMLReader extends FormatReader {
 
   /** Constructs a new OME-XML reader. */
   public OMEXMLReader() {
-    super("OME-XML", "ome");
+    super("OME-XML", new String[] {"ome", "ome.xml"});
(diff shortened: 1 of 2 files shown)


rleigh-dundee:
Tested with:

```
ome.tiff to ome.xml
ome.xml to ome.tiff
ome.xml to ome.xml
```

In all cases the metadata and pixel data are fine.  Looks good to merge.

However, I did find a defect (unrelated to this PR).  If I convert ome.tiff to ome.xml, the TiffData elements are preserved while BinData elements are added.  This means it's technically invalid.  The OMEXML writer should be stripping these out.

sbesson:
As discussed at today's BF standup, closing this and will reopen on `develop` which becomes our primary development branch.

PR #2136, opened 2015-12-04
Andor - Plane Position Z (rebased onto develop)

dgault:
This is the same as https://github.com/openmicroscopy/bioformats/pull/2117 but rebased onto develop.

---

Reported Issue: http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-November/005745.html

This is a follow up to : #1997

To reproduce:

Using uploaded from file QA 16928
Import with bioformats and view metadata
Verify that the PhysicalSizeZ and TheZ are set correctly and PositionZ is 0 for each image
To test the PR:

Follow the same steps above but verify that PositionZ is now set for each image, PositionZUnit should also now be µm rather than reference frame

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
@@ -326,10 +326,12 @@ public class FluoviewReader extends BaseTiffReader {
         ArrayList<Double> uniqueZ = new ArrayList<Double>();
         if (i > 1) {
           zPositions = stamps[i - 2];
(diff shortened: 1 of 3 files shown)


dgault:
--rebased-from https://github.com/openmicroscopy/bioformats/pull/2117

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/31/console. Four remaining failures are Gatan sample files which will need another rebased PR /cc @melissalinkert

PR #2138, opened 2015-12-04
Format tools units automatic conversion (rebased onto develop)

dgault:
This is the same as #2071 but rebased onto develop.

---

For some readers we have native units that are not logical for the purpose resulting in very low values.
See #2044 for an example.

This proposed change to FormatsTools getPhysicalSize would implement the following logic as suggested in the PR mentioned above:
1.check if the value is greater than Constants.EPSILON

2.if so, use whatever unit was supplied
3.if not, convert to the smallest order of magnitude unit such that (1) passes (or the default unit?)
4.if the value really is 0, or is less than Constants.EPSILON for even the smallest defined unit, return null

I have also updated getPhysicalSizeX and Y methods to use the same getPhysicalSize function as we had 2 separate workflows depending on if the units were provided as String or UNITS.MM for example.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1439,7 +1439,17 @@ public final class FormatTools {
     if (unit != null) {
       try {
         UnitsLength ul = UnitsLength.fromString(unit);
-        return UnitsLength.create(value, ul);
+        int ordinal = ul.ordinal();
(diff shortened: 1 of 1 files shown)


dgault:
--rebased-from #2071

sbesson:
Discussed briefly with @melissalinkert. As described in the original PR, this turned out to be a no-op and having this PR integrated is required to rebase https://github.com/openmicroscopy/bioformats/pull/2084. Merging as it is.

PR #2140, opened 2015-12-04
FEITiffReader - Using native units (rebased onto develop)

dgault:
This is the same as #2046 but rebased onto develop.

---

Using metres for physical size rather than micrometres


dgault:
--rebased-from #2046

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/29/console

PR #2141, opened 2015-12-04
AliconaReader - Using native units (rebased onto develop)

dgault:
This is the same as #2044 but rebased onto develop.

---

Using metres rather then micrometers


dgault:
--rebased-from #2044

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/29/console

PR #2143, opened 2015-12-04
Bug fix - Convert values (rebased onto develop)

dgault:
This is the same as #2084 but rebased onto develop.

---

When converting the units the value must also be converted.
Have to use a workaround of using a temporary created Length in order to
get the new Unit as opposed to the UnitsLength

This a a bug fix for the merged PR #2071


dgault:
--rebased-from #2084

sbesson:
Given https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/29, merging this as it is a prerequisite for other rebases and will get refactored.

PR #2147, opened 2015-12-07
ScanR: read core metadata from first non-null TIFF file

melissalinkert:
This fixes the case where the files for the first recorded well were removed.

See https://trello.com/c/a3M5iBxQ/156-secretionb-failure


joshmoore:
Files which were previously blocked are certainly going in now. :thumbsup:

joshmoore:
Data in!

PR #2148, opened 2015-12-08
OBF: add fallback case for unparseable OME-XML (rebased onto develop)

melissalinkert:
This is the same as gh-2068 but rebased onto develop.

----

See https://trello.com/c/ZNENhFMs/5-obf-reader-improvements

We don't have any example OBF files with OME-XML metadata, so code/job stability review is probably the best that can be done.  The idea is to allow image data to be read if the metadata is invalid or otherwise cannot be parsed.

### DIFF
--- a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
+++ b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
@@ -60,6 +60,8 @@ import loci.formats.meta.ModuloAnnotation;
 import loci.formats.meta.OriginalMetadataAnnotation;
 import loci.formats.ome.OMEXMLMetadata;
 import loci.formats.ome.OMEXMLMetadataImpl;
+
+import ome.units.quantity.Length;
 import ome.xml.meta.OMEXMLMetadataRoot;
 import ome.xml.model.BinData;
 import ome.xml.model.Channel;
@@ -807,7 +809,6 @@ public class OMEXMLServiceImpl extends AbstractService implements OMEXMLService
         throw new IllegalArgumentException(
             "Expecting OMEXMLMetadata instance.");
       }
-
       dest.setRoot(ome);
     }
     else {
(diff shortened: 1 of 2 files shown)


melissalinkert:
--rebased-from #2068

bjoernthiel:
@melissalinkert @sbesson 
Why do we have [value - Constants.EPSILON > 0](https://github.com/openmicroscopy/bioformats/blob/v5.1.6/components/formats-api/src/loci/formats/FormatTools.java#L1392) instead of value > 0 here? 
It prevents us from using SI base units or arbitrary units. 
I mean if one or another reader might introduce rounding errors it should handle them individually. 
Or is anything else (e.g. ImageJ) forcing us into these constrains?

sbesson:
https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/34 being green, merging this PR to have all the 5.1.x effort reintegrated into the `develop` line.

melissalinkert:
@bjoernthiel: it's not an issue with the reader introducing rounding errors, it's an issue with the file actually containing very, very small positive values of a physically large unit.  The OME-XML schema requires PhysicalSizeX, PhysicalSizeY, and PhysicalSizeZ to be positive, so values that are small enough that they could be considered rounding errors of 0.0 can cause errors during validation or re-parsing of the generated OME-XML.  ```Constants.EPSILON``` is the minimum safe threshold that we use for avoiding these errors.

We know this is a little limiting, but precise floating point arithmetic is notoriously difficult, particularly when trying to prevent rounding to 0.0 across the entire stack of reading a value from the original file, processing it in Bio-Formats, storing in XML, and reading the value back from XML using Bio-Formats or some other software (which may not be written in Java, and thus may have different rounding conventions).

bjoernthiel:
@melissalinkert @ctrueden I see. But I think we should better switch off the rounding (that usually happens on writing floating point numbers using %f instead of %g). 
Please point me to the lines of code where this "considered rounding errors of 0.0 can cause errors during validation or re-parsing of the generated OME-XML" happens. I'd love to clean that up.

PR #2135, opened 2015-12-03
Add version history for 5.1.7

melissalinkert:
Just what it says on the tin.  I didn't add C++ changes since there don't seem to be any dev_5_1 C++ PRs not tagged for 5.1.6 - happy to add if I missed anything.

PR #2150, opened 2015-12-08
Add version history for 5.1.7 (rebased onto develop)

melissalinkert:
This is the same as gh-2135 but rebased onto develop.

----

Just what it says on the tin.  I didn't add C++ changes since there don't seem to be any dev_5_1 C++ PRs not tagged for 5.1.6 - happy to add if I missed anything.

### DIFF
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -1,6 +1,28 @@
 Version history
 ===============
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
--rebased-from #2135

sbesson:
Staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/about/whats-new.html

PR #2154, opened 2015-12-10
Fix map namespace post schema unification

sbesson:
This should fix the failing `testValidXML` tests from https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/4/console.

### DIFF
--- a/components/ome-xml/src/ome/xml/model/MapPairs.java
+++ b/components/ome-xml/src/ome/xml/model/MapPairs.java
@@ -91,7 +91,7 @@ public class MapPairs implements OMEModelObject {
 
     // -- Constants --
 
-    public static final String MAPNAMESPACE = "http://www.openmicroscopy.org/Schemas/SA/" + VERSION;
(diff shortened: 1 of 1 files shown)


components/ome-xml/src/ome/xml/model/MapPairs.java
```diff
@@ -91,7 +91,7 @@ static Properties loadProperties() {
 
     // -- Constants --
 
-    public static final String MAPNAMESPACE = "http://www.openmicroscopy.org/Schemas/SA/" + VERSION;
+    public static final String MAPNAMESPACE = "http://www.openmicroscopy.org/Schemas/OME/" + VERSION;
     public static final String PAIRSNAMESPACE = "http://www.openmicroscopy.org/Schemas/OME/" + VERSION;
```
melissalinkert: Since these are the same value now, would it make sense to remove one of them?

PR #2156, opened 2015-12-14
Update README.md to current status

simleo:
Java 1.7+ code can now be added to the repo since Java 6 support has been dropped. Also removing the reference to "test_per_commit" as suggested by @sbesson


sbesson:
:+1:

PR #2158, opened 2015-12-15
idr: change group

joshmoore:
Change the group that is in use on the `metadata` branch so that a proper release without conflicting version numbers can be performed.

cc: @sbesson

### DIFF
--- a/ant/version.xml
+++ b/ant/version.xml
@@ -9,7 +9,7 @@ Ant build file for release version numbering.
   <import file="${root.dir}/ant/gitversion.xml" optional="true"/>
 
   <target name="release-version">
-    <property name="release.version.regex" value="v([0-9]+)[.]([0-9]+)[.]([0-9]+)(.*)"/>
+    <property name="release.version.regex" value="IDR-([0-9]+)[.]([0-9]+)[.]([0-9]+)(.*)"/>
     <!-- check for Git versioning information -->
     <property name="git.path" value="${root.dir}/.git"/>
     <available file="${git.path}" property="git.present"/>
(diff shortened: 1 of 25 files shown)


joshmoore:
Temporarily closing to allow @manics' build through.

joshmoore:
Error seen by @manis in `BIOFORMATS-maven/179` was that there was no initial branch point tag:

```
[INFO] --- gmaven-plugin:1.4:execute (default) @ pom-bio-formats ---
     [exec] Result: 128
     [exec] Result: 128
128
fatal: No names found, cannot describe anything.
```

I've pushed one to snoopy and will re-test.

joshmoore:
Still not working. Running `git fetch --tags origin` from the command-line did get the tags, but not from the devspace job. Something seems to resolve different.

sbesson:
Closing. Can be reopened when the conflicts are solved.

joshmoore:
I don't think we want this PR as it stands but things to consider:
 * should I salvage the script changes?
 * do we want to use groupId on the spaces?
 * do we ever want a tag prefix? (This is what worked the *least*)

sbesson:
Sorry coming back late to this. Assuming we don't want to dive into different tag prefixes, a potential workflow would be: a fork of https://github.com/openmicroscopy/bioformats defining its own tags following the usual `v{major}.{minor}.{patch}` convention and pushing artifacts with a different `groupId` then `ome` (which is the `groupId` of the OME repo). In this case, we can certainly salvage the `groupId` part of the script.

joshmoore:
Ok. I'll salvage the groupId part of this and we can go from there.

PR #2159, opened 2015-12-15
Enable basic C++11 features

rleigh-dundee:
See https://trello.com/c/h4pHPV2h/344-minimal-c-11-features .  This is a rework of the held 5.1 PR from October for the new split repositories now that 5.2 is open.

For review: https://github.com/openmicroscopy/bioformats/pull/2159/files?w=1 without whitespace changes, which makes the diff much smaller.

Library features:
- `std::shared_ptr` and related classes from `<memory>`.
- `std::array` from `<array>`

Language features:
- `<::` is not  a trigraph, but a template opening followed by an absolute namespace
- `>>` is two templates closing not a right-shift operator

There are no changes to the program logic in this PR.  It's all basic search and replace:

- `ome::compat::shared_ptr` to `std::shared_ptr` (and likewise for `weak_ptr`, `enable_shared_from_this`, and `make_shared`)
- `ome::compat::array` to `std::array`
- `< ::` to `<::` (for templates)
- `> >` to `>>` (for templates)

The language changes are simply removing extra spaces where they are no longer needed to avoid parsing ambiguity.  E.g. `std::vector<ome::compat::shared_ptr< ::ome::xml::model::OMEModelObject> >` will now be `std::vector<std::shared_ptr<::ome::xml::model::OMEModelObject>>`.

The unused headers are dropped (in ome-compat) and the unneeded feature tests dropped for cmake.

The main point of this PR is not to change to code to take advantage of any significant new C++ functionality, but to verify that the CI build nodes can put the compiler into C++11 mode and successfully build with some very trivial changes.  More useful changes such as `unique_ptr`, `enum class`, `auto` and rvalue references can come later.

See also https://github.com/ome/ome-common-cpp/pull/3

--------

Tested by hand on Windows VC12, Ubuntu 15.10, FreeBSD 10.2, MacOS 10.11.

Testing: Covered entirely by the CI jobs since if it's broken it won't build, or the tests will fail.

--------

Note: None of these changes should affect the current set of supported platforms and compilers; they were all capable of this basic featureset already.  The `shared_ptr` and `array` library features were already used when configured to use C++11; this is simply making them required.  i.e. this is the same in effect as the "ON" rows in https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-5.1-merge-superbuild/; we're basically dropping the "OFF" rows.  The only exception (for further changes) might be centos6; the SCL devtoolset-3 exists for this purpose here.

It would be useful to expand the DEV test matrix now the initial split PR is merged and we're happy with things.  It doesn't need to be comprehensive, but adding CentOS6, Ubuntu14.04 and MacOSX 10.11 + perch back in would be useful to ensure there are no regressions.  Adding the non-superbuild builds back in would also be useful to ensure the changes work with the system boost versions.  Any objection to adding some or all of these?

### DIFF
--- a/components/xsd-fu/python/ome/modeltools/property.py
+++ b/components/xsd-fu/python/ome/modeltools/property.py
@@ -455,7 +455,7 @@ class OMEModelProperty(OMEModelEntity):
                 else:
                     itype = "const %s&" % self.langTypeNS
             elif self.maxOccurs > 1 and not self.parent.isAbstractProprietary:
-                itype = "std::vector<std::shared_ptr<%s> >&" % ns_sep
+                itype = "std::vector<std::shared_ptr<%s>>&" % ns_sep
 
         return itype
     argType = property(_get_argType, doc="""The property's argument type.""")
@@ -508,10 +508,10 @@ class OMEModelProperty(OMEModelEntity):
                     itype = {' const': "const %s&" % self.langTypeNS}
             elif self.maxOccurs > 1 and not self.parent.isAbstractProprietary:
                 itype = {' const':
-                         "const std::vector<std::shared_ptr<%s> >"
+                         "const std::vector<std::shared_ptr<%s>>"
                          % ns_sep,
                          '':
-                         "std::vector<std::shared_ptr<%s> >"
+                         "std::vector<std::shared_ptr<%s>>"
                          % ns_sep}
 
         return itype
@@ -641,10 +641,10 @@ class OMEModelProperty(OMEModelEntity):
             elif (self.maxOccurs > 1 and
                     not self.parent.isAbstractProprietary):
                 itype = {' const':
-                         "const std::vector<std::shared_ptr<%s> >"
+                         "const std::vector<std::shared_ptr<%s>>"
                          % ns_sep,
                          '':
-                         "std::vector<std::shared_ptr<%s> >"
+                         "std::vector<std::shared_ptr<%s>>"
                          % ns_sep}
 
         return itype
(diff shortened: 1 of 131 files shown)


rleigh-dundee:
Green in https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge-superbuild/17/ and https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge-win-superbuild/19/

rleigh-dundee:
Note corresponding generated source changes in https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-merge-generated-sources/57/

rleigh-dundee:
Green in

- https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge/3/
- https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge-superbuild/28/
- https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge-win-superbuild/33/

I expanded the superbuild matrix to include the missing platforms now the 5.1 jobs are disabled and the ome-common split is complete.  I also used the updated jenkins-build to add a non-superbuild job for a smaller number of platforms.  The intent is to show that the C++11 support does not cause any failures on our currently-supported platforms, and that it also does not cause any failures with the system libraries for the non-superbuild case.  The exception is centos6, which is built using devtoolset-3, but as you can see from the build results, it's working perfectly acceptably.  We can certainly change the CI setup; I put the above in place primarily to test this PR to ensure it will not cause unexpected problems.

rleigh-dundee:
Closing for now for a docs PR.

PR #2160, opened 2015-12-16
Shape complex type develop

dgault:
This is a rebased and merged copy of https://github.com/openmicroscopy/bioformats/pull/2113
---

What this means:

Shape and LightSource become abstract classes and are no longer elements
There is now a separate index for each Shape and LightSource type (as for other abstract implementations eg Annotations)
Changes to xsd:

Shape and LightSource become complexTypes
Remove abstractproprietary in favour of abstract
Changes to CodeGen:

Remove abstractproprietary in favour of only abstract
Enable code gen for mask bindata
Changes to templates:

Remove abstractproprietary in favour of only abstract
Add in AnnotationCount methods for each of the Shape and LightSource types (same as annotations are currently handled)
Remove hardcoded PixelsBinData and getType methods for Shape and LightSource
Other changes:

Clean-up existing unit tests etc to use the new indexing for Shape and LightSource
To Do:

Clean up all abstract types to use single indexing for all (removing the need for much of the hardcoded template functions)


rleigh-dundee:
Comment on the original PR, which also applies here:

Great stuff for removing AbstractProprietary!  I don't think it goes quite far enough though--there are two fairly major things I can see which need further work:

- Your patch implements separate methods for each abstract type; it would be nicer if these worked with the generic types.  Example, `instrument.addLightSource(laser);` becoming `instrument.addLaser(laser);`.  I know this is how `Abstract` is currently implemented, but the existing implementation in the generator is currently very suboptimal, and needs changing to generate a single list of the abstract type, rather than a separate list for each concrete type.  This will mean most of the user-facing API can remain as it is, which will make the needed changes vastly smaller.

- There are a lot of hardcoded `getFooCount(idx ...)` methods being added to the templates.  These are all bugs in the generator.  Ideally, we update the template with a *single* instance which is expanded for each `Count` method needed, i.e. making the template truly generic.  All the existing `Count` methods are equally buggy and should be also removed once they are also getting generated.

Update: Sorry, I saw that the first point above was in the "todo" section of the PR description.  I think this needs to happen on this PR, otherwise we will be making a very large change to the model API, which will also affect OMERO and other consumers of the model and metadata interfaces, and then change it back again.  It would be best if the changes were done in this PR, if at all possible.

sbesson:
Briefly discussed with @dgault earlier this morning. A couple of thoughts:
- this is a set API decisions to be made including comments from @rleigh-dundee  https://github.com/openmicroscopy/bioformats/pull/2160#issuecomment-165104753 but also how whether we should differentiate generic shape methods e.g.  `setShapeFillColor()` vs `setRectangleWidth()`
- since this is affecting the generated sources, addressing https://trello.com/c/pe32RPuW/69-fix-generated-sources-jobs post ome-common-cpp decoupling will be an immediate requirement
- along with the API decisions above, we should also discuss with @mtbc how this change will affect OMERO and generally how to allow OMERO integration tests to consume this branch as soon as possible to validate breaking changes
- should this be treated/handled as `2016-DEV1` ?

melissalinkert:
Looks like there is also a compile error in MetadataConverter, so presumably this isn't yet included in most of the DEV builds.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-test_images_good/. The XML validation tests are currently failing with

```
[2016-02-10 02:29:44,838] INFO  Parsing schema path
[2016-02-10 02:29:44,839] INFO  http://www.openmicroscopy.org/Schemas/OME/2016-DEV0/ome.xsd
[2016-02-10 02:29:44,839] INFO  Validating OME-XML
[2016-02-10 02:29:45,050] INFO  Error parsing schema at http://www.openmicroscopy.org/Schemas/OME/2016-DEV0/ome.xsd
org.xml.sax.SAXParseException: s4s-elt-invalid-content.1: The content of 'LightSource' is invalid.  Element 'complexType' is invalid, misplaced, or occurs too often.
	at org.apache.xerces.util.ErrorHandlerWrapper.createSAXParseException(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.util.ErrorHandlerWrapper.error(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDHandler.reportSchemaError(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDAbstractTraverser.reportSchemaError(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDComplexTypeTraverser.handleComplexTypeError(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDComplexTypeTraverser.traverseComplexTypeDecl(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDComplexTypeTraverser.traverseGlobal(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDHandler.traverseSchemas(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.traversers.XSDHandler.parseSchema(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.XMLSchemaLoader.loadSchema(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.XMLSchemaLoader.loadGrammar(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.impl.xs.XMLSchemaLoader.loadGrammar(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at org.apache.xerces.jaxp.validation.XMLSchemaFactory.newSchema(Unknown Source) ~[xercesImpl-2.8.1.jar:na]
	at javax.xml.validation.SchemaFactory.newSchema(SchemaFactory.java:627) ~[na:1.7.0_85]
	at loci.common.xml.XMLTools.validateXML(XMLTools.java:716) ~[formats-common.jar:na]
	at loci.formats.services.OMEXMLServiceImpl.validateOMEXML(OMEXMLServiceImpl.java:556) [formats-api.jar:na]
	at loci.formats.services.OMEXMLServiceImpl.validateOMEXML(OMEXMLServiceImpl.java:502) [formats-api.jar:na]
	at loci.tests.testng.FormatReaderTest.testValidXML(FormatReaderTest.java:1831) [classes/:na]
	at sun.reflect.GeneratedMethodAccessor79.invoke(Unknown Source) ~[na:na]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.7.0_85]
	at java.lang.reflect.Method.invoke(Method.java:606) ~[na:1.7.0_85]
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:80) [testng-6.8.jar:6.8-201210030754]
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:714) [testng-6.8.jar:6.8-201210030754]
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:901) [testng-6.8.jar:6.8-201210030754]
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1231) [testng-6.8.jar:6.8-201210030754]
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:127) [testng-6.8.jar:6.8-201210030754]
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:111) [testng-6.8.jar:6.8-201210030754]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145) [na:1.7.0_85]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615) [na:1.7.0_85]
	at java.lang.Thread.run(Thread.java:745) [na:1.7.0_85]
``` 

These changes also bring some method additions at the `ome.xml.meta` level (e.g. `MetadataStore.setExternalhref`...) which currently break the OMERO build on the regions branch. Minimally this will need corresponding patches at the `OMEROMetadataStoreClient.java/OMEROMetadataStore.java` level and there might be some corresponding DB changes required /cc @mtbc.

For now, I am labeling this PR as `breaking` so that we can have it built and tested on CI without impacting the rest of the regions work.

rleigh-dundee:
In files including

components/specification/transforms/2015-01-to-2016-DEV0.xsl
components/xsd-fu/templates-cpp/AggregateMetadata.template
components/xsd-fu/templates-cpp/DummyMetadata.template
components/xsd-fu/templates-cpp/MetadataRetrieve.template
components/xsd-fu/templates-cpp/OMEXMLMetadata.template
components/xsd-fu/templates-java/FilterMetadata.template
components/xsd-fu/templates-java/MetadataStore.template

this PR introduces trailing whitespace.  While this is generally discouraged, in this case it adds a lot of noise to the generated code diffs.  There are also added/removed newlines which also add noise to these diffs.  Please could these be fixed.

`git show --pretty="format:" --name-only develop..HEAD | xargs git grep " $"` will show up the instances which were added in the PR (plus any which were already present).

components/bio-formats-plugins/src/loci/plugins/util/ROIHandler.java
```diff
@@ -944,4 +944,4 @@ private static void storeOval(OvalRoi roi,
         color.getGreen(), color.getBlue(),color.getAlpha());
   }
 
-}
+}
```
rleigh-dundee: Why is the newline at the EOF dropped?

components/formats-api/src/loci/formats/meta/MetadataConverter.java
```diff
@@ -4358,4 +4358,4 @@ private static void convertRootAttributes(MetadataRetrieve src, MetadataStore de
     catch (NullPointerException e) { }
   }
 
-}
+}
```
rleigh-dundee: Why is the newline at the EOF dropped?

components/formats-gpl/test/loci/formats/utests/InOutCurrentTest.java
```diff
@@ -45,9 +45,7 @@
 import javax.xml.transform.stream.StreamResult;
 
 import loci.formats.ome.OMEXMLMetadataImpl;
-
```
rleigh-dundee: Are the changes in this file necessary?

components/specification/released-schema/2016-DEV0/ome.xsd
```diff
@@ -3584,16 +3592,7 @@ If Offset and ReadLength are specified in conjuction with Compression, then they
...
-                                        <xsd:element ref="Line" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Rectangle" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Mask" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Ellipse" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Point" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Polyline" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Polygon" minOccurs="1" maxOccurs="1"/>
-                                        <xsd:element ref="Label" minOccurs="1" maxOccurs="1"/>
-                                </xsd:choice>
+
```
rleigh-dundee: Drop this newline?

components/specification/released-schema/2016-DEV0/ome.xsd
```diff
@@ -3778,6 +3777,9 @@ If Offset and ReadLength are specified in conjuction with Compression, then they
                         </xsd:documentation>
                 </xsd:annotation>
                 <xsd:complexType>
+                                        <xsd:complexContent>
```
rleigh-dundee: Could this section be reindented (maybe something for a followup PR for the entire file, if it's inconsistent).

components/specification/released-schema/2016-DEV0/ome.xsd
```diff
@@ -4672,4 +4695,5 @@ If Offset and ReadLength are specified in conjuction with Compression, then they
                 </xsd:complexType>
         </xsd:element>
 
-</xsd:schema>
+
+    </xsd:schema>
```
rleigh-dundee: Is this space and newline intentional?

dgault:
There is still some tidying up required around indents, new lines etc. They will all be sorted before any formal review listing.

rleigh-dundee:
@dgault OK.  It was listed for review today, which is why I was going through it!

rleigh-dundee:
Regarding the various `OVERRIDE_` settings:

What's the purpose of `COMPLEX_OVERRIDE`?  Would it not be simpler to replace all use of `prop.name in fu.COMPLEX_OVERRIDE` with `prop.name == "Transform"` which is what it looks like the intent is?  The same also applies to `ANNOTATION_OVERRIDE`; would it not make the code simpler and express the intent more accurately to replace it with `prop.name == "AnnotationRef"`?

`SHAPE_OVERRIDE` appears entirely unused.  Is that intended?  Do we actually need it?

What's the purpose of `ABSTRACT_OVERRIDE`?  I see usage like

```
{% for abstractClass in fu.ABSTRACT_OVERRIDE %}\ 
```

What I'm wondering here is whether it's necessary to do this, or if we could otherwise have

```
{% if klass.isAbstract %}\
```

or similar to handle this case this as it loops over *all* the classes?

The reason for asking about these things is that it would be ideal to remove all these special-cased class names out of `config.py` so that the templates can cope with any model we throw at them.

rleigh-dundee:
In the metadata templates:

```
  def getParent(o):
    if o.parentName in fu.ABSTRACT_OVERRIDE:
      return o.parentName
    return o.name
```

Why is this placed into each template?  Could it go into `object.py` in this function?

```
    def _get_parentName(self):
        parent = self._get_parent()
        name = self.modelBaseType

        if (parent is not None and parent.isAbstract and
                self.name not in config.ANNOTATION_OVERRIDE):
            name = parent.name
```

i.e. add it to the existing function.  If it doesn't make sense to go in this function, could we add a different one with a more descriptive name which would meet the intended purpose?

components/xsd-fu/templates-cpp/MetadataRetrieve.template
```diff
@@ -424,7 +423,6 @@ namespace ome
         virtual bool
         getPixelsBinDataBigEndian(index_type imageIndex,
                                   index_type binDataIndex) const = 0;
-
```
rleigh-dundee: Please don't drop newlines in the templates since it's unnecessary and makes the generated source diff hard to review.

components/xsd-fu/templates-cpp/OMEXMLMetadata.template
```diff
@@ -121,6 +121,52 @@ OMEXMLMetadata::${getPropMethod(is_multi_path[o.name], parent, obj.name, prop.na
         return o->get${prop.methodName}();
 {% end %}\
 {% end %}\
+{% when obj.parentName in fu.ABSTRACT_OVERRIDE and prop.isReference and prop.maxOccurs > 1 %}\
+        // ${o.parentName} is abstract, is reference and occurs more than once
+        // ${obj.name} o = (${obj.name}) root.${".".join(accessor(obj.name, parent, prop)[:-1])};
+        // return o.getLinked${prop.methodName}(${index_name_string(prop.name)}).getID();
+        // DUMMYLINE
```
rleigh-dundee: Stray comment to remove?

rleigh-dundee:
The Shape methods are getting generated twice in MetadataRetrieve; in the example below, `getShapeType` and `getShapeCount` are declared 64 times in `MetadataRetrieve.template`.  I would suggest that they might be inside a loop and moving it outside might help here.  Note the indenting of the doxygen comments is also off for these methods.

```
FAILED: /usr/bin/c++   -Dome_xml_EXPORTS -pedantic -Wall -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wextra -Wformat=2 -Wmissing-declarations -Wno-l
ong-long -Wnon-virtual-dtor -Woverlength-strings -Woverloaded-virtual -Wredundant-decls -Wreorder -Wswitch-default -Wunused-variable -Wwrite-strings -
Wno-variadic-macros -Wno-unused-local-typedef -Wno-language-extension-token -fstrict-aliasing -fPIC -I/home/rleigh/code/bioformats/cpp/lib -Icpp/lib -
isystem /tmp/b/superbuild-install/include -MMD -MT cpp/lib/ome/xml/CMakeFiles/ome-xml.dir/meta/Convert.cpp.o -MF cpp/lib/ome/xml/CMakeFiles/ome-xml.di
r/meta/Convert.cpp.o.d -o cpp/lib/ome/xml/CMakeFiles/ome-xml.dir/meta/Convert.cpp.o -c /home/rleigh/code/bioformats/cpp/lib/ome/xml/meta/Convert.cpp
In file included from /home/rleigh/code/bioformats/cpp/lib/ome/xml/meta/Convert.h:42:0,
                 from /home/rleigh/code/bioformats/cpp/lib/ome/xml/meta/Convert.cpp:38:
cpp/lib/ome/xml/meta/MetadataRetrieve.h:340:9: error: ‘virtual const string& ome::xml::meta::MetadataRetrieve::getShapeType(ome::xml::meta::BaseMetadata::index_type, ome::xml::meta::BaseMetadata::index_type) const’ cannot be overloaded
         getShapeType(index_type ROIIndex, index_type shapeIndex) const = 0;
         ^
cpp/lib/ome/xml/meta/MetadataRetrieve.h:287:9: error: with ‘virtual const string& ome::xml::meta::MetadataRetrieve::getShapeType(ome::xml::meta::BaseMetadata::index_type, ome::xml::meta::BaseMetadata::index_type) const’
         getShapeType(index_type ROIIndex, index_type shapeIndex) const = 0;
```

sbesson:
After discussing with @rleigh-dundee earlier today, Roger has set up the https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-DEV-breaking/ with only one configuration to allow this PR tested both on the Java and on the C++ front while labelled as breaking.

rleigh-dundee:
From the new job, see https://ci.openmicroscopy.org/job/BIOFORMATS-CPP-DEV-breaking/2/BUILD_TYPE=Debug,node=cod/console  It's failing to build the C++ code.  There's a bit of a sticky situation here:

- we're building with the 2013-06 model, but the generator is now no longer compatible with this model.
- we can't upgrade to 2016-DEV0 until the generator works; if it's sufficiently fixed with this PR, then simply updating the C++ model version in `cpp/cmake/XsdFu.cmake` *may* be sufficient for this.  However, it might need some additional work relating to units to make this compile.

rleigh-dundee:
@joshmoore Regarding the brief discussion around units, please see https://github.com/openmicroscopy/bioformats/compare/develop...rleigh-dundee:model-enum-metadata?expand=1 for a proposed update to the schema to add all the missing information, including the enum names.  The code to extract this in xsd-fu property.py isn't added here; it's just an initial proof of concept for some feedback.

sbesson:
@dgault: "This branch has conflicts that must be resolved". Can you merge `origin/develop` into this branch?

dgault:
Additional modifications made to the PR since the original:
- Indexing has been resolved to return to single indexing for Shape and LightSource
- Schema has been updated to use the recommended substitution groups for abstract types Shape and LightSource. Codegen changes have been made to support this. This will provide us with a valid schema which will also validate against the new upgraded XML with concrete Shape types interchangeable.
- Based on feedback the previous getParent function in the templates has been moved to the python object class as a property and renamed as topLevelName
- unused config overrides removed. I have left the annotation and abstract override in to avoid hardcoding property values in the templates. These could possibly be combined or reworked if this is no longer the way we wish to proceed.

Further work for follow up PR's:
- CPP codegen for Units to support generation of 2016 schema
- BinData requires further codegen update to support base64Binary

Testing:
- Builds should remain green with travis and CI passing. Generated sources will appear as unstable due to newly generated source classes.
- Diff of outline of the metadata classes with develop branch. API should remain unchanged.
- Sanity check of any sample images using Shapes or LightSources
- Unit tests of particular interest will be the schema file upgrade and instrument upgrade along with file reader and writer tests. It might also be useful to run a file downgrade test from 2016-DEV0 to a previous revision. XML can be validated to verify that the correct properties have transferred, for example below, attributes ID, FillRule and Text belong to Shape rather than Rectangle, as does the node Transform:

<Rectangle FillRule="EvenOdd" ID="Shape:2" Text="Hi There! (from shape 2)">
<Transform A00="10" A01="30" A02="50" A10="20" A11="40" A12="60"/>
</Rectangle> 

when downgraded should become:
<ROI:Shape FillRule="EvenOdd" ID="Shape:2" Text="Hi There! (from shape 2)">
<ROI:Transform A00="10" A01="30" A02="50" A10="20" A11="40" A12="60"/>
<ROI:Rectangle FillRule="EvenOdd" ID="Shape:2" Text="Hi There! (from shape 2)"></ROI:Rectangle>
</ROI:Shape>

rleigh-dundee:
On the Java side, the code generation is looking much better, thanks.

On the C++ side, it's broken unfortunately.  The concrete LightSource and Shape types are no longer deriving from the abstract LightSource and Shape types.  See https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-breaking-generated-sources/lastSuccessfulBuild/artifact/ome-xml-model-cpp.diff/*view*/  It's also adding in unwanted whitespace in the diffs.

I think that we need to either (1) conditionally use the old logic, adding it back if necessary or (2) temporarily hard-code the relationships in the generator to force it to use the correct behaviour, if that's possible.

dgault:
Something seems to have gone wrong with the CPP build last night, currently investigating as to what. Its certainly producing different results from what I am generating locally.

rleigh-dundee:
@dgault The other issue is that it's not using the generic methods on Shape in Convert.cpp:

```
09:52:15 /opt/hudson/workspace/BIOFORMATS-CPP-DEV-breaking/BUILD_TYPE/Debug/node/cod/ome-xml/cpp/lib/ome/xml/meta/Convert.cpp:779:67: error: ‘setLaserModel’ is not a member of ‘{anonymous}::MetadataConverter::MS {aka ome::xml::meta::MetadataStore}’
09:52:15                    transfer(&MR::getLaserModel,                   &MS::setLaserModel,                   instrumentIndex, lightSource);
```

Convert.cpp would need updating to use the new way of doing things, i.e. the generic methods on Shape rather than the same method duplicated for each concrete type.  That would be the ideal if possible.  But depends on if it's possible to do that with the old model, or if that needs to wait.

components/specification/released-schema/2016-DEV0/ome.xsd
```diff
@@ -2613,9 +2622,13 @@
                         </xsd:documentation>
                 </xsd:annotation>
                 <xsd:complexType>
-                        <xsd:sequence>
-                                <xsd:element ref="Map" minOccurs="0" maxOccurs="1"/>
-                        </xsd:sequence>
+                        <xsd:complexContent>
+            				        <xsd:extension base="LightSource">
```
mtbc: strange indent

dgault: Yeah theres a tab has snuck in there, should be fixed now

rleigh-dundee:
@dgault One additional blocking issue: In `language.py`, the Units* classes are still present in `primitive_type_map`.  They aren't primitives and this breaks the C++ build, so they need to be moved out.

sbesson:
From the Bio-Formats/OMERO stack point of view, http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/61/ is certainly passing with this PR included.

dgault:
CPP jobs are now greenhttps://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-CPP-DEV-merge-trigger/129/

sbesson:
http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-start/68/ is also green with this PR included. Any last veto to getting this merged @joshmoore @rleigh-dundee @mtbc ?

mtbc:
Go for it!

rleigh-dundee:
There are still some whitespace problems in https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-generated-sources/lastSuccessfulBuild/artifact/ome-xml-metadata-cpp.diff/*view*/ -- there are lines removed between methods and also some lines added in places as well.

sbesson:
Thanks all. @rleigh-dundee: thanks for spotting this. I would say let's handle the whitespace issues in a follow-up PR and merge this epic one to unlock the rest of the model work.

PR #2162, opened 2015-12-16
cmake: Add model-source-only option

rleigh-dundee:
This is a temporary change to permit the model source to be generated without a dependency upon ome-common, which isn't needed for source generation.  The source will not build with this option enabled, but the code generator will run.

When the model code is split into a separate repository, this option can be moved there, or redone in a better way.

--------

Testing: Add `-Dmodel-source-only=ON` to source generation job, and it should work again.  Done in https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-merge-generated-sources/ and https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-latest-generated-sources/ -- the latter will require the PR merging before it will go green.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,17 @@ include(XsdFu)
 include(GTest)
 include(Doxygen)
 include(HeaderTest)
-find_package(OME 5.2.0 REQUIRED Compat Common)
+
+# This is a hack to generate the model sources without having to have
+# ome-common available.  It will be removed or moved once the model
+# has been split out.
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
@sbesson @melissalinkert See https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-DEV-merge-generated-sources/57/ - this fixes the generated-sources job.

sbesson:
Thanks @rleigh-dundee. Merging to restore the generated-sources job

PR #2164, opened 2015-12-17
Princeton SPE follow-up

melissalinkert:
Fixes a couple of minor issues noticed when reviewing gh-1965.

To test, compare the output of ```showinf -nopix -omexml``` on the ```curated/pi-spe``` files with and without this PR.  The only visible difference with this PR should be that the valid bits/bitsPerPixel/BitsPerSample values now match the full bit width of the pixel type, instead of being the number of bytes.  The ROI indexing changes shouldn't visibly affect the existing sample data, only files that have multiple ROIs (so careful code review is sufficient).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SPEReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SPEReader.java
@@ -43,6 +43,10 @@ import loci.formats.meta.MetadataStore;
 /**
  * SPEReader is the file format reader for Princeton Instruments SPE .spe files.
  * XML Footer introduced in Princeton Instruments SPE Ver 3.0 not supported
+ *
+ * See public specification document:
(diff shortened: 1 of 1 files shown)


sbesson:
Quickly tested using the sample `spe` files. The visible effects of this changes on the populated `SignificantBits` work as described above and the [automated tests](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/72/) pass with this PR included. Leaving @dgault to have a final review at it at the beginning of 2016.

dgault:
Im happy with all 3 of those changes. 

The bitsPerPixel was certainly wrong and now understanding the ROI and Shape indexing much better, the Shape indexing was also incorrect and should be corrected by these changes.

PR #2168, opened 2016-01-05
TiffSaver: fix recorded offsets to non-interleaved RGB tiles

melissalinkert:
This prevents earlier tiles from being overwritten by later tiles.
Strip-based data, single channels, and interleaved RGB tiles should be
unaffected.

See:

http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-October/005680.html
https://trello.com/c/nFq6AzGi/53-writing-tiff-in-tiles
http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2015-December/003531.html

I was able to reproduce the problem mentioned in the October 2015 thread using:

```bfconvert test_images_good/tiff-big-images/2kx2k.tif 2kx2k-tiled.ome.tiff -tilex 1000 -tiley 1000```

Without this change, ```2kx2k-tiled.ome.tiff``` should not match the original file (mixed up colors and tiles).  With this change, the two files should match when viewed in showinf/ImageJ/ImageMagick.

It would also be good to test the following to make sure that this change doesn't break anything else:

* ```bfconvert test_images_good/tiff-big-images/2kx2k.tif 2kx2k-copy.ome.tiff```
* bfconvert with and without ```-tilex```/```-tiley``` on a single-channel image
* bfconvert with and without ```-tilex```/```-tiley``` on a multi-channel interleaved image (e.g. anything in ```test_images_good/png/```)

As discussed earlier with @sbesson and @joshmoore, happy to exclude if this seems too distracting.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -504,7 +504,7 @@ public class TiffSaver {
     int tileCount = isTiled ? tilesPerRow * tilesPerColumn : 1;
     for (int i=0; i<strips.length; i++) {
       out.seek(out.length());
-      int index = interleaved ? i : i / nChannels;
+      int index = interleaved ? i : (i / nChannels) * nChannels;
       int c = interleaved ? 0 : i % nChannels;
       int thisOffset = firstOffset + index + c * tileCount;
       offsets.set(thisOffset, out.getFilePointer());
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -353,14 +353,14 @@ public class TiffSaver {
                   }
                   else {
                     off = c * blockSize + ndx + n;
+                    int realStrip = c * (nStrips / nChannels) + strip;
                     if (row >= h || col >= w) {
-                      stripOut[strip].writeByte(0);
+                      stripOut[realStrip].writeByte(0);
                     } else if (off < buf.length) {
-                      stripOut[c * (nStrips / nChannels) + strip].writeByte(
-                          buf[off]);
+                      stripOut[realStrip].writeByte(buf[off]);
                     }
                     else {
-                      stripOut[strip].writeByte(0);
+                      stripOut[realStrip].writeByte(0);
                     }
                   }
                 }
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
The tiling is correct for 1000×1000, but try with 64×64 and you see it result in visual artefacts on the right-hand edge:

showinf-tifftile1

Interestingly, this happened for 2kx2k.tif but not for 2kx2k.png.  The other tests all pass.

ximenesuk:
See: https://github.com/openmicroscopy/openmicroscopy/pull/4403#issuecomment-171356138

sbesson:
Excluding as per https://github.com/openmicroscopy/openmicroscopy/pull/4403#issuecomment-171367077

ximenesuk:
SEe: https://ci.openmicroscopy.org/job/OMERO-DEV-breaking-integration-python/473/

melissalinkert:
Commit pushed to fix the edge tile issue noted above.  I think this will also fix the exception noted by the OMERO integration tests, but I'm having trouble verifying locally.  Leaving excluded for now to prevent build disruption tomorrow.

ximenesuk:
I'll test this with a local build tomorrow against the OMERO integration tests.

ximenesuk:
The previously failing test now passes with this PR, feel free to include for further testing.

melissalinkert:
One more commit pushed for https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-build/100/

rleigh-dundee:
With the latest commit, I can't reproduce the corruption at the end of each row above.  All the other cases continue to look correct.

Looks fine to merge to me.  The only minor comment I have is that some comments explaining the difference between`realStrip` vs `strip` would be useful for future maintenance, likewise as a general point for other calculations.  Also, again as a general readability and maintainability point, having calculations without parentheses makes them hard to read unambigously; I know the operator precedence rules, but having parentheses makes the intent absolutely explicit and unambiguous.  (Not just for this PR, but something to consider generally.)

melissalinkert:
Some clarifications pushed for final review.

Re:

```The TiffData/IFD corruption issue with bfconvert noted above is still present, but likely unrelated to this PR.```

I don't see anything relevant above - is this https://trello.com/c/R1x1tmaP/78-bfconvert-writes-invalid-ome-tiff?

rleigh-dundee:
@melissalinkert Yes, that's the card/ticket.  Sorry, I thought I'd also commented about it here when testing.

rleigh-dundee:
Thanks for adding the comments.  Looks good to merge to me.

rleigh-dundee:
--rebased-to #2371

PR #2172, opened 2016-01-06
docs: Add note about missing C++ XSL transforms

rleigh-dundee:
Guidance about missing XSLT transforms.  Should be possible to remove in the next few weeks, but adding to document the current state.  I'll remove it in the PR adding transform support.


mtbc:
Reads fine to me. :+1:

mtbc:
As the staging docs job failed today, for the record here's the rendered version:
paragraph

PR #2174, opened 2016-01-07
FilePattern docs revision

simleo:
This PR extends the javadocs for `FilePattern{,Block}`. Part of the new docs for `FilePattern` are taken from a long non-javadoc section at the end of the file, which has been deleted. A few `FilePattern` methods had a commented out deprecation tag, so they were actually not deprecated. New javadocs have been added for these methods, and they are now checked by the unit test.

### DIFF
--- a/components/formats-bsd/src/loci/formats/FilePattern.java
+++ b/components/formats-bsd/src/loci/formats/FilePattern.java
@@ -48,23 +48,56 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * FilePattern is a collection of methods for handling file patterns, a way of
- * succinctly representing a collection of files meant to be part of the same
+ * Represents a collection of files meant to be part of the same
  * data series.
- *
- * Examples:
- * <ul>
- *   <li>C:\data\BillM\sdub&lt;1-12&gt;.pic</li>
- *   <li>C:\data\Kevin\80&lt;01-59&gt;0&lt;2-3&gt;.pic</li>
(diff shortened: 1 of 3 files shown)


melissalinkert:
A couple of Javadoc warnings: https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-build/91/warnings39Result/

components/formats-bsd/src/loci/formats/FilePattern.java
```diff
@@ -723,7 +897,11 @@ private void buildFiles(String prefix, int ndx, List<String> fileList) {
     return last;
   }
 
-  /* @deprecated */
+  /**
+   * Gets the STEP element for all blocks.
+   *
+   * @return an array containing all STEP elements.
+   */
```
melissalinkert: Just to clarify - your intention was to un-deprecate these three methods?  I'm fine with that given that they're still being used, just want to make sure it isn't accidental before merging.

melissalinkert:
All reads fine to me, just waiting on one clarification and confirmation from the merge build.

components/formats-bsd/src/loci/formats/FilePattern.java
simleo: Actually they were *not* deprecated, since `/* @deprecated */` is not a javadoc, so i just removed some misleading comments and added missing javadocs.

melissalinkert: OK, fair enough.

PR #2176, opened 2016-01-08
minor cleanup for 2016-DEV0/ome.xsd

mtbc:
This PR should not change behavior. Everything should still work and CI should be green. The keen may want to try upgrading an OME-XML file to 2016-DEV0 then validating it.

--no-rebase


rleigh-dundee:
The changes look absolutely fine to me.  I'll test it later.

rleigh-dundee:
Upgrade and validation is fine.  Please merge.

PR #2178, opened 2016-01-11
Synchronize OME-XML samples with the upstream sample changes

sbesson:
In preparation of samples addition/modification changes planned in https://trello.com/c/gskA0JKM/65-schema-samples-generation and since we haven't finalized our workflow for updating/proposing sample files, this PR:
- updates the OME-XML samples under `components/specification/samples` using the content of http://downloads.openmicroscopy.org/images/OME-XML/
- removes all binary samples (OME-TIFF, PNG) from the repository

/cc @mtbc @will-moore


components/specification/samples/2010-04/filter.ome.xml
```diff
@@ -81,9 +81,6 @@
 					<EmissionFilterRef ID="Filter:6"/>
 				</LightPath>
 			</Channel>
-			<Channel ID="Channel:3">
-				<FilterSetRef ID="FilterSet:2"/>
-			</Channel>
```
mtbc: I wonder why this changed.

sbesson: Primarily because it was not passing `xmlvalid`: there was a discrepancy between the defined `sizeC` and the number of `Channel` elements.

mtbc:
Is `libxml2/example.c` not useful? I've really no idea.

components/specification/samples/2010-04/filter.ome.xml
mtbc: Aha, okay.

sbesson:
My understanding from the `README` and the file history is that the `libxml2` content is largely obsolete and has been unused for sample generation process of all the recent schemas. Being one of the people originally involved, @chris-allan might have additional thoughts.

chris-allan:
Back in the day when that code was originally written we had absolutely no C or C++ example code for writing OME-XML in any form. If you guys feel that's been superseded, obsolete or just plain broken and should be removed I can't see any reason to argue with that.

rleigh-dundee:
@sbesson The removal of the samples is why the C++ builds have been failing the last few days.  I can copy the selected few we need into the C++ test directory if we need to do that.  Just to let you know that this is causing the breakage though.

sbesson:
Discussed with @rleigh-dundee:
- I will push a few commits to fix the C++ tests affected by the samples renaming (incl. the OME-TIFF)
- follow-up discussions might follow on rationalizing a central location for broken samples /cc @pwalczysko

sbesson:
Closing this for now as I haven't found the time to get back to it and we are still salvaging some `develop` OME-XML samples showing the new unit concepts.

PR #2180, opened 2016-01-13
Reader/writer discovery

melissalinkert:
This uses https://github.com/ronmamo/reflections to build the list of valid readers and writers at runtime.  Each reader or writer must have a class annotation (```@BioFormatsReader``` or ```@BioFormatsWriter```) that specifies a value from the ```FormatType``` enum.  ```FormatType``` has values that correspond to the sections in ```readers.txt```; the ```FormatType``` assigned to a reader/writer determines its priority during type checking.

Right now, ```ImageReader``` and ```ImageWriter``` will only scan the ```loci.formats``` package; see line 97 of ```ClassList```.  I know that's limiting for anyone wanting to add a third-party reader, but scanning all packages on the classpath adds substantially to the scan time (and thus setId time in most cases).  Suggestions for how to better handle that would be welcome - only thing I can think of at the moment is to have an environment variable that specifies additional packages to search.

The main thing to check will be the status and run time of the merge-full-repository job, as that should validate that the new reader ordering works and that the annotation scan time is not prohibitive.

Leaving excluded until we decide if this is worth getting in to 5.2.0.


melissalinkert:
Closing at least for now after discussion with @sbesson.

PR #2184, opened 2016-01-15
Adding public sample links & deleting old page

hflynn:
Addresses https://trello.com/c/O0uwuqLj/82-review-public-data-samples-policy

Removes the Testing Notes page and instead adds links to public sample images to the individual format pages.

Issues which may need discussed:
- I've added additional links rather than make the existing 'we have datasets' bullet points into links so we distinguish between what we have and what we can share, but this is possibly redundant for formats where we have public sample data and could be replaced with one line.
- I've made the Incell 1000 page cover InCell 2000 too as they use the same reader. We may want to add a separate format page for it instead.


components/autogen/src/format-pages.txt
```diff
@@ -453,10 +454,11 @@ writer = CellH5Writer
 mif = true
 
 [Cellomics]
-extensions = .c01
+extensions = .c01, .dib
```
sbesson: This change should also be propagated to `supported-formats.txt` as well as the specific format page.

hflynn: Ah, I'd missed that! Fix pushed now thanks.

components/autogen/src/format-pages.txt
```diff
@@ -453,10 +454,11 @@ writer = CellH5Writer
...
 developer = `Thermo Fisher Scientific <http://www.thermofisher.com/>`_
 bsd = no
-weHave = * a few Cellomics .c01 datasets
+weHave = * a few Cellomics .c01 datasets /n
```
sbesson: `/n` should be `\n`

components/autogen/src/format-pages.txt
```diff
@@ -453,10 +454,11 @@ writer = CellH5Writer
...
+* `public .dib sample images <http://downloads.openmicroscopy.org/images/HCS/BBBC/>`_ \n
```
sbesson: No need for `\n` as this is the last item.

components/autogen/src/format-pages.txt
```diff
@@ -501,7 +503,8 @@ owner = `GE Healthcare (formerly Applied Precision) <http://www.gelifesciences.c
 bsd = no
 software = `DeltaVision Opener plugin for ImageJ <http://rsb.info.nih.gov/ij/plugins/track/delta.html>`_
 weHave = * a DV specification document (v2.10 or newer, in HTML) \n
-* numerous DV datasets
+* numerous DV datasets \n
+* `public sample images <http://downloads.openmicroscopy.org/images/DV/>`__ \n
```
sbesson: No `\n` else `pixelsRating = Outstanding` gets added to the generated list.

components/autogen/src/format-pages.txt
```diff
@@ -962,11 +965,12 @@ utilityRating = Fair
 reader = OBFReader
 mif = true
 
-[InCell 1000]
+[InCell 1000/2000]
 extensions = .xdce, .tif
 developer = `GE <http://gelifesciences.com/>`_
 bsd = no
 weHave = * a few InCell 1000 datasets
+* `public InCell 2000 sample images <http://downloads.openmicroscopy.org/images/HCS/INCELL2000/>`_ \n
```
sbesson: No trailing `\n`

components/autogen/src/format-pages.txt
```diff
@@ -1789,7 +1795,8 @@ mif = true
 extensions = .tiff, .xml
 developer = `PerkinElmer <http://www.perkinelmer.com/>`_
 bsd = no
-weHave = * a few sample datasets
+weHave = * a few sample datasets \n
+* `public sample images <http://downloads.openmicroscopy.org/images/HCS/Operetta/>`__ \n
```
sbesson: No trailing `\n`

sbesson:
Some inline comments on `format-pages.txt` originating from computing the local diff after regenerating the documentation pages with `ant -f components/autogen/build.xml gen-meta-support gen-structure-table gen-format-pages`. Otherwise the expected Sphinx pages outcome reads well to me. https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/search.html?q=public+sample is a nice way to list all file formats for which we have public samples available.

Leaving @melissalinkert to give her thoughts on the InCell 1000/2000 duality.

hflynn:
Thanks, I'll fix the trailing \n issues now

melissalinkert:
All reads fine to me.

sbesson:
Regenerated the auto-generated documentation page from the last commit and everything seems fine. This PR is introducing a few whitespaces but otherwise everything seems fine.  Note a major implication of the `InCell 1000` -> `InCell 1000/2000` change is that the format page is now renamed as `formats/incell-10002000` potentially breaking stable URLs.

hflynn:
Hmmm, is there a way to have the titles be changed without changing the URL? I hoped not changing the filename would keep it constant.

hflynn:
Is it changed though? https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/formats/incell-1000.html

hflynn:
Or did you mean we should really change the name too to be consistent?

sbesson:
@hflynn: I mean the name of the page is generated by the `autogen` Ant target pasted above. To have a page name not generated from the format automatically, you can use the `pagename = xxx` field though.

hflynn:
Hopefully that's fixed it! (Wasn't quite sure whether I need the `-` or not)

sbesson:
Works now. Re-running https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-docs/ for final validation then we should be able to merge (and rebase on `dev_5_1`?)

hflynn:
Happy to rebase once merged if we think there's likely to be another 5.1.x release

hflynn:
--rebased-to #2189

PR #2186, opened 2016-01-18
C++: Add model transform resolver for discovery and graph-based query

rleigh-dundee:
- Install transforms
- Set transform directory inside and outside the build, and when running tests
- `OMETransformResolver` detects all available transforms and builds a graph (vertices are schemas, edges are transforms between schemas)
- This class permits querying of available schemas and of the set of transforms to apply to change from a given source to target schema.  Schemas are discovered by searching the transforms directory for "{source}-to-{target}.xsl" files.
- Transform quality is determined using the same algorithm as in `components/specification/transforms/util/write-transforms.py`, and used to weight the edges for use in a shortest path search.

Note it is based upon https://github.com/openmicroscopy/bioformats/pull/2185 and so includes the commits for this PR as well.  The actual diff for this PR alone in https://github.com/rleigh-dundee/bioformats/compare/cpp-c++11...rleigh-dundee:cpp-transform-graph?expand=1

Update: Removed C++11 support, including delegating constructors, uniform initialisation, automatic type deduction, range-based for, const iterator methods, fstream string constructors, `tie`, `enum class` and `unique_ptr`.  It is now C++98.

This is the transform graph the code will construct and work from:

ome-transforms

--------

Testing:

In the build log, search for `ome-xml/transform-resolver`.

- Test 1 emits a list of schemas plus a `dot` input graph.  Copy the graph to a file, and run `dot -T ome-transforms.dot > ome-transforms.svg` and look at the SVG file.  This is a readable representation of the internal transform graph, including the transform quality and weighting.  This should be the same as shown above.
- Test 2 emits a list of schemas plus an `ome-transforms.xml` document.  Copy the document to a file and diff with `components/specification/transforms/ome-transforms.xml`.  The two should be identical, confirming that this implementation behaves identically to the existing hardcoded list in `ome-transforms.xml` which is now generated by `write-transforms.py`, i.e. for any given pairing of source and target schema version, the set of transforms is the same.
- Verify that the transforms were installed (check the build archive in the artefacts), and/or the build log.

(You can also find these files in `bioformats-build/cpp/test/ome-xml/data` if you do your own build.  I have them included in the build log for convenience.)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -43,7 +43,7 @@ set(OME_TOPLEVEL_INCLUDES
 function(bf_add_test)
   add_test(${ARGV})
   set_tests_properties(${ARGV0}
-    PROPERTIES ENVIRONMENT "BIOFORMATS_SCHEMADIR=${PROJECT_SOURCE_DIR}/components/specification/released-schema")
+    PROPERTIES ENVIRONMENT "BIOFORMATS_SCHEMADIR=${PROJECT_SOURCE_DIR}/components/specification/released-schema;BIOFORMATS_TRANSFORMDIR=${PROJECT_SOURCE_DIR}/components/specification/transforms")
 endfunction()
 
 add_subdirectory(lib)
--- a/cpp/cmake/TemplateCmdWrapper.cmake.in
+++ b/cpp/cmake/TemplateCmdWrapper.cmake.in
@@ -3,5 +3,6 @@
 set "BIOFORMATS_ICONDIR=@PROJECT_SOURCE_DIR@/cpp/share/icons"
 set "BIOFORMATS_MANDIR=@PROJECT_SOURCE_DIR@/cpp/man"
 set "BIOFORMATS_SCHEMADIR=@PROJECT_SOURCE_DIR@/components/specification/released-schema"
+set "BIOFORMATS_TRANSFORMDIR=@PROJECT_SOURCE_DIR@/components/specification/transforms"
 
 %*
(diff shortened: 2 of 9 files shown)


rleigh-dundee:
https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge-win-superbuild/85/ is green,
https://ci.openmicroscopy.org/view/DEV/job/BIOFORMATS-CPP-DEV-merge-superbuild/71/ is green.  Any of these can be used for the build log and archive to check.

joshmoore:
Looks good! My PNG matches what you've pasted here. Just briefly:

 * `OMETransformResolver` is intended for end users (eventually) or internal-only?
 * And based on that, any pointers needed in the documentation needed beyond what you have in doxygen?

rleigh-dundee:
@joshmoore Initially only intended for internal use.  It's mainly user-visible for unit testing and decoupling from the actual transform logic.  I'm currently thinking of having an `OMETransform` class which will handle the user-facing side of transforms--it will have a method taking as input

- an XML document/string/filename as input
- one the the former as output
- source version (optional--could be derived from input if not too expensive)
- target version (may be optional since we could default to current model version)

Internally, this class can use the `OMETransformResolver` to get the list of transforms to apply.  It can also potentially cache the transforms if needed (right now, I don't think it's necessary, the overhead is a few milliseconds), but it's something which could be implemented if required.  I expect that individual readers will use `OMETransform`, and `OMETIFFReader` will certainly be the initial place we'll use it (and unit tests, of course).

joshmoore:
:+1: when the time comes, it'll be interesting to discuss how this maps to Java.

rleigh-dundee:
@joshmoore We could certainly reimplement it in Java if we wanted, and drop the generated `ome-transforms.xml` entirely.  This would mean no future updates to the generated-but-static file would be required, making model changes a bit simpler since the system will adapt to whatever transforms are detected.  The only possible future change would be adapting the heuristic to change the assigned transform quality if we make additional transforms which don't fit the current pattern.

PR #2187, opened 2016-01-18
AxisGuesser: fix javadocs

simleo:
Fixes javadocs for `AxisGuesser`.


melissalinkert:
All makes sense, merging.

PR #2189, opened 2016-01-19
Adding public sample links & deleting old page (rebased onto dev_5_1)

hflynn:
This is the same as gh-2184 but rebased onto dev_5_1.

----

Addresses https://trello.com/c/O0uwuqLj/82-review-public-data-samples-policy

Removes the Testing Notes page and instead adds links to public sample images to the individual format pages.

Issues which may need discussed:
- I've added additional links rather than make the existing 'we have datasets' bullet points into links so we distinguish between what we have and what we can share, but this is possibly redundant for formats where we have public sample data and could be replaced with one line.
- I've made the Incell 1000 page cover InCell 2000 too as they use the same reader. We may want to add a separate format page for it instead.


hflynn:
--rebased-from #2184

hflynn:
Also fixed OME-TIFF page which hadn't been updated to the 2015 Schema.

sbesson:
All the initial changes from gh-2184 are rendered properly in http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/.

On the schema addition a couple of comments:
- the 2015-01 schema should likely be added to the list for the OME-XML format as well
- the schema addition has not been propagated to the individual OME-TIFF format page. This can be done either here or by opening an auto-generated documentation PR once this is merged.
- this last addition will need to be ported forward to `develop`

hflynn:
I can push a further commit here to fix the OME-XML page and propagate the TIFF change to the format page.
The format pages for OME-TIFF and OME-XML on develop already have the 2015-01 schema listed (that's why I had to fix a conflict on the rebase and noticed it was missing on the 5.1 docs) so there is nothing to port forward.

hflynn:
There, should be sorted now hopefully.

PR #2182, opened 2016-01-14
Allow for simpler fake plate definition

ximenesuk:
This change allows for fake plates to be created by simply having one of the relevant keys in the name set to 1 or more. All non-set keys are then assumed to be equal to 1. See https://trello.com/c/f9tTxVFz/372-improve-fake-parameter-handling

Thus,
```
SPW&plates=1.fake
SPW&plates=1&plateRows=1&plateCols=1&plateAcqs=1&fields=1.fake
```
would both generate the same plate form.

I'm not sure how best to test this. I have done so locally using the breaking build of OMERO - and in fact the fake plate tests in https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/clitest/test_import.py in that build should pass with this PR included in Bio-Formats. The Bio-Formats unit tests can't easily be changed to test this as they don't appear to test filename-style fakes. Any other testing suggestions welcome.

If this is mergeable then the now merged docs PR, https://github.com/openmicroscopy/bioformats/pull/2170, will need an additional PR.

PR #2191, opened 2016-01-20
Allow for simpler fake plate definition (docs)

ximenesuk:
This docs PR supports gh-2182

### DIFF
--- a/docs/sphinx/developers/generating-test-images.txt
+++ b/docs/sphinx/developers/generating-test-images.txt
@@ -44,16 +44,31 @@ To generate a simple plate file:
 
     touch "simple-plate&plates=1&plateAcqs=1&plateRows=1&plateCols=1&fields=1.fake"
     touch "default-plate&plates=1.fake"
+    touch "default-plate&screens=0&plates=1.fake"
+
+These will each create a single plate without a containing screen, by default
+in the first two cases. In the third case setting ``screens`` to zero is used
+to document the lack of a screen. As above a :file:`.fake.ini` file can be
+used.
+
+To generate a simple screen file:
+
+::
+
     touch "default-screen&screens=1.fake"
 
-These will each create a single plate which is within a containing screen,
-by default in the first two cases. To create a plate with no containing screen
-add ``&screens=0`` to the key-value pairs in the file name. As above a
-:file:`.fake.ini` file can be used.
+This will create a screen containing a single simple plate.
 
 To generate a valid plate at least one of ``screens``, ``plates``,
 ``plateAcqs``, ``plateRows``, ``plateCols`` and ``fields`` must be greater
-than zero.
+than zero. If this condition is met then any other plate-specific values set
+to zero will be ignored and the defaults used. So, for example, the file:
+
+::
+
+    one-key-set&screens=0&plates=0&plateRows=0&plateCols=0&plateAcqs=0&fields=1.fake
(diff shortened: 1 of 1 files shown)


docs/sphinx/developers/generating-test-images.txt
```diff
@@ -42,17 +42,18 @@ To generate a simple plate file:
...
+    touch "default-screen&screens=1.fake"
 
-This will create a single plate which is, by default, within a containing
-screen. To create a plate with no containing screen add ``&screens=0`` to the
-key-value pairs in the file name. As above a :file:`.fake.ini` file can be
-used.
+These will each create a single plate which is within a containing screen,
+by default in the first two cases. To create a plate with no containing screen
+add ``&screens=0`` to the key-value pairs in the file name. As above a
+:file:`.fake.ini` file can be used.
```
sbesson: I am confused  by this paragraph. With 722701bd6d088f2537e1a10867583b33e705edce, the first two example are generating an orphaned plate. Isn't it the opposite, i.e. no screen by default and to add to a screen, `&screens=1` should be appended to the file name?

ximenesuk: Sorry, I'll double check the behaviour and try to make the paragraph clearer.

ximenesuk: Fixed below. I blame my extremely badly worded commit: https://github.com/ximenesuk/bioformats/commit/722701bd6d088f2537e1a10867583b33e705edce which states the opposite of what it does!

ximenesuk:
There are still some rows in the key-value table that are unexplained and, for me, not obvious from the code.

 + ` thumbnail` : is this just whether there is a thumbnail in the file itself?
 + `scaleFactor`: what is scaled here?
 + `color` : what does this color apply to?
 +  `color_x` : this is dependent on the one above.

sbesson:
@ximenesuk: as far as I can tell from the source code:
- `color_x` sets the color of the `x` channel (expressed in rgba)
- `color` sets the default color for all channels (and is overriden by `color_x`)
- `scaleFactor` is a scaling factor for the pixel values on each plane
- `thumbnail` toggles on/off the [CoreMetadata.thumbnail](http://downloads.openmicroscopy.org/bio-formats/5.1.7/api/loci/formats/CoreMetadata.html#thumbnail). I assume this would be used in the case of fake multi-series images @melissalinkert ?

melissalinkert:
```thumbnail``` would only be used if multiple series are present; by default it is always false.  Setting it to true for a specific series means that the series is a lower resolution/thumbnail/icon representing one of the other defined series.  It's not a common use case for .fake data - if I recall, we only added it for completeness.

ximenesuk:
Thanks both. For `thumbnail` I will simply add "toggles on/off the `CoreMetadata.thumbnail`". If anyone reall wants to use it I guess they will know what that means.

ximenesuk:
@sbesson I'll re-list this for Monday unless you get time to check a local docs build.

sbesson:
@ximenesuk: quickly browsing through the text, everything reads fine, thanks. Please list for Monday for a final review.

sbesson:
Staged at http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/developers/generating-test-images.html. Reads well. Thanks @ximenesuk.

PR #2192, opened 2016-01-20
cmake: Add Xalan-C checks

rleigh-dundee:
Required by https://github.com/ome/ome-common-cpp/pull/7 to use the exported dependency configuration.  The changes are the same as in this PR (adding the same support files).

--------

Testing: check job is green; configuration and/or linking will fail if there's a problem.


sbesson:
See https://github.com/ome/ome-common-cpp/pull/7#issuecomment-183296732

PR #2181, opened 2016-01-14
Fix FEITiffReader metadata parser. Fix #2179

avandecreme:
According to http://stackoverflow.com/questions/4567636/java-sax-parser-split-calls-to-characters , the processing should be done in endElement instead so that XML element value "1.943794245326921E-09" doesn't get split in two characters calls.

PR #2194, opened 2016-01-20
Fix FEITiffReader stage position parsing.

avandecreme:
I also modified the end element code so that even if the string builder is empty, the parentNames get peeked (to be consistent with the code before #2181 ).


dgault:
The proposed change makes sense and looks good to me.
Files appear as expected in tests and FEI merge build and CI tests are green https://ci.openmicroscopy.org/view/Bio-Formats-DEV/job/BIOFORMATS-DEV-merge-repository-subset/91/

melissalinkert:
Merging - thanks again, @avandecreme.

avandecreme:
Thanks! Glad to contribute back to bio-formats as it is a big building block of ours :)

bramalingam:
--rebased-to #2214

PR #2193, opened 2016-01-20
Remove artifactories repositories from the profiles

sbesson:
Latest Travis builds on the dev_5_1 branch have uncached dependencies and the logs show they are is primarily hitting the OME artifactory to retrieve artifacts that should be retrieved from Maven Central.

This PR amends the changes made in https://github.com/openmicroscopy/bioformats/pull/1867 to remove the OME repositories from the profiles. Is there any reason why `ome.snapshots` and `ome.unstable` should not be used for anything other than publishing?

To test this PR:
- clean your local Maven repository
- run `mvn install`
- in the log especially in the initial steps, check `artifacts.openmicroscopy.org` is only hit on purpose (i.e. for artifacts not available via Maven Central).

PR #2195, opened 2016-01-20
Remove artifactories repositories from the profiles (rebased onto develop)

sbesson:
This is the same as gh-2193 but rebased onto develop.

----

Latest Travis builds on the dev_5_1 branch have uncached dependencies and the logs show they are is primarily hitting the OME artifactory to retrieve artifacts that should be retrieved from Maven Central.

This PR amends the changes made in https://github.com/openmicroscopy/bioformats/pull/1867 to remove the OME repositories from the profiles. Is there any reason why `ome.snapshots` and `ome.unstable` should not be used for anything other than publishing?

To test this PR:
- clean your local Maven repository
- run `mvn install`
- in the log especially in the initial steps, check `artifacts.openmicroscopy.org` is only hit on purpose (i.e. for artifacts not available via Maven Central).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -718,23 +718,6 @@
     <profile>
       <id>merge-build</id>
 
-      <repositories>
-        <repository>
-          <id>ome.unstable</id>
-          <url>http://artifacts.openmicroscopy.org/artifactory/ome.unstable</url>
-        </repository>
-      </repositories>
-
-      <pluginRepositories>
-        <pluginRepository>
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #2193

melissalinkert:
The same test as in https://github.com/openmicroscopy/bioformats/pull/2193#issuecomment-173298851 works locally, so this is fine from my perspective.

PR #2196, opened 2016-01-20
TiffParser: always set the stream for any OnDemandLongArrays

melissalinkert:
This prevents exceptions when an OnDemandLongArray is loaded from a memo
file, resulting in an invalid default stream.

See https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7980

To test, convert the file from QA 16966 to .ome.btf using bfconvert.  Import the .ome.btf into OMERO; without this change, pyramid generation will throw an exception as noted in the forum thread.  With this change, pyramid generation should succeed, and the image should be viewable in OMERO.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -672,11 +672,15 @@ public class TiffParser {
 
     if (ifd.get(IFD.STRIP_BYTE_COUNTS) instanceof OnDemandLongArray) {
       OnDemandLongArray counts = (OnDemandLongArray) ifd.get(IFD.STRIP_BYTE_COUNTS);
-      counts.setStream(in);
+      if (counts != null) {
+        counts.setStream(in);
+      }
     }
     if (ifd.get(IFD.TILE_BYTE_COUNTS) instanceof OnDemandLongArray) {
       OnDemandLongArray counts = (OnDemandLongArray) ifd.get(IFD.TILE_BYTE_COUNTS);
-      counts.setStream(in);
+      if (counts != null) {
+        counts.setStream(in);
+      }
     }
 
     long[] stripByteCounts = ifd.getStripByteCounts();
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/tiff/TiffParser.java
```diff
@@ -672,15 +672,11 @@ else if (map instanceof int[]) {
 
     if (ifd.get(IFD.STRIP_BYTE_COUNTS) instanceof OnDemandLongArray) {
       OnDemandLongArray counts = (OnDemandLongArray) ifd.get(IFD.STRIP_BYTE_COUNTS);
-      if (counts != null && counts.getStream() == null) {
```
joshmoore: The `counts` field is guaranteed to not be null then? or should the `counts != null` remain?

melissalinkert: Sorry, a little over-zealous there.  ```counts != null``` checks re-added.

mtbc:
From `OMERO-DEV-merge-deploy/var/log/PixelData-0.log` we seem to be running into,
```
2016-01-25 09:31:48,762 INFO  [                ome.io.nio.PixelsService] (2-thread-2) Pyramid creation for Pixels:29751 1/1890 (0%).
2016-01-25 09:31:50,032 ERROR [ ome.services.pixeldata.PixelDataHandler] (2-thread-2) Failed to handle pixels 29751
```
https://github.com/snoopycrimecop/bioformats/commit/4337030584a398426d6a9b19e26ad05650f0028e and https://github.com/snoopycrimecop/openmicroscopy/commit/26c5e2ad69645c5f0064458c5ddbdb015e188a52 would make me think that this PR is in the build.

mtbc:
Thank you to @sbesson for much investigation of how to arrange testing of such PRs.

Using http://regions-ci.docker.openmicroscopy.org:8080/job/OMERO-build/5/artifact/src/target/ locally I now get:
```
2016-01-25 11:35:07,681 INFO  [                ome.io.nio.PixelsService] (2-thread-1) SUCCESS -- Pyramid created for pixels id:1
```
and an image that actually looks like something:
image-whole
and zooming in:
image-zoom

Good to merge.

components/formats-bsd/src/loci/formats/tiff/TiffParser.java
joshmoore: :+1:

bramalingam:
--rebased-to #2209

PR #2161, opened 2015-12-16
Unit test & fix for FilePattern

simleo:
This PR adds a new unit test for `FilePattern`. In addition, it changes the `FilePattern(String)` constructor to accept nonexistent single files (see #2131).

PR #2198, opened 2016-01-21
Fix FilePatternTest on Windows

simleo:
This should fix a problem where invalid filename characters break the test on Windows (See @sbesson's comment on #2161). I'm using the conditional here because I could not test it on a Windows box, so I'm waiting for Jenkins' opinion on the matter.


melissalinkert:
This does pass on Windows locally, so :+1:

sbesson:
From https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build-win/88/, https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/105/ and the Travis build, everything is now passing. Thanks @simleo

PR #2199, opened 2016-01-21
docs: Qu for Matlab only supports up to 4 dimensions

carandraug:
The current docs claims that qu-matlab is a toolset analysis of N-dimensional datasets. However, this is not true. On [their website](http://www.scs2.net/next/index.php?id=120):

> It allows the visualization and analysis of 4D (XYZT) datasets with any number of color channels. 

Since colour is another dimension, then technically it should be 5 dimensions. I went with "multi-channel 4 dimensional data" to reflect that qu treats the channel dimension different.


hflynn:
Formats fine, ready to merge, thanks.

bramalingam:
--rebased-to #2206

PR #2200, opened 2016-01-21
doc: install and usage instructions for Octave package (issue ome-release 177)

carandraug:
Adding instructions on how to install bioformats package in Octave as requested in openmicroscopy/ome-release#177

Note: I have added octave on the list above idl, and not in alphabetic order, because on the build the link appears as "GNU Octave" (so it is alphabetic order but of the end result).

### DIFF
--- a/docs/sphinx/users/octave/index.txt
+++ b/docs/sphinx/users/octave/index.txt
@@ -15,24 +15,25 @@ the only difference being the installation steps.
 Installation
 ------------
 
-Download `bioformats_package.jar` from the Bio-Formats
-:downloads:`downloads page <>` and place it somewhere sensible for
-your system (in Linux, this will probably be `/usr/local/share/java` or
-`~/.local/share/java` for a system-wide or user installation respectively).
+#. Download :downloads:`bioformats_package.jar <artifacts/bioformats_package.jar>`
+   and place it somewhere sensible for your system (in Linux, this will
+   probably be `/usr/local/share/java` or `~/.local/share/java` for a
+   system-wide or user installation respectively).
+#. Add `bioformats_package.jar` to Octave's *static* javaclasspath (see
+   `Octave's documentation <https://www.gnu.org/software/octave/doc/interpreter/How-to-make-Java-classes-available_003f.html>`_).
+#. Download the Octave package from the :downloads:`downloads page <>`.
+#. Start octave and install the package with::
 
-Add `bioformats_package.jar` to Octave's *static* javaclasspath (see
-`Octave's documentation <https://www.gnu.org/software/octave/doc/interpreter/How-to-make-Java-classes-available_003f.html>`_).
-
-Download the Octave package from the Bio-Formats
-:downloads:`downloads page <>`.  Start octave and install the package
-with `pkg install path-to-bioformats-octave-version.tar.gz`.
+      >> pkg install path-to-bioformats-octave-version.tar.gz
 
 Usage
 -----
 
 Usage instructions are the same as Matlab.  The only difference is that
-you need to explicitly load the package.  This is done by running
-`pkg load bioformats` at the Octave prompt.
+you need to explicitly load the package.  This is done by running at the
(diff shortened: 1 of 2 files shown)


docs/sphinx/users/octave/index.txt
```diff
@@ -0,0 +1,41 @@
...
+
+`GNU Octave <http://www.octave.org>`_ is a high-level interpreted language,
+primarily intended for numerical computations.
+Being an array programming language, it is naturally suited for image
+processing and handling of N dimensional datasets.
+Octave is distributed under the terms of the GNU General Public License.
+
+The Octave language is Matlab compatible so that programs are easily
+portable.
+Indeed, the Octave bioformats package is exactly the same as Matlab's,
```
hflynn: The MATLAB toolbox has the bioformats_package.jar bundled with it so this may be slightly misleading even though I accept that does only affect the installation.

hflynn:
Thanks! Placement is fine and page formats correctly. One minor quibble where the text is slightly confusing, and also matlab should really be written as MATLAB throughout to be consistent with our other pages. Otherwise this looks good thanks and provided @sbesson is happy with the technical content, I'm happy to merge and do nitpicky tidying myself.

sbesson:
@carandraug: tried to run the installation instructions in a lightweight Octave Docker  (https://hub.docker.com/r/schickling/octave/) and got stuck at

```
octave:2> pkg install bioformats-octave-5.1.7.tar.gz
error: the following dependencies where unsatisfied:
   bioformats needs octave >= 4.0.0
```

Two quick comments:
- the minimum GNU Octave version is certainly worth adding to the documentation page
- do you know of a Dockerfile that would ship Octave 4.0 that could be used for the lighweight testing of Octave-Bio-Formats (typicalle executing bftest.m)?

Regarding the PR itself, would it be worth adding a link to https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/developers/matlab-dev.html in the usage section and add a top-level note to the latter page mentioning these examples should apply equally to MATLAB or GNU Octave?

docs/sphinx/users/octave/index.txt
carandraug: You're right. I meant to say that the code of the package is exactly the same.

carandraug:
sbesson wrote:
> * the minimum GNU Octave version is certainly worth adding to the documentation page

Ok. I will add that.

Another requirement is that Octave must be built with java support.  On the good old style of GNU software, Octave has a thousand of build options. But my experience is that many users don't understand build options and referring it will just create confusion.

I know that at least homebrew (this seems to be the most referenced method to install Octave on MacOSX) disables java by default.


> * do you know of a Dockerfile that would ship Octave 4.0 that could be used for the lighweight testing of Octave-Bio-Formats (typicalle executing bftest.m)?

No.  Linux distributions are very good at keeping up to date with Octave releases, maybe that's way there hasn't been much need for it. I did found [one](https://github.com/mtmiller/docker-octave-dev-centos) to build development versions of Octave in CentOS (mtmiller is one of the Octave core developers).

> Regarding the PR itself, would it be worth adding a link to https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/developers/matlab-dev.html in the usage section and add a top-level note to the latter page mentioning these examples should apply equally to MATLAB or GNU Octave?

Well, I just took the lazy approach of linking to the [users/matlab](https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/users/matlab/) which then links to that. Maybe a text saying that the usage instructions are relevant to Octave in both of those pages?

sbesson:
Thanks for the last commits @carandraug. This now reads all fine to me. Leaving @hflynn to give a final sign-off on this PR before integrating it (and rebasing on `dev_5_1` I would think /cc @bramalingam )

hflynn:
Good to merge from me too, thanks.

bramalingam:
--rebased-to #2208

PR #2056, opened 2015-10-28
Modified setLength(..) method of NIOFileHandle.java for faster save operation.

Stephane-D:
Modified setLength(..) method for faster saving operation.
With the previous implementation, saving a 700 MB 5D dataset in OME TIFF format was taking about 45 minutes, now it takes about 3 minutes. It's still slow but already a nice improvement.

PR #2203, opened 2016-01-22
Modified setLength(..) method of NIOFileHandle.java for faster save operation. (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2056 but rebased onto dev_5_1.

----

Modified setLength(..) method for faster saving operation.
With the previous implementation, saving a 700 MB 5D dataset in OME TIFF format was taking about 45 minutes, now it takes about 3 minutes. It's still slow but already a nice improvement.


bramalingam:
--rebased-from #2056

bramalingam:
@melissalinkert : Can you suggest a good dataset to test this PR?

PR #2207, opened 2016-01-22
OME-TIFF export fixes

melissalinkert:
This should resolve https://trello.com/c/R1x1tmaP/78-bfconvert-writes-invalid-ome-tiff and https://trello.com/c/L0Mq3BSV/351-fill-the-creator-field-when-generating-ome-tiff.

To test, use the file mentioned in https://trac.openmicroscopy.org/ome/ticket/13131.  Run ```bfconvert vessels.ome.tiff -channel 0 c0.ome.tiff``` without this change and verify that the reported error occurs during ```showinf c0.ome.tiff```.  The same command with this change should not result in an error upon ```showinf c0.ome.tiff```.  ```tiffcomment c0.ome.tiff``` should show the following:

* 144 TiffData elements
* no TiffData elements representing C > 0
* Creator (on root OME node) set to ```OME Bio-Formats```

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -199,8 +199,6 @@ public final class FormatTools {
 
   // -- Constants - versioning --
 
-  public static final String CREATOR = "OME Bio-Formats";
-
   public static final Properties VERSION_PROPERTIES = loadProperties();
 
   /** Current VCS revision. */
@@ -220,6 +218,9 @@ public final class FormatTools {
   /** Version number of this release. */
   public static final String VERSION =
     VERSION_PROPERTIES.getProperty("release.version");
+
+  /** Value to use when setting creator/software fields in exported files. */
+  public static final String CREATOR = "OME Bio-Formats " + VERSION;
 
   public static final String PROPERTY_FILE = "version.properties";
 
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -199,6 +199,8 @@ public final class FormatTools {
 
   // -- Constants - versioning --
 
+  public static final String CREATOR = "OME Bio-Formats";
+
   public static final Properties VERSION_PROPERTIES = loadProperties();
 
   /** Current VCS revision. */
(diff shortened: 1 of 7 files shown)


components/formats-bsd/src/loci/formats/out/OMEXMLWriter.java
```diff
@@ -100,6 +102,9 @@ public void setId(String id) throws FormatException, IOException {
       xml = service.getOMEXML(retrieve);
       OMEXMLMetadata noBin = service.createOMEXMLMetadata(xml);
       service.removeBinData(noBin);
+
+      OMEXMLMetadataRoot root = (OMEXMLMetadataRoot) noBin.getRoot();
+      root.setCreator(FormatTools.CREATOR);
```
rleigh-dundee: Would it be helpful to append the release version here as well?  Might be useful for diagnostics in the future, if there's ever any version-specific problems.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -199,6 +199,8 @@
 
   // -- Constants - versioning --
 
+  public static final String CREATOR = "OME Bio-Formats";
+
```
rleigh-dundee: My earlier comment is probably better here.  Could this include the version number?

melissalinkert:
Commit pushed to append the version (currently ```5.2.0-DEV```) to the creator string.

rleigh-dundee:
The changes all look fine and good to merge, thanks.

Shall I make the corresponding change to the C++ code?

melissalinkert:
Wouldn't hurt to make similar changes in C++, but it's probably low priority as the bug only occurs when  both the input and output formats are OME-TIFF.

sbesson:
Looking at the diff, is there any reason not to rebase this fix onto `dev_5_1` @melissalinkert? /cc @bramalingam

melissalinkert:
3c798ee and 257517f are probably safe for dev_5_1, but 4aa464a does change the API of OMEXMLService which may not be the best thing for a patch release.  It fixes something of an edge case (OME-TIFF to OME-TIFF conversion only) so I'm not too worried about it being in 5.1.8, but can rewrite the commit to be non-breaking if anyone feels strongly otherwise.

rleigh-dundee:
Added https://trello.com/c/4xUsxBUY/375-ome-tiff-tiffdata-handling for C++.

sbesson:
@melissalinkert: no strong feeling from me. If needed, we know we can rebase this effort for an upcoming 5.1.x release but for now let's keep it as a 5.2.0 feature.

PR #2208, opened 2016-01-25
doc: install and usage instructions for Octave package (issue ome-release 177) (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2200 but rebased onto dev_5_1.

----

Adding instructions on how to install bioformats package in Octave as requested in openmicroscopy/ome-release#177

Note: I have added octave on the list above idl, and not in alphabetic order, because on the build the link appears as "GNU Octave" (so it is alphabetic order but of the end result).

### DIFF
--- a/docs/sphinx/users/index.txt
+++ b/docs/sphinx/users/index.txt
@@ -98,6 +98,7 @@ Numerical data processing applications
     :maxdepth: 1
     :titlesonly:
 
+    octave/index
     idl/index
     knime/index
     matlab/index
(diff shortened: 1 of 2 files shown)


bramalingam:
--rebased-from #2200

hflynn:
Should be fine to merge, we could re-run the docs build to check staging if you like but there's no reason there should be an issue with this.

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-docs/1157 was re-run with this PR included I think.

sbesson:
See http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/users/octave/index.html - merging.

hflynn:
So it was, I saw the time stamp and didn't check staging. All good.

PR #2209, opened 2016-01-25
TiffParser: always set the stream for any OnDemandLongArrays (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2196 but rebased onto dev_5_1.

----

This prevents exceptions when an OnDemandLongArray is loaded from a memo
file, resulting in an invalid default stream.

See https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7980

To test, convert the file from QA 16966 to .ome.btf using bfconvert.  Import the .ome.btf into OMERO; without this change, pyramid generation will throw an exception as noted in the forum thread.  With this change, pyramid generation should succeed, and the image should be viewable in OMERO.


bramalingam:
--rebased-from #2196

bramalingam:
works as expected. Pyramid generation works and image opens without any issues on develop.

PR #2210, opened 2016-01-25
add folders to model

mtbc:
Adds a new model object `Folder` which may contain Images and ROIs and may itself be in a folder.

--no-rebase


mtbc:
Largely tested via https://github.com/openmicroscopy/openmicroscopy/pull/4441.

dgault:
Having carried out some testing the model changes all look good to me. Builds also appear to be stable.

PR #2211, opened 2016-01-25
Fixing nifti links

hflynn:
http://nifti.nimh.nih.gov/nifti-1/data and http://nifti.nimh.nih.gov/nifti-1/ have been down since last week and it seems the sysadmin has left so unclear if/when they will work again, so replacing with working links.

Should make https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/ green again.

PR #2212, opened 2016-01-25
Fixing nifti links (rebased onto dev_5_1)

hflynn:
This is the same as gh-2211 but rebased onto dev_5_1.

----

http://nifti.nimh.nih.gov/nifti-1/data and http://nifti.nimh.nih.gov/nifti-1/ have been down since last week and it seems the sysadmin has left so unclear if/when they will work again, so replacing with working links.

Should make https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/ green again.


hflynn:
--rebased-from #2211

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-docs/1166/ :+1:

PR #2214, opened 2016-01-26
Fix FEITiffReader stage position parsing. (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2194 but rebased onto dev_5_1.

----

I also modified the end element code so that even if the string builder is empty, the parentNames get peeked (to be consistent with the code before #2181 ).


bramalingam:
--rebased-from #2194

bramalingam:
All tests are passing here : https://ci.openmicroscopy.org/view/Bio-Formats-5.1/job/BIOFORMATS-5.1-merge-repository-subset/218/

PR #2217, opened 2016-01-27
Fixed SlideBook reader dlls for montages

RichardMyers:
fixed bug with xy position reading for montages


melissalinkert:
--rebased-to #2219

PR #2218, opened 2016-01-28
developers/examples/bftest.m: simplify by using Matlab and Octave core functions

carandraug:
This simplifies the creation of a directory on the tmporary directory by making use of Matlab and Octave core functions instead of the java stuff. While this makes it much simpler, the main reason is Octave compatibility since the example uses Matlab dot notation to call java methods that hasn't been implemented in Octave yet.

With this change, the example runs fine in Octave until the call to im2frame (again, a missing feature in Octave).

### DIFF
--- a/docs/sphinx/developers/examples/bftest.m
+++ b/docs/sphinx/developers/examples/bftest.m
@@ -1,9 +1,7 @@
 bfCheckJavaPath();
 
 % Create local file for testing
-java_tmpdir = char(java.lang.System.getProperty('java.io.tmpdir'));
-uuid = char(java.util.UUID.randomUUID());
-tmpdir = fullfile(java_tmpdir, uuid);
+tmpdir = tempname();
 mkdir(tmpdir);
 
 % Create fake file for testing purposes
@@ -54,8 +52,7 @@ imagesc(series1_plane1);
 % displaying-images-end
 
 % animated-movie-start
-v = linspace(0, 1, 256)';
-cmap = [v v v];
+cmap = gray(256);
 for p = 1 : size(series1, 1)
   M(p) = im2frame(uint8(series1{p, 1}), cmap);
 end


sbesson:
Thanks @carandraug, including for review tomorrow.

ximenesuk:
The changes look fine in MATLAB - valid `tmpdir` and `cmap` are created and the script completes. I've not tested in Octave. If that needs doing I will try later otherwise this is good to merge.

carandraug:
> I've not tested in Octave. If that needs doing I will try later otherwise this is good to merge.

It still does not run to the end in Octave because Octave does not have `im2frame`. But with this change will at least run the first half of examples in Octave while also simplifying the code even for Matlab users.

sbesson:
Thanks @carandraug. Looks good as a first step towards a MATLAB/Octave compatible `bftest.m` example.

carandraug:
I have noticed that in some cases, you already make use of [`javaObject()`](https://github.com/carandraug/bioformats/blob/9380a94e941a82c59a2c579899633f34528252f0/docs/sphinx/developers/examples/bftest.m#L163) instead of dot notation. Will a patch that makes use of it acceptable? That would make it Octave compatible.

Also, this example makes use of a function named `feature()`. Indeed, this seems to be a Matlab function. However, it's undocumented. It doesn't show up on Matlab's online documentation and it has no help text whatsoever. This causes problems because Octave will not implement it. Being undocumented, I'm unsure if bioformats should even make use of it in the first place.

sbesson:
> I have noticed that in some cases, you already make use of [`javaObject()`](https://github.com/carandraug/bioformats/blob/9380a94e941a82c59a2c579899633f34528252f0/docs/sphinx/developers/examples/bftest.m#L163) instead of dot notation. Will a patch that makes use of it acceptable? That would make it Octave compatible.

I assume you mean a patch to use `javaObject()` across the board in the `bftest.m` example. If so, no objections from me at least.

> Also, this example makes use of a function named `feature()`. Indeed, this seems to be a Matlab function. However, it's undocumented. It doesn't show up on Matlab's online documentation and it has no help text whatsoever. This causes problems because Octave will not implement it. Being undocumented, I'm unsure if bioformats should even make use of it in the first place.

The main reason for having this function is that it is the best option I know of to determine whether MATLAB is running in an interactive session. Effectively it allows the daily CI job to execute the script without crashing.

carandraug:
>> Also, this example makes use of a function named feature(). Indeed, this seems to be a Matlab function. However, it's undocumented. It doesn't show up on Matlab's online documentation and it has no help text whatsoever. This causes problems because Octave will not implement it. Being undocumented, I'm unsure if bioformats should even make use of it in the first place.
>
> The main reason for having this function is that it is the best option I know of to determine whether MATLAB is running in an interactive session. Effectively it allows the daily CI job to execute the script without crashing.

I didn't realize this can be an issue with Matlab code. The suggestion I have is to use [`xvfb-run`](http://manpages.ubuntu.com/manpages/lucid/man1/xvfb-run.1.html).

The closest I found to documentation of `feature` was on the [Undocumented Matlab blog](http://undocumentedmatlab.com/blog/undocumented-feature-function) and even they warn against using it " [...] actually does often change without prior notice between Matlab releases, so be very careful when using this function in your code."

PR #2219, opened 2016-01-28
Fixed SlideBook reader dlls for montages (rebased onto develop)

melissalinkert:
This is the same as gh-2217 but rebased onto develop.

----

fixed bug with xy position reading for montages


melissalinkert:
--rebased-from #2217

melissalinkert:
Merging as discussed with @sbesson.

PR #2221, opened 2016-01-28
Fix TiffRational integer overflow

melissalinkert:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-January/005834.html

To test, use the ```myfile.tif``` file attached to the thread.  Without this change, ```showinf -nopix``` should confirm that the ```XResolution``` and ```YResolution``` keys in the original metadata table are both 0.  ```tiffinfo``` on the same file should show ```  Resolution: 42.42, 0 (unitless)``` as mentioned in the thread.

With this change, ```showinf -nopix``` should show ```42.419...``` and ```0.0``` for ```XResolution``` and ```YResolution``` respectively.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/BaseTiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/BaseTiffReader.java
@@ -230,8 +230,16 @@ public abstract class BaseTiffReader extends MinimalTiffReader {
     putInt("Model", firstIFD, IFD.MODEL);
     putInt("MinSampleValue", firstIFD, IFD.MIN_SAMPLE_VALUE);
     putInt("MaxSampleValue", firstIFD, IFD.MAX_SAMPLE_VALUE);
-    put("XResolution", firstIFD.getIFDRationalValue(IFD.X_RESOLUTION).doubleValue());
-    put("YResolution", firstIFD.getIFDRationalValue(IFD.Y_RESOLUTION).doubleValue());
+
+    TiffRational xResolution = firstIFD.getIFDRationalValue(IFD.X_RESOLUTION);
+    TiffRational yResolution = firstIFD.getIFDRationalValue(IFD.Y_RESOLUTION);
+
+    if (xResolution != null) {
+      put("XResolution", xResolution.doubleValue());
+    }
+    if (yResolution != null) {
+      put("YResolution", yResolution.doubleValue());
+    }
 
     int planar = firstIFD.getIFDIntValue(IFD.PLANAR_CONFIGURATION);
     String planarConfig = null;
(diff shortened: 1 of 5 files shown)


melissalinkert:
Commits have been pushed to resolve https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-performance/102/ and https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-test_images_good/68/.  Will keep an eye on jobs over the weekend for any further failures.

rleigh-dundee:
The changes look fine and testing shows it works correctly and is good to merge.  Two comments:

- Rather than putting `& 0xffffffffL` in multiple times, would it not be cleaner to add a `readUnsignedInt()` method to `RandomAccessInputStream` so it can be used in other places as well?  Maybe the same for `short` and `long` as well?
- Do we have any unit or integration tests?  (Will this file be added?)

melissalinkert:
Updated to use a new ```readUnsignedInt()``` method instead of ```& 0xffffffffL``` everywhere in TiffParser.  The same problem was encountered with ```LONG```-type fields when investigating https://trac.openmicroscopy.org/ome/ticket/13119, so that's been fixed now too.

Config PR forthcoming for the above file and QA 16912.  This will need re-testing per instructions above as well as confirmation that the image read from QA 16912 now matches what is shown by other software (instead of being all black).

melissalinkert:
See https://github.com/openmicroscopy/data_repo_config/pull/83

bramalingam:
Works as expected (both : myfile.tif and QA16912). Good to merge.

sbesson:
The last commits included in this PR caused the repository builds [79](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/79/) and [80](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/80/) to hang while parsing some `hamamatsu` files. This was confirmed independently by running the repository subset job [with](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/114) and [without](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/115/) this PR included. Excluding this PR and aborting the full-repository job.

melissalinkert:
Commit pushed to fix infinite loop and exclude label removed - see https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/123/

melissalinkert:
Re-including after investigation of ```tiff/samples/libtiff/g3test.tif```.  That file exhibited the same bug as in https://trac.openmicroscopy.org/ome/ticket/13119: the ```RowsPerStrip``` field was large enough to overflow, so without this change the number of strips was negative, so no strips were decoded, the image was all black, and the logic to check for an unsupported compression type was not encountered.  With this change, the strips are detected correctly and decompression is attempted, thus the exception in the tests.  https://github.com/openmicroscopy/data_repo_config/pull/83 has been updated to disable testing of this file; we'll need to reconfigure it if/when we add support for the appropriate codec.

bramalingam:
The basic test on "MyFile.tif" works as expected. Looks good to merge if the config issues for the tests are going to be handled separately.

melissalinkert:
Merging then, since this and associated config PRs have been in for several days and consistently passing (e.g. https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-full-repository/89/).

PR #2224, opened 2016-02-01
Micromanager: parse JSON tag in each TIFF (rebased onto dev_5_1)

sbesson:
This is the same as gh-2213 but rebased onto dev_5_1.

----

See https://trello.com/c/499UGA5U/72-micromanager-stage-positions-and-json

This reads the JSON data from each TIFF and stores the key/value pairs
in the original metadata table.  Stage position values ("*PositionUm")
will also be stored in OME-XML, if they are present.

As each TIFF tag contains a single JSON object, simple parsing of the
key/value pairs without an external library was the easiest solution for now.
We'll likely want to revisit this once we have a better plan for dealing
with JSON data in general though.

To test, use the dataset in ```curated/micromanager/nico/Untitled_4_multi_file``` (config PR forthcoming).  Without this change, ```showinf -omexml``` should not show PositionX, PositionY, or PositionZ values for Plane in the OME-XML.  With this change, all 3 values should be set for every Plane, and the original metadata table should show many entries beginning with ```Plane```.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -445,13 +445,22 @@ public class MicromanagerReader extends FormatReader {
             // using key alone will result in conflicts with metadata.txt values
             addSeriesMeta(String.format("Plane #%0" + digits + "d %s", plane, key), value);
             if (key.equals("XPositionUm")) {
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #2213

bramalingam:
Works as expected. Good to merge.

PR #2225, opened 2016-02-02
Refactor FakeReaderTest

simleo:
This PR simplifies `FakeReaderTest` and makes sure temporary files and directory trees are created only when necessary (except for the root working directory, which is used for the global teardown).

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/FakeReaderTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/FakeReaderTest.java
@@ -37,9 +37,7 @@ import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
 
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.PrintWriter;
-import java.io.RandomAccessFile;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


sbesson:
Sorry for the delay. Overall, tests pass on Linux/Windows, the changes read well and certainly simplify both the test setup and the addition of new tests for each modification of the `FakeReader`. I see that the multi-channel fake file test has been removed in the operation.  Leaving @melissalinkert to have a final look but this seems good to merge.

Would the next step on this test class to build upon the recent fake improvements/documentation efforts made by @ximenesuk and add unit tests for each key supported by the reader before adding new functionalities?

simleo:
@sbesson yes, I've removed that test because it's essentially a duplicate of `testValuesFromFilename`, with `sizeC` instead of `sizeX` as the non-default parameter

melissalinkert:
Only minor comment is that you might consider using ```loci.common.IniWriter``` when creating the ```.ini``` files.  For the current usage it's probably excessive, but if any future tests here need to create more complex ```.ini``` files then it's something to keep in mind.  Merging though, since this is all fine and working.

PR #2227, opened 2016-02-03
getUsedFiles performance fix

simleo:
This PR addresses a performance issue in `FormatReader.getUsedFiles` that can get very serious for formats with a large number of files in multiple series (e.g., `ScreenReader`).

The current implementation iteratively updates a `Vector` with the contents of arrays returned from `getSeriesUsedFiles`. To avoid duplicates, each filename from `getSeriesUsedFiles` is checked to see if it already belongs to the vector before adding. The problem is that `Vector.contains` is O(n), which makes the whole update operation **quadratic**.

A previous attempt at fixing the problem (see #1028) only made things better for single-series images: updates for subsequent series are still performed in the same way.

On my laptop, `ScreenReader.getUsedFiles` on [a TARA screen file](https://github.com/IDR/idr-metadata/blob/master/idr0015-UNKNOWN-taraoceans/screenA/patterns/TARA_HCS1_H5_G100001472_G100001473--2013_09_28_19_45_25_chamber--U01--V01.screen) takes ~60 seconds without the patch and ~3 seconds with the patch. Note that I've replaced `.ome.tif` entries with `.fake` ones to avoid copying real data.

This problem is also present in develop. I'm opening the PR vs metadata first because `ScreenReader`  allows to better evaluate the performance improvement.

NOTE: I have used `LinkedHashSet` instead of `HashSet` for predictable iteration order.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1043,16 +1043,22 @@ public abstract class FormatReader extends FormatHandler
   /* @see IFormatReader#getUsedFiles() */
   @Override
   public String[] getUsedFiles(boolean noPixels) {
-    if (getSeriesCount() == 1) {
-      return getSeriesUsedFiles(noPixels);
+    String[] seriesUsedFiles;
+    int seriesCount = getSeriesCount();
+    if (seriesCount == 1) {
+      seriesUsedFiles = getSeriesUsedFiles(noPixels);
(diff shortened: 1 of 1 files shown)


mtbc:
Code changes look good.

simleo:
See https://trello.com/c/MDPDqacz/148-test-suite-is-breaking-travis re: Travis failure

joshmoore:
Deployed to the idrstaging server as build 66. Merging to be tagged for 0.0.4.

PR #2226, opened 2016-02-03
Micromanager: *_metadata.txt and 51123 tag parsing

melissalinkert:
This fixes several issues with the new datasets from Thomas Julou - see https://github.com/openmicroscopy/data_repo_config/pull/76 for configuration and gh-2213 for discussion.

For each of the four new datasets, check that ```showinf``` on the ```*metadata.txt``` file shows images without an error (with this change).  Without this change, ```showinf``` on ```test_sep/Pos0/metadata.txt``` should work, but all others will throw an exception.

For ```test2_sep```, ```test_stack```, and ```test2_stack```, check each image (this may be easier in ImageJ) and verify that each image is unique and not blank.

For ```test_stack``` and ```test2_stack```, the 51123 tag should be parsed for each plane.  Compare the original metadata tables with and without this change; there should be several new ```Plane #...``` keys with this change.

PR #2229, opened 2016-02-04
getUsedFiles performance fix (develop)

simleo:
Same as #2227, but rebased onto develop.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1043,16 +1043,22 @@ public abstract class FormatReader extends FormatHandler
   /* @see IFormatReader#getUsedFiles() */
   @Override
   public String[] getUsedFiles(boolean noPixels) {
-    if (getSeriesCount() == 1) {
-      return getSeriesUsedFiles(noPixels);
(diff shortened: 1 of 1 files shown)


joshmoore:
NB: :+1: on having this on develop for more thorough testing. Once that's done, I'd propose opening against `dev_5_1` as a candidate for 5.1.9 (in which case the `metadata` might can be closed)

sbesson:
See the FileStitcher failures on https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/106/. Excluding for now to test #2226

simleo:
Now it should be OK. The problem was in the first section, where I was trying to avoid the useless conversion to set and back to array for the single-series case. Returning `getSeriesUsedFiles`'s result directly does not work because it can be `null`, while `getUsedFiles` is expected to always return an array (possibly empty). Is this difference between the two methods documented anywhere?

melissalinkert:
All looks good and builds are back to normal, so merging this.

The potential for returning null is not documented, unfortunately.  I'd be inclined to clarify the Javadocs so that ```null``` is not allowed, and update any misbehaving readers to return an empty array instead - happy to discuss though.

PR #2232, opened 2016-02-05
Micromanager: *_metadata.txt and 51123 tag parsing (rebased onto dev_5_1)

melissalinkert:
This is the same as gh-2226 but rebased onto dev_5_1.

----

This fixes several issues with the new datasets from Thomas Julou - see https://github.com/openmicroscopy/data_repo_config/pull/76 for configuration and gh-2213 for discussion.

For each of the four new datasets, check that ```showinf``` on the ```*metadata.txt``` file shows images without an error (with this change).  Without this change, ```showinf``` on ```test_sep/Pos0/metadata.txt``` should work, but all others will throw an exception.

For ```test2_sep```, ```test_stack```, and ```test2_stack```, check each image (this may be easier in ImageJ) and verify that each image is unique and not blank.

For ```test_stack``` and ```test2_stack```, the 51123 tag should be parsed for each plane.  Compare the original metadata tables with and without this change; there should be several new ```Plane #...``` keys with this change.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -180,7 +180,7 @@ public class MicromanagerReader extends FormatReader {
       }
       if (!noPixels) {
         for (String tiff : pos.tiffs) {
-          if (new Location(tiff).exists()) {
+          if (new Location(tiff).exists() && !files.contains(tiff)) {
             files.add(tiff);
           }
         }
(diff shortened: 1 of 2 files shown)


melissalinkert:
--rebased-from #2226

sbesson:
See https://ci.openmicroscopy.org/view/Bio-Formats-5.1/job/BIOFORMATS-5.1-merge-repository-subset/223/

bramalingam:
Works as expected. Good to merge.

PR #2234, opened 2016-02-09
Fix spider format links

hflynn:
The spider docs have been moved. This should make all the BF docs builds green again once rebased.

PR #2235, opened 2016-02-09
Fix spider format links (rebased onto dev_5_1)

hflynn:
This is the same as gh-2234 but rebased onto dev_5_1.

----

The spider docs have been moved. This should make all the BF docs builds green again once rebased.


hflynn:
--rebased-from #2234

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-docs/1183/ :+1:

PR #2239, opened 2016-02-10
Update copyright headers

snoopycrimecop:
Repository: openmicroscopy/bioformats
Already up-to-date.

Generated by build [BIOFORMATS-5.1-latest-copyright#352](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-latest-copyright/352/).

----
--no-rebase

PR #2240, opened 2016-02-10
Fix copyrights missed by update_copyright

melissalinkert:
Bumps the year to 2016, and fixes text so that ```update_copyright``` can pick these up in the future.

See https://github.com/openmicroscopy/bioformats/pull/2239#issuecomment-182476425


melissalinkert:
Pushed one more set of fixes noticed by @sbesson.  The ```cpp``` changes may need further investigation after 5.1.8, as I'm not entirely sure why those files aren't caught by ```update_copyright``` (visually they seem to match the regex).

sbesson:
:+1: Let's get this merged and investigate the copyright issues on `develop`.

PR #2236, opened 2016-02-10
Adding metadata file notes to Micro-Manager

hflynn:
Explains new Micro-Manager metadata.txt file handling on MM file format page and in the dataset structure table (will need an autogen docs build run to generate table changes in the docs I think).

Sample links still be added.

PR #2241, opened 2016-02-11
Adding metadata file notes to Micro-Manager (rebased onto dev_5_1)

hflynn:
This is the same as gh-2236 but rebased onto dev_5_1.

----

Explains new Micro-Manager metadata.txt file handling on MM file format page and in the dataset structure table (will need an autogen docs build run to generate table changes in the docs I think).

Sample links still be added.

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -434,6 +434,12 @@ presenceRating = Fair
 utilityRating = Good
 reader = MicromanagerReader
 mif = true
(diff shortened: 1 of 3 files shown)


hflynn:
--rebased-from #2236

sbesson:
Looks good both on the staging pages and the [autogenerated documentation commit](https://github.com/snoopycrimecop/bioformats/commit/94b97cd033ac887b80bf388f0869c1e2ced0a613). Merging.

PR #2249, opened 2016-02-16
Debug registered Kryo classes

joshmoore:
In order to pre-register most if not all required Bio-Formats
classes so that bfmemo files are stable between JVM restarts,
we print out classes during Memoization. Output from the full
repo job as well as OMERO should be combined together.

See:

 * https://github.com/romix/akka-kryo-serialization/issues/16#issuecomment-24979653
 * https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/serializer/KryoSerializer.scala#L82
 * http://stackoverflow.com/questions/12277193/automatic-class-registration-in-kryo

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -58,6 +58,7 @@ import org.slf4j.LoggerFactory;
 
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.KryoException;
+import com.esotericsoftware.kryo.Registration;
 import com.esotericsoftware.kryo.io.Input;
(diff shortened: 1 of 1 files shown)


joshmoore:
```
java.lang.ClassCastException: java.lang.String cannot be cast to loci.formats.IFormatReader
	at loci.formats.Memoizer$KryoDeser.loadReader(Memoizer.java:186) ~[classes/:5.2.0-SNAPSHOT]
	at loci.formats.Memoizer.loadMemo(Memoizer.java:890) [classes/:5.2.0-SNAPSHOT]
	at loci.formats.Memoizer.setId(Memoizer.java:677) [classes/:5.2.0-SNAPSHOT]
```

Not sure what's going on here. I'll take a look tomorrow.

PR #2250, opened 2016-02-16
Zeiss TIFF: fix getSeriesUsedFiles to return actual paths on disk

melissalinkert:
CaseInsensitiveLocation.exists() can (by design) return true even if the
path from which it was constructed does not actually exist.  This fixes
ZeissTIFFReader to report the CaseInsensitiveLocation's path after
lookup, rather than the path passed to the constructor.

This should fix intermittent test failures, see
https://trello.com/c/EWS9dntP/101-zeiss-zvi-improve-tile-handling

This will be difficult to completely verify, but the test failures noted on the card do seem to correlate with abnormally long build execution times (~20 hours vs. ~14-16).  It might be easiest to just let this sit for a few weeks to see what happens.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissTIFFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissTIFFReader.java
@@ -217,18 +217,18 @@ public class ZeissTIFFReader extends BaseZeissReader {
     ArrayList<String> files = new ArrayList<String>();
 
     try {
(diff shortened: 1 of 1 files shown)


sbesson:
Except for three isolated failures due to `OutOfMemoryException`, the data repository automated test builds have been consistently passing since [build 86](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/86/). Leaving this PR opened until the end of this working week, then I am planning to integrate it if no related test failure arises.

PR #2248, opened 2016-02-16
Enforce UTF-8 encoding in FakeReaderTest

melissalinkert:
This forces INI files written by ```FakeReaderTest``` to be UTF-8, even on Windows.  As noted in the commit message for 2f80775, this is consistent with the rest of our I/O stack, which enforces UTF-8 for both reading and writing across the board.  To validate the fix, the angstrom case has been re-introduced in ```FakeReaderTest```.

See https://trello.com/c/qTAgtVHT/99-fakereader-add-unicode-units-tests

PR #2252, opened 2016-02-17
Specify the encoding (UTF-8) for fake.ini files

joshmoore:
In order to safely allow encoding unit symbols like
angstroms and micrometers cross-platform, fake.ini
files are now *required* to be UTF-8.

See: https://github.com/openmicroscopy/bioformats/pull/2248


chris-allan:
:+1:

PR #2253, opened 2016-02-18
SDT: fix width padding calculation for single-pixel image

melissalinkert:
See https://trello.com/c/57Xzqhzz/95-sdtreader-issue-in-openbytes

To test, use the file from QA 17017.  Without this change, ```showinf``` on the file will throw an exception as noted in the Trello card.  With this change, the entire (single pixel) image should be read without an error.  A configuration PR for QA 17017 is forthcoming.

/cc @imunro for validation with previously-working version of FLIMfit

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SDTReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SDTReader.java
@@ -150,12 +150,23 @@ public class SDTReader extends FormatReader {
     int bpp = FormatTools.getBytesPerPixel(getPixelType());
     boolean little = isLittleEndian();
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
See https://github.com/openmicroscopy/data_repo_config/pull/82

imunro:
Tested with FLIMfit and all seems to work!
I've never been really clear under what conditions  paddedWisth might not be == to sizeX but there doesn't seem any potential for the code changes to introduce bugs so I'm happy to merge.

melissalinkert:
Merging then, as all current test failures are accounted for by other PRs.

bramalingam:
--rebased-to #2291

PR #2256, opened 2016-02-20
Remove "Is SPW file" global metadata stored when reading OME-XML files

sbesson:
The "Is SPW file" global metadata was added as part of 2e9d34c9f74213c94ed3c6df3584f08a6c64c87c but has unexpected impact when upgrading OME-XML samples.


mtbc:
This looks good to me. :+1: Regions CI is all blue too.

melissalinkert:
I'm a little concerned that this is necessary - could this minimally be documented somewhere in the reader so that we don't add something else breaking to the original metadata table?

melissalinkert:
Merging after discussion with @sbesson - I misunderstood what the original problem was, so ignore my above comment.

PR #2259, opened 2016-02-24
Updating Improvision format links

hflynn:
This should make the latest docs green again.

The current problem is actually just that their redirect is not working properly but given that Improvision doesn't exist as a separate entity any more, I figured best to just update the links anyway. Also removed the Openlab links as although they weren't breaking the build, the content doesn't exist any more, it was just redirecting to a 'We're sorry' page and searching for Openlab on the main PE page gives no useful results.


sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-docs/ is green.

bramalingam:
--rebased-to #2293

PR #2262, opened 2016-02-28
Micromanager: clean up JSON metadata parsing

melissalinkert:
See QA 17040.  This should correctly parse various types of arrays and
keys with child attributes.  As noted previously, this should eventually be replaced
by use of a real JSON library.

To test, use the *.txt files from QA 17040.  Without this change, as noted in the comments for QA 17040:

```
You should see lines like this in the metadata display:
Plane #0 08 39 -0800
Plane #0 ] Camera-SaturatePixels
Plane #1 08 40 -0800
Plane #1 ] Camera-SaturatePixels
et cetera. 
```

With this change, none of the metadata keys should contain ```]``` or ```[``` or be just numbers, the ```Camera-SaturatePixels``` key should be present for every plane, and in general there shouldn't be anything that looks odd.  It's also possible to double-check against the actual JSON strings by turning on TRACE-level debugging, but that requires some patience as the JSON strings do not contain line breaks and are thus a little tricky to read as-is.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -497,32 +497,55 @@ public class MicromanagerReader extends FormatReader {
           }
           String[] tokens = json.split("[\\{\\}:,\"]");
           String key = null, value = null, propType = null;
-          for (String token : tokens) {
+          int nEmptyTokens = 0;
+          for (int q=0; q<tokens.length; q++) {
+            String token = tokens[q];
             if (token.length() == 0) {
+              nEmptyTokens++;
               continue;
             }
+            if (nEmptyTokens == 5 && value == null) {
+              key = null;
+            }
             if (key == null && value == null && propType == null) {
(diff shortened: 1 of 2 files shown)


bramalingam:
Works as expected. Looks good to merge.

sbesson:
The QA file was properly tested as part of https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/89/. @bramalingam: which datasets/MM versions were tested as part of this  PR review?

bramalingam:
I had just tested the QA file alone, based on the testing instructions. Do you want me to try other datasets as well?

sbesson:
@bramalingam: I would definitely check the impact of this PR minimally on current (1.4.22) and even former Micro-Manager datasets if we have any.

bramalingam:
Tested with MM1.4.21, 1.4.22 and one dataset from 1.4.15. In all cases, the test passed without any issues.

bramalingam:
--rebased-to #2294

PR #2264, opened 2016-02-29
cpp: Update naming scheme

rleigh-dundee:
See https://github.com/ome/ome-files-cpp/pull/2 for testing instructions.


cpp/lib/ome/xml/module.cpp
```diff
@@ -54,7 +54,7 @@ namespace
       ("ome-xml-data",
        "OME_XML_PKGDATADIR",
        "OME_XML_HOME",
-       "BIOFORMATS_HOME",
+       "OME_FILES_HOME",
```
joshmoore: The use of `OME_FILES` here seems somewhat like a leaking concern if this will eventually migrate to `ome-model` (or similar), but that is possibly only temporary.

rleigh-dundee: It's not intended to be.  `OME_XML_HOME` would be renamed to `OME_MODEL_HOME` if we name the component.  This is the installation root of that component.  `OME_FILES_HOME` is the installation root of the whole collection, i.e. when we use OME_FILES to refer to the overarching release name.

The purpose of the two names is to allow overriding of the installation root for an individual component, for example for testing purposes, while keeping the same main installation root for everything else.  We already use this at a finer-grained level in the super build, where we override individual data directories e.g. schema and transform directories in the source tree.

This is more of a problem in ome-files, where we use the same environment variable to refer to the component and the whole collection of components.  This is one aspect of what was commented on in the other PR.  This is another case of having the same name for two different things being quite problematic.

joshmoore:
Merging. This may require another B-F milestone (`M3` or `M2.1`) which we can discuss next week.

PR #2274, opened 2016-03-08
FluoviewReader - Null Pointer Fix

dgault:
Issue raised by QA-17055

If MetadataOptions has the MetadataLevel set to minimum then the
readStamps function will not be called and stamps will be equal to null.

This results in the following NPE:
	at loci.formats.in.FluoviewReader.initStandardMetadata(FluoviewReader.java:328)
	at loci.formats.in.BaseTiffReader.initMetadata(BaseTiffReader.java:98)
	at loci.formats.in.BaseTiffReader.initFile(BaseTiffReader.java:577)

To reproduce:
- Import the sample file from QA-17055 via Omero Insight 5.2.2
- Verify that the same Null Pointer Exception is reproduced

To test:
- Update the formats-gpl lib from this branch
- Re import the sample file
- Verify that the file imports sucessfully

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FluoviewReader.java
@@ -324,7 +324,7 @@ public class FluoviewReader extends BaseTiffReader {
         }
         
         ArrayList<Double> uniqueZ = new ArrayList<Double>();
-        if (i > 1) {
+        if (i > 1 && stamps != null) {
           zPositions = stamps[i - 2];
           if (zPositions != null) {
             for (Double z : zPositions) {


mtbc:
I can reproduce the above exception using today's OMERO-DEV-latest artifacts.

mtbc:
Checked via regions-ci build. Good to merge.

melissalinkert:
Does QA 17055 need to be added to ```curated```, or would this have been reproducible with an existing file?  I wonder if it also makes sense to add MetadataLevel data tests (in a future PR), so that this sort of problem is caught earlier on (see https://trello.com/c/o2tgUIb2/381-metadatalevel-data-tests).

dgault:
The issue would have been recreated with a number of files already in our sample set. It would be no harm to add the QA 17055 file too though.

MetadataLevel tests would certainly have helped to catch this issue. If we have a card for it then it would certainly be useful to add at some point.

sbesson:
Given the state of https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/141/console, merging

bramalingam:
--rebased-to #2296

PR #2276, opened 2016-03-10
Gatan: fix physical size parsing and use native units

sbesson:
See https://trello.com/c/11hj6Ixk/47-fix-gatan-dm4-metadata-parsing.

This PR:
- fixes the erroneous metadata parsing reported in https://github.com/openmicroscopy/bioformats/pull/1979#issuecomment-140699287 by excluding the `Scale` key/value pairs associated with `Calibration` groups - the physical sizes of the QA file mentioned in the PR should be 0.01255um in X and Y and 0.05um in Z
- stores the physical sizes using native units instead of converting unilaterally to micrometers

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -483,12 +483,12 @@ public class GatanReader extends FormatReader {
       if (value != null) {
         addGlobalMeta(labelString, value);
 
-        if (labelString.equals("Scale")) {
+        if (labelString.equals("Scale") && !parent.equals("Calibration")) {
           if (value.indexOf(",") == -1) {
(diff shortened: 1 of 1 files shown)


sbesson:
Excluding as a configuration PR will need to be opened to fix the tests against the Gatan samples files

dgault:
Tested with datasets containing both nanometers and micrometers and both are displaying the physical pixel size units correctly. Excluding the Scale value for Configuration sets appears to be correct following from the feedback on https://github.com/openmicroscopy/bioformats/pull/1979#issuecomment-140699287

Gatan tests are green following from the configuration PR
https://ci.openmicroscopy.org/view/Bio-Formats-DEV/job/BIOFORMATS-DEV-merge-repository-subset/148/

Im happy for both this and the configuration PR to be merged

bramalingam:
--rebased-to #2299

PR #2278, opened 2016-03-14
Fixing alicona link

hflynn:
This should fix https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/171/warnings3Result/

PR #2279, opened 2016-03-14
Fixing alicona link (rebased onto dev_5_1)

hflynn:
This is the same as gh-2278 but rebased onto dev_5_1.

----

This should fix https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/171/warnings3Result/


hflynn:
--rebased-from #2278

PR #2280, opened 2016-03-14
Refactor bump_maven_version.py for setting groups

joshmoore:
Originally opened as https://github.com/openmicroscopy/bioformats/pull/2158
this modifies the bump_maven_version script to allow changing the groupId
property. Use of a separate groupId prevents confusing version re-use in
the local repository, which is especially hard to detect in the case of
SNAPSHOTs.

To test:
 * Try a command like `tools/bump_maven_version.py --old-group=ome --new-group=idr 0.0.3-SNAPSHOT` and see if your regular dev workflow continues to work.

### DIFF
--- a/tools/bump_maven_version.py
+++ b/tools/bump_maven_version.py
@@ -1,6 +1,7 @@
 #! /usr/bin/python
 # Script for increasing versions numbers across the code
 
+import sys
 import glob
 import re
 import argparse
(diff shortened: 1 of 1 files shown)


sbesson:
Tested using various combinations of versions (`SNAPSHOT` and non `SNAPSHOT`), old and new groups. All works as expected. Unit tests are likely overkill for this utility script used prior to release only although if you have any idea on how to implement this without too much infrastructure, happy to bite the bullet. Merging.

joshmoore:
Probably only if we periodically (daily) spin up a [devspace](https://github.com/openmicroscopy/devspace). Then we could run the script with a unique group name before kicking off `Trigger`.

PR #2283, opened 2016-03-15
FakeReader: fix close() and reopenFile()

sbesson:
See https://trello.com/c/XkVbBZ2X/111-fakereader-reset-variables-when-reader-is-closed and https://trello.com/c/IdLyg8yn/89-fakereader-reopenfile-fails-with-filenotfoundexception

The unit tests additions should cover these test cases so checking Travis and the daily builds should be sufficient. To be thorough, a simple workflow as follows could be tested and should not raise any exception:

```
IFormatReader reader = new FakeReader();
reader.setId("foo&points=10.fake");
reader.reopenFile();
reader.close();
reader.setId("foo.fake");
```

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -769,6 +769,10 @@ public class FakeReader extends FormatReader {
     }
   }
 
+  @Override
+  public void reopenFile() throws IOException {
+  }
+
   private void fillPhysicalSizes(MetadataStore store) {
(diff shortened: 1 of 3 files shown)


dgault:
FakeReader and Memoizer unit tests are all running successfully. I produced some sample code for the FakeReader similar to that suggested and the reopen command no longer throws an exception and defaults are reset after closing the reader.

My only question would be if other values such as scaleFactor and lut etc also need to reset?

sbesson:
@dgault: it is certainly sensible to reset any value that is conditionally initialized depending on the key/value pairs. Pushed.

dgault:
Tests are all still green so looks good to merge

PR #2284, opened 2016-03-16
ND2: fix detection of planes with interleaved metadata blocks

melissalinkert:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-March/005868.html

To test, use the file referenced in https://github.com/openmicroscopy/data_repo_config/pull/93.  Verify that without this change, ```showinf``` produces an exception as shown in the thread.  With this change, 60 series should be present, each with a single plane.  The images and metadata can be compared with Nikon's NIS Elements viewer.


dgault:
The code changes make sense to me and look good. Testing verified that prior to applying this PR that an exception was thrown as described in the user mail. With the new changes the file opens correctly, showing 60 series, images open correctly and look good, as does the metadata. PR looks good to merge.

PR #2285, opened 2016-03-16
DataTools: add number parsing API 

sbesson:
See http://trac.openmicroscopy.org/ome/ticket/7782

This PR adds a number of utility functions to `DataTools` allowing to generically:
- account for decimal separators in different locales (`.` vs `,`)
- parse a String into Double, Float, Integer...
- handle NumberFormatException thrown when parsing
Unit tests have been added to cover the newly added functions.

As discussed in http://trac.openmicroscopy.org/ome/ticket/7782, following the acceptance of this new API, all readers should be reviewed and string parsing should use this new API. I will review all candidate readers and open a corresponding card. Note this review should likely be done down the development line when we are sure we are not releasing any bug fix 5.1.x release.

### DIFF
--- a/components/formats-common/src/loci/common/DataTools.java
+++ b/components/formats-common/src/loci/common/DataTools.java
@@ -474,6 +474,7 @@ public final class DataTools {
    * @return {@code null} if the string could not be parsed
    */
   public static Float parseFloat(String value) {
+    if (value == null) return null;
     try {
       return Float.valueOf(sanitizeDecimalString(value));
     } catch (NumberFormatException e) {
@@ -488,6 +489,7 @@ public final class DataTools {
    * @return {@code null} if the string could not be parsed
    */
   public static Double parseDouble(String value) {
+    if (value == null) return null;
     try {
       return Double.valueOf(sanitizeDecimalString(value));
     } catch (NumberFormatException e) {
--- a/components/formats-common/src/loci/common/DataTools.java
+++ b/components/formats-common/src/loci/common/DataTools.java
@@ -54,6 +54,9 @@ public final class DataTools {
   // -- Static fields --
   private static final Logger LOGGER = LoggerFactory.getLogger(DataTools.class);
 
+  // Character used for the decimal sign in the current locale
+  private static final char decimalSeparator = new DecimalFormatSymbols().getDecimalSeparator();
+
   // -- Constructor --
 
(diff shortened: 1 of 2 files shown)


components/formats-common/src/loci/common/DataTools.java
```diff
@@ -409,7 +413,99 @@ public static String bytesToHex(byte[] b) {
...
+    } catch (NumberFormatException e) {
+      LOGGER.debug("Could not parse double value", e);
+    }
+    return null;
+  }
+
+  /** Normalizes the decimal separator of a string for the user's locale. */
+  private static String sanitizeDecimalString(String value) {
+    value = value.replaceAll("[^0-9,\\.]", "");
+    char separator = new DecimalFormatSymbols().getDecimalSeparator();
```
mtbc: As this does heap allocation and may be frequently called, I wonder would this line still work if it were a private static final field?

sbesson: Makes sense to me. I will give it a try.

sbesson:
@mtbc: see e43c169

rleigh-dundee:
Would it be better to be returning byte/short/int/long/float/double rather than Byte/Short/Integer/Long/Float/Double?  Or does autoboxing make this completely transparent?  Is there any performance overhead either way?

mtbc:
The underlying `valueOf` methods box them anyway, unfortunately, I don't know why. So, we already have them boxed.

dgault:
The new functions look good, and the performance hit on using Byte/Short/Integer/Long/Float/Double appears to negligible so seems to be the best choice and allows us to return null. The units tests makes sense and cover most of the possible scenarios. The only other test I would add to the unit tests would be for null eg - assertEquals(DataTools.parseFloat(null), null);

For most of the types this should pass fine but for the functions that call sanitizeDecimalString you will get a Null Pointer Exception

joshmoore:
> For most of the types this should pass fine but for the functions that call sanitizeDecimalString you will get a Null Pointer Exception

Good catch.

sbesson:
@dgault: thanks. Last two commits should add unit tests covering `null` and empty string (which works but might be a common case as well) and handle the NPE in the float/double parsing methods. Will relist for Monday.

dgault:
Units tests look good with the new commits and everything passes as expected

PR #2286, opened 2016-03-16
remove aspects of model for Shape

mtbc:
Effects https://github.com/openmicroscopy/design/issues/17#issuecomment-189236477 and https://github.com/openmicroscopy/design/issues/18#issuecomment-189241559 by adjusting the OME-XML schema to,

* remove `LineCap` altogether
* remove non-`Arrow` options for `MarkerStart`, `MarkerEnd`.

Also includes @rleigh-dundee's whitespace changes to avoid later conflicts.


sbesson:
While c9923b5 certainly improves the formatting of the schema, it has some impact on the readability of this PR as it effectively makes the schema changes non-reviewable via the GitHub Pull Request API with the default options. In this case https://github.com/openmicroscopy/bioformats/pull/2286/files?diff=split&w=1 should allow to review the changes. How difficult would it be to have the whitespace changes handled in a completely separate PR?

rleigh-dundee:
@sbesson The individual commits are reviewable and the changes are quite clear when viewed separately.

rleigh-dundee:
The code changes all look sensible.  I'll check the xslt behaviour as well.

rleigh-dundee:
```
XalanXPathException: The function number 'http://exslt.org/strings:tokenize' is not available. (Occurred in entity 'file:///home/rleigh/code/bioformats-test/components/specification/transforms/2015-01-to-2016-DEV0.xsl', at line 81, column 113.)
```

May have been introduced by another recent PR.

mtbc:
I'd guess that's from https://github.com/openmicroscopy/bioformats/commit/30343604e6a3b04b5b31571edfeb9e0482f9bf6c.

rleigh-dundee:
With `xsltproc` the `LineCap` and `MarkerStart`/`MarkerEnd` stripping and setting to `Arrow` all get processed correctly.  Tested with:

```
xsltproc  ../transforms/2015-01-to-2016-DEV0.xsl  2015-01/2013-06-datetests.ome  > test.xml
```

from inside the `samples` directory.

dgault:
Changes all look sensible to me and transforms have been tested and are working as expected. Looks good to merge.

rleigh-dundee:
@dgault Is there any possibility of reworking the xslt change in your recently-merged shape complextype PR to not use `tokenize` since this isn't supported by the Xerces processor?  http://www.heber.it/?p=1088 and http://stackoverflow.com/questions/7425071/split-function-in-xslt-1-0 might be helpful.

dgault:
I can certainly look at replacing the tokenize function in a separate PR. Im a little surprised that Ive never encountered that exception, as Ive run those transforms quite a number of times and I believe the reason I put it in was because other functions were not supported.

rleigh-dundee:
@dgault Thanks.  It's certainly missing from the xalan-c sources, having grepped through them.

sbesson:
The proposed model changes look good and in agreement with the proposal. Additionally all Java tests are passing. Following up on c9923b5, I have a few questions on the impact of this change. Are we effectively adjusting the XML formatting policy to use 2 spaces indentation and unify it with the rest of the Bio-Formats code? If so, is that something, will we need to apply this to the rest of the schemas, the transforms and maybe the OME-XML samples?

rleigh-dundee:
@sbesson We could do so if we wanted, but I thought it might be more conservative to leave "released" schemas unchanged which are already in widespread use.  While the chance of breakage is low, and it's theoretically not changing anything functionally, there's the possibility that parsers might change their behaviour e.g. parsing text content with adjusted whitespace.

sbesson:
Discussed the XML indentation policy with @mtbc @rleigh-dundee @dgault @joshmoore @jburel. Moving forward, using two spaces indentation certainly bring some benefits in terms of readability/editability as mentioned in https://github.com/openmicroscopy/bioformats/pull/2286#issuecomment-197890493 but also unification with the rest of the code base. Following this PR, the following actions were proposed:
- leaving published schemas unchanged, i.e. apply the policy to  schemas from 2016 and above
- modifying the XSLT transforms to comply with the 2 space indentation policy 
- modifying OME-XML samples in the repo and live to comply with the 2 space indentation policy
- adding a Readme and/or extending the Sphinx content to document this policy

PR #2287, opened 2016-03-17
Reindent and untabify the transforms

rleigh-dundee:
- Remove tabs
- Indent is two spaces

Done with emacs (`indent-region`) over each file.  I left out vim modelines since I couldn't validate they worked, and also Emacs local variables, but we could add them either per-file or per-directory if we wished.

Whitespace only diff: https://github.com/openmicroscopy/bioformats/pull/2287/files?w=1

--------

Testing: check all jobs are green.  There should be no change in the transform behaviour.


mtbc:
Good to merge.

sbesson:
Merging as per https://github.com/openmicroscopy/bioformats/pull/2287, thanks for the unification. As discussed elsewhere, the next step will be to document the new policy for XML style.

PR #2289, opened 2016-03-17
Remove non-OME DLLs from the repository

joshmoore:
As discussed with @RichardMyers et al., having these opaque
binaries in a GPL'd codebase puts the OME Team in a precarious
position. It was decided to leave the status quo for 5.1.8
but to remove them early enough so that alternative delivery
methods could be found for the SlideBook6Reader users.

See:

 - https://github.com/openmicroscopy/bioformats/pull/1993
 - https://github.com/openmicroscopy/bioformats/pull/2217


RichardMyers:
Thanks for the heads up @joshmoore et al.   If you can point me in the right direction for how best to move forward that would be much appreciated.   

My guess is we need to either host these dlls on our web site and/or create our own update site, as suggested on the [ImageJ distribution page](http://imagej.net/Distribution).

The next question then is what exactly we should host?  Should we host the entire SlideBook6Reader plugin, or just get the offending dlls to the right place somehow?  

The simple solution is to just throw up an error if the dlls are not found and point to a web page where the right dlls can be downloaded with instructions to put them in the plugin directory.  Maybe I'm overthinking things and that is the best solution.  The problem with this simple solution though is that it would be hard for users to automatically stay updated, which is why perhaps an ImageJ update site makes more sense.

Anyway, thanks in advance for your wisdom on this question.

RichardMyers:
I have  created an ImageJ update site: http://sites.imagej.net/SlideBook/

And it will update the file: jars\bio-formats\SlideBook6Reader.jar

Inside SlideBook6Reader.jar are the two native DLLs removed as part of this pull request:

  META-INF\lib\windows_32\SlideBook6Reader.dll
  META-INF\lib\windows_64\SlideBook6Reader.dll

This seems like the simplest way to make these native files available again and to keep them automatically updated.

If you think there's a problem with this solution for independently managing the native DLLs please just let me know.

joshmoore:
> I have created an ImageJ update site: http://sites.imagej.net/SlideBook/ ...

This is great. Thanks, @RichardMyers!

> If you think there's a problem with this solution for independently managing the native DLLs please just let me know.

From my side, as we discussed elsewhere, the primary issue with just hosting the DLLs as you've done here is that ...

> The simple solution is to just throw up an error if the dlls are not found and point to a web page where the right dlls can be downloaded with instructions to put them in the plugin directory.

... won't work. The error will be caught by Bio-Formats, logged, and the next reader in the [list](https://github.com/openmicroscopy/bioformats/blob/v5.1.8/components/formats-api/src/loci/formats/readers.txt#L38), i.e. [SlidebookReader](https://github.com/openmicroscopy/bioformats/blob/v5.1.8/components/formats-gpl/src/loci/formats/in/SlidebookReader.java), will be chosen. For the moment, I think what you've done is the best solution possible.

On our side, we need to enable you to also host `SlideBook6Reader.java` in your update site. At that point, you could be more aggressive about warning your users:

> ImageJ dialog: "*You have the official SlideBook update site activated, this is a SlideBook file, but the DLLs are improperly installed (for this platform, etc). I'm exiting now*".

I'd propose that when we have that functionality in place we all test out the migration of your reader implementation to do just that.

RichardMyers:
Ah right @joshmoore , I forgot that there is a problem giving the user feedback when a required dll for the reader plugin is missing.

> On our side, we need to enable you to also host SlideBook6Reader.java in your update site. At that point, you could be more aggressive about warning your users...

Sorry, but I don't really understand the connection between us hosting the .java and giving a better warning.   But either way, it seems logical that we host both the java and matching dlls.

Are you just proposing that before we host it, we also need a fix for the case where the SlideBook update site is active but a particular architecture (eg. OS X) is not supported?  

I'm also curious if bio-formats would still need to host some sort of stub SlideBook6Reader class, or does the framework skip readers when the corresponding reader plugin class is missing?

joshmoore:
--rebased-to #2300

joshmoore:
> Sorry, but I don't really understand the connection between us hosting the .java and giving a better warning. But either way, it seems logical that we host both the java and matching dlls.

If you host the .java then you can decide to throw a proper exception if the DLLs aren't in place. Vanilla Bio-Formats can't do that because it can't assume that the user has enabled a given update site.

> Are you just proposing that before we host it, we also need a fix for the case where the SlideBook update site is active but a particular architecture (eg. OS X) is not supported?

Not necessarily. You could conditionally throw the exception on just Windows. You could even assume ImageJ (since it's an update site) or detect ImageJ and open a dialog. Again, not something we can do from vanilla Bio-Formats. (If you want someone to use the SlideBook6 reader outside of ImageJ, though, this would also need to be conditional.)

> I'm also curious if bio-formats would still need to host some sort of stub SlideBook6Reader class, or does the framework skip readers when the corresponding reader plugin class is missing?

Our goal is certainly to *not* require a stub.

joshmoore:
Hi @RichardMyers. The report on the testing of this is available in https://github.com/openmicroscopy/bioformats/pull/2300#issuecomment-207438335 using builds of the upcoming 5.1.9 release. Though the debugging of "No DLLs present" does not seem to be happening, the update site **definitely** enables use of `SlideBook6Reader`. Thanks for making this happen!

From our side, the only other TODO is to know which URL you would like us to point to for explaining to users how to activate the update site.

RichardMyers:
Hi @joshmoore - I talked to our IT guy and he will create a redirect from [www.intelligent-imaging.com/bioformats](http://www.intelligent-imaging.com/bioformats) to the correct entry in our 'technical answers' (basically a FAQ page).   This is not yet done, but in the works so feel free to use that link.

I have also updated SlideBook6Reader.jar on http://sites.imagej.net/SlideBook/ to include a version of SlideBook6Reader.class that pops up a dialog if the dynamic library for the current architecture is not present.  I still need to update the URL of the dialog to also point to the link above.

We also have a (very) beta and untested version of the linux library in that jar.  It probably won't work, but its a start.

The final step is to do a pull request to remove SlideBook6Reader.java from development and 5.1.x branches.  If you want to see my dialog changes, they'll only be in our branch since it necessitated changes to pom.xml and build.properties to add ij.jar.  I don't think you want those changes merged in.

RichardMyers:
@joshmoore - on second thought, removing SlideBook6Reader.java may not be the correct solution.  How would you like to handle this transition?  For example, SLD file tests will fail if we remove it.

My changes are made in commit [ccfb39e](https://github.com/Intelligent-Imaging/bioformats/commit/ccfb39e1cd4c37383aea57f9ed81689e0ca78481) of our fork.

If I do a pull request to openmicroscopy:develop then the SlideBook6Reader.class file will still be included in formats-gpl.jar and the SlideBook6Reader.class file hosted at http://sites.imagej.net/SlideBook/ will not be used.

joshmoore:
> Hi @joshmoore - I talked to our IT guy and he will create a redirect from www.intelligent-imaging.com/bioformats to the correct entry in our 'technical answers' (basically a FAQ page). This is not yet done, but in the works so feel free to use that link.

Great, thanks.

> We also have a (very) beta and untested version of the linux library in that jar. It probably won't work, but its a start.

Interesting! Let us know when you're ready for us to generally point towards this for Linux usage as well. 

> @joshmoore - on second thought, removing SlideBook6Reader.java may not be the correct solution. How would you like to handle this transition? For example, SLD file tests will fail if we remove it.

> If I do a pull request to openmicroscopy:develop then the SlideBook6Reader.class file will still be included in formats-gpl.jar and the SlideBook6Reader.class file hosted at http://sites.imagej.net/SlideBook/ will not be used.

To some degree, you're asking when/how to achieve a.b.c & d.e.f in the following table:

Version | Platform | Reader chosen
------------ | ------------- | -----------
5.1.8 | Linux | SlideBookReader from formats-gpl.jar
5.1.8 | Windows | SlideBook6Reader from formats-gpl.jar
5.1.9 | Linux | SlideBookReader from formats-gpl.jar
5.1.9 | Windows | SlideBook6Reader from formats-gpl.jar + update site
a.b.c | Linux | SlideBook6Reader from formats-gpl.jar + update site
d.e.f | * | SlideBook6Reader from update site (requires readers.txt)
x.y.z | * | SlideBook6Reader from update site

Since we're in the process of releasing 5.1.9, I'd suggest we not try to remove anything more from formats-gpl at the moment but discuss the steps for whatever's next. If you need anything from our side to prepare those steps, let us know.

I *think* that what you're attempting ccfb39e will work, but only if `SlideBook6Reader` is removed from this repo.

RichardMyers:
After 5.1.9 we need to host SlideBook6Reader.java from our update site to enable the missing .dll warning dialog on startup, the question is how do we remove it from the openmicroscopy/bioformats repo in a safe manner.

After 5.1.9 I can put in a pull request to delete SlideBook6Readers.java;  I assume readers.txt stays the same.  Are there other changes I should make?

I assume this will break test scripts.  Should I remove dependencies on SlideBook6Reader.java that might occur in other parts of the codebase?  I'm afraid I'm not really qualified to make those changes.

If your test scripts can be modified to first update SlideBook6Reader.jar from the new SlideBook update site, then maybe that is the only change that needs to be made for the test scripts.

From 5.1.9 user must enable the SlideBook update site, but on non-Windows systems it will silently fail (only a message to the debug log).

From version d.e.f we can expect that the new warning dialog will appear at startup for any users running on systems I haven't added dlls for.  Hopefully we'll have some non-Windows dlls available by then too.

PR #2290, opened 2016-03-18
NativeND2Reader: Fix for the emmission wavelength

bramalingam:
ref: http://trac.openmicroscopy.org/ome/ticket/13179

Testing Instructions : 

Please try opening the 'nd2' file from QA 17099 in FIJI/ImageJ. Without this PR you should be getting a String parsing exception. This PR should fix that issue, and open the image without any issues. Also compare the values for the 'emission wavelengths' between NISElements Viewer and Bio-Formats, and check if the values match.

@melissalinkert @jburel : As always, please let me know if the code change looks fine.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -51,6 +51,7 @@ import loci.formats.meta.MetadataStore;
 import ome.xml.model.primitives.Color;
 import ome.xml.model.primitives.PositiveFloat;
 import ome.xml.model.primitives.PositiveInteger;
+
 import ome.units.quantity.ElectricPotential;
 import ome.units.quantity.Frequency;
 import ome.units.quantity.Length;
@@ -2233,7 +2234,7 @@ public class NativeND2Reader extends FormatReader {
           if (value.endsWith("Active")) {
             int first = key.lastIndexOf(":") + 1;
             int last = key.lastIndexOf(";");
-            if(last-first < 0){
+            if (last-first < 0){
                 last = first + key.substring(first).indexOf(' ');
             }
             try {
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -48,11 +48,9 @@
 import loci.formats.codec.JPEG2000Codec;
 import loci.formats.codec.ZlibCodec;
 import loci.formats.meta.MetadataStore;
-
```
joshmoore: Can you revert the whitespace changes please?

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -2235,9 +2233,12 @@ else if (key.startsWith("Line:")) {
           if (value.endsWith("Active")) {
             int first = key.lastIndexOf(":") + 1;
             int last = key.lastIndexOf(";");
+            if(last-first < 0){
```
joshmoore: But you could add one of the whitespaces after the `if` here :smile: (i.e. `if (last...`)

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/158/console, this PR does not seem to introduce any regression with existing ND2 filesets. Do we have a representative sample of the failure addressed in this PR that should be configured and added to the data repository?

bramalingam:
QA 17099 should throw this error for sure, so that could be a candidate.

rleigh-dundee:
The change looks fine, wavelengths are correct and match NIS elements.

There's still a remaining whitespace removal which needs correcting.

sbesson:
Thanks for the reverted whitespace change. https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/116/ was green with the QA file included.

bramalingam:
--rebased-to #2321

PR #2293, opened 2016-03-18
Updating Improvision format links (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2259 but rebased onto dev_5_1.

----

This should make the latest docs green again.

The current problem is actually just that their redirect is not working properly but given that Improvision doesn't exist as a separate entity any more, I figured best to just update the links anyway. Also removed the Openlab links as although they weren't breaking the build, the content doesn't exist any more, it was just redirecting to a 'We're sorry' page and searching for Openlab on the main PE page gives no useful results.


bramalingam:
--rebased-from #2259

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-docs/1194/ is green. Merging.

PR #2268, opened 2016-03-02
Zeiss CZI: turn off prestitched flag for separate scenes

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13095

To test, use the file from QA 16882.  Without this change, ```showinf -minmax``` should show that the first series has non-zero pixels, and the remaining four series have all zero pixels.  With this change, the same test should show that all five series have non-zero pixels (which should also be obvious from looking at the images).  The images and dimensions can also be compared against Zeiss' ZEN software.

PR #2295, opened 2016-03-18
Zeiss CZI: turn off prestitched flag for separate scenes (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2268 but rebased onto dev_5_1.

----

Fixes https://trac.openmicroscopy.org/ome/ticket/13095

To test, use the file from QA 16882.  Without this change, ```showinf -minmax``` should show that the first series has non-zero pixels, and the remaining four series have all zero pixels.  With this change, the same test should show that all five series have non-zero pixels (which should also be obvious from looking at the images).  The images and dimensions can also be compared against Zeiss' ZEN software.


bramalingam:
--rebased-from #2268

sbesson:
Closing to combine all CZI changes in one PR (to minimize conflict resolution).

PR #2300, opened 2016-03-18
Remove non-OME DLLs from the repository (rebased onto dev_5_1)

joshmoore:
This is the same as gh-2289 but rebased onto dev_5_1.

----

As discussed with RichardMyers et al., having these opaque
binaries in a GPL'd codebase puts the OME Team in a precarious
position. It was decided to leave the status quo for 5.1.8
but to remove them early enough so that alternative delivery
methods could be found for the SlideBook6Reader users.

See:

 - https://github.com/openmicroscopy/bioformats/pull/1993
 - https://github.com/openmicroscopy/bioformats/pull/2217


joshmoore:
--rebased-from #2289

joshmoore:
Testing instructions:

 * Using Windows 32bit or 64bit (potentially both)
 * Find a working SlideBook file (e.g. `test_images_good/slidebook`).
 * Check that the SlideBook6 reader prints a warning that the DLLs are not found. (This may require turning on debugging)
 * Enable http://sites.imagej.net/SlideBook/ in ImageJ following the [regular process](How_to_follow_a_3rd_party_update_site)
 * Re-open the same file. The debug message should not be printed.

bramalingam:
Testing Results:

Opening the image on ImageJ : 5.1.8 (No update sites)
The image opens, the original metadata does not have any metadata added by the slide book reader.
There is a warning for missing dll's.

Opening the image on FiJI : 5.1.8 (No update sites)
The image opens, the original metadata does not have any metadata added by the slide book reader.
There is "no" warning for missing dll's.

Opening the image on FIJI : 5.1.9_Snapshot (No update Sites).
The image opens, the original metadata "has" metadata added by the slide book reader.
There is "no" warning for missing dll's.

Opening the image on FIJI : 5.1.9_Snapshot (With SlideBookReader update Sites).
The image opens, the original metadata "has" metadata added by the slide book reader.
There is "no" warning for missing ell's.

The thing that am confused about is the lack of warnings on FIJI with regards to dll's. I am testing on a Windows 8, FIJI  : ImageJ 1.50g, java 1.8.0_66 (64bit)

joshmoore:
> The thing that am confused about is the lack of warnings on FIJI with regards to dll's.

The warnings are really debug statements. There was nothing in the extra debug window?

> Opening the image on FIJI : 5.1.9_Snapshot (With SlideBookReader update Sites).
The image opens, the original metadata "has" metadata added by the slide book reader.

This final result is a bit surprising. This may mean that the update site is not yet working for us, and would count as a blocker against this PR. i.e. we need something that will get us back to "The image opens, the original metadata does not have any metadata added by the slide book reader." in the "5.1.9_Snapshot (With SlideBookReader update Sites)." case.

bramalingam:
My bad : 

The version BF was somehow switched while running the updates and it went back to 5.1.8 in my previous testing. 

So yes:
For  FIJI : 5.1.9_Snapshot (With SlideBookReader update Sites).
"The image opens, the original metadata does not have any metadata added by the slide book reader." 

The warning messages are still an issue, but they are not blockers. So I guess this PR is good to merge now.

sbesson:
Thanks. Merging this and the companion `develop` PR. What are the next steps in terms of documenting the new workflow for getting `dlls`? /cc @hflynn

PR #2302, opened 2016-03-18
Fake reader: HCS with regions

sbesson:
See https://trello.com/c/hzWe7ZNR/115-fakereader-fix-regions-with-hcs

The default `XMLMockObjects.createPlate()` was creating a `MicrobeamManipulation` object with an associated `ROI` which was conflicting with the annotation indexing in the `FakeReader` semantics. The default is modified to remove this creation although it might be possible to selectively enable it using a flag i.e. `XMLMockObjects.createPlate(int, boolean)`.

The flip effect of this change is that fakes should now be able to combine HCS and Regions features, i.e. the following

```
touch "HCSanalysis&plates=10&plateRows=16&plateCols=24&rectangles=100.fake"
```

should allow us to manipulate objects that approach the scale at which folders should be tested. /cc @mtbc

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -1083,9 +1083,9 @@ public class FakeReader extends FormatReader {
     final XMLMockObjects xml = new XMLMockObjects();
     OME ome = null;
     if (screens==0) {
-      ome = xml.createPopulatedPlate(plates, rows, cols, fields, acqs);
+      ome = xml.createPopulatedPlate(plates, rows, cols, fields, acqs, false);
     } else {
-      ome = xml.createPopulatedScreen(screens, plates, rows, cols, fields, acqs);
+      ome = xml.createPopulatedScreen(screens, plates, rows, cols, fields, acqs, false);
     }
     getOmeXmlMetadata().setRoot(new OMEXMLMetadataRoot(ome));
     // copy populated SPW metadata into destination MetadataStore
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -505,6 +505,7 @@ public class FakeReader extends FormatReader {
     boolean falseColor = false;
     boolean metadataComplete = true;
     boolean thumbnail = false;
+    boolean withMicrobeam = false;
 
     int seriesCount = 1;
     int lutLength = 3;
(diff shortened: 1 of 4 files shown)


joshmoore:
I have some slight concern that the change will leave microbeam even more under-tested than it already is since this method is also used throughout OMERO. Would be good to at least plan for re-enabling the second codepath.

sbesson:
@joshmoore: last 2 commits should restore backwards compatibility of the `XMLMockObjects` method and specifically override the `Microbeam` creation in the case of the fake HCS.

components/formats-bsd/src/loci/formats/in/FakeReader.java
```diff
@@ -1083,9 +1083,9 @@ private int populateSPW(MetadataStore store, int screens, int plates, int rows,
     final XMLMockObjects xml = new XMLMockObjects();
     OME ome = null;
     if (screens==0) {
-      ome = xml.createPopulatedPlate(plates, rows, cols, fields, acqs);
+      ome = xml.createPopulatedPlate(plates, rows, cols, fields, acqs, false);
```
rleigh-dundee: Should the reader provide the option to call this method with w`ithMicrobeam=true`?  Likewise for createPopulatedScreen?

sbesson: Good point. It would certainly be good to have this feature preserved if anyone depended on it. Let me see how hard it is to fix the AnnotationRef indexation issue in this PR. Otherwise, I'll tackle it in a follow-up PR.

sbesson:
Pushed. Now

```
touch "HCSanalysis&plates=10&plateRows=16&plateCols=24&withMicrobeam=true&rectangles=100.fake"
```

should turn the microbeam addition on while allowing  regions of interest to be created for each image . Unit tests have been added to `FakeReaderTest` to cover the new keyword semantics.

Note there is still a  bug in the case of multiple screens with microbeam i.e. `./showinf -omexml "HCSanalysis&screens=2&withMicrobeam=true.fake"` fails at the validation step. However, this was already the case on `dev_5_1`. As this is not a regression, I am inclined to limit the scope of this PR and address the issue it in a separate one.

joshmoore:
Can I propose we file that as a follow-on? Otherwise, I like the `withMicrobeam` (though of course, I wonder what will come next...). Good to merge from my point-of-view.

sbesson:
Filed as https://trello.com/c/pBQgFTMS/117-fakereader-fix-microbeam-with-multiple-screens. Merging

PR #2303, opened 2016-03-18
cpp: ome-xml: Add enum value list macros

rleigh-dundee:
This is to allow direct code generation with the C++ preprocessor by looping over the available enum values.

A better approach would be to use C++11 variadic templates, but this is not yet a possibility, and they too are broken on Windows.  See http://stackoverflow.com/questions/35360908/alternative-to-expanding-templates-in-a-switch-statement/ for further details.

This is the second attempt to make this build.  It's the same as before, but the nested loops are skipped on Windows due to the preprocessor being broken.

Purpose: It will be used to simplify code working with model enums, but the primary intent is model units support (to be added in a future PR).

See also: https://github.com/ome/ome-files-cpp/pull/5

Generated source changes: https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-generated-sources/lastSuccessfulBuild/artifact/ome-xml-model-cpp.diff/*view*/

--------

Testing: check the builds are green; it's tested by the `enum` unit test (and `pixelproperties` test in ome-files).

Optional (because it's tedious) manual verification, on Linux or MacOS X since it requires make to run the preprocessor:

Clone the following repositories:

- ome-cmake-superbuild (branch: develop or develop/merge/files-cpp)
- ome-common-cpp (branch: develop or develop/merge/files-cpp)
- bioformats (branch: rleigh-dundee/preprocess-enums or develop/merge/files-cpp)
- ome-files-cpp (branch: rleigh-dundee/preprocess-enums or develop/merge/files-cpp)

Then run:

```
mkdir /tmp/b
cd /tmp/b
cmake -Dome-common-dir=/path/to/ome-common-cpp -Dome-xml-dir=/path/to/bioformats -Dome-files-dir=/path/to/ome-files-cpp -Dsource-caache=~/sources -Dbuild-prerequisites=OFF /path/to/ome-cmake-superbuild
make
[build will eventually complete]
cd ome-xml-build/cpp/test/ome-xml
make enum.cpp.i
[Look at the contents of CMakeFiles/enum.dir/enum.cpp.i]
cd ../../../..
cd ome-files-build/lib/ome-files
make PixelProperties.cpp.i
[Look at the contents of CMakeFiles/ome-files.dir/PixelProperties.cpp.i]
```

For both of these files, you should see the macros in the original source files (enum.cpp and PixelProperties.cpp) expanded into switch statements, one per enum value.  It's right at the end of both files; ignore the header junk at the top.

### DIFF
--- a/components/xsd-fu/templates-cpp/OMEXMLModelEnum.template
+++ b/components/xsd-fu/templates-cpp/OMEXMLModelEnum.template
@@ -55,6 +55,9 @@
 
 #include <ome/xml/model/enums/EnumerationException.h>
 
+// All values in the ${klass.langType} enumeration.
+#define ${fu.CLASS_PREFIX}_VALUES {% for value in klass.possibleValues %}(${enum_value_name(value, False).upper()}){% end %}
+
 {% end %}\
 {% if fu.SOURCE_TYPE == "source" %}\
 #include <algorithm>
--- a/components/xsd-fu/xsd-fu
+++ b/components/xsd-fu/xsd-fu
@@ -56,11 +56,18 @@ from getopt import gnu_getopt, GetoptError
 from glob import glob
 from xml import sax
 
-def guard(filename):
+def macroname(filename):
     filename = "_".join(filename.split(os.path.sep))
     filename = filename.upper()
     filename = re.sub("[^A-Z0-9]", "_", filename)
     return filename
+
+def guard(filename):
+    return macroname(filename)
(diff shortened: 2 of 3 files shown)


joshmoore:
Merging based on https://ci.openmicroscopy.org/view/Files-DEV/job/OME-FILES-CPP-DEV-merge-superbuild/161/ et al.

PR #2305, opened 2016-03-22
Zeiss CZI: fixes for files with multiple mosaics and positions

melissalinkert:

To test, use the file in ```curated/zeiss-czi/niko``` (it will be ~100GB).  The images are quite large; the easiest first test would be to select a few ~2kx2k tiles and check the display in ImageJ vs Zeiss' ZEN.  It would be good to try importing into OMERO as well, to validate that all of the pixel data is read correctly, but this will take a while as pyramids will need to be generated.

Configuration PR is: https://github.com/openmicroscopy/data_repo_config/pull/96

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -314,6 +314,19 @@ public class ZeissCZIReader extends FormatReader {
     Arrays.fill(buf, (byte) 0);
     RandomAccessInputStream stream = new RandomAccessInputStream(currentId);
     try {
+      int minTileX = Integer.MAX_VALUE, minTileY = Integer.MAX_VALUE;
+      for (SubBlock plane : planes) {
+        if ((plane.seriesIndex == currentSeries && plane.planeIndex == no) ||
+          (plane.planeIndex == previousChannel && validScanDim))
+        {
+          if (plane.row < minTileY) {
+            minTileY = plane.row;
+          }
+          if (plane.col < minTileX) {
+            minTileX = plane.col;
+          }
+        }
+      }
       for (SubBlock plane : planes) {
         if ((plane.seriesIndex == currentSeries && plane.planeIndex == no) ||
           (plane.planeIndex == previousChannel && validScanDim))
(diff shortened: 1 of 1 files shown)


bramalingam:
BF : SeriesCount = 8. SizeX = 33291, SizeY = 20493 (Whole Image split into 8 series)
ZenLite : ImageSize : SizeX = 136283 SizeY = 44372 (Whole Image with 980 tiles)

I went through the PositionX and PositionY values to calculate the maximum values for the montage,
and 

maximum Value of PositionX = 102816 , now adding the sizeX to this gives : (102816+33291) = 136107 which is lesser than 136283 (From Zen Lite).

and maximum Value PositionY = 23868 , now adding the sizeY to this gives : (23868 + 20493) = 44361 which is lesser than 44372(From Zen Lite).

So, in conclusion: The pixel data seems to fit into the image montage (if stitched together), but seems to be missing a few pixels at the edges. I am unsure if this expected, will leave it up to @melissalinkert  to decide.

All the metadata matches between BF and ZenLite for this image.

melissalinkert:
Assuming that the image display has been validated against Zen, and given that we don't stitch scenes (the 8 series in this case) where Zen does, then that's more or less what I'd expect.  To do the arithmetic a different way, our 8 series corresponds to 4x2 stitched tiles == 33291x4 x 20493x2 == 133164 x 40986, so within the bounds of the dimensions reported by Zen.

sbesson:
@bramalingam: what is your call on image display validation? Could we paste screenshots from Bio-Formats and Zen into the PR comments?

bramalingam:
That will not help, cause the image has low contrast and it will hard to infer anything from those screenshots. Melissa knows better about this, and I would suggest we go ahead and merge this.

bramalingam:
--rebased-to #2329

sbesson:
--rebased-to #2330

ehrenfeu:
Might be a stupid question, but is this somewhere available as a pre-release build? Jenkins or whatever? We'd be keen on testing! :wink:

sbesson:
@ehrenfeu: this bug fix should be soon released as 5.1.9. If you are keen on testing pre-release Continuous Integration builds, you should be able to download them from the [latest 5.1 CI build](https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-latest/) which includes the backported PR. See also https://www.openmicroscopy.org/site/support/contributing/ci-bio-formats.html for more information.

PR #2311, opened 2016-03-25
source-archive: Set antversion.xml and GitVersion.cmake mtime

rleigh-dundee:
Correct  timestamp

```
ant release
```

Check `ant/gitversion.xml` and `cmake/GitVersion.cmake` in the two tar archives.


sbesson:
Retested with the last version of the commit. Works as expected.

bramalingam:
--rebased-to #2312

PR #2312, opened 2016-03-25
source-archive: Set antversion.xml and GitVersion.cmake mtime (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2311 but rebased onto dev_5_1.

----

Correct  timestamp

```
ant release
```

Check `ant/gitversion.xml` and `cmake/GitVersion.cmake` in the two tar archives.

### DIFF
--- a/tools/source-archive.py
+++ b/tools/source-archive.py
@@ -6,6 +6,7 @@ from optparse import OptionParser
 import os
 from subprocess import call
 import sys
+import time
 import zipfile
 import tarfile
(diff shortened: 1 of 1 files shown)


bramalingam:
--rebased-from #2311

sbesson:
Tested with the artifacts of https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-build/1134/. Works as expected.

PR #2313, opened 2016-03-28
Leica LIF: fix multi-dimensional data from color cameras

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13112.

This fixes the case when a color camera is used (so each plane has 3 RGB
channels), and multiple channels are acquired potentially in addition to
multiple timepoints and/or Z sections.

To test, use the files from QA 16911 and QA 16919.  Compare the images as shown in Leica's LAS X or LAS AF Lite software against what is shown by ImageJ with and without this change.  Also verify that with this change, no error messages are shown when opening either file in ImageJ.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -2020,17 +2020,22 @@ public class LIFReader extends FormatReader {
     ms.indexed = !ms.rgb;
     ms.imageCount = ms.sizeZ * ms.sizeT;
     if (!ms.rgb) ms.imageCount *= ms.sizeC;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Config PR: https://github.com/openmicroscopy/data_repo_config/pull/99

dgault:
Code changes make sense and the two QA files now open without any of the errors that were thrown before applying the PR. The images look and display as expected also. Config PR has also been tested and is green - https://ci.openmicroscopy.org/view/Bio-Formats-DEV/job/BIOFORMATS-DEV-merge-repository-subset/207/

Changes look good to merge.

Issue #2237, opened 2016-02-10
Exporting to ICS can throw ArrayIndexOutOfBoundsException


ctrueden:
Macro to reproduce:
```javascript
run("Organ of Corti (2.8M, 4D stack)");
outDir = "/home/curtis/Desktop";
run("Bio-Formats Exporter", "save=" + outDir + "/organ-of-corti.ics write_each_channel");
```
Relevant stack trace:
```
(Fiji Is Just) ImageJ 2.0.0-rc-44/1.50e; Java 1.8.0_66 [64-bit]; Linux 3.19.0-39-generic; 55MB of 1065MB (5%)
 
java.lang.ArrayIndexOutOfBoundsException: 15
	at loci.formats.out.ICSWriter.saveBytes(ICSWriter.java:130)
	at loci.formats.FormatWriter.saveBytes(FormatWriter.java:124)
	at loci.plugins.out.Exporter.run(Exporter.java:767)
	at loci.plugins.LociExporter.run(LociExporter.java:75)
```

Exception is the same using both release (Bio-Formats 5.1.7) and dev version (Bio-Formats 5.1.7-DEV 62c4dd810, 9 February 2016).

Reported by @tischi on [this ImageJ forum thread](http://forum.imagej.net/t/bug-in-bio-formats-exporting-ics/823).

bramalingam:
Thank you for submitting the issue Curtis. The same was reproducible locally, and a card has been created on this regard. 
https://trello.com/c/L7dtKWsl/96-bio-format-exporter-issues

ctrueden:
Thanks @bramalingam !

joshmoore:
Closing in favor of the card. Thanks, guys.

PR #2315, opened 2016-03-31
ICS writer: fix dimension population for split files

melissalinkert:
See https://github.com/openmicroscopy/bioformats/issues/2237 and
https://trello.com/c/L7dtKWsl/96-bio-format-exporter-issues

To test, use the macro from gh-2237 and test cases (1) and (2) from the Trello card.  Verify that no exceptions are thrown during conversion, and that the resulting files are ~4 MB each and are reported by showinf as having 15 Z sections (1 timepoint, 1 channel).  Also verify that converting the ```Organ of corti``` sample to a single .ics file does not behave differently with this change; be sure to test conversion in both ImageJ and bfconvert, and check each plane in the resulting file against the original data.

Note that there are likely still some edge cases here.  The "correct" way to do this across all writers would be to add API similar to suggestions in https://trac.openmicroscopy.org/ome/ticket/8215, so that a writer can be given a mapping between plane indexes and file names prior to ```setId``` - and then we're not trying to guess what the final dimensions of each file might be.  That needs more thought and probably is not in scope for 5.2.0 though.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/ICSWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/ICSWriter.java
@@ -34,6 +34,8 @@ package loci.formats.out;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 
 import loci.common.RandomAccessInputStream;
 import loci.common.RandomAccessOutputStream;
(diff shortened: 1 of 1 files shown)


sbesson:
Note that `ConversionTest.testICS` fails with this PR included - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-build/181

melissalinkert:
Commit pushed which should fix the test failure.

rleigh-dundee:
Changes look reasonable and testing with the ImageJ macro and with bfconvert demonstrates the correct behaviour without errors.  Merge build is now green.  Please merge.

PR #2318, opened 2016-04-01
Fixes #2317 - infer channel count from channel names

### LINKED ISSUE #2317: FlowSightReader error
I have a .cif file from a user that fails to read with the following error:

    Channel count (1) does not match number of channel names (6) in string "C0|C1|C21|C3|C4|C5

This is an error in interpreting one of the TIF tags in the file; I will contribute a patch that fixes it today.

LeeKamentsky:
See #2317 for more details. The fix is pretty surgical - the only effect is to try harder for a case that previously threw an exception.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FlowSightReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FlowSightReader.java
@@ -150,9 +150,7 @@ public class FlowSightReader extends FormatReader {
     if (channelNamesString != null) {
       channelNames = channelNamesString.split("\\|");
       if (channelNames.length != channelCount) {
-        throw new FormatException(String.format(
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thanks, @LeeKamentsky.  This reads fine to me; adding the ```include``` label so that it is included in tomorrow's Jenkins builds.

sbesson:
Automated builds have been passing against our set of sample files with this PR included. @LeeKamentsky: would you have a sample file available that would reproduce the error fixed by this PR which could be added to our data repository?

sbesson:
Merging. As proposed by @mtbc, a follow-up PR should be opened to add a `DEBUG` level statement logging this particular use case.

PR #2319, opened 2016-04-04
Codegen - Adding BinData

dgault:
This PR will allow for our codegen to handle raw binary data. For the Mask implementation in particular this requires some unique behaviour to be added. In our schema a Mask extends base64binary, which in Java would be the equivalent of extending a byte[]. The metadata API reflects this by allowing you to set BinData as such - setMaskBinData(byte[] binData, int ROIIndex, int shapeIndex)

From a Java perspective it would make more sense to have the raw binary as an additional attribute on the BinData object, however this would require larger schema changes and break the existing API in favour of adding new functions that perhaps don't accurately reflect the schema. 

As such I have modified the codegen so that if a model object has a primitive base then it shall instead extend the default type and the primitive base will be added as an attribute with the tag baseAttribute. Properties tagged as being baseAttributes will be exposed in the API at the top level, ie setMaskBinData as opposed to setMaskBinDataBinData. 

This will see the BinData model object add the following property along with getters and setters. The getters and setters will then be used by the top level API functions, leaving the metadata API unchanged.
// BinData property
private byte[] binData;

Reasons for taking this approach:
- Limit schema changes as much as possible
- Metadata API should remain unchanged
- Simplifies templates by removing previous specialisations for shape and lightsource abstract 
- Removes last remaining hardcoded bin data functions

In Java we are using a raw byte[], the reason for using this as opposed to List<Byte>, ByteBuffer, InputStream etc was firstly for efficiency in storage and in performance when converting back and forth from the XML. In CPP this is currently listed as a string for simplicity but this would certainly welcome suggestions on a suitable alternative.

Note: The CPP implementation requires the Shape complex type PR - https://github.com/openmicroscopy/bioformats/pull/2160 
As such the CPP is not going to validate against the 2013 schema

Testing:
I am in the process of defining a suitable testing strategy.
- Check that all tests remain green
- Verify that the new attribute has been added to BinData model along with getters and setters
- Verify that metadata API remains unchanged


sbesson:
Note this PR impacted the OMERO build and various Bio-Formats merge builds in this morning run. Marking as `breaking` so that we can relaunch the merge jobs but have a CI suite to test its impact.

dgault:
Mask Bin Data has now been added to FakeReader and the FakeReaderTest.
Additionally, a sample file with Mask BinData has been added to inbox with an accompanying config - https://github.com/openmicroscopy/data_repo_config/pull/103

components/formats-api/src/loci/formats/meta/MetadataConverter.java
```diff
@@ -1021,6 +1021,10 @@ private static void convertImages(MetadataRetrieve src, MetadataStore dest,
           try {
             Boolean bigEndian = src.getPixelsBinDataBigEndian(i, q);
             dest.setPixelsBinDataBigEndian(bigEndian, i, q);
+            NonNegativeLong length = src.getPixelsBinDataLength(i, q);
```
rleigh-dundee: Should each pair be in a separate try/catch block so they always get called if the previous set failed for any reason?

components/xsd-fu/templates-cpp/AggregateMetadata.template
```diff
@@ -397,27 +399,29 @@ namespace ome
...
 
 {% end debug %}\
+{% if len(model.opts.lang.getSubstitutionTypes()) == 0 %}\
 {% if fu.SOURCE_TYPE == "header" %}\
-        // Documented in base class.
-        index_type
-        getPixelsBinDataCount(index_type imageIndex) const;
+        // Documented in base class.    
+        index_type    
+        getPixelsBinDataCount(index_type imageIndex) const;
```
mtbc: not sure we need all these trailing spaces?

components/formats-bsd/src/loci/formats/in/FakeReader.java
```diff
@@ -986,7 +990,18 @@ private void fillRegions(MetadataStore store, int imageIndex) {
...
+        store.setMaskX((double)ROI_SPACING, roiCount, 0);
+        store.setMaskY((double)ROI_SPACING, roiCount, 0);
+        store.setMaskWidth((double)ROI_SPACING, roiCount, 0);
+        store.setMaskHeight((double)ROI_SPACING, roiCount, 0);
         store.setImageROIRef(roiID, imageIndex, roiRefCount);
+        byte[] rawBytes = new byte[ROI_SPACING*ROI_SPACING];
+        random.nextBytes(rawBytes);
+        String base64encoded = DatatypeConverter.printBase64Binary(rawBytes);
+        byte[] binData = base64encoded.getBytes(StandardCharsets.UTF_8);
+        store.setMaskBinData(binData, roiCount, 0);
```
rleigh-dundee: Why are we dealing with Base64 encoding at this level?  Why isn't the API dealing solely with raw bytes?

Should this conversion to/from Base64 not be done internally by the BinData object, either in the getter/setter or when serialising to/from XML?

components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
```diff
@@ -656,8 +656,14 @@ else if (proc instanceof ShortProcessor) {
...
-            boolean littleEndian =
-                    !w.getMetadataRetrieve().getPixelsBinDataBigEndian(0, 0).booleanValue();
+            boolean littleEndian = false;
+            if (w.getMetadataRetrieve().getPixelsBigEndian(0) != null)
+            {
+              littleEndian = !w.getMetadataRetrieve().getPixelsBigEndian(0).booleanValue();
+            }
+            else if (w.getMetadataRetrieve().getPixelsBinDataCount(0) == 0) {
+              littleEndian = !w.getMetadataRetrieve().getPixelsBinDataBigEndian(0, 0).booleanValue();
+            }
```
rleigh-dundee: This two step pattern is so common, I do wonder if we shouldn't implement it directly inside `getPixelsBigEndian` so it works correctly all the time for every caller?

components/specification/released-schema/2016-DEV0/ome.xsd
```diff
@@ -3882,7 +3885,7 @@ PlaneInfo (as Plane)
       <xsd:complexContent>
         <xsd:extension base="Shape">
           <xsd:sequence>
-            <xsd:element ref="BinData" minOccurs="1" maxOccurs="unbounded"/>
+            <xsd:element ref="BinData" minOccurs="1" maxOccurs="1"/>
```
rleigh-dundee: Do we want a similar constraint on `Pixels`?

components/xsd-fu/python/ome/modeltools/language.py
```diff
@@ -347,6 +361,7 @@ def __init__(self, namespace, templatepath):
         self.primitive_type_map[namespace + 'double'] = 'double'
         self.primitive_type_map[namespace + 'anyURI'] = 'std::string'
         self.primitive_type_map[namespace + 'hexBinary'] = 'std::string'
+        self.primitive_type_map['base64Binary'] = 'std::string'
```
rleigh-dundee: Depending upon the outcome of the comment above regarding having Base64 in the API, this might need adjusting.  Ideally it would use the VariantPixelBuffer type, but that would be a layering violation; I'd like to move this out of the ome::files library into a separate repository to allow reuse in the lower levels.  Maybe it would be better as a model primitive, with PixelBuffer in ome::common or another separate nd library?

components/xsd-fu/templates-cpp/AggregateMetadata.template
```diff
@@ -397,27 +399,29 @@ namespace ome
         // -- Entity counting (manual definitions) --
 
 {% end debug %}\
+{% if len(model.opts.lang.getSubstitutionTypes()) == 0 %}\
 {% if fu.SOURCE_TYPE == "header" %}\
-        // Documented in base class.
-        index_type
-        getPixelsBinDataCount(index_type imageIndex) const;
+        // Documented in base class.
```
rleigh-dundee: There's a lot of trailing whitespace to clean up here.  Please do a `git grep '[[:space:]]$'` over your changes.

rleigh-dundee:
I've had initial look over and noted some things for consideration.  We can discuss them when you're back, and I'll be happy to look over in more detail once some of the needed changes have been made.  Overall though, looks like great progress, thanks!

components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
dgault: It certainly does get reused a lot and a nicer solution would be preferred. The logic was originally found in the hardcoded getPixelsBinDataBigEndian, I would have preferred to try and and keep these sort of exceptions outside of the codegen if possible, otherwise it would almost be simpler to go back to the hardcoded function. I guess ideally it would seem to make sense to only have a single value of big endian on the Pixels object.

components/formats-bsd/src/loci/formats/in/FakeReader.java
dgault: There is already some new logic around that attribute in both the update and asXML functions, so it should be straightforward to make the change to encode and decode within the BinData object itself if that is the desired solution.

components/specification/released-schema/2016-DEV0/ome.xsd
dgault: If that makes the most sense then it is an easy change from the model code gen perspective. I had left it for now as it matched up with the api calls unlike for Mask, an update to apply the restriction on Pixels would change the API functions to remove the binDataIndex parameter.

sbesson: @rleigh-dundee: what would be the impact of this additional constraint on our samples using multiple BinData elements? Given the impact of these changes when it comes to propagating across the stack, I am inclined to focus our efforts on `MaskBinData` as initially proposed and have a separate discussion about `PixelsBinData`.

rleigh-dundee: @sbesson I'm not sure, I was just asking the question in case there was value in consistency here.  Leaving it is absolutely fine.

components/xsd-fu/templates-java/OMEXMLModelObject.template
```diff
@@ -221,7 +223,8 @@ ${customUpdatePropertyContent[prop.name]}
 		String value_textContent = element.getTextContent();
 		if (value_textContent.trim().length() > 0) {
 {% if prop.type in model.opts.lang.primitive_base_types %}\
```
rleigh-dundee: Does this conditional (and the other instances of it) need `&& prop.type == "base64Binary"` so that it will avoid breakage if we add additional types to `primitive_base_types` in the future?

rleigh-dundee:
All the changes look good.  I've commented on the use of primitive_base_types in conditionals above, and there are some trivial whitespace issues in:

- cpp MetadataRetrieve/OMEXMLMetadata templates (added/removed newlines between methods)
- indentation in BinData.java (` byte[] rawBytes = DatatypeConverter.parseBase64Binary` and `String encodedString = DatatypeConverter.printBase64Binary`)
- `import javax.xml.bind.DatatypeConverter;` is present in all generated model objects, but is only needed by BinData; the import could be made conditional
- Removed whitespace after `// Parents: ...` in the Java metadata classes

If the above whitespace issues could be fixed up, it would keep the structure of the generated code clean and keep the changes to a minimum for review.

That's all minor stuff though, looks like the PR is good for all the main changes.

Thanks,
Roger

dgault:
Ive updated with changes based on feedback from @rleigh-dundee 
The one thing I haven't changed yet is the indentation on the Java metadata. The incorrect use of spaces rather than tabs has a large number of occurrences, not just from this PR. A separate PR to tidy these up may be the best option.

sbesson:
Thanks for the review. Given the current state of https://ci.openmicroscopy.org/view/Breaking/job/BIOFORMATS-DEV-breaking-trigger/175/, removing the breaking label from this PR so that it can be included in the next merge build /cc @mtbc. As discussed, @rleigh-dundee should be able to use the breaking queue for his units support work.

rleigh-dundee:
Thanks.  There are some remaining whitespace issues in the java diffs, but they are minor and we could fix them up later if we wanted.

@sbesson I think this should be OK to merge if you're also happy with it.

sbesson:
@rleigh-dundee: thanks for the thorough review, we are investigating issues in the OMERO integration jobs now this PR is exposed across the stack so I would prefer to wait until we have some more information on this front to get this PR in. /cc @mtbc But let me know if keeping this open for another few days blocks your progress on the C++ front.

rleigh-dundee:
It looks like I'll have to rebase my current units work on top of this branch merged into develop, then rebase it back onto develop once this PR is merged.  Shouldn't block anything though.

rleigh-dundee:
@sbesson I now have a branch waiting to open a PR for, but this is now a blocker since there's a small conflict.  If the stubs @mtbc added are OK for OMERO (looked fine to me), then would it be possible to merge these?  Once that's done, I can rebase my branch and open a PR.

sbesson:
Thanks for the heads up Roger. We are having issues with a few OMERO integration tests (ExporterTest) which would be good to narrow down before we get this merged. Let's have a discussion tomorrow on the current state of this effort so that we unblock your work but also know what needs to be addressed.

sbesson:
Closing/reopening to include new tests from #2401.

pwalczysko:
Bug: Clients do not handle masks imported/exported as ome-tiff and reimported to OMERO from different softwares (which display fine in Web and Insight after first import as original formats)
 - from previously imported ``...mias/frans`` plate, export the well A1 as ome.tiff (need to use Batch image export script for this selecting ome-tiff in the script window). e.g. use https://eel.openmicroscopy.org/merge/webclient/?show=well-313059 -> this image has 184 ROIs and 11 of them are masks which load and display successfully in the clients)
 - reimport this ome-tiff, and observe that
       - Web shows correct ROI count in the RHP, but in full viewer and Show ROIs -> endless loading is happening
       - Insight shows also correct ROI count in the RHP, but in Measurement Tool, when clicked, warning "Cannot load ROIs" appear followed by much diminished amount of ROIs shown in the MT table (6 out of 184 and no masks)

Prior to export
screen shot 2016-05-13 at 11 19 15

After reimport of ome-tiff in RHP insight ROI count is okay 
screen shot 2016-05-13 at 11 21 16

But the reimported image does not load ROIs in Measurement Tool
screen shot 2016-05-13 at 11 22 29

Web is just loading forever, not displaying anything (ROI count in RHP also correct though)

screen shot 2016-05-13 at 11 23 39

pwalczysko:
Trying to repeat the above workflow of mias/frans & masks export and reimport, see  https://github.com/openmicroscopy/bioformats/pull/2319#issuecomment-219007029 on **eel latest** failed completely , because already at the export as ome-tiff stage the Batch image export script reported failure on the masks export:

```
WARNING:omero.gateway:InternalException on <class 'omero.gateway.OmeroGatewaySafeCallWrapper'> to <a97d9d36-410e-4dd7-9843-2fb76a64e7ddomero.api.Exporter> generateTiff((<ServiceOptsDict: {'omero.session.uuid': 'c919a601-bb30-4bf7-9c8a-375c9e86b099', 'omero.client.uuid': '5b0328a5-6df7-4d9b-ab4f-d63eed9d97f3'}>,), {})
Traceback (most recent call last):
  File "/opt/hudson/workspace/OMERO-DEV-latest-deploy/src/dist/lib/python/omero/gateway/__init__.py", line 4212, in __call__
    return self.f(*args, **kwargs)
  File "/opt/hudson/workspace/OMERO-DEV-latest-deploy/src/dist/lib/python/omero_api_Exporter_ice.py", line 170, in generateTiff
    return _M_omero.api.Exporter._op_generateTiff.invoke(self, ((), _ctx))
InternalException: exception ::omero::InternalException
{
    serverStackTrace = java.lang.ClassCastException: ome.xml.model.Mask cannot be cast to ome.xml.model.Dichroic
	at ome.xml.model.LightPath.link(LightPath.java:210)
	at ome.xml.model.OMEModelImpl.resolveReferences(OMEModelImpl.java:147)
	at ome.xml.meta.OMEXMLMetadataImpl.resolveReferences(OMEXMLMetadataImpl.java:108)
	at ome.xml.meta.OMEXMLMetadataImpl.dumpXML(OMEXMLMetadataImpl.java:102)
	at loci.formats.services.OMEXMLServiceImpl.getOMEXML(OMEXMLServiceImpl.java:464)
	at ome.services.blitz.impl.ExporterI.getMetadataBytes(ExporterI.java:560)
	at ome.services.blitz.impl.ExporterI.access$400(ExporterI.java:76)
	at ome.services.blitz.impl.ExporterI$2.doWork(ExporterI.java:382)
	at sun.reflect.GeneratedMethodAccessor306.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
...
```

pwalczysko:
The images exported from the clients (the mias/frans well A1 was exported using Batch image export from Web, the ``image manipulated by emils script`` was exported from insight using the ``export as ome-tiff`` option) are located on

``team/pwalczysko//PR2319_of_dgault``

pwalczysko:
Bug: ``Emils script created mask case``
  - have a mask on leica image created by https://gist.github.com/emilroz/3858429bccd3012bf478 script (python script which uploads a mask created by other means (e.g. in ImageJ) and attaches it to an image already sitting on an OMERO server) -> note you cannot simply use insight-ij plugin for this workflow, but rather save the mask locally directly from IJ and then use the script to upload the mask to OMERO
  - export the image as ome-tiff, reimport it back to OMERO
  - observe exactly analogous behaviour as described for the mias/frans plate well A1, see https://github.com/openmicroscopy/bioformats/pull/2319#issuecomment-219007029

pwalczysko:
xml header masks ome-tiff workflow:
  - take ``/ome/data_repo/inbox/ome-tiff/mask-bin-data/single-channel-bindata.ome.tif`` and import it to eel merge using insight
  - inspect it using insight and Web, observe that the clients report 0 ROIs and consequently you cannot see any mask in either of them
  - export the image as ome-tiff using web
  - reimport the resulting image to omero using insight, no surprises there, again it has 0 ROIs
  - the location of the image reexported from web see https://github.com/openmicroscopy/bioformats/pull/2319#issuecomment-219016316

sbesson:
Having discussed the current status at length with @pwalczysko, @dgault, @joshmoore and @rleigh-dunde. From the OMERO side, this PR does not show any regression for our Mask/ROI workflows and opens the door to interesting user-facing functionalities such as full round-tripping of images with ROIs containing masks /cc @jburel. 
On the BF side, all tests are green and the full-repository job has been passing. It might be worth reviewing the MIASReader (and other readers) to review/enable the mask population in setId. Next step is to build on top of these efforts on the C++ front to make progress on the support of the 2015-01model.

Congratulations on another epic codegen PR.

PR #2307, opened 2016-03-22
Transforms - String tokenize xslt 1.0

dgault:
This change is required as the Xerces processor does not support the previously implemented XSLT 2.0 solution. Issue was raised in a previous PR - https://github.com/openmicroscopy/bioformats/pull/2286#issuecomment-197889060

To test:
	•	Ensure that the Schema Upgrade tests run successfully

PR #2304, opened 2016-03-22
Fixes for the APL Reader

dgault:
#2307 This PR is a follow up to the user reported issues in https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8013#p16861
The associated ticket is http://trac.openmicroscopy.org/ome/ticket/13174

There are a number of associated QA files provided in QA-17057 and QA-17107

Testing showed a range off issues causing various failures. This PR will address the following fixes:
- Files which do not have all the dimensions listed will be defaulted
- Files which have filename and file path switched, if the filename does not exist the filepath will be parsed to retrieve the filename 
- Files where the filename is unreadable, some of the data from the mtb files are using unkown charsets. Without being able to tell which charset is used the values, most importantly the filename will be incorrect. In this case the filepath can also be parsed to retrieve the filename. 
- Previously we had retrieved the channel count from the column 'Color Channels'. In some of the provided sample files this is not present but the channel count can instead be found by checking the column 'Image Type' for the values RGB or Mono

To test:
Verify that the provided QA files load sucessfullly
- The files in Plain.zip can be used to test the default dimensions
- The files in LifeScienceDemo will test the instance when filename and path are swapped
- The files in Example2 show the case with the unidentified charset
- The files in MaterialsDemo use the ImageType column instead of Color Channels

Note: The files in ExampleDatabase still do not load due to a remaining unresolved issue

PR #2322, opened 2016-04-04
Fixes for the APL Reader (rebased onto dev_5_1)

dgault:
This is the same as [gh-2304](https://github.com/openmicroscopy/bioformats/pull/2304) but rebased onto dev_5_1.

----

#2307 This PR is a follow up to the user reported issues in https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8013#p16861
The associated ticket is http://trac.openmicroscopy.org/ome/ticket/13174

There are a number of associated QA files provided in QA-17057 and QA-17107

Testing showed a range off issues causing various failures. This PR will address the following fixes:

Files which do not have all the dimensions listed will be defaulted
Files which have filename and file path switched, if the filename does not exist the filepath will be parsed to retrieve the filename
Files where the filename is unreadable, some of the data from the mtb files are using unkown charsets. Without being able to tell which charset is used the values, most importantly the filename will be incorrect. In this case the filepath can also be parsed to retrieve the filename.
Previously we had retrieved the channel count from the column 'Color Channels'. In some of the provided sample files this is not present but the channel count can instead be found by checking the column 'Image Type' for the values RGB or Mono
To test:
Verify that the provided QA files load sucessfullly

The files in Plain.zip can be used to test the default dimensions
The files in LifeScienceDemo will test the instance when filename and path are swapped
The files in Example2 show the case with the unidentified charset
The files in MaterialsDemo use the ImageType column instead of Color Channels
Note: The files in ExampleDatabase still do not load due to a remaining unresolved issue

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/APLReader.java
+++ b/components/formats-gpl/src/loci/formats/in/APLReader.java
@@ -302,10 +302,14 @@ public class APLReader extends FormatReader {
     int calibratedHeight = DataTools.indexOf(columnNames, "Height");
     int calibratedWidth = DataTools.indexOf(columnNames, "Width");
     int path = DataTools.indexOf(columnNames, "Image Path");
+    if (path == -1) {
+      path = DataTools.indexOf(columnNames, "Path");
+    }
     int filename = DataTools.indexOf(columnNames, "File Name");
     int magnification = DataTools.indexOf(columnNames, "Magnification");
     int width = DataTools.indexOf(columnNames, "X-Resolution");
     int height = DataTools.indexOf(columnNames, "Y-Resolution");
+    int imageType = DataTools.indexOf(columnNames, "Image Type");
     int imageName = DataTools.indexOf(columnNames, "Image Name");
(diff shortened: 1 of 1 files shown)


dgault:
--rebased-from https://github.com/openmicroscopy/bioformats/pull/2304

bramalingam:
https://ci.openmicroscopy.org/view/Bio-Formats-5.1/job/BIOFORMATS-5.1-merge-repository-subset/234/ is green.

bramalingam:
Except for 

```
ls29010:bftools-7 bramalingam$ ./showinf /Users/bramalingam/Desktop/QA/17107/ExampleDatabase/ExampleDatabase.apl 
Checking file format [Olympus APL]
Initializing reader
APLReader initializing /Users/bramalingam/Desktop/QA/17107/ExampleDatabase/ExampleDatabase.apl
```

All other files under the QA posts pass. And this is an expected failure as well, since this PR just covers a fraction of the issues. The rest to follow in a separate PR.

This PR in its current state looks good to merge.

@dgault

PR #2323, opened 2016-04-04
CZI reader bugfix : For wrongly identified master-files in multi-file datasets

bramalingam:
ref: https://trac.openmicroscopy.org/ome/ticket/13173

Testing Instructions : 
Try opening all the files individually from QA 17056 on FIJI/ImageJ. Without this PR, all files under the datasets "WT 12 *.czi and WT 0 *.czi" will look the same as : "WT 12.czi and WT 0.czi" respectively.
With this PR the pixel data should look different for the individual files. Note: This is not a multi-file dataset, its just a random coincidence that the user has named them like this.

Please try opening a multi-file dataset as well : QA 7271 and QA 7276
and check if we still open these files properly. (This test should not fail and should open the pixeldata without any issues).

cc : @melissalinkert @jburel

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -140,6 +140,10 @@ public class ZeissCZIReader extends FormatReader {
 
   private String zoom;
   private String gain;
+  
+  private String masterFile;
+  private String userSelectedFile;
+  private Boolean wrongMasterFile = false;
 
   private ArrayList<Channel> channels = new ArrayList<Channel>();
   private ArrayList<String> binnings = new ArrayList<String>();
@@ -463,7 +467,8 @@ public class ZeissCZIReader extends FormatReader {
 
     // switch to the master file if this is part of a multi-file dataset
     String base = id.substring(0, id.lastIndexOf("."));
-    if (base.endsWith(")") && isGroupFiles()) {
+    System.out.println(!wrongMasterFile);
+    if (base.endsWith(")") && isGroupFiles() && !wrongMasterFile) {
       LOGGER.info("Checking for master file");
       int lastFileSeparator = base.lastIndexOf(File.separator);
       int end = base.lastIndexOf(" (");
@@ -474,6 +479,8 @@ public class ZeissCZIReader extends FormatReader {
         base = base.substring(0, end) + ".czi";
         if (new Location(base).exists()) {
           LOGGER.info("Initializing master file {}", base);
+          masterFile = base;
+          userSelectedFile = id;
           initFile(base);
           return;
         }
@@ -619,6 +626,12 @@ public class ZeissCZIReader extends FormatReader {
     // finish populating the core metadata
 
     int seriesCount = positions * acquisitions * mosaics * angles;
+
+    if (seriesCount == 1 && masterFile != null && !wrongMasterFile){
+        wrongMasterFile = true;
+        System.out.println(userSelectedFile);
+        initFile(userSelectedFile);
+    }
 
     ms0.imageCount = getSizeZ() * (isRGB() ? 1 : getSizeC()) * getSizeT();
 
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -140,6 +140,10 @@
 
   private String zoom;
   private String gain;
+  
+  private String masterFile;
+  private String userSelectedFile;
+  private Boolean wrongMasterFile = false;
```
sbesson: Should these be reset in the `close()` method?

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -617,8 +625,16 @@ protected void initFile(String id) throws FormatException, IOException {
     ms0.sizeT *= phases;
 
     // finish populating the core metadata
-
     int seriesCount = positions * acquisitions * mosaics * angles;
+    //Checkpoint for identifying wrong masterfiles.
+    //Current assumption: Datasets with master files will have seriesCount>1,
+    //This strategy is not the safest way, will need to be reviewed in future.
+    if (seriesCount == 1 && masterFile != null && !wrongMasterFile){
```
sbesson:
```
sbesson@c000176:apache_repo $ bftools/showinf -autoscale -nometa 17056/WT\ 0\ \(2\).czi 
Checking file format [Zeiss CZI]
Initializing reader
ZeissCZIReader initializing 17056/WT 0 (2).czi
Checking for master file
Initializing master file 17056/WT 0.czi
Initialization took 0.097s

Reading core metadata
filename = 17056/WT 0 (2).czi
...
````

Similarly to what's been done above, we might want to add a logging statement explaining that we are switching back to the selected file else this is confusing.

**Comment by sbesson** [comment]
Testing using the input QA file (which we might want to add it to the curated data repository) went as expected and all other QA files using a master file have been successfully tested without regressions by https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/189/.

Apart from my logging comment, my main concern is the non reinitialization of the new variables used to track the master file identification. It seems like re-using the reader could lead to unwanted state, e.g.:

```
reader = new ZeissCZIReader();
reader.setId('file_with_wrong_master (2).czi');
reader.close();
reader.setId('file_with_real_master (2).czi')
```

would have `wrongMasterFile` incorrectly set to `true`.

**Comment by bramalingam** [comment]
Thanks Seb. Resetting the variables definitely something I am looking at. And the wrongMasterFile will be set to true, if and only if the seriesCount==1 (Which can definitely happen theoretically, but we can definitely deal with that later, if we can get hold of datasets satisfying this condition).

**Comment by bramalingam** [comment]
Ahhh I see your point the above issue that you have discussed is completely related to resetting the variables

**Comment by mtbc** [comment]
Should this be relisted for review?

**Comment by melissalinkert** [comment]
I think the last commit here broke things.  I now see:

```
$ showinf -version
Version: 5.2.0-DEV
VCS revision: d451f5f13e660cfb9875ddcce0dcf0080e6e4d45
Build date: 20 April 2016
$ showinf -nopix WT\ 0\ \(2\).czi
Checking file format [Zeiss CZI]
Initializing reader
ZeissCZIReader initializing WT 0 (2).czi
Checking for master file
Initializing master file WT 0.czi
Initialization took 0.82s

Reading core metadata
filename = WT 0.czi
Used files:
        WT 0.czi
        /home/melissa/qa-17056/WT 0 (2).czi
        /home/melissa/qa-17056/WT 0 (3).czi
        /home/melissa/qa-17056/WT 0 (4).czi
Series count = 1
Series #0 :
        Image count = 2
        RGB = false (1) 
        Interleaved = false
        Indexed = true (false color, 8-bit LUT: 3 x 256)
        Width = 1024
        Height = 1024
        SizeZ = 1
        SizeT = 1
        SizeC = 2
        Thumbnail size = 128 x 128
        Endianness = intel (little)
        Pixel type = uint8
        Valid bits per pixel = 8
        Metadata complete = false
        Thumbnail series = false
        -----
        Plane #0 <=> Z 0, C 0, T 0
        Plane #1 <=> Z 0, C 1, T 0
```

when I'd expect the used file list to just include ```WT 0 (2).czi``` (with no logging to indicate the master file was used).

sbesson:
Closing. Can be reopened when the fix is pushed,

PR #2324, opened 2016-04-05
Hamamatsu VMS: fix dimensions of full-resolution image

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/11535.

To test, use the three .vms files from ```curated/hamamatsu-vms/openslide```, and compare against the three .ndpi files from ```curated/hamamatsu/openslide```.  The full resolution image for the .ndpi files (== series 0) should be a factor of 2 smaller than the full resolution image for the corresponding .vms.  The whole image can't be opened with showinf/ImageJ, but spot checking several (x, y, w, h) .ndpi tiles against the corresponding (x * 2, y * 2, w * 2, h * 2) .vms tile should give confidence that the tiling is now being done correctly.

Importing into OMERO would also be something to check, but note that this should result in pyramid generation since pyramids are not stored in these .vms datasets.  Not sure how easy that would be with an open PR, so if it's easier to test in OMERO after merging and follow up with a card (if needed) then that works for me.

Also, this is a breaking reader change, so likely not a backporting candidate.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/HamamatsuVMSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/HamamatsuVMSReader.java
@@ -38,6 +38,7 @@ import loci.common.IniParser;
 import loci.common.IniTable;
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
+import loci.common.Region;
 import loci.common.services.ServiceException;
 import loci.formats.CoreMetadata;
 import loci.formats.FormatException;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Corresponding config PR: https://github.com/openmicroscopy/data_repo_config/pull/101

sbesson:
Added to https://trello.com/c/NE6UFb5k/70-better-document-breaking-reader-changes

mtbc:
Needs listing for review?

melissalinkert:
Listed for Monday now.

bramalingam:
Did the non-omero testing on Matlab. Works as expected. As suggested doing the Omero testing is not viable currently. So this PR is good to merge in its current state.

We can always do the Omero import testing at a later state, when we start consuming BF snapshot's in develop.

melissalinkert:
Any reason not to merge this?

mtbc:
We can test open BF PRs via the regions devspace: this morning's build merges,
```
2016-05-06 04:01:09,632 [   scc.merge] INFO    # PR 2324 melissalinkert 'Hamamatsu VMS: fix dimensions of full-resolution image'
```
I have imported these files into `PR-2324-ndpi` and `PR-2324-vms` on `regions-omero.docker.openmicroscopy.org` as `user-14`.

mtbc:
> should be a factor of 2 smaller

does seem to be true. :+1:

mtbc:
At least the first image looks the same in both. (Pyramids take ages.)

PR #2325, opened 2016-04-06
PerkinElmer UltraVIEW VoX link

mtbc:
fixes the broken link to the microscope information

PR #2327, opened 2016-04-06
PerkinElmer UltraVIEW VoX link (rebased onto dev_5_1)

sbesson:
This is the same as gh-2325 but rebased onto dev_5_1.

----

fixes the broken link to the microscope information


sbesson:
--rebased-from #2325

sbesson:
See https://ci.openmicroscopy.org/view/Bio-Formats-5.1/job/BIOFORMATS-5.1-merge-docs/1209/

PR #2326, opened 2016-04-06
remove mentions of deleted project SubImager

mtbc:
The CellProfiler folks have deleted the subimager repo as not having been a good idea in the first place. Perhaps in some later PR we may wish to suggest some alternative.

PR #2328, opened 2016-04-06
remove mentions of deleted project SubImager (rebased onto dev_5_1)

sbesson:
This is the same as gh-2326 but rebased onto dev_5_1.

----

The CellProfiler folks have deleted the subimager repo as not having been a good idea in the first place. Perhaps in some later PR we may wish to suggest some alternative.


sbesson:
--rebased-from #2326

sbesson:
See https://ci.openmicroscopy.org/view/Bio-Formats-5.1/job/BIOFORMATS-5.1-merge-docs/1209/

PR #2329, opened 2016-04-07
Zeiss CZI: fixes for files with multiple mosaics and positions (rebased onto dev_5_1)

bramalingam:
This is the same as gh-2305 but rebased onto dev_5_1.

----

To test, use the file in ```curated/zeiss-czi/niko``` (it will be ~100GB).  The images are quite large; the easiest first test would be to select a few ~2kx2k tiles and check the display in ImageJ vs Zeiss' ZEN.  It would be good to try importing into OMERO as well, to validate that all of the pixel data is read correctly, but this will take a while as pyramids will need to be generated.

Configuration PR is: https://github.com/openmicroscopy/data_repo_config/pull/96


bramalingam:
--rebased-from #2305

PR #2330, opened 2016-04-07
Zeiss CZI: 5.1.9 backports

sbesson:
This is the same as gh-2268 but rebased onto dev_5_1.

----

Fixes https://trac.openmicroscopy.org/ome/ticket/13095

To test, use the file from QA 16882.  Without this change, ```showinf -minmax``` should show that the first series has non-zero pixels, and the remaining four series have all zero pixels.  With this change, the same test should show that all five series have non-zero pixels (which should also be obvious from looking at the images).  The images and dimensions can also be compared against Zeiss' ZEN software.

This is the same as gh-2305 but rebased onto dev_5_1.

----

To test, use the file in curated/zeiss-czi/niko (it will be ~100GB). The images are quite large; the easiest first test would be to select a few ~2kx2k tiles and check the display in ImageJ vs Zeiss' ZEN. It would be good to try importing into OMERO as well, to validate that all of the pixel data is read correctly, but this will take a while as pyramids will need to be generated.

Configuration PR is: openmicroscopy/data_repo_config#96

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -631,7 +631,7 @@ public class ZeissCZIReader extends FormatReader {
     LOGGER.trace("prestitched = {}", prestitched);
     LOGGER.trace("scanDim = {}", scanDim);
 
-    if (mosaics == seriesCount &&
+    if (((mosaics == seriesCount) || (positions == seriesCount)) &&
       seriesCount == (planes.size() / getImageCount()) &&
       prestitched != null && prestitched)
     {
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #2268

sbesson:
--rebased-from #2305

bramalingam:
QA 16882 with this PR on omero, 
screen shot 2016-04-08 at 14 52 52

and
curated/zeiss-czi/niko:
Same as : https://github.com/openmicroscopy/bioformats/pull/2305#issuecomment-206295542

So the manual testing passes as well.
Along with the subset jobs,
https://ci.openmicroscopy.org/job/BIOFORMATS-5.1-merge-repository-subset/235/console

Looks good to merge.

@sbesson

PR #2331, opened 2016-04-07
Fake: fix HCS fake files with multiple screens

sbesson:
See https://trello.com/c/pBQgFTMS/117-fakereader-fix-microbeam-with-multiple-screens

This PR fixes an indexing bug in the `XMLMockObjects.createPlate()` method which caused all fake HCS files with multiple screens to be invalid as multiple `Experiment` objects were created with the same ID. Unit tests are also added to `FakeReaderTest` to cover the multiple screens test case and include some testing of the `MicrobeamManipulation` option.

To test these changes, check fake HCS files with more than one screen have a valid OME-XML:

```
$ showinf -nopix -omexml "SPW&screens=2.fake
$ showinf -nopix -omexml "SPW&screens=2&plates=3.fake
$ showinf -nopix -omexml "SPW&screens=2&withMicrobeam=true.fake
```


dgault:
Changes look good and FakeReaderTest passes. Tested that fake files validate with multiple screens, plates and withMicrobeam, all validated with no errors. Looks good to merge.

PR #2282, opened 2016-03-15
Improved OutOfMemory handling in automated tests

melissalinkert:
See builds 99-103 of https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-full-repository/

This does a better job of detecting OOMs in thumbnail tests, as well as more accurately reporting the amount of free memory available.

While the latest builds are failing primarily on new CellSens datasets, I wasn't able to find any obvious memory leaks specific to CellSens; as discussed with @joshmoore and @sbesson, it's likely that we're reaching the limit of how much data can reasonably be tested at once.

Apart from code review, the easiest way to test this is likely to observe all of the jobs for a week or two to verify that OOM-related failures are no longer present.

PR #2333, opened 2016-04-09
Add ROI import to ImageJ overlay

stelfrich:
The PR adds a choice for the user to determine whether to import ROIs to the ImageJ ROI manager or create a new overlay and add them to that. The choice is disabled if the _Display ROIs_ is disabled.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/importer-options.txt
+++ b/components/bio-formats-plugins/src/loci/plugins/in/importer-options.txt
@@ -192,6 +192,17 @@ info = <b>Display ROIs</b> - \
   Adds any ROIs in the file to ImageJ's ROI manager.
 default = false
 
+[roiMode]
+label = ROIs_Import Mode
+info = <b>ROI Import Mode</b> - \
+  ROIs can be imported to the ROI Manager or an Overlay.
+type = string
+save = false
+default = ROI Manager
+values = \
+  ROI Manager, \
+  Overlay
+
 [specifyRanges]
 type = boolean
 label = Specify_range for each series
--- a/components/bio-formats-plugins/src/loci/plugins/in/importer-options.txt
+++ b/components/bio-formats-plugins/src/loci/plugins/in/importer-options.txt
@@ -189,13 +189,13 @@ default = false
 type = boolean
 label = Display_ROIs
 info = <b>Display ROIs</b> - \
-  Adds any ROIs in the file to ImageJ's ROI manager.
+  Adds any ROIs in the file to ImageJ.
 default = false
 
 [roiMode]
-label = ROIs_Import Mode
+label = ROIs_Import Mode:
 info = <b>ROI Import Mode</b> - \
-  ROIs can be imported to the ROI Manager or an Overlay.
+  ROIs can be imported to the ROI manager or an overlay.
 type = string
 default = ROI Manager
 values = \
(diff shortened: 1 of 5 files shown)


stelfrich:
There are failing tests for the merge to develop: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-full-repository/130/changes

I am a little lost on how to investigate these issues..

sbesson:
@stelfrich: the failures in the job are due to `OutOfMemory` exceptions (see https://github.com/openmicroscopy/bioformats/pull/2282). More generally, the repository jobs are testing regressions due to changes in the readers under `formats-gpl/formats-bsd`. As your changes are primarily modifying the ImageJ/Fiji plugin, you can safely ignore these failures in the context of this PR. Thanks for the active checking though.

dgault:
@stelfrich: Hi Stefan, I have finished testing out your PR today. The feature certainly looks like it could be very useful and the dialog additions and overlay option look good and are all working well.

The only issue I encountered was when using the ROI Manager selection, the ROI Handler wasnt  recognising the selection in roisMode.equals(ImporterOptions.ROIS_MODE_MANAGER) 

This can be resolved if you update the 2 instances of the above in ROI Handler with:
roisMode.equalsIgnoreCase(ImporterOptions.ROIS_MODE_MANAGER)

stelfrich:
Thanks for the feedback, @dgault! I have integrated your comment.

> The only issue I encountered was when using the ROI Manager selection, the ROI Handler wasnt recognising the selection in `roisMode.equals(ImporterOptions.ROIS_MODE_MANAGER)`

How did you notice (just out of curiosity)? Everything works fine for me with `equals`.

dgault:
Not sure what the difference was but for me the default value for the option was "ROI Manager" and the ROIS_MODE_MANAGER Sting was using "ROI manager" with a lowercase m. 

I've retested with the update and everything is working as expected. PR looks good to merge from my perspective.

components/bio-formats-plugins/src/loci/plugins/in/importer-options.txt
```diff
@@ -189,9 +189,19 @@ default = false
...
-  Adds any ROIs in the file to ImageJ's ROI manager.
+  Adds any ROIs in the file to ImageJ.
 default = false
 
+[roiMode]
+label = ROIs_Import Mode:
+info = <b>ROI Import Mode</b> - \
+  ROIs can be imported to the ROI manager or an overlay.
+type = string
+default = ROI Manager
```
sbesson: On top of the `equalsIgnoreCase` fix added in https://github.com/openmicroscopy/bioformats/pull/2333/commits/0e45bd4b44e61dd57404a35418b496a157159e88, is it worth using the same nomenclature as `ROIS_MODE_MANAGER` for `default` and `values` i.e. `ROI manager` with lower case `m`?

stelfrich: I also found that yesterday, @sbesson, but decided to leave the fix in place. Would you rather have the naming consistent with `equals`?

sbesson: @stelfrich: thinking mostly in terms of maintenance, imposing a consistent naming scheme might minimize similar issues if the new `roiMode` gets reused anywhere else so I would vote for that.

stelfrich: I have updated the PR to fix that inconsistency. Thanks for the feedback, @sbesson!

sbesson:
Retested with the last merge build with a fake file with ellipses

screen shot 2016-04-14 at 08 53 56

Everything works as expected. Thanks @stelfrich.

PR #2337, opened 2016-04-11
Add officially supported versions to the LaVision Imspector format

sbesson:
Our current representative set of LaVision Imspector files included in the daily automated tests only contain files generated with the 4.0 and 4.1 versions of ImspectorPro. /cc @imunro

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -1222,6 +1222,7 @@ Commercial applications that support LIF include: \n
 [LaVision Imspector]
 extensions = .msr
 developer = `LaVision BioTec <http://www.lavisionbiotec.com/>`_
+versions = 4.0, 4.1
 bsd = no
 weHave = * a few .msr files
 pixelsRating = Fair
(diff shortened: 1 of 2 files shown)


imunro:
Looks good to me - I would point out though, for completion, that it's possible that non-FLIM files from Ver 5.0 might work fine. I simply don't know.

sbesson:
@imunro: fair enough. This is certainly a statement we could extend to most formats. Does it come down to defining more precisely the meaning of "Officially Supported Versions"? /cc @hflynn

hflynn:
We don't define unsupported anywhere, I would stick with listing only those we *know* work.

To my mind, supported = we test this format and will commit to fixing bugs reported for it

imunro:
I wasn't suggesting a change to the PR just wanted to make sure it was clear how much we don't know. Looks good to merge to me.

hflynn:
Looks fine on staging btw, good to merge from my pov too

bramalingam:
--rebased-to #2339

sbesson:
--rebased-to #2340

Issue #2255, opened 2016-02-19
NIOFileHandle large buffer size detracts from ImageJ importer performance


etarena:
We noticed that when opening OME-TIFF datasets produced by WiscScan, the initialization step took a long time. Profiling the code, we found that much time was spent in `NIOByteBufferProvider` in the `allocateDirect` method, copying data from the `FileChannel` into the buffers. The default read-only buffer size is set to 1048576 bytes, which gets allocated for every TIFF file in the dataset, causing the `channel.read` call to take a long time populating each buffer.

The issue can be significantly mitigated by calling `NIOFileHandler.setDefaultBufferSize(int)` with a smaller value. Here is a quick benchmark of the difference with various buffer sizes, when reading an (fill in size of dataset) OME-TIFF dataset from an SMB networked file system:
```
Buffer size: 16 Process time: 90566
Buffer size: 1024 Process time: 92513
Buffer size: 8192 Process time: 96178
Buffer size: 65536 Process time: 111689
Buffer size: 1048576 Process time: 313106
```
Here is the script which generated the benchmark:
```javascript
var list = newArray(16, 1024, 8192, 65536, 1048576);
for (i = 0; i < list.length; i++) {
	call("loci.common.NIOFileHandle.setDefaultBufferSize",list[i]);
	start = getTime();
	run("Bio-Formats", "open=[/Volumes/data/Jayne/test data sets for Ellen/3ROI 2filters 2 channels 3TP/H150901R1 012016 postW_TL1 2dpfDish1 krtGFP_SHG 20x890nmp50_160_z2_5i2g580f520_445u5ROI3TP6_C0_TP0_SP0_FW0.ome.tiff] autoscale color_mode=Default open_all_series view=Hyperstack stack_order=XYCZT");
	end = getTime();
	print("Buffer size: " + list[i] + " Process time: " + (end - start));
	run("Close All");
}
```
You can find the data at the above path on LOCI ftp accessible via the usual credentials (just message a LOCI programmer if you need them). 

As you can see above, the smaller the buffer size the faster the data is opened within ImageJ.  We could certainly hardcode the buffer size within ImageJ - but we are wondering if you have another idea? 

Many thanks for your insight.

melissalinkert:
Thank you for reporting this.  I don't see the dataset after logging in to dev.loci.wisc.edu (the last ftp server known to me) - could you please upload the files to https://www.openmicroscopy.org/qa2/qa/upload/ instead?

[...]

PR #2338, opened 2016-04-11
NIOFileHandle: reduce default buffer size

melissalinkert:
See https://github.com/openmicroscopy/bioformats/issues/2255

Compared two consecutive runs of ```test-automated``` on necromancer (4096MB, 4 threads) with 4 directories.  Without this PR:

run | directory | scan time | total time
-----|-------------|-------------|--------------
0 | test_images_good | 107.38s | 32m26s
1 | test_images_good | 74.706s | 33m53s
0 | curated/nd2 | 31.481s | 79m3s
1 | curated/nd2 | 27.971s | 70m31s
0 | curated/zeiss-zvi | 2.953s | 20m42s
1 | curated/zeiss-zvi | 2.266s | 19m4s
0 | curated/ome-tiff | 19.218s | 31m48s
1 | curated/ome-tiff | 14.13s | 28m48s

With this PR:

run | directory | scan time | total time
-----|-------------|-------------|--------------
0 | test_images_good | 42.031s | 26m7s
1 | test_images_good | 34.343s | 23m29s
0 | curated/nd2 | 18.659s | 73m48s
1 | curated/nd2 | 15.17s | 62m42s
0 | curated/zeiss-zvi | 3.168s | 21m2s
1 | curated/zeiss-zvi | 2.074s | 16m30s
0 | curated/ome-tiff | 19.402s | 20m8s
1 | curated/ome-tiff | 9.293s | 18m49s

For each directory/table, run 1 was executed immediately after run 0.  Tests with this PR were run on a separate day from tests without this PR, to reduce impact of filesystem/OS caching.  The above has me reasonably confident that this won't cause harm; this is another PR that will need to be in the builds for a few days before we consider merging though.  The total run times of all DEV-merge jobs, especially https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-full-repository/, will need to be monitored - I would expect some reduction on average for anything running tests against data_repo.

### DIFF
--- a/components/formats-common/src/loci/common/NIOFileHandle.java
+++ b/components/formats-common/src/loci/common/NIOFileHandle.java
@@ -63,7 +63,7 @@ public class NIOFileHandle extends AbstractNIOHandle {
   //-- Static fields --
 
   /** Default NIO buffer size to facilitate buffered I/O. */
-  protected static int defaultBufferSize = 1048576;
+  protected static int defaultBufferSize = 16;
 
   /**
    * Default NIO buffer size to facilitate buffered I/O for read/write streams.


mtbc:
Ready for review?

melissalinkert:
Would be good to have https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-full-repository/ green again before fully reviewing, so that averages of passing builds can be compared.

mtbc:
A few green runs lately.

dgault:
Tests have been green for quite some time with this PR included and it hasn't been responsible for any of the failures. 

Local testing using test-automated and a number of sample directories shows that the average run times after the PR is included are reduced as expected and inline with the initial testing mentioned on the PR. The average times on https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-full-repository/ are a little harder to judge as the run times can almost double on a night to night basis, but the times remain in a similar range to prior to the inclusion of this PR so it is certainly not having any negative impact. 

As such the PR should be good to merge without any concerns and should have a positive impact over the long term.

sbesson:
Thanks for the review @dgault. Discussing with @melissalinkert, the last thing we could consider before merging is a review of the impact of this PR at the OMERO level. Concretely that would mean reviewing the performance at both the import and viewing level for a set of images across various formats. Do we have existing metrics/scenarios that could be reused here @pwalczysko?

pwalczysko:
> ...various formats

All formats ? Big images ? Plates ?

Metrics - stored metrics are very sporadic and scarce. Metrics from when ? (which version of OMERO ?)
If the question is "Do we have something fully ready for the latest OMERO and easy to go...?" then the answer is no.

sbesson:
Two additional comments:
- https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository-win/1 launched to have this change validated on Windows as well
- should we also review all the places where `16` has been passed as the buffer size and use the default `RandomAccessInputStream(String)` constructor instead?

 ```
$ git grep RandomAccess.*16 | wc
      22     137    2803
```

**Comment by melissalinkert** [comment]
:+1: to Windows validation.  I don't think cases of ```new RandomAccessInputStream(..., 16)``` should be changed, as that will change behavior when ```NIOFileHandler.setDefaultBufferSize(...)``` is called.

**Comment by sbesson** [comment]
Once we are done with model changes, the current plan is to come back to this PR and run a bulk import of our set of files for various formats using https://github.com/openmicroscopy/openmicroscopy/pull/4614 rebased into the mainline. To be discussed into more detail with @johsmooore. Note this might move us towards a more streamlined workflow allowing to address questions raised in https://github.com/openmicroscopy/bioformats/pull/2338#issuecomment-219687452 in the future /cc @pwalczysko @jburel

File: components/formats-common/src/loci/common/NIOFileHandle.java [hunk 1]
```diff
@@ -63,7 +63,7 @@
   //-- Static fields --
 
   /** Default NIO buffer size to facilitate buffered I/O. */
-  protected static int defaultBufferSize = 1048576;
+  protected static int defaultBufferSize = 16;
```
mtbc: That's a big difference. I wonder if we ought also to test some intermediate value?

sbesson:
Briefly discussed with @melissalinkert, given the impact of this PR and the team agenda/availabilities for the upcoming weeks, I do not see us doing a proper testing across formats. Instead, the proposal is to reduce the scope of this PR to fixing OME-TIFF only (which is the original issue) and do the testing using our OME-TIFF samples both on Bio-Formats and OMERO (import time + plane access/visualization). Would that be more reasonable in terms of testing burden @pwalczysko?

melissalinkert:
Closing for now in favor of https://github.com/openmicroscopy/bioformats/pull/2489.  The two PRs aren't mutually exclusive, so we can always re-open this if needed/as time permits.

PR #2340, opened 2016-04-12
Add officially supported versions to the LaVision Imspector format (rebased onto dev_5_1)

sbesson:
This is the same as gh-2337 but rebased onto dev_5_1.

----

Our current representative set of LaVision Imspector files included in the daily automated tests only contain files generated with the 4.0 and 4.1 versions of ImspectorPro. /cc @imunro


sbesson:
--rebased-from #2337

PR #2341, opened 2016-04-12
Linking to slidebook6 page

hflynn:
Updates the slidebook format page to point to how to update the new DLLs. For now the /info/slidebook URL needs to redirect to https://www.openmicroscopy.org/site/support/slidebook but once 3i are sorted, it should be changed to point at their page (see #2289 for context)

cc @sbesson - can you set up the redirect please?

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -22,11 +22,8 @@ preserved. \n
 Free software from 3I can export the files to OME-TIFF post-acquisition, see \n
 https://www.slidebook.com/reader.php.\n
 \n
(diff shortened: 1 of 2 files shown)


sbesson:
Redirect into place. Note that the URL does not render as an hyperlink in the generated documentation page as it needs to be prefixed with `http://`.

hflynn:
Link fixed and can confirm redirect works, thanks

hflynn:
http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/formats/3i-slidebook.html - link rendered correctly now and works

hflynn:
--rebased-to #2344

PR #2343, opened 2016-04-13
Resolve Plane population errors for Zeiss LSM and OME-TIFF

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12978

To test, use the following datasets mentioned in the ticket:

* QA 16996 (OME-TIFF)
* QA 11328 (Zeiss LSM)
* QA 16981 (Zeiss LSM)
* QA 17012 (Zeiss LSM)

Verify that without this change, ```showinf -nopix -omexml``` shows validation errors indicating invalid ```Plane``` elements in the generated OME-XML.  With these changes, there should be no ```Plane``` validation errors.  The Zeiss LSM datasets should also be checked against Zeiss' ZEN Lite software, as the fix for those datasets was to introduce rotation and phase support to the Zeiss LSM reader.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1013,7 +1013,7 @@ public class OMETiffReader extends FormatReader {
       }
     }
 
-    MetadataTools.populatePixels(metadataStore, this, false, false);
+    MetadataTools.populatePixels(metadataStore, this, true, false);
     for (int i=0; i<acquiredDates.length; i++) {
       if (acquiredDates[i] != null) {
         metadataStore.setImageAcquisitionDate(
--- a/components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
@@ -970,11 +970,6 @@ public class ZeissLSMReader extends FormatReader {
       if (splitPlanes) ms.imageCount *= getSizeC();
     }
 
-    for (int c=0; c<getEffectiveSizeC(); c++) {
-      String lsid = MetadataTools.createLSID("Channel", series, c);
-      store.setChannelID(lsid, series, c);
-    }
-
     // NB: the Zeiss LSM 5.5 specification indicates that there should be
     //     15 32-bit integers here; however, there are actually 16 32-bit
     //     integers before the tile position offset.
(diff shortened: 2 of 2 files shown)


melissalinkert:
Configuration PR: https://github.com/openmicroscopy/data_repo_config/pull/106

dgault:
The LSM files all load correctly without any errors and look as expected when comparing with datasets loaded in Zen. 

I can verify that prior to the code changes all 4 QA files reported back validation errors. With the PR applied the 3 Zeiss LSM files verify correctly as expected, however I am still seeing validation errors on the OME Tiff.

The validation issues on the OME Tiff are as below and consist of 2 different errors:
cvc-attribute.3: The value '0/38/0: Theta' of attribute 'ID' on element 'AnnotationRef' is not valid with respect to its type, 'AnnotationID'.
cvc-pattern-valid: Value '0/0/0: Theta' is not facet-valid with respect to pattern '(urn:lsid:([\w\-\.]+\.[\w\-\.]+)+:\S+:\S+)|(\S+:\S+)' for type 'AnnotationID'.
cvc-attribute.3: The value '0/0/0: Theta' of attribute 'ID' on element 'DoubleAnnotation' is not valid with respect to its type, 'AnnotationID'.
cvc-pattern-valid: Value '0/1/0: Theta' is not facet-valid with respect to pattern '(urn:lsid:([\w\-\.]+\.[\w\-\.]+)+:\S+:\S+)|(\S+:\S+)' for type 'AnnotationID'.
Error validating document: 157 errors found

melissalinkert:
The Annotation validation errors are expected - the LSIDs in the file are actually invalid (tiffcomment should show this), and we're not currently doing anything to check and correct them.  I'm inclined to think that's outside the scope of this PR, but it's worth keeping in mind in the larger ongoing "how far do we go to fix invalid OME-TIFFs" discussion (https://trello.com/c/QwV2GwDn/7-metadata-validity-and-validation is the most obvious card I could find, there may be others).

dgault:
That all makes sense now, in that case this PR should be good to merge. I wasn't able to find the QA files under inbox or curated but once those are copied over and the merge subset job run then the configuration PR should be good also.

melissalinkert:
All QA files should be copied under curated now.

melissalinkert:
See https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/211/ and https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/212/

dgault:
With the tests passing then this and the config PR should both be good to merge

PR #2344, opened 2016-04-13
Linking to slidebook6 page (rebased onto dev_5_1)

hflynn:
This is the same as gh-2341 but rebased onto dev_5_1.

----

Updates the slidebook format page to point to how to update the new DLLs. For now the /info/slidebook URL needs to redirect to https://www.openmicroscopy.org/site/support/slidebook but once 3i are sorted, it should be changed to point at their page (see #2289 for context)

cc @sbesson - can you set up the redirect please?


hflynn:
--rebased-from #2341

hflynn:
see http://www.openmicroscopy.org/site/support/bio-formats5.1-staging/formats/3i-slidebook.html
Should be good to merge for the release

PR #2345, opened 2016-04-14
Fixes crash when loading large .nii files

cniedwor:
As discussed in the mailing list, I changed planeSize to long to prevent the seek position calculation from overflowing on large .nii files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -147,7 +147,7 @@ public class NiftiReader extends FormatReader {
   {
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
-    int planeSize = FormatTools.getPlaneSize(this);
+    long planeSize = FormatTools.getPlaneSize(this);
     pixelFile.seek(pixelOffset + no * planeSize);
     readPlane(pixelFile, x, y, w, h, buf);
 


sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-April/005934.html for the corresponding ome-users thread. Thanks for the PR  @cniedwor.

rleigh-dundee:
The change itself looks fine.

For the general case though, could we review the use of `int` in our interfaces?  Should most of them be `long` instead for safety?  In this specific example, `getPlaneSize()` is returning `int`.  This will overflow for images larger than `sqrt(2^(32-1)-1) = 46340.95` which for big images is not that big at all.  Could all of our dimension sizes be upgraded to `long`, both in the reader and writer APIs and in helpers such as this?

sbesson:
@rleigh-dundee: completely agreed on the principle. As proposed by @joshmoore  in https://trello.com/c/CMIqklLG/124-findbugs-for-int-long, Maven Fingbugs should flag these as warnings. We might have to make decisions on which warnings we can and want to fix in a future 5.2.0 milestone.

bramalingam:
Not able to reproduce the exception on Mac (ImageJ : Java 1.6.0_65 or ImageJ: Java 1.8.0_60, 64 bit, Bio-Formats 5.1.9): The image opens without any issues. And the image opens without any issues with the changes (on Bio-Formats Develop) as well, so if the code changes are acceptable, I will leave it upto @sbesson and @rleigh-dundee to decide on the course of action here.

sbesson:
@bramalingam: did you try opening the last plane of the image with and without this PR?

bramalingam:
Apologies, my testing scenario had not right. I was using "Use virtual stack" and was always looking at the first few planes. Never tried to reach the last plane. With the corrected scenario, I was able to reproduce the issue. And the PR is able fix the same. Looks good to merge.

PR #2349, opened 2016-04-15
Fix csv pattern blocks

simleo:
This PR fixes the detection of comma-separated pattern blocks.

The first commit addresses a minor problem where the `fixed` and `numeric` properties where not set for csv blocks, although they are general enough to make sense in all cases.

The second commit fixes a more serious problem where block strings where checked for range structure (`<START-STOP:STEP>`) first and then for csv structure. This made pattern blocks with dashes in their csv elements unusable. For instance, without this PR, FilePatternBlock would try to interpret `<Cy3-B,DAPI>` as a range from `Cy3` to `B,DAPI`, leading to a `NumberFormatException`.

Note that a side effect of this is that now single element blocks, e.g., `<foo>`, are considered comma-separated lists with one element, rather than ranges with step 1, so the `begin, end, step` properties are left to `null`. I'd be very surprised if this broke anything, given that a single-element pattern is a corner case that has no practical purpose.

### DIFF
--- a/components/formats-bsd/src/loci/formats/FilePatternBlock.java
+++ b/components/formats-bsd/src/loci/formats/FilePatternBlock.java
@@ -194,35 +194,22 @@ public class FilePatternBlock {
     String trimmed = block.substring(
         BLOCK_START.length(), block.length() - BLOCK_END.length()
     );
-    int dash = trimmed.indexOf("-");
-    String b, e, s;
-    if (dash < 0) {
-      // check if this is an enumerated list
-      int comma = trimmed.indexOf(",");
-      if (comma > 0) {
-        elements = trimmed.split(",");
-        setNumeric();
-        setFixed();
-        return;
-      }
-      else {
-        // no range and not a list; assume entire block is a single value
-        b = e = trimmed;
-        s = "1";
-      }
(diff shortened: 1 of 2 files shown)


joshmoore:
@simleo : no objections to having this open first on `develop` for discussion with @melissalinkert / @sbesson / et al but for deployment it will need to be against `dev_5_1` or optimally `metadata`.

simleo:
@joshmoore the reason I've opened this onto `develop` first is that `metadata` is still missing stuff from https://trello.com/c/AfFMb376/76-filestitcher-stack-issues, most notably the unit test for `FilePatternBlock`. If you want, I can start rebasing PRs stemming from the above card onto `metadata`, so that we'll be ready to backport this one as well once it's been reviewed.

sbesson:
Code changes look good and newly added unit tests cover the semantics described in the PR description. Waiting for the repository job to confirm this has no breaking impact on stitching in the data repository

sbesson:
As https://ci.openmicroscopy.org/view/Bio-Formats-DEV/job/BIOFORMATS-DEV-merge-full-repository/141/  is green, merging.

PR #2190, opened 2016-01-20
Add missing formats-bsd tests to the testng config file

simleo:
Adds `FilePattern{,Block}Test` so they are run with `ant test`

PR #2145, opened 2015-12-07
FilePatternBlock: fixes and unit test

simleo:
This PR adds a unit test for `FilePatternBlock` and fixes a few problems in its behavior, most notably the fact that alphanumeric range delimiters can't go beyond `"p"` due to a misinterpretation of the `radix` argument in `BigInteger`.

**To test:** check that the code compiles and *all* tests pass.

PR #2350, opened 2016-04-15
Fixes and unit test for FilePatternBlock (metadata)

simleo:
Same changes as #2145, but rebased onto metadata.

**Update:** also included (part of) #2190 so that the new test is run by `ant`


joshmoore:
Deployed to the idrstaging server as build 66. Merging to be tagged for 0.0.4.

PR #2351, opened 2016-04-18
Fix libBlitzBioFormats link

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/213/warnings3Result/ - I deleted the link initially thinking it hadn't been released since 2010, then re-read the page and realised it's now part of the XuvTools download. Not sure if we want to remove the link and leave the text though - the info on the page is rather out of date, linking to LOCI and broken doc links.

### DIFF
--- a/docs/sphinx/developers/jace/overview.txt
+++ b/docs/sphinx/developers/jace/overview.txt
@@ -29,4 +29,4 @@ external library.
 
 .. _jar2lib: http://loci.wisc.edu/software/jar2lib
 .. _WiscScan: http://loci.wisc.edu/software/wiscscan
-.. _Video Spot Tracker: 
+.. _Video Spot Tracker: http://cismm.cs.unc.edu/resources/software-manuals/video-spot-tracker-manual
--- a/docs/sphinx/developers/jace/overview.txt
+++ b/docs/sphinx/developers/jace/overview.txt
@@ -18,6 +18,9 @@ Other projects using the JACE C++ bindings include:
 
 -  WiscScan_ which uses the JACE C++ bindings to write
    :model_doc:`OME-TIFF <ome-tiff>` files.
+-  :doc:`/users/xuvtools/index` which uses an adapted version of the JACE C++
+   bindings called
+   `BlitzBioFormats <http://www.xuvtools.org/doku.php?id=devel:libblitzbioformats>`_.
 -  `Video Spot Tracker`_ which uses the JACE C++ bindings to add Bio-Formats
    support since version 8.10.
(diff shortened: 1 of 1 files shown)


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/ is green again but I'll hold off rebasing until we settle on a decision re: keeping the link or not.

melissalinkert:
NB: I downloaded 1.8.1-beta4 (the latest I could find, from 2011) and ran the following:

```
> java -cp .\blitzbioformats.jar loci.formats.tools.ImageInfo -version
Version: 4.3-DEV
SVN revision: 7443
Build date: 8 May 2011
```

Maybe worth noting that somewhere if the link is kept, so it's at least clear that we can't really provide support (and upgrading would be pretty scary)?

hflynn:
If that is the latest version, I'd be inclined to at least delete the link if not the entire mention of it to be honest.

hflynn:
Docs pages says "Other projects **using** the JACE C++ bindings include" if they aren't using it anymore, we should probably just delete the entire bullet point.

rleigh-dundee:
Removal would be fine with me.  On the other hand, it's purely ancillary information about JACE-using projects, so isn't harmful--we aren't recommending its use, and we could specifically add such a caution.

Related to that, JACE itself doesn't work with Java8, so the whole thing is of much less utility than previously.  See https://trello.com/c/ZizTu2RX/4-bfcpp-cppwrap .  So we might want to remove the whole page for 5.2.0 if it's going to be largely unusable...

hflynn:
I'd favour removal over adding caveats but will follow the majority opinion.

We should probably discuss the status of JACE for 5.2 on a card, the Bf developer docs are going to need some attention now all the C++ stuff has been migrated anyway.

sbesson:
No strong opinion on keeping vs removing the link in this PR, can see it either way. As noted by @rleigh-dundee in https://github.com/openmicroscopy/bioformats/pull/2351#issuecomment-211921277, we ceertainly need to clarify the JACE status in the context of BF 5.2.0.

hflynn:
Already added a card @sbesson - https://trello.com/c/9aehklS3/127-tidy-up-dev-docs-for-non-java

hflynn:
Okay, made the decision to remove the reference since no-one has spoken out against that strongly. Will rebase to dev_5_1 to fix the merge build.

hflynn:
--rebased-to #2362

PR #2352, opened 2016-04-18
FilePattern: allow pattern blocks in the dirname

simleo:
This PR removes a restriction in `FilePattern.getPrefix` that prevents the definition of patterns where one or more blocks appear in the dirname (rather than in the basename). This allows to match layouts where, for instance, data for different channels is stored in different directories:

```
/tmp/c<1,2>/foo.fake
```

Note that this is not hypotetical: we have exactly this problem with one of the IDR datasets.

As a side effect, this also removes a semantic discrepancy between `FilePattern.getPrefix()` and `FilePattern.getPrefix(int i)`, so that now the former behaves like the latter with an argument of `0`.

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/FilePatternTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/FilePatternTest.java
@@ -322,7 +322,7 @@ public class FilePatternTest {
     assertEquals(fp.getCount(), new int[] {2});
     assertEquals(fp.getElements(), new String[][] {{"0", "1"}});
     assertEqualsNoOrder(fp.getFiles(), new String[] {
-      "z0/foo.tif", "z1/foo.tif",
+      "z0" + suffix, "z1" + suffix,
     });
     assertEquals(fp.getFirst(), new BigInteger[] {BigInteger.ZERO});
     assertEquals(fp.getLast(), new BigInteger[] {BigInteger.ONE});
--- a/components/formats-bsd/test/loci/formats/utests/FilePatternTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/FilePatternTest.java
@@ -299,4 +299,34 @@ public class FilePatternTest {
     assertNotNull(fp.getErrorMessage());
   }
 
+  @Test
+  public void testVarDir() {
+    String[] prefixes = {"z"};
+    String[] blocks = {"<0-1>"};
+    String suffix = SEPARATOR + "foo.tif";
+    String pattern = mkPattern(prefixes, blocks, suffix);
+    FilePattern fp = new FilePattern(pattern);
+    assertTrue(fp.isValid());
+    assertFalse(fp.isRegex());
+    assertEquals(fp.getPattern(), pattern);
+    assertEquals(fp.getPrefixes(), prefixes);
+    for (int i = 0; i < prefixes.length; i++) {
+      assertEquals(fp.getPrefix(i), prefixes[i]);
+    }
+    assertEquals(fp.getPrefix(), prefixes[0]);
+    assertEquals(fp.getBlocks(), blocks);
+    for (int i = 0; i < blocks.length; i++) {
+      assertEquals(fp.getBlock(i), blocks[i]);
+    }
+    assertEquals(fp.getSuffix(), suffix);
+    assertEquals(fp.getCount(), new int[] {2});
+    assertEquals(fp.getElements(), new String[][] {{"0", "1"}});
+    assertEqualsNoOrder(fp.getFiles(), new String[] {
+      "z0/foo.tif", "z1/foo.tif",
+    });
+    assertEquals(fp.getFirst(), new BigInteger[] {BigInteger.ZERO});
+    assertEquals(fp.getLast(), new BigInteger[] {BigInteger.ONE});
+    assertEquals(fp.getStep(), new BigInteger[] {BigInteger.ONE});
+  }
+
 }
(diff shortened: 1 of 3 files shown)


components/formats-bsd/test/loci/formats/utests/FilePatternTest.java
```diff
@@ -299,4 +299,34 @@ public void testBadPatterns(String pattern) {
...
+    assertEquals(fp.getPrefix(), prefixes[0]);
+    assertEquals(fp.getBlocks(), blocks);
+    for (int i = 0; i < blocks.length; i++) {
+      assertEquals(fp.getBlock(i), blocks[i]);
+    }
+    assertEquals(fp.getSuffix(), suffix);
+    assertEquals(fp.getCount(), new int[] {2});
+    assertEquals(fp.getElements(), new String[][] {{"0", "1"}});
+    assertEqualsNoOrder(fp.getFiles(), new String[] {
+      "z0/foo.tif", "z1/foo.tif",
```
sbesson: Note this assert is failing on Windows. From the https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-build-win/188/ workspace,

```
java.lang.AssertionError: Arrays not equal: [z0/foo.tif, z1/foo.tif] and [z0\foo.tif, z1\foo.tif]
```

simleo: Forgot about windows again :( Will fix it ASAP

sbesson:
Note this PR was responsible for the early failure of https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/140/ as it breaks the semantics of some MIAS datasets

```
[sbesson@ome-c6100-1 BIOFORMATS-DEV-merge-build]$ ./tools/showinf -nopix /ome/data_repo/curated/mias/frans/NEO-Exp13072008-long-protokol/001-365700055641/Well0001/mode1_z000_t000_im1_1.tif 
Checking file format [MIAS]
Initializing reader
MIASReader initializing /ome/data_repo/curated/mias/frans/NEO-Exp13072008-long-protokol/001-365700055641/Well0001/mode1_z000_t000_im1_1.tif
Building list of TIFF files
Exception in thread "main" loci.formats.FormatException: Unsupported block 'repo/curated/mias/frans/neo-exp13072008-long-protokol/001-365700055641/well0001/mode
	at loci.formats.in.MIASReader.initFile(MIASReader.java:660)
	at loci.formats.FormatReader.setId(FormatReader.java:1431)
	at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:651)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1034)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1123)
```

With this PR excluded, https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/215/ is back to normal. Excluding for now.

simleo:
It seems that what we are considering as a *limitation* of `FilePattern` in IDR is actually relied upon as a *feature* by `MIASReader`. We should be able to fix this by changing the `MIASReader` code so that it's aware of the new way things work. I can't look into this until I fix (or get stuck trying to fix) the current IDR issue though.

sbesson:
Discussed with @simleo, removing the `exclude` label. This PR can be labelled as `--breaking` instead and tested against the `mias` data repository subset using the [automated test job](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/) /cc @dgault

simleo:
--breaking

simleo:
8d7c2610bbea86842b422c9335f5d9b2322bef7e should hopefully fix `MIASReader` (works on my laptop).

Both with the old (without this PR) and the new `FilePattern`, the following pattern is discovered for the first well:

```
/ome/data_repo/curated/mias/frans/NEO-Exp13072008-long-protokol/001-365700055641/Well0001/mode<1-2>_z000_t000_im<1-2>_<1-2>.tif
```

However, with the new `FilePattern`, the first prefix is `"/ome/data_repo/curated/mias/frans/NEO-Exp13072008-long-protokol/001-365700055641/Well0001/mode"`, not `"mode"`. I could have left the `FilePattern` call unchanged and cut the directory part from the first prefix after pattern detection, but since here `MIASReader` is not interested in the directory part at all (when looking at pattern blocks), I just had `FilePattern` perform pattern discovery on the basenames. Since now it has to analyze shorter strings, it should be slightly faster as a side effect.

sbesson:
Note that https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/79/ passed against the `mias` fileset with this PR. Removing the breaking label for tomorrow.

dgault:
Tests are still green and the new file patterns appear to be recognised as expected. The code changes make sense to me and the new unit test for the variable directories also looks good. The only small thing I noticed was in the MIASReader it might be to have a null pointer check on firstTiff.getParentFile().listFiles(). Otherwise these changes all look good to me.

simleo:
@dgault thanks for the review. Did you manage to set up a scenario where you actually get the `NullPointerException` on `firstTiff.getParentFile().listFiles()`? According to my tests, the only problematic situation is that in which a well directory has empty subdirectories, but in that case you get an `ArrayIndexOutOfBoundsException` on `new Location(tiffFiles[0])` three lines above. This has to be fixed, but it's not related to this PR. I will open a card.

Note that if a well directory is *totally* empty (no files and no subdirectories) it gets skipped on line 541, so that does not lead to the `ArrayIndexOutOfBoundsException`.

dgault:
I was basing it of a potential of return null in list files for a empty directory but wasn't able to reproduce an actual scenario either. As you say this should instead result in the ArrayIndexOutOfBoundsException getting thrown earlier. I think its fair to say that its not directly related to this PR and it should be fine to merge the PR as it is.

PR #2356, opened 2016-04-18
Fix reader delegation when the legacy reader is enabled but not working

melissalinkert:
This fixes the edge case where ```setUseLegacy(true)``` was called, but the
legacy reader throws an exception during setId triggering the native
reader instead.

Prior to this commit, the result of calling openBytes after setId would
have been a slightly cryptic exception indicating invalid core metadata.
Now, calling openBytes after setId should pass off to the native reader
without error.

Testing is easiest in ImageJ, preferably on a non-Windows machine.  First open ImageJ and select ```Plugins > Bio-Formats > Bio-Formats Plugins Configuration```, then select ```Nikon ND2``` in the ```Formats``` tab and make sure the ```Use Nikon's ND2 library...``` box is ticked (it shouldn't be by default).  Without this change, opening either of the .nd2 files in test_images_good/nd2 should result in an IllegalArgumentException.  With this change, the same test should result in the file being opened without error.

After testing this PR, reverting the ImageJ configuration change is recommended - I only discovered this issue because I'd forgotten to untick the box after a previous round of testing a few months ago.

### DIFF
--- a/components/formats-api/src/loci/formats/DelegateReader.java
+++ b/components/formats-api/src/loci/formats/DelegateReader.java
@@ -192,7 +192,7 @@ public abstract class DelegateReader extends FormatReader {
   /* @see IFormatReader#get8BitLookupTable() */
   @Override
   public byte[][] get8BitLookupTable() throws FormatException, IOException {
-    if (useLegacy || (legacyReaderInitialized && !nativeReaderInitialized)) {
+    if (callLegacyReader()) {
       return legacyReader.get8BitLookupTable();
     }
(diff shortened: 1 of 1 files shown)


mtbc:
Without this change I get a `PositiveInteger` exception on both the files with that option checked. (Without that option all works fine.)

With this change both the files work regardless of if the option is checked. Also, the code change looks sensible.

Good to merge.

mtbc:
(Out of curiosity is there an easy way to completely reset BF plugin config to default?)

PR #2359, opened 2016-04-19
cpp: transform test: Handle MapAnnotation downgrade

rleigh-dundee:
With the patched xalan source from https://github.com/ome/ome-cmake-superbuild/pull/56 it's now safe to test downgrade failure for map annotations.  This patch adds the ability to know if a test will throw and exception or not, and check for this accordingly.

**Do not merge before https://github.com/ome/ome-cmake-superbuild/pull/56 **

--------

Testing: Check jobs are green.  The tests will segfault and fail if it's not handled, or else fail if the test passes rather than failing as required.

### DIFF
--- a/cpp/test/ome-xml/transform.cpp
+++ b/cpp/test/ome-xml/transform.cpp
@@ -58,6 +58,7 @@ using boost::filesystem::directory_iterator;
 struct TransformTestParameters
 {
   path file;
+  bool throws;
 };
 
 template<class charT, class traits>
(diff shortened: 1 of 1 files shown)


sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/OME-FILES-CPP-DEV-merge/173/
https://ci.openmicroscopy.org/view/Failing/job/OME-FILES-CPP-DEV-merge-superbuild/204/
https://ci.openmicroscopy.org/view/Failing/job/OME-FILES-CPP-DEV-merge-win-superbuild/227/

PR #2360, opened 2016-04-19
TestNG - OutOfMemory handling

dgault:
This is building upon the commits in https://github.com/openmicroscopy/bioformats/pull/2282
This has extended the Out Of Memory exception handling to other tests in the FormatReaderTest.

There are the possibilities of some memory leaks with the tests but those are still being investigated.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -314,9 +314,12 @@ public class FormatReaderTest {
         try {
           b = reader.openThumbImage(0);
         }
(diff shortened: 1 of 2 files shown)


sbesson:
https://ci.openmicroscopy.org/view/Bio-Formats-DEV/job/BIOFORMATS-DEV-merge-full-repository/141/ turned green with this PR included. :+1: Let's see if we can maintain the job stability over a week. 

The only addition I can think of is whether we should have some console logging when OOM are catched and handled. This would allow us to parse our full-repository logs and check the number of skipped tests due to OOM remains under control (1-5).

PR #2362, opened 2016-04-20
Fix libBlitzBioFormats link (rebased onto dev_5_1)

hflynn:
This is the same as gh-2351 but rebased onto dev_5_1.

----

See https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/213/warnings3Result/ - I deleted the link initially thinking it hadn't been released since 2010, then re-read the page and realised it's now part of the XuvTools download. Not sure if we want to remove the link and leave the text though - the info on the page is rather out of date, linking to LOCI and broken doc links.

### DIFF
--- a/docs/sphinx/developers/jace/overview.txt
+++ b/docs/sphinx/developers/jace/overview.txt
@@ -29,4 +29,4 @@ external library.
 
 .. _jar2lib: http://loci.wisc.edu/software/jar2lib
 .. _WiscScan: http://loci.wisc.edu/software/wiscscan
-.. _Video Spot Tracker: 
+.. _Video Spot Tracker: http://cismm.cs.unc.edu/resources/software-manuals/video-spot-tracker-manual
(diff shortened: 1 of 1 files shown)


hflynn:
--rebased-from #2351

docs/sphinx/developers/jace/overview.txt
```diff
@@ -32,4 +29,4 @@ external library.
 
 .. _jar2lib: http://loci.wisc.edu/software/jar2lib
 .. _WiscScan: http://loci.wisc.edu/software/wiscscan
-.. _Video Spot Tracker: http://cismm.cs.unc.edu/resources/software-manuals/video-spot-tracker-manual
```
sbesson: Was this link removed on purpose?

hflynn: No! Sorry, will fix that

hflynn:
Fixed, thanks for spotting that!

PR #2364, opened 2016-04-20
LiFlimReader: fix ExposureTime check and use native units

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-April/005940.html

This change should fix the issue reported in the ome-users thread while preserving compatiblity with existing `li-flim` filesets. Additionally, the exposure time should now be stored using the native unit in the metadata.

To test this PR, check the automated tests pass and review the sample files under `curated/li-flim`.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
@@ -55,6 +55,7 @@ import loci.formats.UnsupportedCompressionException;
 import loci.formats.meta.MetadataStore;
 
 import ome.units.quantity.Time;
+import ome.units.unit.Unit;
 import ome.units.UNITS;
 
 /**
@@ -135,6 +136,7 @@ public class LiFlimReader extends FormatReader {
   private Map<Integer, ROI> rois;
   private Map<Integer, String> stampValues;
   private Double exposureTime;
+  private Unit<Time> exposureTimeUnit = UNITS.S;
 
   /** True if gzip compression was used to deflate the pixels. */
   private boolean gzip;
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
The changes look sensible, https://ci.openmicroscopy.org/view/Bio-Formats-DEV/job/BIOFORMATS-DEV-merge-full-repository/ has been green for several days with this PR open.

One discrepancy I am seeing is with `li-flim/samples/lifetime-extended.fli`:

```
filename = li-flim/samples/lifetime-extended.fli
Series count = 1
Series #0 :
        Image count = 15
        RGB = true (6) 
        Interleaved = true
        Indexed = false (true color)
        Width = 32
        Height = 24
        SizeZ = 3 (1 null x 2 Frequency)
        ************ SizeZ dimension mismatch ************
        SizeT = 5 (2 null x 2 Phase)
        ************ SizeT dimension mismatch ************
        SizeC = 6 (effectively 1)
```

This indicates that the reader is screwing up the calculation for e.g. the total plane count.

And then:
```
Exception in thread "main" java.lang.IllegalArgumentException: Cannot construct image with 6 channels
```

It's not 6 samples, it's 6 separate channels, so I'm not sure why this is too taxing for the viewer.

Other than that oddity, which is likely not specific to this PR, it looks good to merge.

sbesson:
Agreed on the discrepancy although I suspect this is independent of this PR. Could you confirm the 5.1.9 command-line tools give the same output ?

rleigh-dundee:
@sbesson Yes, it's the same with 5.1.9, so shouldn't prevent this PR being merged.  It looks fine to merge to me.

PR #2368, opened 2016-04-25
AxisGuesser: public axis prefixes

simleo:
This PR changes `AxisGuesser`'s static fields that define known axis prefixes (`AxisGuesser.{Z,C,T,S}`) to `public`. Users need to know what pattern block prefixes lead to automatic axis assignments, and this is indeed documented in the [sphinx docs](https://www.openmicroscopy.org/site/support/bio-formats5.1/formats/pattern-file.html).

Another problem with those fields is that their names are too short, making the code harder to read and debug. Since they're going public, this was the last chance to make them longer, so I've changed them to `AxisGuesser.{Z,C,T,S}_PREFIXES`

### DIFF
--- a/components/formats-bsd/src/loci/formats/AxisGuesser.java
+++ b/components/formats-bsd/src/loci/formats/AxisGuesser.java
@@ -40,6 +40,8 @@ import loci.common.Location;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableSet;
+
 /**
  * AxisGuesser guesses which blocks in a file pattern correspond to which
  * dimensional axes (Z, T, C or S), potentially recommending an adjustment in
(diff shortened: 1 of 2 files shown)


components/formats-bsd/src/loci/formats/AxisGuesser.java
```diff
@@ -71,18 +71,18 @@
   public static final int S_AXIS = 4;
 
   /** Prefixes indicating space dimension. */
-  protected static final String[] Z = {
+  public static final String[] Z_PREFIXES = {
```
joshmoore: As a general heads up, `public static final` arrays make findbugs unhappy.

simleo:
@joshmoore findbugs has every right to complain. I've changed those fields to immutable sets (not lists, since order does not actually matter). This also allowed to make part of the constructor's code cleaner and more efficient.

sbesson:
Changes make sense https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/loci/formats/AxisGuesser.html documents the new public fields and unit test are all passing. Merging.

components/formats-bsd/src/loci/formats/AxisGuesser.java
```diff
@@ -71,18 +73,24 @@
   public static final int S_AXIS = 4;
 
   /** Prefixes indicating space dimension. */
-  public static final String[] Z_PREFIXES = {
+  public static final ImmutableSet<String> Z_PREFIXES = ImmutableSet.of(
```
joshmoore: Not to harp, but we should make sure that having a google object in the public API is something we want since we won't be able to change the implementation without a breaking change.

sbesson: Fair enough. I had looked for other usages of guava across the code. Despite being consumed in several places, there is no precedent of using google objects as public API.

simleo:
I'm very hesitant to use external libraries unless it's necessary, and not just for the public API. In this case, however, I couldn't find a proper immutable collection in the standard Java API. There's [Collections.unmodifiableSet](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableSet(java.util.Set)), but it's just an unmodifiable view of an underlying collection that's still mutable. Would that be enough? Or maybe there is another solution I'm not aware of. Any suggestions?

components/formats-bsd/src/loci/formats/AxisGuesser.java
simleo: I'm very hesitant to use external libraries unless it's necessary, and not just for the public API. In this case, however, I couldn't find a proper immutable collection in the standard Java API. There's [Collections.unmodifiableSet](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableSet(java.util.Set)), but it's just an unmodifiable view of an underlying collection that's still mutable. Would that be enough? Or maybe there is another solution I'm not aware of. Any suggestions?

joshmoore: If there's no reference to the mutable collection, then an unmodifiable view is fine.

simleo: @joshmoore fix in #2382

joshmoore: Ta.

PR #2371, opened 2016-04-25
TiffSaver: fix recorded offsets to non-interleaved RGB tiles (5.1)

rleigh-dundee:
--rebased-from #2168

See the above PR for testing instructions.
Also see https://www.openmicroscopy.org/community/viewtopic.php?t=8042&p=17018#p17018 for a sample SVS file (http://openslide.cs.cmu.edu/download/openslide-testdata/Aperio/JP2K-33003-1.svs) which fails to convert with bfconvert without this PR.


sbesson:
Retested the workflow described in https://www.openmicroscopy.org/community/viewtopic.php?t=8042&p=17018. SVS conversion into TIFF worked as expected.

sbesson:
Retested following the instructions of https://github.com/openmicroscopy/bioformats/pull/2168 against 2k x 2k TIFF and PNG images with and without tiling options. No more mismatch post conversion. Merging.

PR #2372, opened 2016-04-25
Logging improvements

sbesson:
See http://trac.openmicroscopy.org/ome/ticket/13151

This PR reviews the `loci.common.DebugTools` class and its logging framework implementations `loci.common.Log4jTools` and `loci.common.LogbackTools` to:
- respect logging frameworks (log4j/logback) initialized via a binding-specific configuration file
- prevent `DebugTools.enableLogging(String)` from overriding initialized logger levels

### Major  API changes:
- addition of a new `DebugTools.setRootLevel(String)` to set the level of the root logger with its `Log4jTools` and `LogbackTools` implementations
- addition of `DebugTools.enableLogging()` the core function for initialization the logging system which is conditional to the return code of  `DebugTools.isEnabled()` and delegates the initialization to the corresponding logging framework classes
- modification of `DebugTools.enableLogging(String)` to set the root logger level via `DebugTools.setRootLevel(String)` on sucessful initialization of the logging framework via `DebugTools.enableLogging()`
- improvement of the `LogbackTools.isEnabled()` implementation to return true if
  1. a binding-specific configuration file has been used to initialize the logging framework or
  2. the logging framework has been initialized via `DebugTools` (using an internal registration mechanism via the logger context properties)

### Plugins/tools
- ImageJ plugin is unmodified and should call `DebugTools.enableLogging(INFO)`. Using a binding-specific configuration file (`logback.xml`) should take precedence over this logging call
- Command-line tools: should be initialized via `tools/logback.xml`. All options-specific logging calls are converted to use `DebugTools.setRootLevel(String)`
- MATLAB toolbox: the `DebugTools.isEnabled` condition is now moved to the downstream method and removed as superfluous

### Test coverage

Tests are available for the 2 main logging frameworks supported by Bio-Formats
- logback: a new `DebugToolsTest` class has been added to the TestNG suite and should be included in Maven/Ant test targets
- log4j: the principal consumer being the MATLAB toolbox, the `DebugTools` happens via the `TestBfInitLogging` unit tests and should be run in the Jenkins MATLAB jobs

### DIFF
--- a/components/formats-common/src/loci/common/DebugTools.java
+++ b/components/formats-common/src/loci/common/DebugTools.java
@@ -116,8 +116,7 @@ public final class DebugTools {
    * Attempts to enable SLF4J logging via logback or log4j without an
    * external configuration file.
    * This will first check whether logging has been enabled either via a
-   * configuration file or a previous call. If not enabled, calls
-   * {@link #setRootLevel(String)} at INFO level.
+   * configuration file or a previous call.
    *
    * @return {@code true} if logging was successfully enabled by this method
    */
--- a/components/formats-common/src/loci/common/DebugTools.java
+++ b/components/formats-common/src/loci/common/DebugTools.java
@@ -47,7 +47,7 @@ import java.util.Arrays;
  */
 public final class DebugTools {
 
-  final static String[][] TOOLCLASSES = new String[][] {
+  final static String[][] TOOL_CLASSES = new String[][] {
     new String[] {"loci.common.", "LogbackTools"},
     new String[] {"loci.common.", "Log4jTools"}
   };
@@ -74,7 +74,7 @@ public final class DebugTools {
    * @return {@code true} if logging has been successfully enabled
    */
   public static synchronized boolean isEnabled() {
-    for (String[] toolClass : TOOLCLASSES) {
+    for (String[] toolClass : TOOL_CLASSES) {
       try {
         Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
         Method m = k.getMethod("isEnabled");
@@ -96,7 +96,7 @@ public final class DebugTools {
    * @param level A string indicating the desired level
    */
   public static synchronized void setRootLevel(String level) {
-    for (String[] toolClass : TOOLCLASSES) {
+    for (String[] toolClass : TOOL_CLASSES) {
       try {
         Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
         Method m = k.getMethod("setRootLevel", String.class);
@@ -120,7 +120,7 @@ public final class DebugTools {
    */
   public static synchronized boolean enableLogging() {
     if (isEnabled()) return false;
-    for (String[] toolClass : TOOLCLASSES) {
+    for (String[] toolClass : TOOL_CLASSES) {
       try {
         Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
         Method m = k.getMethod("enableLogging");
--- a/components/formats-common/src/loci/common/DebugTools.java
+++ b/components/formats-common/src/loci/common/DebugTools.java
@@ -122,7 +122,19 @@ public final class DebugTools {
    * @return {@code true} if logging was successfully enabled by this method
    */
   public static synchronized boolean enableLogging() {
-    return enableLogging("INFO");
+    if (isEnabled()) return false;
+    for (String[] toolClass : TOOLCLASSES) {
+      try {
+        Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
+        Method m = k.getMethod("enableLogging", String.class);
+        m.invoke(null);
+        return true;
+      }
+      catch (Throwable t) {
+        // no-op. Ignore error and try the next class.
+      }
+    }
+    return false;
   }
   
   /**
@@ -137,19 +149,9 @@ public final class DebugTools {
    * @return {@code true} if logging was successfully enabled by this method
    */
   public static synchronized boolean enableLogging(String level) {
-    if (isEnabled()) return false;
-    for (String[] toolClass : TOOLCLASSES) {
-      try {
-        Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
-        Method m = k.getMethod("enableLogging", String.class);
-        m.invoke(null, level);
-        return true;
-      }
-      catch (Throwable t) {
-        // no-op. Ignore error and try the next class.
-      }
-    }
-    return false;
+    boolean status = enableLogging();
+    if (status) setRootLevel(level);
+    return status;
   }
 
   /**
--- a/components/formats-common/src/loci/common/DebugTools.java
+++ b/components/formats-common/src/loci/common/DebugTools.java
@@ -126,7 +126,7 @@ public final class DebugTools {
     for (String[] toolClass : TOOLCLASSES) {
       try {
         Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
-        Method m = k.getMethod("enableLogging", String.class);
+        Method m = k.getMethod("enableLogging");
         m.invoke(null);
         return true;
       }
--- a/components/formats-common/src/loci/common/DebugTools.java
+++ b/components/formats-common/src/loci/common/DebugTools.java
@@ -80,7 +80,7 @@ public final class DebugTools {
         Method m = k.getMethod("isEnabled");
         return (Boolean) m.invoke(null);
       }
-      catch (Throwable t) {
+      catch (ReflectiveOperationException t) {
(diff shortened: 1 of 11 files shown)


melissalinkert:
Does the change in intention of ```enableLogging``` need to propagate to ```loci.common.Log4jTools``` and ```loci.common.LogbackTools```?

When testing in ImageJ - bioformats_package.jar, no log config file - I see ```DEBUG``` output on the console when I'd expect to see ```INFO```, presumably because ```isEnabled()``` always returns true for logback.  Does it make sense to fix that here as well?

sbesson:
Excluding for now. As suggested in https://github.com/openmicroscopy/bioformats/pull/2372#issuecomment-214837806, logback always [configures itself automatically](http://logback.qos.ch/manual/configuration.html) making the current implementation of `DebugTools.isEnabled()` effectively a no-op.

sbesson:
Commits pushed to review the logging system initialization logic, add unit tests for the default logback framework.  The description of the PR description has been updated to describe the nature of the new breaking changes. One method unmodified so far is `DebugTools.enableIJLogging` which still overrides the root logger level e.g. causing all the logs at the `INFO` level in the bio-formats-plugins tests. Should this be modified to reuse the same logic as `DebugTools.enableLogging()` and be conditional to the status of `DebugTools.isEnabled()` or do we want to preserve the current behavior?

melissalinkert:
API changes seem fine to me.  The ImageJ test does now work - I see ```INFO```-level output with no config file, and the level changes appropriately when a config file is on the classpath.  ```enableIJLogging``` is only called in one place (loci.plugins.in.ImportProcess:678); I think the logic needs to stay the same, since it's intentionally adding a new appender regardless of any other underlying configuration.  If there are any strong feelings otherwise though, happy to test a new commit.

I was initially a little skeptical of the s/```enableLogging```/```setRootLevel``` in the command line tools, but testing both with and without a config file produces the output I'd expect, so :+1: across the board.

sbesson:
@melissalinkert: you are correct, `enableIJLogging` should not follow the same pattern as `enableLogging`.
 Coming back to the original [community thread](https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7993), I guess my main worry is rather the unconditional setting of the root logger level by the method i.e. anyone using `BF.openImagePlus()` is forced into either `DEBUG` or `INFO` level independently of the logging initalization.
7b4df8c gives an improvement proposal by only setting the root logger level to debug if `IJ.debugMode` is true and preserving the current level otherwise. An immediate effect of this change is that the ImageJ plugin `ImporterTest` should now have all logs at `INFO` level silenced from the console output as per the `logback.xml` initialization. Let me know if you think this would be too intrusive/breaking for existing workflows.

components/formats-common/src/loci/common/DebugTools.java
```diff
@@ -47,8 +47,12 @@
  */
 public final class DebugTools {
 
-  // -- Constructor --
+  final static String[][] TOOLCLASSES = new String[][] {
```
simleo: I'd call this `TOOL_CLASSES` according to the convention on constant names

components/formats-common/src/loci/common/DebugTools.java
```diff
@@ -65,17 +69,12 @@ public static String getStackTrace(Throwable t) {
   }
 
   /**
-   * Checks whether SLF4J logging has been enabled via logback or log4j
+   * Checks whether the SLF4J logging has been enabled via logback of log4j.
```
simleo: I'd revert this change

components/formats-common/src/loci/common/DebugTools.java
```diff
@@ -89,24 +88,43 @@ public static synchronized boolean isEnabled() {
...
+  public static synchronized void setRootLevel(String level) {
+    for (String[] toolClass : TOOLCLASSES) {
       try {
         Class<?> k = Class.forName(toolClass[0] + toolClass[1]);
-        Method m = k.getMethod("enableLogging", String.class);
+        Method m = k.getMethod("setRootLevel", String.class);
         m.invoke(null, level);
+        return;
+      }
+      catch (Throwable t) {
```
simleo: Catching `Throwable` is pretty dangerous, as it swallows up everything. It could hide any type of error. We should catch `ClassNotFoundException` / `NoSuchMethodException` and let other errors slip through undisturbed.

simleo:
All changes seem good to me, except for a few notes I've left inline. I've checked all three command line tools with and without this patch, and now `logback.xml` isn't ignored anymore. The `-debug` and `-trace` options in `showinf` override the `logback.xml` setting, which is also good.

The fact that setting the log level now works, however, exposes a problem! CLI tools use log messages for **all** communication, so if I set `root level="ERROR"` in logback.xml, `showinf` does not do **anything**. Even worse, if I specify an existing output file to `bfconvert`, the tool seems to hang indefinitely, because it's waiting for an answer to a question that the user hasn't even seen (`"Do you want to overwrite it? ([y]/n)"`).

sbesson:
Pushed 2 additional commits. While trying to address the `Throwable` comment, https://github.com/sbesson/bioformats/commit/f1a6f4b4b6e4c0f62ab2d13400d4c85084b230a2 caused more errors in the build so I propose to address this as a bug fix in a follow-up PR.

Thanks for the extensive testing of the command-line tools. As this PR is mostly concerned about the logging API and the default command-line tools are working, I propose to have the command-line tools classes reviewed separately to identify the calls that should be using `System.out` as opposed slf4j.

components/formats-common/src/loci/common/DebugTools.java
mtbc: `ReflectiveOperationException` can be a useful superclass for this kind of code.

jburel:
I will move ``"Bio-Formats"`` to a common static variable. since it is invoked in few locations.

jburel:
Matlab test added for the changes in this PR. https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-matlab/lastCompletedBuild/testReport/(root)/TestBfInitLogging/ passed

sbesson:
Last commits should address @jburel's https://github.com/openmicroscopy/bioformats/pull/2372#issuecomment-218397600 but also @simleo's https://github.com/openmicroscopy/bioformats/pull/2372#discussion_r62660546.

simleo:
+1 from me, `ReflectiveOperationException` seems exactly what we need here

components/formats-common/src/loci/common/LogbackTools.java
```diff
@@ -51,31 +51,52 @@
 
   // -- Constructor --
 
+  public static final String CALLER = "Bio-Formats";
```
joshmoore: As mentioned elsewhere, I'd tend to make a field like this private unless there's a reason we want `LogbackTools.CALLER` as part of our public API. As this stands, changing or removing the variable would require a deprecation cycle.

sbesson: Agreed - see https://github.com/openmicroscopy/bioformats/pull/2372/commits/2207de006f83e701f16137023930cd30fdd7444d

sbesson:
Having received thumbs up from @joshmoore and others on the last set of changes, merging this.

components/formats-common/src/loci/common/DebugTools.java
sbesson: Interestingly switching  from `Throwable` to `ReflectiveOperationException` broke the ImageJ plugin when using `loci_tools.jar` (as shown by the testing of 5.2.0-m4 /cc @pwalczysko @bramalingam) as it does not ctach the `NoClassDefFoundError` thrown in:

```
java.lang.NoClassDefFoundError: ch/qos/logback/core/Context
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at loci.common.DebugTools.enableLogging(DebugTools.java:132)
	at loci.common.DebugTools.enableLogging(DebugTools.java:159)
	at loci.plugins.LociImporter.run(LociImporter.java:70)
	at ij.IJ.runUserPlugIn(IJ.java:216)
	at ij.IJ.runPlugIn(IJ.java:180)
	at ij.Executer.runCommand(Executer.java:137)
	at ij.Executer.run(Executer.java:66)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.ClassNotFoundException: ch.qos.logback.core.Context
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
```

mtbc: Interesting, might be worth catching `LinkageError` more broadly commenting why it can happen.

PR #2373, opened 2016-04-26
Unit test for FilePatternBlock

simleo:
Rebase of #2161 onto `metadata`. It also adds the test to `testng.xml` (in `develop` this happened in #2190). It does *not* include the change to `FilePattern` from 4329e2fe021bf252f0d133ee001c4827e38e3048, because that's already been merged to `metadata`.

### DIFF
--- a/components/formats-bsd/src/loci/formats/FilePattern.java
+++ b/components/formats-bsd/src/loci/formats/FilePattern.java
@@ -703,51 +703,6 @@ public class FilePattern {
     return files.toArray(new String[files.size()]);
   }
 
-  // -- Main method --
-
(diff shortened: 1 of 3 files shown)


joshmoore:
cppwrap failed with:

```
2016-04-26 10:09:57,688 [main] WARN  loci.formats.Memoizer - config: minimumElapsed wrong type: null
2016-04-26 10:09:57,689 [main] WARN  loci.formats.Memoizer - config: minimumElapsed wrong type: null


The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.

```

@sbesson : is there something that should be backported?

PR #2379, opened 2016-04-27
Update cpp downloads link in sidebar

hflynn:
See https://trello.com/c/K0EpMdAx/128-bf-docs-sidebar-menu-still-links-to-bf-c-downloads

Staged at http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/

Not to be backported

### DIFF
--- a/docs/sphinx/themes/globalbftoc.html
+++ b/docs/sphinx/themes/globalbftoc.html
@@ -1,5 +1,5 @@
 <h3><a href="{{ pathto(master_doc) }}">{{ _('Bio-Formats') }}</a></h3>
 {{ toctree()}}
-<a href="http://downloads.openmicroscopy.org/latest/bio-formats5.1/">{{ _('Bio-Formats Java Downloads') }}</a></br>
+<a href="http://downloads.openmicroscopy.org/latest/bio-formats5.2/">{{ _('Bio-Formats Downloads') }}</a></br>
 <a href="http://downloads.openmicroscopy.org/latest/ome-files-cpp/">{{ _('OME Files C++ Downloads') }}</a></br>
 <a href="//www.openmicroscopy.org/site/about/licensing-attribution">{{ _('Licensing') }}</a>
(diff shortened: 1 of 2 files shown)


docs/sphinx/themes/globalbftoc.html
```diff
@@ -1,5 +1,5 @@
 <h3><a href="{{ pathto(master_doc) }}">{{ _('Bio-Formats') }}</a></h3>
 {{ toctree()}}
-<a href="http://downloads.openmicroscopy.org/latest/bio-formats5.1/">{{ _('Main Downloads') }}</a></br>
-<a href="http://downloads.openmicroscopy.org/latest/bio-formats-cpp5.1/">{{ _('C++ Downloads') }}</a></br>
+<a href="http://downloads.openmicroscopy.org/latest/bio-formats5.1/">{{ _('Bio-Formats Java Downloads') }}</a></br>
+<a href="http://downloads.openmicroscopy.org/latest/ome-files-cpp/">{{ _('C++ Downloads') }}</a></br>
```
rleigh-dundee: Should be "OME Files C++"

hflynn:
Can we make a decision re: backporting because I realise if we do want to put this on the 5.1.10 release, I'd need to update the conf too 'else the remaining C++ docs will be pointing to a different download page to the sidebar cc/ @sbesson

sbesson:
@hflynn: as all 0.1.x releases of OME Files C++ are using milestone releases of Bio-Formats 5.2.0 for the OME-XML C++ component, I would be inclined to *not* backport this change to Bio-Formats 5.1.x.

docs/sphinx/themes/globalbftoc.html
```diff
@@ -1,5 +1,5 @@
 <h3><a href="{{ pathto(master_doc) }}">{{ _('Bio-Formats') }}</a></h3>
 {{ toctree()}}
-<a href="http://downloads.openmicroscopy.org/latest/bio-formats5.1/">{{ _('Main Downloads') }}</a></br>
-<a href="http://downloads.openmicroscopy.org/latest/bio-formats-cpp5.1/">{{ _('C++ Downloads') }}</a></br>
+<a href="http://downloads.openmicroscopy.org/latest/bio-formats5.1/">{{ _('Bio-Formats Java Downloads') }}</a></br>
```
sbesson: With the OME Files C++ clarification, this could simply become `Bio-Formats Downloads`. If we agree not to backport, we should be able to point at the latest bio-formats5.2 downloads.

hflynn:
That sounds like a good argument for not backporting to me, I'll push the necessary changes to the config.

rleigh-dundee:
Looks fine to me, thanks.

Issue #2367, opened 2016-04-25
New warining message about SlideBook6Reader.dll with 5.1.9


sebi06:
Hi guys,

when i use the bioformats_package.jar from 5.1.9 from python I get those warnings. But when I switch back to version 5.1.8 those warnings do not pop up. I both cases my CZI file is read correctly, but I am just wondering...
When I open the file directly from within Fiji no warning appears no matter what version (5.1.8 or 5.1.9) no  warning appears. 

Sebi 

C:\Anaconda\python.exe C:/Users/M1SRH/Documents/Python_Projects/BioFormatsRead/test_get_image6d.py
09:36:35.054 [Thread-0] DEBUG loci.common.NIOByteBufferProvider - Using mapped byte buffer? false
Apr 25, 2016 9:36:35 AM org.scijava.nativelib.BaseJniExtractor extractJni
INFORMATION: Couldn't find resource META-INF/lib/windows_64/ SlideBook6Reader.dll
Apr 25, 2016 9:36:35 AM org.scijava.nativelib.NativeLibraryUtil loadNativeLibrary
WARNUNG: IOException creating DefaultJniExtractor
java.io.IOException: Couldn't find resource META-INF/lib/windows_64/ SlideBook6Reader.dll
	at org.scijava.nativelib.BaseJniExtractor.extractJni(BaseJniExtractor.java:144)
	at org.scijava.nativelib.NativeLibraryUtil.loadNativeLibrary(NativeLibraryUtil.java:265)
	at loci.formats.in.SlideBook6Reader.<clinit>(SlideBook6Reader.java:93)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Unknown Source)
	at loci.formats.ClassList.<init>(ClassList.java:127)

bramalingam:
Hi,

Thank you for reporting your issue.

Bio-Formats reads a file with each reader detect the format. In our recent 5.1.9 release of Bio-Formats, we removed the native binary files required for the Slidebook reader from our repository (https://github.com/openmicroscopy/bioformats/pull/2300) which led to the warnings for the “missing Slidebook dll’s” in your script.

I have created a ticket on this regard, https://trac.openmicroscopy.org/ome/ticket/13220, and we are currently working on a fix. We are looking at releasing a patched version at the earliest.

Hope that helps.

Best,
Balaji

sbesson:
Closing in favor of the ticket.

sbesson:
[...]

PR #2381, opened 2016-04-28
Slidebook DLL warnings

sbesson:
See https://trac.openmicroscopy.org/ome/ticket/13220 and https://github.com/openmicroscopy/bioformats/issues/2367

The [removal of the 3i Slidebook DLLs](https://github.com/openmicroscopy/bioformats/pull/2300) from the source code repository in Bio-Formats 5.1.9 caused warnings to be systematically printed for non-Fiji/ImageJ Windows users of Bio-Formats.In particular, when constructing an `ImageReader` and initializing the `SlideBook6Reader`, the [following line](https://github.com/scijava/native-lib-loader/blob/native-lib-loader-2.1.0/src/main/java/org/scijava/nativelib/NativeLibraryUtil.java#L275) in the native library loading library is now throwing an exception at the WARN level due to the DLL removal.

To mitigate this issue, an initial proposal was to use the logging to silence these warnings. The main caveat here is that `native-lib-loader` uses `java.util.logging`. While possible this approach would require the introduction of another logging dependency `jul-to-slf4j` in the middle of a stable release.

This PR proposes an alternative fix and modifies the `SlideBook6Reader` to perform the native library check at runtime rather than at creation time. The static `libraryFound` flag is initialized via a private method `isLibraryFound()` which should only be executed in `isThisType(String, boolean)` after the `FormatReader.isThisType(String, booolean)` call has returned true i.e. once it is established the id is a Slidebook file.

This PR should be primarily tested on Windows using both the ImageJ/Fiji plugin and at least another tool (command-line, MATLAB...). When opening a non-Slidebook file, no DLL warning should be thrown when calling the opening commands:

```
showinf.bat test.fake  % Command-line
bfopen('path\to\test.fake'); %MATLAB
```

When opening a Slidebook file without the [3i Slidebook DLLs](http://www.openmicroscopy.org/info/slidebook) included, the missing DLLs warnings should be printed:

```
showinf.bat \path\to\slidebook.sld % Command-line
bfopen('path\to\slidebook.sld'); % MATLAB
```

/cc @joshmoore @bramalingam @RichardMyers

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
@@ -79,9 +79,11 @@ public class SlideBook6Reader  extends FormatReader {
 
 	// -- Static initializers --
 
+	private static boolean initialized = false;
 	private static boolean libraryFound = false;
 
-	static {
+	private static boolean isLibraryFound() {
+		if (initialized) return libraryFound;
 		try {
 			// load JNI wrapper of SBReadFile.dll
 			NativeLibraryUtil.Architecture arch = NativeLibraryUtil.getArchitecture();
@@ -102,6 +104,8 @@ public class SlideBook6Reader  extends FormatReader {
 			LOGGER.warn("Insufficient permission to load native library", e);
 			libraryFound = false;
 		}
+		initialized = true;
+		return libraryFound;
 	}
 
 	// -- Constructor --
--- a/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/SlideBook6Reader.java
@@ -138,7 +138,7 @@ public class SlideBook6Reader  extends FormatReader {
 	/* @see loci.formats.IFormatReader#isThisType(String, boolean) */
 	public boolean isThisType(String file, boolean open) {
 		// Check the first few bytes to determine if the file can be read by this reader.
-		return libraryFound && super.isThisType(file, open);
+		return super.isThisType(file, open) && isLibraryFound();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


dgault:
Tested on Windows in Fiji and through the command line tools. Non Slidebook files open without any warnings as expected. Slidebook files produce the missing dll resource warning which is as desired. PR looks good to merge.

sbesson:
Thanks @dgault for the testing. Apologies for forgetting about this point it in the description but did you happen to retest the workflow described in https://github.com/openmicroscopy/bioformats/pull/2300#issuecomment-198507977 to confirm this causes no regression while selecting the SlideBook6Reader when the relevant DLLs are found? @joshmoore: anything else you would like to see happening here?

dgault:
Retested after applying the Fiji install of SlideBook6Reader. With the update applied the warning messages are no longer present when importing Slidebook files. Files load as expected without any issues.

joshmoore:
WFM.

sbesson:
--rebased-to #2392

ctrueden:
Thanks, this is a nice fix!

> The main caveat here is that native-lib-loader uses java.util.logging. 

Note that we (as in: the SciJava consortium team) control `native-lib-loader`, so we could certainly move the logging to SLF4J. Would that be helpful? If so, I'm happy to take care of it, since I'm sure it would be easy. Just let me know!

sbesson:
@ctrueden: thanks for the input. In the context of the patch release, this PR was proposed to mitigate the issue for the majority of the users without causing any dependency bump. We had certainly planned to open the discussion with Scijava regarding the `native-lib-loader` logging in the context of Bio-Formats 5.2.0 where breaking changes can happen. One possibility we had envisioned was to include the `jul-to-slf4j` bindings to control the logging. Arguably, having `native-lib-loader` logging system migrated to slf4j might simplify this further.

PR #2382, opened 2016-04-28
AxisGuesser: do not use Guava

simleo:
Addresses a problem found by @joshmoore in #2368 after it was merged, i.e., the fact that I've used an external library in the public API. With this change, each set of prefixes not a true immutable set, but rather an unmodifiable view of a mutable `HashSet`. However, there's no way to modify the `HashSet` because only the unmodifiable wrapper has a reference to it.

### DIFF
--- a/components/formats-bsd/src/loci/formats/AxisGuesser.java
+++ b/components/formats-bsd/src/loci/formats/AxisGuesser.java
@@ -34,13 +34,17 @@ package loci.formats;
 
 import java.io.IOException;
 import java.math.BigInteger;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.Arrays;
 
 import loci.common.Location;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.ImmutableSet;
+import static java.util.Collections.unmodifiableSet;
+
 
 /**
  * AxisGuesser guesses which blocks in a file pattern correspond to which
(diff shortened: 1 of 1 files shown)


mtbc:
Just wondering, is it not okay to declare the variable as `Set` but to assign an `ImmutableSet` to it?

joshmoore:
> Just wondering, is it not okay to declare the variable as Set but to assign an ImmutableSet to it?

@mtbc would know better than anyone :smile:

mtbc:
Just thinking this PR might well be changing more than was required, not that I object either way.

joshmoore:
> Main thing IIRC is not to try to hand ImmutableSets to Ice.

@mtbc : not that it applies here, but **what**?! That's a surprise.

@simleo : As long as the field itself is just `Set` either of the solutions is fine. Sorry for all the noise.

simleo:
I'll let you guys decide. I prefer this one because it does not use Guava at all, so if one day we have to get rid of Guava everywhere (maybe because it becomes unmaintained and full of security bugs) there will be one less pain in the neck.

mtbc:
No strong opinion so long as you don't mind my using Guava everywhere else. :smiley: Certainly not meaning to block this so much as ensuring there wasn't something going on that I didn't understand (goodness knows this world is otherwise full of such stuff!).

simleo:
@mtbc I certainly don't mind! Staying away from external libraries is my personal obsession, and I don't do it myself if using one (or more) means you can do something with much less code.

PR #2385, opened 2016-04-29
cmake: Remove VC11 faux variadic templates workaround

rleigh-dundee:
See https://github.com/ome/ome-cmake-superbuild/pull/64


joshmoore:
Merging with https://github.com/ome/ome-cmake-superbuild/pull/64#issuecomment-216232380

PR #2388, opened 2016-05-02
Misc AxisGuesser and FilePattern fixes

simleo:
Rebases #2352, #2187, #2368 and #2382 onto metadata.


joshmoore:
Merging since there's one more dependent PR.

PR #2389, opened 2016-05-02
ScreenReader enhancements

simleo:
Adds two new features to `ScreenReader`:

 * An `AxisTypes` setting can be added to each `Well` entry to override the automatic type guessing performed by `AxisGuesser`. The value should be a string of `Z|C|T` characters, one for each pattern block (e.g., `AxisTypes = CT`)
 * A `ChannelNames` setting can be added to each `Well` entry to override the channel names. The value should be a comma-separated list, e.g., `ChannelNames = DAPI,Cy3`

Also adds a new `overrideAxisTypes` method to `FileStitcher`, which allows to externally set the axis types before `setId` is called. In this case, `FileStitcher` uses the new `AxisGuesser(int[], ...)` (see 19ef8e25d6a8244575a78d031c8715331da6e337) that sidesteps axis types guessing (i.e., the `AxisGuesser` is only used for `getAdjustedOrder`).

Note that, without `overrideAxisTypes`, `ScreenReader` would have to call `FileStitcher.setAxisTypes` *after* the call to `setId`, so `AxisGuesser` would perform its guessing work for nothing. After that, `ScreenReader` would have to call `reader.setId` again to make the wrapping `DimensionSwapper` pick up the changes, with an additional performance hit.

### DIFF
--- a/components/formats-bsd/src/loci/formats/AxisGuesser.java
+++ b/components/formats-bsd/src/loci/formats/AxisGuesser.java
@@ -114,6 +114,24 @@ public class AxisGuesser {
   /** Whether the guesser is confident that all axis types are correct. */
   protected boolean certain;
 
+  // -- Helpers --
+
+  private boolean swapZT(int sizeZ, int sizeT, boolean haveZ, boolean haveT) {
+    boolean wrongZ = haveZ && !haveT && sizeZ > 1 && sizeT == 1;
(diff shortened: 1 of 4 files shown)


joshmoore:
Travis re-open

joshmoore:
Merging per https://trello.com/c/zvShIIM8/243-idr0016-screenreader-images-for-a-single-plate-in-separate-folders-for-each-channel -- first plates have been imported.

joshmoore:
--rebased-to #3012

PR #2390, opened 2016-05-02
Merge dev_5_1 into metadata

joshmoore:
This brings the current version to 5.1.10-SNAPSHOT and will require an OMERO PR to sync.


joshmoore:
Merging lest this grow without my noticing it.

PR #2394, opened 2016-05-03
Move to develop .travis.yml

joshmoore:
@sbesson suggests that using the smaller Java7+ matrix from develop might fix the metadata branches builds.


joshmoore:
Assuming this is handled by https://github.com/openmicroscopy/bioformats/pull/2417

PR #2391, opened 2016-05-03
Fix SVI Huygens link

hflynn:
BF docs builds were unstable this morning because this link was broken. Replacement link is their homepage as found via google search. Should make the docs green again.

PR #2395, opened 2016-05-03
Fix SVI Huygens link (rebased onto dev_5_1)

hflynn:
This is the same as gh-2391 but rebased onto dev_5_1.

----

BF docs builds were unstable this morning because this link was broken. Replacement link is their homepage as found via google search. Should make the docs green again.


hflynn:
--rebased-from #2391

PR #2396, opened 2016-05-03
Tests and logging

sbesson:
See https://trello.com/c/ISuLct1l/125-formats-bsd-unit-tests-take-too-long-to-run

This PR starts addressing test issues at the logging level with the following changes:
- the Ant test targets now use the `logback.xml` under `ant` (like the Maven test targets)
- the `FATAL` logging configuration in `ant/logback.xml` is removed.

The impact of this PR should be the following:
- Ant-based tests should have reduced logging (https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/ and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build-win) especially the Formats BSD tests should have no more logs due to the JPEG2000 tests
- Formats BSD Maven-based tests should now hide the stack trace due to the ClassList error. As a results the log  of the Maven components in the [Travis matrix build](http://travis-ci.org/openmicroscopy/bioformats) should not overflow the 10K lines.

### DIFF
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -65,7 +65,7 @@ Type "ant -p" for a list of targets.
       <then>
         <junit haltonfailure="true" fork="true">
           <classpath>
-            <pathelement location="${root.dir}/tools/"/><!-- logback.xml -->
+            <pathelement location="${root.dir}/ant/"/><!-- logback.xml -->
             <pathelement location="${artifact.dir}/junit-4.8.2.jar"/>
             <pathelement location="${test-classes.dir}"/>
             <pathelement location="${classes.dir}"/>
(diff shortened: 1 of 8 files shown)


simleo:
Tested locally with `ant test` and `mvn test`. Line counts for the test output are as follows:

          | Without the PR  | With the PR |
| ------- | --------------- | ----------- |
| **ant** | 203037          | 4286        |
| **mvn** | 9185            | 3509        |

So this brings ant's line count well within the 10K limit, and it also lowers maven's, which was not too far from 10K. We should merge this soon, to avoid breaking the Maven Travis build.

I've also tested this patch on the metadata branch:

          | Without the PR  | With the PR |
| ------- | --------------- | ----------- |
| **ant** | 268404          | 70498       |
| **mvn** | 75640           | 70007       |

In this case we're still well above the 10K limit, so there must be another dominant factor. Still useful to have, but it looks like rebasing this alone will not make things better.

joshmoore:
@simleo : @sbesson recently pointed out that the `meta{data,morph}` PR (in combination with the Memoizer config work) may have introduced a recursion problem which is most visible in the Memoizer warn logging.

PR #2397, opened 2016-05-04
Updating volocity links

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/1245/warnings3Result/
Browsing around, it seems like PE are in the process of updating their website and some of their redirects are a bit shot! This is an older page but the new one (http://www.perkinelmer.com/category/cellular-imaging-software) has no mention of Volocity so it seemed like the best option. Hopefully it'll stick around!

PR #2398, opened 2016-05-04
Updating volocity links (rebased onto dev_5_1)

hflynn:
This is the same as gh-2397 but rebased onto dev_5_1.

----

See https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/1245/warnings3Result/
Browsing around, it seems like PE are in the process of updating their website and some of their redirects are a bit shot! This is an older page but the new one (http://www.perkinelmer.com/category/cellular-imaging-software) has no mention of Volocity so it seemed like the best option. Hopefully it'll stick around!

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -2247,7 +2247,7 @@ mif = true
 
 [Volocity Library Clipping]
 extensions = .acff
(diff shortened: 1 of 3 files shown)


hflynn:
--rebased-from #2397

sbesson:
https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-5.1-merge-docs/1246/ is green.

PR #2401, opened 2016-05-05
Schema Transform Test

dgault:
This is a new Bio-Formats test based off https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/ExporterTest.java

- The main goal of the test is to validate the downgrade and upgrade process of schema transforms.
- The base image data is generated from XMLMockObjects.
- Images are generated for simple images, images with ROI's and images with Annotations
- For each of the above there is a test to validate the image from XMLMockObjects, test the downgrade transforms and test the upgrade transforms
- The transforms performed are parsed from the catalog file ome-transforms.xml
- Additionally upon completing the transforms in each test I have read the image in OMEXMLReader which will in turn upgrade to the current schema and validate it

### DIFF
--- a/components/formats-bsd/test/spec/schema/Schema_Transform_Test.java
+++ b/components/formats-bsd/test/spec/schema/Schema_Transform_Test.java
@@ -1,6 +1,33 @@
 /*
- *   Copyright 2006-2016 University of Dundee. All rights reserved.
- *   Use is subject to license terms supplied in LICENSE.txt
+ * #%L
+ * BSD implementations of Bio-Formats readers and writers
+ * %%
+ * Copyright (C) 2005 - 2016 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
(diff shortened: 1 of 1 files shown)


sbesson:
Does this new test file need to be added to `components/formats-bsd/test/spec/testng.xml` to be picked up by the Maven/Ant test targets?

dgault:
The components/formats-bsd/test/spec/testng.xml file only references the package spec.* which should include this test.
Im not aware of anywhere else it needs to be added but I could be wrong on that

sbesson:
You are absolutely correct. Looking at the Formats BSD test reports for the Travis Maven build, [with this PR](https://travis-ci.org/openmicroscopy/bioformats/jobs/128017488)

```
Tests run: 1103, Failures: 0, Errors: 0, Skipped: 0
```

while from [develop](https://travis-ci.org/openmicroscopy/bioformats/jobs/127821114): 

```
Tests run: 1031, Failures: 0, Errors: 0, Skipped: 0
```

components/formats-bsd/test/spec/schema/Schema_Transform_Test.java
```diff
@@ -0,0 +1,803 @@
+/*
```
sbesson: The header might need to be reformatted to comply with the Bio-Formats styling.

components/formats-bsd/test/spec/schema/Schema_Transform_Test.java
```diff
@@ -0,0 +1,830 @@
...
+                            NodeList tt = node.getElementsByTagName(TARGET);
+                            for (int l = 0; l < tt.getLength(); l++) {
+                                parseTarget((Element) tt.item(l), umap);
+                            }
+                            //parse the map
+                            Iterator<Entry<String, List<String>>> kk = 
+                                    umap.entrySet().iterator();
+                            Entry<String, List<String>> e;
+                            while (kk.hasNext()) {
+                                e = kk.next();
```
mtbc: It doesn't suffice to check `umap.get(schema)` or similar?

rleigh-dundee:
Just as a general question, could some of the logic here be moved into XMLTools as helpers to apply transforms?  For example, getting the list of transforms to apply for a given source and target schema version?  And applying the list of transforms?  This would allow the Bio-Formats code to drop the hardcoded transform sequences, and also for the unit test data to match the behaviour in the code.

components/formats-bsd/test/spec/schema/Schema_Transform_Test.java
dgault: The upgrades map could probably be built directly from umap without the need for the iterator in this case. Initially this code was merely replicating the parsing used in the ExporterTest as closely as possible.

dgault:
I think if people are happy with the logic used here to parse the catalog file and build up the transforms lists then it could perhaps be worth moving it into XMLTools at some stage. The only additional code that Bio-Formats uses when going through the upgrade process is to tidy up namespaces after the transforms 

      transformed = transformed.replaceAll("<ns.*?:", "<");
      transformed = transformed.replaceAll("xmlns:ns.*?=", "xmlns:OME=");
      transformed = transformed.replaceAll("</ns.*?:", "</");

rleigh-dundee:
@dgault Ye gods, that "tidying" is really awful, directly mangling the raw XML like that!  Could we safely delete it entirely?  While I'm sure there may have been historical reasons for adding it, do they exist today?

dgault:
I could certainly run some initial tests on it but I would be nervous removing it completely without some thorough testing across the repo.

sbesson:
I would think the primary reason for having this PR in the first place was to catch potential errors in codegen PRs like https://github.com/openmicroscopy/bioformats/pull/2319 early (rather than at the OMERO level).

Reading through the comments, I see 2 distinct proposals:
- migrating the transforms logic to `components/specification` to make it more generic/reusable
- clean up and modernize the `XMLTools` class

My primary concern here is the conflict with all the work we already scheduled for the upcoming milestone (see https://trello.com/b/OHTqY4pc/bio-formats-5-2-0). For each of these proposals, do we have an estimate of the time but also the person who would be in charge? My proposal is to have  this PR merged and make progress on Bindata for OME Files C++ 0.2.0 immediately. We can initiate the scoping on the XMLTools refactoring /cc @jburel

mtbc:
:+1: for breaking out cards for the further ideas for addressing in later PRs but preferably in some way that doesn't lose them in the sea of what we might or might not do over the coming decade.

rleigh-dundee:
@sbesson Certainly no objection from me to tackle these things in followup PRs.

PR #2402, opened 2016-05-05
Add 5.1.10 version history

hflynn:
Adds version history for 5.1.10 release. Will be staged at https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/about/whats-new.html

PR #2409, opened 2016-05-06
Add 5.1.10 version history (rebased onto develop)

hflynn:
This is the same as gh-2402 but rebased onto develop.

----

Adds version history for 5.1.10 release. Will be staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/about/whats-new.html


hflynn:
--rebased-from #2402

PR #2151, opened 2015-12-08
Updates for Metamorph plates with one file per channel

melissalinkert:
See https://trello.com/c/vpc8NKAC/163-idr-metamorph-stk-per-channel.

The Metamorph changes use stage labels starting with ```Scan ``` to detect when a whole plate is saved in a single .stk file.  That doesn't allow for Z stacks/time series as well as a plate being stored, but I imagine we could extend it if/when we have an example.

The AxisGuesser changes allow a pattern block such as ```<53bp1,dapi>``` to be automatically detected as 2 channels, instead of 2 Z sections

PR #2411, opened 2016-05-09
Fix HCS variant of MetaMorph STK

simleo:
Fixes a problem in HCS-style `.stk` datasets (see #2151) that was causing `MetamorphReader.openBytes` to always return the first plane regardless of the requested index.

#2151 introduced a convention where a `.stk` file whose UIC4 `StageLabel`s are `"Scan A01", "Scan A02", ...` is interpreted as a plate. In particular, image planes are mapped to different series. The problem is that #2151 does not change anything in `MinimalTiffReader.openBytes`, where TIFF IFDs are mapped to plane indices, not series indices.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
@@ -95,6 +95,8 @@ public class MinimalTiffReader extends FormatReader {
 
   protected boolean noSubresolutions = false;
 
+  protected boolean seriesToIFD = false;
+
   /** Number of JPEG 2000 resolution levels. */
   private Integer resolutionLevels;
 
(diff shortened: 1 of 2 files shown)


components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
```diff
@@ -95,6 +95,8 @@
 
   protected boolean noSubresolutions = false;
 
+  protected boolean seriesToIFD = false;
```
melissalinkert: Presumably needs to be reset when ```close()``` is called?

joshmoore:
Present on orca-3 when attempting to load a newly imported plate:

```
2016-05-10 08:01:41,389 WARN  [                   loci.formats.Memoizer] (2-thread-4) config: minimumElapsed wrong type: null
```

See http://orca-3.openmicroscopy.org:48888/webclient/?show=plate-251

joshmoore:
Travis re-open

joshmoore:
Merging for IDR-0.0.5

PR #2410, opened 2016-05-09
Revert "Fix generateDS link"

sbesson:
Reverts openmicroscopy/bioformats#2404

PR #2404, opened 2016-05-06
Fix generateDS link

hflynn:
Dave Kuhlman released an updated version yesterday and the URL has changed.

Should make https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/ green again.

Staged at http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/developers/xsd-fu.html#special-thanks

PR #2412, opened 2016-05-09
Revert "Fix generateDS link" (rebased onto dev_5_1)

sbesson:
This is the same as gh-2410 but rebased onto dev_5_1.

----

Reverts openmicroscopy/bioformats#2404


sbesson:
--rebased-from #2410

mtbc:
Staged at https://www.openmicroscopy.org/site/support/bio-formats5.1-staging/developers/xsd-fu.html#special-thanks. Good to merge.

PR #2414, opened 2016-05-11
log when using workaround for .cif issue from bad TIF tag

mtbc:
Satisfies https://github.com/openmicroscopy/bioformats/pull/2318#issuecomment-218255604 by logging the issue now handled by https://github.com/openmicroscopy/bioformats/pull/2318/commits/1187fe7ea98b715136819996a05a6e038fb5fed7. Unfortunately I am not aware that we have the original data that triggers the issue.


sbesson:
Looks good. Unfortunately we do not have any representative data for the original issue. Our curated samples of the Amnis FlowSight format open without issue when opened with the command-line tools with the debug mode activated. @melissalinkert: is there any additional testing that would be useful here?

melissalinkert:
No objections here - what's under curated is everything we have, so no further testing that I can think of.

PR #2415, opened 2016-05-11
Memoizer.wrap: fix handling of MetadataOptions

simleo:
Addresses a problem where `Memoizer.wrap` was issuing a warning if any of the relevant `MetadataOptions` was `null`. Since `MetadataOptions.getMetadataOption` returns `null` to signal that the requested option is not set, getting a `null` is not a sign that something went wrong, so I've changed `wrap` to silently return the unwrapped reader in this case.

**NOTE: this was breaking the Travis build by increasing the Maven test output from ~9K to ~75K lines.**

Other problems fixed here:

* If the whole `options` container was `null`, `wrap` returned `null`, while the docs state that the original reader would be returned unwrapped (which seems the sensible thing to do, so I've changed the code to match the docs and not the other way around)
* The code was giving up too early by returning the unwrapped reader after encountering an invalid value for `inplace`, while it can still return a `Memoizer` if it gets a valid `cachedir`

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -555,24 +555,30 @@ public class Memoizer extends ReaderWrapper {
     }
     String k = Memoizer.class.getName();
     Object elapsed = options.getMetadataOption(k + ".minimumElapsed");
-    Object inplace = options.getMetadataOption(k + ".inPlace");
-    Object cachedir = options.getMetadataOption(k + ".cacheDirectory");
     if (!(elapsed instanceof Long)) {
+      if (null != elapsed) {
         LOGGER.warn("config: minimumElapsed wrong type: {}", elapsed);
-        return r;
-    } else if (!(inplace instanceof Boolean)) {
-        LOGGER.warn("config: inplace wrong type: {}", inplace);
-        return r;
-    } else if (cachedir != null && !(cachedir instanceof File)) {
-        LOGGER.warn("config: cachedir wrong type: {}", cachedir);
-        return r;
+      }
+      return r;
     }
-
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/Memoizer.java
```diff
@@ -327,6 +327,17 @@ public void saveStop() throws IOException {
 
   }
 
+  private static Object getMetadataOption(
```
mtbc: I wonder if there's useful scope for making this a `private static <X> X get...(... Class<X> type)` or somesuch then having the logging instead catch the exception from `type.cast`. (Something like that anyway.)

simleo:
All Travis builds are OK now, except for the `openjdk6` ones: those will always fail because with the latest rebases we have introduced Java 7 code, so we should just remove `openjdk6` from the Travis matrix.

joshmoore:
:+1: for removing JDK6. Here or in a separate PR.

sbesson:
@simleo: as per https://github.com/openmicroscopy/bioformats/pull/2416, removing `cppwrap` from the Travis matrix might also speed up build times.

simleo:
JDK6 removal: #2417. I guess we could merge that one first, then I can rebase this. Then we can also rebase and merge #2416

joshmoore:
Merged, @simleo

simleo:
Thanks @mtbc, I've addressed your comment in 90781b028a021ea07d41daf1d14b67b8d7bad843

mtbc:
That change looks great, thank you.

simleo:
After rebasing on top of #2417, Travis seems finally happy again

joshmoore:
Merging to re-open the other PRs to get them green.

PR #2419, opened 2016-05-14
Model enum units metadata

rleigh-dundee:
This PR is the first part of the C++ units support needed for the 2015 model.  The primary change is the introduction of extended enum metadata to the schema, for use by the generator (it's all `xsd:appinfo/xsdfu` data so isn't a model change).  This allows the enum name to be different from the language enum name, which removes the need to hardcode the differences in the generator.  It also has alternative names to allow for plurals, language differences and alternative regexes, which also removes hardcoded behaviour from the generator and supplementary configuration files (`enum_handler.cfg`).

- Unit language enum names are now the full unit name, matching OMERO.  The symbol names were inconsistent, caused clashes with system headers for C++, and had capitalisation and encoding issues with the micron symbol.  The corresponding code changes are the bulk of this PR.
- Binning names are not hardcoded in the generator; I've chosen a proper name rather than a mixture of names and symbols, which is a breaking change but is a better choice; can be reverted if compatibility is more important
- The code generator passes through a full copy of the parsed schemas, since bugs in generateds strip out needed information like per-enum metadata; this makes much more information available than previously.  Currently it's only used for the enum metadata.
- The C++ generator changes aren't tested here since we don't use the new model yet.  Future work extends (and corrects) some of the changes made here.

--------

Testing:

- Check jobs are green
- Review generated code changes for Java and C++ https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-generated-sources/lastSuccessfulBuild/artifact/ome-xml-model-java.diff
- Verify that the enum name changes are correct and acceptable
- Verify that downstream users are not affected by the changes; if they are then they will need updating

--------

Followup work

- If the Binning and FontFamily changes to ome.xsd are acceptable, I'd like to move all enum definitions to the top-level like this for consistency with the code generator--it would mean only a single method of handling enums in the generator which would remove some unnecessary complexity from it
- Additional C++ PRs for MapAnnotations and 2015 model generation


components/xsd-fu/python/ome/modeltools/property.py
```diff
@@ -345,6 +351,50 @@ def _get_unitsType(self):
...
         doc="""The property's units type.""")
 
+    def _get_enumProperties(self):
+        if self.isEnumeration:
+            if self.enumProps is None:
+                self.enumProps = dict()
+
+                try:
+                    root = self.model.schemas['ome']
+                    enum = root.find("{http://www.w3.org/2001/XMLSchema}simpleType[@name='%s']" % self.langType)
```
sbesson: Note that using an xpath expression requires ElementTree 1.3+ (see http://stackoverflow.com/a/4809350). However, our Centos6-based CI bundles an older version of the library which explains the [failures of the build](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-build).

rleigh-dundee: Yes, this is quite annoying.  I was about to make a card for this.  A wider question would be whether we should or could drop Python 1.6 support for the bioformats build.  We could switch the jobs to centos 7 nodes or use Python 2.7 on CentOS 6.

rleigh-dundee: See https://trello.com/c/fzgLYbvK/142-drop-python-2-6-support

sbesson:
Barring the Python 2.6 issue discussed above, my main concern about the current state of this PR is the complete removal of the Java unit enums introduced in Bio-Formats 5.1.0.

For consumers of Bio-Formats who have been adjusting their code to account for units, there is no doubt this breaking change would force them to modify their code once again to comply with the new units. Completely dropping this API without having scoped the upcoming steps and warned about it can put both ourselves and our community in a precarious position. On the other hand, I appreciate the homogenization of the unit system across our stack and the overall intent.

Couple of immediate suggestions/thoughts to move forward:
- no strong veto to using the new enums proposed in the schema especially if they match the model/C++/OMERO. I would reintroduce the old enums in `UNITS.java` and deprecate them instead of dropping them.
- should we use the occasion to work on a proper documentation page for the current units system in the Bio-Formats developer documentation? /cc @hflynn 

Once we have frozen the 2016 model, we should definitely take the time to go through the units to-do column and define our next actions.

rleigh-dundee:
@sbesson UNITS.java updated to retain the old names, and (I think) I've fixed the Python 2.6 issues, we'll see shortly.

Regarding documentation, I'm not sure yet.  Might be worth waiting until all the C++ units stuff is done; we might want to port some of that back to Java; UNITS.java is hideous, and might be replaceable entirely.

sbesson:
Note that one side effect of the current state of this PR is that it will break the unit conversion logic in `FormatTools.getPhysicalSize()` which currently relies on the units order - so we will have to make sure https://trello.com/c/7A22zVaV/60-refactor-units-conversions is reviewed.

joshmoore:
:+1: for the deprecation. Minor comments from me:

 * if a squash is trivial it might be worth it
 * Use of the `@Deprecated` annotation would probably be useful at the same time. (It updates the javadoc as well, so if only one is going to be used, I'd side with the annotation)

rleigh-dundee:
@joshmoore I originally tried `@Deprecated` but it seems that it's Java8-only for static members; it failed to compile for me.

I can certainly squash the commits.

joshmoore:
> but it seems that it's Java8-only for static members; it failed to compile for me.

Hmmm.... odd. 

```
$ cat a.java
@Deprecated
public class a {

  @Deprecated
  public static final String t = "t";

  public static void main(String[] args) {
    System.out.println(t);
  }

}
$ javac a.java
$ java a
t
$ javac -version
javac 1.7.0_95
```

Do we perhaps still have a 1.6 setting somewhere?

rleigh-dundee:
@joshmoore Unsure, but I don't think so.  I got the error with OpenJDK8 on Linux.  I'll paste the error message in a short while when I can repeat it.

rleigh-dundee:
@sbesson The enum order is unchanged though; the deprecated UNITS constants are a separate thing.  If you look at the generated code changes, UnitsLength is unchanged in terms of the unit ordering.  (The problems noted with it still need fixing independently of this PR though.)

rleigh-dundee:
@joshmoore Hmm, I can't reproduce the failure now, so I've switched to `@Deprecated` and squashed as requested.  Changes which are logically separate or might be revertable in the future have been left separate.

components/xsd-fu/templates-java/OMEXMLModelEnum.template
```diff
@@ -72,11 +95,22 @@ public enum ${klass.langType} implements Enumeration
...
+{% if klass.enumProperties is None or not value in klass.enumProperties or klass.enumProperties[value].get('enum', None) is None %}\
     if ("${value}".equals(value))
     {
       return ${enum_value_name(value, klass.isUnitsEnumeration).upper()};
     }
 {% end %}\
+{% end %}\
+
+{% for value in klass.possibleValues %}\
+{% end %}\
```
dgault: Is this an old debug loop which can be removed?

components/xsd-fu/templates-java/OMEXMLModelEnum.template
```diff
@@ -61,7 +61,30 @@ import ome.xml.model.primitives.*;
 
 public enum ${klass.langType} implements Enumeration
 {
-  ${", ".join(['%s("%s")' % (enum_value_name(value, klass.isUnitsEnumeration), value) for value in klass.possibleValues])};
+{% for value in klass.possibleValues %}\
+{% if klass.enumProperties is not None and value in klass.enumProperties and klass.enumProperties[value].get('enum', None) is not None %}\
+{% if value == klass.possibleValues[-1] %}\
+  ${klass.enumProperties[value].enum}("${value}") /** ${klass.enumDocumentation[value]} */;
```
sbesson: Minor thought: would it be worth splitting this into two lines so that the enum documentation can also server as Javadoc ?

components/xsd-fu/templates-java/OMEXMLModelEnum.template
rleigh-dundee: Unsure, but it can certainly be removed.  Pushed now.

components/xsd-fu/templates-java/OMEXMLModelEnum.template
rleigh-dundee: I thought this was valid javadoc syntax for enum values?  No matter though, I've pushed a change to move the comments onto the preceding line.

sbesson:
@rleigh-dundee: you are correct about `EnumLength`, I got confused by the Javadoc which is ordering enums alphabetically however the ordering is unchanged in the generated classes.

Do the the breaking changes in the `ome.xml.enums` classes need to be communicated or are the `EnumHandler` classes the user-facing classes for manipulating enums on the Java side?

With regard to the order of follow-up PRs, no strong veto to generalizing the enum definition work initiated in this PR and they decrease the code generation burden but let us make sure the C++ work takes priority.

rleigh-dundee:
@sbesson I think we definitely need to communicate that the enums have changed.  But the impact is likely very low.  If you're using the UNITS statics or the metadata store, you'll not be affected, which should be most people.  Only people who are using the enums directly will be affected.

components/xsd-fu/templates-java/OMEXMLModelEnumHandler.template
```diff
@@ -91,10 +91,15 @@ public class ${klass.langType}EnumHandler implements IEnumerationHandler {
 {% for value in klass.possibleValues %}\
     p.put("^\\\s*${value}\\\s*", "${value}");
 {% end %}\
-{% if klass.langType in config %}\
     // BEGIN custom enumeration mappings
```
dgault: Perhaps an if statement around the comment here to check if klass.possibleValues is not none and greater than zero would be useful. Otherwise the comment appears in the generated code even if there are no custom enum mappings

dgault:
Other than the 2 minor comments made, the templates, resulting generated code and enum renaming  look good to me.

components/xsd-fu/templates-java/OMEXMLModelEnumHandler.template
rleigh-dundee: The conditional ends up being a bit more complex than that--this is always true--so I redid it to make the custom mapping line comments more informative which should hopefully have the same effect.

sbesson:
From my last review of the generated sources diff, in addition to the deprecation of the short `ome.units.UNITS` enums in favor of the long unit names discussed above, the breaking changes arising from this PR concern the following enum classes:
- [ome.xml.model.enums.Binning](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/Binning.html)
- [ome.xml.model.enums.PixelType](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/PixelType.html)
- [ome.xml.model.enums.UnitsElectricPotential](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsElectricPotential.html)
- [ome.xml.model.enums.UnitsFrequency](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsFrequency.html)
- [ome.xml.model.enums.UnitsLength](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsLength.html)
- [ome.xml.model.enums.UnitsPower](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsPower.html)
- [ome.xml.model.enums.UnitsPressure](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsPressure.html)
- [ome.xml.model.enums.UnitsTemperature](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsTemperature.html)
- [ome.xml.model.enums.UnitsTime](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/javadoc/ome/xml/model/enums/UnitsTime.html)

The enum constants defined in the last seven classes should likely be checked against OMERO [UnitsElectricPotential](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsElectricPotential.html), [UnitsFrequency](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsFrequency.html), [UnitsFrequency](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsFrequency.html), [UnitsLength](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsLength.html), [UnitsPower](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsPower.html),  [UnitsPressure](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsPressure.html),  [UnitsTemperature](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsTemperature.html) and [UnitsTime](http://downloads.openmicroscopy.org/omero/5.2.3/api/ome/model/enums/UnitsTime.html) to make sure we are consistent.

sbesson:
Discussed the PR and its potential impact with @joshmoore and @rleigh-dundee. Added a [documentation task](https://trello.com/c/FvyQj7OP/144-document-unit-changes) and a [review card](https://trello.com/c/CCldRgeg/145-final-review-of-the-model-breaking-changes) to complete as part of this effort for the model freeze. Merging.

PR #2421, opened 2016-05-16
Do not run upgrade check when passing -version

sbesson:
Fixes  https://trac.openmicroscopy.org/ome/ticket/13226

As indicated in the command help, this option should display the current library version and exit. This commit skips the initial upgrade check if the argument is passed.

To test this PR, run the command line tools *offline* and check:
- `showinf -version` and `bfconvert -version` do not throw a `java.net.UnknownHostException`
- `showinf test.fake` and `bfconvert test.fake` throw a `java.net.UnknownHostException`

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -197,7 +197,6 @@ public class ImageInfo {
           return true;
         }
         else if (args[i].equals("-nopix")) pixels = false;
-        else if (args[i].equals(VERSION)) printVersion = true;
         else if (args[i].equals("-nocore")) doCore = false;
         else if (args[i].equals("-nometa")) doMeta = false;
         else if (args[i].equals("-nofilter")) filter = false;
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -93,6 +93,7 @@ public class ImageInfo {
   private static final String NEWLINE = System.getProperty("line.separator");
 
   private static final String NO_UPGRADE_CHECK = "-no-upgrade";
+  private static final String VERSION = "-version";
 
   private static final ImmutableSet<String> HELP_ARGUMENTS =
       ImmutableSet.of("-h", "-help", "--help");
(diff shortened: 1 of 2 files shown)


mtbc:
1. The last part of the ticket description still occurs: If I *do* still give `bfconvert` the `-no-upgrade` option then I'm still bizarrely told, `Found unknown command flag: -version; exiting.`
1. Shouldn't the upgrade check fail slightly more gracefully? Stack traces shouldn't occur in routine situations. (I'm not actually sure why we bother telling the user that they're offline but, if it's important to, then we could maybe do so less scarily?)

Still, this PR does appear to be an improvement, thank you! The code refactoring in general looks positive.

sbesson:
Last commit should fix the first issue and also handle the `-version` argument similarly to most command-line programs, i.e. return early independently of the next arguments, so that the following should all return the version

```
showinf -debug -version
showinf -version
showinf -version test.fake
showinf -version -foo

bfconvert -debug -version
bfconvert -version
bfconvert -version test.fake
bfconvert -version test.fake test.ome.tiff
bfconvert -version -foo
````

Also as reported in https://trello.com/c/FHWeDznP/140-cli-system-out-vs-logging, the version is now printed via `System.out.println`. This can be tested by changing `tools/logback.xml` to put the logging OFF and checking the commands above still output the version.

The second proposal also makes sense but having review the various usage of the code, I would be inclined to start tackling https://trac.openmicroscopy.org/ome/ticket/8260 first and migrate these generic functions (print version, run upgrade check) to an utility class.

**Comment by mtbc** [comment]
Could we reopen the ticket for tackling the second proposal?

File: components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java [hunk 1]
```diff
@@ -191,8 +192,12 @@ public boolean parseArgs(String[] args) {
...
     for (int i=0; i<args.length; i++) {
       if (args[i].startsWith("-")) {
-        if (args[i].equals("-nopix")) pixels = false;
-        else if (args[i].equals("-version")) printVersion = true;
+        if (args[i].equals(VERSION)){
+          printVersion = true;
+          return true;
+        }
+        else if (args[i].equals("-nopix")) pixels = false;
+        else if (args[i].equals(VERSION)) printVersion = true;
```
mtbc: didn't we check this just above?

mtbc:
If I give `-version` *and* something to actually do, it just prints the version: that's as intended? (Not sure what https://github.com/openmicroscopy/bioformats/pull/2421#issuecomment-220018171 has in mind.)

sbesson:
@mtbc: yes `-version` shows the version and exits similar to e.g. `psql --version omero`.

mtbc:
Then, good to merge!

melissalinkert:
Good to merge from my side as well.

PR #2422, opened 2016-05-17
FilePatternReader: fix openBytes

simleo:
Adds the following to `FilePatternReader.openBytes`:

* a `helper.setSeries` call to so that the correct planes are returned when stitching multi-series files
* a `helper.close` call to match `helper.setId` (avoids file leaks)

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FilePatternReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FilePatternReader.java
@@ -134,7 +134,9 @@ public class FilePatternReader extends FormatReader {
     int planeIndex = fileIndexes[getSeries()][no][1];
     helper.setId(files[fileIndex]);
     helper.setSeries(getSeries());
-    return helper.openBytes(planeIndex, buf, x, y, w, h);
+    byte[] plane = helper.openBytes(planeIndex, buf, x, y, w, h);
+    helper.close();
+    return plane;
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/in/FilePatternReader.java
```diff
@@ -133,7 +133,11 @@ public FilePatternReader() {
     int fileIndex = fileIndexes[getSeries()][no][0];
     int planeIndex = fileIndexes[getSeries()][no][1];
     helper.setId(files[fileIndex]);
-    return helper.openBytes(planeIndex, buf, x, y, w, h);
+    int oldSeries = helper.getSeries();
+    helper.setSeries(getSeries());
+    byte[] plane = helper.openBytes(planeIndex, buf, x, y, w, h);
+    helper.setSeries(oldSeries);
```
joshmoore: Happy to test as is, but this might should be in a `finally {}`

components/formats-bsd/src/loci/formats/in/FilePatternReader.java
```diff
@@ -133,7 +133,11 @@ public FilePatternReader() {
     int fileIndex = fileIndexes[getSeries()][no][0];
     int planeIndex = fileIndexes[getSeries()][no][1];
     helper.setId(files[fileIndex]);
-    return helper.openBytes(planeIndex, buf, x, y, w, h);
+    int oldSeries = helper.getSeries();
+    helper.setSeries(getSeries());
```
melissalinkert: This will break the case where ```FilePatternReader``` has multiple series, but each file has a single series.  I don't know off hand if any of the IDR datasets exhibit this, but would be worth verifying before merging.

I'd also expect just ```helper.setSeries(getSeries())``` to be sufficient (without saving/resetting the series afterwards), but if there is a reason for the extra calls it might be good to document it.

joshmoore:
Merging for IDR-0.0.5

PR #2424, opened 2016-05-18
Tests and logging (metadata)

simleo:
Rebases #2396 onto metadata. Brings the Maven test output line count from 9158 down to 3525 (Maven 3.0.5 on my laptop), reducing the risk that Travis will break again soon.

### DIFF
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -65,7 +65,7 @@ Type "ant -p" for a list of targets.
       <then>
         <junit haltonfailure="true" fork="true">
           <classpath>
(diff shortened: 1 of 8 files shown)


joshmoore:
Makes sense and green. Was this a manual rebase to have lost the original commit messages?

simleo:
@joshmoore yes, I've applied the whole diff and made a single commit. Is that bad practice?

joshmoore:
@simleo : it loses the authorship (in this case from @sbesson) . Typically a rebase is used. Considering this is a minor PR on a minor branch (in a minor arm of a minor galaxy). Merging. Happy to discuss the workflow elsewhere.

PR #2427, opened 2016-05-18
ome-xml: Use language-specific container types for maps

rleigh-dundee:
This makes some (non-breaking) model changes which result in some breaking model (but not metadata) API changes.  The overall aim is to allow the C++ code generation to function; the existing approach was fundamentally incompatible with metadata conversion in terms of having a usable concrete map container instance which could be safely used with get/set methods on the metadatastore to transfer metadata between stores.  It can also be considered to be fixing some historical design warts with the Map model and implementation which clean it up and make the code generator, templates and resulting API saner, and should perhaps have been done at the time we originally introduced these features.

The type used with the MetadataStore can possibly be reverted if this is unacceptable.  The model changes are needed, however.

Also note that we might need to make the corresponding changes to the 2015 schema to get C++ to work with it; that wouldn't be a breaking change, but it might be something we want to think carefully about before doing.

See also: https://github.com/openmicroscopy/openmicroscopy/pull/4671 for the corresponding OMERO changes.

--------

Summary of changes

model:
  - `Map` is a complexType, not an element.  This removes the generation of the useless `Map` model object.  This does not change the schema; it was never used as a concrete element.  This is needed for code generation sanity (the generated inheritance hierarchy).
  - Drop `MapPairs`; this is also non-breaking since it was also never used directly in OME-XML.
  - `MapAnnotation` `Value`, `ImagingEnvironment` `Map` and `GenericExcitationSource` `Map` are now implemented in terms of the `Map` complexType.
  The above changes do not change the validation of the OME-XML in any shape or form.

xsd-fu:
  - Add Map support to the C++ templates; it's currently disabled when using the 2013-06 schema.
  - At the model object level, redundant `getPairs()` calls are eliminated entirely.  The model object template handles `List<MapPair` serialisation directly.

cpp:
  - Add `OrderedMultimap` model primitive and tests to validate duplicate keys and key/value pairs, and preservation of insertion order.
  - Drop (unused) `MapPairs` model object.

java:
  - Simplify xsd-fu template usage; List<MapPair> is a model primitive.
  - Drop `MapPairs`; this type is now redundant.
  - Unit tests check that the same key may be used multiple times and that insertion order is preserved.  They also verify that the same key/value pair may be resused.

Source changes: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-generated-sources/219/

--------

Testing:

- Check generated source code differences
    - https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-generated-sources/lastSuccessfulBuild/artifact/ome-xml-metadata-java.diff/*view*/
    - https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-generated-sources/lastSuccessfulBuild/artifact/ome-xml-model-java.diff/*view*/
- Check kyro can serialise readers using the different types of Map annotation
- Check OMERO can import files with map annotations
- You'll see that there are no C++ model changes in this PR; these will be enabled with the 2016-DEV0/2015 model in a followup PR; what's here has been built and tested on my `model-units-cpp3` branch (from which this PR was factored out).  C++ testing is limited to checking the jobs remain green.


rleigh-dundee:
I'll drop the now redundant `java.util.Map` additions from the generated and static sources; they were used at one point but were dropped.

dgault:
Im currently working on trying to get some unit tests for the serialisation of some of these MultiMap types, once done that should perhaps help with the testing of this PR also.

components/formats-bsd/test/loci/formats/utests/GenericExcitationMapTest.java
```diff
@@ -132,20 +136,21 @@ public void testGenericExcitationSourceMapContent() throws Exception {
...
+    assertPair(dataMap, 3, "b", "4");
     assertPair(dataMap, 4, "e", "5");
+    assertPair(dataMap, 5, "c", "6");
 
   }
 
-  void assertPair(Map dataMap, int idx, String name, String value) {
-    assertEquals(name, dataMap.getPairs().get(idx).getName());
-    assertEquals(value, dataMap.getPairs().get(idx).getValue());
+  void assertPair(LinkedListMultimap dataMap, int idx, String name, String value) {
```
mtbc: Make `LinkedListMultimap` a `LinkedListMultimap<String, String>` and you won't need the casts.

components/formats-bsd/test/loci/formats/utests/GenericExcitationMapTest.java
```diff
@@ -76,12 +79,13 @@ public void setUp() throws Exception {
     // Add a GenericExcitationSource with an Map
     GenericExcitationSource geSource = new GenericExcitationSource();
     geSource.setID("LightSource:0");
-    Map dataMap = new Map();
-    assertPair(dataMap, 0, "a", "1");
-    assertPair(dataMap, 1, "b", "2");
-    assertPair(dataMap, 2, "c", "3");
-    assertPair(dataMap, 3, "d", "4");
-    assertPair(dataMap, 4, "e", "5");
+    LinkedListMultimap dataMap = LinkedListMultimap.create();
```
mtbc: `<String, String>` is good habit even if compiler doesn't insist

components/formats-bsd/test/loci/formats/utests/MapAnnotationTest.java
```diff
@@ -129,14 +130,21 @@ public void testMapAnnotationValueContent() throws Exception {
...
+    assertEquals("d", dataMap.entries().get(0).getKey());
+    assertEquals("2", dataMap.entries().get(0).getValue());
+    assertEquals("c", dataMap.entries().get(0).getKey());
+    assertEquals("3", dataMap.entries().get(0).getValue());
+    assertEquals("b", dataMap.entries().get(0).getKey());
+    assertEquals("4", dataMap.entries().get(0).getValue());
+    assertEquals("e", dataMap.entries().get(0).getKey());
+    assertEquals("5", dataMap.entries().get(0).getValue());
+    assertEquals("c", dataMap.entries().get(0).getKey());
+    assertEquals("6", dataMap.entries().get(0).getValue());
```
mtbc: This is strange stuff. It really passes? (Can't tell easily locally, I assume because of decoupling.) Would an iterator help?

components/formats-bsd/test/loci/formats/utests/GenericExcitationMapTest.java
rleigh-dundee: D'oh.  Makes perfect sense.

components/formats-bsd/test/loci/formats/utests/MapAnnotationTest.java
rleigh-dundee: Definitely an oversight, and it should definitely be failing.  Maybe it's not being run for some reason?

WRT decoupling, it should run directly with "ant jars test".  That's how I'm running them anyway.

mtbc: Ah, I was using netbeans, but it may actually be a problem with my project setup.

rleigh-dundee:
https://github.com/rleigh-dundee/bioformats/compare/model-map-refactor...rleigh-dundee:model-map-refactor-oldcontainer?expand=1 are the needed changes to retain the internal model refactoring, but preserve the metadatastore API.  It's less than ideal, and I'd prefer we don't use this unless strictly necessary, but it's here if there's no better choice.  Note this is only altering the Java metadatastore; the rest of the original PR is still used, and MapPairs/Map are still deleted.

rleigh-dundee:
The last two bits are cleanup only; they don't have any functional effect, but the main changes clean up the model sufficiently that some of the special-case hacks can be dropped.

components/formats-bsd/src/loci/formats/in/FakeReader.java
```diff
@@ -865,9 +866,10 @@ private void fillAnnotations(MetadataStore store, int imageIndex) {
       annotationID = ANNOTATION_PREFIX + annotationCount;
       store.setMapAnnotationID(annotationID, annotationMapCount);
       store.setMapAnnotationNamespace(ANNOTATION_NAMESPACE, annotationMapCount);
-      List<MapPair> mapValue = new ArrayList<MapPair>();
+      LinkedListMultimap<String, String> mapValue = LinkedListMultimap.create();
       for (int keyNum=0; keyNum<10; keyNum++) {
-        mapValue.add(new MapPair("keyS" + imageIndex + "N" + keyNum, "val" + (keyNum+1)*(annotationCount+1)));
+          mapValue.put("keyS" + imageIndex + "N" + keyNum,
```
mtbc: here and elsewhere you seem to be shifting from the usual two-space indent

mtbc:
This PR does seem to simplify the code somewhat.

sbesson:
Looking at this PR, the companion OMERO one and the generated sources diff, some of the concerns I can see raised are:

- what are the pros/cons of changing our policy to using Guava in the public API? - see also recent discussion in  https://github.com/openmicroscopy/bioformats/pull/2382
- what is the impact of these API changes on the caching `Memoizer` wrapper where recent investigation seems to indicate issues with Guava maps?
- more globally, what is the impact of the API changes for the workflows extensively using key/value pairs notably on the IDR side?

For changes with such impact, I think all concerns need to be seriously discussed across the various components of the project and notably include the IDR team. Also the testing steps need to be more concrete/detailed and likely performed at scale as we has been successfully done for the previous model/API changes over the last few months. Given the imminent timelines, @rleigh-dundee can you schedule a call to discuss these changes?

mtbc:
Regarding memoizers I wonder if it would help to look at contributing to https://github.com/magro/kryo-serializers/tree/master/src/main/java/de/javakaffee/kryoserializers/guava for whichever we need.

rleigh-dundee:
@mtbc Indentation should now be fixed
@sbesson Map tests added and should now pass as well

Happy to have a wider discussion regarding the details.  Since we're using a standard (not immutable) container here, are there any problems with guava at all for LinkedListMultimap?  If it's OK, then that's the major concern removed regarding kryo/memo stuff.  The main reason for the change is to use a standard container API rather than our own ad-hoc (and entirely unnecessary and non-standard) one, which would greatly ease interoperability; `MapPair` never needed to exist.

mtbc:
My guess is that the issue isn't the immutability, it's the lack of a no-argument constructor.

rleigh-dundee:
OK.  Let's stick with the existing approach for this PR then.  The last commit reverts the interface change, while retaining the internal refactoring/cleanup.  It's something which should be revertable in the future if this is desired.

rleigh-dundee:
Green in https://ci.openmicroscopy.org/view/Files-DEV/job/OME-FILES-CPP-DEV-breaking/103/

rleigh-dundee:
Source changes: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-generated-sources/219/

components/ome-xml/pom.xml
```diff
@@ -46,6 +46,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.google.guava</groupId>
```
sbesson: With https://github.com/commit/f2892ff0b9261bfe22515e3ef1d791f86b4fb0bf, I would think this and the corresponding ant dependency addition can be dropped.

rleigh-dundee: Will do.

rleigh-dundee: By the way, is there any way of having a PR in both the regular merge build and the breaking build at the same time?  This one is now no longer breaking and should be fine to move over to the regular build, as are some of the others.

sbesson: @rleigh-dundee: the defaut merge command for the [breaking push job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-push/) uses `merge -D none -I label:breaking -E label:exclude`. What you suggest should be achievable either by dropping `-D none` to include all organization PRs by default or by including specific PRs by number e.g  `merge -D none -I breaking -E exclude -I #xxx`.

components/formats-api/src/loci/formats/meta/MetadataConverter.java
```diff
@@ -884,7 +884,7 @@ private static void convertImages(MetadataRetrieve src, MetadataStore dest,
       catch (NullPointerException e) { }
 
       try {
-        List<MapPair> map = src.getImagingEnvironmentMap(i);
+        java.util.List<ome.xml.model.MapPair> map = src.getImagingEnvironmentMap(i);
```
sbesson: This could be reduced to `List<MapPair>` (i.e. resulting in no change in this file)

components/formats-bsd/src/loci/formats/in/FakeReader.java
```diff
@@ -865,9 +865,10 @@ private void fillAnnotations(MetadataStore store, int imageIndex) {
       annotationID = ANNOTATION_PREFIX + annotationCount;
       store.setMapAnnotationID(annotationID, annotationMapCount);
       store.setMapAnnotationNamespace(ANNOTATION_NAMESPACE, annotationMapCount);
-      List<MapPair> mapValue = new ArrayList<MapPair>();
+      java.util.List<ome.xml.model.MapPair> mapValue = new java.util.ArrayList<MapPair>();
```
sbesson: Same as above

components/formats-gpl/utils/writeMapAnnotationsExample.java
```diff
@@ -59,13 +58,10 @@ public static void main(String[] args) throws Exception{
...
 
-        //Create MapPair Object and add to List
-        ArrayList<MapPair> mapList = new ArrayList<MapPair>();
-        MapPair mapPair;
-        mapPair = new MapPair("Example Key","Example Value");
-        mapList.add(mapPair);
-        mapPair = new MapPair("Bio-Formats Version", FormatTools.VERSION);
-        mapList.add(mapPair);
+        // Create MapPair Object and add to List
+        java.util.List<ome.xml.model.MapPair> mapList = new java.util.ArrayList<ome.xml.model.MapPair>();
```
sbesson: Re-use import and drop the full package name as this is an example?

sbesson:
On the [Java code generated sources](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-generated-sources), the following block has been removed from  `Shape.java`
```
-		if (transform != null)
-		{
-			// Element property Transform which is complex (has
-			// sub-elements)
-			Shape_element.appendChild(transform.asXMLElement(document));
-		}
```


A few minor items spotted (can be addressed separately):
- all `ome.xml.model` classes now include an unnecessary top-level `import java.util.Map`
- the usages of `java.util.List<ome.xml.model.MapPair>` could be collapsed to `List<MapPair>`

I will start testing the `MapAnnotation`, `ImagingEnvironment` and `GenericExcitationSource` changes.

rleigh-dundee:
@sbesson I've pushed the corrections for the package namespace dropping.  I'll push a Map one shortly.

On the codegen front, note for this PR the changes are here: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-generated-sources/219/ (the current diffs include changes from the other PRs)  That erroneous change will likely need fixing up in one of them.

cpp/test/ome-xml/CMakeLists.txt
```diff
@@ -64,6 +64,10 @@ if(BUILD_TESTS)
   target_link_libraries(constrained-numeric OME::XML)
   target_link_libraries(constrained-numeric ome-test)
 
+  add_executable(orderedmultimap orderedmultimap.cpp)
+  target_link_libraries(orderedmultimap)
```
dgault: Does this need to be linked to the OME XML library? - target_link_libraries(orderedmultimap OME::XML)

rleigh-dundee: No, it's a header-only class so doesn't use any of the library facilities.

rleigh-dundee: That said, the line is redundant and could be removed entirely.

rleigh-dundee:
@sbesson The latest code generation changes are: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-generated-sources/246/

It fixes up the remaining instances of `List<MapPair>` in the code, and drops an unnecessary `List` import.

dgault:
Having retested through this today, Im happy with the CPP changes here.

The existing 2013-06 build continues to build as normal without any changes in the generated cpp source. From a CPP model and metadata standpoint it means that we don't have any breaking changes with the PR as we have never been building the later models unlike in the Java.

Judging from the generated changes when combined with https://github.com/openmicroscopy/bioformats/pull/2442 the CPP changes for the later models appear to be working as expected also with the OrderedMultiMap being used correctly.

All tests also appear to be green from the CPP side.

sbesson:
Tested various sample files using the new `Map` complex type including the `mapannotation.ome.xml` sample file as well as `instrument.ome.xml` with the following diff

```
sbesson@ls30630:bioformats (map_Samples) $ git diff
diff --git a/components/specification/samples/2015-01/instrument.ome.xml b/components/specification/samples/2015-01/instrument.ome.xml
index 533028d..6309000 100644
--- a/components/specification/samples/2015-01/instrument.ome.xml
+++ b/components/specification/samples/2015-01/instrument.ome.xml
@@ -21,6 +21,14 @@
                  SerialNumber="A123456">
       <Arc Type="Xe"/>
     </LightSource>
+    <LightSource ID="LightSource:2">
+      <GenericExcitationSource>
+        <Map>
+          <M K="LightSourceKey-1">LightSourceValue-1</M>
+          <M K="LightSourceKey-2">LightSourceValue-2</M>
+        </Map>
+      </GenericExcitationSource>
+    </LightSource>
     <Detector ID="Detector:0" Type="CCD" Manufacturer="OME Detectors" Model="Standard CCD Mk2"/>
     <Objective Correction="UV" Immersion="Air" LensNA="1.2" NominalMagnification="60"
                CalibratedMagnification="60.12" WorkingDistance="20" ID="Objective:0"
@@ -53,6 +61,11 @@
     <OME:AcquisitionDate>2010-02-23T12:51:30</OME:AcquisitionDate>
     <Description>This image is linked to the Control Set</Description>
     <InstrumentRef ID="Instrument:0"/>
+    <ImagingEnvironment Temperature="-80.0" TemperatureUnit="°C">
+      <Map>
+        <M K="ImagingKey-1">ImagingValue-1</M>
+      </Map>
+    </ImagingEnvironment>
     <OME:Pixels DimensionOrder="XYCZT" ID="Pixels:0:0" PhysicalSizeX="10000.0"
                 PhysicalSizeY="10000.0" Type="uint8" SizeC="1" SizeT="1" SizeX="6" SizeY="4" SizeZ="1">
       <OME:Channel Color="-2147483648" ID="Channel:0:1">
```

Tested various workflows: command-line display (`showinf -omexml`) and conversion (`bfconvert`), import into an OMERO server with this PR included, download and re-export as OME-TIFF. Most of them worked as expected with the maps being properly handled and the corresponding `ImagingEnvironment_Map` and `GenericExcitationSource_Map` DB tables being populated. The export functionality loses the instrument and the map annotation however this is an export deficiency independent of this PR. /cc @mtbc 

Depending on the progress on the scoping in https://trello.com/c/fB3wiIf4/138-ome-xml-samples-model-coverage, we might consider including changes similar as the ones above to the new 2016 samples to extend our coverage of the model.

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/192/ for the validation against the data repository with this PR included. Next step will be to document the recent set of changes brought to the model in order to bring C++ support for the recent model versions. /cc @hflynn.

PR #2430, opened 2016-05-23
Memoizer pre register readers

dgault:
This is a a basic improvement to pre register with Kryo the reader classes prior to saving or loading a memo file. A hash value is generated based on the class name, however due to a restriction of Kryo we have to use the absolute value of the hash code.

This does not guarantee that the generated id's will be unique, if we later add readers there is a possibility that that a clash may occur in which case the KryoException will be thrown. What it  does give us is consistent registering of reader classes, hopefully preventing any issues between separate jvm sessions.

One major point is that any previously generated memo files will be required to be regenerated and will be considered invalid.

Testing this is a tricky prospect as there are a number of possible workflows for the Memoizer. Previous issues have been largely reported round the ScreenReader as it requires multiple readers to be registered. A real world scenario which would be a good test case would be:
- Clearing existing memo files
- Loading a dataset which creates the new memo files
- Verifying that the new memo files load
- Kill the JVM process and relaunch
- Verify that the existing memo files still load correctly


joshmoore:
@dgault : thanks for this. Excluded since I don't have time for a regeneration at this point. My primary concern looking at this would be a Class name hash collision. Any thoughts?

dgault:
It is certainly a possibility, there is always the potential of a clash, added to the fact that the Kryo registration requires a positive int, thus increasing the potential collisions even further. If there were any collisions discovered then the implementation would simply be unusable. Im still trying to think of a cleverer solution which would be scalable without the need for hardcoding.

sbesson:
Closing for now. Can be reopened later when @joshmoore needs to reinitiate the `memoization` work.

PR #2434, opened 2016-05-25
Schema publication

sbesson:
See https://trello.com/c/zkbHhnNO/31-schema-release-process and https://github.com/openmicroscopy/bioformats/pull/1985 for an earlier version of this PR

This PR streamlines the XSD schema release workflow as follows:

- unifies the schemas generation under a single root to include both pre and post 2007 schemas including @rleigh-dundee's suggestion in https://github.com/openmicroscopy/bioformats/pull/1985#issuecomment-142570844
- removes deprecated `web-extras` and `Samples` which is managed and released independently under a different location
  
To test this PR

- run ``cd components/specification && ./publish`` and check no error is thrown
- review the diff between the `published` folder generated under specification and the live schemas. The pre-2007 schemas should be added to the the output folder and the relative links should no longer use `../XMLschemas/*`
- to help computing the difference, released schemas without the generation documentation have been synced to the `master` branch of https://github.com/sbesson/schemas. It should be possible to copy the content of the `published` folder and perform a `git diff`.

For deployment/URLs:
- all XSD schemas should be deployed under a single root location served at `http://www.openmicroscopy.org/Schemas` i.e. use `http://www.openmicroscopy.org/Schemas/OME/$release/ome.xsd`
- for the pre-2007 schemas currently served at `http://www.openmicroscopy.org/XMLSchemas`. there are 2 possibliities. Either leave them unmodified or also unify the URLs and let `http://www.openmicroscopy.org/XMLSchemas` be an alias of `http://www.openmicroscopy.org/Schemas`
- deployment can happen either by archiving the content of `./publish` via a job and deploying using `scc deploy`. If the Git repository/review process described above is sensible, we could also formalize the usage of https://github.com/ome/schemas which should allow to review diffs between releases and use a clone of this repository on the production server.

Follow-up efforts:
- the generated oXygen documentation needs review. Since the current workflow requires the xsd schemas to be published and live *before* the documentation is generated, the proposal is to generate them independently (and reproducibly) under a different root location. The generated documentation can still be served at `http://www.openmicroscopy.org/Schemas/Documentatation/Generated/OME-$release` or we can use the opportunity to move to a a different URL /cc @hflynn

### DIFF
--- a/components/specification/publish
+++ b/components/specification/publish
@@ -152,7 +152,7 @@ contains() {
     fi
 }
 
-devmarker="dev"
+devmarker="DEV"
 root=published
 
 
--- a/components/specification/publish
+++ b/components/specification/publish
@@ -163,10 +163,6 @@ mkdir -p "$root"
 
 for dir in released-schema/20*
 do
-    # Don't copy legacy or other material
-    if [ "$dir" = "released-schema/additions" ]; then
-        continue
-    fi
 
     # Copy dev material, but don't make index
     if contains $dir $devmarker; then
--- a/components/specification/publish
+++ b/components/specification/publish
@@ -217,7 +217,7 @@ do
 
         # Special case copying of legacy schemas
         if [ "$year" = "2003" ]; then
-            path="${root}/XMLschemas/${base}/${month}"
+            path="${root}/Schemas/${base}/${month}"
             path=$(echo "$path" | sed -e 's;/2003-\(.*\);/\1;')
             reldate="2003 ($month)"
         else
@@ -292,7 +292,6 @@ done
 # $3=filter (current or legacy)
 # $4=verbose (true or false)
 # $5=schema set (current or legacy)
-# $6=schema location (current=Schemas, legacy=XMLschemas)
 createlink() {
     name=$(basename "$1")
     release="$2"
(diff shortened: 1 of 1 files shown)


rleigh-dundee:
For the changes in the PR, I've reviewed the script and the published data, and everything looks fine.
The script changes are good.  The published changes are good.  I compared by

- running `publish
- copying `published/*/*` to a test directory
- deleting `Samples` from the test directory
- merging this PR and re-running `publish`
- diffing the two directories with `-urN`

The only changes are the fixed up relative URLs, and the dropping of the 2016-DEV0 schema page.  Looking at the generated HTML pages, the links are all fine, and the pages contain all the historical schemas, back to 2003 when appropriate.

I think this looks fine to merge.  A simple redirect would allow the URLs to be unified and not break the XMLSchemas URL.

I'm unsure that a dedicated git repository is necessary here--it's possible to test without git quite easily and have that made available with a jenkins job; it's just a diff -urN on the old and new published content.

sbesson:
Thanks for the review. I realized that the `2016-DEV0` schema is copied into `published` although unlisted in the landing page. I will try to push an extra commit to exclude development schemas/transforms.

Regarding the Git repository, agreed it is not necessary, the only drawback at the moment is that we  have no tag/commit corresponding to the latest schema release e.g. if we decide to rework the `index.html` pages. Assuming everyone is happy with computing the diff against the last release or the production content, let's put this idea on the backburner for now.

rleigh-dundee:
@sbesson When we "release" a schema, we could always tag in bioformats.git, e.g. schema/2015-01 so it's not confused with release tags.  The same could also apply when ome-xml is split out into a separate repo.  Then getting back to that point is as simple as checking out (or branching from) that tag and running the publish script.

rleigh-dundee:
The latest change dropping the development transforms/schema looks fine.

Please merge.

PR #2435, opened 2016-05-25
Memoizer: Add failIfMissing flag for perf

joshmoore:
If `failIfMissing` is set to true then no `setId` call
will be performed on the underlying reader. Instead, if
`loadMemo` returns null, a `FormatException` will be thrown
signaling that no cache is available.


joshmoore:
Merging for IDR-0.0.5

aleksandra-tarkowska:
--rebased-to #2806

Issue #2431, opened 2016-05-25
ZeissCZIReader throws a NumberFormatException


gatagat:
Hi all,

we have a couple of files that make the ``ZeissCZIReader`` to fail (using bioformats from within imagej).

Trying to check the file with showinf results in the following stacktrace:

    Exception in thread "main" java.lang.NumberFormatException: null
        at java.lang.Integer.parseInt(Integer.java:443)
        at java.lang.Integer.parseInt(Integer.java:514)
        at loci.formats.in.ZeissCZIReader.translateExperiment(ZeissCZIReader.java:2268)
        at loci.formats.in.ZeissCZIReader.translateMetadata(ZeissCZIReader.java:1329)
        at loci.formats.in.ZeissCZIReader.initFile(ZeissCZIReader.java:731)
        at loci.formats.FormatReader.setId(FormatReader.java:1426)
        at loci.formats.ImageReader.setId(ImageReader.java:835)
        at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:651)
        at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1028)
        at loci.formats.tools.ImageInfo.main(ImageInfo.java:1117)

The bioformats version should be up to date:

    ./showinf -version
    Version: 5.1.10
    VCS revision: 3125500841deff07687b25eb3f430dca62dee648
    Build date: 6 May 2016

Let me know where can I upload the test data.

Thanks for any help.

Best regards,
Tomas

gatagat:
A similar issue has been discussed on the ome-users mailing list:
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2015-January/004988.html
From the discussion there it seems it was resolved and a patch merged, however I see the error with the current version (5.1.10).

Another mention of the issue just two months ago on the imagej mailing list:
http://permalink.gmane.org/gmane.comp.java.imagej/40304

dgault:
Hi Tomas,

Thank you for reporting this issue. There have been similar issues with NumberFormatExceptions reported and fixed in the ZeissCZIReader previously, however looking at the code there may still be further checks that need put in place for certain metadata combinations.

If you could upload a sample file I will be able to reproduce and further investigate the issue - https://www.openmicroscopy.org/qa2/qa/upload/ 

With Thanks,
David Gault

dgault:
Hi Tomas,

Thank for providing a QA file in order to help reproduce this issue.
[...]

PR #2437, opened 2016-05-26
ZeissCZIReader - Number Format Exception fix

dgault:
Issue was raised in issue: https://github.com/openmicroscopy/bioformats/issues/2431
Sample QA filed provided is QA-17199

Null pointer occured when the position object is not null but the values of child are null. This resulted in a number format exception while attempting to parse the null value to integer.

I have put in place extra null pointer and isEmpty checks to deal with this scenario.

To reproduce:
- Attempt to load the file provided in QA-17199
- Verify that a Number Format Exception is thrown

To test:
- Verify that all tests and builds remain green
- Verify that the file provided in QA-17199 imports successfully without errors and displays as expected


components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -2264,9 +2264,11 @@ private void translateExperiment(Element root) throws FormatException {
         Element group = (Element) groups.item(i);
 
         Element position = getFirstNode(group, "Position");
-        if (position != null) {
-          int tilesX = Integer.parseInt(getFirstNodeValue(group, "TilesX"));
-          int tilesY = Integer.parseInt(getFirstNodeValue(group, "TilesY"));
+        String tilesXValue = getFirstNodeValue(group, "TilesX");
+        String tilesYValue = getFirstNodeValue(group, "TilesY");
```
sbesson: Is it worth making use of the data parsing API introduced in https://github.com/openmicroscopy/bioformats/pull/2285 to catch all cases of `NumberFormatException`?

dgault: Using DataTools to handle the exceptions would be a better solution, I will make the necessary adjustments here.

sbesson:
See also the [ome-users thread](http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-June/006056.html) and the corresponding QA 17229. Both QA files should be included in the testing of this PR and ideally one of them should be added to the curated data repository to prevent future regressions.

dgault:
Having gone through and made the changes to use DataTools throughout the reader I have come across a few additional problems with this method. As such I have checked in a small sample relevant to this PR as an example.

In short, we end up with exceptions either way we parse the int:
Original - `int a = Integer.parseInt(intString); ` - Throws a NumberFormatException
DataTools - `int a = DataTools.parseInteger(intString); ` - Throws a NullPointerException when assigning null to a primitive int

To avoid the NullPointerException we could then use:
Integer a = DataTools.parseInteger(intString);

But often we are then using these values in calculations or converting to Lengths etc, causing NullPointerExceptions down the line.

sbesson:
@dgault: you are absolutely correct about the generic `null` handling. For the construction of `Length` quantities, my impression is that reusing our `FormatTools`  helper methods should safely handle all exceptions e.g.:

```
Length xPos = FormatTools.getPhysicalSizeX(DataTools.parseDouble(value), UNITS.REFERENCEFRAME)
```

sbesson:
@dgault: apologies, I realized these are positions we are parsing. In that case, `FormatTools.getPhysicalSize()` is too restrictive as it explicitly checks for positive floats as opposed to floats. I have a staging branch adding `FormatTools.getPosition()` helper methods which I can open a PR against your branch if you want. Alternatively, we can restrict the scope of this PR to fixing the reported `NumberFormatException` issue and deal with the other number parsing elsewhere. Let me know what you think is the easiest.

dgault:
@sbesson If there is no immediate rush to have this PR in then it might be as well waiting and fixing the parsing properly with the new helper methods.

sbesson:
No immediate rush from my point of view. Opened #2448 to add the helper methods discussed above.

gatagat:
Can this PR be merged? Can we somehow help it?

Without this PR we are not able to read our data and it is kind of frustrating that the fix is ready for months, but still not available in fiji.

sbesson:
Apologies for the delay @gatagat, we have working on the final stages of the development of Bio-Formats 5.2.0 since the beginning of May and we froze our code changes early August in preparation of the release last week. Unfortunately, we did not get back to this PR in time for it to be included in in the 5.2.0 release.

Now Bio-Formats 5.2.0 is out and shipped via Fiji, we should be able to deliver releases of Bio-Formats with bug fix/regression at a faster pace. As you rightly noticed, this PR is almost ready barring usage of the new API and validation by the automated tests so we will make sure it gets included in Bio-Format 5.2.1 due in the upcoming weeks.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/285/ for the passing repository tests. Merging for 5.2.1.

PR #2438, opened 2016-05-26
OME-TIFF: fix NPE when closing reader for partial multi-file filesets

sbesson:
See https://trac.openmicroscopy.org/ome/ticket/13107

At the Bio-Formats command-line tools level, this PR adds a `reader.close()` call if a `FormatException` is thrown by `reader.setId()` reproducing the NPE described in the ticket. In addition these exceptions are not handled more gracefully using logging and should return a non-zero return code.

This PR should be tested at using either one of the QA files linked to the ticket or any of the [multi-file OME-TIFF filesets](https://www.openmicroscopy.org/site/support/ome-model/ome-tiff/data.html#multi-file-ome-tiff-filesets) after deleting one of the TIFF files:
- at the Bio-Formats level, `showinf /path/to/broken_image.ome.tiff` should print an error statement and the full stack trace should be shown when using `showinf -debug /path/to/broken_image.ome.tiff`
- at the OMERO level, importing the partial fileset should still fail but the error should not longer be a NPE but instead a message about the nature of the error.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -1034,7 +1034,12 @@ public class ImageInfo {
 
     // initialize reader
     long s = System.currentTimeMillis();
-    reader.setId(id);
+    try {
+      reader.setId(id);
+    } catch (FormatException exc) {
+      reader.close();
+      throw(exc);
+    }
     long e = System.currentTimeMillis();
     float sec = (e - s) / 1000f;
     LOGGER.info("Initialization took {}s", sec);
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -1038,7 +1038,9 @@ public class ImageInfo {
       reader.setId(id);
     } catch (FormatException exc) {
       reader.close();
-      throw(exc);
+      LOGGER.error("Failure during the reader initialization");
+      LOGGER.debug("", exc);
+      return false;
     }
     long e = System.currentTimeMillis();
     float sec = (e - s) / 1000f;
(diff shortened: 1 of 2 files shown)


rleigh-dundee:
The changes look sensible.

However, I'm afraid I've so far been unable to reproduce the original problem with the QA or sample datasets with develop without your PR merged in, so can't verify it solves the original problem.

sbesson:
@rleigh-dundee: did you try importing a partial dataset onto OMERO as described in the ticket ?

rleigh-dundee:
@sbesson Not yet, so far I was testing with bioformats alone.  But without being able to reproduce the original problem.

sbesson:
Understood. https://github.com/openmicroscopy/bioformats/pull/2438/commits/64ebf58d0d7896e0a7cbcd687b131cd8da2a8972 allows to reproduce the NPE on Bio-Formats alone. I can look into splitting this commit into a different PR if that helps the testing.

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
```diff
@@ -1034,7 +1034,14 @@ public boolean testRead(String[] args)
 
     // initialize reader
     long s = System.currentTimeMillis();
-    reader.setId(id);
+    try {
+      reader.setId(id);
+    } catch (FormatException exc) {
+      reader.close();
```
joshmoore: No objections to this being here, but we should probably make it a bug for any reader to throw from `setId` **without** closing the underlying files. Not sure if FindBugs could help us with that.

sbesson: Agreed. The most appropriate proposal I can come with might be to migrate this `try/catch` behavior into `FormatReader.setId(String)`. As there might be caveats we want to discuss with /cc @melissalinkert , I propose to push this change out of the PR and start by fixing the NPE.

joshmoore: I'd propose leaving the change (though you could catch most exceptions and call `close()` I'd think), but that we also look at the wider proposal as well.

bramalingam:
Both command-line and Omero import workflows, work as expected. Looks good to merge.

sbesson:
Captured discussion on auto-close as https://trello.com/c/AZNC7S6V/152-formatreader-auto-close-file-on-exception.

PR #2442, opened 2016-06-01
cpp: Upgrade to 2016-DEV0 model

rleigh-dundee:
Add units, quantities, map annotations, folders to the generator, templates etc.
Also includes a number of minor bugfixes.

Note this is based upon https://github.com/openmicroscopy/bioformats/pull/2427.

Source changes: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-generated-sources/250/

I wonder if this PR breaks any records, at a total 58058 lines of changes in 7919 static and 50141 generated (22856 model, 35202 metadata).  While generated from just 4901 lines of already overly verbose XML schema, the information density in this code is very low.


rleigh-dundee:
Green in https://ci.openmicroscopy.org/view/Files-DEV/job/OME-FILES-CPP-DEV-breaking/103/

rleigh-dundee:
Note: Rebased onto the latest https://github.com/openmicroscopy/bioformats/pull/2427 to make the code generation look sensible, as well as the history.

components/formats-gpl/utils/writeMapAnnotationsExample.java
```diff
@@ -39,6 +39,7 @@
 import loci.common.services.ServiceFactory;
 import loci.formats.meta.IMetadata;
 
+import ome.xml.model.*;
```
dgault: The import model.* could be replaced with import ome.xml.model.MapPair;

dgault:
I know they were commits that I mad, but just looking at the removal of abstractproprietry from the code gen and wanting to clarify, are we sure that it is ok if no longer support the building of previous model versions? (Of course we still support the import of older model files) Is there any reason we would ever want to build those older model objects from the latest release?

components/xsd-fu/python/ome/modeltools/property.py
```diff
@@ -415,6 +432,17 @@ def _get_isReference(self):
...
+        returnList = []
+        if self.model.opts.lang.hasSubstitutionGroup(self.name):
+            for o in sorted(self.model.objects.values(), lambda x, y: cmp(x.name, y.name)):
+                if o.parentName == self.name:
+                    returnList.append(o.name)
+        return returnList
+    concreteClasses = property(
+        _get_concreteClasses,
+        doc="""Concrete instance types for an abstract type.""")
+
```
dgault: The get_concreteClasses function must have snuck in with one of my commits to revert to an older schema, it isnt being used and can be removed

components/xsd-fu/templates-cpp/DummyMetadata.template
```diff
@@ -466,54 +480,28 @@ namespace ome
...
+        throw MetadataException("DummyMetadata", "get${abstractClass}Type",
                                 "intentionally not implemented");
       }
 {% end source %}\
 
 {% if fu.SOURCE_TYPE == "header" %}\
         // Documented in base class.
         index_type
-        get${abstractClass}Count(${indexes_string(v)}) const;
+        get${abstractClass}Count(${indexes_string(v[:-1])}) const;
```
dgault: Should this be indexes_string_dummy also?

components/xsd-fu/templates-cpp/OMEXMLMetadata.template
```diff
@@ -1,4 +1,5 @@
 {% def indexes_string(indexes) %}${", ".join(map(lambda x: lang.index_string(x),indexes))}{% end %}\
+{% def indexes_string_dummy(indexes) %}${", ".join(map(lambda x: lang.index_string(x, dummy=True),indexes))}{% end %}\
```
dgault: As the indexes_string_dummy is not being used in this template any more (had previously been in by mistake by the looks of it), the definition can probably be removed

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
```diff
@@ -1,12 +1,3 @@
-{% python
-  def getConcreteClasses(abstractClass):
-    returnList = []
-    if model.opts.lang.hasSubstitutionGroup(abstractClass):
-      for o in sorted(model.objects.values(), lambda x, y: cmp(x.name, y.name)):
-          if o.parentName == abstractClass:
-            returnList.append(o.name)  
-    return returnList
```
dgault: I had made a previous comment about removing getConcreteClasses from property, but now I see you've moved it to there from the template. In which case both changes are fine to leave as they are

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
```diff
@@ -396,7 +391,7 @@ ${customContent}\
 {% if klass.langBaseType == 'std::string' %}\
           this->impl->value = text;
 {% end %}\
-        }
+        } //a
```
dgault: Rouge comment snuck in

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
```diff
@@ -480,11 +510,12 @@ ${customUpdatePropertyContent[prop.name]}
 {% if prop.instanceVariableType == 'std::string' %}\
               set${prop.methodName}(element.getAttribute("${prop.name}"));
 {% end %}\
-            }
+            } // b
+{% end choose %}\
 {% end %}\
-          }
+          } // c
```
dgault: 2 more 'b' and 'c' comments

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
```diff
@@ -994,7 +1071,7 @@ ${customUpdatePropertyContent[prop.name]}
       }
 {% end source %}\
 {% end %}\
-{% when prop.maxOccurs > 1 and not klass.isAbstract and not klass.isAbstractProprietary %}
+{% when prop.maxOccurs > 1 and not klass.isAbstract %}\
```
dgault: In the cpp generated model code there are a number of places where new lines have disappeared. 
I cant see anywhere particular in the template that removes them but perhaps adding the forward slash (correctly) onto these template conditionals means that the set functions above for example no longer have any spacing between them.

dgault:
In the cpp model generated code there are a number of empty conditions such as in update functions:
if (element.hasAttribute("ExposureTimeUnit"))
{         
}

Do we need extra code to deal with the units attributes in these cases?

dgault:
Finally finished reviewing all the changes here. I've covered all of the code changes, with a number of comments as above. I've also reviewed the generated source changes for both the model and metadata on C++ and Java.

The important thing here is that everything builds and all the tests are good on both platforms. The metadata API changes also appear to be as expected. As the C++ API is moving up 2 schemas the changes here are significant but there is nothing that raises concern from my side. That being said with such significant API updates it would always be good to have a few more eyes scan the cpp metadata diffs.

The only real potential discussion point is the previously mentioned question about supporting the building of older schemas from the latest code base.

components/formats-gpl/utils/writeMapAnnotationsExample.java
rleigh-dundee: Should be fixed now.

rleigh-dundee:
@dgault 

```
if (element.hasAttribute("ExposureTimeUnit"))
```

These should now be dropped (not generated).

Regarding the 2013-06 model, I don't think we will want to ever generate it again.  The intent is to only support 2016 here; 2015 was previously the goal but since it required schema changes it wasn't possible to achieve that, so we switched to 2016 instead.  I've cleaned up some of the 2013 support in the templates; if any other support is lingering which wasn't reverted by your changes or mine, it can be deleted as well.

I'm still working though the other comments you made; hopefully I'll get to them early tomorrow.

rleigh-dundee:
The latest commit fixes the problem noted by @sbesson in https://github.com/openmicroscopy/bioformats/pull/2427#issuecomment-223927569

components/xsd-fu/templates-cpp/DummyMetadata.template
rleigh-dundee: Never for headers, only for sources where we don't use the arguments.  The sole reason for doing this is to avoid compiler warnings about unused arguments.  They all look consistent from that point of view.

components/xsd-fu/templates-cpp/OMEXMLMetadata.template
rleigh-dundee: Now dropped, thanks.

components/xsd-fu/templates-cpp/OMEXMLModelObject.template
rleigh-dundee: I'll look over this in more detail tomorrow.  As a general rule, every conditional uses `%}\` to end every line to avoid any trailing characters contributing to the output, including newlines, so that every line in the output is explicit and obvious.  This was due to some subtle problems in the past.

rleigh-dundee:
That's it for template tidying for this PR I think.  There are some whitespace issues in the metadata templates, but it's hard to fix when the total generated code delta is ~95klines, 60klines of which are metadata code.  It would probably be more prudent to fix as a followup where the diffs would be readable.

I think I've addressed all the points raised so far, unless I missed something.  Any further feedback welcome.

dgault:
The updated changes look good to me. I agree that a separate PR to fully tidy up and clean whitespace issues in the templates is probably the best approach.

sbesson:
All CI jobs have been green in the last few days and successfully test our set of OME-XML samples including the new ones added in https://github.com/openmicroscopy/bioformats/pull/2447. I can confirm the Java generated sources diff is now fixed and nothing in the large C++ generated sources diff raised immediate concerns as well. Merging this to bring  C++ support for the new model. Congratulations to everyone involved in the last few months especially @rleigh-dundee and @dgault.

Do we have a follow-up card to tidy up the whitespaces in the templates? Actually on the Java side, we might use this opportunity to comply with the indentation style by using spaces instead of tabs.

PR #2445, opened 2016-06-07
FakeReader: add methods to read special pixels

simleo:
Adds methods to read out the metadata info (S, No, Z, C, T) that `FakeReader` encodes in each plane. This also serves as a source of documentation for the feature.


joshmoore:
Nice.

sbesson:
All looks good to me. The Javadoc is clear and the addition of `public static` method is certainly sensible and reusable. In the long-run, if we start expanding FakeReader to support other types of pixel data than hardcoded gradients, we might considering deriving a companion `writeSpecialPixels()`.

Is it worth mentioning the existence of these special pixels in the corresponding Sphinx documentation page?

simleo:
> Is it worth mentioning the existence of these special pixels in the corresponding Sphinx documentation page?

I'm not sure, they are only useful for testing purposes. Moreover, I think we should make them optional in the future (see Trello card)

sbesson:
Makes sense. Merging for now.

PR #2446, opened 2016-06-08
Formats-bsd: unit tests performance

sbesson:
See https://trello.com/c/ISuLct1l/125-formats-bsd-unit-tests-take-too-long-to-run

This PR proposes to reduce the time spent in the default Formats BSD unit tests with the main bottleneck being the `SixteenBitLosslessJPEG2000Test` class which loops over every short value from -32,768 to +32,767.

#### Test classes

- add an increment field in `SixteenBitLosslessJPEG2000Test` controlling the loop of `testLosslessPixels`
- set this increment to `1024` by default
- define another test class and use the TestNG Factory annotation to set the increment to `1` and test all short values

#### Build infrastructure

- leave `testng.xml` consumed by the default Maven `test` target unchanged
- define a new `testng-long-running.xml` configuration file and `test-long-running` Ant target allowing to run the full test using the entire Short range.

### Testing instructions
- check Travis is green with this PR included and the time spent in the Maven matrix component is reduced and/or check that `mvn test` completes faster
- check that `ant test` runs the `SixteenBitLosslessJPEG2000LongRunningTest` and tests all short values

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000LongRunningTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000LongRunningTest.java
@@ -42,7 +42,7 @@ public class SixteenBitLosslessJPEG2000LongRunningTest {
 
   @Factory
   public Object[] factoryMethod() {
-    return new Object[] { new SixteenBitLosslessJPEG2000Test(1)};
+    return new Object[] {new SixteenBitLosslessJPEG2000Test(1)};
   }
 
 }
--- a/components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000LongRunningTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000LongRunningTest.java
@@ -1,0 +1,48 @@
+/*
+ * #%L
+ * BSD implementations of Bio-Formats readers and writers
+ * %%
+ * Copyright (C) 2016 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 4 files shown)


components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000LongRunningTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Long-running test case which outlines the problems seen in    
+ * trac.openmicroscopy.org/ome/ticket/6278.
+ */
+public class SixteenBitLosslessJPEG2000LongRunningTest {
+
+  @Factory
+  public Object[] factoryMethod() {
+    return new Object[] { new SixteenBitLosslessJPEG2000Test(1)};
```
mtbc: spacing

components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000Test.java
```diff
@@ -66,12 +66,21 @@
   private static final Logger LOGGER =
     LoggerFactory.getLogger(SixteenBitLosslessJPEG2000Test.class);
 
+  private int increment = 1024;
+
+  public SixteenBitLosslessJPEG2000Test() {
+  }
+
+  public SixteenBitLosslessJPEG2000Test(int increment) {
+      this.increment=increment;
```
mtbc: spaces around =

components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000Test.java
```diff
@@ -66,12 +66,21 @@
   private static final Logger LOGGER =
     LoggerFactory.getLogger(SixteenBitLosslessJPEG2000Test.class);
 
+  private int increment = 1024;
+
+  public SixteenBitLosslessJPEG2000Test() {
+  }
```
mtbc: a more idiomatic pattern might be to have `private final int increment;` and in this constructor do `this(1024);` but this approach is fine too

mtbc:
`ant test`: Without this PR I have 943 Formats-BSD unit tests running in 3m34 and with this PR I have 943 tests running in 3m56.
`mvn test`: Without this PR it runs in 3m37 and with this PR it runs in 2m31 where for both `SixteenBitLosslessJPEG2000Test` is green in `formats-bsd/target/surefire-reports/`.

The above seem typical over multiple runs. Slightly slower `ant test` seems worth rather faster `mvn test`. This is looking good.

sbesson:
Thanks for the testing @mtbc. Your last metrics raises the question of whether the 1min increase for the default Maven unit tests is significant enough or if we need to spend more time identifying the bottlenecks /cc @simleo

components/formats-bsd/test/loci/formats/utests/SixteenBitLosslessJPEG2000Test.java
simleo: You mean `this(increment)`?

mtbc: No, as how the no-arg constructor could specify 1024 default. (Fewer state changes = easier reasoning.)

mtbc: (The pattern's perhaps a bit like `doChange` or `newUserInGroup` in `AbstractServerTest` where we chain the methods to add arguments to cover for our lack of default values in Java, yet another thing the language could have learned from Modula-3.)

simleo:
On my laptop, the running time for the whole Maven test suite is 3:49 without and 2:19 with the patch. The BSD section goes down to 1:05 from 2:37, so it's more than twice as fast. The BSD suite is still the longest running at 1:05, but `SixteenBitLosslessJPEG2000Test` was the main bottleneck and this PR removes it. We should merge, then we can look for the next big offender.

sbesson:
@mtbc @simleo: pushed

mtbc:
Lovely, thank you. :+1:

rleigh-dundee:
This looks fine and good to merge.

The only question I have WRT travis is whether we want to be able to selectively enable these long running tests for a given matrix combination e.g. jdk7.  If ant is going to be dropped in the future, then I think it would be nice to have the option to run the long running tests with maven.  Maybe this is already possible?

sbesson:
Thanks, merging this. Re https://github.com/openmicroscopy/bioformats/pull/2446#issuecomment-225135922, one of the major scoping tasks associated with the Ant -> Maven transition proposal is certainly to review and list all the tasks which are specific to one build system.

PR #2448, opened 2016-06-09
Add helper methods to FormatTools for creating stage position lengths

sbesson:
See discussion in https://github.com/openmicroscopy/bioformats/pull/2437 for more context. Similarly to the existing `FormatTools.getPhysicalSize()` methods, this PR adds a set of three methods to the `FormatTools` class allowing to format a stage position from an input `Double` and an input unit:
- `FormatTools.getStagePosition(Double, String)`
- `FormatTools.getStagePosition(Double, Unit<Length>)`

The use case for these methods is the parsing of string values across the readers in a safe manner. In particular, combined with `DataTools` utility methods, the following forms:

```
Length pos = FormatTools.getStagePosition(DataTools.parseDouble(value), unit);
```

should allow to handle most invalid cases and prevent the usage of `try/catch` statement.

Unit tests have been added to `FormatToolsTest` to cover the semantics of the new methods. In particular it should be tested that the array of objects in the data providers cover most use cases.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1437,7 +1437,57 @@ public final class FormatTools {
     }
     return new Time(value, UNITS.SECOND);
   }
-  
+
+  /**
+   * Formats the input value for the position into a length.
+   *
+   * @param value  the value of the position
+   * @param unit   the unit of the position. If {@code null} or invalid,
+   *               default to reference frame.
+   *
+   * @return       the physical size formatted as a {@link Length}
+   */
+  public static Length getPosition(Double value, String unit) {
+    if (value == null || value == Double.POSITIVE_INFINITY ||
+        value == -Double.POSITIVE_INFINITY) {
+      LOGGER.debug("Expected float value for position; got {}", value);
+      return null;
+    }
+
+    if (unit == null) return new Length(value, UNITS.REFERENCEFRAME);
+
+    try {
+      return UnitsLength.create(value, UnitsLength.fromString(unit));
+    } catch (EnumerationException e) {
+      LOGGER.debug(e.getMessage());
+      return new Length(value, UNITS.REFERENCEFRAME);
+    }
+  }
+
+  /**
+   * Formats the input value for the position into a length.
+   *
+   * @param value  the value of the position
+   *
+   * @return       the physical size formatted as a {@link Length}
+   */
+  public static Length getPosition(Double value) {
+   return getPosition(value, UNITS.REFERENCEFRAME);
+  }
+
+  /**
+   * Formats the input value for the position into a length of the given unit.
+   *
+   * @param value  the value of the position
+   * @param unit   the unit of the position. If {@code null} or invalid,
+   *               default to reference frame.
+   *
+   * @return       the position formatted as a {@link Length}
+   */
+  public static Length getPosition(Double value, Unit<Length> unit) {
+      return getPosition(value, unit.getSymbol());
+  }
+
   public static Length getPhysicalSize(Double value, String unit) {
     if (value != null && value != 0 && value < Double.POSITIVE_INFINITY) {
       if (unit != null) {
@@ -1554,7 +1604,7 @@ public final class FormatTools {
 
   /**
    * Formats the input value for the physical size in Z into a length in
-   * microns
+   * microns.
    *
    * @param value  the value of the physical size in Z in microns
    *
@@ -1586,7 +1636,6 @@ public final class FormatTools {
    * @param unit   the unit of the physical size in Z
    *
    * @return       the physical size formatted as a {@link Length}
-
    */
   public static Length getPhysicalSizeZ(Double value, Unit<Length> unit) {
       return getPhysicalSize(value, unit.getSymbol());
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1441,12 +1441,15 @@ public final class FormatTools {
 
 
   /**
-   * Formats the input value for the position into a length of the given unit.
+   * Formats the input value for the stage position into a length of the given
+   * unit.
    *
-   * @param value  the value of the position
-   * @param unit   the unit of the position
+   * @param value  the value of the stage position
+   * @param unit   the unit of the stage position
    *
-   * @return       the position formatted as a {@link Length}
+   * @return       the stage position formatted as a {@link Length}. Returns
+   *               {@code null} if {@code value} is {@code null} or infinite
+   *               or {@code unit} is {@code null}.
    */
   public static Length getStagePosition(Double value, Unit<Length> unit) {
     if (value == null || value.isInfinite()) {
@@ -1466,10 +1469,14 @@ public final class FormatTools {
    * Formats the input value for the stage position into a length of the given
    * unit.
    *
-   * @param value  the value of the position
-   * @param unit   the unit of the position
+   * @param value  the value of the stage position
+   * @param unit   the unit of the stage position. If the string cannot be
+   *               converted into a base length unit, the stage position length
+   *               will be constructure using the default reference frame unit.
    *
-   * @return       the physical size formatted as a {@link Length}
+   * @return       the stage position formatted as a {@link Length}. Returns
+   *               {@code null} under the same conditions as
+   *               {@link #getStagePosition(Double, String)}.
    */
   public static Length getStagePosition(Double value, String unit) {
       Unit<Length> baseunit = null;
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1444,52 +1444,42 @@ public final class FormatTools {
    * Formats the input value for the position into a length of the given unit.
    *
    * @param value  the value of the position
-   * @param unit   the unit of the position. If {@code null} or invalid,
-   *               default to reference frame.
+   * @param unit   the unit of the position
    *
    * @return       the position formatted as a {@link Length}
    */
-  public static Length getPosition(Double value, Unit<Length> unit) {
+  public static Length getStagePosition(Double value, Unit<Length> unit) {
     if (value == null || value.isInfinite()) {
-      LOGGER.debug("Expected float value for position; got {}", value);
+      LOGGER.debug("Expected float value for stage position; got {}", value);
       return null;
     }
 
-    if (unit == null) unit = UNITS.REFERENCEFRAME;
+    if (unit == null) {
+      LOGGER.debug("Expected valid unit for stage position; got {}", unit);
+      return null;
+    }
 
     return new Length(value, unit);
   }
 
-
   /**
-   * Formats the input value for the position into a length of the given unit
+   * Formats the input value for the stage position into a length of the given
+   * unit.
    *
    * @param value  the value of the position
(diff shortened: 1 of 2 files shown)


components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1437,7 +1437,57 @@ public static Time getTime(Double value, String unit) {
...
+   * Formats the input value for the position into a length of the given unit.
+   *
+   * @param value  the value of the position
+   * @param unit   the unit of the position. If {@code null} or invalid,
+   *               default to reference frame.
+   *
+   * @return       the position formatted as a {@link Length}
+   */
+  public static Length getPosition(Double value, Unit<Length> unit) {
+      return getPosition(value, unit.getSymbol());
```
joshmoore: Someone may have used `UnitsLength.fromString(unit)` for type safety to call this method. I'd suggest making this method the definitive one, and `(Double, String)` makes use of it.

sbesson: I might need a units refresher here since `UnitsLength.fromString(String)` will return `UnitsLength` and I am looking for a way to come back to `Unit<Length>`. Or should we support both signatures? In many ways, this comes back to our need to document the recommendations for using [the unit system](https://trello.com/c/FvyQj7OP/144-document-unit-changes)

melissalinkert: ```getBaseUnit``` in ```ome.xml.model.enum.handlers.UnitsLengthEnumHandler``` should convert a ```UnitsLength``` into ```Unit<Length>```.  Supporting both might make sense - ```UnitsLength``` is easier to work with if you're parsing a unit from a string, but ```Unit<Length>``` is a lot easier if you know definitively which unit to use.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1437,7 +1438,61 @@ public static Time getTime(Double value, String unit) {
...
+   * Formats the input value for the position into a length of the given unit.
+   *
+   * @param value  the value of the position
+   * @param unit   the unit of the position. If {@code null} or invalid,
+   *               default to reference frame.
+   *
+   * @return       the position formatted as a {@link Length}
+   */
+  public static Length getPosition(Double value, Unit<Length> unit) {
+    if (value == null || value == Double.POSITIVE_INFINITY ||
```
mtbc: value.isInfinite()

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1437,7 +1438,60 @@ public static Time getTime(Double value, String unit) {
...
+   *
+   * @return       the position formatted as a {@link Length}
+   */
+  public static Length getPosition(Double value, Unit<Length> unit) {
+    if (value == null || value.isInfinite()) {
+      LOGGER.debug("Expected float value for position; got {}", value);
+      return null;
+    }
+
+    if (unit == null) unit = UNITS.REFERENCEFRAME;
```
rleigh-dundee: Why the hardcoding of reference frame as a fallback?  This is a situation where it would be appropriate to throw an exception.  Note that for every model property, we *know* the default unit type and it's reasonable to require a valid unit here.

As an example, this is the C++ `update()` logic for `Pixels/PhysicalSizeY`

```
        if (element.hasAttribute("PhysicalSizeY"))
          {
            // Attribute property PhysicalSizeY with unit companion PhysicalSizeYUnit
            std::string unitSymbol("µm");
            if (element.hasAttribute("PhysicalSizeYUnit"))
              {
                unitSymbol = element.getAttribute("PhysicalSizeYUnit");
                if(unitSymbol.empty())
                  unitSymbol = "µm";
              }

            set_quantity(element.getAttribute("PhysicalSizeY"), unitSymbol,
                         *this, &Pixels::setPhysicalSizeY,
                         "Pixels", "PhysicalSizeY");
          }
```

Here we default to µm if unspecified in the input.  But for any *API* called by a *reader*, the reader itself must already know the unit system it's dealing with.  If it doesn't this it must itself pass in reference frame as the unit type.  I don't think this type of hardcoded fallback is appropriate.  This is the reader's responsibility.

components/formats-api/test/loci/formats/utests/FormatToolsTest.java
```diff
@@ -49,58 +49,26 @@
...
+      {FormatTools.FLOAT, -2147483648L, 2147483647L},
+      {FormatTools.DOUBLE, -2147483648L, 2147483647L},
+    };
   }
 
-  public void testDefaultMinMaxDouble() throws IllegalArgumentException {
-    long[] lim = FormatTools.defaultMinMax(FormatTools.DOUBLE);
-    assertEquals(lim[0], -2147483648);
-    assertEquals(lim[1], 2147483647);
+  @Test(dataProvider = "pixelTypes")
```
rleigh-dundee: While this clean up of the PixelType min/max range checks is fine, it's unrelated to length unit helpers and I think should be in a separate PR.

sbesson: Makes sense. I'll push the commit away.

components/formats-api/src/loci/formats/FormatTools.java
sbesson: Explicitly passing a valid unit type is certainly what we have enforced in our readers so far.
Thus the proposal of handling  `null`/invalid units in the same way we handle `null`/invalid values is certainly sensible to me. I will adjust the method/tests accordingly.

What are the implications for default methods like `getPosition(Double)`? Is there a need to define such unit-less methods (and use the default unit specified in the model) for other consumers of the API than our own readers?

rleigh-dundee: I don't think `getPosition(Double)` makes sense to have.  I think we should require the caller to specify the unit in all cases, even when that unit is "reference frame" so as to be completely unambiguous.

In the case that the value *or* the unit are `null`, I think we should be simply returning `null` or throwing an exception here, so that nulls are propagated through or are required to be handled.

The main benefit I see in this PR is the ability to create a `Length` using either the enum value or a string.  But I wonder if this couldn't be added at a lower level--as a factory method for the ome-xml Quantity class, for example, to make use of it simple and direct.  And it would also make it work with all unit types rather than just lengths.

sbesson: There is another benefit with these methods on top of creating `Length` objects which is the value validation e.g. floats for position or positive floats for physical sizes in agreement with the model. I also agree this is something there is some generic content we should be able to migrate upstream to the `ome-xml` component eventually. 

For now, the  goal is to reduce the logic  implemented in every reader in favor of a more generic :

```
Length posx = FormatTools.getPosition(DataTools.parseDouble(value), UNITS.REFERENCE_FRAME);
```

rleigh-dundee:
A few additional points:

- Why `getPosition` and not `getLength`?  This is usable by readers in contexts where it's not a position; I would count physical sizes in that category.
- I think that the method name could maybe indicate its function a bit better, maybe getXXXLength where XXX indicates what it's actually doing here.
- Irrespective of the other comments, I'd like `getPosition(Double value)` to be dropped; I think we should keep things simple and require readers to explicitly state their intentions rather than rely on "magic" defaults as side-effects which can result in implicit assumptions pervading the codebase.

sbesson:
After discussion with @rleigh-dundee and @joshmoore, pushed additional commit addressing most of the critical comments and moved the unit tests optimization  to https://github.com/openmicroscopy/bioformats/pull/2452. @rleigh-dundee will take a look at the quantity specialization with the `ome.xml` primitives e.g. in the case of PhysicalSize.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1437,7 +1438,50 @@ public static Time getTime(Double value, String unit) {
...
-  
+
+
+  /**
+   * Formats the input value for the position into a length of the given unit.
+   *
+   * @param value  the value of the position
+   * @param unit   the unit of the position
+   *
+   * @return       the position formatted as a {@link Length}
```
joshmoore: "or null if the value is infinite or the unit null"

sbesson: :+1: I will update the Javadocs in a final commit once we have all signed off on the expected behavior.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1437,7 +1438,50 @@ public static Time getTime(Double value, String unit) {
...
+   *
+   * @return       the physical size formatted as a {@link Length}
+   */
+  public static Length getStagePosition(Double value, String unit) {
+      Unit<Length> baseunit = null;
+      try {
+        baseunit = UnitsLengthEnumHandler.getBaseUnit(
+          UnitsLength.fromString(unit));
+      } catch (EnumerationException e) {
+        LOGGER.debug(e.getMessage());
```
rleigh-dundee: Do we want to set `baseunit` to reference frame here to at least preserve the value?  Or discard everything if the unit is invalid?

rleigh-dundee:
All looks good.  Just one outstanding question about the error behaviour for `Length getStagePosition(Double value, String unit)`.

components/formats-api/src/loci/formats/FormatTools.java
sbesson: The idea of the last commit was to mirror the behavior of `getStagePosition(Double, Unit<Length>)` where invalid units return `null`. Looking at it again, in the case of an `EnumerationException`, we might want to `return null` immediately after logging the exception.
Adding extra granularity to choose whether or not to fallback to the default unit is possible but I it would be good to clarify of the use cases before adding the API.

sbesson:
Coming back to this and answering https://github.com/openmicroscopy/bioformats/pull/2448#issuecomment-225909006, the last few commits now align the behavior with the other `FormatTools.getXXX(Double, String)` methods especially with regard to invalid base units. Unit tests and Javadocs have been updated accordingly.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1437,7 +1438,59 @@ public static Time getTime(Double value, String unit) {
...
+   *
+   * @param value  the value of the stage position
+   * @param unit   the unit of the stage position
+   *
+   * @return       the stage position formatted as a {@link Length}. Returns
+   *               {@code null} if {@code value} is {@code null} or infinite
+   *               or {@code unit} is {@code null}.
+   */
+  public static Length getStagePosition(Double value, Unit<Length> unit) {
+    if (value == null || value.isInfinite()) {
```
mtbc: interesting we object to infinity but not NaN

rleigh-dundee: Yes.  I'm not sure why we object to infinity in the first place--it's a valid value which might legitimately occur.

rleigh-dundee:
The latest changes all look fine to me.  There's just the Inf/NaN policy which could be cleared up.

sbesson:
Reading other `FormatTools` helper methods, they are typically enforcing  constraints on individual values stricter than the ones defined in the schema (non-infinite, non-Nan, minimum epsilon values...). My understanding of these bans is that they are driven by the reality of the files we are dealing with where legitimate from a schema but non-physical values are read that we do not want to store in the metadata (or maintain in automated tests configuration files). Is that an accurate representation @melissalinkert?

Based on the comments above, I propose to include these new APIs for the 5.2.0-m5 milestone and adjust their implementation later on to add non-NaN checks.

PR #2450, opened 2016-06-10
xsd-fu: Clean up Java templates

rleigh-dundee:
Remove tabs and trailing whitespace.  No functional changes.  This is a followup to the recent schema code generation work.

https://github.com/openmicroscopy/bioformats/pull/2450/files?w=1 shows the non-whitespace changes (i.e. none).

--------

Testing: Check jobs are green; review source changes.

### DIFF
--- a/components/xsd-fu/templates-java/AggregateMetadata.template
+++ b/components/xsd-fu/templates-java/AggregateMetadata.template
@@ -307,7 +307,7 @@ public class AggregateMetadata implements IMetadata
 
 
         // -- Entity counting (manual definitions) --
- 
+
   public int getBooleanAnnotationAnnotationCount(int booleanAnnotationIndex) {
     for (Iterator iter = delegates.iterator(); iter.hasNext();) {
       Object o = iter.next();
@@ -470,8 +470,8 @@ public class AggregateMetadata implements IMetadata
       }
     }
     return null;
-  }    
-    
+  }
+
   public int get{% if is_multi_path[abstractClass] %}${k}{% end %}${abstractClass}Count(${indexes_string(v[:-1])})
   {
     for (Iterator iter = delegates.iterator(); iter.hasNext();) {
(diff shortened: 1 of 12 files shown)


sbesson:
Thanks @rleigh-dundee. As we are tackling the whitespace, what's the cost of unifying the Java templates to use the 2 space indentation used for the Java codebase in Bio-Formats? `AggregateMetadata.template` for instance has a mixture of 2 and 4 space indentations.

rleigh-dundee:
@sbesson The main cost is the tedious editing of the templates to get the indentation levels correct.  It's potentially quite fiddly.  I did it for the C++ templates when I originally wrote them, and we can certainly do this for Java, but it's not possible to automatically do due to the conditionals in the templates--you can only easily re-indent a block between conditionals.  Certainly do-able, but it's probably a couple of days work while this PR was 5 minutes with `M-x untabify` in Emacs.

sbesson:
Despite the large diff generated in https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-generated-sources/255/, the non-whitespace diff is indeed null:

```
$ git diff -w snoopycrimecop/develop/merge/generated_sources snoopycrimecop/develop/latest/generated_sources  -- components/ome-xml/
$
```

On the https://github.com/openmicroscopy/bioformats/pull/2450#issuecomment-225460772 discussion, I certainly agree with regard to the tediousness of the task. Should we consider parallelizing the work and dividing the Java templates amongst ourselves as we have done for other cleanup epics?

sbesson:
@rleigh-dundee: if you haven't started, I will open a PR against your branch reducing the indentation to 2.

rleigh-dundee:
@sbesson Why don't we just merge this PR as is and do that separately?

sbesson:
Makes sense. Merging, then.

rleigh-dundee:
@sbesson OK, thanks.

I've not started work on the indentation yet, by the way.  When I did this for C++, I did it by reindenting the generated code and then adjusting the templates until the automatic reindentation resulted in zero changes.

PR #2451, opened 2016-06-13
Command-line tools: common methods refactoring

sbesson:
See https://trac.openmicroscopy.org/ome/ticket/8260 and https://trac.openmicroscopy.org/ome/ticket/13226 

This PR initiates the refactoring of the common functions used across the various command-line interfaces into a central class under `formats-bsd`. Currently this refactoring is limited to the version printing and the upgrade checking but more could be migrated there in the future.

Summary of changes:
- add a `loci.formats.tools.CommandLineTools` class defining both command-line arguments and methods
- migrates arguments and code used in `showinf`, `bfconvert`, `tiffcomment` and `xmlvalid`
- decouples the version and upgrade check  logic by running the upgrade check *after* the version handling
- gracefully handles `UnknowHostException` in the `UpgradeChecker` allowing to prevent stack traces when using the tools offline - see https://trac.openmicroscopy.org/ome/ticket/13226

To test this PR:
- test the `showinf/bfconvert` functionalities are unchanged with a mixture of `-version/-no-upgrade` is unmodified
- test `showinf/bfconvert/tiffcomment/xmlvalid` offline and check a log message is sent at the `WARN` level instead of a stack trace during the upgrade check.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tools/CommandLineTools.java
+++ b/components/formats-bsd/src/loci/formats/tools/CommandLineTools.java
@@ -1,0 +1,55 @@
+/*
+ * #%L
+ * BSD implementations of Bio-Formats readers and writers
+ * %%
+ * Copyright (C) 2016 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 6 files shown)


dgault:
Code looks fine to me and testing shows that all tools behave as normal in online mode.
Showinf and bfconvert also behave as expected when using both the version and no-upgrade options.
When used offline no stack trace is shown and 'Failed to reach the update site' is displayed instead.

PR looks good from my perspective.

components/formats-bsd/src/loci/formats/tools/CommandLineTools.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * An utility class for various methods used in command-line tools
+ */
+public final class CommandLineTools {
+
+  // -- Constants --
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CommandLineTools.class);
+  public static final String VERSION = "-version";
+
```
mtbc: spacing here is a little odd

mtbc:
Overall looking good, a definite improvement.

sbesson:
One final thought here in terms of repository layout: would it make sense to migrate `TiffComment.java`, `XMLValidate.java`, the new `CommandLineTools.java` and other command-line tool classes under the `bio-formats-tools` component?

mtbc:
That sounds sensible to me.

PR #2453, opened 2016-06-13
Unify Java templates indentation to 2 spaces

sbesson:
Follow-up work of https://github.com/openmicroscopy/bioformats/pull/2450. This commit should update the style of the generated Java classes to match the rest of the code base.

See https://github.com/openmicroscopy/bioformats/pull/2453/files?w=1 for the diff without whitespace changes.

Check all Bio-Formats jobs are green and that the non-whitespace difference between the `develop/latest/generated_sources` and `develop/merge/generated_sources` branches of @snoopycrimecop's fork is empty.

### DIFF
--- a/components/xsd-fu/templates-java/AggregateMetadata.template
+++ b/components/xsd-fu/templates-java/AggregateMetadata.template
@@ -12,71 +12,71 @@
 {% def index_string(name) %}int ${index_name_string(name)}{% end %}\
 \
 {% def counter(parent, obj, indexes) %}\
-        public int get{% if is_multi_path[o.name] %}${parent}{% end %}${obj.name}Count(${indexes_string(indexes[:-1])})
-        {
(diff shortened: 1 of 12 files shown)


rleigh-dundee:
Looks good to merge to me.  Both the PR diff and the generated source changes diff are whitespace-only (the latter checked by hand).  The new indentation looks good to me, though I admit I haven't read all 163557 lines of the diff!

I tried to verify that the indentation style matched that specified in `config/BFSourceFormat.xml`.  It doesn't--the braces and newlines in arguments don't match; but I'm not sure most of the source code meets the latter requirement anyway, so I don't think this minor discrepancy should block merging.  That significantly more tedious cleanup could be done separately, if desired.

sbesson:
Barring vetoes, I propose to merge this PR and use follow-up PRs to perform additional stylistic as appropriate /cc @melissalinkert @dgault

PR #2454, opened 2016-06-14
xsd-fu: Specialise C++ Quantity types by both unit and value

rleigh-dundee:
Tweak the unit types inside the python generator.  This essentially changes `Quantity<Unit, double>` to `Quantity<Unit, Value>`.  Note `double` is the default type in the Quantity template and so not actually generated.  Example: `Quantity<Length>` becomes `Quantity<Length, NonNegativeFloat>`.  This involves no code changes--it's completely transparent from an end use point of view.  Unless you were using values out of range of course, in which case it will now correctly throw an exception!

Future work: A corresponding change to the Java code generator would be useful, if there's a way to do that with the existing Quantity and Unit generics.

--------

Testing: Check jobs are green.  Also check generated source changes.

### DIFF
--- a/components/xsd-fu/python/ome/modeltools/language.py
+++ b/components/xsd-fu/python/ome/modeltools/language.py
@@ -292,11 +292,11 @@ class Java(Language):
     def getDefaultModelBaseClass(self):
         return "AbstractOMEModelObject"
 
-    def typeToUnitsType(self, valueType):
-        return self.model_unit_map[valueType]
+    def typeToUnitsType(self, unitType):
+        return self.model_unit_map[unitType]
 
-    def typeToDefault(self, valueType):
-        return self.model_unit_default[valueType]
+    def typeToDefault(self, unitType):
+        return self.model_unit_default[unitType]
 
     def index_signature(self, name, max_occurs, level, dummy=False):
         """Makes a Java method signature dictionary from an index name."""
@@ -376,8 +376,11 @@ class CXX(Language):
     def getDefaultModelBaseClass(self):
         return "detail::OMEModelObject"
 
-    def typeToUnitsType(self, valueType):
-        return "%s::Quantity<%s > " % (self.omexml_model_quantity_package, valueType)
+    def typeToUnitsType(self, unitType, valueType=None):
+        if valueType is None:
+            return "%s::Quantity<%s > " % (self.omexml_model_quantity_package, unitType)
+        else:
+            return "%s::Quantity<%s, %s > " % (self.omexml_model_quantity_package, unitType, valueType)
 
     def index_signature(self, name, max_occurs, level, dummy=False):
         """Makes a C++ method signature dictionary from an index name."""
--- a/components/xsd-fu/python/ome/modeltools/property.py
+++ b/components/xsd-fu/python/ome/modeltools/property.py
@@ -208,8 +208,14 @@ class OMEModelProperty(OMEModelEntity):
         name = self.langType
         if isinstance(self.model.opts.lang, language.CXX):
             if self.hasUnitsCompanion:
-                name = self.model.opts.lang.typeToUnitsType(
-                    self.unitsCompanion.langTypeNS)
+                if name == "double":
+                    name = self.model.opts.lang.typeToUnitsType(
+                        self.unitsCompanion.langTypeNS)
+                else:
+                    name = self.model.opts.lang.typeToUnitsType(
+                        self.unitsCompanion.langTypeNS,
+                        "%s::primitives::%s"
+                        % (self.model.opts.lang.omexml_model_package, name))
             elif self.isEnumeration:
                 name = ("%s::enums::%s"
                         % (self.model.opts.lang.omexml_model_package, name))
@@ -231,8 +237,11 @@ class OMEModelProperty(OMEModelEntity):
         mstype = None
 
         if self.hasUnitsCompanion:
-            mstype = self.model.opts.lang.typeToUnitsType(
-                self.unitsCompanion.metadataStoreArgType)
+            if isinstance(self.model.opts.lang, language.Java):
+                mstype = self.model.opts.lang.typeToUnitsType(
+                    self.unitsCompanion.metadataStoreArgType)
+            elif isinstance(self.model.opts.lang, language.CXX):
+                mstype = self.langTypeNS
 
         if self.name == "Transform":
             if isinstance(self.model.opts.lang, language.Java):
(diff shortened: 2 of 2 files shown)


sbesson:
Two immediate questions:
- is there a failing test/sample that could be used to test the initial behavior and the improvement proposed in this PR? 
- is there a concrete implementation proposal for the Java counterpart of this PR? Or should this be more generally discussed as part of the Units overhaul?

rleigh-dundee:
For Java, nothing concrete at the moment, but if you look at e.g. quantity.Length:

```
public class Length extends Quantity implements Comparable<Length>
{
  private static final int SEED1 = 45;
  private static final int SEED2 = 56;
  Number value;
  Unit<ome.units.quantity.Length> unit;
```

Here, `Number` could be turned into a generic Value which could be `PositiveFloat` or whatever.  It might also be possible to replace all these classes with `Quantity<Unit, Value>` with a little refactoring.  Or keep them as empty classes with the specialised Quantity as the base.

With regard to tests, we don't have any failing examples at the moment.  Since it inherits the behaviour of the primitive value types, which *do* have comprehensive test coverage, I didn't think it would add much of value, but we can add a few test cases as extra checks if you prefer.

sbesson:
A couple of thoughts while reading your proposal:
- specializing `Quantity` and/or its subclasses in Java (e.g. `Length<PositiveFloat>`) would certainly be matching the changes proposed in this PR. My feeling is this would be a large breaking change for the entire Java API and should be discussed both at the Bio-Formats and OMERO level.
- changing the type of `value` to be more generic would likely have less breaking impact in terms of some of the API. In this case I assume we would still do the mapping of the expected type for each element attribute at the codegenerated level?

Semi-related, while looking at the Java templates, some portions of the generated Java code enforce the model constraints internally using the existing `Quantity` API while using enum handlers e.g. [UnitsLengthEnumHandler.getQuantity(T, UnitsLength)](http://downloads.openmicroscopy.org/bio-formats/5.1.10/api/ome/xml/model/enums/handlers/UnitsLengthEnumHandler.html#getQuantity(T,%20ome.xml.model.enums.UnitsLength)) - in https://github.com/openmicroscopy/bioformats/blob/develop/components/xsd-fu/templates-java/OMEXMLModelObject.template#L288. Is it worth reviewing how much this covers while we discuss a new quantity API?

sbesson:
Given the open questions on the Java side and the late hour, I propose to close this proposal and review it as part of the next breaking model series. In the short term, properly documenting our existing Units API in both Java and C++ implementations should be a priority. From there, we certainly need to identify the critical limitations and prioritize our future work while keeping in mind implementation across both languages.

rleigh-dundee:
Sorry, I didn't see the earlier questions.

Regarding the Java behaviour, agreed that it's internally implemented in update().  However, it's not enforced when setting or serialising back to XML.  So it's a partial solution only and unenforced by the Quantity instance.

The Handler logic only caters for numbers, not the model primitives.  It's fairly horrible, and likely shouldn't have been placed here; this could have been part of Quantity.

For this PR, the behaviour is correct for C++.  Given that this is a place where the language forces the implementations to differ, I don't think we need to tie the C++ implementation to the current Java limitations, or wait on Java to be fixed.  It's needed for the C++ implementation to enforce constraints, or else every single quantity will have zero checking at any point in serialisation, modification and deserialisation.  We already implemented Quantity as a template; this is just adjusting the specialisation to be correct.

sbesson:
> Regarding the Java behaviour, agreed that it's internally implemented in update(). However, it's not enforced when setting or serialising back to XML. So it's a partial solution only and unenforced by the Quantity instance.

> The Handler logic only caters for numbers, not the model primitives. It's fairly horrible, and likely shouldn't have been placed here; this could have been part of Quantity.

I was thinking of using the internal implementation as a component that could be reviewed and fixed within a stable series and give ourselves the time to properly refactor the quantity API.

> For this PR, the behaviour is correct for C++. Given that this is a place where the language forces the implementations to differ, I don't think we need to tie the C++ implementation to the current Java limitations, or wait on Java to be fixed. It's needed for the C++ implementation to enforce constraints, or else every single quantity will have zero checking at any point in serialisation, modification and deserialisation. We already implemented Quantity as a template; this is just adjusting the specialisation to be correct.

Having different implementations specific to each language is absolutely fair and expected. I do not disagree about the effectiveness of the C++ implementation but rather have some concerns about the understanding of its consequences for the rest of our team. On top of the cost of porting the code across language (https://github.com/openmicroscopy/bioformats/pull/1806 being a concrete example), do we have a concrete proposal for the corresponding Java API changes and their impact on the units code both in Bio-Formats and OMERO code that could be reviewed?

Obviously with all the effort put into the development of OME Files C++, these considerations need to go both ways and the C++ implementation will have to be included as part of the core API design.

rleigh-dundee:
@sbesson There's already a fairly wide split in terms of functionality.  The C++ primitive types (PositiveFloat etc.) are actual native numeric types, which can be freely used and modified anywhere and with any standard algorithms completely transparently.  The Java primitives are construct-only and non-modifiable and are not in any way transparent to use.

The Quantity types follow the same pattern.  The C++ Quantity is also a native numeric type and is also freely substitutable and modifiable anywhere.  It supports all the standard arithmetic operators with addition and subtraction of other compatible quantities (with transparent conversion) and dimensionless values for multiplication/division/modulus.  I.e. it can be freely used with standard algorithms where this makes logical sense, and unit conversion is explicit at construction time or implicit with arithmetic operations where the LHS unit is preserved and the RHS is converted.  The Java quantity is again construct only with a given unit type and value; you can do explicit unit conversions, but that's it.  All of the functionality is static and explicit.

The point I want to make here is that the C++ implementation is always going to differ at this level, and that merging or not merging this PR won't have any impact on these fundamental design considerations.  There's room to clean up and improve the Java implementation and make it *slightly* more similar to the C++ implementation, but it will never reach the same level of direct usability and transparency due to the language differences.  In terms of the impact, those basic design choices have already been made.  The small refinement of the C++ API in this PR doesn't really change any of those considerations.  But doing it here and now will avoid an API break in the future, since this does change the Quantity type specialisations used in the model and metadata APIs.

sbesson:
@rleigh-dundee: thanks for the thorough explanation, I had certainly overlooked the fact the C++ quantity API departed so much from the Java implementation, especially on the native numeric types front.

>  In terms of the impact, those basic design choices have already been made.

This is true albeit also the main concern on my side is that I do not recall these choices and their implications being clearly discussed or stated in the [C++ unit support addition](https://github.com/openmicroscopy/bioformats/pull/2442).
For future development of OME Files, we will need to find ways to consider both languages while designing new APIs and communicate these choices early with the rest of the team.

> The small refinement of the C++ API in this PR doesn't really change any of those considerations. But doing it here and now will avoid an API+ABI break in the future, since this does change the Quantity type specialisations used in the model and metadata APIs.

I agree this PR is fundamentally an API+ABI breaking change and want to make sure we have a good understanding of its implications. From the discussion above, I accept the fact we have already diverged enough on the quantity API so that merging this PR does not bring us significantly further apart. My understanding from the above is that on the Java side, a significant refactoring would be required to include similar features: is that correct or are there some immediate actions that could include some model enforcements at least partly? /cc @joshmoore @mtbc @dgault

sbesson:
@rleigh-dundee: thanks. From my side (but I would expect from anyone new to this repository), having these design principles clearly laid out is hugely helpful and having a reference document containing them would be appreciated either in the documentation or as a PR against https://github.com/openmicroscopy/design.

Merging this PR as I agree it does not lead us further away from the previous state and matches the C++ design which has already been implemented so far. Thinking about keeping the APIs in sync, now that we have a functional C++ implementation, the OME Files team priority over the upcoming months will be the development of the Java equivalent so that we can refactor our native readers, add the functionalities we need to deliver like tabular data and consume them in our stack.

PR #2455, opened 2016-06-14
cpp: ome-xml: Throw less exceptions in OMEModelObject::getChildrenByTagName

rleigh-dundee:
Rather than waiting for an exception to be thrown by the
Element RAII wrapper, check whether it will throw
beforehand.  This isn't strictly necessary, but it makes
the processing slightly more efficient and, more
importantly, makes it easier to debug the model code
since only important exceptions will be thrown.

--------

Testing: Check builds are green.  There's no change in functionality since this is purely an optimisation, so nothing specific to test with respect to behaviour changes.

### DIFF
--- a/cpp/lib/ome/xml/model/detail/OMEModelObject.cpp
+++ b/cpp/lib/ome/xml/model/detail/OMEModelObject.cpp
@@ -108,10 +108,17 @@ namespace ome
             {
               try
                 {
-                  common::xml::dom::Element child(pos->get(), false);
-                  if (child && name == stripNamespacePrefix(common::xml::String(child->getNodeName())))
+                  // This pointer check is unnecessary--the Element
+                  // class would throw; but this avoids the need to
+                  // throw and catch many exceptions during document
+                  // processing.
+                  if (dynamic_cast<const xercesc::DOMElement *>(pos->get()))
                     {
-                      ret.push_back(child);
+                      common::xml::dom::Element child(pos->get(), false);
+                      if (child && name == stripNamespacePrefix(common::xml::String(child->getNodeName())))
+                        {
+                          ret.push_back(child);
+                        }
                     }
                 }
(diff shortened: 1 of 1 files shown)


cpp/lib/ome/xml/model/detail/OMEModelObject.cpp
```diff
@@ -108,10 +108,17 @@ namespace ome
             {
               try
                 {
-                  common::xml::dom::Element child(pos->get(), false);
-                  if (child && name == stripNamespacePrefix(common::xml::String(child->getNodeName())))
+                  // This pointer check is unnecessary--the Element
+                  // class would throw; but this avoids the need to
```
dgault: Should this line state the exception that the Element class is throwing?

dgault:
It certainly does simplify things. I notice the catch block below also simply has a comment, would it be sensible to perhaps have a logged debug statement in the event of an invalid element?

cpp/lib/ome/xml/model/detail/OMEModelObject.cpp
rleigh-dundee: It's in the catch block parameter immediately below, so should be fairly obvious.

rleigh-dundee:
@dgault Element is a RAII wrapper over Xerces, it's expected and intended behaviour that it throw if it's not constructed with a valid DOMElement* pointer, so logging wouldn't be too useful--it would end up logging every DOMNode that's not a DOMElement, including every bit of whitespace etc.  That's all expected and isn't an error, so I think just ignoring that stuff should be fine.  Since the child nodes can be any valid Node type, we are simply selecting those we are interested in.

dgault:
That makes sense, if I look at it as a selection moreso than an exception then it would be a heck of a lot of logging which wouldn't of use then. Other than that it looks fine to me.

Should be good to merge.

PR #2456, opened 2016-06-15
Formats API - Reader and Writer Validation

dgault:
New validate property has been added to MetadataOptions with the default option set to true. As such support for MetadataOptions has been moved to FormatHandler level to allow for MetadataOptions to be used with both readers and writers. 

Validation occurs in the setID function after the file has been initialised and metadata populated to the store. This has been added for FormatReader and Writer as well as ImageReader and Writer. This should cover all implementations of IFormatReader and Writer that we have, in the case of ReaderWrappers the underlying reader will provide the validation.

ImageInfo has been updated to remove existing manual validation in favour of setting the property through the API. ImageConverter has been updated to add a new argument to match that of ImageInfo, it also sets the validation on both the reader and writer used for the conversion.

For the nightly test suite, testValidXML currently provides some validation coverage. As the new validation properties are currently on by default this should provide coverage through the existing tests with exceptions being thrown for any invalid samples we may have in the repo.

For testing the important thing is to test a range of possible readers such as:
- OME-TIFF
- OME-XML
- Tiff based format
- Non Tiff based format

It is also required to test the following workflows:
- Full repo tests should be green
- Through Fiji plugin, reading and exporting
- Using showinf and bfconvert tools (using both the default and the -novalid option)

Potential points for discussion:
- Is it correct to default validation to true?
- Is a FormatException suitable in the event of failure or simply log a warning?

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -190,9 +190,6 @@ public abstract class FormatReader extends FormatHandler
    */
   protected MetadataStore metadataStore = new DummyMetadata();
 
-  /** Metadata parsing options. */
-  protected MetadataOptions metadataOptions = new DefaultMetadataOptions();
-
   private ServiceFactory factory;
   private OMEXMLService service;
 
@@ -585,36 +582,6 @@ public abstract class FormatReader extends FormatHandler
   /** Return a properly configured loci.formats.meta.FilterMetadata. */
   protected MetadataStore makeFilterMetadata() {
     return new FilterMetadata(getMetadataStore(), isMetadataFiltered());
-  }
-
-  // -- IMetadataConfigurable API methods --
-
-  /* (non-Javadoc)
-   * @see loci.formats.IMetadataConfigurable#getSupportedMetadataLevels()
-   */
-  @Override
-  public Set<MetadataLevel> getSupportedMetadataLevels() {
-    Set<MetadataLevel> supportedLevels = new HashSet<MetadataLevel>();
-    supportedLevels.add(MetadataLevel.ALL);
-    supportedLevels.add(MetadataLevel.NO_OVERLAYS);
-    supportedLevels.add(MetadataLevel.MINIMUM);
-    return supportedLevels;
-  }
-
-  /* (non-Javadoc)
-   * @see loci.formats.IMetadataConfigurable#getMetadataOptions()
-   */
-  @Override
-  public MetadataOptions getMetadataOptions() {
-    return metadataOptions;
-  }
-
-  /* (non-Javadoc)
-   * @see loci.formats.IMetadataConfigurable#setMetadataOptions(loci.formats.in.MetadataOptions)
-   */
-  @Override
-  public void setMetadataOptions(MetadataOptions options) {
-    this.metadataOptions = options;
   }
 
   // -- IFormatReader API methods --
@@ -1463,7 +1430,7 @@ public abstract class FormatReader extends FormatHandler
         ((OMEXMLMetadata) store).resolveReferences();
         setupService();
         
-        if (validate) {
+        if (getMetadataOptions().isValidate()) {
           try {
             String omexml = service.getOMEXML((MetadataRetrieve)store);
             if (!XMLTools.validateXML(omexml)) {
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1467,13 +1467,13 @@ public abstract class FormatReader extends FormatHandler
           try {
             String omexml = service.getOMEXML((MetadataRetrieve)store);
             if (!XMLTools.validateXML(omexml)) {
-              LOGGER.warn("Invalid XML when retrieving OME-XML from OMEXMLMetadata object.");
+              throw new FormatException("Invalid XML when retrieving OME-XML from OMEXMLMetadata object.");
             }
             if (!service.validateOMEXML(omexml)) {
-              LOGGER.warn("Invalid OME-XML when retrieving OME-XML from OMEXMLMetadata object.");
+              throw new FormatException("Invalid OME-XML when retrieving OME-XML from OMEXMLMetadata object.");
             }
           } catch (ServiceException e) {
-            LOGGER.warn("OMEXMLService unable to create OME-XML metadata object.", e);
+            throw new FormatException("OMEXMLService unable to create OME-XML metadata object.", e);
           }
         }
         
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1434,13 +1434,13 @@ public abstract class FormatReader extends FormatHandler
           try {
             String omexml = service.getOMEXML((MetadataRetrieve)store);
             if (!XMLTools.validateXML(omexml)) {
-              throw new FormatException("Invalid XML when retrieving OME-XML from OMEXMLMetadata object.");
+              LOGGER.warn("Invalid XML when retrieving OME-XML from OMEXMLMetadata object.");
             }
             if (!service.validateOMEXML(omexml)) {
-              throw new FormatException("Invalid OME-XML when retrieving OME-XML from OMEXMLMetadata object.");
+              LOGGER.warn("Invalid OME-XML when retrieving OME-XML from OMEXMLMetadata object.");
             }
           } catch (ServiceException e) {
-            throw new FormatException("OMEXMLService unable to create OME-XML metadata object.", e);
+            LOGGER.warn("OMEXMLService unable to create OME-XML metadata object.", e);
           }
         }
         
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1434,7 +1434,7 @@ public abstract class FormatReader extends FormatHandler
           try {
             String omexml = service.getOMEXML((MetadataRetrieve)store);
             service.validateOMEXML(omexml);
-          } catch (ServiceException e) {
+          } catch (ServiceException | NullPointerException e) {
             LOGGER.warn("OMEXMLService unable to create OME-XML metadata object.", e);
           }
         }
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1429,7 +1429,7 @@ public abstract class FormatReader extends FormatHandler
       if (store instanceof OMEXMLMetadata) {
         ((OMEXMLMetadata) store).resolveReferences();
         setupService();
-        
+
         if (getMetadataOptions().isValidate()) {
           try {
             String omexml = service.getOMEXML((MetadataRetrieve)store);
@@ -1438,7 +1438,7 @@ public abstract class FormatReader extends FormatHandler
             LOGGER.warn("OMEXMLService unable to create OME-XML metadata object.", e);
           }
         }
-        
+
         for (int series=0; series<getSeriesCount(); series++) {
           setSeries(series);
 
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -46,13 +46,16 @@ import loci.common.DataTools;
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.common.services.DependencyException;
+import loci.common.services.ServiceException;
 import loci.common.services.ServiceFactory;
+import loci.common.xml.XMLTools;
 import loci.formats.in.DefaultMetadataOptions;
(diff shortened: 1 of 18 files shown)


melissalinkert:
Few general comments:

1. it would be good to put the validation and error reporting block into a separate method, so it isn't repeated in each of the classes where this check is performed
2. strong +1 to logging a warning instead of throwing an exception upon validation errors.  I think there are going to be cases (e.g. for old or third-party OME-XML/OME-TIFF) where the XML won't strictly validate - but that shouldn't stop the reader from giving a "best effort" to return everything it can read (and throwing an exception upon setId means the images are no longer accessible).  Or maybe that's another configuration option?
3. I'm not opposed to validation by default, but I'd expect this to have some performance impact, so minimally would be good to have a prominent doc page on this feature.  Might also be worth considering an INFO-level message to the effect of "you have validation turned on, so if you think this is too slow now do XYZ".

dgault:
I can certainly pull the validation out into a seperate block.

Regarding point number 2, the unit tests are indeed showing that there are many validation failures, most being things such as lsid id's not matching the desired patterns etc. So warnings rather than exceptions looks to be much the better option.

Regards point 3, running a large number of unit tests and it does have a notable impact especially when there are a significant number of warnings being printed, perhaps reverting the default to off may be worth considering.

components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java
```diff
@@ -71,5 +73,21 @@ public MetadataLevel getMetadataLevel() {
...
+  public boolean isValidate() {
+    return validate;
+  }
+
+  /* (non-Javadoc)
+   * @see loci.formats.in.MetadataOptions#setValidate(boolean)
+   */
+  @Override
+  public void setValidate(boolean validateMetadata) {
+    this.validate = validate;
```
sbesson: See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/267/findbugsResult/

components/formats-api/src/loci/formats/FormatReader.java
```diff
@@ -1459,7 +1429,16 @@ public void setId(String id) throws FormatException, IOException {
       if (store instanceof OMEXMLMetadata) {
         ((OMEXMLMetadata) store).resolveReferences();
         setupService();
-
+        
+        if (getMetadataOptions().isValidate()) {
+          try {
+            String omexml = service.getOMEXML((MetadataRetrieve)store);
+            service.validateOMEXML(omexml);
```
sbesson: From https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-test_images_good/, `NullPointerException` should also be handled if `omexml` is `null`.

sbesson:
Couple of points following a discussion with @dgault and @joshmoore.

After some investigation, the memoizer failures in the [MATLAB tests](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-matlab/415/) are related to the usage of `bioformats_package.jar` rather than the individual JARs. Similar outcome can be observed while using the command-line tools with `-cache` option. I will work on a PR to try and fix the packaging.

Reviewing the effects of the XML validation, a couple of comments:
- we have a lot of invalid content across our mock classes as shown in the Formats BSD unit tests which might be worth cleaning
- if we enable this option by default, we might want to up priority of the [default offline XML validation](https://trello.com/c/dRkIAjMP/10-offline-xml-validation)
- under some circumstances `./tools/showinf -nopix -cache`, the XML validation is logged twice. Does that mean the XML is validated multiple times

dgault:
I've gone through the Formats BSD unit tests fixing any validation errors encountered, so that should at least be one set clean for the next build.

sbesson:
[Maven tests](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-maven/74/) are now much cleaner thanks. Also file opening/caching using the command-line tools seems to be fixed. However, it seems the [MATLAB memoizer functionality](and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-matlab/) is still broken:

```
>> bfCheckJavaPath();
>> loci.common.DebugTools.enableLogging('DEBUG');
>> r=loci.formats.Memoizer(bfGetReader(),0);
processor is INTEL_64 os.arch is amd64
architecture is LINUX_64 os.name is linux
architecture is LINUX_64 os.name is linux
platform specific path is META-INF/lib/linux_64/
mappedLib is libturbojpeg.so
URL is jar:file:/opt/hudson/workspace/BIOFORMATS-DEV-breaking-matlab/artifacts/bfmatlab/bioformats_package.jar!/META-INF/lib/linux_64/libturbojpeg.so
URL path is file:/opt/hudson/workspace/BIOFORMATS-DEV-breaking-matlab/artifacts/bfmatlab/bioformats_package.jar!/META-INF/lib/linux_64/libturbojpeg.so
Extracting 'jar:file:/opt/hudson/workspace/BIOFORMATS-DEV-breaking-matlab/artifacts/bfmatlab/bioformats_package.jar!/META-INF/lib/linux_64/libturbojpeg.so' to '/tmp/libturbojpeg5686861998647302882.so'
>> r.setId('test.fake')
deleting invalid memo file: /fc/homes/sbesson/.test.fake.bfmemo
start[1467712557381] time[160] tag[loci.formats.Memoizer.loadMemo]
Loaded properties from: services.properties
Added interface interface loci.formats.services.POIService and implementation class loci.formats.services.POIServiceImpl
Added interface interface loci.formats.services.MDBService and implementation class loci.formats.services.MDBServiceImpl
Added interface interface loci.formats.services.JPEGTurboService and implementation class loci.formats.services.JPEGTurboServiceImpl
Added interface interface loci.formats.services.JAIIIOService and implementation class loci.formats.services.JAIIIOServiceImpl
Added interface interface loci.formats.services.WlzService and implementation class loci.formats.services.WlzServiceImpl
Added interface interface loci.formats.services.JHDFService and implementation class loci.formats.services.JHDFServiceImpl
Added interface interface loci.formats.services.NetCDFService and implementation class loci.formats.services.NetCDFServiceImpl
Added interface interface loci.formats.services.EXIFService and implementation class loci.formats.services.EXIFServiceImpl
Added interface interface loci.formats.services.MetakitService and implementation class loci.formats.services.MetakitServiceImpl
Added interface interface loci.formats.services.LuraWaveService and implementation class loci.formats.services.LuraWaveServiceImpl
Added interface interface loci.formats.services.OMEXMLService and implementation class loci.formats.services.OMEXMLServiceImpl
Using mapped byte buffer? false
loci.formats.in.FakeReader.initFile(test.fake)
Parsing schema path
Validating OME-XML
No validation errors found.
saved to temp file: /fc/homes/sbesson/.test.fake.bfmemo3457280658340907828
start[1467712558229] time[198] tag[loci.formats.Memoizer.saveMemo]
saved memo file: /fc/homes/sbesson/.test.fake.bfmemo (40845 bytes)
start[1467712557334] time[1093] tag[loci.formats.Memoizer.setId]
>> r.close()
Location.mapFile: embedded-stream.raw -> null
Location.mapFile: embedded-stream.raw -> null
>> r.setId('test.fake')
deleting invalid memo file: /fc/homes/sbesson/.test.fake.bfmemo
start[1467712562628] time[51] tag[loci.formats.Memoizer.loadMemo]
loci.formats.in.FakeReader.initFile(test.fake)
Parsing schema path
Validating OME-XML
No validation errors found.
saved to temp file: /fc/homes/sbesson/.test.fake.bfmemo1582770655526703220
start[1467712562696] time[91] tag[loci.formats.Memoizer.saveMemo]
saved memo file: /fc/homes/sbesson/.test.fake.bfmemo (40843 bytes)
start[1467712562628] time[159] tag[loci.formats.Memoizer.setId]
>> r.close()           
```

Would it be worth logging the exception leading to invalid file deletion at  `DEBUG` level in `Memoizer.java`?

dgault:
Ok from what I can gather the Matlab build on trout is using R2013b which has a xercesImpl.jar version 2.9, whereas we are using version 2.8.1

This could be the cause of the Matlab errors in Memoizer:

`deleting invalid memo file: /opt/hudson/workspace/BIOFORMATS-DEV-breaking-matlab/.test.fake.bfmemo
Other Exception: tried to access class org.apache.xerces.util.XMLAttributesImpl$Attribute from class org.apache.xerces.util.XMLAttributesImpl$AttributeConstructorAccess
start[1467736926903] time[53] tag[loci.formats.Memoizer.loadMemo]`

sbesson:
@dgault: the `xerces` version conflict is not completely unsurprising. Also disabling the XML validation using the same approach as the command-line tools i.e.

```
r = loci.formats.Memoizer(bfGetReader(), 0);
options = loci.formats.in.MetadataOptions();
options.setValidate(false);
r.setMetadataOptions(options);
r.setId('test.fake');
```

is sufficient to restore the memoization ability in MATLAB.

Removing the `breaking` label to include this PR into tomorrow's merge build and consider it for inclusion in the API freeze. A couple of points that we may need to discuss as part of the final decision:
- setting the default validation option to `false` by default
-  `xerces` dependency and version number (esp. with https://github.com/openmicroscopy/bioformats/pull/2473)
- potential impact of this PR without the complete review of the `MetadataOptions` stack (cf IDR work)

joshmoore:
Primary concern at the API level (as discussed on zoom) is whether or not this will ever need to be extensible/configurable. i.e. like logging levels is there a "DEBUG" validation versus a "FATAL" validation. At that point, the `boolean` could still be used to enable/disable the entire apparatus, but if we see anything more complex coming, it might be worth considering.

Otherwise, there are a few minor(?) implementation changes here which confuse me, but they may be necessary:

 * https://github.com/openmicroscopy/bioformats/pull/2456/files#diff-3226e29a1c55ed5a5aa34239af9c1e8fR173
 * https://github.com/openmicroscopy/bioformats/pull/2456/files#diff-234fed54f2f1d834179bb0682cc14babR824
 * https://github.com/openmicroscopy/bioformats/pull/2456/files#diff-2a27cd8fb03fc0ccfa2a41c05d2bc9bcR268

(And the setupService method looks ripe for refactoring, but without digging, I don't know where it would live)

rleigh-dundee:
Whether we need the OMEXMLService at all is one additional question I would have (unrelated to this PR), since the service interface and implementation live alongside the metadata/model code it no longer seems to be necessary when you can use the OMEXMLMetadatastore directly.  All the service stuff could likely be made into static methods in MetadataTools or equivalent.

components/formats-api/src/loci/formats/in/MetadataOptions.java
```diff
@@ -46,4 +46,10 @@
    */
   MetadataLevel getMetadataLevel();
 
+  /** Specifies whether or not to validate files when reading. */
+  void setValidate(boolean validate);
+
+  /** Returns true if files should be validated when read.*/
+  boolean isValidate();
+
```
rleigh-dundee: Trailing whitespace

components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java
```diff
@@ -71,5 +73,21 @@ public MetadataLevel getMetadataLevel() {
   public void setMetadataLevel(MetadataLevel level) {
     this.level = level;
   }
+
```
rleigh-dundee: Trailing whitespace

rleigh-dundee:
There's trailing whitespace introduced in most of the files in this PR (12 cases); please could they be cleaned up?

rleigh-dundee:
Overall looks good.  I'd certainly like to see this made the default once any validation warnings in the full repo test are cleaned up, since correctness of our use of the metadata model is quite important to be getting right.

melissalinkert:
Latest commits read fine to me, so no veto to merging as-is.  It might be nice to have clearer javadocs in the new ```MetadataOptions``` methods at some point - we all know what they mean, but for anyone else it wouldn't necessarily be clear that those check/enable validation of the generated OME-XML, not the file itself.  Certainly doesn't need to block this PR though.

sbesson:
Merging this PR as this is adding a feature in sync with the C++ API and will help our tools flagging invalid OME-XML early with the caveats mentioned above:
- when evaluating advanced versions of this API (logging, failures), we should certainly evaluate flexible implementations such as the one that was proposed in the `metadata` branch
- the new API Javadocs can be reviewed in a separate PR
- the automated reader tests front already run a `testValidXML` per file so I would expect no additional need here (last [build](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/221/) had 2 unrelated transient failures)
- the command-line tools options in https://github.com/openmicroscopy/bioformats/pull/2456/commits/119b6f5fd978cf6061b0df208b015f1094a6d3ea are slightly confusing and for `showinf` you already need to pass `-omexml` to store the OME-XML metadata. While keeping the default value in the `DefaultMetadataOptions` API to `false` is a good compromise for now, my preference would be to set `validate` as the default in our command-line tools `showinf/bfconvert` and use `-no-valid` to disable it as currently. This can be evaluated in a separate PR as well.

PR #2440, opened 2016-05-27
Support for more native libraries

RichardMyers:
Remove static check for windows x32/x64 architecture; now checks at
runtime if correct dynamic library exists.  SlideBook ImageJ update site
also now has OS X x32, x64 and Linux x64 libraries.

PR #2457, opened 2016-06-15
ClassList: discovery improvements 

sbesson:
This PR is an incremental step towards the reader discoverability described in https://github.com/openmicroscopy/design/issues/42. The changes in this Pull Request are restricted themselves to `ClassList` which is used in conjunction with the `readers.txt/writers.txt` configuration files in `ImageReader/ImageWriter` to instantiate a series classes implementing the `IFormatReader/IFormatWriter` interface.

### Main features

The major changes proposed in this PR are the following:
- refactors the `ClassList` notably decomposes the constructor with a configuration file into  a set of public methods can be tested independently and expose more granularity for the generation of class lists
- improves error handling when dealing with invalid input streams
- add `append()`, `prepend()` methods allowing to combine `ClassList` objects
- add a mechanism for storing a map of options in addition to a list of classes into a `ClassList` object
- add logic for parsing options stored as comma-separated key/value pairs in a configuration file

### Testing

All changes described above as well as generic constructor/method tests for the `loci.formats.ClassList` class should be covered via a set of unit tests in a new `ClassListTest` class using the `java.util.Iterable` interface is used as the base class. 

### Applications

The current state of this PR tries to address the following use cases:
- optional/external/discoverable readers could be annotated into the `readers.txt` class using a controlled key/value pair:

  ```
loci.formats.in.SlideBook6Reader[optional=true]  # sld
```

- additional per-reader options can be specified at the readers.txt level or even parsed externally e.g. using the case of https://github.com/openmicroscopy/bioformats/pull/2458, the following line

```
loci.formats.in.FlexReader[granularity=infinite,depth=10]  # flex
```

would store options `{loci.formats.in.FlexReader.granularity=infinite, loci.formats.in.FlexReader.depth=10}` into the `ClassList` options. A proposed subsequent step at the `ImageReader` level would be to pass these options to the `MetadataOptions` /cc @joshmoore @dgault

### DIFF
--- a/components/formats-api/src/loci/formats/ClassList.java
+++ b/components/formats-api/src/loci/formats/ClassList.java
@@ -101,24 +101,10 @@ public class ClassList<T> {
   {
     this(base);
     if (file == null) return;
-    classes = parseFile(file, location, true);
+    classes = parseFile(file, location);
   }
 
   // -- ClassList API methods --
-
-  /**
-   * Parses a list of classes from a configuration file.
-   * @param file Configuration file containing the list of classes.
-   * @param location Class indicating which package to search for the file.
-   *        If {@code null}, 'file' is interpreted as an absolute path name.
-   * @return A list of classes parsed from the file
-   * @throws IOException if the file cannot be read.
-   */
-  public List<Class<? extends T>> parseFile(String file, Class<?> location)
-    throws IOException
-  {
-    return parseFile(file, location, false);
-  }
 
    /**
     * Parses a list of classes from a configuration file.
@@ -128,7 +114,7 @@ public class ClassList<T> {
     * @return A list of classes parsed from the file
     * @throws IOException if the file cannot be read.
     */
-  public List<Class<? extends T>> parseFile(String file, Class<?> location, boolean strict)
+  public List<Class<? extends T>> parseFile(String file, Class<?> location)
     throws IOException
   {
     List<Class<? extends T>> parsedClasses = new ArrayList<Class<? extends T>>();
@@ -183,7 +169,7 @@ public class ClassList<T> {
         LOGGER.debug("", exc);
       }
       if (c == null) {
-        if (strict) LOGGER.error("\"{}\" is not valid.", line);
+        LOGGER.error("\"{}\" is not valid.", line);
         continue;
       }
       parsedClasses.add(c);
--- a/components/formats-api/src/loci/formats/ClassList.java
+++ b/components/formats-api/src/loci/formats/ClassList.java
@@ -90,7 +90,7 @@ public class ClassList<T> {
    * @param file Configuration file containing the list of classes.
    * @param base Base class to which all classes are assignable.
    * @param location Class indicating which package to search for the file.
-   *  If null, 'file' is interpreted as an absolute path name.
+   *        If {@code null}, 'file' is interpreted as an absolute path name.
    * @throws IOException if the file cannot be read.
    */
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: components/formats-api/src/loci/formats/ClassList.java [hunk 1]
```diff
@@ -90,26 +92,63 @@ public ClassList(String file, Class<T> base) throws IOException {
...
+  {
+    return parseFile(file, location, false);
+  }
+
+   /**
+    * Parses a list of classes from a configuration file.
+    * @param file Configuration file containing the list of classes.
+    * @param location Class indicating which package to search for the file.
+    *        If {@code null}, 'file' is interpreted as an absolute path name.
+    * @return A list of classes parsed from the file
```
mtbc: `@param strict ...`

rleigh-dundee:
I'm likely not the best person to give this a detailed review, but from my reading of it, it looks sensible.

sbesson:
After some discussions with @joshmoore while trying to come up with a generic interface worth maintaining, the last set of commits propose a second API change allowing to annotate classes using key/value pairs into the readers.txt file. The PR description has been updated to reflect the latest proposal.

melissalinkert:
Makes sense - this is pretty cool.  Two main questions from my side:

1. Is there a list of "approved" keys?  It might be nice to allow anything, but log on anything not on the list (minimally for catching typos).
2. If you want to append/prepend a class outside readers.txt, is it worth having an easier way to set options for that class as well, i.e. without constructing a well-formatted string and calling ```parseOptions```?

sbesson:
@melissalinkert: last few commits should add an `addOption(String, String)` method (better names welcome), improve the options string parsing ;pgoc and provide a first proposal for whitelisting option keys. On the latter, a couple of questions:

- as the proposal is to store reader options as `package.reader.key, value` should the whitelisting apply to the whole key or the trailing part of it?
- do we need a mechanism for allowing the user to override whitelisted values?

melissalinkert:
Latest commits look good to me.

> as the proposal is to store reader options as package.reader.key, value should the whitelisting apply to the whole key or the trailing part of it?

Probably just the trailing part, otherwise that's going to quickly get annoying for keys that apply to multiple readers.

> do we need a mechanism for allowing the user to override whitelisted values?

I wouldn't think so, since the key/value is stored no matter what, but not strongly opposed if you or anyone else thinks that would be useful.

sbesson:
All understood. Leaving @joshmoore to have a final look/say on the state of this PR. Otherwise I would propose to merge this API change (which is non breaking) and get the whitelisting improvement done and tested alongside an implementation of this PR separately.

joshmoore:
Few last thoughts:

 * Good to eventually have the format from https://github.com/openmicroscopy/bioformats/pull/2457/files#diff-b3f8513ea964544bab39d91b1641bc5eR147 and perhaps an example of the configuration in the class javadoc
 * There's still the issue of the "well-known keys" to discuss. I assume that's for follow-on PRs and a more general doc'ing effort.
 * Adjustments may be necessary as we start to use this in anger; so as long as that's pre-GA, :thumbsup:

sbesson:
Captured follow-up as https://trello.com/c/k1RljMtO/162-classlist-documentation (docs) and  https://trello.com/c/EPqD1xkr/163-externalize-slidebook6reader (implementation/keys).

mtbc:
Now merged, anything further to mention on #2440 @sbesson?

PR #2463, opened 2016-06-22
docs: Fix licensing URL in global TOC

rleigh-dundee:
Missing `http:` which breaks the docs when not staged on openmicroscopy.org


hflynn:
👍

PR #2468, opened 2016-06-22
Remove dead alicona 3D spec doc link

hflynn:
This is the second time I've had to fix this link and as it's no longer hosted by the company and is 13 years old, I propose just dropping the link.

Should make https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/ green again.


sbesson:
Agreed. The pain of maintaining this external link is not worth the effort.

PR #2469, opened 2016-06-23
cmake: Find OME components individually

rleigh-dundee:
See https://github.com/ome/ome-common-cpp/pull/21 for full description and testing instructions.

### DIFF
--- a/cpp/cmake/Version.cmake
+++ b/cpp/cmake/Version.cmake
@@ -1,7 +1,7 @@
 # #%L
 # Bio-Formats C++ libraries (cmake build infrastructure)
 # %%
-# Copyright © 2006 - 2016 Open Microscopy Environment:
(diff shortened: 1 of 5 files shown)


cpp/cmake/Version.cmake
```diff
@@ -1,7 +1,7 @@
 # #%L
 # Bio-Formats C++ libraries (cmake build infrastructure)
 # %%
-# Copyright © 2006 - 2016 Open Microscopy Environment:
+# Copyright © 2006 - 2015 Open Microscopy Environment:
```
sbesson: This can be reverted

rleigh-dundee: Note it's just a copy of what was in the Version.cmake in all the other repos--it needs updating across the board.

sbesson: Understood. Let's do that in a separate effort once all jobs are green.

PR #2470, opened 2016-06-23
docs: Rename .txt to .rst

rleigh-dundee:
This is a straightforward rename, done with:

```
find docs/sphinx -name '*.txt' | grep -v CMakeLists.txt | while read f; do git mv "$f" "${f%.txt}.rst"; done
```

and then changing `source_suffix` in `conf.py.in` to `.rst`.

--------

Testing:

Jobs should remain green; there should be no changes in the docs build.

Check that the format page autogeneration works and generates `.rst` files rather than `.txt`.


melissalinkert:
Is the corresponding update to the format page autogeneration to be done here, or planned for a follow-up PR?

rleigh-dundee:
@melissalinkert It can be done here as well.  Is that the only additional piece which would need updating?

hflynn:
Builds are all green and the staging pages seem fine so I guess this is basically a policy decision.

hflynn:
Change would have to be made against all the relevant repos and re-done on dev_5_1 here and dev_5_2 on the ome-docs too. That seems an awful lot of work for a change that doesn't achieve very much to my mind. It would also make rebasing older develop PRs in the OMERO docs annoying and I do have several candidates on my list for that. So overall, I'm edging towards too much work for too little benefit personally.

rleigh-dundee:
Why would we need to make this on 5.1 at all?  There's no need to retrospectively fix up the older branches.

hflynn:
Probably less needed here given 5.2 timeline but on OMERO we need to be able to rebase stuff to 5.2.x incase something comes up before we get 5.3 out - urgent fixes are enough bother without adding to our workflow by making rebasing a pain.

rleigh-dundee:
@hflynn You can cherry-pick between branches and git will handle the renaming.

hflynn:
what about snoopy rebasing?

rleigh-dundee:
It should also be fine.  It uses the common parent of the two histories and should handle the renaming.  But as I said, doing the change is trivial.  I can open the necessary PRs this afternoon if needed.  It's definitely not a hard task.

sbesson:
Given the number of open questions to be answered and the fact @hflynn is unavailable next week, closing this PR for now. We can work thought the file renaming plan including its implications for the team workflow at future times when we have low-activity on this component.

rleigh-dundee:
This was made out to be much more of a problem than it would have been.  Git can cope with renaming when cherry-picking and rebasing.  I could open a PR to rename the other branch right now.

hflynn:
Ok, but Friday afternoon when I'm trying to get the model docs out and I'm away all next week isn't the best time.

rleigh-dundee:
Why?  It's a trivial rename we could easily get done on Monday.

PR #2472, opened 2016-06-24
Catch LinkageError to fix loci_tools with ImageJ

sbesson:
The switch from throwable to ReflectiveOperationException does not catch `java.lang.NoClassDefFoundError` thrown by `Log4jTools`.

To test this PR, check `loci_tools.jar` on ImageJ stalls when running the Bio-Formats Importer without this PR but works as expected with this PR included.


mtbc:
Code change looks fine to me.

dgault:
Testing with ImageJ works as expected with this PR included. Change looks good.

PR #2473, opened 2016-06-28
JPEG - Read Physical Size X and Y

dgault:
This is a reworked version of an old PR - https://github.com/openmicroscopy/bioformats/pull/2036
Original thread - https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7877&p=16128#p16128

This PR adds functionality to read the Exif tags TAG_Y_RESOLUTION and TAG_X_RESOLUTION.
Originally this PR was closed as these tags were not supported by the metadata-extractor library we were using and suitable replacements could not be found.

I have since updated the library to the latest revision enabling us to access the required tags.

To test:
We currently are not using the Physical Size X and Y, this PR simply provides users the ability to set the values on the Exif tags and have them remain as part of their metadata. The main change here is the upgrade of the metadata-extractor jar. As such it is important that this has no side affects and previous functionality is unchanged.  Ensuring that all builds and repo runs remain green would be a good indictor that the upgraded jar is working as expected.


dgault:
I have updated to remove the Missing Xerces test for EXIFService.
With the updated Exif library there is no longer a Xerces dependancy with metadata-extractor, as such the dependancy check was removed from EXIFServiceImpl and the unit test can also be removed.

sbesson:
For reference the relevant commit removing the xerces dependency in metadata-extractor is https://github.com/drewnoakes/metadata-extractor/commit/ad0e9ace448219ce62a4d6d5fbec7deae0258711.

The last merge commit seems to have reverted https://github.com/openmicroscopy/bioformats/pull/2473/commits/29b28ec4d3a06a92a9f8be58d597a99bbc34cf55 so the build are still failing.

components/formats-bsd/src/loci/formats/in/JPEGReader.java
```diff
@@ -207,8 +211,28 @@ protected void initFile(String id) throws FormatException, IOException {
         }
 
         HashMap<String, String> tags = exif.getTags();
-        for (String tagName : tags.keySet()) {
-          addGlobalMeta(tagName, tags.get(tagName));
+        for (String tagName : tags.keySet()) {  
+          if (tagName.equals(exif.getTagName(ExifSubIFDDirectory.TAG_X_RESOLUTION))) {
```
sbesson: Should we also parse the ResolutionUnit Exif tag to construct the the physical size using the correct unit?

dgault: I had removed Resolution Unit from the previous iteration of the PR as the documented options for it were inches or cm, which didn't really seemed tailored to this application. We could add it back in again though.

components/formats-bsd/src/loci/formats/in/JPEGReader.java
```diff
@@ -207,8 +211,28 @@ protected void initFile(String id) throws FormatException, IOException {
...
+            try {
+              Length sizeX = FormatTools.getPhysicalSizeX(Double.parseDouble(tags.get(tagName)));
```
sbesson: As mentioned in a conversation, my reading of the EXIF spec (see e.g http://www.exiv2.org/tags.html) is that the `{XY}Resolution` value should be map to the inverse of our `PhysicalSize{XY}`.

components/ome-xml/src/ome/xml/model/Map.java
```diff
@@ -0,0 +1,179 @@
+/*
```
sbesson: This file might have been erroneously reintroduced.

dgault: Im not even sure how that ended up in the commit, removing it now

sbesson:
@dgault: the `test-no-xerces` target and its references in the [Ant build file](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/build.xml) should also be dropped to fix the [failing build](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/282)

sbesson:
@dgault: in addition to the bug reported by FindBugs and fixed in https://github.com/openmicroscopy/bioformats/pull/2473/commits/36b3cc3e07899204db348b4170e86df704428d3d, I briefly tested the PR using the sample in http://www.exiv2.org/sample.html which should contain the tags for which support is added in this PR. The metadata does not seem to be parsed as part of this PR. Could it be that restricting the EXIF service implementation to the first EXIF directory is too limiting and we need to consider other directories ?

As a side-effect of this PR, the `metadata-extractor` bump seems to have removed the only place in our code with a dependency on `xercesImpl` so we might want to clean up this dependency in a follow-up PR.

dgault:
Having implemented the reading of multiple directories and done some further testing with that sample file i've come across 2 potential issues:
1 - For that sample the resolution tags aren't on an ExifSubIFDDirectory, should we therefore read tags from all available directories?
2 - Fiji ships with the older version of the metadata-extractor jar, which I believe is being used elsewhere in their application. As there are breaking API changes between the 2 versions we then get conflict issues when running through the BF plugin

sbesson:
Re tag parsing, I can certainly see the reader parsing metadata from all the directories although there might be some performance issues to consider.

Good point about `metadata-extractor` and Fiji. Quickly checking `mvn dependency:tree` from https://github.com/fiji/fiji/, it seems that Bio-Formats is the main dependency requiring it but @ctrueden might want to chime in here about potential caveats.

ctrueden:
Indeed, the only reason Fiji ships `metadata-extractor` is because it is a dependency of Bio-Formats. Nothing else uses it directly. You can verify this by viewing the dependency hierarchy of `fiji/fiji` in e.g. Eclipse.

Fiji certainly wants to be shipping compatible versions of Bio-Formats dependencies. The Bio-Formats upgrade tooling only updates the Bio-Formats JARs, not its dependencies—new versions of those must be uploaded manually.

But this will only matter when we update to Bio-Formats 5.2.0, right? Not the 5.1.x line?

sbesson:
@ctrueden: sorry for the delay. You are right, this dependency bump is only for Bio-Formats 5.2.0 which is scheduled for the end of next month. Let us know what is the best process to record and communicate all the dependency changes that might have to be handled manually.

On the `xerces` front, following https://github.com/openmicroscopy/bioformats/pull/2473#issuecomment-229893100, it seems like this PR coupled with https://github.com/openmicroscopy/bioformats/pull/2456 seems to have broken the [Maven BSD tests](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-maven/253/).

sbesson:
Discussed with @dgault, removing the exclude label so that he can manage exclusion or inclusion in breaking/merge builds via comments.

dgault:
Im removing this from breaking as the Memoizer tests are being skipped due to the tmp directory not being writable.

The failed test is an odd one, as it failed the MemoizerTest for certain tests but not others. It also failed on the second read of the file as opposed to the initial. I have tested this locally with every combination of jars being included and excluded and have been unable to reproduce the same error. In fact having run the same tests that failed through debug mode with various configurations I never saw it attempt to register the same class that it failed with. However the fact it happened in that test run shows that it must be possible under certain conditions.

In the failed test we were seeing an IllegalAccessError when Kryo was attempting to register the class:
com.sun.org.apache.xerces.internal.util.XMLAttributesImpl

There are 2 versions of XMLAttributesImpl:
- com.sun.org.apache.xerces.internal.util.XMLAttributesImpl
- org.apache.xerces.util.XMLAttributesImpl

The second is version found in our Xerces library and which would have been part of the metadata extractor dependancy which is no longer in place. I assume this is the version which normally would have been picked up by Kryo, but with it missing it located the first. This version is located in the Java 1.7 packages but has accessibility restrictions limiting it (which would make sense for the IllegalAccessError).

As such although I have been unable to reproduce the access error it would seem that we do need to include the Xerces library as a runtime dependency for the Memoizer. As such I have added the dependency for the version (2.8.1) which we have in the jars folder and build properties. This should simply allow us to upgrade the metadata-extractor while maintaining the Xerces dependency which it has dropped.

dgault:
I have updated this to expand the list of directories being searched and allowed for more flexible handling of the units tag. The units should be as below but will also handle string representations as these can be found in certain samples and throw a number format exception otherwise.

0x0128	ResolutionUnit	     int16u
1 = None 
2 = inches 
3 = cm

The X and Y Resolution tags have also been found with String representations which are not expected. See http://www.exiv2.org/sample.html for an example which contains "140/1136000 inches"

In these cases parsing all possible combinations could be problematic. In these cases its probably best that we handle those which are as in the spec and any others will get added as a piece of globalMetadata. This is how the current implementation works as in this PR.

components/formats-bsd/src/loci/formats/services/EXIFServiceImpl.java
```diff
@@ -82,26 +79,48 @@ public void initialize(String file) throws ServiceException, IOException {
   @Override
   public HashMap<String, String> getTags() {
     HashMap<String, String> tagMap = new HashMap<String, String>();
-    if (directory != null) {
-      for (Tag tag : directory.getTags()) {
-        tagMap.put(tag.getTagName(), tag.getDescription());
+    if (directories != null) {
+      for (Directory directory : directories) {
```
sbesson: Taking another look at the EXIF spec, I realize this might not achieve what we want. The `Image` and the `Thumbnail` EXIF tags are located in different IFDs but the tags overlap. In the case of  `XResolution`  this means the tag will be overwritten by the ThumbnailXResolution if present.

components/formats-bsd/src/loci/formats/services/EXIFServiceImpl.java
```diff
@@ -82,26 +79,48 @@ public void initialize(String file) throws ServiceException, IOException {
...
+        for (Tag tag : directory.getTags()) {
+          tagMap.put(tag.getTagName(), tag.getDescription());
```
sbesson: Not this PR fault, but I realize `tag.getDescription()` converts the values in the EXIF tag using some string formatting - see e.g. https://github.com/drewnoakes/metadata-extractor/blob/2.9.1/Source/com/drew/metadata/exif/ExifDescriptorBase.java#L262 which might prevent proper parsing in `JPEGReader`.

components/formats-bsd/src/loci/formats/in/JPEGReader.java
```diff
@@ -207,8 +212,47 @@ protected void initFile(String id) throws FormatException, IOException {
...
+        }
+        for (String tagName : tags.keySet()) {  
+          if (tagName.equals(exif.getTagName(ExifSubIFDDirectory.TAG_X_RESOLUTION))) {
+            try {
+              double resolutionX = Double.parseDouble(tags.get(tagName));
+              Length sizeX = FormatTools.getPhysicalSizeX((1/resolutionX),units);
+              store.setPixelsPhysicalSizeX(sizeX, 0);
+            }
+            catch(NumberFormatException e) {
+              LOGGER.warn("EXIF Tag - Resolution X not in correct format, expected double, found : " + tags.get(tagName));
```
sbesson: My main concern with the usage of `WARN` level is that it can be invasive e.g. in the Fiji console and/or OMERO logs. Alternatively we could use `DataTools.parseDouble()` which will log incorrect formatting at `DEBUG` level.

sbesson:
A couple of issues noted with the parsing of the EXIF tags. A slightly more concerning issue is that running `showinf -omexml -nopix ...` on some EXIF sample files, I get an empty OME-XML while it is properly populated from the `HEAD` of `develop`: 

```
[sbesson@ome-c6100-1 ~]$  /opt/hudson/workspace/BIOFORMATS-DEV-merge-build/tools/showinf -version
Version: 5.2.0-DEV
VCS revision: fdd1fd2ae881a463c333106712bb4a7ad3d3d94f
Build date: 4 August 2016
[sbesson@ome-c6100-1 ~]$ /opt/hudson/workspace/BIOFORMATS-DEV-merge-build/tools/showinf -nopix -omexml /ome/data_repo/curated/jpeg/exif/DCP_0433_com.JPG 
Checking file format [JPEG]
Initializing reader
JPEGReader initializing /ome/data_repo/curated/jpeg/exif/DCP_0433_com.JPG
Parsing schema path
Validating OME-XML
No validation errors found.
Populating metadata
Parsing JPEG EXIF data
EXIF Tag - Resolution X not in correct format, expected double, found : 72 dots per inch
EXIF Tag - Resolution Y not in correct format, expected double, found : 72 dots per inch
Parsing schema path
Validating OME-XML
No validation errors found.
Initialization took 1.58s

Reading core metadata
filename = /ome/data_repo/curated/jpeg/exif/DCP_0433_com.JPG
Series count = 1
Series #0 -- DCP_0433_com.JPG:
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (true color)
	Width = 1760
	Height = 1168
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Thumbnail size = 128 x 84
	Endianness = motorola (big)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata
Aperture Value: f/8.0
Color Space: sRGB
Component 1: Y component: Quantization table 0, Sampling factors 2 horiz/2 vert
Component 2: Cb component: Quantization table 1, Sampling factors 1 horiz/1 vert
Component 3: Cr component: Quantization table 1, Sampling factors 1 horiz/1 vert
Components Configuration: YCbCr
Compression: JPEG (old-style)
Compression Type: Baseline
Copyright: KODAK DC280 ZOOM DIGITAL CAMERA 
Data Precision: 8 bits
Date/Time Digitized: 2000:11:02 21:29:27
Date/Time Original: 2000:11:02 21:29:27
Exif Image Height: 1168 pixels
Exif Image Width: 1760 pixels
Exif Version: 2.10
Exposure Bias Value: 0 EV
Exposure Index: 100
Exposure Time: 1/90 sec
F-Number: f/8.0
File Modified Date: Mon May 19 19:41:15 +01:00 2014
File Name: DCP_0433_com.JPG
File Size: 688847 bytes
File Source: Digital Still Camera (DSC)
Flash: Flash did not fire
FlashPix Version: 1.00
Focal Length: 6.3 mm
Image Height: 1168 pixels
Image Width: 1760 pixels
Interoperability Index: Recommended Exif Interoperability Rules (ExifR98)
Interoperability Version: 1.00
JPEG Comment: Intel(R) JPEG Library, version 1,5,4,36
Make: EASTMAN KODAK COMPANY
Makernote: [572 values]
Max Aperture Value: f/2.9
Metering Mode: Average
Model: KODAK DC280 ZOOM DIGITAL CAMERA
Number of Components: 3
Orientation: Top, left side (Horizontal / normal)
Resolution Unit: Inch
Scene Type: Directly photographed image
Sensing Method: One-chip color area sensor
Shutter Speed Value: 1/90 sec
Thumbnail Length: 8453 bytes
Thumbnail Offset: 1488 bytes
White Balance: Unknown
YCbCr Positioning: Center of pixel array

Reading metadata

Generating OME-XML (schema version 2016-06)
<OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2016-06" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2016-06 http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd"/>

[sbesson@ome-c6100-1 ~]$
```

Marking this PR as `breaking` to investigate whether it is the cause of this issue.

sbesson:
After investigation, the issue reported in https://github.com/openmicroscopy/bioformats/pull/2473#issuecomment-237559415 is unrelated to this PR but rather a side-effect of the https://github.com/openmicroscopy/bioformats/pull/2456 which seems to effect all readers extending `LegacyReader`. Opening these files using `showinf -nopix -omexml -novalid` is sufficient to restore the OME-XML.

sbesson:
Discussed with @dgault, given the remaining questions about the metadata-extractor API and how we use it to populate EXIF metadata in our model, closing this PR in the scope of Bio-Formats 5.2.0 and postponing the dependency bump. In the meantime, we can keep collecting samples and examples of usage for reading these tags.

eduramiba:
Is JPEG physical size available somehow? I am producing JPEG images and I would like to add that information in a compatible way with bioformats.

Thanks,
Eduardo

dgault:
Hi Eduardo,

Im afraid currently Bio-Formats does not read physical sizes for JPEG. As you can see from this PR it is something which we had tried to address previously but failed to find a simple solution for it. Do you have the option to produce any other format other than JPEG such as TIFF?

eduramiba:
Hi David,
Thank you for your response.

I tried to use other formats, but unfortunately tiff produces very large files for me compared to jpeg. For example several megabytes tiff against a few hundred kilobytes jpeg. Are there other supported formats that can be lossy or better compressed?

Are XResolution and YResolution not suitable for some reason?

Regards,
Eduardo

sbesson:
Hi @eduramiba,

regarding the consideration about file size, the TIFF layout supports a number of lossy or non-lossy compression schemes. Testing the conversion of a 2048x2048 single plane using a few of these schemes:

```
$ bfconvert "test&sizeX=2048&sizeY=2048.fake" test.jpg
$ bfconvert "test&sizeX=2048&sizeY=2048.fake" test.tiff
$ bfconvert "test&sizeX=2048&sizeY=2048.fake" test-lzw.tiff -compression LZW
$ bfconvert "test&sizeX=2048&sizeY=2048.fake" test-zlib.tiff -compression zlib
$ bfconvert "test&sizeX=2048&sizeY=2048.fake" test-jpeg.tiff -compression JPEG
$ bfconvert "test&sizeX=2048&sizeY=2048.fake" test-jpeg2000.tiff -compression JPEG-2000
```

gives the following sizes

```
sbesson@ls30630:tiff-compression $ ls -alh
total 19088
drwxr-xr-x   8 sbesson  wheel   256B 18 Aug 10:15 .
drwxrwxrwt  32 root     wheel   1.0K 18 Aug 10:29 ..
-rw-r--r--   1 sbesson  wheel   1.6M 18 Aug 10:15 test-jpeg.tiff
-rw-r--r--   1 sbesson  wheel   802K 18 Aug 10:15 test-jpeg2000.tiff
-rw-r--r--   1 sbesson  wheel   2.2M 18 Aug 10:15 test-lzw.tiff
-rw-r--r--   1 sbesson  wheel   608K 18 Aug 10:15 test-zlib.tiff
-rw-r--r--   1 sbesson  wheel   122K 18 Aug 10:14 test.jpg
-rw-r--r--   1 sbesson  wheel   4.0M 18 Aug 10:14 test.tiff
```

eduramiba:
Thanks @sbesson I will try it.

As you can see the smallest tiff file is 5 times as large as the normal jpg file. That's too much for just adding phyisical distance metadata.

And other lossy tiff files such as test-jpeg.tiff are even 13 times larger!

sbesson:
Agreed re TIFF with JPEG compression. I think it boils down to https://github.com/ome/bioformats/issues/3370 i.e. the Bio-Formats code currently does not allow to control the compression quality which is set to the maximum value.

Back to your original question about supporting `XResolution`/`YResolution`, unfortunately the original testing of this PR included some false positives related to concurrent modifications made at the time. Populating the metadata correctly  from the EXIF tags looks like the correct approach but we will need to review the impact on our QA samples to make sure nothing is lost.

I think the two options to help you move forward are:

- test the creation of TIFF files using JPEG compression with low quality using either some third party software or Bio-Formats after addressing #3370
- resurrect this work especially now the dependencies like `metadata-extractor` have been already upgraded.

Do you have any prefenrec and optionally some capacity for driving either of these options?

eduramiba:
Thanks @sbesson I will take a look into that issue.
I am using `metadata-extractor` to read JPEGs metadata in my backend server to extract information of the produced images.

mtbc:
With images that compress well to JPEG I've often found zlib-compressed TIFF to be rather smaller if I use fairly large tiles instead of the BF default strips.

sbesson:
Thanks @mtbc, I completely forgot about setting tile sizes. Extending my previous example

```
bfconvert "test&sizeX=2048&sizeY=2048.fake" test-zlib-512x512.tiff -compression zlib -tilex 512 -tiley 512
bfconvert "test&sizeX=2048&sizeY=2048.fake" test-zlib-1024x1024.tiff -compression zlib -tilex 1024 -tiley 1024
bfconvert "test&sizeX=2048&sizeY=2048.fake" test-zlib.tiff -compression zlib 
``` 
 produces

```
sbesson@ls30630:tmp $ ls -alh test-zlib*
-rw-r--r--  1 sbesson  wheel    18K 19 Aug 11:12 test-zlib-1024x1024.tiff
-rw-r--r--  1 sbesson  wheel    22K 19 Aug 11:12 test-zlib-512x512.tiff
-rw-r--r--  1 sbesson  wheel   608K 19 Aug 11:12 test-zlib.tiff
```

PR #2474, opened 2016-06-28
Windows header portability checks and fixes

rleigh-dundee:
See: https://github.com/ome/ome-common-cpp/pull/22


mtbc:
Diff looks good to me, see when CI is happy.

Issue #2308, opened 2016-03-22
[bug] Micromanager: reader crashes with stack format and large datasets


julou:
Hello,

Following [your fixes to parse config-specific metadata in MM datasets](https://github.com/openmicroscopy/bioformats/pull/2226), I realised that I focused on providing small datasets (for obvious practical reasons). Hence the current MM reader (5.1.8) crashes when the dataset is split over several tiff files.

In more details:
- when MM datasets are stored in stack format and the dataset size exceeds 4.7gb, it's split over several files (e.g. prefix_1_MMStack.ome.tif, prefix_1_MMStack_1.ome.tif, etc)
- because MM can save one tif file (or set of tif files if > 4.7gb) per position, it is sometimes useful to use BF to open PosXX (in fact it would be my primary use case!)
- with BF 5.1.7, the following readers were used depending on the input path:
    - first tif file of the dataset: OME_TIF reader
    - any other tif file: TIF reader
- with current BF 5.1.8, the following readers are used depending on the input path:
    - first tif file of the dataset: OME_TIF reader
    - any other tif file with associated *_metadata.txt file: MM reader (OK provided the position dataset isn't >4.7gb)
    - any other tif file (without *_metadata.txt file): TIF reader

I would be happy to provide you with a demo large dataset, or you can easily create one using MM's demo config and the Multi Dimensional Acquisition window (select e.g. enough time points so that the dataset size is >4.7gb).

Also, as far as I understand it, it means that there is no easy way to open the first position of a dataset (because when this path is given as input, the OME-TIF reader is triggered which will load the full dataset). Although I don't understand how the reader selection works (but that you want to keep a way to trigger the OME-TIF reader for such datasets), I find this very cumbersome! What would be the other possible designs?

Thank you in advance for your attention. Best,
Thomas

sbesson:
Hi @julou ,

many thanks for reporting this issue. We are currently investigating it together with @bramalingam following the workflow you recommended above i.e. creating a large multi-dimensional acquisition using the demo config of MM 1.4.22.

This initial investigation raised a couple of questions regarding your issue:
- we assume you are using a recent version of Micro-Manager, i.e. 1.4.22 or later, is this correct?
- which tool are you using to open the images with Bio-Formats: ImageJ/Fiji or something else?
[...]

PR #2475, opened 2016-06-30
Micromanager: fix multi-file image stacks and dimensionorder

sbesson:
Fixes https://github.com/openmicroscopy/bioformats/issues/2308

This Pull Request fixes the `MicroManagerReader` reader to handle large datasets saved as image stack and split over multiple files (i.e. datasets larger than 4GB). While testing a sample dataset, a second issue was fixed by improving the way each image dimensionality is parsed from the metadata file.

### Summary of changes

- in `parsePosition()`, the number of IFDs were added to the current plane at the end of each loop as well as in the call to `parseKeyValue()` causing an `ArrayIndexOutOfBoundsException` when dealing with multiple files
- the `dimensionOrder` was previously unconditionally set to `XYZCT` causing wrong files to be returned by `Position.getFile()` under some conditions. For the sample multi-file dataset, this caused `parsePosition()` to scan twice the IFDs of the same file and `parseKeyValue()` to fail with an  `ArrayIndexOutOfBoundsException`. Instead we make use of the `SlicesFirst` metadata value as documented in https://micro-manager.org/wiki/Micro-Manager_File_Formats#Writing_Images to store the correct `DimensionOrder`

### Testing

In addition to the existing datasets available under `curated/micromanager`, a new large fileset was generated using Micro-Manager 1.4.22 with the demo config with 3 channels, 21 z-sections and 140 timepoints composed of 2 OME-TIFF files and a metadata text file. This fileset will be migrated to the `public` folder and be available under http://downloads.openmicroscopy.org/images/Micro-Manager/1.4.22/.

This fileset should fail to open without this PR and open properly with this PR included with the dimensions indicated above (might require to bump the default memory if using the command-line tools). Additionally, the OMERO workflow should be tested as the fileset should be importable and viewable.

The configuration files for a couple of filesets will be updated separately to match the `DimensionOrder` changes described above. Some attention should be put on `curated/micromanager/nico/Untitled_4_multi_file/` where the `Plane` metadata was previously incorrectly populated and should now be fixed by this change. The expected deltaT and position value can be checked using Micro-Manager.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -565,7 +565,6 @@ public class MicromanagerReader extends FormatReader {
     // using key alone will result in conflicts with metadata.txt values
     for (int i=plane; i<plane+nPlanes; i++) {
       addSeriesMeta(String.format("Plane #%0" + digits + "d %s", i, key), value);
-      if (i >= p.positions.length) continue;
       if (key.equals("XPositionUm")) {
         try {
           p.positions[i][0] = new Double(value);
@@ -718,6 +717,13 @@ public class MicromanagerReader extends FormatReader {
         }
         else if (key.equals("Slices")) {
           ms.sizeZ = Integer.parseInt(value);
+        }
+        else if (key.equals("SlicesFirst")) {
+          if (value.equals("false")) {
+            ms.dimensionOrder = "XYCZT";
+          } else {
+            ms.dimensionOrder = "XYZCT";
+          }
         }
         else if (key.equals("PixelSize_um")) {
           p.pixelSize = new Double(value);
@@ -910,7 +916,7 @@ public class MicromanagerReader extends FormatReader {
     if (getSizeZ() == 0) ms.sizeZ = 1;
     if (getSizeT() == 0) ms.sizeT = 1;
 
-    ms.dimensionOrder = "XYZCT";
+    if (ms.dimensionOrder == null) ms.dimensionOrder = "XYZCT";
     ms.interleaved = false;
     ms.rgb = false;
     ms.littleEndian = false;
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -541,7 +541,7 @@ public class MicromanagerReader extends FormatReader {
                 }
               }
               if (!value.equals("PropVal")) {
-                parseKeyAndValue(key, value, digits, (plane * nIFDs) + i, 1);
+                parseKeyAndValue(key, value, digits, plane + i, 1);
               }
               propType = null;
               key = null;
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
```diff
@@ -718,6 +718,13 @@ else if (key.equals("Frames")) {
         else if (key.equals("Slices")) {
           ms.sizeZ = Integer.parseInt(value);
         }
+        else if (key.equals("SlicesFirst")) {
+          if (value.equals("false")) {
+            ms.dimensionOrder = "XYCZT";
+          } else {
+            ms.dimensionOrder = "XYZCT";
+          }
```
dgault: This conditional statement seems to be setting the same dimension order either way

dgault:
The code changes here look fine and I can reproduce and see the fixes in the plane metadata for incorrect ordering. 

I am unable to reproduce the ArrayIndexOutOfBoundsException though when using the dataset from http://downloads.openmicroscopy.org/images/Micro-Manager/1.4.22/sebastien/ or samples from curated. Am I testing with the wrong filesets?

sbesson:
@dgault: to reproduce the exception, you could either use the 5.1.10 command-line tools (you might have to up `BF_MAX_MEM`) or try to import the dataset into OMERO without this PR included. Since this is a MM image file stack, you need to point at the `*_metadata.txt` file from the dataset to pick the correct reader, e.g. 

```
$ BF_MAX_MEM=4g ~/bftools/showinf -nopix /ome/data_repo/public/Micro-Manager/1.4.22/sebastien/Test_Dataset_MultiPosition_Stack_MMStack_Pos0_metadata.txt 
Checking file format [Micro-Manager]
Initializing reader
MicromanagerReader initializing /ome/data_repo/public/Micro-Manager/1.4.22/sebastien/Test_Dataset_MultiPosition_Stack_1_MMStack_Pos0_metadata.txt
Reading metadata file
Populating metadata
Finding image file names
Building list of TIFFs
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 66015625
	at loci.formats.in.MicromanagerReader.parseKeyAndValue(MicromanagerReader.java:576)
	at loci.formats.in.MicromanagerReader.parsePosition(MicromanagerReader.java:544)
	at loci.formats.in.MicromanagerReader.initFile(MicromanagerReader.java:298)
	at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:651)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1028)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1117)
```

dgault:
Ah that makes sense now, I wasn't pointing to the txt file. I can now reproduce the ArrayIndexOutOfBoundsException on 5.1.10 and verify that it is no longer present with this PR in place.

I've also retested the Plane metadata on curated/micromanager/nico/Untitled_4_multi_file and can confirm that it has been updated correctly. Lastly the images also display as expected. 

PR should be good to merge.

julou:
hi sebastien,
can I test this by adding the OME repository in fiji's downloader? or how to get a preview of BF5.2?

thanks,
Thomas

dgault:
Hi Thomas,

The Fiji updater at the moment only supports 5.1 builds. If you wish to test this you can download the latest 5.2 build from http://downloads.openmicroscopy.org/bio-formats/5.2.0-m5/ 

To update Fiji you will need to replace the jars in the folder jars/bioformats and the bio-formats_plugins.jar under the plugins directory.

Note: The 5.2 builds are still in internal development and are not fully supported 

David

PR #2477, opened 2016-07-06
xsd-fu - MetadataConverter validation

dgault:
This PR adds, during code generation, a comparison of the properties used in MetadataStore, MetadataRetrieve and MetadataConverter.

Any discrepancies between these files will list the properties missing during the output of xsd-fu

To test:
- Run the code generation either directly from xsd-fu or through other means eg msv generate-sources
- Verify that the output lists validation differences (can search for the string MetadataValidation)
- Currently the ShapeType and LightSourceType are the only properties returned

### DIFF
--- a/components/xsd-fu/xsd-fu
+++ b/components/xsd-fu/xsd-fu
@@ -637,6 +637,8 @@ def metadataMain(model, opts):
                 os.path.join(relpath, "DummyMetadata"),
                 model, opts)
 
+    validateMetadataConverter()
+
 def docGenMain(model, opts):
     """
     Documentation main() that does not parse the command line and spits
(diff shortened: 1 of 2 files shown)


sbesson:
Note from https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-generated-sources/282


```
IOError: [Errno 2] No such file or directory: u'/opt/hudson/workspace/BIOFORMATS-DEV-merge-generated-sources/cpp/generated/cpp/lib/ome/xml/meta/MetadataStore.h'
```

dgault:
Well that location looks like it should be correct and the workspace has the correct directory but no generated files. Im guessing it does a dryrun of the cpp generation. Added in an option to skip the validation if the dryrun option is selected

dgault:
Results appear to be as expected from the latest generated sources build https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-generated-sources/284/consoleFull

     [exec] MetadataValidation: LightSourceType is present in MetadataRetrieve but missing from MetadataStore
     [exec] MetadataValidation: ShapeType is present in MetadataRetrieve but missing from MetadataStore
     [exec] MetadataValidation: LightSourceType is present in MetadataConverter but missing from MetadataStore
     [exec] MetadataValidation: ShapeType is present in MetadataConverter but missing from MetadataStore

sbesson:
Looks good and a nice way to include validation as part of the build. A couple of comments:
- the `cpp` part shows a lot of extra missing attributes which seem to be in `Convert.h`  
```
+ make gensrc
Scanning dependencies of target gensrc
[  0%] Generating meta/MetadataStore.h, meta/MetadataRetrieve.h, meta/AggregateMetadata.h, meta/FilterMetadata.h, meta/DummyMetadata.h, meta/OMEXMLMetadata.h
MetadataValidation: LightSourceType is present in MetadataRetrieve but missing from MetadataStore
MetadataValidation: ShapeType is present in MetadataRetrieve but missing from MetadataStore
MetadataValidation: DatasetAnnotationRef is present in MetadataStore but missing from MetadataConverter
MetadataValidation: MicrobeamManipulationLightSourceSettingsWavelength is present in MetadataStore but missing from MetadataConverter
...
```
- the `ShapeType`, `LightSourceType` type attributes are expected since these are not attributes like `ExperimentType`, immediately we might want to whitelist them in the metadata validation
- once the validation issues are fixed, should we consider failing the build if the validation step fails?
- in terms of cleanup, is there any reason to keep https://github.com/openmicroscopy/bioformats/blob/develop/cpp/test/ome-xml/scripts/metadata-convert-coverage.cmake.in or can this be dropped in favor of this `xsd-fu` addition @rleigh-dundee ?

components/xsd-fu/xsd-fu
```diff
@@ -694,6 +697,43 @@ def parentlist(model, name, pre):
...
+    retrieveDiff = storeSet.symmetric_difference(set(retrieveSet))
+    converterDiff = storeSet.symmetric_difference(set(convertSet))
+    for match in retrieveDiff:
+        if match in retrieveSet:
+            print("MetadataValidation: %s is present in MetadataRetrieve but missing from MetadataStore" % match)
+        else:
+            print("MetadataValidation: %s is present in MetadataStore but missing from MetadataRetrieve" % match)
+    for match in converterDiff:
+        if match in convertSet:
+            print("MetadataValidation: %s is present in MetadataConverter but missing from MetadataStore" % match)
```
sbesson: Could reuse `opts.lang.getConverterName()` as `MetadataConverter` is Java only

sbesson:
Last commit looks good. Only remaining bits are the false positives related to the grepping of `Convert.h`  and the `ShapeType`/`LightSourceType` attributes.

sbesson:
Looking at https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-generated-sources/314, the white listing works as expected with no reference to `LightSourceType/ShapeType`. The only point to clarify is the `Convert.h` and the potential redundancy with the C++ test script which could be dropped in favor of this solution. This can happen on @rleigh-dundee's return and be addressed in a future PR. Merging.

PR #2479, opened 2016-07-07
Tiffcomment: add note about ImageDescription tag

sbesson:
See https://trello.com/c/pqx8FSLk/160-can-comment-ome-tiff-but-not-plain-tiff for reference

This PR adds a note to the Sphinx documentation page and the `tiffcomment` usage to mention the `ImageDescription` tag requirement in the TIFF file.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/TiffComment.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/TiffComment.java
@@ -61,7 +61,7 @@ public class TiffComment {
       System.out.println();
 
       System.out.println("This tool requires an ImageDescription tag to be " +
-        "in place in the TIFF file. ");
+        "present in the TIFF file. ");
       System.out.println();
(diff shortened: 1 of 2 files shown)


mtbc:
Do there happen to be any third-party utilities that will add an `ImageDescription` tag to one's TIFF or if there isn't a preexisting one are people basically screwed? (If yes, might be nice to mention; if not, doesn't block this PR which improves the documentation anyway.)

rleigh-dundee:
I would use `present` rather than `in place` in all these descriptions.  Otherwise, it all looks good.

@mtbc You are basically screwed.  In theory an entirely new IFD could be written, but that's a lot more work than adjusting a single offset.

sbesson:
Pushed terminology change suggested above.

rleigh-dundee:
Looks fine and good to merge, thanks.

PR #2480, opened 2016-07-07
Handle NaN values in FormatTools.getStagePosition()

sbesson:
Add corresponding unit tests checking the returned length is null

Following up of the discussion in https://github.com/openmicroscopy/bioformats/pull/2448#discussion_r69686940

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1452,7 +1452,7 @@ public final class FormatTools {
    *               or {@code unit} is {@code null}.
    */
   public static Length getStagePosition(Double value, Unit<Length> unit) {
-    if (value == null || value.isInfinite()) {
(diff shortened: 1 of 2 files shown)


mtbc:
Perhaps we need corresponding `minExclusive`, `maxExclusive` in the XSD to also exclude such from the OME-XML.

dgault:
Im happy with the handling of infinity and Nan from https://github.com/openmicroscopy/bioformats/pull/2448 and this PR. All the FormatTools unit tests are passing and green.

I do also like the proposal from @mtbc to limit the fields in the xsd in a follow up PR in the future

sbesson:
Merging for now and thumbs up for considering the enforcement at the model level. Would that require defining new primitives of the like of `PositiveFloat`?

dgault:
Yeah we have a few other primitives such as PositiveFloat, though it is the only one making use of minExclusive. I don't believe there would be much required at the code gen level as we should be able to handle the addition ok.

Issue #2467, opened 2016-06-22
Bug in CZI Lightsheet Reader


StephanPreibisch:
Hi,

I realised a pretty significant bug in the Zeiss Lightsheet CZI reader. It does read the sizes of the individual stacks wrongly (assumes the biggest stack size for all of them), which is a huge problem as we need to correct it by hand for every acquisition, otherwise post-processing fails:

'Information|Image|V|View|SizeZ #1': 78
WRONG, this view really has 73 slices

'Information|Image|V|View|SizeZ #2': 78
CORRECT, this view has 78 slices

'Information|Image|V|View|SizeZ #3': 78
WRONG, this view really has 77 slices

'Information|Image|V|View|SizeZ #4': 78
WRONG, this view really has 54 slices

It just shows the maximum number of slices of all views, not the correct amount per view. I uploaded the corresponding dataset here: http://preibischlab.mdc-berlin.de/download/dros-3.czi (1.7GB)

It has 4 angles (0,90,180,270), two channels (488 & 561), one illumination direction and one timepoint.

Please let me know if I can do anything else to help fix this.

All the best from Berlin,
Stephan

StephanPreibisch:
Hi, I forgot to mention, that this did work before and broke at some point. This is really quite a big problem for analysing the data, do you have any idea when you have time to look into it?

If it is not possible in the near future I would start implementing a work-around to detect zero planes in the opened stacks to correct. Please let me know. Thanks a lot, Stephan

dgault:
Hi Stephan,

First off my apologies for the long delay in responding to this issue and thank you for making available a sample file to test with.

I will begin investigating this today to see what possible solutions we have and follow up with you with my findings.

With Thanks,
David Gault

StephanPreibisch:
Great, thanks a lot!

dgault:
Hi Stephan,

Just as a follow up to this issue, I have been working on a solution which appears to resolve these issues. This will have to go through further testing and a PR opened and reviewed before it can be scheduled for any future releases. Once I have the Pull Request opened I shall let you know so you can keep track of the fix.

With Thanks,
David Gault

StephanPreibisch:
Thanks so much David, we really need this fix urgently. If I can help somehow please let me know.

All the best, Stephan

PR #2481, opened 2016-07-07
Zeiss CZI: add support for slides

melissalinkert:
See https://trello.com/c/OHKk0BiI/161-jpeg-xr.

This adds more polished support for slide data from Axio Scan.Z1 systems.  Slides should now be detected as a complete full-resolution image (instead of each tile being a separate series), pyramid sub-resolutions are detected, and label/overview images are detected.

This is excluded for now, pending conclusion of m5 and a configuration PR.  Further testing instructions will be added once the configuration PR is set - I'm just in the process of importing the 13 files for which configurations have changed, verifying that the new behavior is correct, and assembling supporting screenshots.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -951,6 +951,8 @@ public class ZeissCZIReader extends FormatReader {
       name = imageName;
     }
 
+    store.setInstrumentID(MetadataTools.createLSID("Instrument", 0), 0);
+
     int indexLength = String.valueOf(getSeriesCount()).length();
     for (int i=0; i<getSeriesCount(); i++) {
       store.setImageInstrumentRef(MetadataTools.createLSID("Instrument", 0), i);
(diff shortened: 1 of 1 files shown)


melissalinkert:
Additional commits have been added to reduce the number of existing files affected by this PR to just three.  A configuration PR is now open (https://github.com/openmicroscopy/data_repo_config/pull/117), so removing the exclude label.

The list of files to test and corresponding screenshots in Zen are in the config PR - double-checking the configuration changes and comparing ImageJ vs. Zen should be sufficient.

dgault:
The code changes here make sense and look good. The slide data now appears to be handled and stored correctly. Additionally this also allows enables us to fix other issues such as https://github.com/openmicroscopy/bioformats/issues/2467 by enabling use of coreIndex as opposed to seriesIndex.

Ive tested comparisons with affected files between the Zen screenshots provided and importing through Bio-Formats via Fiji. The results of the testing can be found on the associated configuration PR. 

With all tests passing both this and the configuration PR look good to merge.

sbesson:
Thanks, merging this and the configuration PR. In terms of documentation, we should be able to add `pyramid = yes` to the `format-pages.txt` file. Is there anything else that needs updating.

PR #2482, opened 2016-07-08
Update version history for m5

hflynn:
See https://trello.com/c/9mgNoo67/21-version-history-prs

Staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/about/whats-new.html

m4 content is edited as plan is to build up content but end up with only 1 version history entry for 5.2.0.

### DIFF
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -1,13 +1,14 @@
 Version history
 ===============
 
-5.2.0-m4 (2016 June 24)
+5.2.0-m5 (2016 July 7)
 -----------------------
 
 Top-level Bio-Formats API changes:
 
 * Java 1.7 is now the minimum supported version
 * the native-lib-loader dependency has been bumped to version 2.1.4
+* the xalan dependency has been bumped to version 2.7.2
 * all the ome.jxr classes have been deprecated to make clear that there is no
   JPEG-XR support implemented in Bio-Formats as yet
 * the DataTools API has been extended to add a number of utility functions to:
@@ -18,13 +19,18 @@ Top-level Bio-Formats API changes:
   (log4j/logback) initialized via a binding-specific configurations file and
   to prevent `DebugTools.enableLogging(String)` from overriding initialized
   logger levels (see :doc:`/developers/logging` for more information)
+* helper methods have been added to FormatTools allowing a stage position to
+  be formatted from an input Double and an input unit
+* the Formats API has also been updated to add a new validate property to
+  MetadataOptions and support for MetadataOptions has been moved to
+  FormatHandler level to allow it to be used by both Readers and Writers
 
 The Data Model version 2016-06 has been released to introduce
 `Folders <http://blog.openmicroscopy.org/data-model/future-plans/2016/05/23/folders-upcoming/>`_,
 and to simplify both the graphical aspects of the model and code generation.
-Full details will be available in the
-:model_doc:`OME Model and Formats Documentation <>` shortly. OME-XML changes
-include:
+Full details are available in the
+:model_doc:`OME Model and Formats Documentation <schemas/june-2016.html>`.
+OME-XML changes include:
 
 * `Map` is now a complexType rather than an element and `MapPairs` has been
   dropped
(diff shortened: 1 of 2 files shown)


docs/sphinx/about/whats-new.txt
```diff
@@ -96,6 +102,11 @@ Other general improvements include:
 * fixes for the detection of csv pattern blocks by `FilePatternBlock`
 * bioformats_package.jar now includes bio-formats-tools as a dependency so
   ImageConverter, ImageFaker and ImageInfo classes are included in the bundle
+* initial work on
+  `Reader discoverability <https://github.com/openmicroscopy/design/issues/42>`_ now allows
+  optional/external Readers to be annotated into the Readers class and
```
mtbc: "annotated into" is a strange phrase

hflynn: I took the wording from https://github.com/openmicroscopy/bioformats/pull/2457 - happy to receive an alternative suggestion

hflynn: Would be make sense if I added 'using a key/value pair' or 'Map annotation'?

mtbc: Not really. :smiley: (I'd love to be able to say something more helpful here but I'm too baffled; I'm happy to be outvoted if other readers don't share my confusion!)

sbesson: maybe something along the lines `initial work on Reader discoverability now allows to annotate readers in the `readers.txt` configuration file using key/value pairs`?
I will have a go at the external reader use case in a few weeks so that could be described separately for the next milestone.

hflynn:
Freudian slip typo aside, hopefully that reads better now!

docs/sphinx/about/whats-new.txt
```diff
@@ -110,6 +120,9 @@ Other general improvements include:
    - upgrade check no longer run when passing -version
    - common methods refactoring
    - showinf improvements to preload format
+* Micromanager fixes:
```
sbesson: Micro-Manager or µManager

mtbc:
@hflynn: Much better, thank you.

docs/sphinx/about/whats-new.txt
```diff
@@ -110,6 +120,9 @@ Other general improvements include:
    - upgrade check no longer run when passing -version
    - common methods refactoring
    - showinf improvements to preload format
+* Micro-Manager fixes:
```
sbesson: :+1: but should this go under the rest of the format fixes section?

hflynn: D'oh! Will push a fix now

dgault:
These changes all look correct to me. PR should be good to merge.

docs/sphinx/about/whats-new.txt
```diff
@@ -1,13 +1,14 @@
...
 
-5.2.0-m4 (2016 June 24)
+5.2.0-m5 (2016 July 7)
 -----------------------
 
 Top-level Bio-Formats API changes:
 
 * Java 1.7 is now the minimum supported version
 * the native-lib-loader dependency has been bumped to version 2.1.4
+* the xalan dependency has been bumped to version 2.7.2
```
sbesson: @ctrueden: `xalan/serializer 2.7.2` and `native-lib-loader 2.1.4` should be the only non-OME dependency version bumps in Bio-Formats 5.2.0 due in a few days. Is there a specific process on the Fiji deployment side or should we coordinate before pushing the new Bio-Formats JARs?

ctrueden: Fiji already ships `native-lib-loader` version `2.1.4`. And it does not ship `serializer` (see [scijava/pom-scijava@0814257](08142570262d44036882e4f730030b861ee63a9c) for details).

One question: does BF 5.2.0 still support Java 6? If not, then please do not upload it to the Fiji update site, but rather to the Java-8 update site. I have added `Linkert` and `Sbesson` wiki users to the list of accounts which may upload to the Java-8 site.

The only other action item is to bump the `bio-formats.version` constant in pom-scijava. You can file a PR—otherwise I will take care of it later.

mtbc: BF 5.2.0 requires Java 7.

sbesson: @ctrueden: thanks for the quick answer. As @mtbc said, Java 6 support is dropped in Bio-Formats 5.2.0. I will update the release instructions to use the correct update site and we can open a PR directly in `pom-scijava`.

Does this mean Bio-Formats 5.2.x will be conditionally available in Fiji depending on the activation of the Java-8 update site and is there any way to mark this condition if so? /cc @hflynn

ctrueden: Yes, old Fiji installations which originally shipped with Java 6, and have not been updated to use Java 8 instead, will now be "dead ended" at the final Bio-Formats 5.1.x release, since that is what is currently available from the Fiji update site.

However, current (since early 2016) downloads of Fiji ship with a Java 8 JRE, and the Java-8 update site enabled by default. So these new installations will receive Bio-Formats 5.2.0 automatically.

See also http://imagej.net/2016-05-10_-_ImageJ_HOWTO_-_Java_8,_Java_6,_Java_3D

Eventually, the plan is to transition everything from Java-8 back onto the ImageJ and Fiji update sites. The prerequisite for doing that is to update the launch mechanism to use a minimal Java-6-compatible stub, which checks the Java version, and complains if it is too old. Once that is in place, we can switch everything back, so that old installations report the need to upgrade to the Java version rather than silently failing to start.

PR #2486, opened 2016-07-17
MemoizerTest: check return value of setWritable(false)

melissalinkert:
This prevents asserts from being called when the directory permissions
haven't changed, as that will guarantee that the memo permissions tests
will fail.

Following discussion with @aleksandra-tarkowska, some Docker builds fail the 3 ```testGetMemoFilePermissions*``` tests in ```loci.formats.utests.MemoizerTest``` (see https://gist.github.com/aleksandra-tarkowska/c12d83c769b536a5888636180a51ca7f#file-bf-log-L2846).  While it be good to understand why the directory permissions can't be changed in that case, this PR should allow the tests to pass in the meantime and perhaps will help to track down the underlying problem.

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/MemoizerTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/MemoizerTest.java
@@ -279,8 +279,10 @@ public class MemoizerTest {
       // Check existing non-writeable memo directory returns null
       if (File.separator.equals("/")) {
         // File.setWritable() does not work properly on Windows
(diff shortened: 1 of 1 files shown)


aleksandra-tarkowska:
so sorry, my docker files was executing wrong comamnd: `mvn` instead of `mvn install -DskipTests`

aleksandra-tarkowska:
after adding user (uid 1000) all is passing, looks like defult maven container using root is not suitable then volume is mounted. All good

PR #2487, opened 2016-07-18
cpp: LICENSE: Add BSD licence text

rleigh-dundee:
See https://github.com/ome/ome-common-cpp/pull/23


mtbc:
So we have this BSD license in this `cpp` directory but the GPL in a parent directory; it's sufficiently understood that this overrides that for this bit?

rleigh-dundee:
This is duplicating what's already done for `formats-bsd`, so is following existing practice.  It will hopefully be resolved when we split the model out of bioformats.git.

mtbc:
:+1: Fair enough.

mtbc:
This now looks good to me. :+1:

PR #2493, opened 2016-07-24
Added .nii.gz support to NiftiReader...

ericbarnhill:
… Also ended the switch block with a `FormatException` if the suffix is not identified, to better handle future issues with reading NIfTI.

Using `mvn install` to build, the resulting `formats-gpl-5.2.0-SNAPSHOT.jar`, with the patched `NiftiReader`, was tested by dragging and dropping into `Fiji.app/jars/bio-formats` and deleting the earlier version, `formats-gpl-5.1.10.jar`. Fiji was then able to read .nii.gz files directly so the patch works. It was also tested to make sure it still reads `.nii`, `.hdr` and `.img` files and it does.

Gzipped NIfTI files are arguably the most common NIfTI in fMRI imaging and this PR will enable ImageJ to open them without a problem.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
+import loci.common.GZipHandle;
 import loci.formats.CoreMetadata;
 import loci.formats.FormatException;
 import loci.formats.FormatReader;
@@ -228,6 +229,11 @@ public class NiftiReader extends FormatReader {
     else if (id.endsWith(".nii")) {
       pixelsFilename = id;
       pixelFile = in;
+    } else if (id.endsWith(".nii.gz")) {
+      pixelsFilename = id;
+      pixelFile = new RandomAccessInputStream(new GZipHandle(id));
+    } else {
+    	throw new FormatException("File does not have one of the required NIfTI extensions (.hdr, .nii, .nii.gz)");
     }
 
     in.order(little);


sbesson:
Thanks for this PR @ericbarnhill. It is now included as part of our daily builds and did not seem to cause any regression in our [repository tests](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/252). We will review it and get back to you in the upcoming work days.

bramalingam:
Tested it on files from : http://nifti.nimh.nih.gov/nifti-1/data 
The PR works as expected on both .nii.gz  and .nii files.

sbesson:
Thanks @bramalingam. Can you list the filesets that were tested so that we can add them to our public NIfTi samples tested daily? The [repository tests](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/237/) certainly all passed with this PR included.

Looking at the code changes, there are a couple of other places that will need to be updated as well @ericbarnhill:
- the [contructor](https://github.com/ericbarnhill/bioformats/blob/53639ce0ba658219ebab380e9cfdf48c086951bc/components/formats-gpl/src/loci/formats/in/NiftiReader.java#L85) should also list the `.nii.gz` extension and the `datasetDescription` should be updated accordingly
Anything else you can think of @melissalinkert ?
- the [format-pages.txt](https://github.com/openmicroscopy/bioformats/blob/develop/components/autogen/src/format-pages.txt#L1497) will also need `.nii.gz` to be aded so that the generated format page includes the new extension. Note that `origin/develop` will need to be merged into this PR first as we already modifed this line to add a missing `.nii` in https://github.com/openmicroscopy/bioformats/pull/2500.

bramalingam:
File sets:
ome/data_repo/inbox/NIfTI/avg152T1_LR_nifti.nii
ome/data_repo/inbox/NIfTI/avg152T1_LR_nifti.nii.gz
ome/data_repo/inbox/NIfTI/avg152T1_RL_nifti.nii
ome/data_repo/inbox/NIfTI/filtered_func_data.nii
ome/data_repo/inbox/NIfTI/zstat1.nii

melissalinkert:
I was going to say lines 99 and 105 should be updated, but looks like ```checkSuffix``` already handles that case, so all good apart from what @sbesson has already listed.

rleigh-dundee:
Does the `nii.gz` extension also need adding to the readers.txt list in addition to the constructor?

sbesson:
@ericbarnhill: can you let us know if you have some time to push the changes suggested above? Otherwise happy to merge this PR and open a follow-up PR to propagate the extension addition.

sbesson:
Thanks @ericbarnhill, after running our automated tests, we noticed that a few cases still need to be handled. Replacing this PR by https://github.com/openmicroscopy/bioformats/pull/2508 which is on top of your initial commit.

components/formats-gpl/src/loci/formats/in/NiftiReader.java
```diff
@@ -228,6 +229,11 @@ protected void initFile(String id) throws FormatException, IOException {
     else if (id.endsWith(".nii")) {
       pixelsFilename = id;
       pixelFile = in;
+    } else if (id.endsWith(".nii.gz")) {
+      pixelsFilename = id;
+      pixelFile = new RandomAccessInputStream(new GZipHandle(id));
+    } else {
+    	throw new FormatException("File does not have one of the required NIfTI extensions (.hdr, .nii, .nii.gz)");
```
dgault: extensions list is missing the .img

PR #2494, opened 2016-07-25
Update OME server page

hflynn:
This removed the old LOCI trac ticket link which is making the docs build unstable (see https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/337/warnings3Result/) and while I was at it, I figured I may as well update the 'Upgrading' section as well as there is no point it linking to the current release when we don't support OME server any more and haven't done since 5.0.0.

Staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/users/ome-server/index.html once built

### DIFF
--- a/docs/sphinx/users/ome-server/index.txt
+++ b/docs/sphinx/users/ome-server/index.txt
@@ -94,17 +94,15 @@ To reset things back to how they were:
    name='import_formats';
 
 Lastly, please note that Li-Cor L2D files cannot be imported into an OME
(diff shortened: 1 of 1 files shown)


hflynn:
On staging now, build green again - https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/338/

sbesson:
:+1: for both additions. Thanks.

PR #2495, opened 2016-07-25
Rearrange 5.2 history & mark breaking readers

hflynn:
See https://trello.com/c/NE6UFb5k/70-better-document-breaking-reader-changes - on further thought it seemed best to mark those format fixes which constitute breaking changes in the existing list, something we can easily carry forward for all further releases, rather than add another separate list to repeat info.

I rearranged the page at the same time although maybe we want to put the top level API changes back at the top and just leave the model stuff at the bottom, not sure what other people think - more recent releases have format fixes first but going back to the last major release, they were lower down the list.

### DIFF
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -35,7 +35,7 @@ Java format support improvements include:
    - Zeiss LSM
        - fixed Plane population errors
    - Nifti
-       - fixed planeSize to prevent crashes when loading large files (thanks 
+       - fixed planeSize to prevent crashes when loading large files (thanks
          to Christian Niedworok)
    - LiFlim
        - fixed ExposureTime check and units usage
@@ -54,7 +54,8 @@ Java format support improvements include:
          over multiple files
 
 †Denotes a major breaking change to the reader (typically modification of core
-metadata).
+metadata). Code changes or re-import may be necessary in ImageJ/FIJI and
+OMERO.
 
 Top-level Bio-Formats API changes:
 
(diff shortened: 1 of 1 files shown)


docs/sphinx/about/whats-new.txt
```diff
@@ -98,6 +53,30 @@ Java format support improvements include:
...
+* the native-lib-loader dependency has been bumped to version 2.1.4
+* the xalan dependency has been bumped to version 2.7.2
+* all the ome.jxr classes have been deprecated to make clear that there is no
+  JPEG-XR support implemented in Bio-Formats as yet
+* the DataTools API has been extended to add a number of utility functions to:
+   - account for decimal separators in different locales
+   - parse a String into Double, Float, Integer etc
+   - handle NumberFormatException thrown when parsing Unit tests
+* the Logging API has been updated to respect logging frameworks
+  (log4j/logback) initialized via a binding-specific configurations file and
```
mtbc: should configurations be plural?

docs/sphinx/about/whats-new.txt
```diff
@@ -98,6 +53,30 @@ Java format support improvements include:
        - fixed handling of large datasets saved as image stacks and split
          over multiple files
 
+†Denotes a major breaking change to the reader (typically modification of core
+metadata).
```
joshmoore: "Code changes or re-import may be necessary in ImageJ and OMERO"

dgault:
I agree with the comment from Josh that a line to inform people of what a breaking reader change would actually mean would be useful.

Otherwise the new reordering works well and the content reads fine to me.

hflynn:
Thanks All, fix pushed.

sbesson:
@mtbc @dgault @joshmoore: are you guys fine with the last set of changes currently staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/about/whats-new.html?

mtbc:
Fine with me.

joshmoore:
:+1: but one thought (don't know if it's possible): can the "†Denotes a major breaking change to the reader " come right after "bug fixes for many formats including:" without ruining the page?

hflynn:
I added it manually @joshmoore so yes, I can stick it at the top instead. Will change it now.

dgault:
Everything looks good to me with the latest set of updates

hflynn:
Assume this is good to merge now - probably better to do the final additions for rc1 on a fresh PR (although I can do them here if preferred)

PR #2498, opened 2016-07-26
Add SPC-FIFO public images & support note

hflynn:
See https://trello.com/c/hjZpniE2/166-make-becker-hickl-fifo-spc-files-public
This adds a link to the now public samples and a note regarding what the files are currently supported as suggested by @imunro 

Will be staged at http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/formats/becker-hickl-fifo.html


imunro:
looks good to me

PR #2501, opened 2016-07-27
Re-include OME-XML and OME-TIFF samples when testing valid OME-XML

sbesson:
This PR reverts skips in https://github.com/openmicroscopy/bioformats/commit/dbd5f824f8c581d5f8476f03d5f580b4090ac077 to investigate issues when converting/validating OME-XML /cc @dgault @rleigh-dundee

PR #2502, opened 2016-07-27
Fix for invalid setPixelsBinDataBigEndian

dgault:
This is a fix for the generation of invalid OME XML when files are read through OMETiffReader.
The setPixelsBinDataBigEndian is only called if the PixelsBinData > 0 otherwise setPixelsBigEndian is used instead.

This can be tested using the FormatReaderTests with the following PR included - https://github.com/openmicroscopy/bioformats/pull/2501

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -248,7 +248,14 @@ public class OMETiffReader extends FormatReader {
       }
 
       for (int i=0; i<meta.getImageCount(); i++) {
-        meta.setPixelsBinDataBigEndian(Boolean.TRUE, i, 0);
+        if (meta.getPixelsBinDataCount(i) > 0) {
+          for (int j=0; j<meta.getPixelsBinDataCount(i); j++) {
+            meta.setPixelsBinDataBigEndian(Boolean.TRUE, i, j);
+          }
+        }
+        else {
+          meta.setPixelsBigEndian(Boolean.TRUE, i);
+        }
         MetadataTools.verifyMinimumPopulated(meta, i);
       }
       return meta.getImageCount() > 0;
@@ -1056,7 +1063,14 @@ public class OMETiffReader extends FormatReader {
     int realSeries = getSeries();
     for (int i=0; i<getSeriesCount(); i++) {
       setSeries(i);
-      store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, 0);
+      if (meta.getPixelsBinDataCount(i) > 0) {
+        for (int j=0; j<meta.getPixelsBinDataCount(i); j++) {
+          store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, j);
+        }
+      }
+      else {
+        store.setPixelsBigEndian(Boolean.TRUE, i);
+      }
     }
     setSeries(realSeries);
     return store;
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -248,13 +248,11 @@ public class OMETiffReader extends FormatReader {
       }
 
       for (int i=0; i<meta.getImageCount(); i++) {
+        meta.setPixelsBigEndian(Boolean.TRUE, i);
         if (meta.getPixelsBinDataCount(i) > 0) {
           for (int j=0; j<meta.getPixelsBinDataCount(i); j++) {
             meta.setPixelsBinDataBigEndian(Boolean.TRUE, i, j);
           }
-        }
-        else {
-          meta.setPixelsBigEndian(Boolean.TRUE, i);
         }
         MetadataTools.verifyMinimumPopulated(meta, i);
       }
@@ -1067,9 +1065,6 @@ public class OMETiffReader extends FormatReader {
         for (int j=0; j<meta.getPixelsBinDataCount(i); j++) {
           store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, j);
         }
-      }
-      else {
-        store.setPixelsBigEndian(Boolean.TRUE, i);
       }
     }
     setSeries(realSeries);
--- a/components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
+++ b/components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
@@ -58,7 +58,7 @@ if is_octave()
 else
     java_true = java.lang.Boolean.TRUE;
 end
-metadata.setPixelsBigEndian(java_true, 0, 0);
+metadata.setPixelsBigEndian(java_true, 0);
 
 % Set dimension order
 dimensionOrderEnumHandler = javaObject('ome.xml.model.enums.handlers.DimensionOrderEnumHandler');
--- a/components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
+++ b/components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
@@ -58,7 +58,7 @@ if is_octave()
 else
     java_true = java.lang.Boolean.TRUE;
 end
-metadata.setPixelsBinDataBigEndian(java_true, 0, 0);
+metadata.setPixelsBigEndian(java_true, 0, 0);
 
 % Set dimension order
 dimensionOrderEnumHandler = javaObject('ome.xml.model.enums.handlers.DimensionOrderEnumHandler');
--- a/components/formats-bsd/src/loci/formats/gui/ImageViewer.java
+++ b/components/formats-bsd/src/loci/formats/gui/ImageViewer.java
@@ -379,7 +379,7 @@ public class ImageViewer extends JFrame implements ActionListener,
         omeMeta = omexmlService.createOMEXMLMetadata();
         omeMeta.setImageID(MetadataTools.createLSID("Image", 0), 0);
         omeMeta.setPixelsID(MetadataTools.createLSID("Pixels", 0), 0);
-        omeMeta.setPixelsBinDataBigEndian(false, 0, 0);
+        omeMeta.setPixelsBigEndian(false, 0);
 
         String order = "XYCZT";
         if (in != null) order = in.getDimensionOrder();


components/formats-bsd/src/loci/formats/in/OMETiffReader.java
```diff
@@ -1056,7 +1063,14 @@ public MetadataStore getMetadataStoreForConversion() {
...
     for (int i=0; i<getSeriesCount(); i++) {
       setSeries(i);
-      store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, 0);
+      if (meta.getPixelsBinDataCount(i) > 0) {
+        for (int j=0; j<meta.getPixelsBinDataCount(i); j++) {
+          store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, j);
+        }
+      }
+      else {
+        store.setPixelsBigEndian(Boolean.TRUE, i);
```
rleigh-dundee: I don't think this should be in an else block--I think it should be set unconditionally irrespective of whether any BinData elements are present.  And I think it should be set using `!isLittleEndian()` for consistency and to match what's set in core metadata.

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
```diff
@@ -1056,7 +1063,14 @@ public MetadataStore getMetadataStoreForConversion() {
     int realSeries = getSeries();
     for (int i=0; i<getSeriesCount(); i++) {
       setSeries(i);
-      store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, 0);
+      if (meta.getPixelsBinDataCount(i) > 0) {
+        for (int j=0; j<meta.getPixelsBinDataCount(i); j++) {
+          store.setPixelsBinDataBigEndian(new Boolean(!isLittleEndian()), i, j);
```
rleigh-dundee: If an OME-TIFF did happen to contain BinData, why are we setting this single attribute at this point?  It would be overriding the value provided in the data, which could result in reading corrupted data if it was switched from the original value.  Would it not be better to remove the entire `if` block and ignore it entirely?

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
```diff
@@ -1056,7 +1063,14 @@ public MetadataStore getMetadataStoreForConversion() {
     int realSeries = getSeries();
     for (int i=0; i<getSeriesCount(); i++) {
       setSeries(i);
```
rleigh-dundee: I think this entire function belongs at the end of `initFile`, and should be removed entirely once moved.  Why aren't we setting these values along with the rest of the metadata?  I think this is historical cruft which could be dropped.

rleigh-dundee:
Also note the solution for C++ in https://github.com/ome/ome-files-cpp/pull/26/files#diff-bd3658acf7c3f261bc1815ae159c232c which implements some of the comments above.

sbesson:
From 1378 failures, the [repository subset job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/258) is now down to 21. Some of these failures might require the samples files themselves to be fixed. Marking this PR and #2501 as `breaking` to allow work on them without disturbing the repository jobs for the other PRs.

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
sbesson: From [our specification](http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Pixels_BigEndian0), `Pixels.BigEndian` can be optional and should definitely match the value of `BinData.BigEndian` if the latter exists i.e. `!isLittleEndian()`. No strong objection to setting it unconditionally.

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
sbesson: It looks the code generation improvements forces us to revisit the API specifically added in https://github.com/openmicroscopy/bioformats/commit/de5dbaf9baea1b84ee34909dcc3cdcc518fe2613.

A couple of points that might require clarification in particular:
- is `BinData.BigEndian` still a requirement for converting file formats or can we use the MetadataStore populated by the OMETiffReader directly with an `IFormatWriter`
- similarly is there still a need for a converted OME-XML when displaying in `ImageInfo` especially now the validation happens on read and no longer after conversion with https://github.com/openmicroscopy/bioformats/pull/2456 ?

As this is public API, I would not drop `getMetadataStoreForDisplay()/getMetadataStoreForConversion`() but deprecating them would be sensible if there is no more clear usage for them.

rleigh-dundee: These are technically "public", but are not part of `IFormatReader`--they are specific to `OmeTiffReader` and using them requires hairy casts to get access to them.  `getMetadataStoreForConversion` has zero public use inside bioformats (and omero) and `getMetadataStoreForDisplay` has a single user in `ImageInfo`.  The special mangling for display could be moved into `ImageInfo` or `MetadataTools`.

If any writer is dependent upon `BinData.BigEndian` in the absence of any BinData elements, I would suggest we should fix those writers.  The valid metadata should be usable in all cases, without the need for special-case mangling which makes it invalid.  The fact these methods exist and aren't in `IFormatWriter` is a strong indication that there's something not quite right and that this is a workaround.

If we do retain these methods, I'd suggest they could simply wrap `getMetadataStore` and return the valid store without mangling, and we can then move the logic elsewhere which mangles things for special purposes (or eliminate it entirely).

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
rleigh-dundee: A deeper question for me would be this: will an OME-TIFF ever contain BinData elements?  If it contains TiffData, then it can't since they are mutually exclusive.  When using a metadatastore for writing, the question for me is this: who has the responsibility of cleaning up the store to match what is being written?  If we're adding BinData, that's fine, but whatever is adding it should be adding it so that the store is coherent: remove all the TiffData, and when you add the BinData, add the BinData in full, rather than a single property, so it's still valid.

Should, for example, the OME TIFF writer strip out all BinData, TiffData and MetadataOnly elements prior to writing? Should FormatWriter do this on a deep copy of the store?  Or should each reader, or FormatReader override getMetadataStore to pass a cleaned up copy?

Right now, we seem to be forcing the OmeTiffReader to partially mangle the store, but without any cleanup of TiffData.  Why this reader and not the others?  I feel that having that sanitisation in a more generic place like MetadataTools which would be called by FormatWriter would be a cleaner and more general solution to the problem.

sbesson: As we cannot mix `BinData` and `TiffData` elements, the only valid use case I can think  would be a multi-image OME-TIFF where one image using `TiffData` elements and the second uses `BinData`.
I am not sure whether this is a scenario we ever wanted to support although our format specification does not clearly discard it.

rleigh-dundee: I think that if such an image did exist, it would not be read successfully by OmeTiffReader.  It doesn't support pixel data except if in a plane referred to by TiffData, and we would have mangled the BigEndian attribute based upon the endianness of the TIFF container when we set it on Pixels and then on each BinData element.  So I think it's probably safe to assume that BinData isn't handled properly at all by this reader.

I'd be more interested to know specifically which code is dependent upon *consuming* this BinData.BigEndian attribute, assuming that there is a consumer and this isn't merely historical cruft.

rleigh-dundee:
```
git grep -l getPixelsBinDataBigEndian
components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
components/formats-api/src/loci/formats/MetadataTools.java
components/formats-api/src/loci/formats/meta/MetadataConverter.java
components/formats-bsd/src/loci/formats/gui/AWTImageTools.java
components/formats-bsd/src/loci/formats/gui/BufferedImageWriter.java
components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
components/formats-bsd/src/loci/formats/out/APNGWriter.java
components/formats-bsd/src/loci/formats/out/ICSWriter.java
components/formats-bsd/src/loci/formats/out/JPEG2000Writer.java
components/formats-bsd/src/loci/formats/out/JavaWriter.java
components/formats-bsd/src/loci/formats/out/OMEXMLWriter.java
components/formats-bsd/src/loci/formats/out/TiffWriter.java
components/formats-bsd/src/loci/formats/out/V3DrawWriter.java
components/test-suite/src/loci/tests/testng/FormatReaderTest.java
cpp/lib/ome/xml/meta/Convert.cpp
```

Basically every writer.  But the logic is generally (from `TiffWriter`):

```
    if (retrieve.getPixelsBigEndian(series) != null) {
      littleEndian = !retrieve.getPixelsBigEndian(series).booleanValue();
    }
    else if (retrieve.getPixelsBinDataCount(series) == 0) {
      littleEndian = !retrieve.getPixelsBinDataBigEndian(series, 0).booleanValue();
    }
```

So long as every writer is using `getPixelsBigEndian` and only using `getPixelsBinDataBigEndian` as a fallback, then it looks like it's safe to drop this from `OmeTiffReader`.  Arguably, they should only be using `getPixelsBigEndian`, since it's the reader's responsibility to set this correctly.  The fallback shouldn't be needed in every writer.

dgault:
Changes added based on the discussion from earlier. Kept setPixelsBigEndian for the isType but setting it unconditionally now, and removed it altogether from getMetadataStoreForConversion

sbesson:
As discussed in a call earlier today with @mtbc @rleigh-dundee and @dgault and similarly to https://github.com/ome/ome-files-cpp/pull/26, the proposed next step for 5.2.0 is the full deprecation of the `getMetadataStoreForDisplay/getMetadataStoreForConversion` in favor of using the MetadataStore populated by the reader directly for:
1- displaying the OME-XML validated by the reader in `showinf` (following up validation improvements in https://github.com/openmicroscopy/bioformats/pull/2456) like for other formats
2- passing to an `IFormatWriter` while converting an OME-TIFF

@melissalinkert: can you think of the scenarios where this would be problematic and these conversion API should be retained?

melissalinkert:
The proposal certainly makes sense.

Neither call to ```setPixelsBinDataBigEndian``` in ```OMETiffReader``` should be necessary, as both only existed to fulfill the requirements of ```MetadataTools.verifyMinimumPopulated```.  That should mean that ```getMetadataStoreForConversion``` is a no-op, so deprecation is safe (and line 1029 can be removed?).

If not done already, it might be a good idea to test converting one or two existing OME-TIFF files to each supported output format in both showinf and ImageJ.  I don't see anything immediately suspicious, but as ```getPixelsBinDataBigEndian``` is still widely used as a fall-back in case ```getPixelsBigEndian``` is null, that would help with confidence that nothing is broken by this change.

components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
```diff
@@ -58,7 +58,7 @@
 else
     java_true = java.lang.Boolean.TRUE;
 end
-metadata.setPixelsBinDataBigEndian(java_true, 0, 0);
+metadata.setPixelsBigEndian(java_true, 0, 0);
```
sbesson: Extra `,0` should be removed - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-matlab/442/testReport/

sbesson:
As per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/243/, merging this PR thanks. @dgault: let's coordinate to open one or two follow-up PR to:
- deprecate the `getMetadataFor*` methods in the `OMETiffReader` as done in https://github.com/ome/ome-files-cpp/pull/26/commits/5bc636b474780a75b71dbc7b8c826a42fbf40fbe
- investigate and drop the call to `getMetadataForConversion` as suggested in https://github.com/openmicroscopy/bioformats/pull/2502#issuecomment-235901568 (likely to be tested via an OME-TIFF -> OME-TIFF conversion)
- drop the usage of `getMetadataForDisplay` in `ImageInfo` so that the displayed OME-XML is exactly the read/validated/populated one
- semi-related, in the command-line tools, I would reset the OME-XML validation flag back to true if `--ome-xml` is passed to `showinf` to be in agreement with OME Files C++

PR #2503, opened 2016-07-28
Fix timestamp annotation 2016-06 OME-XML sample

sbesson:
For some reason, an extra `>` got inserted in the sample during the generation making it invalid. Since this sample contains non-POSIX timestamps, it is excluded from the C++ tests and could not be detected. The re-enabling of the XML tests for the OME-XML/OME-TIFF readers in https://github.com/openmicroscopy/bioformats/pull/2501 successfully flagged this file as invalid.

To test this PR check that `xmlvalid` fails on the OME-XML file without this PR but passes with the PR included.


dgault:
Fix looks good and xml validates now. From the failed tests this appears to be the only affected file so this sample should be good to merge.

PR #2504, opened 2016-07-29
update schema link to 2016-06 for publication

mtbc:
Under http://www.openmicroscopy.org/Schemas/ --

> Generated documentation for the schemas is available at: http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2015-01/ome.html

should now be OME-2016-06.


hflynn:
👍

PR #2506, opened 2016-08-01
if offline, skip LocationTest cases that require connectivity

mtbc:
Fix https://trac.openmicroscopy.org/ome/ticket/13272 by allowing Maven tests to pass even when offline. Specifically, check that:

* when online, *all* `LocationTest` methods pass
* when offline, *some* `LocationTest` methods skip, the rest pass
* in both cases, `mvn test` succeeds.

### DIFF
--- a/components/formats-common/test/loci/common/utests/LocationTest.java
+++ b/components/formats-common/test/loci/common/utests/LocationTest.java
@@ -139,8 +139,8 @@ public class LocationTest {
     for (int i=0; i<files.length; i++) {
       skipIfOffline(i);
       String msg = files[i].getName();
-      assertEquals(msg, files[i].canRead(), mode[i].indexOf("r") != -1);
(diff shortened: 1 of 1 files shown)


components/formats-common/test/loci/common/utests/LocationTest.java
```diff
@@ -107,16 +111,36 @@ public void setup() throws IOException {
       "rw", "", "rw", "", "r", "","rw"
     };
 
+    isRemote = new boolean[] {
+      false, false, false, true, true, true, false
+    };
+  }
+
+  @BeforeClass
+  public void checkIfOnline() throws IOException {
```
mtbc: note that the online check does *not* depend on `Location`

sbesson:
Tested locally online and offline with this PR included - got 3 tests skipped when running the tests offline and all of them passed online (as confirmed by Travis). Thanks for tackling this.

PR #2508, opened 2016-08-01
Add nii.gz support to NiFTi reader

sbesson:
Supersedes https://github.com/openmicroscopy/bioformats/pull/2493

This PR adds support for gzipped compressed `.nii.gz` filesets by using  `checkSuffix()` which supports compression extensions rather than `String.endsWith()`. The `RandomAccessInputStream` constructor internally uses `Location.getHandle()` which in turns handles compressed files. 

The new extension is also registered in the reader, `readers.txt` and `format-pages.txt` and the documentation pages are regenerated accordingly.

A data repository PR will be opened to add the configuration for the public `nii.gz` samples available at http://nifti.nimh.nih.gov/nifti-1/data.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
+import loci.common.GZipHandle;
 import loci.formats.CoreMetadata;
 import loci.formats.FormatException;
 import loci.formats.FormatReader;
@@ -228,6 +229,11 @@ public class NiftiReader extends FormatReader {
     else if (id.endsWith(".nii")) {
       pixelsFilename = id;
       pixelFile = in;
+    } else if (id.endsWith(".nii.gz")) {
+      pixelsFilename = id;
+      pixelFile = new RandomAccessInputStream(new GZipHandle(id));
+    } else {
+    	throw new FormatException("File does not have one of the required NIfTI extensions (.hdr, .nii, .nii.gz)");
     }
 
     in.order(little);
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -147,6 +147,7 @@ public class NiftiReader extends FormatReader {
   {
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
+    pixelFile.seek(0);
     long planeSize = FormatTools.getPlaneSize(this);
     pixelFile.seek(pixelOffset + no * planeSize);
     readPlane(pixelFile, x, y, w, h, buf);
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -230,7 +230,7 @@ public class NiftiReader extends FormatReader {
       pixelsFilename = id;
       pixelFile = in;
     } else {
-      throw new FormatException("File does not have one of the required NIfTI extensions (.hdr, .nii, .nii.gz)");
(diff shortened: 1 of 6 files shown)


hflynn:
Doc additions look fine, I'll check the staging pages when this is built tomorrow

melissalinkert:
Closing and reopening ```pixelFile``` in each openBytes call resolves the remaining test failures in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/268/consoleText, which suggests that this exposes an underlying I/O bug (probably in GZipHandle?).  Should be able to investigate further Tuesday afternoon, unless anyone else wants to have a go.

hflynn:
Staging pages like fine thanks

ericbarnhill:
Looks great and sorry if I was sluggish to respond. I wasn't up on everything being discussed, and felt I needed time to sit down and look it over before contributing further.

melissalinkert:
@sbesson: unfortunately not any closer to understanding the root cause of the test failures in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/268/.  There are two workarounds:

- explicitly close and reopen ```pixelFile``` before seeking/reading the buffer in ```openBytes```
- call ```pixelFile.seek(0)``` in ```openBytes``` before seeking to the correct offset

Various things tried to track this down:

- removed ```pixelFile``` variable and used ```in``` everywhere in the Nifti reader (no change)
- checked the file pointers reported in each line of ```openBytes``` (same values reported in passing vs. failing hash tests)
- tried a minimal example of initializing two readers with the same ```.nii.gz``` file, reading the first plane for each and writing each resulting byte array to a separate file (MD5s of output files match)
- removed ```BufferedInputStream``` from the chain of ```InputStreams``` in ```GZipHandle``` in case there was a stale buffer (no change)
- removed calls to ```stream.mark``` in ```StreamHandle```, in case a stale mark position was affecting seeking (no change)
- tested with both OpenJDK 1.7.0_79 and Oracle JDK 1.8.0_91 in case of a difference in how ```GZIPInputStream``` was implemented (no change)
- gzipped a few .bmp files, generated configurations and verified that ```test-automated``` passed

...all of which is to say, I can't reproduce outside of ```test-automated```, and forcing a call to the underlying ```GZipHandle.resetStream()``` is the only thing that has fixed the tests so far.

I'm out of ideas at the moment, and likely won't have much more time to investigate this week - might be necessary to defer a proper fix until 5.2.1 at this point.

components/formats-gpl/src/loci/formats/in/NiftiReader.java
```diff
@@ -225,9 +226,11 @@ protected void initFile(String id) throws FormatException, IOException {
       pixelsFilename = id.substring(0, id.lastIndexOf(".")) + ".img";
       pixelFile = new RandomAccessInputStream(pixelsFilename);
     }
-    else if (id.endsWith(".nii")) {
+    else if (checkSuffix(id, "nii")) {
       pixelsFilename = id;
       pixelFile = in;
+    } else {
+      throw new FormatException("File does not have one of the required NIfTI extensions (.hdr, .nii, .nii.gz)");
```
dgault: Extensions list is missing the .img

dgault:
Having tested this with the public fileset from our inbox (as the http://nifti.nimh.nih.gov/nifti-1/data site appears to be offline) the files all open without any issues and the new extensions are recognized fine.

The one issue I noticed was when comparing the files to when I opened them with FSLViewer they appear to  vertically inverted or upside down. This appears to be a legacy issue though and not related to this PR in particular.

sbesson:
@dgault pushed the extension addition.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/276/console for the testing against the NIfTi samples (including the `.nii.gz`).

dgault:
With the tests green everything looks good for .nii.gz file extensions. PR is good to merge.

PR #2510, opened 2016-08-04
Fixed so odd widths are not increased by one pixel

RichardMyers:
I'm just trying to make a minor fix to the SlideBook6Reader.java file, but it appears that I made a mistake in my pull request.  

 This change is only relevant for the openmicroscopy:master branch because SlideBook6Reader.java is removed in the openmicroscopy:develop branch.


sbesson:
@RichardMyers: the `master` branch of Bio-Formats is not used for development or release and we are not planning any release of Bio-Formats 5.1.x so I will close this PR.
As discussed elsewhere, Bio-Formats 5.2.0 which we expect to release in less than 2 weeks has externalized the SlideBook6Reader. So you should be able to test and commit your fixes directly using the 3i Slidebook JAR file and use your update site for distribution.

Let us know if you need a pre-release version of Bio-Formats 5.2.0 and instructions for testing these changes.

PR #2511, opened 2016-08-05
Quicktime download links & reader clarifications

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/353/warnings3Result/ - Apple have moved their download links under support.apple.com/downloads breaking our links.

In investigating this I discovered the other links we have (in the java-library.txt & pict-mactintosh-picture.txt format page) are referencing Quicktime for Java which seems to have been depreciated in 2009 so I'm not sure how to handle these. @sbesson suggests the reader using it may be essentially legacy anyway, so we could remove the reference from https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/java-library.html altogether, but that still leaves https://www.openmicroscopy.org/site/support/bio-formats5.2/formats/pict-macintosh-picture.html - should this reader become legacy @melissalinkert ?

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -1969,7 +1969,7 @@ reader = QuesantReader
 extensions = .mov
 owner = `Apple Computer <http://www.apple.com/>`_
 bsd = yes
-software = `QuickTime Player <http://www.apple.com/quicktime/download/>`_
+software = `QuickTime Player <https://support.apple.com/downloads/quicktime>`_
 weHave = * a `QuickTime specification document <http://developer.apple.com/documentation/Quicktime/QTFF/>`_ (from 2001 March 1, in HTML) \n
 * several QuickTime datasets \n
 * the ability to produce more datasets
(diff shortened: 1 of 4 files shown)


hflynn:
Removed links, legacy issue can be handled on a different PR

components/autogen/src/format-pages.txt
```diff
@@ -1982,19 +1981,19 @@ metadataRating = Very good
...
 utilityRating = Poor
-reader = NativeQTReader
+reader = NativeQTReader, LegacyQTReader
 writer = QTWriter
 notes = Bio-Formats has two modes of operation for QuickTime: \n
 \n
-* QTJava mode requires `QuickTime <http://www.apple.com/quicktime/download/>`_ to be \n
-  installed (32-bit JVM only, not supported with 64-bit). \n
+* The Legacy QTJava mode requires QuickTime for Java to be installed \n
+  (no longer available from Apple) and is 32-bit JVM only, not supported with 64-bit. \n
```
ximenesuk: The end of this sentence is a bit clunky (as it was originally):

"...and is 32-bit JVM only, not supported with 64-bit."

I assume this means that QuickTime for Java needs a 32-bit JVM and won't run with a 64-bit JVM? If so I wonder if it could be phrased a little better?

docs/sphinx/formats/quicktime-movie.txt
```diff
@@ -58,15 +61,15 @@ Utility: |Poor|
 
 Bio-Formats has two modes of operation for QuickTime: 
 
-* QTJava mode requires `QuickTime <http://www.apple.com/quicktime/download/>`_ to be 
-  installed (32-bit JVM only, not supported with 64-bit). 
+* The Legacy QTJava mode requires QuickTime for Java to be installed 
+  (no longer available from Apple) and is 32-bit JVM only, not supported with 64-bit.
```
ximenesuk: ditto (clearly).

hflynn:
Hopefully that's clearer now @ximenesuk

PR #2513, opened 2016-08-05
Add note that Gatan reader doesn't support stacks

hflynn:
See https://trello.com/c/KMTuzwmG/172-gatan-format-page-stacks

staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/formats/gatan-digital-micrograph.html


ximenesuk:
This looks good to merge.

PR #2514, opened 2016-08-05
DelegateReader - Remove call to super.setId(id);

dgault:
This is a followup to issues spotted in https://github.com/openmicroscopy/bioformats/pull/2473#issuecomment-237785374 which is as a result of https://github.com/openmicroscopy/bioformats/pull/2473

After investigation this was caused by the delegateReader making 2 calls to setID, once using the super.setId and then another to setId on the legacy or native reader. The first call was initialising a blank OMEXML object in AbstractOMEXMLMetadata as part of the dumpXML command. This meant that any future calls to dumpXML were using the blank root object.

As the first call to super is redundant this can be removed. This will allow the delegateReader to initialise with only a single call to setID. It may also be worth investigating if the dumpXML function is required to set the root object.

This issue only impacts calls through the command line tools were the validation is on by default.

### DIFF
--- a/components/formats-api/src/loci/formats/DelegateReader.java
+++ b/components/formats-api/src/loci/formats/DelegateReader.java
@@ -272,16 +272,17 @@ public abstract class DelegateReader extends FormatReader {
   /* @see IFormatHandler#setId(String) */
   @Override
   public void setId(String id) throws FormatException, IOException {
-    super.setId(id);
     if (useLegacy && !nativeReaderInitialized && !legacyReaderInitialized) {
       try {
         legacyReader.setId(id);
         legacyReaderInitialized = true;
+        currentId = legacyReader.getCurrentFile();
       }
       catch (FormatException e) {
         LOGGER.debug("", e);
         nativeReader.setId(id);
         nativeReaderInitialized = true;
+        currentId = nativeReader.getCurrentFile();
       }
     }
     else {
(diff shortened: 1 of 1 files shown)


sbesson:
Various runs of the repository subset jobs against the affected formats ([100](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/100/), [103](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/103), [104](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/104)) seem to indicate this PR works as expected. Only exception is the `poly3.jpg` file ([99](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/99/)) which has tests failing depending of the JDK.

Barring veto, I propose to exclude the last file from the subset covered by automated tests and move this PR to the regular merge builds.

sbesson:
Tested via OMERO by importing a variety of the formats using the delegate reader (i.e. all formats with a legacy/alternate reader) `nd2`, `jpeg`, `quicktime` and `tiff`. Imports proceeded as expected. For each format, `showinf -omexml` now populates the OME-XML metadata as expected. Running Windows automated tests against each of the formats in the data repository.

sbesson:
With https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/254/ green, all looks good. Merging

PR #2515, opened 2016-08-09
ImageConverter: Allow disabling of LUT conversion

ghost:
See https://trac.openmicroscopy.org/ome/ticket/13273

Testing:

Run `bfconvert test.image [-nolookup] test.ome.tiff`.  Run `showinf` and `tiffinfo` on the result to check that the image remains viewable and has the correct photometric interpretation.  Given a source image with a lookup table, the OME-TIFF output should contain photometric=palette; with `-nolookup` it will contain photometric=minisblack.

### DIFF
--- a/docs/sphinx/users/comlinetools/conversion.txt
+++ b/docs/sphinx/users/comlinetools/conversion.txt
@@ -112,6 +112,15 @@ name pattern, then the other must be included too.  The only exception is if
 
       bfconvert -nooverwrite /path/to/input /path/to/output
 
+.. option:: -nolookup
+
+    To disable the conversion of lookup tables, leaving the output
+    file without any lookup tables::
+
+      bfconvert -nolookup /path/to/input /path/to/output
+
+    .. versionadded:: 5.2.0
+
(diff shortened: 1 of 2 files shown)


dgault:
Tested the PR with multiple files and the new option works as expected. Default behaviour works as before and when the -nolookup is added the LUT are ignored as expected and tiffinfo displays Photometric Interpretation: min-is-black

docs/sphinx/users/comlinetools/conversion.txt
```diff
@@ -112,6 +112,15 @@ name pattern, then the other must be included too.  The only exception is if
...
       bfconvert -nooverwrite /path/to/input /path/to/output
 
+.. option:: -nolookup
+
+    To disable the conversion of lookup tables, leaving the output
+    file without any lookup tables::
+
+      bfconvert -nolookup /path/to/input /path/to/output
+
+    .. versionadded:: 5.2.0
```
sbesson: This should be updated to 5.2.1

rleigh-codelibre:
Updated version.

PR #2518, opened 2016-08-11
Docs: Update developer code examples

dgault:
Due to limitations of the tiling example provided this has been removed until further investigation and work can be carried out. The example that was provided did not function correctly across the board and relied on specific parameters and files being used.

The examples have also been updated to fix the asRetrieve function which was no longer correct. Instead it has been moved to OMEXMLService instead of MetadataTools


rleigh-dundee:
This all looks just fine.  Good to merge once it's green and the docs build is green.

PR #2519, opened 2016-08-11
Docs: New Physical Size Example

dgault:
This PR adds a new sample file and brief documentation regarding the reading of physical size dimensions and converting them to display in another quantity unit.

The need for this was raised in a discussion of the ImageJ forums - http://imagej.1557.x6.nabble.com/bioformats-retrieve-calibration-in-JAVA-td5014616.html#a5014627

The new sample file provided is based on the ImageJ example provided here https://gist.github.com/ctrueden/6282856

### DIFF
--- a/docs/sphinx/developers/ReadPhysicalSize.java
+++ b/docs/sphinx/developers/ReadPhysicalSize.java
@@ -62,30 +62,27 @@ public class ReadPhysicalSize {
         
         Unit<Length> targetUnit = UNITS.MICROMETER;
         
-        for (int series=0; series<reader.getSeriesCount(); series++) {
-            reader.setSeries(series);
-            for (int image=0; image<reader.getImageCount(); image++) {
-                Length physSizeX = omeMeta.getPixelsPhysicalSizeX(image);
-                Length physSizeY = omeMeta.getPixelsPhysicalSizeY(image);
-                Length physSizeZ = omeMeta.getPixelsPhysicalSizeZ(image);
+        for (int image=0; image<omeMeta.getImageCount(); image++) {
+            Length physSizeX = omeMeta.getPixelsPhysicalSizeX(image);
+            Length physSizeY = omeMeta.getPixelsPhysicalSizeY(image);
+            Length physSizeZ = omeMeta.getPixelsPhysicalSizeZ(image);
                 
-                System.out.println("Physical calibration - Series: " + series + " Image: " + image);
+            System.out.println("Physical calibration - Image: " + image);
                 
-                if (physSizeX != null) {
-                    Length convertedSizeX = new Length(physSizeX.value(targetUnit), targetUnit);
-                    System.out.println("\tX = " + physSizeX.value() + " " + physSizeX.unit().getSymbol()
-                                       + " = " + convertedSizeX.value() + " " + convertedSizeX.unit().getSymbol());
-                }
-                if (physSizeY != null) {
-                    Length convertedSizeY = new Length(physSizeY.value(targetUnit), targetUnit);
-                    System.out.println("\tY = " + physSizeY.value() + " " + physSizeY.unit().getSymbol()
-                                       + " = " + convertedSizeY.value() + " " + convertedSizeY.unit().getSymbol());
-                }
-                if (physSizeZ != null) {
-                    Length convertedSizeZ = new Length(physSizeZ.value(targetUnit), targetUnit);
-                    System.out.println("\tZ = " + physSizeZ.value() + " " + physSizeZ.unit().getSymbol()
-                                       + " = " + convertedSizeZ.value() + " " + convertedSizeZ.unit().getSymbol());
-                }
+            if (physSizeX != null) {
+                Length convertedSizeX = new Length(physSizeX.value(targetUnit), targetUnit);
+                System.out.println("\tX = " + physSizeX.value() + " " + physSizeX.unit().getSymbol()
+                            + " = " + convertedSizeX.value() + " " + convertedSizeX.unit().getSymbol());
+            }
+            if (physSizeY != null) {
+                Length convertedSizeY = new Length(physSizeY.value(targetUnit), targetUnit);
+                System.out.println("\tY = " + physSizeY.value() + " " + physSizeY.unit().getSymbol()
+                            + " = " + convertedSizeY.value() + " " + convertedSizeY.unit().getSymbol());
+            }
+            if (physSizeZ != null) {
+                Length convertedSizeZ = new Length(physSizeZ.value(targetUnit), targetUnit);
+                System.out.println("\tZ = " + physSizeZ.value() + " " + physSizeZ.unit().getSymbol()
+                            + " = " + convertedSizeZ.value() + " " + convertedSizeZ.unit().getSymbol());
             }
         }
         reader.close();
--- a/docs/sphinx/developers/ReadPhysicalSize.java
+++ b/docs/sphinx/developers/ReadPhysicalSize.java
@@ -2,7 +2,7 @@
  * #%L
  * OME Bio-Formats package for reading and converting biological file formats.
  * %%
- * Copyright (C) 2005 - 2016 Open Microscopy Environment:
+ * Copyright (C) 2016 Open Microscopy Environment:
  *   - Board of Regents of the University of Wisconsin-Madison
  *   - Glencoe Software, Inc.
  *   - University of Dundee
--- a/docs/sphinx/developers/ReadPhysicalSize.java
+++ b/docs/sphinx/developers/ReadPhysicalSize.java
@@ -25,23 +25,13 @@
 
 import java.io.IOException;
 
-import loci.common.services.DependencyException;
-import loci.common.services.ServiceException;
-import loci.common.services.ServiceFactory;
 import loci.formats.FormatException;
-import loci.formats.FormatTools;
 import loci.formats.ImageReader;
-import loci.formats.ImageWriter;
 import loci.formats.MetadataTools;
 import loci.formats.meta.IMetadata;
-import loci.formats.services.OMEXMLService;
 import ome.units.UNITS;
 import ome.units.quantity.Length;
 import ome.units.unit.Unit;
-import ome.xml.model.enums.DimensionOrder;
-import ome.xml.model.enums.EnumerationException;
-import ome.xml.model.enums.PixelType;
-import ome.xml.model.primitives.PositiveInteger;
 
 /**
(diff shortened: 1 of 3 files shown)


docs/sphinx/developers/file-reader.txt
```diff
@@ -79,6 +79,15 @@ table below together with the appropriate accessor method:
 
 All file formats are guaranteed to accurately report core metadata.
 
+Bio-Formats also converts and stores additional information which can be stored and retrieved 
+from the OME Metadata. These fields can be accessed in a similar way to the core metadata above.
```
rleigh-dundee: OME-XML rather than OME.

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @param inputFile the file to be read
+     */
+    public ReadPhysicalSize(String inputFile) {
+        ImageReader reader = new ImageReader();
+        IMetadata omeMeta = MetadataTools.createOMEXMLMetadata();
+        reader.setMetadataStore(omeMeta);
+        reader.setId(inputFile);
+        
+        Length physSizeX = omeMeta.getPixelsPhysicalSizeX(0);
```
rleigh-dundee: We might want to add a comment here to the effect that this is for only the first series, and it might make sense to add a sanity check here by checking the number of series is >0, with getImageCount.  Alternatively, loop over and print out the values for all series.

What's here is fine, but it would make it clearer for example purposes and best practices to follow to have the extra check/loop here.

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,92 @@
...
+        Length physSizeY = omeMeta.getPixelsPhysicalSizeY(0);
+        Length physSizeZ = omeMeta.getPixelsPhysicalSizeZ(0);
+        
+        System.out.println("Physical calibration:");
+        if (physSizeX != null) {
+            System.out.println("\tX = " + physSizeX.value() + " " + physSizeX.unit().getSymbol()
+                               + " = " + physSizeX.value(UNITS.MICROMETER) + " microns");
```
rleigh-dundee: Can't we do a full unit conversion here to a second Length and then use its getSymbol for the micron symbol?  That would perhaps better demonstrate the utility of the unit conversion functionality.

rleigh-dundee:
What's here all looks fine, but there are a few improvements I suggested which might make it a bit better as an example.

docs/sphinx/developers/file-reader.txt
```diff
@@ -79,6 +79,15 @@ table below together with the appropriate accessor method:
...
+An example of such values would be the physical size of dimensions X, Y and Z. The accessor methods 
+for these properties return a :javadoc:`Length <ome/units/quantity/Length.html>` object which 
+contains both the value and unit of the dimension. These lengths can also be converted to other units using 
+:javadoc:`value(ome.units.unit.Unit) <ome/units/quantity/Length.html#value(ome.units.unit.Unit)>`
+An example of reading and converting these physical sizes values can be found in 
+:download:`ReadPhysicalSize.java <ReadPhysicalSize.java>`
```
sbesson: +1 Once we have defined a location for documenting the units system (see https://trello.com/c/FvyQj7OP/144-document-unit-api) we might also introduce a reference to the relevant page.

dgault:
New commits have the following changes based on the feedback:
- Fixed OME to OME-XML in the file-reader.txt
- Example now loops through series and imagecount
- Conversion now uses a new Length and prints using the getSymbol method

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    public ReadPhysicalSize(String inputFile) {
+        ImageReader reader = new ImageReader();
+        IMetadata omeMeta = MetadataTools.createOMEXMLMetadata();
+        reader.setMetadataStore(omeMeta);
+        reader.setId(inputFile);
+        
+        Unit<Length> targetUnit = UNITS.MICROMETER;
+        
+        for (int series=0; series<reader.getSeriesCount(); series++) {
```
rleigh-dundee: getSeriesCount and getImageCount do the same thing in the reader and metadatastore, respectively.  Doesn't this end up printing out the data n^2 times?  You can probably ignore series/setSeries entirely for this example--the iteration over the store images should be sufficient?

rleigh-dundee:
Is this example code (and any other example code) build and tested as part of the regular bio-formats build?

sbesson:
@rleigh-dundee: we do not have the Bio-Formats examples tested regularly although this issue has certainly been raised. An initial step had been done in order to collect various utility scripts across the code but unifying these examples and setting up the testing framework has not been prioritized yet.

One important decision concerns whether these examples should be:
-  maintained in the individual components under e.g. `utils` or even in a third-party repo, tested directly using the regular `testng` infrastructure and copied to the Sphinx documentation using the `autogen` component like the format pages 
- maintained under `docs/sphinx/developers/examples` and tested via some testing infrastructure to put into place.

As I understood, the C++ implementation uses a version of the second option. Do you have any hindsight on how something similar should or should not be implemented in Java?

rleigh-dundee:
@sbesson Wouldn't adding a build.xml/pom file here and adding this to the top-level build be sufficient?   It doesn't need to be particularly fancy, just to build and run them to ensure they are buildable and runnable.

docs/sphinx/developers/ReadPhysicalSize.java
dgault: The getSeriesCount and getImageCount are certainly returning different values, ie you could have 5 series with each series having 10 images etc. Though having thought about it, it really doesn't make any sense to check the physical size on each image, these are going to be fixed values based on the hardware used rather than settings that will change during acquisition. Perhaps, as you had suggested earlier, a commented note to say its being read from the first image would make more sense?

sbesson:
Briefly discussed the C++ framework with @rleigh-dundee, we might look into expanding the build system to execute these classes in 5.2.x. 

Also @dgault would it be possible to expand https://github.com/openmicroscopy/bioformats/blob/develop/docs/sphinx/developers/export2.txt as part of this PR to also show an example of `setPhysicalSize{X,Y,Z}` with a `Length`. Then we should have most of the concrete use cases of units covered (getting, converting, setting).

docs/sphinx/developers/ReadPhysicalSize.java
rleigh-dundee: @dgault We're mixing up two different things here though.  In the FormatReader API, we have "series", and in the MetadataStore API we have "images", and these are the same thing.  When exploring the MetadataStore contents, it doesn't make any sense to call setSeries because it won't have any effect upon what you're iterating over.

This is where the terminology is confused.  FormatReader uses Series and Image; MetadataStore uses Image and Plane.  And Series==Image and Image==Plane (!)

sbesson: In agreement with what is done in the MATLAB example (see bftest.m), I would also suggest to use the MetadataStore API here, loop over the number of OME images/series and skip all usage of  `setSeries()` to streamline the example.

Agreed on the confusing terminology which arises from the original Bio-Formats API implementation and predates the introduction of the OME data model. Maybe, this is something we can investigate and clarify in the next development series. One proposal would be to deprecate `FormatReader.getImageCount` in favor of an unambiguous `FormatReader.getPlaneCount` (as done in `FormatWriter`).

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+ * %%
+ * Copyright (C) 2005 - 2016 Open Microscopy Environment:
```
sbesson: 2016

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,101 @@
...
+import ome.xml.model.enums.PixelType;
+import ome.xml.model.primitives.PositiveInteger;
+
+/**
+ * Example class that shows how to read and convert the physical X, Y and Z dimensions of a file
+ * Bio-Formats version 5.2 or later.
+ */
+public class ReadPhysicalSize {
+    
+    /**
```
sbesson: use two spaces indentation across the example

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,101 @@
...
+import ome.units.quantity.Length;
+import ome.units.unit.Unit;
+import ome.xml.model.enums.DimensionOrder;
+import ome.xml.model.enums.EnumerationException;
+import ome.xml.model.enums.PixelType;
+import ome.xml.model.primitives.PositiveInteger;
+
+/**
+ * Example class that shows how to read and convert the physical X, Y and Z dimensions of a file
+ * Bio-Formats version 5.2 or later.
```
sbesson: I would even mention 5.1 even if the examples is introduced in 5.2, it should be backwards-compatible

docs/sphinx/developers/ReadPhysicalSize.java
```diff
@@ -0,0 +1,100 @@
...
+ * Bio-Formats version 5.1 or later.
+ */
+public class ReadPhysicalSize {
+
+  /**
+   * Construct a new ReadPhysicalSize that will read and convert the physical dimensions of a file
+   *
+   * @param inputFile the file to be read
+   */
+  public ReadPhysicalSize(String inputFile) {
```
sbesson: Exceptions need to be caught  `throws IOException, FormatException`

sbesson:
Once the exceptions are declared/caught, the last version of the read example compiles and works as expected against samples files including both multi-image files and images with/without physical sizes. The setter addition also looks good. @rleigh-dundee: are you fine with the last set of changes? We can look into refactoring the setter example into a file and run documentation examples as part of 5.2.x.

rleigh-dundee:
@dgault The last set look fine.  Minor things I spotted:

- Why is a class contructor being used for the example rather than a static function?
- A number of the variables could be `final`
- There's a missing newline at the end of the file

sbesson:
Thanks, retested the last commit locally. Once Travis validates it, this is good to merge.

dgault:
Updated again based on feedback:
- Moved to a static method rather than class constructor
- Newline at end
- Using final variables
- Removed unused imports

PR #2520, opened 2016-08-11
Docs: Eclipse instructions for using Maven

dgault:
This PR adds documentation on using Bio-Formats as a Maven dependancy through Eclipse. Sample code for the pom dependancy as well as the custom repository are provided. The repository can be added in either the pom or the users settings file, so its worth ensuring that the instructions are clear for either purpose.

Some additional things of note:
- Is the repo location correct (http://artifacts.openmicroscopy.org/artifactory/repo) ?
- Should we recommend using the bioformats-package or individual componets?

### DIFF
--- a/docs/sphinx/developers/building-bioformats.txt
+++ b/docs/sphinx/developers/building-bioformats.txt
@@ -114,7 +114,7 @@ and JAR dependencies (stored in the Maven repository in
 :file:`~/.m2/repository`) on the build path, depending on which projects are
 currently open.
 
-We recommend using Eclipse 4.3 (Kepler), specifically -
+We recommend using Eclipse 4.3 (Kepler) or later, specifically -
 "Eclipse IDE for Java developers". It comes with m2e installed
 (http://eclipse.org/downloads/compare.php?release=kepler).
 
@@ -123,6 +123,8 @@ dependency can be added to the project pom file and should include the
 desired Bio-Formats version. Using bioformats_package as the artifactId
 will include the complete Bio-Formats package, or individual components 
 can be chosen as desired.
+
+::
 
   <dependency>
     <groupId>OME</groupId>
@@ -134,6 +136,8 @@ In order to include this Bio-Formats dependency a custom repository
 must also be added to the project pom or ${user.home}/.m2/settings.xml 
 The repositories element is inherited so for a group of projects the 
 repositories element can be defined at the top of your inheritance chain. 
+
+::
 
   <repositories>
     <repository>
--- a/docs/sphinx/developers/building-bioformats.txt
+++ b/docs/sphinx/developers/building-bioformats.txt
@@ -120,7 +120,7 @@ We recommend using Eclipse 4.3 (Kepler) or later, specifically -
 
 Bio-Formats can be included as a dependency in any Maven project. The 
 dependency can be added to the project pom file and should include the
-desired Bio-Formats version. Using bioformats_package as the artifactId
+desired Bio-Formats version. Using `bioformats_package` as the artifactId
 will include the complete Bio-Formats package, or individual components 
 can be chosen as desired.
 
@@ -133,7 +133,7 @@ can be chosen as desired.
   </dependency>
 
 In order to include this Bio-Formats dependency a custom repository 
-must also be added to the project pom or ${user.home}/.m2/settings.xml 
+must also be added to the project pom or :file:`${user.home}/.m2/settings.xml`.
 The repositories element is inherited so for a group of projects the 
 repositories element can be defined at the top of your inheritance chain. 
 
--- a/docs/sphinx/developers/building-bioformats.txt
+++ b/docs/sphinx/developers/building-bioformats.txt
@@ -147,7 +147,7 @@ repositories element can be defined at the top of your inheritance chain.
     </repository>
   </repositories>
 
-Alternatively you can import the Bio-Formats source by choosing
+If you wish you can instead import the Bio-Formats source by choosing
 :menuselection:`File --> Import --> Existing Maven Projects` from the menu
 and browsing to the top-level folder of your Bio-Formats working copy.
 Alternatively, run the Eclipse Maven target with :command:`mvn
--- a/docs/sphinx/developers/building-bioformats.txt
+++ b/docs/sphinx/developers/building-bioformats.txt
@@ -127,7 +127,7 @@ can be chosen as desired.
 ::
 
   <dependency>
-    <groupId>OME</groupId>
+    <groupId>ome</groupId>
     <artifactId>bioformats_package</artifactId>
     <version>5.2.0</version>
   </dependency>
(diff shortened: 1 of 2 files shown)


docs/sphinx/developers/building-bioformats.txt
```diff
@@ -114,11 +114,40 @@ and JAR dependencies (stored in the Maven repository in
...
+
+  <repositories>
+    <repository>
+      <id>OME</id>
+      <name>Bio-Formats Repo</name>
+      <url>http://artifacts.openmicroscopy.org/artifactory/repo</url>
+    </repository>
+  </repositories>
+
+Alternatively you can import the Bio-Formats source by choosing
```
ximenesuk: Given the next sentence also starts with "Alternatively," it might be better to start this one with, "You can also...|, say.

ximenesuk:
Other than the comment above this reads okay but I hadn't quite appreciated the scope for any workflow testing here. I might defer to @mtbc for a fuller check on the use configuration given here.

mtbc:
Staged at http://www.openmicroscopy.org/site/support/bio-formats5.2-staging/developers/building-bioformats.html#eclipse.

docs/sphinx/developers/building-bioformats.txt
```diff
@@ -114,11 +114,40 @@ and JAR dependencies (stored in the Maven repository in
...
+Bio-Formats can be included as a dependency in any Maven project. The 
+dependency can be added to the project pom file and should include the
+desired Bio-Formats version. Using bioformats_package as the artifactId
+will include the complete Bio-Formats package, or individual components 
+can be chosen as desired.
+
+::
+
+  <dependency>
+    <groupId>OME</groupId>
```
mtbc: This doesn't work for me unless I lower-case `ome`.

docs/sphinx/developers/building-bioformats.txt
```diff
@@ -114,11 +114,40 @@ and JAR dependencies (stored in the Maven repository in
...
 
-We recommend using Eclipse 4.3 (Kepler), specifically -
+We recommend using Eclipse 4.3 (Kepler) or later, specifically -
 "Eclipse IDE for Java developers". It comes with m2e installed
 (http://eclipse.org/downloads/compare.php?release=kepler).
 
-You can then import the Bio-Formats source by choosing
+Bio-Formats can be included as a dependency in any Maven project. The 
+dependency can be added to the project pom file and should include the
+desired Bio-Formats version. Using bioformats_package as the artifactId
```
mtbc: maybe backticks for bioformats_package?

docs/sphinx/developers/building-bioformats.txt
```diff
@@ -114,11 +114,40 @@ and JAR dependencies (stored in the Maven repository in
...
+::
+
+  <dependency>
+    <groupId>OME</groupId>
+    <artifactId>bioformats_package</artifactId>
+    <version>5.2.0</version>
+  </dependency>
+
+In order to include this Bio-Formats dependency a custom repository 
+must also be added to the project pom or ${user.home}/.m2/settings.xml
```
mtbc: `:file:` markup? something anyway for this xml file.

mtbc: end sentence with period

docs/sphinx/developers/building-bioformats.txt
```diff
@@ -114,11 +114,40 @@ and JAR dependencies (stored in the Maven repository in
...
+The repositories element is inherited so for a group of projects the 
+repositories element can be defined at the top of your inheritance chain. 
+
+::
+
+  <repositories>
+    <repository>
+      <id>OME</id>
+      <name>Bio-Formats Repo</name>
+      <url>http://artifacts.openmicroscopy.org/artifactory/repo</url>
```
mtbc: In OMERO our `ivysettings.xml` uses `http://artifacts.openmicroscopy.org/artifactory/maven`; I wonder if we should change that to `repo`, @sbesson?

docs/sphinx/developers/building-bioformats.txt
```diff
@@ -114,11 +114,40 @@ and JAR dependencies (stored in the Maven repository in
 :file:`~/.m2/repository`) on the build path, depending on which projects are
 currently open.
 
-We recommend using Eclipse 4.3 (Kepler), specifically -
+We recommend using Eclipse 4.3 (Kepler) or later, specifically -
 "Eclipse IDE for Java developers". It comes with m2e installed
 (http://eclipse.org/downloads/compare.php?release=kepler).
 
-You can then import the Bio-Formats source by choosing
+Bio-Formats can be included as a dependency in any Maven project. The
```
mtbc: Why is this good stuff buried in the "Eclipse" subsection?

docs/sphinx/developers/building-bioformats.txt
sbesson: Having looked a bit further at the Artifactory documentation, `/repo` is the URL suffix used for the [Virtual repository](https://www.jfrog.com/confluence/display/RTF/Virtual+Repositories) which aggregates *all* local and remote repositories. As this might include unwanted/experimental JARs (in our case `ome.unstable/ome.experimental`), I think we should rather expose to the `/maven` URL.

docs/sphinx/developers/building-bioformats.txt
dgault: Do you think this should be pulled out under a separate Maven heading? Then we could keep the NetBeans and Eclipse specific commands under the original headings.

mtbc: That's a good idea. :+1: Maybe even on a different page if it's no longer about building BF itself but using it.

sbesson: If we were to restructure content, I would suggest this could migrated to http://www.openmicroscopy.org/site/support/bio-formats5.2/developers/java-library.html which is about using Bio-Formats. This should be done in sync with @hflynn who is going to update this page though.

hflynn: Moving content to that page sounds like a good plan. If the content is nearly sorted here, you could do it on this PR and I'll prep my table of dependencies but not open the PR until this is merged.

mtbc: Fine with me to break content out into a separate section here ready for @hflynn to move later.

docs/sphinx/developers/java-library.txt
```diff
@@ -1,7 +1,43 @@
 Using Bio-Formats as a Java library
 ===================================
 
-If you wish to make use of Bio-Formats within your own software, you can
+Bio-Formats as a Maven dependency
+-----------------
```
sbesson: These should have the same size as the title otherwise Sphinx will fail with the warnings treated as errors options turned on.

sbesson:
@mtbc: staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/developers/java-library.html.

mtbc:
Good to merge.

PR #2524, opened 2016-08-12
Document ome-xml unit classes

rleigh-dundee:
- Link to current name for deprecated units
- Document Unit and Quantity classes
- Document concrete Quantity types

### DIFF
--- a/components/ome-xml/src/ome/units/quantity/Temperature.java
+++ b/components/ome-xml/src/ome/units/quantity/Temperature.java
@@ -28,24 +28,29 @@ package ome.units.quantity;
 import ome.units.unit.Unit;
 
 /**
- * A wrapper for the Temperature class from the units implementation.
+ * Length quantity.
  *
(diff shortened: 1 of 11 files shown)


components/ome-xml/src/ome/units/quantity/Temperature.java
```diff
@@ -28,24 +28,29 @@
 import ome.units.unit.Unit;
 
 /**
- * A wrapper for the Temperature class from the units implementation.
+ * Length quantity.
```
dgault: Temperature quantity

dgault:
Documentation all reads well and deprecated units are all correct.

The only one nitpicky thing is that they all use 'create an' which reads a bit off for certain types eg Length 'Create an Length' -> 'Create a Length'

rleigh-codelibre:
@dgault Should all be fixed now.

dgault:
Everything looks good with the changes. PR is ready to merge.

PR #2525, opened 2016-08-15
Set SPHINXOPTS option to fail on warnings in Travis build

sbesson:
While reviewing https://github.com/openmicroscopy/bioformats/pull/2520, I realized that Travis was no longer treating warnings as error as we do in our CI jobs. This commit should restore this behavior.

See https://travis-ci.org/sbesson/bioformats/builds/152396660 for an expected of failure on top of this commit.

### DIFF
--- a/tools/test-build
+++ b/tools/test-build
@@ -62,8 +62,7 @@ sphinx_html()
 {
     (
         cd docs/sphinx
-        export SPHINXOPTS="-W"
-        ant -Dsphinx.warnopts=$SPHINXOPTS html
+        ant -Dsphinx.warnopts=-W html
      )
  }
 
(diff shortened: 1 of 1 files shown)


tools/test-build
```diff
@@ -62,6 +62,7 @@ sphinx_html()
 {
     (
         cd docs/sphinx
+        export SPHINXOPTS="-W"
```
rleigh-dundee: Why are we exporting this in the evironment when it's being passed on the command-line to ant?  Why not just use `-Dsphinx.warnopts=$SPHINXOPTS=-W` ?

sbesson: I copied over the other Sphinx target defining SPHINXOPTS but you are right this is clearer. Will push a commit.

rleigh-dundee:
Looks fine to me.  Please merge.

PR #2528, opened 2016-08-16
minor code cleanup for formats classes

mtbc:
Should slightly improve performance, maybe readability, but not make behavioral changes. Check that code looks like it will do the same as before and that CI passes. Partly about introducing good patterns before people copy them into further code. See http://trac.openmicroscopy.org/ome/ticket/12385.


sbesson:
NB: https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-full-repository/41/ reports 35 failures which are most likely unrelated to this PR but rather related to JDK specifics (see [this card](https://trello.com/c/l2oMl30u/37-investigate-jdk-specific-automated-test-failures)).

@mtbc: would you expect these changes to affect serialization when reading a cached 5.2.0 reader using a 5.2.x Bio-Formats?

mtbc:
Gah, yes, some would affect serialization; Kryo just keeps on biting me. I'd forgotten why I was trying to get this work in before I left: we have to keep the caches valid over minor releases? Might honestly be easier to just close the PR and repeat the work again some other time rather than try to unpick what will affect what.

sbesson:
It all narrows down to enforcing the policy we want to adopt with regard to stable releases. Model/API and reader core metadata breaking changes are certainly disallowed. Serialization breaking change could arguably be classified in the same category as core metadata breaking change espeically as the latter will involve the former.
On the other hand, the major component of our software stack relying on the serialization stability is OMERO and OMERO 5.3.0 will not be released before a couple of releases of Bio-Formats 5.2.x. /cc @jburel @joshmoore The main concern is whether anyone else in the community would rely on serialization stability in a stable line.

The conservative approach would certainly be to have this change scheduled for 5.3.0 but happy to hear other thoughts.

mtbc:
Well, BIOFORMATS-DEV-merge-full-repository seems to be staying green anyway.

components/formats-gpl/src/loci/formats/in/MetamorphReader.java
```diff
@@ -352,7 +351,7 @@ protected void initFile(String id) throws FormatException, IOException {
       LOGGER.info("Initializing " + id);
       // find an associated STK file
       String stkFile = id.substring(0, id.lastIndexOf("."));
-      if (stkFile.indexOf(File.separator) != -1) {
+      if (stkFile.indexOf(File.separatorChar) != -1) {
```
joshmoore: This is a surprise. Could this have unintended consequences?

components/formats-gpl/src/loci/formats/in/MetamorphReader.java
```diff
@@ -937,7 +936,7 @@ else if (exposureTimes.size() == 1 && exposureTimes.size() < getSizeC()) {
 
       for (int p=0; p<getImageCount(); p++) {
         int[] coords = getZCTCoords(p);
-        Double deltaT = Double.valueOf(0);
+        Double deltaT = 0d;
```
joshmoore: Why autoboxing here and not elsewhere? (For the record, I'm for auto(un)boxing)

mtbc: Probably just one I happened to notice, I'm progressively getting them over time!

joshmoore:
:+1: for:
 * Vector -> ArrayList
 * sizing new arrays
 * import cleanups
 * more `final`s
 * use of `contain`, `isEmpty`, etc
 * fewer object instantiations

Few other comments inline. I also remember one example of a change in semantics between Vector and List which we had recently. Order of enumeration? That should be checked, since it did impact end users.

As for the serialization question, depending on the target, I don't see a big issue. Though we *do* need to define a serialization policy and stick to it like glue.

components/formats-gpl/src/loci/formats/in/MetamorphReader.java
mtbc: Standard libraries simply have,
```java
public static final String separator = "" + separatorChar;
```
They're the same, this is really just about switching to calling a method that can assume searching for a single character instead of having to care about search string length.

joshmoore: TIL

mtbc: http://docs.oracle.com/javase/7/docs/api/java/io/File.html#separatorChar and http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#indexOf%28int%29 for convenient double-checking. :smiley:

components/formats-gpl/src/loci/formats/in/MetamorphReader.java
joshmoore: But it seems like you are actively moving to non-auto elsewhere. (e.g. use of `.booleanValue()`)

mtbc: ... its removal? Just trying to reduce visual clutter that has no operational effect.

joshmoore: Ah, ignore me. My eyes got crossed reading the diff (red became green). We agree and it's very nice. Thank you.

mtbc:
I don't remember a `Vector` to `List` issue; perhaps you're thinking of `Stack` to `Deque`? Definitely an order of enumeration issue there.

joshmoore:
Ah, quite possibly. Thanks for reassuring me.

mtbc:
Merge into same version of Bio-Formats as #2557 to reduce serializer breakage.

PR #2533, opened 2016-08-17
Remove unnecessary BinData/MetadataOnly handling in ImageInfo

sbesson:
The addition of MetadataOnly elements for OME-XML containing no BinData or TiffData elements should now be handled via the MetadataTools API. There is no reason to have some special handling in the ImageInfo class which should return the OME-XML using the MetadataRetrieve interface directly.

To review this PR, run 

```
showinf -nopix -omexml /path/to/ome.xml
```

on a valid OME-XML file e.g. one of the OME samples and check the `MetadataOnly` block is no longer populated but `BinData` blocks are available.


mtbc:
I wonder what to make of `BinData` elements with `Length="0"`!

mtbc:
This is a welcome improvement. Seems to me to be **good to merge**.

PR #2534, opened 2016-08-17
cpp: ome-xml: Quantity: Allow unit conversions for arbitrary value types

rleigh-codelibre:
The existing code works, but isn't sufficiently flexible.  While it
permits quantities of any numerical type, it only supports a double
value type for unit conversions.  This refactors the code to support
conversion of arbitrary value types.

- Transform the internal source files into public headers
- Generalise conversion helpers

Purpose: it corrects a defect found when writing the unit usage examples.

----------

Testing: handled by the existing unit tests.  Check jobs remain green.


dgault:
The changes here to the make these public headers makes a lot of sense and the code all looks good.
The unit tests appear to be all passing and the builds are green. This PR looks good to merge.

PR #2535, opened 2016-08-18
Add Maven module to compile and run examples

sbesson:
### Background

See https://trello.com/c/C58pWYqf/33-investigate-maven-target-for-running-examples
Several cards have reported the need to regularly test the Bio-Formats examples code. See for instance the discussion while adding  a new example in our documentation in https://github.com/openmicroscopy/bioformats/pull/2519. What is missing is a simple framework allowing to expand developer examples and get them tested by the CI build.

### Implementation

This PR declares a new Maven module under `docs/sphinx`, compiles the examples and uses the `maven-exec-plugin` to execute them as part of the regular Maven build.

A few implementations choices are proposed in the first version of this PR open to discussion at this stage:

- the examples are migrated to `docs/sphinx/developers/examples` making it the default location for adding new example files. The main advantage of this approach is that it does not require an extra build step to copy files to the Sphinx documentation so that example classes can be directly added, built and referenced in the documentation via the `:download:` target
- a single top-level `ExampleSuite` class is maintained in the same examples directory that executes all examples. This should reduce the need to modify the `pom.xml` everytime an example is added.
- the `maven-install-plugin` plugin is skipped so that no artifact is effectively installed/uploaded to a repository


joshmoore:
Primary thoughts:

 * :+1:
 * Where/when all will this be run? By travis as well, right? Anything needed in CI to check for "WARN"?
 * I'd likely opt for removing `ExampleSuite` in favor all specifications in `pom.xml` or move the argument from `pom.xml` to `ExampleSuite`.

sbesson:
The addition of the Maven module was certainly designed to have it run in Travis as well as OME CI (and fail the build if the examples break). The `logback.xml` was largely introduced to reduce the DEBUG-level logging but there is no significance associated with `WARN` so I can simply turn the console output OFF.

Will push a commit to remove `ExampleSuite` in favor of individual specifications for now. My initial idea was to reduce the number of places to maintain when adding a new developers example but either way it needs to be registered with the correct arguments in either `ExampleSuite` or `pom.xml`.

joshmoore:
```
[INFO] Bio-Formats Sphinx documentation ................... SUCCESS [  1.154 s]
```

:thumbsup:

PR #2536, opened 2016-08-19
CZI Reader - Correct dimensions per series

dgault:
This is a PR to fix an issue raised in https://github.com/openmicroscopy/bioformats/issues/2467

Previously the dimension values (Z and C) across each series would be identical, with the largest value be replicated across each series.

With the PR in place the dimensions will be calculated and stored for each individual series giving them the correct values.

Testing:
- Unit tests and data repo jobs should remain green
  Using the sample file provided in https://github.com/openmicroscopy/bioformats/issues/2467
- Verify that without the PR the same dimensions are displayed for each series
- Verify that with the PR applied that the correct dimensions for each individual series are stored
- Verify that each series still opens and displays correctly but with the correct metadata

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1268,7 +1268,7 @@ public class ZeissCZIReader extends FormatReader {
     for (int i=0; i <getSeriesCount(); i++){
       setSeries(i);
       CoreMetadata seriesCore = core.get(i);
-      seriesCore.imageCount = getSizeZ() * getSizeC();
+      seriesCore.imageCount = getSizeZ() * (isRGB() ? 1 : getSizeC()) * getSizeT();
     }
     setSeries(oldSeries);
   }
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -804,6 +804,7 @@ public class ZeissCZIReader extends FormatReader {
     }
 
     assignPlaneIndices();
+    calculateSeriesDimensions();
 
     if (maxResolution > 0) {
       tileWidth = new int[core.size()];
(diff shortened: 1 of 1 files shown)


sbesson:
Note this PR causes a couple of errors on existing CZI datasets including some `ArrayOutOfBoundExceptions` - see https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/277/

sbesson:
With the last commit, all filesets in our repository open correctly. As expected, some of the files in our data repositories now have Z/T dimensions marked as incorrect - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/108/console

For some of these filesets, e.g. qa-10301, the XZYCT dimensions were preserved across series on purpose to be consistent the sparse nature of the data - see https://github.com/openmicroscopy/bioformats/pull/1386 for a related discussion. Is there a way to discriminate between these different use cases /cc @melissalinkert

melissalinkert:
I can confirm that the number of Z sections for each series in Zen matches what is listed in #2467.  The Z slider goes up to 78 for each of the 4 series (views), but scrolling past the last valid plane for each results in a blank gray plane with no pixel values.

In this case, all of the missing planes are at the end - so all valid planes are contiguous and there are no holes.  If I read/remember correctly, qa-10301 has missing planes in between valid planes, so that is probably the way to decide whether or not the reduce the dimension count.

melissalinkert:
Pasting Zen screenshots here since the sample file is already public:

View 1 with Z slider set to 78 (notice no pixel values):

image

View 2 with Z slider set to 78 (very dark, but pixel values present):

image

View 4 with Z slider set to 1 past the end of the valid stack:

image

sbesson:
Given the comments/screenshots above, I am worried of the impact of systematically reducing the image dimensionality when planes have missing pixel values also since it will conflict with some users expectations.

Coming back to the context of the [original issue](https://github.com/openmicroscopy/bioformats/issues/2467) and looking briefly at the code using the provided dataset, it seems we should be to identify planes with no corresponding subblocks in [this statement](https://github.com/openmicroscopy/bioformats/blob/v5.2.1/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java#L1085). We do not have a proper API to do this but would be this detection be a valuable first step in your analytical context @StephanPreibisch? If so what would be reasonable here e.g. would a logging statement reporting such series/planes be useful/sufficient?

StephanPreibisch:
Hi, I am not sure about everything you said. Here is a recap of what I know for sure: 
The correct stack sizes per view/angle as defined by the user itself are somewhere in the CZI as previous versions of bioformats reported it correctly. 

It is displayed wrongly in Zen itself, but only because of limitations of Zen itself (slider issues), the Zeiss people told me.

I wrote a workaround in my software, so I would like to solve it properly. I can comment again next week in more detail when I have more time.

Thanks a lot,
Stephan

sbesson:
@StephanPreibisch: all understood and thanks for the extra information about Zen which we have definitely used as our reference tool for validating the data handling so far. Happy to discuss more of the details whenever we have time.
As a side note, if it turns our we need to give users some flexibility in the way these files should be handled, this issue might be a another candidate for the MetadataOptions API /cc @joshmoore @simleo

sbesson:
Closing for now as the https://github.com/openmicroscopy/bioformats/pull/2648 work takes priority. We can rework this change once the new options API is in place.

PR #2537, opened 2016-08-19
Make the INI parser less stringent

ctrueden:
This fixes a bug which was [reported on the ImageJ forum](http://forum.imagej.net/t/2443).

The sample non-working file has been uploaded as [QA 17305](http://qa.openmicroscopy.org.uk/qa/feedback/17305/).

-------

Here is a sample TIFF description header produced by a Hamamatsu instrument which previously failed to parse using the `IniParser`:

```
Created by Hamamatsu Inc.
Fri, 19 Aug 2016 15:42:31 Central Daylight Time

[ CALIBRATION ]
title = Default
units = Pixels
symbol = Px
factor = 1.000000
scale_len = 100.000000
[ CAPTURE DEVICE ]
 Camera Type = D_CAM
 Camera Name = C12345-67CU S/N: 987654
 Camera Size = 2048x2048
 Bit Depth = 16-Bit
 Binning = 1
 Capture Region = (X0=300, Y0=900, Width=1600, Height=324)

[ CAPTURE SETTINGS ]
 eOffset1 = 100.00
 eCoeff1 = 0.49000
 Exposure1 = 0.450000 s

[ CAPTURE TIME ]
 Time_From_Start = 00:00:32.2370
 Time_From_Last = 00:00:15.2410

[ EVENT MARKER ]
Event1+Event2
```

The issue is that some of the lines do not have an equals sign, nor enclosing square brackets. So we have a couple of choices here:

1. Detect this not-quite-an-INI flavor of TIFF metadata, and parse it using something other than the `IniParser`.
2. Loosen up the `IniParser` to do a "best effort" rather than failing.

This commit chooses the latter option, since it was a simple change.

In short: let's follow Postel's Law here: permissive when reading, conservative when writing.

### DIFF
--- a/components/formats-common/src/loci/common/IniParser.java
+++ b/components/formats-common/src/loci/common/IniParser.java
@@ -177,7 +177,10 @@ public class IniParser {
 
       // parse key/value pair
       int equals = line.indexOf("=");
-      if (equals < 0) throw new IOException(no + ": bad line");
+      if (equals < 0) {
+        LOGGER.debug("Ignoring line {}", no);
+        continue;
+      }
       String key = line.substring(0, equals).trim();
       String value = line.substring(equals + 1).trim();
       attrs.put(key, value);


components/formats-common/src/loci/common/IniParser.java
```diff
@@ -177,7 +177,10 @@ public IniList parseINI(BufferedReader in)
 
       // parse key/value pair
       int equals = line.indexOf("=");
-      if (equals < 0) throw new IOException(no + ": bad line");
+      if (equals < 0) {
+        LOGGER.debug("Ignoring line %d", no);
```
mtbc: From `showinf -debug` console output:
```
Ignoring line %d
```
I don't think the expected String format parsing is happening.

mtbc:
I wonder if there's some multi-line value format where we should skip lines until indent returns to where it was, but that's just a random thought outwith this PR's scope.

This PR does allow the uploaded image to be viewed without errors, thank you for the sample data.

sbesson:
@ctrueden: thanks for the PR, barring https://github.com/openmicroscopy/bioformats/pull/2537#discussion_r75633176, this looks good and we will make sure the sample file gets added to the data repository.

A wider question related to the PR description arises from the fact the INI format specification is quite loosely defined. While being permissive in the context of the file reader is completely sensible, is there any other use case of the INI reader that would justify maintaining the previous state via some mechanism e.g. via a `strict` or `legacy` flag?

components/formats-common/src/loci/common/IniParser.java
ctrueden: Doh, thanks. I meant to double check that syntax before pushing but forgot. Fixed.

ctrueden:
> is there any other use case of the INI reader that would justify maintaining the previous state via some mechanism e.g. via a strict or legacy flag?

Not to my knowledge, which is why I went ahead with this change.

Certainly a `strict` toggle could be added later if the need arises.

sbesson:
Understood. Merging this for 5.2.1 on the basis of green tests in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/285/.

PR #2539, opened 2016-08-22
ImageJ plugin: fix file pattern prefix

simleo:
What it does
-----------
Fixes the stitching of file patterns in the ImageJ plugin, where the dirname part was being duplicated.

Background
----------
#2352 changed the semantics of `FilePattern.getPrefix` to allow pattern blocks in the whole filename, rather than in the basename alone. Now the first prefix of a file pattern string is whatever comes before the first pattern block, *including* parent directories, if any. This means that the logic in `FilePatternDialog.harvestResults`, which rebuilds the pattern string from prefixes and blocks, has to start from an empty string.

Testing
-------

Place two single-plane images corresponding to different time points (or z sections, or channels: all that matters is that they have the same XY size) in a newly created directory, e.g.:

```
/tmp/ij_test/foo_t0.tif
/tmp/ij_test/foo_t1.tif
```

Download the latest Fiji, update the BF plugin to make sure it's the one released with 5.2.0 and use the bio-formats importer to open the first file, checking "Group files with similar names" in the import options. Press OK, then OK again in the "Bio-Formats File Stitching" dialog box. You should get an import error, and the Log window should show a `FileNotFoundException` for `/tmp/ij_test/tmp/ij_test/foo_t0.tif` (duplicate dir).

Now close Fiji, build the version of Bio-Formats corresponding to this PR and update Fiji's plugin with it. You can use https://github.com/fiji/fiji/blob/master/bin/update-bf.sh: comment out the `mvn dependency:get ...` line and supply the snapshot version as the second arg (e.g., `5.2.1-SNAPSHOT`). Restart Fiji and retry the above import. This time it should work fine.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
@@ -211,8 +211,7 @@ public class FilePatternDialog extends ImporterDialog {
       }
     }
     else if (useRanges) {
-      String pattern =
-        originalID.substring(0, originalID.lastIndexOf(File.separator) + 1);
+      String pattern = "";
       for (int i=0; i<counts.length; i++) {
         BigInteger first = new BigInteger(firsts[i]);
         BigInteger fileCount = new BigInteger(counts[i]);


pwalczysko:
Situation preceding this PR: 
  - getting the error ``java.io.FileNotFoundException: /Users/pwalczysko/Desktop/test-ij/Users/pwalczysko/Desktop/test-ij/image_t1.tif (No such file or directory)`` which is saying that he cannot find the second file (...t0.tif is the first file I was opening, the ....t1.tif is the second file which should have been found automatically). I suppose this is what was meant in the header of this PR.

simleo:
@pwalczysko actually the bug causes the whole pattern to be corrupted, so it should fail to open *any* matching file. I suppose that's simply the first that BF attempted to open.

pwalczysko:
Works as expected. The two images open in a stack which has the two planes, indicated as timepoints or z planes according to the naming of the two images (...t0.tiff etc. names == timepoints, ...01.tiff etc. names == z planes).
Ready to merge.

PR #2546, opened 2016-08-25
Dimension Swap Dialog - Incorrect dimension sizes

dgault:
Issue was discovered by user email - http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-August/006129.html
Associated QA-17311

To reproduce:
- Using Fiji import an image with different dimension sizes for each series
- Select the Swap Dimensions
- Select a series with dimensions differing that of the last series
- Verify that the dimensions shown in the Swap Dialog are incorrect
- Verify that an IllegalArgumentException is thrown when attempting to open the image

To test:
- With the PR applied reproduce the steps above
- Verify that the correct dimensions are displayed in the dialog
- Verify that the image opens as expected
- Retest with other series and combinations of series

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/SwapDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/SwapDialog.java
@@ -68,14 +68,14 @@ public class SwapDialog extends ImporterDialog implements ItemListener {
     GenericDialog gd = new GenericDialog("Dimension swapping options");
 
     String[] labels = {"Z", "C", "T"};
-    String[] sizes = new String[] {
-      String.valueOf(reader.getSizeZ()), String.valueOf(reader.getSizeC()),
-      String.valueOf(reader.getSizeT())
-    };
     for (int s=0; s<seriesCount; s++) {
       if (!options.isSeriesOn(s)) continue;
       reader.setSeries(s);
 
+      String[] sizes = new String[] {
+          String.valueOf(reader.getSizeZ()), String.valueOf(reader.getSizeC()),
+          String.valueOf(reader.getSizeT())
+        };
       gd.addMessage("Series " + (s + 1) + ":\n");
 
       for (int i=0; i<labels.length; i++) {


will-moore:
Using image: data_repo/test_images_metadata/leica-lif/Rosier.lif
I reproduced the same stack trace as reported without this PR. Then with the updated bio-formats_plugins.jar I got different Dimensions in the "Swap Dimensions" dialog for each Series.

However, I got a different error with the updated jar that I didn't see before.
```
(Fiji Is Just) ImageJ 2.0.0-rc-49/1.51d; Java 1.8.0_66 [64-bit]; Mac OS X 10.10.5; 48MB of 10129MB (<1%)

java.lang.NoSuchFieldError: MICROMETER
	at loci.plugins.in.Calibrator.applyCalibration(Calibrator.java:71)
```

dgault:
The error you are seeing here is due to a combination of Bio-Formats 5.1.x and 5.2.x jars being used. The units work done in 5.2 will have introduced the new longer units enums.

Either the remaining jars will need to be replaced with the latest 5.2 versions or you can update them via Fiji, using the 'Help -> Update Fiji ' option

sbesson:
@bramalingam what was the output of the retesting of this PR?

will-moore:
Sorry - didn't realise you wanted re-testing. If I know what all the "remaining jars" are I'm happy to repeat.

dgault:
The easiest way to update the jars will be to use the Fiji updater ('Help -> Update Fiji'), otheriwse you can simply replace the jars found in the Fiji directory under jars/Bio-Formats.

The latest 5.2.1 release jars can be used http://downloads.openmicroscopy.org/bio-formats/5.2.1/ with alongside the new bio-formats-plugins jar with this PR included.

dgault:
Yeah, just replacing the jars in /jars/bioformats with bioformats_package.jar works just fine.

will-moore:
After downloading the bioformats_package.jar and replacing the /jars/bioformats with it, I'm not seeing this PR fix working with the latest bio-formats_plugins.jar from https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/lastSuccessfulBuild/artifact/artifacts/. I'm still seeing the dimensions in the swap dialog remain same as previous image and get the same exception java.lang.IllegalArgumentException: order is unexpected length (4).
If I revert back to the /jars/bioformats/ directory then I see this PR fixed, BUT I MICROMETER error as before. Seems I can't get both fixes together!
With the MICROMETER error it tells me I am using 5.1.10-DEV. How do I tell the Fiji updater that I want to use BF 5.2?

dgault:
Ah, your right the formats package will also have the old plugins version which will cause the mismatch. If you use the bio-formats-package from https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/lastSuccessfulBuild/artifact/artifacts/ then that should contain everything you need

will-moore:
Great, that works! So I think this is good to merge.
However.... I did find another related bug (when I really tried to break things!):
If I open 2 images at once, swapping dimensions on the second image (dims are 60, 1, 1) didn't automatically swap them, so I was able to select 60 for more than 1 dimension, which then gave me an error: 
```
java.lang.IllegalArgumentException: order is unexpected length (3)
	at loci.formats.DimensionSwapper.swapDimensions(DimensionSwapper.java:92)
```
screen shot 2016-08-31 at 11 25 48
Happy for this to be fixed later if that makes sense.

sbesson:
Thanks for the testing and the extra bug @will-moore  My impression while looking at `SwapDialog` is that this code was not written to handle the swapping of multiple series e.g. only three listeners are defined. I don't think this bug should block this PR.
Given the above, would it make sense to block the swapping dialog box if more than one series is selected or would it be too breaking for consumers? /cc @melissalinkert

PR #2548, opened 2016-08-26
Maven examples: add top-level example suite class

sbesson:
See https://trello.com/c/lFM1xUyH/40-sphinx-example-suite-class

Following the migration of all Java examples under `docs/sphinx/developers/examples`, this PR adds a top-level class to run all examples with a single command and uses this class in the top-level `pom.xml`.

To review this PR:
- check Travis is green as the Maven matrix builds will execute this class as part of the test phase of the `docs/sphinx` component
- run `mvn` locally and check `docs/sphinx/target/classes` contains the fake files used for running the example and the OME-TIFF files generated from the conversion/export examples.

A few comments:
- except for `ReadPhysicalSize` which outputs some content, there is no console trace of the example executed. Is it worth adding some logging/`System.out.println` to the main class
- I used BSD to license `ExampleSuite` and realized the existing examples are GPL licensed. There is no GPL specificity so we could consider moving everything to BSD.

### DIFF
--- a/docs/sphinx/developers/examples/ExampleSuite.java
+++ b/docs/sphinx/developers/examples/ExampleSuite.java
@@ -36,6 +36,14 @@ import java.io.File;
 
 public class ExampleSuite {
 
+  public static void execute(String name, String[] args) throws Exception {
+    System.out.println("Executing " + name);
+    Class c = Class.forName(name);
+    Object passedArgs[] = {args};
+    c.getMethod("main", args.getClass()).invoke(null, passedArgs);
+    System.out.println("Success");
+  }
+
   /**
    * Execute a series of examples using the test files
    *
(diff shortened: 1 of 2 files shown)


dgault:
Running maven locally the example suite appears to have run correctly, though as mentioned the readPhysicalSize example is the only one which provides any sort of feedback.

Perhaps it would be worth adding extra logging to say that the example suite is being run and which examples are included. As these are examples rather than tests, other than catching an exception there is no real way of determining success or failure, so simply a notification that a particular example has been run in likely sufficient.

Regarding the licensing, moving to BSD would seem like a sensible approach for these if there is nothing GPL specific included in any of them.

sbesson:
@dgault: the last commit should now allow `ExampleSuite` to print some stdout output for each run example.

dgault:
With the latest commit the output from a local mvn build is generating the extra logging as expected.

PR looks good to merge.

Issue #2490, opened 2016-07-20
Add zero padding to bfconvert pattern to fix sorting 


burgerga:
When I use `bfconvert` on a large XYZCT stack with 

    bfconvert  sample.nd2 output_xy%sz%zc%ct%t.ome.tif

I get the following output files:

```
output_xy0z0c0t0.ome.tif
output_xy0z0c0t10.ome.tif
output_xy0z0c0t11.ome.tif
output_xy0z0c0t12.ome.tif
output_xy0z0c0t13.ome.tif
output_xy0z0c0t14.ome.tif
output_xy0z0c0t15.ome.tif
output_xy0z0c0t16.ome.tif
output_xy0z0c0t17.ome.tif
output_xy0z0c0t18.ome.tif
output_xy0z0c0t19.ome.tif
output_xy0z0c0t1.ome.tif
output_xy0z0c0t20.ome.tif
output_xy0z0c0t21.ome.tif
output_xy0z0c0t22.ome.tif
output_xy0z0c0t23.ome.tif
output_xy0z0c0t24.ome.tif
output_xy0z0c0t25.ome.tif
output_xy0z0c0t26.ome.tif
output_xy0z0c0t27.ome.tif
```

Would it be possible to add zero padding (in a printf like way) to get natural sorting?

dgault:
Hi Gerhard,

There are 2 possible ways we could do this:
- Automatically add the zero padding based on the max size of the parameters
- Add a new option to the API to allow for formatting (like printf)

I have put together some initial code which would enable the first option - https://github.com/dgault/bioformats/commit/1ac64b36629d92d15656d2837a3ea237b77b9165

As this has the potential for an API change there will have to be further discussion on the best possible solution. If you have any preferences on which you would like to see it would be good to know.

With Thanks,
David

burgerga:
@dgault Hi David,

Thanks for your fast response!
I think I prefer the first option, because it's simple to use (and to implement ;) ). At them moment I can't think of use cases where option 2 would be worth the effort.

sbesson:
Should be fixed in [Bio-Formats 5.2.2](https://www.openmicroscopy.org/community/viewtopic.php?f=11&t=8118).

PR #2550, opened 2016-08-29
BF Convert - Optional Zero Padding

dgault:
This PR is aimed to add a zero padding option when using bfconvert tool.
- When the -padded option is set the filename parameters for series, z, c and t should be padded with zeros. The padding length is not configurable but is based on the image parameters.
- When the new -padded option is not set the default behaviour should be preserved 

The request has been raised in various places:
https://github.com/openmicroscopy/bioformats/issues/2490
https://trac.openmicroscopy.org/ome/ticket/12678
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-August/006130.html

To test:
- With the new parameter set verify that bfconvert (on an image with a dimension or series parameter > 10) will be padded with zeros:
For example - bfconvert  testing.nd2 output_xy%sz%zc%ct%t.ome.tif -padded
Should result in filenames such as:

| Original | New Padded name |
|---------|-----------------------|
| output_xy0z0c0t0.ome.tif | output_xy0z000c0t0.ome.tif |
| output_xy0z9c0t0.ome.tif | output_xy0z009c0t0.ome.tif |
| output_xy0z10c0t0.ome.tif | output_xy0z010c0t0.ome.tif |
| output_xy0z19c0t0.ome.tif | output_xy0z019c0t0.ome.tif |
| output_xy0z20c0t0.ome.tif | output_xy0z020c0t0.ome.tif |
| output_xy0z100c0t0.ome.tif | output_xy0z100c0t0.ome.tif |

- Verify that the original default behaviour still works as before when the new option is not set


components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -1024,13 +1024,17 @@ public static String getTileFilename(int tileX, int tileY,
    * @throws IOException Never actually thrown.
    */
   public static String getFilename(int series, int image, IFormatReader r,
-    String pattern) throws FormatException, IOException
+    String pattern, boolean padded) throws FormatException, IOException
```
sbesson: As this is breaking public API, can you also re-add the previous method signature calling this implementation with  `padded=false`?

melissalinkert:
Changes read fine to me, so no veto to merging as-is assuming tests pass.  Is the plan to enable this in ImageJ as well?

I do wonder if it would be useful to allow the user to configure the number of digits, either per parameter or across the board, but that might just be something to consider for later on.

sbesson:
Exposing this API in the `bio-formats-plugins` exporter functionality certainly makes sense since one of the requests was raised in the context of the Fiji exporter. I would consider the additional granularity as outside the scope of this PR.

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -144,6 +145,7 @@ private boolean parseArgs(String[] args) {
         else if (args[i].equals("-autoscale")) autoscale = true;
         else if (args[i].equals("-novalid")) validate = false;
         else if (args[i].equals("-validate")) validate = true;
+        else if (args[i].equals("-padded")) zeroPadding = true;
```
sbesson: This option should likely be documented in the usage string as well.

dgault:
New commits add documentation for both the usage string as well as the sphinx doc.
A new option has also been added to the Plugins exporter and the functionality to pad the filenames added when writing dimensions to separate files.

hflynn:
Sphinx docs addition looks good

components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
```diff
@@ -546,12 +552,21 @@ else if (cal.frameInterval != 0.0 && cal.getTimeUnit().equals("sec")) {
...
+                  cPlaces = "%0" + String.valueOf(sizeC).length() + "d";
+                }
+                
                 int nextFile = 0;
                 for (int z=0; z<(splitZ ? sizeZ : 1); z++) {
                     for (int c=0; c<(splitC ? sizeC : 1); c++) {
                         for (int t=0; t<(splitT ? sizeT : 1); t++) {
-                            outputFiles[nextFile++] = base + (splitZ ? "_Z" + z : "") +
-                                    (splitC ? "_C" + c : "") + (splitT ? "_T" + t : "") + ext;
+                            outputFiles[nextFile++] = base + (splitZ ? "_Z" + String.format(zPlaces, z) : "") +
```
sbesson: What is preventing us from consuming  `FormatTools.getFilename()` directly here: is it because it requires an `IFormatReader` that we don't have at this point?

dgault: Well we dont have a reader but we could rework the FormatTools function to get around that. It also writes a slightly different filename as it doesn't have the series parameter. Again we could rework FormatTools to add the new parameters that we need.

sbesson: Definitely worth discussing the effort but I think there is value in using this RFE to refactor/unify our usage of the API. Initial thoughts on a additional refactoring were  along the lines of:

```
public static String getFilename(int series, int image, IFormatReader r,
    String pattern, boolean padded) throws FormatException, IOException
  {
     MetadataStore store = r.getMetadataStore();
     MetadataRetrieve retrieve = store instanceof MetadataRetrieve ?
       (MetadataRetrieve) store : new DummyMetadata();
     return getFilename(series, image, retrieve, pattern, padded) 
  }

public static String getFilename(int series, int image, MetadataRetrieve retrieve,
    String pattern, boolean padded) throws FormatException, IOException
  {
```

A secondary advantage of the proposal above is that I suspect I allows us to cover it in `FormatToolsTest` by initializing an `IMetadata` object and testing various combinations of dimensionalities/patterns.

sbesson:
"This branch has conflicts that must be resolved". Could you merge `origin/develop` to resolve it?

dgault:
With the latest changes the option is now also available when exporting through Fiji, as such this scenario should be added to the tests for this PR:

- Open an image in Fiji (on an image with a dimension or series parameter > 10) and choose to export it
- Select to save dimensions to separate files (test various combinations)
- Verify without the padded checkbox ticked that there is no zero padding
- Verify with the padded checkbox selected that zero padding is used and all dimensions are correct

bramalingam:
Both Bftools and Fiji workflows, work as expected. Good to merge.

Agree with @melissalinkert on the user configurable padding. But that can be dealt with later (if need be)

sbesson:
I am getting some `Invalid image count: 0` exce[tion when exporting a fake file `test&size=14&sizeT=11&sizeC=3.faketest&size=14&sizeT=11&sizeC=3.fake` with any of the split options checking using ImageJ with the merge JAR. @bramalingam: can you confirm? Which scenario/test file did you use exactly  with which options in your previous comment?

dgault:
Added a fix for the image count exception, I believe the issue would have been isolated to .fake files only

bramalingam:
@sbesson : I used some Nd2 files and DV files from data_repo/curated. Did not test this PR with Fake files.

sbesson:
Understood, I would suggest we retest @dgault's last commit with 1- OME-TIFF/OME-XML, 2- Fake files (as it is easy to check various dimensionality) and 3- some proprietary files (DV/ND2 is fine) and various combinations of export options to be thorough.

rleigh-codelibre:
Would it be worth making this default, and instead having a `nopad` option to disable it?  Having the filenames sort properly by default is quite desirable.

sbesson:
I am not certain we want to change the default behavior in a patch release. But using zero-padding by default is certainly worth considering for the next minor version bump. Is it something we would apply both to the command-line tools and the Image plugin?

@bramalingam: any update on the testing front?

rleigh-codelibre:
@sbesson It certainly doesn't need to happen for this release; doing it for a later release would be appreciated.

bramalingam:
Tested : 
1) ImageJ : padded and non-padded workflows for dv files : 
screen shot 2016-09-08 at 15 06 12

2)bftools : bfconvert of dv files and fake files.

And in all cases the test returned the expected results.
Looks good to merge.

Issue #2551, opened 2016-08-30
DataTools not Thread-Safe anymore


dietzc:
Hi all,

Some of the methods in https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-common/src/loci/common/DataTools.java are not thread-safe anymore since openmicroscopy/bioformats@87730bd.

Reproducible with e.g. calling https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java#L2499 from two threads at the same time.

Thanks!

Christian

dietzc:
To be more precise. https://github.com/openmicroscopy/bioformats/blob/87730bdcf8677d0e889f0e20c2c0d4278ddebe6e/components/formats-common/src/loci/common/DataTools.java#L61 is the problem.

dgault:
A PR to address this has been opened - https://github.com/openmicroscopy/bioformats/pull/2552

sbesson:
@dietzc I am sure you saw it, Bio-Formats 5.2.2 is now released and should include the regression fix. Let us know if you still encounter issues on this front and many thanks for your quick report.

dietzc:
@sbesson thank you very much :+1: Keep up the great work!

PR #2552, opened 2016-08-31
Use ThreadLocal with NumberFormat

dgault:
Issue raised in - https://github.com/openmicroscopy/bioformats/issues/2551

As NumberFormat in Java is not thread-safe it is recommended to create a separate format instance for each thread. See Synchronization under https://docs.oracle.com/javase/8/docs/api/java/text/NumberFormat.html  

To achieve this I have wrapped the NumberFormat instance in ThreadLocal to provide a separate instance for each individual thread.

To reproduce:
An example of a failing test can be found here: https://github.com/sbesson/bioformats/commit/8697cf56c575e02532b307e61858995f90b8014d

### DIFF
--- a/components/formats-common/src/loci/common/DataTools.java
+++ b/components/formats-common/src/loci/common/DataTools.java
@@ -58,8 +58,8 @@ public final class DataTools {
   // -- Static fields --
   private static final Logger LOGGER = LoggerFactory.getLogger(DataTools.class);
 
-  private static final ThreadLocal <NumberFormat> nf =
-      new ThreadLocal <NumberFormat>() {
+  private static final ThreadLocal<NumberFormat> nf =
+      new ThreadLocal<NumberFormat>() {
         @Override protected NumberFormat initialValue() {
             return DecimalFormat.getInstance(Locale.ENGLISH);
         }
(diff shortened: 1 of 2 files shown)


components/formats-common/src/loci/common/DataTools.java
```diff
@@ -58,7 +58,12 @@
   // -- Static fields --
   private static final Logger LOGGER = LoggerFactory.getLogger(DataTools.class);
 
-  private static NumberFormat nf = DecimalFormat.getInstance(Locale.ENGLISH);
+  private static final ThreadLocal <NumberFormat> nf =
```
mtbc: no space before `<`

mtbc:
Nice solution.

sbesson:
:+1: Is it worth including the unit test mentioned in the description as part of this PR to minimally cover thread safety for this API in the future (suggestion to make it more reliable welcome)?

dgault:
I think having the unit test is certainly beneficial, added it as part of this PR.

mtbc:
Nice test too, and it fails without your fix! Good to merge. :+1:

PR #2553, opened 2016-09-01
add enumeration values for acquisition mode

mtbc:
Increments OME-XML schema version attribute value and adds acquisition modes,
* BrightField
* SweptFieldConfocal
* SPIM

and parsing for,
* Laser Scan Confocal
* Swept Field Confocal

The schema namespace is left unchanged.

### DIFF
--- a/components/specification/released-schema/2016-06/ome.xsd
+++ b/components/specification/released-schema/2016-06/ome.xsd
@@ -24,7 +24,7 @@
             xmlns:xsd="http://www.w3.org/2001/XMLSchema"
             xmlns:OME="http://www.openmicroscopy.org/Schemas/OME/2016-06"
             xmlns:xml="http://www.w3.org/XML/1998/namespace"
-            version="1"
+            version="2"
             elementFormDefault="qualified">
 
   <xsd:import namespace="http://www.w3.org/XML/1998/namespace" schemaLocation="http://www.w3.org/2001/xml.xsd"/>
--- a/components/specification/released-schema/2016-06/ome.xsd
+++ b/components/specification/released-schema/2016-06/ome.xsd
@@ -670,11 +670,9 @@
         <xsd:simpleType>
           <xsd:restriction base="xsd:string">
             <xsd:enumeration value="WideField"/>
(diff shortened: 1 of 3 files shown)


rleigh-codelibre:
Looks good to me in principle.

However, do we want to add new enumerations to the end of the list to preserve ABI?  We don't have a fixed ABI for C++, so probably not too important there at this point in time, but it would break the ABI by changing the enum values. How do enum additions affect Java?  Not necessarily something which needs affect this PR, but something to bear in mind for how we handle ABI changes in the future.

sbesson:
@rleigh-codelibre thanks for raising the ABI considerations and agreed that we might want to investigate how to minimize breakages in the future - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-generated-sources/lastSuccessfulBuild/ for the current enum generated changes on C++ and Java.

rleigh-codelibre:
The latest changes look fine, and the generated source changes show all enums are appended which will preserve the existing enum values.

This looks good to merge, thanks.

hflynn:
Will this require updates to the Model docs too @mtbc ?

mtbc:
Just the autogenerated really, though it's probably also worth a line in "what's new"; we'll catch that in the routine diffs I suggested.

As for the more general issue of extending enumerations, probably the docs checklist item on https://trello.com/c/0zNp8HvU/167-enumerations-addition will cover it, @sbesson?

sbesson:
Documentating the changes in the version 2 of the 2016-06 schema certainly sounds reasonable, I'll add a note to the checklist. Otherwise all tests are passing including the integration testing at the OMERO level which auto-detects new enumeration. Parsing the logs of the [last full repository job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-full-repository/293/ for improperly parsed values shows that the new enumeration are properly picked up.

```
sbesson@lifesci-51950:target $ grep WARN.*Unknown * | wc -l
  171504
sbesson@lifesci-51950:target $ grep WARN.*Acquisition * | wc -l
       0
```

PR #2554, opened 2016-09-01
IniParser unit tests

sbesson:
See https://trello.com/c/RsYkLAFK/18-tests-add-ini-parser-unit-tests

Following https://github.com/openmicroscopy/bioformats/pull/2537, this PR adds a new unit test class covering the various forms of INI files currently handled by the `loci.common.IniParser` class. It also fixes the warnings in `IniParser` reported by `mvn findbugs:check` by refactoring the header parsing logic.

To test this PR, check Travis is green and/or run the Maven tests on `formats-common` locally and review whether all use cases of the implementation in `IniParser.parseINI(BufferedReader)` are unit tested.

### DIFF
--- a/components/formats-common/src/loci/common/IniParser.java
+++ b/components/formats-common/src/loci/common/IniParser.java
@@ -143,18 +143,18 @@ public class IniParser {
       }
 
       // check for chapter header
-      if (isHeader(line, "{")) {
-        chapter = parseHeader(line, "{", "}");
+      if (isHeader(line, '{')) {
+        chapter = parseHeader(line, '{', '}');
         continue;
       }
 
       // check for section header
-      if (isHeader(line, "[")) {
+      if (isHeader(line, '[')) {
         attrs = new IniTable();
         list.add(attrs);
 
         // strip brackets
-        String header = parseHeader(line, "[", "]");
+        String header = parseHeader(line, '[', ']');
         if (chapter != null) header = chapter + ": " + header;
 
         attrs.put(IniTable.HEADER_KEY, header);
@@ -206,15 +206,17 @@ public class IniParser {
   // -- Helper methods --
 
   /** Checks whether the input line is a INI header **/
-  private boolean isHeader(String line, String start) {
-    return (line != null && line.length() > 1 && line.startsWith(start));
+  private boolean isHeader(String line, char start) {
+    return (line != null && line.length() > 1 && line.charAt(0) == start);
   }
 
   /** Parse a header line given input delimiters **/
-  private String parseHeader(String line, String start, String end) {
+  private String parseHeader(String line, char start, char end) {
     if (line == null || line.length() <= 1) return null;
-    if (!line.startsWith(start)) return null;
-    if (line.endsWith(end)) return line.substring(1, line.length() - 1);
+    if (line.charAt(0) != start) return null;
+    if (line.charAt(line.length() - 1) == end) {
+      return line.substring(1, line.length() - 1);
+    }
     return line.substring(1);
   }
 
--- a/components/formats-common/src/loci/common/IniParser.java
+++ b/components/formats-common/src/loci/common/IniParser.java
@@ -178,7 +178,7 @@ public class IniParser {
       }
       String key = line.substring(0, equals).trim();
       String value = line.substring(equals + 1).trim();
-      attrs.put(key, value);
+      if (!key.isEmpty() && !value.isEmpty()) attrs.put(key, value);
       no += num;
     }
     return list;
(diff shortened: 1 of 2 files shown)


components/formats-common/test/loci/common/utests/IniParserTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertFalse;
```
mtbc: You don't actually use this.

components/formats-common/test/loci/common/utests/IniParserTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * Unit tests for {@link loci.common.IniParser}.
+ */
+public class IniParserTest {
+  private IniParser parser = new IniParser();
+  final Charset utf8charset = Charset.forName("UTF-8");
```
mtbc: not sure how you decided which of these was `private` and which of them `final`?

components/formats-common/test/loci/common/utests/IniParserTest.java
```diff
@@ -0,0 +1,135 @@
...
+import loci.common.IniTable;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+import static org.testng.AssertJUnit.assertEquals;
```
mtbc: Why not simply `org.testng.Assert.assertEquals`? That would have the bonus of making your arguments in `assertEquals` the correct way around.

mtbc:
Overall this is looking pretty good, thank you for adding it. None of my comments are blockers.

mtbc:
```
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

sbesson:
The last commit is slightly orthogonal from the initial intent but is based on the unit tests coverage and should fix the Findbugs warnings thrown by `mvn findbugs:check` in the `IniParser` class. Happy to move it to a separate PR if that's desirable.

components/formats-common/src/loci/common/IniParser.java
```diff
@@ -210,6 +205,19 @@ public static BufferedReader openTextResource(String path, Class<?> c) {
 
   // -- Helper methods --
 
+  /** Checks whether the input line is a INI header **/
+  private boolean isHeader(String line, String start) {
+    return (line != null && line.length() > 1 && line.startsWith(start));
```
mtbc: probably either make start a `char` or use `line.length() > start.length()`

dgault:
The new tests run successfully and the code reads well and looks good.

The list of test cases appears to provide good coverage, the only cases I could think of as being untested would be the handling of invalid headers for example empty string, single char in a string, missing end char.

sbesson:
@dgault added a few more tests for some cases with single characters, empty/null string, empty string headers. https://github.com/openmicroscopy/bioformats/pull/2554/commits/35de522498152c5a1800f48748b35a9af00d8dfb also handles the special case of lines with a single `=` character.

dgault:
Latest updates and new tests all look good, that seems like pretty comprehensive testing. PR is good to merge.

sbesson:
While the test_images_good automated tests [failed](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-test_images_good/305/) earlier this morning, excluding this PR let them to [pass](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-test_images_good/307/) in a subsequent build. So my suspicion is that https://github.com/openmicroscopy/bioformats/pull/2554/commits/35de522498152c5a1800f48748b35a9af00d8dfb broke an API which some of the readers rely on. I will revert this commit, update the tests and investigate which formats are affected.

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/127/ to reproduce failures from 35de522498152c5a1800f48748b35a9af00d8dfb and https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/128/ for the success after the last commit. Removing `breaking` label.

In the long-run the proper way to address this in a more transparent way might be to work torwards various implementations of the `IniParser` and let the individual readers choose their implementation. /cc @joshmoore

joshmoore:
_nods_ i.e. this file comes under semver.

sbesson:
Merging as per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/283

PR #2556, opened 2016-09-02
Use chunkmap for Nikon ND2 files to speed up processing

csachs:
Hello,

this is a 're-open' of #1232, which I couldn't reopen after pushing the patch rebased to current develop.

It's been two years since I proposed this; and I still think that it's very useful for people working with ND2 files, especially for large files (hundreds of GB) over network (eg. 100 Mbit), the current state is … suboptimal.

Given the great speed advantage possible with this, I think it's definitely worthwhile to debug any remaining problems (@melissalinkert mentioned wrong C/Z/T demensions for certain non-public files).
Maybe with a heuristical approach the chunk map code could just be used when it's known to deliver the right results. (If no 'complete' bugfix can be found.) (However, given my knowledge of ND2 files I'm unaware of any scenarios where wrong dimensions should even be possible, except for damaged files).

I'd love to offer help with integrating this and/or working on necessary changes.

Best regards,
Christian

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -586,17 +586,12 @@ public class NativeND2Reader extends FormatReader {
           if(useChunkMap) {
             long lastImagePosition = 0;
             for(ChunkMapEntry entry : allChunkPositions.values()) {
-              if(entry.position < in.getFilePointer()) {
-                continue;
-              }
               if(entry.name.startsWith("ImageDataSeq")) {
                 lastImagePosition = entry.position;
                 imageOffsets.add(new Long(entry.position + 16));
                 imageLengths.add(new int[] {nameLength, (int)entry.length, getSizeX() * getSizeY()});
                 imageNames.add(entry.name.substring(12));
                 // assumption nameLength is constant throughout the file for image blocks!
-              } else {
-                break;
               }
             }
             in.seek(allChunkPositions.higherKey(lastImagePosition));
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -26,11 +26,7 @@
 package loci.formats.in;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Hashtable;
-import java.util.Map;
-import java.util.TreeMap;
+import java.util.*;
 
 import loci.common.ByteArrayHandle;
 import loci.common.Constants;
@@ -619,13 +615,26 @@ public class NativeND2Reader extends FormatReader {
           if(useChunkMap && chunkmapSkips == 0) {
             ChunkMapEntry lastImage = null;
 
+            // sanity check: see if the chunk we just found is actually in the chunkmap ...
+
+            long lookupPosition = in.getFilePointer() - 28;
+
+            Long lookupResult = allChunkPositions.floorKey(lookupPosition);
+
+            if(lookupResult == null || lookupResult != lookupPosition) {
+              // if not, deactivate chunkmap processing and try classic
+              useChunkMap = false;
+              in.seek(lookupPosition);
+              continue;
+            }
+
             for(ChunkMapEntry entry : allChunkPositions.values()) {
-              if(!entry.name.startsWith("ImageDataSeq")) {
+              if((entry.position + 28) < in.getFilePointer()) {
                 continue;
               }
 
-              if((entry.position + 28) < in.getFilePointer()) {
-                continue;
+              if(!entry.name.startsWith("ImageDataSeq")) {
+                break;
               }
 
               if(lastImage!=null) {
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -586,12 +586,17 @@ public class NativeND2Reader extends FormatReader {
           if(useChunkMap) {
             long lastImagePosition = 0;
             for(ChunkMapEntry entry : allChunkPositions.values()) {
+              if(entry.position < in.getFilePointer()) {
+                continue;
+              }
               if(entry.name.startsWith("ImageDataSeq")) {
                 lastImagePosition = entry.position;
                 imageOffsets.add(new Long(entry.position + 16));
                 imageLengths.add(new int[] {nameLength, (int)entry.length, getSizeX() * getSizeY()});
                 imageNames.add(entry.name.substring(12));
                 // assumption nameLength is constant throughout the file for image blocks!
+              } else {
+                break;
               }
             }
             in.seek(allChunkPositions.higherKey(lastImagePosition));
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Map;
+import java.util.TreeMap;
 
 import loci.common.ByteArrayHandle;
 import loci.common.Constants;
@@ -347,6 +348,8 @@ public class NativeND2Reader extends FormatReader {
     // better performance with the seek/skip pattern used here
     in = new RandomAccessInputStream(id, BUFFER_SIZE);
 
+    boolean useChunkMap = true; // could be deactivated here ...
+
     channelColors = new HashMap<String, Integer>();
 
     if (in.read() == -38 && in.read() == -50) {
@@ -378,6 +381,75 @@ public class NativeND2Reader extends FormatReader {
       boolean foundAttributes = false;
       boolean useLastText = false;
       int blockCount = 0;
+
+      class ChunkMapEntry {
+        public String name;
+        public long position;
+        public long length;
+
+        public String toString() {
+          return String.format("ChunkMapEntry<%s@%d(%d)>", name, position, length);
+        }
+      }
+
+      TreeMap<Long, ChunkMapEntry> allChunkPositions = new TreeMap<Long, ChunkMapEntry>();
+
+      if(useChunkMap) {
+        /* In modern ND2 files, the chunk map is stored near the end, and contains
+         * a list of blocks and their offsets. By using these offsets instead of
+         * scanning through the whole file, an enormous speed up can be achieved.
+         *
+         * Implementation: Read the chunk map beforehand, process the file as normally,
+         * once the first ImageDataSeq block is reached, add all images and skip past the
+         * image data to process remaining metadata.
+         * I haven't read through all of NativeND2Reader, but I hope to have the least
+         * chance of inadvertedly breaking something by this approach.
+         */
+        String chunkMapSignature = "ND2 CHUNK MAP SIGNATURE 0000001";
+        in.seek(in.length() - 40);
+
+        if(!in.readString(chunkMapSignature.length()).equals(chunkMapSignature)) {
+          useChunkMap = false;
+          LOGGER.info("ND2 Warning: No chunk map found!");
+        } else {
+          in.skipBytes(1);
+
+          long chunkMapPosition = in.readLong();
+          in.seek(chunkMapPosition);
+
+          int tmpLenOne = in.readInt();
+          int tmpLenTwo = in.readInt();
+          long chunkMapLength = in.readLong();
+
+          chunkMapPosition += 16 + tmpLenTwo;
+          long chunkMapEnd = chunkMapPosition + chunkMapLength;
+
+          while(in.getFilePointer() + 1 + 16 < chunkMapEnd) {
+
+            char b = (char) in.readByte();
+            while (b != '!') {
+              name.append(b);
+              b = (char) in.readByte();
+            }
+
+            ChunkMapEntry entry = new ChunkMapEntry();
+            entry.name = name.toString();
+            name.delete(0, name.length());
+
+            if(entry.name.equals(chunkMapSignature))
+              break;
+
+            entry.position = in.readLong();
+            entry.length = in.readLong();
+
+            allChunkPositions.put(entry.position, entry);
+
+            LOGGER.info("ND2 {}", entry.toString());
+          }
+        }
+
+        in.seek(0);
+      }
 
       // search for blocks
       byte[] sigBytes = {-38, -50, -66, 10}; // 0xDACEBE0A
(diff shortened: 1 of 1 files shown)


sbesson:
Hi @csachs, many thanks for resurrecting this thread. Speeding up the reader initialization using some heuristics to only target ND2 files makes complete sense.

I ran the [automated tests job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-repository-subset/120) against our sample ND2 files with this PR included. There are currently 60 failures spread across 10 files (out of 194). Unfortunately, at the moment, we are unable at the moment to share these sample files for privacy reasons although we will attempt to contact their submitters.

In the meantime, is there some metadata or debugging information which you could help in refining the assumptions of this PR?

csachs:
Hello, thanks for the link and feedback. Looking through the logs for failing files I can unfortunately not see any pattern without knowing more about the contained data…
Some errors seem to be off-by-one errors, and I just currently realize that there might be a problem in the chunk map entry processing … I always assumed that, within the ND2 file, once the first image appears, all images will show up consecutively; thus the code skips towards the last image after the first appears and everything has been added according to the chunk map.
For my files that worked, and apparently for quite many (184/194?!), but I've just seen that some files I have do have further metadata blocks after the first image data block, which would then be skipped. I have to investigate this.
Opening one or two of the non-working files with debug logging enabled should print out all chunk map entries, which might provide some hints whether this is the case in the current situation.

I'll try to fix what I think might be wrong; let's see if that helps …

Best regards, Christian

sbesson:
Hi Christian,

here are the chunk map entries for one of the failing files as shown by `showinf -nopix -debug`: [rfp_h2a_cells_02.chunkmap.txt](https://github.com/openmicroscopy/bioformats/files/455170/rfp_h2a_cells_02.chunkmap.txt). I see you have pushed an extra commit. Let me know when you would like this PR to be tested against the `nd2` sample files again.

csachs:
Thanks for the output; I guess my current patch will not change anything for this file.
If it's not too much work, feel free to re-run the version with the last commit to see if less (or more?!) errors occur. Unfortunately, I currently have no idea what the exact problem is, if it remains with the patch. I'll take a look at how dimensionality is calculated and if the chunk-map code might interfere there, but for debugging I guess I might need a file making problems (I so far did not encounter one in our archives).
Obviously, if anyone of the OME team has some time to spare, they might look into the issue as well with the files…

Regards, Christian

sbesson:
Unfortunately the last commit seems to have created some sort of infinite loop when opening one of the sample files - the full debug log is at [962_xy1.chunkmap.txt](https://github.com/openmicroscopy/bioformats/files/455272/962_xy1.chunkmap.txt).

csachs:
Okay, thanks for the feedback. That didn't went too well then, reverting the commit to keep the branch working. However, as I currently have no idea what's the problem and no example file to debug, I can't promise to come up with a solution…

sbesson:
Quick heads up. I checked some of the failing files and they indeed have non-image blocks interleaved into `ImageDataSeq` blocks. While https://github.com/openmicroscopy/bioformats/pull/2556/commits/31d08bf8df431f23d9df13d2389a54723ab3e584 makes sense, the issue described in https://github.com/openmicroscopy/bioformats/pull/2556#issuecomment-244768477 is that for this particular file the very first `entry.position` at 90112 is found after the file pointer is at 90140 so `lastImagePosition` is never updated. and the input stream is reset to position 0.

sbesson:
Wrong button. Apologies.

csachs:
Before I revert the commit again, I wonder, did the changed version (which led to the infinite loop in one file) maybe worked better on other files?

sbesson:
@csachs unfortunately no, 31d08bf alone is not does not fix the failures reported above. I will give another try after adding some extra logging statements later in the week and get back to you.

sbesson:
Hi @csachs, apologies for the delay on answering this thread. We have been busy working on the development of [Bio-Formats 5.3](http://blog.openmicroscopy.org/data-model/future-plans/2016/11/01/bf-status/) for the last two months.

We contacted some of the owners of our sample ND2 filesets and asked them if we could make their data data. We received positive feedback from those we managed to join. After agreement on the licensing (CC-BY), we will be making ND2 files available under the canonical http://downloads.openmicroscopy.org/images/ND2 URL.

As of today, one fileset should already be available. Following https://github.com/openmicroscopy/bioformats/pull/2556#issuecomment-244731165, the `rfp_h2a_cells_02.nd2` and `b16_pdtB+y50__crop.nd2` files should allow to reproduce the failures detected by the non-regression tests with the latest state of this PR.

csachs:
Thanks @sbesson !
I will take a look :)

csachs:
Okay, couldn't resist but debugged a bit about these issues. First – these are broken files:

`rfp_h2a_cells_02.nd2` has a minor problem; apparently plane 973 wasn't added to the chunk map somehow, the old code obviously skipped it, leading to wrong plane counts leading to wrong calculated dimensionalities. The fix now checks if the gap between any two adjacent chunks is large enough for a plane to fit in, and if so, runs the normal block scanning code.

`b16_pdtB+y50__crop.nd2` is worse, the positions from the chunk map and the real positions don't correspond. This leads to a general problem: If a chunk map is there, but it's wrong, one is in obivous trouble: Either one trusts it, which can lead to problems accessing the images, or one checks it, and in the worst case checks every single position, then there is no advantage of having it in the first place.
I've mitigated this somewhat by performing a heuristic check, 10 blocks equally spaced in the file are read and checked if the positions are sane. If not, the chunk map processing is deactivated. Obviously, if one does not hit a broken block right away, this might still lead to broken processing.

Can you run your automated checks to see if this mitigates all the remaining problems?

For the case of broken files, I would propose a generally different approach: Have the old mode as 'recovery'-mode, activatable via configuration; so if someone has broken files, they still may open them, but correct files can still be processed fast. Via the GUI, there seems to be these `components/bio-formats-plugins/src/loci/plugins/util/LociPrefs.java` settings, are they generally (eg. with the bio-formats cli tools, or embedded in OMERO) accessible? So far they appear to be ImageJ specific. Is this still the preferred way to add options? It seems to greatly contradict separation/modularization efforts to have the reader in one file, it's possible settings somewhere completely different ...

Best regards,
Christian

sbesson:
Hi @csachs, thanks for the extra commit. I reluanched the non-regression tests and it seems there are still [a few failures](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/160/) although the total count is decreasing. We'll have a look at the samples, see if anything else can be made public and get back to you regarding the mismatching chunkmaps.

On the second part of your answer, the `LociPrefs` was designed in the context of the ImageJ plugins. The generic API to pass options to the reader is the [MetadataOptions](http://downloads.openmicroscopy.org/bio-formats/5.2.4/api/loci/formats/in/MetadataOptions.html) API. Quite timely, in the context of Bio-Formats 5.3.0, we have been working on extending our default implementation of this interface to support generic key/value pairs - see https://github.com/openmicroscopy/bioformats/pull/2679. It would be interesting to know if this is something worth consuming here?

csachs:
Hi @sbesson , thanks for running the tests. Somewhat sad to see that there are still problems, I assume the files [1] might be broken as well. I wonder how to proceed best now, would it be an option to add (somewhat log-spamming) debugging code to this branch, and have you run it on the test datasets?
An approach would be to print the chunk map block positions, as wells all whole-file-scan block positions by the regular bioformats code, and compare that. Given that discrepancies there have caused the former two files to fail, I have some hope that it might still be a problem, somehow.

I'll take a look in the `MetadataOptions` API, ultimately it depends on if you/the bioformats team thinks this alternative behavior should be controlled via user configuration.

[1] `/qa-7534/130606_JCC719_003.nd2` `/qa-9507/control003.nd2` `/nd2/guillaume/Lig_drugs051607.nd2` `/nd2/karsten/100720104_TestglasSchott_MM63_C10h.nd2` `/nd2/benjamin/B2.nd2`

sbesson:
We just published another [set of ND2 samples](http://downloads.openmicroscopy.org/images/ND2/jonas/) including the `130606_JCC719_003.nd2` file mentioned and I will look into the ownership of the remaining ones. If this is still insufficient to fix the remaining issues, adding some debugging statements to compare the chunk map code to the whole-file-scan block positions might certainly be a possibility.

csachs:
Let's try it again ... feel free to restart the test suite once Travis is OK.
In the case of `130606_JCC719_003.nd2`, the file contained blocks not present in the chunk map …

sbesson:
Thanks for the last commits. So we're down to [one remaining failing file](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/162). I did not manage to successfully contact the data owner but I will give it another try. If that does not suffice, we can look into enabling some debugging code in this PR or a separate branch.

csachs:
Thanks for running the tests. `/benjamin/B2.nd2`, right? It has two `ND2 FILEMAP `, which is probably rather strange. Can you runthe code with debug enabled, e.g. via `showinf -debug`, so I can take a look at the chunk map debugging statements? My first workaround attempt would be to disable chunk map processing if two chunk maps are found (so it would be interesting to see if the first chunk map is stored as a chunk in the last chunk map).
By the way, how is ground truth (dimensionality etc.) established for these? I guess many of these would not properly open in NIS Elements…

sbesson:
@csachs here is the output of the debug log while opening the nd2 file: [B2.nd2.txt](https://github.com/openmicroscopy/bioformats/files/636252/B2.nd2.txt).

With regard to the determination of the ground truth, in most cases we use NIS Elements Viewer for the ground truth, e.g. for `130606_JCC719_003.nd2`

screen shot 2016-12-07 at 11 05 09

csachs:
Hmm, thanks for quickly running this. Unfortunately with the current debugging output I can't spot the problem … there are some discrepancies between the found blocks and the chunk map entries, but only in the non-image blocks, which are handled by the normal code anyways.
So adding debug output would be an option? What would be most convenient for you,  a commit here, or another branch?

sbesson:
@csachs whatever is easier for you. No objection to pushing a commit here. To some extent, the commit history can also serve as a documentation of the investigation leading to the (hopefully) final set of change.

csachs:
Okay, please just run the new code on the non-working file in with debug output enabled.
I've split the `initFile` method into two different subroutines, because it contained code for reading completely different, unrelated formats ...
(The current commit calls the reader code twice to produce extensive debug information …)

In my opinion, the file generally badly needs refactoring … it feels very spaghetti-esque, and contains many deprecated Java patterns. (Based on a commit I spotted, Java 1.7 is the new minimal version bioformats supports, right? So changing many parts to 1.7 Java should be ok?)
If some refactoring commits would be welcome, I'd look into this at some later time, as a different PR.

sbesson:
@csachs yes the minimum Java version has been set to 1.7 since Bio-Formats 5.2.0.  [This file](https://github.com/openmicroscopy/bioformats/files/639346/B2.nd2.txt) contains the output of the following command with your last commits:

```
showinf -nopix -debug -nometa B2.nd2
```

csachs:
Thanks for running it. This file is rather broken as well. I doubt the image contents of plane 3 and 6 (zero-based) is meaningful? It overlaps with the next image.
```
* Parsing ImageDataSeq|0! at 81936 length 8686088 (new blockcount 6)
* Parsing ImageDataSeq|1! at 8773648 length 8686088 (new blockcount 7)
* Parsing ImageDataSeq|2! at 17481744 length 8686088 (new blockcount 8)
* Parsing ImageDataSeq|3! at 26165264 length 8686088 (new blockcount 9) <--
* Parsing ImageDataSeq|4! at 34856976 length 8686088 (new blockcount 10)
* Parsing ImageDataSeq|5! at 43556880 length 8686088 (new blockcount 11)
* Parsing ImageDataSeq|6! at 52240400 length 8686088 (new blockcount 12) <--
* Parsing ImageDataSeq|7! at 60932112 length 8686088 (new blockcount 13)
* Parsing ImageDataSeq|8! at 69623824 length 8686088 (new blockcount 14)
```
C.f. 'distance' between beginning of ImageDataSeq|3 and ImageDataSeq|2!:
```
26165264 - 17481744 = 8683520
8683520 < 8686088 (image block size)
```

I have adapted the heuristical image block finder to cater for this as well.
Personally, I think that this is rather something which should be addressed by regular vs. recovery mode. All the files which made problems seem to be broken in one or the other way.
You could try to run the version of the last commit again on the Travis automated testing suite.

sbesson:
I re-rean the jobs and it seems the last change [broke another file](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/165/). Here is the output of the
[B2.nd2 log](https://github.com/openmicroscopy/bioformats/files/641927/B2.nd2.txt) file with the latest commit.

csachs:
Had some days off … mh, bad to hear. The output for `B2.nd2` is not correct I assume? I don't know how it should look like … unfortunately, I had removed/commented out the debugging code, so an approach now would be to repeat the cycle, add back in the debugging code, run it for these files and see what the last commit changed, and if this was at all a reasonable approach…

sbesson:
Sorry for the delay @csachs , we have been busy working on our Bio-Formats 5.3.1 release. To have a precise estimate of the impact of the current state of this PR, I temporarily excluded the file failing the `setId` https://github.com/openmicroscopy/bioformats/pull/2556#issuecomment-265986293 and re-ran the [automated tests](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/168/). It turns out that the `B2.nd2` file is now passing the tests but two other files in our repository are now failing this test (i.e. 3 in total incl. the excluded file).

We certainly agree many of these problematic files are in a weird if not broken state. That being said, maintaining support for the large range of ND2 files that have been submitted to us throughout the years is part of the contract of this library.

I understand that working on this complex format without having access to the failing files is extremely complex. As there is already substantial changes and we may have some bandwidth to look at it again, I would propose the following:

- create a dedicated integration branch e.g. `nd2_chunkmap`
- change the base branch of this PR and integrate it this integration branch
- let us open a PR trying to address the remaining file and move the discussion there 

Main question is whether we should use the latest state of this PR or revert to e.g. 759511e which had only one file to fix. Thoughts?

csachs:
No problem; and this approach sounds very reasonable. I would propose 759511e aswell for that branch then. (However the question might still be how B2.nd2 can be addressed in a sane way…)
But I'll probably not be able to proceed with bug fixing during the holidays.

sbesson:
Sounds good. Can you force push away the last set of commits (maybe keep them in a separate branch if we need to come back to this)?

csachs:
Done. I'll keep a backup in nd2-chunkmap-patch-backup ( csachs/bioformats@f196082a13a8a9e28271e31c0cc49e1f0bec168d ) while we work on this.

sbesson:
Thanks @csachs, merging this into the integration branch. @melissalinkert: I will review the semantics of the [CI job](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/) to include this branch and its PRs in the merge command.

PR #2557, opened 2016-09-02
desynchronize string buffers

mtbc:
Bio-Formats is full of `StringBuffer` which is synchronized. This PR replaces many readers' use of these with `StringBuilder`.


rleigh-codelibre:
The changes all look fine to me.

Does any specific testing need doing here, or is checking the job status sufficient?

mtbc:
I'd say the full repo job should suffice.

PR #2559, opened 2016-09-05
switch to String.indexOf(int) in various classes including Metakit

mtbc:
Inspired by https://github.com/openmicroscopy/bioformats/pull/2528#discussion_r76218340 -

> this is really just about switching to calling a method that can assume searching for a single character instead of having to care about search string length

this is a mechanical change so that a simpler library method can be used. It may not make much difference but sets a good example for others writing further code based on our existing code. https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-full-repository/ suffices for testing.

In the event of conflicts I'll remove the problem changes from this PR.

### DIFF
--- a/components/formats-api/src/loci/formats/ClassList.java
+++ b/components/formats-api/src/loci/formats/ClassList.java
@@ -176,14 +176,14 @@ public class ClassList<T> {
   public void parseLine(String line)
    {
       // Ignore characters following # sign (comments)
-      int ndx = line.indexOf("#");
+      int ndx = line.indexOf('#');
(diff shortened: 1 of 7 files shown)


rleigh-codelibre:
This looks good to merge once the full-repo test for today is green (might be tomorrow before completion).

sbesson:
Merging as per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/281/

PR #2123, opened 2015-11-27
InCell: attempt to find TIFF files in Analyzer 2000 datasets

melissalinkert:
Filenames won't necessarily be given in the .xdce file, so this attempts
to guess at the correct file names.  This won't cover all cases, but
should work as a place to start.  It also fixes field position
population to allow for positions being recorded out of order.

See https://trello.com/c/9V2wbtvD/139-idr0017-breinig-drugscreen

PR #2562, opened 2016-09-05
InCell: attempt to find TIFF files in Analyzer 2000 datasets (develop)

simleo:
Rebases #2123 onto develop.

I've tested locally against a small subset of our InCell data and got no errors. If this does not break the build, we can move forward and add a new test case from idr0017.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -91,7 +91,7 @@ public class InCellReader extends FormatReader {
 
   private int wellRows, wellCols;
   private Map<Integer, int[]> wellCoordinates;
-  private List<Length> posX, posY;
+  private Map<Integer, Length> posX, posY;
 
   private boolean[][] exclude;
 
@@ -106,6 +106,9 @@ public class InCellReader extends FormatReader {
   private Double gain;
   private Double temperature;
   private Double refractive;
+
+  private List<String> exFilters = new ArrayList<String>();
+  private List<String> emFilters = new ArrayList<String>();
 
   // -- Constructor --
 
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/InCellReader.java
```diff
@@ -1018,9 +1063,10 @@ else if (qName.equals("Exposure") && openImage) {
       else if (qName.equals("offset_point")) {
         String x = attributes.getValue("x");
         String y = attributes.getValue("y");
+        int index = Integer.parseInt(attributes.getValue("index"));
```
sbesson: This line fails some of the [automated test](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/286/consoleText) with a NPE. It might be worth using `DataTools.parseInteger` and handle the case where `index == null`.

simleo:
@sbesson looks like the original PR assumed that `index` is always present in `offset_point`. We can add a static counter and fall back to the old behavior when the attribute is missing. This should work fine when *all* `offset_point` elements are missing the attribute, otherwise the counter will probably override some indices in an unpredictable fashion. We might argue that in this case the file can be considered malformed though. BTW I don't think we are testing for `offset_points` being in the order specified by `index` attributes anyway. One of the datasets that passed the test on my laptop has an unsorted sequence of `offset_point` elements in it.

I'll mark this for exclusion while the bug is being sorted out.

simleo:
--exclude

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/287/ :+1:

simleo:
Test case: openmicroscopy/data_repo_config#128

sbesson:
Merging as per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/281/ and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/289/ (IDR dataset)

PR #2564, opened 2016-09-07
test: quantity: Split into multiple object files

rleigh-codelibre:
The number of templates exceeded the symbol limit on some platforms when built with debug symbols.

Fixes https://trello.com/c/yFMSI6L1/24-re-enable-brill

Testing: Check all builds are green.  Will fix the failing debug build on FreeBSD (though I'm surprised it didn't also break Windows which has a similar limit).

### DIFF
--- a/cpp/test/ome-xml/CMakeLists.txt
+++ b/cpp/test/ome-xml/CMakeLists.txt
@@ -83,7 +83,17 @@ if(BUILD_TESTS)
   target_link_libraries(model OME::XML)
   target_link_libraries(model ome-test)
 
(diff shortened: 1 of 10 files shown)


sbesson:
Today's builds are passing on all platforms including Windows and FreeBSD. The console output correctly indicates that a set of tests is executed for each supported quantity. Once https://ci.openmicroscopy.org/view/Failing/job/OME-FILES-CPP-DEV-merge-superbuild/361/ completes successfully (previous failure is independent of this PR), this should be ready to merge.

Issue #2238, opened 2016-02-10
Bio-Formats Exporter can deadlock on OS X


ctrueden:
When attempting to reproduce #2237 on OS X, the Bio-Formats Exporter dialog never appears on my OS X 10.11 system. Instead, I get a deadlock:
```
Found one Java-level deadlock:
=============================
"Bio-Formats Exporter":
  waiting to lock monitor 0x000000011a806048 (object 0x00000006444d46d8, a java.lang.Object),
  which is held by "AWT-EventQueue-0"
"AWT-EventQueue-0":
  waiting to lock monitor 0x000000011a804e68 (object 0x000000066aa10ff8, a java.awt.Component$AWTTreeLock),
  which is held by "Bio-Formats Exporter"

Java stack information for the threads listed above:
===================================================
"Bio-Formats Exporter":
	at com.apple.laf.AquaFileSystemModel.getRowCount(AquaFileSystemModel.java:194)
	- waiting to lock <0x00000006444d46d8> (a java.lang.Object)
	at javax.swing.JTable.getRowCount(JTable.java:2662)
	at javax.swing.plaf.basic.BasicTableUI.createTableSize(BasicTableUI.java:1692)
	at javax.swing.plaf.basic.BasicTableUI.getPreferredSize(BasicTableUI.java:1733)
	at javax.swing.JComponent.getPreferredSize(JComponent.java:1661)
	at javax.swing.ScrollPaneLayout.preferredLayoutSize(ScrollPaneLayout.java:495)
	at java.awt.Container.preferredSize(Container.java:1788)
	- locked <0x000000066aa10ff8> (a java.awt.Component$AWTTreeLock)
	at java.awt.Container.getPreferredSize(Container.java:1773)
	at javax.swing.JComponent.getPreferredSize(JComponent.java:1663)
	at java.awt.BorderLayout.preferredLayoutSize(BorderLayout.java:719)
	- locked <0x000000066aa10ff8> (a java.awt.Component$AWTTreeLock)
	at java.awt.Container.preferredSize(Container.java:1788)
	- locked <0x000000066aa10ff8> (a java.awt.Component$AWTTreeLock)
	at java.awt.Container.getPreferredSize(Container.java:1773)
	at javax.swing.JComponent.getPreferredSize(JComponent.java:1663)
	at java.awt.BorderLayout.preferredLayoutSize(BorderLayout.java:719)
[...]

PR #2566, opened 2016-09-08
ICS Writer - incorrect planes

dgault:
This issue was raised via the Image J forum - http://forum.imagej.net/t/release-of-bio-formats-5-2-0/2523/9

When exporting to ics format the ordering of the image dimensions was incorrect with images from different channels being displayed in the wrong position.

This issue started occuring after other fixes to the ICSWriter in this PR: https://github.com/openmicroscopy/bioformats/pull/2315

The original issue that the PR was solving was: https://github.com/openmicroscopy/bioformats/issues/2237

Note there is an additional issue which can lock up the exporter, this is still open and is not aimed to be fixed yet - https://github.com/openmicroscopy/bioformats/issues/2238

**To reproduce:**
- Import a suitable multi dimensioned image using Bio-Formats
- Export the image to the ICS format
- Open the original image and the ICS image and compare the image planes to verify that they differ

**To test:**
Ensure that the test cases from the original PR pass:
1) Tried checking the same macro in Fiji https://github.com/openmicroscopy/bioformats/issues/2237 to export the file with a splitC to non ".ome" extensions and the export worked. (This meant the issue was not at the exporter level)

2) Tried bfconvert with splitC to convert it into ".ics" files. Even though this does not throw any exceptions, it still does not do the conversion properly. Ensure each channel file is correctly generated.

Additional the following test cases should be used:
3) Import a suitable multi dimensioned image using Bio-Formats
- Export the image to the ICS format
- Open the original image and the ICS image and compare the image planes to verify they match

4) Import a suitable multi dimensioned image using Bio-Formats
- Export the image to the ICS format using combinations of split dimensions
- Open the original image and each ICS dimension image to verify the files have been generated correctly

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/ICSWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/ICSWriter.java
@@ -163,7 +163,7 @@ public class ICSWriter extends FormatWriter {
         pixels.skipBytes(bytesPerPixel * rgbChannels * (sizeX - w - x));
       }
     }
-    lastPlane = no;
+    lastPlane = realIndex;
     if (lastPlane != getPlaneCount() - 1 || uniqueFiles.size() > 1) {
       overwriteDimensions(getMetadataRetrieve());
     }
--- a/components/formats-bsd/src/loci/formats/out/ICSWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/ICSWriter.java
@@ -164,9 +164,7 @@ public class ICSWriter extends FormatWriter {
       }
     }
     lastPlane = realIndex;
-    if (lastPlane != getPlaneCount() - 1 || uniqueFiles.size() > 1) {
-      overwriteDimensions(getMetadataRetrieve());
-    }
+    overwriteDimensions(getMetadataRetrieve());
 
     pixels.close();
     pixels = null;
--- a/components/formats-bsd/src/loci/formats/out/ICSWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/ICSWriter.java
@@ -342,7 +342,7 @@ public class ICSWriter extends FormatWriter {
 
     if (rgbChannels > 1) {
       dimOrder.append("ch\t");
-      sizes[nextSize++] = pos[1] + 1;
+      sizes[nextSize++] = rgbChannels;
     }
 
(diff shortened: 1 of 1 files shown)


bramalingam:
Tested this PR on : smb://squig.openmicroscopy.org.uk/ome/data_repo/test_images_good/dv/CFPNEAT01_R3D.dv

And the PR does not make a difference with BF5.2.1:
1) The export with default options, has the dimensions wrong (1 Z-section missing) and the first Z-section of Channel 2 is wrong as well.
2) The export with split-channels selected exports the image properly.

sbesson:
@dgault did you build the artifacts locally? Given the opening timestamp, this PR was not included into this morning's merge build.

dgault:
This appears to be a new issue discovered. It is present without this PR and manifests itself differently with the PR. Its a bug that we should probably aim to fix as part of this PR if possible. I will attempt to get a resolution ready for retesting tomorrow.

bramalingam:
@sbesson : Nope downloaded the artifacts from the CI server. Sorry for the late response, somehow missed your comment.

dgault:
Last commit will fix the additional bug which Balaji found while testing. The bug related to the image dimensions being written in the ics metadata.

Without this PR the bug would export with a size Z one smaller that actual, resulting in the last z plane appearing missing but being displayed as the first plane of the next channel. With the PR it was resulting in the channel size being one smaller than actual.

sbesson:
Note the last commit failed the minimal conversion tests under https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/.

sbesson:
In addition to the functional testing, is there room for having the scenario of this PR included in the test coverage? Looking minimally in the code, the most appropriate place testing the writer logic at the moment might be the [ConversionTest](https://github.com/openmicroscopy/bioformats/blob/v5.2.2/components/test-suite/src/loci/tests/testng/ConversionTest.java) which is currently executed in an Ant-only target of the `test-suite` component. Alternatively should we think of refactoring the testing of our writers?

bramalingam:
Tested it with the same dv file (using ImageJ) from test_images_good(CFPNEAT01_R3D). Works as expected. Looks good to merge.

PR #1819, opened 2015-05-27
Automated tests: add support for external configuration directory

sbesson:
Currently automated tests are relying upon configuration files (`.bioformats`) being present under the same tree as the data repository. The goal of this infrastructure PR is to remove this requirement and allow external configuration tree to bee used to test the data repository. Ultimately, this should simplify our workflow when modifying readers and configuration files /cc @jburel

To test this PR:

- check the overall set of changes is sensible. Some refactoring in the test classes to use helper methods and hide parameters of type `${property}` set by `ant`
- check the existing test logic is untouched without specifying a configuration directory
  ```
ant test-config -Dtestng.directory=/path/to/data  # should create/append .bioformats files in-place
ant test-automated -Dtestng.directory=/path/to/data  # should test using local .bioformats files
```
- test the configuration files can be generated under a custom directory using `test-config`
  ```
ant test-config -Dtestng.directory=/path/to/data -Dtestng.configDirectory=/path/to/config 
```
Check that for each section, the entries of `.bioformats` file should now be sorted by key. This is meant to simplify diff examination when updating these files.
- modify the created `.bioformats` file under `/path/to/config` and test the new ant variable picks up this configuration file instead of the local file under the data directory tree using
  ```
ant test-automated -Dtestng.directory=/path/to/data -Dtestng.configDirectory=/path/to/config
```

Other scenarios that can be tested:
- test a more complex tree (multiple level of hierarchies under `/path/to/data`)
- test multiple data folders
- test removing the `.bioformats` files under the data directory with the `configDirectory` property set does not affect the tests (for future data repo cleanup)

PR #2574, opened 2016-09-13
FormatReaderTestFactory: list files as returned by getCurrentFile

simleo:
This PR addresses a problem I've encountered while building a test case based on a plate from `idr0009-simpson-secretion`. The dataset is structured as follows:

```
AcquisitionLog.dat
data/
.DS_Store
experiment_descriptor.dat
experiment_descriptor.xml
```

Where `data` contains the individual TIFF files. The configuration file generated with `gen-config` looked fine:

```
[experiment_descriptor.xml global]
...

[experiment_descriptor.xml series_0]
[experiment_descriptor.xml series_1]
...
```

However, the test failed, complaining about `AcquisitionLog.dat` not being configured. Here is a breakdown of what happens:

* To build the list of files to be tested, `FormatReaderTestFactory` walks the entire tree and builds an initial list with everything in it (almost everything, as .DS_Store is immediately discarded as a hidden file); then it walks down that list in order, and uses `getUsedFiles` to remove redundant files (i.e., those that are already taken into account by readers of files above them).

* Due to the black magic performed by `ScanrReader`, all files in the set are actually found by pointing `setId` to any of them. Since `AcquisitionLog.dat` is the first one, all other files are discarded.

* `FormatReaderTest` receives `AcquisitionLog.dat` as id, but uses `getCurrentFile` to write the config file. `ScanrReader`'s wisdom, however, ensures that no matter what was passed to `setId`, `getCurrentFile` always returns the path of the XML file in the set.

* Since the above list is rebuilt while running tests, the only file listed is, again, `AcquisitionLog.dat`. This is not mentioned in the config file, though, so the test fails.

This PR fixes (in this case, at least) the problem by using `getCurrentFile` to represent test targets in the list, so that there is no mismatch when this is compared with config contents. However, it seems weird to rebuild the list while running tests. A more long-term solution would be to change the code so that only what's written in the config file is considered while actually running tests.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -30,7 +30,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Arrays;
 
 import loci.common.DataTools;
@@ -202,14 +202,14 @@ public class FormatReaderTestFactory {
     }
 
     // remove duplicates
-    Set<String> fileSet = new HashSet<String>(files);
-    Set<String> minimalFiles = new HashSet<String>();
+    Set<String> fileSet = new LinkedHashSet<String>(files);
+    Set<String> minimalFiles = new LinkedHashSet<String>();
     FileStitcher reader = new FileStitcher();
     while (!fileSet.isEmpty()) {
       String file = fileSet.iterator().next();
       try {
         reader.setId(file);
-        Set<String> auxFiles = new HashSet<String>(
+        Set<String> auxFiles = new LinkedHashSet<String>(
             Arrays.asList(reader.getUsedFiles())
         );
         fileSet.removeAll(auxFiles);
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -29,6 +29,9 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.Arrays;
 
 import loci.common.DataTools;
 import loci.formats.FileStitcher;
@@ -70,7 +73,7 @@ public class FormatReaderTestFactory {
 
   @Factory
   public Object[] createInstances() {
-    List files = new ArrayList();
+    List<String> files = new ArrayList<String>();
 
     // parse explicit filename, if any
     final String nameProp = "testng.filename";
@@ -199,19 +202,18 @@ public class FormatReaderTestFactory {
     }
 
     // remove duplicates
-    int index = 0;
+    Set<String> minimalFiles = new HashSet<String>();
     FileStitcher reader = new FileStitcher();
-    while (index < files.size()) {
-      String file = (String) files.get(index);
+    for (String file: files) {
       try {
         reader.setId(file);
-        String[] usedFiles = reader.getUsedFiles();
-        for (int q=0; q<usedFiles.length; q++) {
-          if (files.indexOf(usedFiles[q]) > index) {
-            files.remove(usedFiles[q]);
-          }
-        }
-        files.set(index, reader.getCurrentFile());
+        Set<String> auxFiles = new HashSet<String>(
+            Arrays.asList(reader.getUsedFiles())
+        );
+        String masterFile = reader.getCurrentFile();
+        auxFiles.remove(masterFile);
+        minimalFiles.removeAll(auxFiles);
+        minimalFiles.add(masterFile);
       }
       catch (Exception e) { }
       finally {
@@ -220,9 +222,8 @@ public class FormatReaderTestFactory {
         }
         catch (IOException e) { }
       }
-
-      index++;
     }
+    files = new ArrayList(minimalFiles);
 
     // create test class instances
     System.out.println("Building list of tests...");
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -202,14 +202,17 @@ public class FormatReaderTestFactory {
     }
 
     // remove duplicates
+    Set<String> fileSet = new HashSet<String>(files);
     Set<String> minimalFiles = new HashSet<String>();
     FileStitcher reader = new FileStitcher();
-    for (String file: files) {
+    while (!fileSet.isEmpty()) {
+      String file = fileSet.iterator().next();
       try {
         reader.setId(file);
         Set<String> auxFiles = new HashSet<String>(
             Arrays.asList(reader.getUsedFiles())
         );
+        fileSet.removeAll(auxFiles);
         String masterFile = reader.getCurrentFile();
         auxFiles.remove(masterFile);
         minimalFiles.removeAll(auxFiles);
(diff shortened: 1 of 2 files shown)


sbesson:
Briefly tested via https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/132/console after reverting one of the configuration commits manually excluding some files/folders for `cellr` datasets. A couple of immediate issues/thoughts:
- the  logic works fine as long as the grouping is independent of the selected file (which might be the case for `incell`). In the case of the build above, the TIFF files from the subfolders are scanned first but they are not grouped with the `.apl` file in the directory above causing "not configured" error. I wonder if ordering by files first and then going into subfolders would be sufficient
- as shown by the build, some old configuration files will need to be updated as their headers do not use `getCurrentFile()` systematically. Having a full run of the full-repository would give us a more general idea of the impact.
- the final proposal in the PR description (using the config tree instead of scanning the file system) certainly makes sense, has clear advantages and is worth discussing before we put more effort into this.

simleo:
Not knowing that `ScanrReader` could discover all files starting from any of them, my initial idea for processing the list was to:

* immediately discard "unreadable" files (i.e., not matched by any reader), since those would either be not relevant or part of a set whose "master" file is somewhere else in the list. I thought `AcquisitionLog.dat` could be discarded in this way because it is unreadable if isolated, but in this case it is read and drags all other files along with it. From what you wrote, it looks like this strategy might work in the `cellr` case though.

* build a set of all file names that are just "parts" of a larger set, then walk the list again and discard any file that can be found in that set. In this case, however, if you feed **any** file to the reader, **all** files are returned by `getUsedFiles`, so if you're not careful you risk discarding everything. This is where `getCurrentFile` might help, as a way to ask the reader "which of these is the master file?".

I think mixing the two approaches might work, and testing both on `incell` and `cellr` should be fine as a preliminary check.

simleo:
I've tested one of the failing `cellr` samples (`colin/07-06-21`) locally. Although the test does fail if the current config file is used, after rebuilding the config file *with this PR* it passes. So maybe this PR works in other cases too, but there has to be consistency between config generation and test execution. Here is the diff between (one entry of) the current config file and the one generated with this PR:

```diff
-[07-06-21.apl series_0]
+[07-06-21_d.mtb series_0]
 ChannelName_0 = null
+Description = null
 DimensionOrder = XYCZT
 FalseColor = true
 Indexed = false
@@ -28,11 +29,13 @@
 SizeZ = 1
 ThumbSizeX = 128
 ThumbSizeY = 128
+Tile_MD5 = 7d6f2abc48abb7f28545f1f50a3ad2b4
 channel_count = 1
-header = 07-06-21.apl series_0
+header = 07-06-21_d.mtb series_0
```

The file name is the `.mtb` one (the one returned by `getCurrentFile`), and there are two new entries (presumably added after the config was generated). Another difference is that the current config file does not have sorted entries in each section, so I had to do the sorting myself in order to generate the above diff.

How hard would it be to re-generate all config files according to this PR, so that it can be evaluated in a consistent environment?

sbesson:
I have mixed thoughts on regenerating all the config files,. Given the data repository size, I would expect the config diff to be large and fairly unreviewable except as per the repository tests outcome. On the other hand, we would have start from a cleaner state with less hardcoded exclusion and the files would also benefit from other improvements e.g. ordered keys per section. 

Few suggestions:
- how would we discriminate between cases where individual files were excluded because of the `test-suite` API limitation (cf `cellr` above) or excluded on purpose instead (redundancy, unsupported fileset)?
- as this issue primarily affects multi-file formats, should we target this subset first?

simleo:
In f77849a I've changed the logic so that "auxiliary" files (i.e., all used files except for the "master" one as returned by `getCurrentFile`) are discarded no matter what order they appear in the initial list. This allows to simplify config files (less need of `test = false` entries) and might also fix the current `cellr` problems by itself, although I think it's more likely that we'll have to regenerate (at least part of the) config files anyway.

simleo:
In c037f1b auxiliary files are also removed from the original collection as soon as they are detected. In cases where the reader picks up all files starting from an aux file (e.g., `scanr`) this avoids a potentially high number of unnecessary calls to `setId` (other aux files are not considered at all).

simleo:
Re: 2a8494b: I was using this locally to fix an unhandled OOM during the recursive search for config files that led to a cryptic `The factory method class loci.tests.testng.FormatReaderTestFactory.createInstances() threw an exception` message from TestNG. Since the latest build [failed with the same message](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-test_images_good/572/console), I'm pushing this as a quick hotfix. Cathing all exceptions, however, is generally bad (`Exception` was only marginally better than `Throwable`), so we have to choose a different path for the long term. Among the possible actions (some of these have been discussed with @sbesson):

 * Enforce a clear contract so that INI parsing always fails with a predictable exception (possibly a custom one)
 * In the recursive search for config files, do not blindly throw anything at the INI parser. Perhaps the safest course would be to consider only files named `.bioformats` and enforce this across the board

In any event, the fact that the test fails with a cryptic message is really bad (I've lost one day chasing unicorns). We should find a way to tell TestNG *not* to swallow all error info (set verbosity level to -1?), then remove most try/catch blocks from the test code.

simleo:
openmicroscopy/data_repo_config#137 is required with this PR. It also provides an example of how config files can be simplified with the new file list building logic.

sbesson:
On the filtering, it might be sufficient to condition the file parsing to the output of  `TestTools.isConfigFile()` as we do when initializing the reader for instance.

More generally, with regard to the options listed in https://github.com/openmicroscopy/bioformats/pull/2574#issuecomment-249140225, the historical approach has been to test a set of files within a folder using a configuration file called `.bioformats` stored in the same folder containing the reference metadata. To be able to handle multiple branches, we introduced the possibility of [suffixing these files](https://github.com/openmicroscopy/bioformats/pull/673) e.g. `.bioformats_5.0` so that multiple configuration files could co-exist within the same directory. More recently, some effort was put into [externalizing these configuration files](https://github.com/openmicroscopy/bioformats/pull/1819) and migrating our test infrastructure to use different configuration directories to separate in branches/versions. My feeling is that with the latter we effectively deprecated the usage of configuration file suffixes and might want to deprecate/drop it in our test infrastructure accordingly in 5.3.0 unless there is some remaining usage of this API /cc @melissalinkert @joshmoore

components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
```diff
@@ -199,29 +202,32 @@ public String getProperty(String key) {
     }
 
     // remove duplicates
-    int index = 0;
+    Set<String> fileSet = new HashSet<String>(files);
```
sbesson: The main downside of this operation is that it works under the assumption that the master/current file selection independently of the initially selected file in `setId()`. Unfortunately, this does not hold for many multi-file readers which do not implement such logic (incl. `ome-tiff`, `micromanager`...).

Since the file header is expected to match the output of `getCurrentFile()` for the first scanned file of the fileset, is there a way to maintain the previous listing order?

dgault:
All current builds and existing tests are green as expected with the various corresponding config PR's.

I've also been retesting this PR by generating a number of configs across various formats and comparing with the originals to ensure that the same config properties are being generated. I was unable to find any issues that would raise a flag in this regard and as such this PR looks ready to merge. 

The issues with the upper case suffixes found in the MetaMorphReader is not directly related to this PR and can be handled separately at a later stage.

sbesson:
With the passing [https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/305/](automated tests), merging this and the companion configuration PRs. Thanks for lowering the barrier to adding/maintaining configuration files.

PR #2579, opened 2016-09-15
Ant: use mvn_ant_tasks plugin

sbesson:
See https://trello.com/c/pWCmg33k/24-investigate-maven-ant-tasks for more context.

The scope of this PR is to investigate the usage of the [Maven Ant tasks](http://maven.apache.org/ant-tasks/) in order to manage the dependencies for the Ant build system. While the long term goal might be to drop Ant and decouple the components into smaller repositories, this is targetted for a stable series release and should be non-breaking.

The choice of [Maven Ant tasks](http://maven.apache.org/ant-tasks/) for the dependency management system was proposed as a third-party library which minimal impact on the code which integrates nicely with the Maven repository. Its main caveat is that it is unmaintained. As the mid-term goal is to decouple components and use Maven for the Java build lifecycle, it is expected this decision will have minimal impact. Might we need to reconsider this choice, Apache Ivy would be a viable replacement Ant-compatible, already used by the OMERO build system and able convert `pom.xml` into `ivy.xml` descriptors for dependencies.

We limit the implementation in this PR to the dependencies used for compile, test and runtime of the various components from the source code repository and facilitate the integration of new libraries using exclusively `pom.xml` for declaring these dependencies.

The following changes are made as part of this PR:
- all JAR bundles  have been removed from the `jar` directory
- `ant/java.xml` has been modified to add a `mvn-init` target retrieving the dependencies from the `pom.xml` and creating refid (`compile.classpath`, `test.classpath`, `runtime.classpath`) allowing to use it in the various tasks (without copy)
- the `jar` target in `ant/java.xml` has been modified to also install the compiled jar into the local Maven repository
- the various Ant targets tested by the Travis build (`compile`, `utils`) have been modified to fix the build
- new builds files have been defined in the `components/bundles/{bioformats_package, loci-tools}` for the combined JARs
- test targets have been fixed. For BSD/GPL  Ant-only target testing missing dependencies, the Ant `restrict` semantics is used to exclude dependencies from the test classpath
- all `build.properties` files have been cleaned up to removed hardcoded lists of dependencies
- `copy-jars` is now made as a no-op target.

Impact of development workflow:
- runtime dependencies are no longer copied to `artifacts`. To use the command-line tools, it is necessary to use `ant clean tools`

Proposed for later:
- deprecation of the `osgi`, `deps-*`, `copy-jars` targets
- deprecation of the `ome-tools` bundle


mtbc:
This is looking pretty nice. After `ant jars tools` my resulting `showinf` complains,
```
ClassNotFoundException: org.slf4j.LoggerFactory
```
not sure what further step is needed to fix that.

sbesson:
On the command-line tools, this is related to `copy-jars` question as this target used to copy all runtime dependencies to `artifacts` thus making `./tools/showinf` functional out-of-the-box. Note this was already not working with the Maven build system as the Bio-Formats JARs are only installed in the local Maven repository. 
If we want to maintain the statu quo, one possibility would be to modify copy-jars to copy the runtime dependencies under `artifacts`. Alternatively, this step could simply be part of the logic of the `dist-bftools` target.

mtbc:
Weird, now when I try the same, the `tools` target gives me with `bio-formats-tools`,
```
/home/mtbc/src/bioformats/ant/toplevel.xml:789: src '/home/mtbc/src/bioformats/artifacts/guava-17.0.jar' doesn't exist.
```
Not sure what I did differently.

sbesson:
The `tools` command will currently not work for similar reasons as above i.e. it was relying on  `copy-jars` to create the bundle. I'll look into reusing `components/bundles/*/pom.xml`.

rleigh-codelibre:
Just done some initial testing, and to echo @mtbc's comments it's looking pretty nice.  With a fixed-up/replacement for copy-jars/dist-bftools, plus getting the tests working, it will be pretty nice.  It's definitely looking like the way to go.

melissalinkert:
Also looks good to me, barring already noted issues with building the bundles.  I do think it would be good to document somewhere (perhaps in a follow-up PR) that OSGi will no longer be supported with Ant, even though I doubt this is used by very many people.

sbesson:
Commits pushed to fix the bundles creation by setting up an Ant `build.xml` under the corresponding directories. Additionally it should restore the command-line tools workflow when using the `tools` target:

```
$ ant clean jars tools
$ ./tools/showinf -version
Version: 5.2.2-DEV
VCS revision: feea2bd733018fffd1bf25dd99126577b887efdf
Build date: 19 September 2016
```

The generated bundles seem to contain most of the expected content although some might be missing (e.g. `xalan`). It might be useful to compute the jar content diff between the Maven generated and the Ant generated package.

Regarding `osgi`, 693f31c and 1bb0310 are certainly premature and outside the scope of this PR. I will likely force push them away from this PR. Instead, I propose to mark the following targets/components as deprecate in 5.2.x and drop them in 5.3.0:

- the `osgi` targets
- the `ome-tools` bundle which has not been maintained or built
- the `copy-jars` target (which should become a no-op by the end of this PR)
- potentially all the `deps-*` target which are fully replaced by the dependency management introduced here

ctrueden:
> While the long term goal might be to drop Ant

I am curious why this work is needed... i.e., in what ways the existing Maven build system falls short of requirements?

sbesson:
@ctrueden: thanks for bringing up the question as this was also raised and discussed internally. The reason for going with this approach rather than switching entirely to Maven at this point is mostly related to the amount of work involved. Many fundamental aspects of the Bio-Formats stack still heavily rely on Ant and were never converted to Maven including in no particular order the automated repository  tests, the documentation, the creation of various bundles, all the non-Artifactory release process, half of the continuous integration jobs and several other targets.

The idea of this proposal is that it is non-breaking and allows us to keep up with our existing releases while switching to Maven exclusively as the dependency management system in a first step. The next proposed step will be to start decoupling the low-level components into their own repository similarly to http://imagej.net/Architecture and let these low-level components be managed by the Maven build system.

ctrueden:
@sbesson Great, thanks for the info. Makes sense.

rleigh-codelibre:
I've not found any problems with the latest changes, testing on Linux or Windows.  It may be the case that I haven't touched some more esoteric functionality, e.g. integration tests, but all the basics appear to be completely functional, and this includes all the top-level targets.  It's looking like it's close to being good to include in a breaking build.

sbesson:
Thanks for the review both under Linux/Windows. Note that this PR has been included in the last couple of runs of the breaking build and already streamlines the [archived artifacts](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-build/lastSuccessfulBuild/artifact/artifacts/). One target remaining to fix is the `target-test-runner` under `test-suite`.

rleigh-codelibre:
Hmm, maybe ignore the findbugs above; I probably don't have it installed.

sbesson:
Closing. To be reopened against `dev_5_3`

PR #2582, opened 2016-09-19
ZeissCZIReader : Fix imageCount for RGB images

bramalingam:
**Reference:** 
https://trac.openmicroscopy.org/ome/ticket/13239
http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-May/006030.html

**Testing Instructions:** 
Open the following Image from QA17192 in ImageJ/Fiji,
Conv 2 3 colors.czi

and 
1) check if the image opens. (without this PR, you should get an exception: as shown in the ticket)
and
2) check if there are 9 channels.
and,
1) Split the image into three stacks (3 channels each),
2) And view the images in composite mode: [Image-->Color-->make Composite]

this will allow you to see the 3 channels as 3RGB images individually and these can be compared with Zen Lite.

@melissalinkert

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -715,7 +715,7 @@ public class ZeissCZIReader extends FormatReader {
       seriesCount *= mosaics;
     }
 
-    ms0.imageCount = getSizeZ() * (isRGB() ? 1 : getSizeC()) * getSizeT();
+    ms0.imageCount = getSizeZ() * (isRGB() ? getSizeC()/3 : getSizeC()) * getSizeT();
 
     LOGGER.trace("Size Z = {}", getSizeZ());
     LOGGER.trace("Size C = {}", getSizeC());


sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/301/ looks good with this PR included, thanks @bramalingam. Can you also open a configuration PR for the corresponding QA file while this changeset is reviewed functionally?

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -715,7 +715,7 @@ protected void initFile(String id) throws FormatException, IOException {
       seriesCount *= mosaics;
     }
 
-    ms0.imageCount = getSizeZ() * (isRGB() ? 1 : getSizeC()) * getSizeT();
+    ms0.imageCount = getSizeZ() * (isRGB() ? getSizeC()/3 : getSizeC()) * getSizeT();
```
mtbc: This isn't simply `getEffectiveSizeC()` is it? (I really don't know, I've wondered how all this works!)

bramalingam: /**
   * Gets the effective size of the C dimension, guaranteeing that
   * getEffectiveSizeC() * getSizeZ() * getSizeT() == getImageCount()
   * regardless of the result of isRGB().
   */
  int getEffectiveSizeC();

Thus, getEffectiveSizeC() is basically the whole of (isRGB() ? getSizeC()/3 : getSizeC()) and I guess it is calculated later, once the imageCount is set at this level. This is how I think it works, but I could be wrong. @melissalinkert @sbesson does that sound right?

dgault: Yeah getEffectiveSizeC works based on the Z, T and image count. So although it would usually return the value we are using here, it can only be used once the image count is set so the code here is correct.

dgault:
This fix appears to be working as described. Merge jobs and tests are all green. I also copied over the QA file to inbox and ran the new config tests which also pass - https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/303/console

PR looks good to merge

sbesson:
Thanks both. Will merge once the full repository job is green. Is it possible to produce a screenshot for the bug fix?

sbesson:
Merging as per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/293/

Issue #2576, opened 2016-09-15
Include version number in btools zip


burgerga:
Hi I saw that 5.2.2 was out (even saw my name in release notes ;) )
But at the moment it's a bit tricky to find out which version you're running.

When I run any tool from bioformats I get

```
*** A new stable version is available. ***
*** Install the new version using:     ***
***   'upgradechecker -install'        ***
```

and I don't think the `upgradechecker` is included in the zip.

Also the zip itself and the download website don't tell you which version you will download. Would it be possible to include that information? Maybe via a CHANGELOG, README, or something...

EDIT I see that on http://www.openmicroscopy.org/site/support/bio-formats5.2/users/comlinetools/ there is 5.2.2 in the top left corner, but still it could be clearer I think.

sbesson:
Hi @burgerga. As an immediate feature, most of the command-line tools have a `-version` option which will print the version, the sha1 and the build date and exit

```
$ showinf -version
Version: 5.2.2
VCS revision: 00d095359df2ae798a1ccc6c394347a2f0aebd94
Build date: 13 September 2016
```

Agreed on the need for information, the most appropriate file to consider including into the distributed command-line tools might be https://github.com/openmicroscopy/bioformats/blob/develop/docs/sphinx/about/whats-new.txt. Thoughts?

burgerga:
Can't believe I overlooked the obvious cli argument, must have been the lack of coffee :P
Yes, the whats-new.txt would be nice, but maybe renamed to the conventional NEWS or CHANGELOG.

sbesson:
@burgerga: renaming the file for distribution looks sensible. Feel free to open a Pull Request implementing your proposal directly (NB: the Ant target creating the command-line tools bundle is at https://github.com/openmicroscopy/bioformats/blob/v5.2.2/ant/toplevel.xml#L995) or let us know if we should capture this on our side as a RFE.

burgerga:
I can do it, but maybe it's best that you decide what conventions you want to follow. I like the conventions from http://keepachangelog.com/ (CHANGELOG) but maybe you like GNU (NEWS) conventions better?

simleo:
@burgerga @sbesson note that this is not simply a rename. Rather, we need to add a new plain text redering for `whats-new.txt`. See https://trello.com/c/mRvqiTsP/46-add-news-file

rleigh-codelibre:
[...]

PR #2584, opened 2016-09-21
Add a NEWS file to the bftools zip

burgerga:
Copy whats-new file to bftools zip as is, but renamed to NEWS.rst

See #2576


rleigh-codelibre:
This looks fine to me and the name is good, and it's all packaged properly.  Looks good to merge; thanks for doing this!

hflynn:
👍

rleigh-codelibre:
The only followup comment would be to ask whether we should repeat this for the other zip archives (matlab, octave), and also for the source releases.

sbesson:
Thanks @burgerga for another contribution, merging. I would agree with https://github.com/openmicroscopy/bioformats/pull/2584#issuecomment-248866028 especially as the Octave package already includes it.

PR #2585, opened 2016-09-22
DataTools: remove internal usages of deprecated API

sbesson:
See https://trello.com/c/b4oDFIo4/21-remove-instances-of-deprecated-sanitizedouble

This PR completes the deprecation of the `DataTools.sanitizeDouble()` in https://github.com/openmicroscopy/bioformats/pull/2285 and updates the few places across the code still relying on the deprecated API to use `DataTools.parse*` instead.

This PR should have no functional effect mostly cleanup. To test it:
- check the build does not warn about these deprecated classes anymore
- check the automated tests are still passing with this PR included
- optionally open a couple of files from the two affected formats and check the metadata is unchanged without and with this PR e.g. by comparing the ouput of `showinf -omexml-only` 

Open questions:
- how to handle ND2 metadata where the value is -1?

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Handler.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Handler.java
@@ -511,17 +511,17 @@ public class ND2Handler extends BaseHandler {
         }
       }
       else if ("dPosX".equals(prevElement) && qName.startsWith("item_")) {
-        final Double number = Double.valueOf(DataTools.sanitizeDouble(value));
+        final Double number = DataTools.parseDouble(value);
         posX.add(new Length(number, UNITS.REFERENCEFRAME));
         metadata.put("X position for position #" + posX.size(), value);
       }
       else if ("dPosY".equals(prevElement) && qName.startsWith("item_")) {
-        final Double number = Double.valueOf(DataTools.sanitizeDouble(value));
+        final Double number = DataTools.parseDouble(value);
         posY.add(new Length(number, UNITS.REFERENCEFRAME));
         metadata.put("Y position for position #" + posY.size(), value);
       }
       else if ("dPosZ".equals(prevElement) && qName.startsWith("item_")) {
-        final Double number = Double.valueOf(DataTools.sanitizeDouble(value));
+        final Double number = DataTools.parseDouble(value);
         posZ.add(new Length(number, UNITS.REFERENCEFRAME));
         metadata.put("Z position for position #" + posZ.size(), value);
       }
@@ -563,7 +563,7 @@ public class ND2Handler extends BaseHandler {
         rois.add(roi);
       }
       else if (qName.equals("dPinholeRadius")) {
-        pinholeSize = new Double(DataTools.sanitizeDouble(value));
+        pinholeSize = DataTools.parseDouble(value);
         metadata.put("Pinhole size", value);
       }
       else if (qName.endsWith("ChannelColor")) {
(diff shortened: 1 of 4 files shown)


sbesson:
The current state of this PR uncovered a bug in `DataTools.sanitizeDouble()` which was converting `-1` as `1` and thus storing incorrect metadata values into the MetadataStore. A configuration PR has been opened to fix a few Nikon ND2 samples where the `PhysicalSize{X,Y}` values should be `null` as opposed to `1.0` prior to this PR.

This leaves the question of how to handle metadata fields where the positiveness is not enforced by the data model e.g. `LensNA`. With this PR, the produced OME-XML can now be: 

```
<Objective CalibratedMagnification="-1.0" Correction="Other" ID="Objective:0:0" Immersion="Other" LensNA="-1.0"/>
```

It is unclear to be whether we want to special case -1.0 when parsing some fields in the `ND2Handler` /cc @melissalinkert

mtbc:
One may want to create a card listing properties whose numerical constraints we may want to tighten in the next major schema release. While I am no microscopist, I too am doubtful about negative apertures.

components/formats-gpl/src/loci/formats/in/ND2Handler.java
```diff
@@ -690,25 +688,20 @@ else if (key.equals("sObjective") || key.equals("wsObjectiveName") ||
...
-          }
+          mag = DataTools.parseDouble(m);
         }
         if (magIndex + 1 < tokens.length) immersion = tokens[magIndex + 1];
       }
       else if (key.endsWith("dTimeMSec")) {
-        long v = (long) Double.parseDouble(DataTools.sanitizeDouble(value));
-        if (!ts.contains(new Long(v))) {
-          ts.add(new Long(v));
+        Long v = DataTools.parseLong(value);;
```
mtbc: don't need two semicolons

components/formats-gpl/src/loci/formats/in/ND2Handler.java
```diff
@@ -690,25 +688,20 @@ else if (key.equals("sObjective") || key.equals("wsObjectiveName") ||
...
+        if (!ts.contains(v)) {
```
mtbc: I wonder if `ts` should have been `LinkedHashSet` or somesuch, outwith this PR though.

components/formats-gpl/src/loci/formats/in/ND2Handler.java
```diff
@@ -950,4 +933,16 @@ public boolean isDimensions(String key) {
...
 
+  /**
+   * Parses the physical size from a key.
+   * The expected format for the key is "- Step <value>".
+   */
+  private Double parsePixelsSizeZFromKey(String key) {
+    int space = key.indexOf(" ", key.indexOf("Step") + 1);
+    if (space == -1) return null;
+
+    int last = key.indexOf(" ", space + 1);
```
mtbc: `' '` :smiley:

dgault:
- Code changes here all look fine to me and the new test ND2Handler test covers the necessary scenarios.
- Builds and automated tests are green and I can find no deprecated warnings for the files affected
- Manual checks on ND2 and Improvision files show metadata remains unchanged as expected.

With regards the handling of negative values, I would agree with @mtbc that my preferred approach would be to add a positive restriction at the model level as part of a major release. That would at least make it very clear what values we would handle and what is considered invalid. For the time being though returning a null is certainly a better option than an incorrect positive value

sbesson:
Card created to record the proposed model enforcements. At the moment, we are not making any assumption and storing exactly what is read from the metadata table as long as it satisfies the current model definitions, including `-1` values. As this definitely fixes incorrect values, I propose to get this merged and review this in 5.3.0 maybe via a metadata option allowing to choose whether `-1` should be read as `null` /cc @melissalinkert

melissalinkert:
Reads fine to me, no veto to merging, and definitely +1 to changing the definition of ```LensNA``` at the model level before enforcing constraints in readers.  I'd be a little hesitant to use the metadata option API for this case if the model is going to be changed anyway, but that probably depends a bit upon the timeline for the next model release.

PR #2586, opened 2016-09-22
Add versioning policy to docs

hflynn:
See https://trello.com/c/nuexeTkT/16-versioning-documentation-page

Staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/about/index.html

### DIFF
--- a/docs/sphinx/about/index.txt
+++ b/docs/sphinx/about/index.txt
@@ -59,9 +59,42 @@ than being updated whenever a new version of :omerodoc:`OMERO <>` is released.
 We expect this to result in more frequent releases to get fixes out to the
 community faster.
 
-The version number is three numbers separated by dots e.g. 4.0.0. See the
-:doc:`version history <whats-new>` for a list of major changes in each
-release.
+See the :doc:`version history <whats-new>` for a list of major changes in
+each release.
+
+.. _versioning-policy:
+
+Versioning policy
+^^^^^^^^^^^^^^^^^
+
+Bio-Formats does not currently conform to `strict semantic versioning <http://semver.org>`_.
+The following set of rules describe the current policy (using
+`RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`_):
+
+- The version number MUST take the form X.Y.Z where X, Y, and Z are
+  non-negative integers, and MUST NOT contain leading zeroes.
+- The patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards
+  compatible bug fixes are introduced. A bug fix is defined as an internal
+  change that fixes incorrect behavior.
+- The minor version Y MUST be increased when backwards incompatible changes
+  (model breaking API) are introduced to the public API. It MAY include patch
+  level changes.
+- The minor version MUST be increased if the version of a non-OME/external
+  dependency is bumped.
+
+The exception to this policy is serialization. Serialization functionality was
+implemented as a ReaderWrapper called Memoizer in Bio-Formats 5.0.0 and is
+exposed to the community via a public API. Currently:
+
+- Major/minor/patch version bumps of Bio-Formats are not backwards-compatible
+  with regard to serialization.
+- Consumers with code relying on Bio-Formats caching stability should
+  not upgrade their Bio-Formats version for now.
+- Changes breaking the serialization should be grouped together as much as
+  possible in order to minimize the number of breakages per series.
+
+See `this GitHub issue <https://github.com/openmicroscopy/design/issues/55>`_
+for further details.
 
 Why Java?
 ---------
--- a/docs/sphinx/about/index.txt
+++ b/docs/sphinx/about/index.txt
@@ -80,7 +80,7 @@ The following set of rules describe the current policy (using
 - Either the minor version Y or the major version X MUST be incremented when
   backwards-incompatible changes (model-breaking API) are introduced to the
   public API. These version increases MAY also include patch level changes.
-- Either the minor version or the major version MUST be increased if the
+- Either the minor version or the major version MUST be incremented if the
   version of a non-OME/external dependency is updated.
 
 The exception to this policy is serialization. Serialization functionality was
--- a/docs/sphinx/about/index.txt
+++ b/docs/sphinx/about/index.txt
@@ -77,8 +77,8 @@ The following set of rules describe the current policy (using
 - The patch version Z MUST be incremented if only backwards
   compatible bug fixes are introduced. A bug fix is defined as an internal
   change that fixes incorrect behavior.
-- Either the minor version Y or the major version X MUST be increased when
-  backwards incompatible changes (model breaking API) are introduced to the
+- Either the minor version Y or the major version X MUST be incremented when
+  backwards-incompatible changes (model-breaking API) are introduced to the
   public API. These version increases MAY also include patch level changes.
 - Either the minor version or the major version MUST be increased if the
   version of a non-OME/external dependency is updated.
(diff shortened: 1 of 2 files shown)


hflynn:
Only additional point is I wonder whether it would be helpful to also reference this policy in the developer docs on the pages referring to memoizer?

docs/sphinx/developers/java-library.txt
```diff
@@ -172,6 +172,10 @@ The complete list of current dependencies is as follows:
       - xml-apis:xml-apis:1.3.04
       - `Apache License v2.0`_
 
+As described in :ref:`versioning-policy`, the minor version number of a
+Bio-Formats release will always be increased if the version of a
+non-OME/external dependency is bumped.
```
mtbc: would something like "changed" read better than "bumped"? (it also may not be a simple increment)

docs/sphinx/about/index.txt
```diff
@@ -59,9 +59,42 @@ than being updated whenever a new version of :omerodoc:`OMERO <>` is released.
...
+Versioning policy
+^^^^^^^^^^^^^^^^^
+
+Bio-Formats does not currently conform to `strict semantic versioning <http://semver.org>`_.
+The following set of rules describe the current policy (using
+`RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`_):
+
+- The version number MUST take the form X.Y.Z where X, Y, and Z are
+  non-negative integers, and MUST NOT contain leading zeroes.
+- The patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards
```
mtbc: a patch version can = 0?

docs/sphinx/about/index.txt
```diff
@@ -59,9 +59,42 @@ than being updated whenever a new version of :omerodoc:`OMERO <>` is released.
...
+  compatible bug fixes are introduced. A bug fix is defined as an internal
+  change that fixes incorrect behavior.
+- The minor version Y MUST be increased when backwards incompatible changes
+  (model breaking API) are introduced to the public API. It MAY include patch
```
mtbc: A confusion here is that presumably, say, 5.2.5 to 5.3.0 would also suffice even though Y got smaller?

docs/sphinx/about/index.txt
```diff
@@ -59,9 +59,42 @@ than being updated whenever a new version of :omerodoc:`OMERO <>` is released.
...
+  level changes.
+- The minor version MUST be increased if the version of a non-OME/external
+  dependency is bumped.
+
+The exception to this policy is serialization. Serialization functionality was
+implemented as a ReaderWrapper called Memoizer in Bio-Formats 5.0.0 and is
+exposed to the community via a public API. Currently:
+
+- Major/minor/patch version bumps of Bio-Formats are not backwards-compatible
+  with regard to serialization.
```
mtbc: Presumably the Memoizer API *is* generally backwards-compatible across patch versions, the issue is specifically limited to re-reading the cache memo files generated by previous versions? So we still won't break your code so long as you can cope with having to remake the files.

docs/sphinx/about/index.txt
hflynn: I copied this from Seb's design text and can remove it if it's confusing. An x.y.0 release isn't a patch release though, is it.

docs/sphinx/about/index.txt
hflynn: Should I add 'unless the major version is increased instead'?

docs/sphinx/about/index.txt
mtbc: The text seems to suggest that the patch version is simply the third number of the version number tuple. Besides, we would increment it from 0 to 1 for a first patch.

docs/sphinx/about/index.txt
mtbc: Unfortunately we don't define the major version anywhere. (These issues are partly coming up because the text is very much of a "read me pedantically, this is a formal definition" style.) Could refer to, say, "major version X (from X.y.z)" or somesuch I guess?

sbesson: Agreed on these. Obviously, our mid-term goals is to *be* semver so that we can refer to semver.org instead of partially duplicating it. Worth adding `X is the major version, Y is the minor version, and Z is the patch version` to the first sentence?

mtbc: That'd certainly help.

docs/sphinx/about/index.txt
```diff
@@ -59,9 +59,42 @@ than being updated whenever a new version of :omerodoc:`OMERO <>` is released.
...
-release.
+See the :doc:`version history <whats-new>` for a list of major changes in
+each release.
+
+.. _versioning-policy:
+
+Versioning policy
+^^^^^^^^^^^^^^^^^
+
+Bio-Formats does not currently conform to `strict semantic versioning <http://semver.org>`_.
```
mtbc: maybe "yet" instead of "currently" to indicate that we are hoping to someday!

sbesson: If we want to be adventurous I'll make a proposal in that sense, we could even say `Up to Bio-Formats 5.2.0`.

docs/sphinx/developers/java-library.txt
sbesson: Updated ? cf http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

mtbc: Sure, that works well.

hflynn:
Fixes pushed, hopefully clearer now

docs/sphinx/about/index.txt
```diff
@@ -67,27 +67,30 @@ each release.
...
+  version, Y is the minor version and Z is the patch version.
+- The patch version Z MUST be incremented if only backwards
   compatible bug fixes are introduced. A bug fix is defined as an internal
   change that fixes incorrect behavior.
-- The minor version Y MUST be increased when backwards incompatible changes
-  (model breaking API) are introduced to the public API. It MAY include patch
-  level changes.
-- The minor version MUST be increased if the version of a non-OME/external
-  dependency is bumped.
+- Either the minor version Y or the major version X MUST be increased when
```
mtbc: incremented

docs/sphinx/about/index.txt
```diff
@@ -67,27 +67,30 @@ each release.
...
 
 The exception to this policy is serialization. Serialization functionality was
 implemented as a ReaderWrapper called Memoizer in Bio-Formats 5.0.0 and is
 exposed to the community via a public API. Currently:
 
 - Major/minor/patch version bumps of Bio-Formats are not backwards-compatible
-  with regard to serialization.
+  with regard to serialization in that cached memo files written with a
+  previous version may not be readable by later versions and may need to be
+  remade.
```
mtbc: "rewritten", perhaps, but :+1: in general :smiley:

docs/sphinx/about/index.txt
```diff
@@ -59,9 +59,45 @@ than being updated whenever a new version of :omerodoc:`OMERO <>` is released.
...
+- The version number MUST take the form X.Y.Z where X, Y, and Z are
+  non-negative integers, and MUST NOT contain leading zeroes. X is the major
+  version, Y is the minor version and Z is the patch version.
+- The patch version Z MUST be incremented if only backwards
+  compatible bug fixes are introduced. A bug fix is defined as an internal
+  change that fixes incorrect behavior.
+- Either the minor version Y or the major version X MUST be incremented when
+  backwards-incompatible changes (model-breaking API) are introduced to the
+  public API. These version increases MAY also include patch level changes.
+- Either the minor version or the major version MUST be increased if the
```
mtbc: incremented

mtbc:
It's looking good to me!

sbesson:
Also reads well to me. Any final comments/thoughts @melissalinkert @dgault @simleo ?

dgault:
The versioning info here appears clear and well documented. I cant think anything else that would need added, looks good from my perspective.

melissalinkert:
Looks good to me as well.

PR #2587, opened 2016-09-23
Add max supported versions for MicroManager& ScanR

hflynn:
See https://trello.com/c/9qBOB08S/26-additional-format-version-ranges

Staged at https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/formats/micro-manager.html and https://www.openmicroscopy.org/site/support/bio-formats5.2-staging/formats/olympus-scanr.html


dgault:
Both versions are correct according to the info on the trello card and the staged pages are displaying fine. Looks good to merge.

PR #2588, opened 2016-09-26
ScanR: return blank planes instead of skipping missing wells (develop)

simleo:
Rebases #1995 onto develop.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ScanrReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ScanrReader.java
@@ -519,7 +519,7 @@ public class ScanrReader extends FormatReader {
       if (next == originalIndex &&
         missingWellFiles == nSlices * nTimepoints * nChannels * nPos)
       {
-        wellNumbers.remove(well);
+        next += nSlices * nTimepoints * nChannels * nPos;
       }
     }
     nWells = wellNumbers.size();


simleo:
The error (https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/311/console) is due to the fact that with this patch the reader finds 9 additional series in `curated/scanr/ruben/plate_002`. Since these are not configured, this leads to currentTable being set to `null` in `Configuration.java`

simleo:
**TL; DR: I think the current version of the reader does the "right thing" on the ruben sample and we should fix the secretion bug in some other way**

`ruben/plate_002` is a plate with 4 rows, 2 columns and 9 fields (subpos). TIFF files for the last well are missing. The current approach (i.e., without this PR) is to completely skip the well: the series count is set to 9 * 7 and the last well is missing in OMERO (white, non-clickable in the GUI). With this PR, the series count is set to 9 * 8 and blank (all-zero) planes are returned where TIFF files are missing (the last well is present and fully black in OMERO).

This was done to sidestep import failures in `idr0009-simpson-secretion`, but I don't think we should keep it as a permanent solution. A blank plane is a valid image plane and can be hard to tell from other planes in a dataset, not just visually. In automated analysis, blank data could be processed for no reason, increasing computation time and leading to garbage results. Downstream from the reader, there's no way to tell which planes were originally missing.

This falls into the sparse data handling domain: see https://github.com/openmicroscopy/bioformats/issues/2467 and the discussion on https://github.com/openmicroscopy/bioformats/pull/2536 for a similar problem. @sbesson proposed putting the user in control of how missing planes are handled via metadata options. While adding flexibility is certainly good, the default behaviour should be to skip planes, for the reasons discussed above.

Is there a way to fix the secretion bug that does not involve returning fake planes? Other multi-file readers skip planes without causing errors. For instance, see plates from `idr0008-rohn-actinome/screenB` (`CellWorxReader`).

/cc @joshmoore @melissalinkert

joshmoore:
Only heads up from my side is that it would be unfortunate if we had to reimport.

sbesson:
Closing for now to focus on Bio-Formats 5.3.0.

PR #2589, opened 2016-09-27
ScanR: read core metadata from first non-null TIFF file (develop)

simleo:
Rebases #2147 onto develop.

**NOTE:** the bug this PR fixes is triggered by #2588 (the `tiffs` array changes order so that a `null` entry appears at index 0), so we should merge that one first.


sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/311/console. Marking as exclude.

simleo:
Looks like the test error is due to openmicroscopy/bioformats#2588

sbesson:
Same comment as https://github.com/openmicroscopy/bioformats/pull/2588#issuecomment-261878317

PR #2590, opened 2016-09-29
Deltavision: use readUnsignedShort() to read time dimensions

sbesson:
See  http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-September/006183.html and https://trac.openmicroscopy.org/ome/ticket/13261

For large dimensions (over 32767), using readShort() wraps the value to a negative. This commit explores using readUnsignedShort() instead to retrieve the  time dimension while remaining in agreement with the DV specification.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -307,7 +307,7 @@ public class DeltavisionReader extends FormatReader {
     int filePixelType = in.readInt();
 
     in.seek(180);
-    int rawSizeT = in.readShort();
+    int rawSizeT = in.readUnsignedShort();
     int sizeT = rawSizeT == 0 ? 1 : rawSizeT;
 
     int sequence = in.readShort();


mtbc:
This code change looks good to me.

dgault:
Change looks fine for me. There is nothing in the Deltavision spec to say if negative values are allowed or not, but they wouldn't appear to make any sense in this context. Looks good to merge.

sbesson:
Was this tested in e.g. ImageJ against one of the provided sample files? Is there a screenshot?

dgault:
Yes it was tested with the large 40000 time point Deltavision file, opened via the Bio-Formats plugin in Fiji without any exception.

I've attached screenshots of the before and after when opening in Fiji

Without the PR:
screen shot 2016-09-30 at 15 30 40

With the PR:
screen shot 2016-09-30 at 15 28 50

sbesson:
While trying to create a configuration file for the relevant Deltavision file, I encountered an issue with the configuration generation as the decimal values of the timestamps are truncated in the configuration file resulting in failures in `PlanePositions` tests. As there are 40K values, this is something that should be investigated/fixed separately. Merging as per https://github.com/openmicroscopy/bioformats/pull/2590#issuecomment-250759973 to fix support in 5.2.3

PR #2591, opened 2016-09-30
Deprecate the unused osgi Ant targets

sbesson:
The osgi targets are not built or released. The ome-tools bundle is release with other components but is unmaintained and unused. 
This commit deprecates these build targets in anticipation of their removal in upcoming releases.


mtbc:
I don't think I'd even noticed the OME tools bundle before. Is it unrelated to `bftools.zip`?

sbesson:
Completely unrelated, it might have been introduced as a very low-level component (dependencies are https://github.com/openmicroscopy/bioformats/blob/v5.2.2/ant/toplevel.properties#L137)

mtbc:
Noted in Trello to mention this in whatsnew?

sbesson:
There is a card for the 5.2.3 version history including this PR in the checklist /cc @hflynn

mtbc:
:+1: Good to merge.

PR #2592, opened 2016-09-30
BaseTiffReader - Index out of bounds exception 

dgault:
Issue was discovered while testing QA-17365 as part of https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8124&p=17519#p17519

An object of type String [] is read for the creation date of the file, but the array itself is empty leading to an index out of bounds exception.

To test:
- All unit tests and data repo tests must remain green
- Read image from QA-17365 using Bio-Formats and verify that an array out of bounds exception is thrown
- Apply the PR
- Reopen the file and verify that no exception is thrown

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/tiff/BaseTiffReaderMock.java
+++ b/components/formats-bsd/test/loci/formats/utests/tiff/BaseTiffReaderMock.java
@@ -1,3 +1,35 @@
+/*
+ * #%L
+ * BSD implementations of Bio-Formats readers and writers
+ * %%
+ * Copyright (C) 2016 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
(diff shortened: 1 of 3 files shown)


sbesson:
Looks good and works as expected with the uploaded QA file. Is there some room for a unit test or would you rather open a configuration PR for the QA file?

dgault:
We are actually missing the testing of invalid values from the unit tests so I will get those written for the base tiff tests as part of this PR.

dgault:
The BaseTiffReader is not specifically covered by existing Unit Tests. I was originally going to add a full unit test for this abstract reader but it was becoming much larger than the scope of this PR. As such I have broken out only the necessary tests specific for this PR change (anything related to the creation date).

components/formats-bsd/test/loci/formats/utests/tiff/BaseTiffReaderMock.java
```diff
@@ -0,0 +1,29 @@
+package loci.formats.utests.tiff;
```
sbesson: Add standard BSD header ?

components/formats-bsd/test/loci/formats/utests/tiff/BaseTiffReaderTest.java
```diff
@@ -0,0 +1,130 @@
+/*
+ * #%L
+ * BSD implementations of Bio-Formats readers and writers
+ * %%
+ * Copyright (C) 2005 - 2016 Open Microscopy Environment:
```
sbesson: `s/2005 - 2016/2016` ?

PR #2596, opened 2016-10-04
Use canonical paths in the test suite

simleo:
Although #2574 has simplified config file generation, the logic that removes files already being tested as part of a different set is not robust enough. openmicroscopy/data_repo_config#146 (see comments on `metamorph/chris`) is an example of how it can fail due to different path strings referring to the same file. This PR changes the above code to use canonical paths to resolve ambiguities. With this, I was able to regenerate the `metamorph/chris` config in a few seconds (see openmicroscopy/data_repo_config#147).

--breaking

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -209,9 +209,11 @@ public class FormatReaderTestFactory {
       String file = fileSet.iterator().next();
       try {
         reader.setId(file);
-        Set<String> auxFiles = new LinkedHashSet<String>(
-            Arrays.asList(reader.getUsedFiles())
-        );
+        String[] usedFiles = reader.getUsedFiles();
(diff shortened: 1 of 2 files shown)


sbesson:
Reads fine to me, I assume we should check 1- if this causes a performance overhead 2- how it performs across operating systems. I see this PR is labelled as breaking. Is the idea to have the breaking full-repository test it or should we include into the daily build (and run the Windows repository jobs as well)?

simleo:
I've labeled it as breaking because of the potential impact on the daily build. Subsequently, though, I've tested it together with openmicroscopy/data_repo_config#147 on breaking-repository-subset, so now we can be a bit more confident. I guess we can remove the breaking label from both PRs and see how it goes. My understanding is that Java guarantees `File.getCanonicalPath` to return a unique absolute path on all platforms.

simleo:
Breaks `test_images_good` because it resolves symlinks. Closing in favour of #2601

PR #2599, opened 2016-10-05
Ant: uses mvn_ant_tasks to manage dependencies

sbesson:
Replaces https://github.com/openmicroscopy/bioformats/pull/2579. See https://trello.com/c/pWCmg33k/24-investigate-maven-ant-tasks for more context.

The scope of this PR is to investigate the usage of the [Maven Ant tasks](http://maven.apache.org/ant-tasks/) in order to manage the dependencies for the Ant build system. While the long term goal might be to drop Ant and decouple the components into smaller repositories, this is targetted for a stable series release and should be non-breaking.

The choice of [Maven Ant tasks](http://maven.apache.org/ant-tasks/) for the dependency management system was proposed as a third-party library which minimal impact on the code which integrates nicely with the Maven repository. Its main caveat is that it is unmaintained. As the mid-term goal is to decouple components and use Maven for the Java build lifecycle, it is expected this decision will have minimal impact. Might we need to reconsider this choice, Apache Ivy would be a viable replacement Ant-compatible, already used by the OMERO build system and able convert `pom.xml` into `ivy.xml` descriptors for dependencies.

We limit the implementation in this PR to the dependencies used for compile, test and runtime of the various components from the source code repository and facilitate the integration of new libraries using exclusively `pom.xml` for declaring these dependencies.

The following changes are made as part of this PR:
- all JAR bundles  have been removed from the `jar` directory
- `ant/java.xml` has been modified to add a `mvn-init` target retrieving the dependencies from the `pom.xml` and creating refid (`compile.classpath`, `test.classpath`, `runtime.classpath`) allowing to use it in the various tasks (without copy)
- the `jar` target in `ant/java.xml` has been modified to also install the compiled jar into the local Maven repository
- the various Ant targets tested by the Travis build (`compile`, `utils`) have been modified to fix the build
- new builds files have been defined in the `components/bundles/{bioformats_package, loci-tools}` for the combined JARs
- test targets have been fixed. For BSD/GPL  Ant-only target testing missing dependencies, the Ant `restrict` semantics is used to exclude dependencies from the test classpath
- all `build.properties` files have been cleaned up to removed hardcoded lists of dependencies
- `copy-jars` is now made as a no-op target.

Impact of development workflow:
- runtime dependencies are no longer copied to `artifacts`. To use the command-line tools, it is necessary to use `ant clean tools`


dgault:
Reading through the changes and everything makes sense and looks good from my perspective. 
- From the latest build, the jars have been removed from the workspace as noted and the build is running successfully.  
- Artifacts folder contains all the correct jars including loci-tools and bioformats_package. 
- The javadoc is also present in a zipped folder as expected.
- The missing dependency tests are running correctly and passing.
- bftools and bfmatlab are present and zipped 

Is there anything else additional that I should be specifically looking for or testing here?

sbesson:
Briefly discussed with @dgault. Given there are no obvious blockers and follow-up tasks have already been captured for `target-test-runner`, merging this PR and moving forward with the rest of the 5.3.0 work.

PR #2600, opened 2016-10-05
Bio-Formats: CZIReader : Add series metadata,populate position list

bramalingam:
ref: http://trac.openmicroscopy.org/ome/ticket/13290

Testing Instructions : 
Please use the code pasted in the ticket in matlab to check if the seriesMetadata is populated.
And also with FIJI/ImageJ to check if the original metadata table is populated appropriately for all the planes/series.

@melissalinkert @sbesson : As always comments on the code, would be really helpful.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -964,29 +964,22 @@ public class ZeissCZIReader extends FormatReader {
       indices.add(i);
       indexIntoPlanes.put(c, indices);
       //Add series metadata : populate position list
+      int nameWidth = String.valueOf(getSeriesCount()).length();
       for (DimensionEntry dimension : p.directoryEntry.dimensionEntries) {
-          if (dimension == null) {
-            continue;
-          }
-          switch (dimension.dimension.charAt(0)) {
-              case 'S':
-                  setSeries(p.coreIndex);
-                  String sIndex = "";
-                  int seriesId = p.coreIndex + 1;
-                  //add padding to make sure the original metadata table is organized properly in ImageJ
-                  if (seriesId < 10) {
-                      sIndex = "000" + String.valueOf(seriesId);
-                  }
-                  else if (seriesId >= 10 && seriesId < 100){
-                      sIndex = "00" + String.valueOf(seriesId);
-                  }
-                  else if (seriesId >= 100 && seriesId < 1000){
-                      sIndex = "0" + String.valueOf(seriesId);
-                  }
-                  addSeriesMetaList("Positions|" + "SeriesIndex|" + sIndex + "|", dimension.start);
-                  break;
-          }
-       }
+        if (dimension == null) {
+          continue;
+        }
+        switch (dimension.dimension.charAt(0)) {
+          case 'S':
+            setCoreIndex(p.coreIndex);
+            int seriesId = p.coreIndex + 1;
+            //add padding to make sure the original metadata table is organized properly in ImageJ
+            String sIndex = String.format("Positions|Series %0" + nameWidth + "d|", seriesId);
+            addSeriesMetaList(sIndex, dimension.start);
+            break;
+        }
+      }
+      setCoreIndex(0);
     }
 
     if (channels.size() > 0 && channels.get(0).color != null && !isRGB()) {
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -963,6 +963,30 @@ else if (segment instanceof Attachment) {
...
+                  //add padding to make sure the original metadata table is organized properly in ImageJ
+                  if (seriesId < 10) {
+                      sIndex = "000" + String.valueOf(seriesId);
+                  }
+                  else if (seriesId >= 10 && seriesId < 100){
+                      sIndex = "00" + String.valueOf(seriesId);
+                  }
+                  else if (seriesId >= 100 && seriesId < 1000){
+                      sIndex = "0" + String.valueOf(seriesId);
+                  }
```
melissalinkert: Probably worth using ```String.format(...)``` instead of the if/else here, so that this doesn't need to be updated if/when we encounter >= 10000 series.  ```NativeND2Reader``` does something similar: https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java#L1785

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -963,6 +963,30 @@ else if (segment instanceof Attachment) {
...
       indices.add(i);
       indexIntoPlanes.put(c, indices);
+      //Add series metadata : populate position list
+      for (DimensionEntry dimension : p.directoryEntry.dimensionEntries) {
+          if (dimension == null) {
+            continue;
+          }
+          switch (dimension.dimension.charAt(0)) {
+              case 'S':
+                  setSeries(p.coreIndex);
```
melissalinkert: This should likely be ```setCoreIndex(p.coreIndex)```, otherwise the indexing may be wrong when working with files that contain pyramids.

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -963,6 +963,30 @@ else if (segment instanceof Attachment) {
...
+                  else if (seriesId >= 10 && seriesId < 100){
+                      sIndex = "00" + String.valueOf(seriesId);
+                  }
+                  else if (seriesId >= 100 && seriesId < 1000){
+                      sIndex = "0" + String.valueOf(seriesId);
+                  }
+                  addSeriesMetaList("Positions|" + "SeriesIndex|" + sIndex + "|", dimension.start);
+                  break;
+          }
+       }
```
melissalinkert: ```setCoreIndex(0)``` should be added here, to restore the reader state to what it was before iterating over ```p.directory.dimensionEntries```.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/317/console for failing tests. Marking as breaking.

bramalingam:
Subset jobs have been triggered with the latest set of changes : https://ci.openmicroscopy.org/view/Breaking/job/BIOFORMATS-DEV-breaking-repository-subset/150/

sbesson:
Tested using the MATLAB workflow described in the ticket. I can confirm the series metadata table is populated while it was empty before:

```
>> keys = seriesMetadata.keys;
for i=1:seriesMetadata.size()
key = keys.nextElement();
value = seriesMetadata.get(key);
fprintf('%s = %g\n', key, value)
end
Positions|Series 006| #2 = 2
Positions|Series 006| #1 = 2
Positions|Series 336| #3 = 2
Positions|Series 336| #2 = 2
Positions|Series 336| #1 = 2
Positions|Series 259| #3 = 0
Positions|Series 259| #2 = 0
Positions|Series 259| #1 = 0
Positions|Series 137| #3 = 1
...
```

Is the semantics of this extended metadata clear enough for an end-user? Are we talking about Position for each Series as defined by the OME Model? What is the meaning of `#xx` for each series?

bramalingam:
In this image, its the channel numbers. But it would be interesting to see how the same one behaves in a multi-dimensional stack. To my knowledge, the current implementation might be found wanting again (any comments on how we can handle that, apart from hard-coding the channel/time-points as well like the Position/Series?)

melissalinkert:
The position index typically won't change between different planes in the same series, so for the future it may make sense to use ```addSeriesMeta``` instead of ```addSeriesMetaList``` for clarity.  That probably needs further investigation though, and the code changes here all look great otherwise, so happy for this to be merged as-is.

PR #2601, opened 2016-10-05
Test suite: use canonical paths to remove duplicates

simleo:
This PR has the same purpose as #2596, but it restores the original string representation of each path after removing redundancies. This prevents problems on tests that rely on symlinks (such as `test_images_good`).

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -245,6 +245,11 @@ public class FormatReaderTestFactory {
     }
     files = new ArrayList<String>();
     for (String s: minimalFiles) {
+      if (!originalPath.containsKey(s)) {
+        String msg = "No match found for " + s;
+        LOGGER.error(msg);
+        throw new RuntimeException(msg);
+      }
       files.add(originalPath.get(s));
     }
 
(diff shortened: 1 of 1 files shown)


simleo:
61cc3130e2fe751a35b91d3ba143c503a49808e4 deals with a problem that may occur when only part of the files in a multi-file image are symlinked to a different directory. For instance, if:

* a dataset consists of files `A` and `B` in a directory `/foo`
* you link `/foo/A` to `/bar/A` and run the test on `/bar`
* the reader (initialized with `A`) identifies both files as used files (some readers resolve symlinks while looking for auxiliary files), but returns `B` from `getCurrentFile`

Then:

* only `foo/B` survives the loop that removes redundancies
* The `originalPath` map contains only `/bar/A => /foo/A`, so `originalPath.get(s)` is `null`

Without 61cc3130e2fe751a35b91d3ba143c503a49808e4, the `null` was passed on to `FormatReaderTest.configTree.get`, leading to an NPE, while now the test breaks as soon as this problem is encountered and logs the name of the missing item.

The solution on the data repo side is to link **all** files in a multi-file set when generating a symlinked subset.

sbesson:
Given https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-full-repository/51/, moving out of breaking.

sbesson:
Looks good.  This has been tested using the [formats_subset](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/339/parameters/) directory or the [individual format options](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/346/parameters/). Once the [daily repository job](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/337) passes, this should be ready to merge.

PR #2602, opened 2016-10-05
Temporarily disable permissions memoizer unit tsts

sbesson:
The File.setWritable(boolean) logic of these tests was already deficient under
Windows. These tests have been failing under the Docker devspace as well. This
commit removes the @Test annotations for separate investigation.


sbesson:
Closing as I found the reason for the failing tests. They were executed with the `root` accound on Jenkins (fixed by https://github.com/openmicroscopy/devspace/pull/53/commits/c53212c906fda40cb04d45189d0b790163927adb). With a standard user everything works as expected. Likely we'll have to live with the fragility of these tests for now until we start reworking the memoization.

PR #2604, opened 2016-10-07
Drop deprecated OME-Tools bundle

dgault:
This is a follow up to #2599

The OME-Tools bundle was deprecated as of Bio-Formats 5.2.3 (see https://github.com/openmicroscopy/bioformats/pull/2591), this PR removes the bundle completely


hflynn:
I've just spotted that there is still mention of OME Tools bundle on https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/components.html#all-components - do you want to delete that on this PR and I'll open a docs PR to remove it for 5.2.x or would it be easier if I just rebase my PR to this branch too? (cc @sbesson )

sbesson:
@hflynn: :+1: we will certainly need to review the developer docs once all these components are removed from the core repository. I would not worry too much about a 5.2.x PR. All builds are passing (once the `tools-ome` target is removed from the CI builds) Merging.

PR #2266, opened 2016-03-01
Mark all ome.jxr classes as deprecated

sbesson:
See https://trello.com/c/urA9fvFj/104-deprecate-jpeg-xr-classes

Add references to the corresponding mailing list thread and blog post

PR #2605, opened 2016-10-07
Drop deprecated OME-JXR

dgault:
This is a follow up to #2599

The OME-JXR bundle has already been deprecated https://github.com/openmicroscopy/bioformats/pull/2266 
This PR removes the module and all build targets for ome-jxr.

Additionally as the ome-jxr component is isolated and not being used as part of any other components, the entire component has also been removed.


sbesson:
All usual CI builds (Travis + Jenkins Ant, Maven) are passing with this PR included. As discussed in [the corresponding card](https://trello.com/c/ugcioB23/18-remove-ome-jxr-component), we are retiring an component which is both unmaintained and completely unused internally. The previous artifacts remain available if anyone was relying on this API. Will ask the team for a last round of vetoes before merging.

PR #2606, opened 2016-10-08
ConversionTest: migration to Formats BSD

sbesson:
See https://trello.com/c/jGrPs6c9/16-fix-integration-tests for the general work on the integration/performance test on this branch

As explained in the [developer documentation](https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/components.html), the `test-suite` component is dedicated to Bio-Formats integration tests requiring concrete data.

While reviewing the Ant targets of this component, two of them do not fulfill this requirement: `test-minimal-convert` and `test-tiff-compress`. In addition, these targets also need to be invoked independently (see the BIOFORMATS-DEV-merge-build) and are not covered by the Travis build.

This PR proposes to address this issue for `ConversionTest` by migrating it under the `formats-bsd` unit tests (as all writers are under this component). The test is also registered in the `testng.xml` so that is can be picked up by the default Maven and Ant `test` phases of the component. The `test-minimal-convert` target is removed and the CI jobs will require a corresponding update.

To test this PR:
- check Travis and the CI jobs for the `dev_5_3` branch are green
- check Maven ant/or Ant `test` include the `ConversionTest` as part of the executed test suite

PR #2607, opened 2016-10-10
CompressDecompressTest: migration to Formats BSD

sbesson:
See https://github.com/openmicroscopy/bioformats/pull/2606 and https://trello.com/c/jGrPs6c9/16-fix-integration-tests for the general work on the integration/performance test on this branch

This PR migrates the Tiff compression/decompression test from `test-suite` to `formats-bsd` and registers it in teh `testng.xml` so that it is picked up as part of the `test` phase of Ant/Maven. Additionally, the logic for computing checksum from a byte array is duplicated into the test itself.

To check this PR:
- check Travis and the CI jobs for the dev_5_3 branch are green
- check Maven ant/or Ant test include the CompressDecompressTest as part of the executed test suite

NB: as a very simple RFE based on this work, the logic for computing checksums out of byte arrays could be refactored down to `DataTools`. This might be a very simple feature allowing us to test the `ome-common-java` integration if we wanted to /cc @rleigh-codelibre


components/formats-bsd/test/loci/formats/utests/CompressDecompressTest.java
```diff
@@ -44,6 +54,19 @@
  */
 public class CompressDecompressTest {
 
+  /** Calculate the MD5 of a byte array. */
+  public static String md5(byte[] b) {
```
mtbc: should this be `private` for now if it might move?

mtbc: Note also that Guava gives us things like `md5().hashBytes(b).toString()` so we may not even really need a helper.

mtbc:
Good to merge (above comments are merely airing thoughts).

components/formats-bsd/test/loci/formats/utests/CompressDecompressTest.java
```diff
@@ -1,34 +1,46 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
  * #L%
  */
 
-package loci.tests.testng;
+package loci.formats.utests;
 
-import static org.testng.AssertJUnit.fail;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
```
mtbc: don't need these

components/formats-bsd/test/loci/formats/utests/CompressDecompressTest.java
```diff
@@ -1,34 +1,46 @@
...
+
+import com.google.common.hash.Hashing;
 
+import static org.testng.AssertJUnit.fail;
 import org.testng.annotations.Test;
 
+import loci.common.DataTools;
```
mtbc: don't need this

mtbc:
Thank you. :+1: **Good to merge** once CI is green.

PR #2608, opened 2016-10-10
Formats-common: component removal

sbesson:
This is one of the first steps of the [model splitting](https://trello.com/c/f1Bfb7rB/403-model-splitting)

This PR removes the entire `formats-common` component from the Bio-Formats source code repository. This component is been filtered out and reworked as the new https://github.com/ome/ome-common-java repository.

The `formats-common` dependency is declared using `pom.xml`, currently pinned to 5.2.3 and should be retrieved from the OME artifactory. Since the model components are depending on the I/O common package, the POMs are also simplified when necessary to use transitive dependencies e.g. in `components/formats-bsd/build.xml`.

To test this PR, check the builds and tests are passing.


rleigh-codelibre:
You might possibly want to pick up the changes from https://github.com/rleigh-codelibre/bioformats/commit/a3a8dd6b9e6bdde3e5162b96cf153cfc8762196a#diff-2cccd7bf48b7a9cc113ff564acd802a8 (just the one file), though it can wait for later.

rleigh-codelibre:
I created https://github.com/rleigh-codelibre/bioformats/tree/common-split to test this in bioformats.  It fails a test, so I think that either the integration work in bioformats is wrong, or there's an issue with the ome-common jar.  Any help appreciated.

sbesson:
@rleigh-codelibre: looking at https://github.com/sbesson/bioformats/compare/formats-common_removal...rleigh-codelibre:common-split?expand=1, I suspect we first need to push the `ome-common-java` snapshot so that the Travis build can download them to their local Maven repository. But it is definitely the way to check the integration with out component.

Re https://github.com/rleigh-codelibre/bioformats/commit/a3a8dd6b9e6bdde3e5162b96cf153cfc8762196a#diff-2cccd7bf48b7a9cc113ff564acd802a8, I have not been touching these bits as I was not sure on how to treat these components in the developer documentation. Should we still list and treat `ome-common-java` and others as components of Bio-Formats or should we list them as OME dependencies e.g. alongside https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/java-library.html?

rleigh-codelibre:
Regarding the documentation, listing alongside the other third-party dependencies would also be fine with me.

rleigh-codelibre:
The changes are all correct, it builds and tests with ant and maven.  It looks fine to merge as is.

Which jobs are testing this?  The regular -DEV ones or are there specific 5.3 jobs?

sbesson:
Since the 5.3.0 is completely modifying the structure/content of the classical CI jobs (see https://trello.com/c/aoMEGJ9E/10-ci-ant-jobs), this is tested via a  set of jobs in a devspace. Once `dev_5_3` is more mature and ready to be merged back into `develop`, these job changes will be propagates back to the ci-master jobs. 

At the moment though, these CI jobs are only running the build + unit tests on the Maven & Ant build systems. There are no integration tests from `test-suite` running yet. I except to ramp up on these in the upcoming days. With regard to merging, we could wait until we have a few of these integration tests running or get this in to move forward with the next step of the decoupling. Main question for you is whether we should unblock this PR once https://github.com/ome/ome-common-java/pull/1 is merged.

rleigh-codelibre:
I think it would be fine to merge this before https://github.com/ome/ome-common-java/pull/1 is merged; I'll turn the first commit on my test branch into a followup PR to enable the use of the ome-common jar, once we have it as a downloadable released artefact.  I don't think we need to wait; what you have here is all good for the purposes of the 5.3 branch.

hflynn:
https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/components.html specifically states that it covers components of "the Bio-Formats code repository". I suggest we list formats-common with the dependencies on https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/java-library.html but reference the section from the component overview for 5.3 to acknowledge that this is the first release where it's been decoupled.

PR #2609, opened 2016-10-10
Ant: remove the pdf generation from the top-level documentation target

sbesson:
See https://trello.com/c/noU6gku8/14-review-docs-strategy

With this PR, the top-level `docs-sphinx` target used by the release jobs amongst others should no longer produce a PDF but still a zip containing the HTML sources /cc @hflynn 

While reviewing this we might want to make sure that the names of the Javadocs vs Sphinx docs zips are sensible/consistent as we will expose them in the downloads pages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,14 +24,13 @@ jdk:
 env:
   - BUILD=maven_findbugs
   - BUILD=maven
-  - BUILD=sphinx_html
   - BUILD=ant
 
 matrix:
   fast_finish: true
 
 before_install:
-  - if [[ $BUILD == 'sphinx_html' ]]; then pip install --user flake8 Sphinx==1.2.3; fi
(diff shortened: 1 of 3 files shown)


hflynn:
bio-formats-doc-VERSION.zip and bio-formats-javadoc.zip seems a fine naming scheme to me (although maybe we want to consider putting the version number in the javadoc zip too? Although I see most of the other zips don't have it so it's probably fine as is), job artifacts present and correct and builds fine locally too. 

Good to merge from me.

sbesson:
Thanks. Merging. Appending a version number was certainly proposed in https://github.com/openmicroscopy/bioformats/pull/2581. What are our thoughts about resurrecting this effort as part of 5.3 maybe by targetting the Javadocs first?

hflynn:
If people think it's valuable to have the version number, adding it to the javadocs now seems like a good plan to me.

PR #2614, opened 2016-10-13
Use ome-common 5.3.0 release from maven central

rleigh-codelibre:
- Switch to using maven central or ome-common in place of the old formats-common.
- Update the documentation to use ome-common and replace links which are no longer valid in the source tree; handle the differences between the split repos when building the documentation

Testing: Check that the ant and maven builds continue to work.  Check the documentation links.

Will the javadoc links need fixing up to use a different location for the ome-common documentation; it's currently using the -latest doc links, but they might disappear after merging?

/cc @sbesson

### DIFF
--- a/docs/sphinx/build.xml
+++ b/docs/sphinx/build.xml
@@ -21,15 +21,33 @@ Type "ant -p" for a list of targets.
   <target name="init" depends="release-version">
     <property name="sphinx.release" value="${release.shortversion}"/>
     <property name="sphinx.version" value="${release.major}.${release.minor}"/>
-    <property name="sphinx.source.branch" value="develop"/>
-    <property name="sphinx.source.user" value="openmicroscopy"/>
+    <if>
+      <isset property="sphinx.source.branch"/>
+      <then>
+        <property name="sphinx.bioformats.source.branch" value="${sphinx.source.branch}"/>
+        <property name="sphinx.common_java.source.branch" value="${sphinx.source.branch}"/>
+      </then>
+    </if>
+    <property name="sphinx.bioformats.source.branch" value="develop"/>
+    <property name="sphinx.common_java.source.branch" value="master"/>
+    <if>
+      <isset property="sphinx.source.user"/>
+      <then>
+        <property name="sphinx.openmicroscopy_source.user" value="${sphinx.source.user}"/>
+        <property name="sphinx.ome_source.user" value="${sphinx.source.user}"/>
+      </then>
+    </if>
+    <property name="sphinx.openmicroscopy_source.user" value="openmicroscopy"/>
+    <property name="sphinx.ome_source.user" value="ome"/>
     <property name="sphinx.omerodoc.uri" value="http://www.openmicroscopy.org/site/support/omero5.1"/>
 
     <copy file="conf.py.in" tofile="conf.py" overwrite="true"/>
     <replace file="conf.py" token="@sphinx_srcdir@" value="."/>
     <replace file="conf.py" token="@sphinx_builddir@" value="."/>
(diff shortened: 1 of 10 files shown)


docs/sphinx/developers/components.txt
```diff
@@ -11,12 +11,31 @@ can also be built with Maven by running :command:`mvn` in the component's
...
+
+Provides I/O classes that unify reading from files on disk, streams or files
+in memory, compressed streams, and non-file URLs.  The primary entry points
+are :javadoc:`Location <loci/common/Location.html>`,
+:javadoc:`RandomAccessInputStream <loci/common/RandomAccessInputStream.html>`
+(for reading), and :javadoc:`RandomAccessOutputStream
+<loci/common/RandomAccessOutputStream.html>` (for writing).
+
+In addition to I/O, there are several classes to assist in working with XML
+(:javadoc:`XMLTools <loci/common/xml/XMLTools.html>`), date/timestamps
```
sbesson: Having double checked, the Javadoc does no aggregate any dependency Javadoc incl. both external and OME dependencies. It should be possible to link to the `ome-common` Javadoc as we already do for e.g. https://docs.oracle.com/javase/8/docs/api/.

docs/sphinx/build.xml
```diff
@@ -21,15 +21,33 @@ Type "ant -p" for a list of targets.
...
     <property name="sphinx.omerodoc.uri" value="http://www.openmicroscopy.org/site/support/omero5.1"/>
 
     <copy file="conf.py.in" tofile="conf.py" overwrite="true"/>
     <replace file="conf.py" token="@sphinx_srcdir@" value="."/>
     <replace file="conf.py" token="@sphinx_builddir@" value="."/>
-    <replace file="conf.py" token="@sphinx_source_branch@" value="${sphinx.source.branch}"/>
-    <replace file="conf.py" token="@sphinx_source_user@" value="${sphinx.source.user}"/>
+    <replace file="conf.py" token="@bioformats_source_branch@" value="${sphinx.bioformats.source.branch}"/>
+    <replace file="conf.py" token="@common_java_source_branch@" value="${sphinx.common_java.source.branch}"/>
+    <replace file="conf.py" token="@openmicroscopy_source_user@" value="${sphinx.openmicroscopy_source.user}"/>
```
sbesson: The downside of is that these are external variables we might need to maintain across all documentation jobs (as we do not have yet the logic to read it from e.g. the pom).
For the current value i.e. 2 source code links in a see also box, I wonder if we should not simply drop these source code links from the Sphinx documentation.

hflynn:
Additions to components doc page read fine.

sbesson:
Discussed with @rleigh-codelibre and @hflynn. The agreement was to remove the source code links to the low-level components and simplify the logic to create links to the GitHub repo. On the Javadocs front, generating aggregated/cross-linked Javadocs sets is one of the remaining challenges post repository splitting. @rleigh-codelibre is investigating how to do this using Javadoc plugin but this might be tested as part of the upcoming `ome-model` decoupling and should not block the acceptance of this PR.

rleigh-codelibre:
@sbesson I've made the necessary changes which includes

- removing source links
- removing the ome-common repo information (I've left the ome user since that's going to be useful for repos under ome rather than openmicroscopy)
- making the javadoc links work for ome-common class references in the sphinx docs with a new extlink

The links in the ant and maven javadocs aren't present at the moment and will need fixing in a later PR or with a aggregate javadoc repo or similar.

sbesson:
All builds and unit tests are passing. The new components overview page in the Sphinx documentation makes sense and points at the correct Javadoc. I will create a card to review the Sphinx/Javadoc status post m1.

PR #2615, opened 2016-10-14
OME-TIFF: fix Plane population to preserve existing The* values

melissalinkert:
See
https://trello.com/c/aTV56qEX/6-fix-ome-tiff-regression-plane-population

To test, check the output of ```showinf -nopix -omexml``` with the files from QA 17375 and 16996 and verify that the ```The*``` values on the ```Plane```s are correct and that there are no relevant validation errors.  For QA 16996, ```TheZ``` and ```TheT``` were set in the file, but ```TheC``` was missing.  For QA 17375, all three values are set, but the order does not necessarily match ```DimensionOrder``` and some planes do not have a ```Plane``` element at all.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1022,13 +1022,10 @@ public class OMETiffReader extends FormatReader {
 
     MetadataTools.populatePixels(metadataStore, this, false, false);
     for (int i=0; i<meta.getImageCount(); i++) {
-      setCoreIndex(i);
-
       // make sure that TheZ, TheC, and TheT are all set on any
       // existing Planes
       // missing Planes are not added, and exising TheZ, TheC, and
       // TheT values are not changed
-      ArrayList<Integer> populatedPlanes = new ArrayList<Integer>();
       for (int p=0; p<meta.getPlaneCount(i); p++) {
         NonNegativeInteger z = meta.getPlaneTheZ(i, p);
         NonNegativeInteger c = meta.getPlaneTheC(i, p);
@@ -1046,11 +1043,8 @@ public class OMETiffReader extends FormatReader {
           t = new NonNegativeInteger(0);
           metadataStore.setPlaneTheT(t, i, p);
         }
-
-        populatedPlanes.add(getIndex(z.getValue(), c.getValue(), t.getValue()));
       }
     }
-    setCoreIndex(0);
     for (int i=0; i<acquiredDates.length; i++) {
       if (acquiredDates[i] != null) {
         metadataStore.setImageAcquisitionDate(
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1020,7 +1020,37 @@ public class OMETiffReader extends FormatReader {
       }
     }
 
-    MetadataTools.populatePixels(metadataStore, this, true, false);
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
This is certainly an improvement on the existing behaviour.  However, a few design questions:

- Why is this correction done in the `OMETiffReader` rather than `MetadataTools` as for the existing metadata correction logic?
- While this is filling a required field, it's still making changes to user data which may be quite wrong; making all coordinates zero if unset is probably the best thing we can do, but I question whether we should be doing this **at all**.  Maybe throwing an exception would be more appropriate than altering data?  There is a limit to the amount of assumptions we can make when mangling invalid data, and at some point it's safer to give up rather than tolerating bad input blindly.
- Could or should we be doing this sort of correction in the transforms rather than in code?

melissalinkert:
The correction is done in ```OMETiffReader``` since other readers should not have anything in their ```MetadataStore``` prior to ```MetadataTools.populatePixels``` being called.

Throwing an exception instead is an option, and if that's the majority opinion, that's fine.  If the invalidity is caught during ```isThisType```, this means that the file would be picked up by the plain TIFF reader (which means incorrect ZCT dimensions and limited access to any other metadata defined in the OME-XML); if caught later on during ```setId```, the file will likely not be importable at all.

I can see the case for doing this in transforms instead, but won't have time to investigate what that would look like in time for the proposed 5.2.4 release date.

Really my goal here was to do a bit better than a straight up revert of the commit that introduced this regression, since that would arguably be a regression as well.  A separate longer-term plan for how and how much invalid data to handle in OME-TIFFs is certainly in order, but that's not within the scope of what this PR was trying to accomplish.

sbesson:
My 2c on the 3 questions above in disorder:
- re validation/autocorrection, with regard to milestones, I would suggest we actively start driving this via the [extended options API](https://trello.com/c/HLILyhL0/12-metadataoptions-api) we are planning for 5.3.0, give the user more control to decide whether to validate/auto-correct and discuss what should be default
- implementing the fix at the transform level is also something that would be a valid 5.3.0 target using the new model repo. However, am I correct that this would not work in the case of OME-TIFF files using the same schema as Bio-Formats schema as no transform would apply?

rleigh-codelibre:
Regarding the transforms, we would likely need to make it apply a "fixes" transform in addition the standard set.  We might want to make that generic so it can be sourced by the existing transforms as well, so we correct on upgrades so we have valid OME-XML at all stages.

sbesson:
Functionally, I can reproduce the metadata issue with the last release Bio-Formats and the QA 17375

screen shot 2016-10-14 at 16 26 50

and the timestamp ordering is fixed with this PR

screen shot 2016-10-14 at 16 27 43

As for QA 16996, I can confirm the OME-XML is still invalid as noted in the original PR but the `Plane` required `theC` field remains filled with the default `0` value.

rleigh-codelibre:
@melissalinkert I think the rationale makes sense and that this is fine as an incremental improvement.  There's just the couple of questions I asked on the PR about the redundant bits of code.

melissalinkert:
Both code review comments above should be resolved by 398a6b6.

rleigh-codelibre:
@melissalinkert Yes, that looks fine, thanks.

sbesson:
[The automated tests on the full repository](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/320/) passed with the latest state of this PR so did the [test of the QA file](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/323/) submitted in the original thread. Merging for 5.2.4.

PR #2620, opened 2016-10-17
Remove findbugs from Travis and simplify the yml configuration file

sbesson:
The `findbugs:findbugs` Maven phase was added to the Travis test script as a way to flag errors. However this is generated a report while the command to fail the build is `findbugs:check`. The latter generates many warnings across components which cannot be fixed in the main Bio-Formats source code without significant work. Instead these should be fixed and activated in each decoupled component.

This commit removes `findbugs:findbugs:` from the Travis test and simplifies the YML file to clarify the fact we are testing a 2x2 matrix (Ant/Maven with JDK 7/8).

To test this PR, check Travis is green and the time of  `oraclejdk8/maven` component is reduced to match roughly the time of the `openjdk8/maven` component.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,16 +13,12 @@ cache:
 addons:
   apt_packages:
     - git
-    - cmake
-    - build-essential
-    - libboost-all-dev
 
 jdk:
   - oraclejdk8
   - openjdk7
 
 env:
-  - BUILD=maven_findbugs
   - BUILD=maven
   - BUILD=ant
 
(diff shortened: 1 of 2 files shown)


mtbc:
Should we create a card for coordinating a findbugs cleanup effort with activating `findbugs:check` as the last step or would that be too ambitious?

sbesson:
@mtbc given the output of the command, my impression is certainly that activating `findbugs:check` at the level of the entire Bio-Formats source code would be quite ambitious. Using the ongoing decoupling of the codebase into smaller repositories to fix the split code and let the `findbugs` check pass seems more feasible. Thus I would be inclined to start driving this cleanup at this level - see e.g. https://github.com/ome/ome-common-java/issues/3 for the `ome-common` component.

rleigh-codelibre:
The changes look fine and travis is green.  Please merge.

Seconded on scheduling findbugs cleanup in the split components.  Doing it bioformats in due course would also be desirable if possible.

PR #2623, opened 2016-10-18
Fix other ZEISS links

hflynn:
Fixes up the ZEISS links which may be redirected but are still upsetting the build -https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/443/warnings3Result/

(I also fixed up the CZI links while I was at it, given that was redirecting too, and linking to the CZI specific page twice was inconsistent with the other pages where the ZEISS developer link goes just to the homepage)

Should make the merge build green again.

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -2377,8 +2377,8 @@ writer = WlzWriter
 
 [Zeiss Axio CSM]
 extensions = .lms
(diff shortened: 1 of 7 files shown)


hflynn:
build is green - https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/444/

PR #2624, opened 2016-10-20
Remove ome-poi component

dgault:
This PR is a follow up to https://github.com/ome/ome-poi/pull/1#issuecomment-255055401 

The formats-gpl POM has been updated to consume the new ome-poi 5.3.0-SNAPSHOT for testing.

I have tested locally by doing a full clean and Maven build. The build was successful and I can see the correct snapshot jar in ~/.m2/repository/org/openmicroscopy/ome-poi


sbesson:
Looks good. Is the plan to remove the `components/ome-poi` component and all related build targets in a separate PR or should we do it as part of this one?

dgault:
We can modify this PR sure and do it all in one go. I will update it with all of the targets and the component removed.

sbesson:
Last commit has been tested via the `devspace` jobs, both Ant and Maven builds are green, so is Travis. Tested briefly using the command-line tools generated by the Ant job. The packaging works as expected:

```
$ jar tf bftools/bioformats_package.jar  | grep ome-poi
META-INF/maven/org.openmicroscopy/ome-poi/
META-INF/maven/org.openmicroscopy/ome-poi/pom.properties
META-INF/maven/org.openmicroscopy/ome-poi/pom.xml
```

and `showinf` run against a few file formats using the poi fork (simplepci, fv1000) works as expected (metadata and pixel data). Additionally the OMERO Java integration tests have been running using the 5.3.0-SNAPSHOT and are passing.

Barring any veto, I suggest we get this PR in and launch the automated tests on the curated data repository or minimally the affected formats. We can parallely work on the release of ome-poi:5.3.0 to Central Maven and open a follow-up PR to bump the component version.

PR #2625, opened 2016-10-21
Bugfixes for Leica LIFReader

michaelgoelzer:
Fixed the following issues:
- Pixel size calculated incorrectly
- import of combined timeseries/tilescan resulted in messed up frame order
- timestamps for recent lif files were not imported


components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -316,21 +317,14 @@ public boolean isThisType(RandomAccessInputStream stream) throws IOException {
...
-    // seek instead of skipBytes to prevent dangerous int cast
-    in.seek(in.getFilePointer() + tile * planeSize * getImageCount());
-    in.skipBytes(bytesToSkip * getSizeY() * no);
+    seekStartOfPlane(no, offset, planeSize);
 
     if (bytesToSkip == 0) {
       readPlane(in, x, y, w, h, buf);
     }
     else {
+      // seek instead of skipBytes to prevent dangerous int cast
```
mtbc: Slightly confused here: "instead of skipBytes" but appears to use skipBytes? :+1: for comments though!

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -1049,6 +1081,8 @@ private void translateMetadata(Element root) throws FormatException {
 
     tileCount = new int[imageNodes.size()];
     Arrays.fill(tileCount, 1);
+    tileBytesInc = new long[imageNodes.size()];
+    Arrays.fill(tileBytesInc, 0);
```
mtbc: The Java spec guarantees that a new long[] is 0-filled.

components/formats-gpl/src/loci/formats/in/LIFReader.java
michaelgoelzer: you are right, that is silly. The comment was there before (old file line 326), apparently i did not move it to the correct location (new file line 378)

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -1947,7 +2020,7 @@ else if (unit.equals("m")) {
             if (ms.sizeZ == 1) {
               ms.sizeZ = len;
               bytesPerAxis.put(nBytes, "Z");
-              physicalSizeZ = (physicalLen * len) / (len - 1);
+              physicalSizeZ = physicalLen;
```
sbesson: Note this change should result in the same value of the physical size Z but will affect all the physical sizes in X and Y which should force us to regenerate the configuration files for the automated testing. @melissalinkert: in https://github.com/openmicroscopy/bioformats/commit/a99279ff4c0c8bb8519b4a938168ec2742a28ebc, was there a rational for correcting `physicalLen` for the Z dimension but not the X and Y?

michaelgoelzer: This was detected because for non-equal x/y dimension sizes the bioformats importer assumed the pixels were not square anymore. If you read in e.g. a 1024*512 pixel image and the pixels are not square according to ome-xml metadata, this should be a failed test case. I can provide lif files if needed.

michaelgoelzer:
Hey sbesson and mtbc, thanks for the very quick replies.
Unfortunately I will be unavailable until 2016-11-02, so apologies for leaving this here and then not being able to follow up quickly. But once I return, I will address any issues/questions.

components/formats-gpl/src/loci/formats/in/LIFReader.java
michaelgoelzer: I think I fixed the issue with unreasonable acquisition dates, and also corrected a bug which overcompensated local time zones (compensation was done despite lif timestamps being in UTC to begin with).

As far as I understand, the timestamps we write in the OME-XML are still ambiguous though. loci.common.DateTools.convertDate calculates a UTC date, but I believe the returned string format "yyyy-MM-dd'T'HH:mm:ss" is typically interpreted as local time. But I am by no means an expert on this topic, and it is also probalby out of scope for this pull request.

Please let me know if I can do anything to help with the pixel size topic (e.g. provide example files) or anything else. Apologies for the long delay, I was ordered to finish some other assignment first.

sbesson: Thanks @michaelgoelzer. With the latest state of this PR, all [automated test failures](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/377/) against our existing Leica LIF samples are related to physical size differences in X and Y which need to be adjusted on our side.

Re pixel size sample files, I found a couple of Leica LIF files (`curated/leica-lif/oliver/Multi-channel Measurements.lif`) in our repository with mismatching X/Y dimensions. We'll use them for testing this PR and let you know if we need additional files.

melissalinkert: Apologies for being late to the discussion here.  The original rationale for dividing ```physicalSizeZ``` by ```len - 1``` but dividing ```physicalSize{X,Y}``` by ```len``` is that we typically model the physical Z size as the physical distance between two Z sections, or the distance the stage has to move (i.e. the plane itself has 0 thickness).  Physical X and Y sizes are typically modeled as the width and height of each pixel, not the space between two pixels.  As an example, a 1x1 image with 1 Z plane would typically have non-0 physical width and height, but physical Z size of 0.  With the current logic, the same 1x1x1 image would have all 0 physical sizes.

michaelgoelzer: I agree that what you describe is the most intuitive way to interpret sizes. The decision to not store it like that in lif-files was probably driven by the desire to make no distinction between x, y and z dimensions.
And since setting z-size to a nonzero value for "2D" images would probably confuse users quite a lot (even if that would arguably be the most honest representation), treating x and y similar to z was the route that was chosen.

sbesson:
Hi @michaelgoelzer, thanks for the last commit pushes. A few comments and suggestions:

- having reviewed the specification, the pixel sizes XY fix makes complete sense especially in the context of analytical workflow. Internally, we will need to updating configuration files for the automated tests against all our Leica LIF samples. Additionally this change might require a very clear documentation of its impact for anyone who has been reading this metadata using Bio-Formats. In that sense, we might consider adding a legacy option allowing to read the pixel size the old way (see https://github.com/openmicroscopy/bioformats/pull/2681) .
- the timestamp improvements also address a ticket reported as https://trac.openmicroscopy.org/ome/ticket/12793. There is an open question there on what should be the behavior if the number of timestamps is lower than the dimension count.
- this PR is quite large and affects different aspects of the Leica-LIF format, my main concern is the testing but also the fact that the entire set of changes remains on hold until the PR is merged. Can I suggest we split this PR into 2 or 3 separate individual ones, one per bug fix? This would allow a more efficient review and discussion but also potentially a faster distribution of some bug fixes.

michaelgoelzer:
As for the question about what if the number of timestamps is lower than the dimension count:
In this case I would assume that the acquisition was stopped prematurely. In the example in the ticket we have:
* 2 channels, 21 z-planes, 3 timepoints
* 94 timestamps (where we would expect 2*21*3 = 126)
This can be interpreted as follows: The first two timepoints were acquired in full (2*21*2 = 84 timestamps), the acquisition was stopped after 10 planes of the third timepoint were acquired. The remaining 11 z-planes lack a timestamp, but they should also lack images (i.e. these frames are black). I can not verify this for the file in question because I can't access it, but barring a bug during acquisition this is the planned behaviour. Timestamps of non-acquired frames do not get some kind of default value in LAS X, they simply do not exist.

Splitting this PR into separate ones: Sure, why not? I will try to, and hope I don't break everything since I am not very familiar with git/github.

As for the legacy option for the pixel format: In general I am not overly fond of the idea of purposefully using a bugged version, but I also don't have enough insight into what kinds of analysis toolchains might be used to decide how big of a problem that change would pose.

michaelgoelzer:
Split into pull requests
#2727 
#2728
#2729

sbesson:
Thanks @michaelgoelzer. Closing this in favor of the new PRs where the discussion can keep happening.

PR #2627, opened 2016-10-21
Update ome-poi to 5.3.0

dgault:
This PR is a follow up to https://github.com/openmicroscopy/bioformats/pull/2624

The ome-poi version has been changed to 5.3.0 from 5.3.0-SNAPSHOT.
The repository previously used for the snapshot has been removed in favour of Maven central.

To test: Verify that builds remain green


sbesson:
Both Travis and the set of CI jobs ran to completion with this PR. Merging.

PR #2631, opened 2016-10-24
Remove specification, xsd-fu and ome-xml components, and all C++ code

rleigh-codelibre:
Use the jars from the new https://github.com/ome/ome-model repository in place of the removed components

- Remove the specification, xsd-fu and ome-xml components
- Update maven build to use external specification and ome-xml jar
- Remove the components from the ant build
- Remove all C++ code under `cpp`

Testing: Check builds remain green.  Check build works with maven and ant, and that all tests pass.


rleigh-codelibre:
`714 files changed, 21 insertions(+), 199818 deletions(-)`

Nearly made it to 200kloc deleted.  Are there 182 more lines we can delete?

hflynn:
https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/components.html will also need updating but given we don't have the staging builds sorted yet, it may be better to do that clean-up on a following PR.

hflynn:
I guess https://www.openmicroscopy.org/site/support/bio-formats5.2/developers/xsd-fu.html should probably move to the new repo too really, or is (some of) this content still relevant to the main repo?

rleigh-codelibre:
@hflynn I think all the model-related docs could be moved here soon.

rleigh-codelibre:
@hflynn Regarding the *5.2* docs link, since these changes are not for 5.2, I don't think those docs need updating.  The docs for 5.3+ should be updated.

hflynn:
@rleigh-codelibre yes, sorry, I pasted the 5.2 link because the 5.3 docs aren't being built as yet, I should have removed the version specific bit of the link to avoid confusion.

melissalinkert:
Only minor nitpick from my side is that it might be nice to define the version for ```org.openmicroscopy:ome-xml``` in the top level ```pom.xml```, as we do for other dependencies.  Otherwise, this looks fine to me assuming builds all pass.

dgault:
Both ant and maven builds look good and are building correctly (using a clean build). 
The 5.3.0-SNAPSHOT jars are being used as expected. Changes to the POM files and the ant build files all look good and make sense.

PR looks good to merge.

sbesson:
NB: 5.3.0 model artifacts are now uploaded and available via [Maven Central](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.openmicroscopy%22)

PR #2633, opened 2016-10-25
Remove mdbtools component

dgault:
This is a follow up PR to https://github.com/ome/ome-mdbtools/pull/1

- Remove the mdbtools components
- Update maven build to use external mdbtools snapshot jar - https://oss.sonatype.org/content/repositories/snapshots/org/openmicroscopy/ome-mdbtools/
- Remove mdbtools from the ant build

Testing: 
- Builds should remain green
- Clean maven build should pass
- Ant build should pass
- All tests should continue to pass


components/formats-gpl/pom.xml
```diff
@@ -208,7 +208,7 @@
                 <additionalClasspathElement>${basedir}/../../ant/</additionalClasspathElement>
               </additionalClasspathElements>
               <classpathDependencyExcludes>
-                <classpathDependencyExclude>${project.groupId}:mdbtools-java</classpathDependencyExclude>
+                <classpathDependencyExclude>org.openmicroscopy:ome-mdbtools</classpathDependencyExclude>
```
sbesson: A corresponding change should also be done in `components/formats-gpl/build.xml` in the `test-no-mdb` target to exclude the `ome-mdbtools` jar.

sbesson:
With the last commit, the `test-no-mdb` passes as expected. Waiting for the Bio-Formats tests and the OMERO integration tests to complete and Travis to turn green. Also tested with a couple of sample Zeiss LSM files which use the MDB service by opening them and comparing the OME-XML output between this PR and 5.2.4. All works as expected.

PR #1912, opened 2015-08-04
Fix automated tests to handle symlinks with external configuration

sbesson:
Follow up of https://github.com/openmicroscopy/bioformats/pull/1894 and https://github.com/openmicroscopy/bioformats/pull/1819.

While the external configuration logic works accurately on Linux and Windows for the main folder, this PR attempts to solve the case of data folders containing symlinks e.g. `test_images_good`. With the data repository structure being organized as follows:

```
/ome/data_repo/test_images_good      # folder containing symlinks to the data /ome/data_repo/from_skyking/
/ome/data_repo/config/breaking           # folder containing configuration files
```
the goal of this PR is to resolve the symlinks and the configuration files with the following command:

```
ant test-automated -Dtestng.directory=/ome/data_repo/test_images_good -Dtestng.configDirectory=/ome/data_repo/config/breaking/test_images_good
```

The main caveat of this PR is the extensive usage of `File.getCanonicalPath()` which may have some impact on performance while scanning files. This performance should be assessed both on Linux and Windows

To be tested via the breaking jobs.

/cc @melissalinkert

PR #2635, opened 2016-10-25
Revert changes to handle symlinks with external configuration

simleo:
Reverts #1912.

Discussed with @sbesson: these changes were needed to support the peculiar link structure in `test_images_good` and should not be needed anymore now that we're moving to a "proper" subset. On one hand, this revert simplifies the code; on the other hand, I've found out that the path relocation feature added in #1912 was generating wrong paths with the new subset layout.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/ConfigurationTree.java
+++ b/components/test-suite/src/loci/tests/testng/ConfigurationTree.java
@@ -98,19 +98,9 @@ public class ConfigurationTree {
     if (rootDir == null) {
       throw new IllegalArgumentException("rootDir cannot be null.");
     }
-
+    this.rootDir = new File(rootDir).getAbsolutePath();
     if (configDir != null) {
-      Location rootLocation = new Location(rootDir);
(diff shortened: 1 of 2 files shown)


sbesson:
Moving out of breaking - see https://github.com/openmicroscopy/bioformats/pull/2601#issuecomment-257176167

sbesson:
Same comment as https://github.com/openmicroscopy/bioformats/pull/2601#issuecomment-257415213 with regard to the testing impact.

This cleans up a lot of logic introduced specifically for the `test_images_good` folder. As we redesigned our subset folder, there is no remaining rationale for maintaining this logic. The only remaining consideration is whether the `TestTools` methods removed in 779496f should be considered as public API and deprecated rather than dropped.

sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/347 works against a symlinked subset (`formats_subset`) with this PR included. Merging.

PR #2637, opened 2016-10-25
All component versions are defined in root pom

rleigh-codelibre:
Also fix up doc links to use versioned ome-xml javadocs, as for ome-common.

Testing: Check and and maven builds continue to work and that jobs remain green.  Check matlab links for MetadataStore/Retrieve.

### DIFF
--- a/ant/global.properties
+++ b/ant/global.properties
@@ -15,15 +15,3 @@ artifact.dir = ${root.dir}/artifacts
 # copyright strings to use when generating javadocs
 copyright.begin = <i>Copyright &#169;
 copyright.end   = Open Microscopy Environment</i>
-
-domain.prefix = org.openmicroscopy
-
-log4j.version = 1.2.17
-logback.version = 1.1.1
-slf4j.version = 1.7.6
-kryo.version = 2.24.0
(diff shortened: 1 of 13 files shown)


rleigh-codelibre:
@sbesson The "interesting side effect" is definitely an unintentional feature.  It was on my list of things to check in more detail.

Now the C++ is gone entirely, we could strip out all the gitversion stuff.  While the source tree contains jars, we still need the -dfsg source release so retaining the release script would be useful for the time being.

The Sphinx docs when built via ant should pick up the same versions if we pass them through, so we could certainly make it use the same scheme.

sbesson:
Discussed the impact/benefits of this PR with @rleigh-codelibre @dgault @hflynn and @jburel. Using a single release version definitely reduces maintenance and should not affect the documentation workflow.

In terms of development workflow but also release testing, this means that we cannot use staging tags anymore and will need to bump the `pom.xml` versions for producing release candidates/milestones releases.

Merging this so that @dgault can bump `ome-mdbtools.version` and @rleigh-codelibre can open a follow-up tidy up for the Sphinx version.

PR #2638, opened 2016-10-26
Move decoupled components & fix links

hflynn:
This fixes https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/453/warnings3Result/ or will do once it's merged in the right branch!

Restructured the component developer page to list the external components at the bottom and replaced the source code links with full external links for now - we can replace with proper `:extlinks:`  when we've decided if we want them to be snoopy compatible or just simple replacements.

cc @sbesson


docs/sphinx/developers/components.txt
```diff
@@ -255,11 +184,82 @@ tests) are included in this component.  These tests are primarily run by the
...
+.. _specification:
+
+`Model specification <https://github.com/ome/ome-model/tree/master/specification>`_:
+
+All released and in-progress OME-XML schema files are contained in this
+component.  The specification component is also the location of all XSLT
+stylesheets for converting between OME-XML schema versions, as well as example
+OME-XML files in each of the released schema versions.
+
+.. _xsd-fu:
```
sbesson: Briefly, following https://github.com/ome/ome-model/pull/2, `xsd-fu` is no longer a proper component, is no longer exposed and consumed and eventually might migrate inside `ome-xml` as a source generation tool. I would be inclined to remove from this page and the related `developers/xsd-fu` one. The latter can be resurrected as part of the model and formats documentation when it gets migrated to the ome-model repo.

hflynn:
xsd-fu removed. And yes, a card for dev docs edits/updates needed for 5.3 would be very handy, thanks.

docs/sphinx/developers/components.txt
rleigh-codelibre: Additionally, the xsd-fu documentation dates back to when it had to be run by hand once per schema release.  It's both self-documenting in that all the options are documented in the tool, and how to invoke it is documented by the build systems which invoke it, so running it by hand isn't something which is needed in practice.  Even when we are working on it or the templates, we test it by running ant/cmake, not by hand-crafting a command-line.  So I don't think there's much use for this page; we can certainly add references to the relevant bits of the builds in the new repo though.

PR #2639, opened 2016-10-26
Bump ome-mdbtools version to 5.3.0

dgault:
Bumps the version of ome-mdbtools from 5.3.0-SNAPSHOT to 5.3.0
- Removes the test repo as jar is now located on maven central
- Uses the new versioning as introduced in https://github.com/openmicroscopy/bioformats/pull/2637


sbesson:
Travis is green so are the devspace builds. Integrating into the `dev_5_3` branch

PR #2644, opened 2016-10-27
Build versioning from maven

rleigh-codelibre:
- Update source-archive script
- Don't embed or use git version or release date in `version.properties` (remove `version.properties`)
- Don't use the shortversion
- Embed all version metadata in `MANIFEST.MF`; update ant build to match what maven embeds, and convert `FormatTools` to use this data in place of the property file.
- Improve consistency of version/date/git commit in the various about/help messages (version; date; git commit/tag)
- Update docs to use the maven version
- Drop unnecessary source copying with ant

Testing: check ant, maven builds, docs versioning, `ant release` and that the source release builds with ant, maven, `ant docs-sphinx`.


sbesson:
Two general comments while looking at the PR content. 

- on the release versioning front, this PR is a follow-up of https://github.com/openmicroscopy/bioformats/pull/2637 and effectively removes the historical discrepancy between the Ant build system (which uses Git-tag based version) and the Maven build system (which uses the top-level pom.xml version). As mentioned on the previous PR, my impression is that this effectively reduces unnecessary complexity.
- on the VCS revision front, I am a bit more worried about the proposed changes and the information lost. While release artifacts should ideally be uniquely identified by their release version provided the `groupId` does not vary, this is not the case in development mode when working with SNAPSHOTs. Introspecting the revision via the `FormatTools` API definitely had some utility in typical scenarios like testing the command-line tools with and without a PR or for automated/integration tests  which are consumed from an unkowne SNAPSHOTS installed in the shared `~/.m2` repository. Is there some alternate way to return the revision which produced a set of artifacts?

I will carry out some further testing on Monday.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-DEV-merge-push#783](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/783/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/783/consoleText) for more details.
Possible conflicts:
  - PR #2643 snoopycrimecop 'Update POM version numbers to 5.3.0-SNAPSHOT'
    - pom.xml

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-DEV-merge-push#785](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/785/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/785/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - pom.xml

rleigh-codelibre:
@sbesson The changes subsequently made should address the concerns about the loss of git versioning.  This is restored, but rather than hardcode it via the properties file, we use the jar manifest to introspect it.  There are two minor downsides:

- the version information is not present when running the formats-api unit tests (since it's not using the jar, but the plain class files); this doesn't affect the tests at all, but just mentioning
- the behaviour when packed into an uber-jar might show different versions; in the case of bioformats-package and loci-tools, this isn't a problem, since they have identical metadata.  If we're intending to move away from uber-jars this is less of a problem; it will only affect people who repack our stuff into their uber-jars, and even then it's purely informational so won't break anything.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-DEV-merge-push#786](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/786/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/786/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - pom.xml

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-DEV-merge-push#787](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/787/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/787/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - pom.xml

rleigh-codelibre:
@sbesson @dgault I've fixed up the version conflict so this should now be reviewable.

components/formats-gpl/build.properties
```diff
@@ -13,9 +13,8 @@ component.java-version   = 1.7
 component.deprecation    = true
 
 component.resources-bin  = loci/formats/bio-formats-logo.png \
-                           loci/formats/meta/*.xsl \
                            loci/formats/utests/2008-09.ome \
                            lib/**/*.dll
-component.resources-text = loci/formats/*.txt
+component.resources-text =
```
dgault: Can this property just be removed altogether?

rleigh-codelibre: Probably.  This is purely for consistency with the other components.

dgault:
The changes look fine from my perspective and the builds for both ant and maven are working as expected with the correct versioning. The docs build and versioning also looks to be correct and as expected.

I've been trying to think of alternative possibilities to using the manifest but haven't come up with any better proposals.

rleigh-codelibre:
If we prefer to stick with the properties file, it's easy enough to revert the use of the manifest back to the properties file, and retain the other improvements.

dgault:
Im happy with the move to the manifest from my side anyway, I cant see it being problematic and should provide the required info in any scenario in which it is needed.

components/formats-bsd/src/loci/formats/Memoizer.java
```diff
@@ -578,9 +576,10 @@ public boolean versionMismatch() throws IOException {
       }
 
       // REVISION NUMBER
-      if (!versionChecking && !FormatTools.VCS_REVISION.equals(revision)) {
-        LOGGER.info("Different Git version: {} not {}",
-          revision, FormatTools.VCS_REVISION);
+      if (!versionChecking &&
+          FormatTools.VERSION.endsWith("-SNAPSHOT")) {
+        LOGGER.info("Development version: {} not {}",
+          FormatTools.VERSION);
```
melissalinkert: Missing parameter here?

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -201,45 +205,95 @@
...
+
+    if (attr.getValue("Implementation-Version") != null)
+      VERSION = attr.getValue("Implementation-Version");
+    else
+      VERSION = "(unknown version)";
+    CREATOR = "OME Bio-Formats " + VERSION;
+    if (attr.getValue("Implementation-Build") != null)
+      VCS_REVISION = attr.getValue("Implementation-Build");
+    else
+      VCS_REVISION = "(unknown revision)";
```
melissalinkert: Could the if/else blocks here please have ```{``` and ```}``` for readability and consistency with the rest of the repo?

components/formats-bsd/src/loci/formats/Memoizer.java
```diff
@@ -578,9 +576,10 @@ public boolean versionMismatch() throws IOException {
       }
 
       // REVISION NUMBER
-      if (!versionChecking && !FormatTools.VCS_REVISION.equals(revision)) {
-        LOGGER.info("Different Git version: {} not {}",
-          revision, FormatTools.VCS_REVISION);
+      if (!versionChecking &&
+          FormatTools.VERSION.endsWith("-SNAPSHOT")) {
```
melissalinkert: I wonder if it makes sense to have a ```boolean IS_SNAPSHOT``` (or similar) field in ```FormatTools```?

rleigh-codelibre: If it is something which needs checking in multiple places that would make sense.  Does any other code need to know this?

rleigh-codelibre:
@melissalinkert I've pushed some extra comments to address your comments.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -201,45 +205,95 @@
...
+
+  static
+  {
+    Manifest manifest = loadManifest();
+    Attributes attr;
+
+    if(manifest != null)
+      attr = manifest.getMainAttributes();
+    else
+      attr = new Attributes();
```
melissalinkert: One more set of missing braces?

components/formats-bsd/src/loci/formats/Memoizer.java
melissalinkert: I don't think we're currently checking for this elsewhere, I can just imagine that it might be useful going forward, especially for external API consumers.

sbesson:
See https://ci.openmicroscopy.org/view/Breaking/job/BIOFORMATS-DEV-breaking-build/667, the Javadoc target need to be adjusted for the changes in the source classes.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -201,45 +205,96 @@
...
+   */
+  public static final String VCS_REVISION;
 
-  /** Current VCS revision (short form). */
-  public static final String VCS_SHORT_REVISION =
-    VERSION_PROPERTIES.getProperty("vcs.shortrevision");
+  /** Current VCS revision (short form).
+   * @deprecated Use the general {@link #VCS_REVISION} field.
+   */
+  public static final String VCS_SHORT_REVISION;
```
sbesson: Also use an `@Deprecated` decorated to warn on compile?

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -201,45 +205,96 @@
...
 
   /** Value to use when setting creator/software fields in exported files. */
-  public static final String CREATOR = "OME Bio-Formats " + VERSION;
+  public static final String CREATOR;
 
-  public static final String PROPERTY_FILE = "version.properties";
+  /**
+   * @deprecated The property file is no longer used.
+   */
+  public static final String PROPERTY_FILE = null;
```
sbesson: Same as above

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -201,45 +205,96 @@
...
 
+  /**
+   * @deprecated This method should no longer be used.  The properties
+   * are now obtained from the jar manifest.
+   */
   static Properties loadProperties() {
```
sbesson: Same as above

rleigh-codelibre:
@sbesson Green in https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-breaking-build/668/

OK to remove the breaking label?

rleigh-codelibre:
@sbesson Can this be merged?

sbesson:
@rleigh-codelibre: from https://github.com/openmicroscopy/bioformats/pull/2644#pullrequestreview-6802551, there was just one question/proposal to use `git rev-parse HEAD` rather than `git describe` for the `Implementation-Build` metadata field to unify further Ant/Maven. Is that something that should be done in a separate PR?

rleigh-codelibre:
@sbesson No, sorry, lost track of that.  I'll do that tonight.

rleigh-codelibre:
@sbesson Done.

sbesson:
Retested with the last commit, the command-line tools are now working as expected

```
$ ./bftools/showinf -version
Version: 5.3.0-SNAPSHOT
Build date: 4 November 2016
VCS revision: 29ccb495af92043896c258e37ecbaf8f3ab988ab
```
so is the ImageJ plugin

screen shot 2016-11-04 at 15 05 55

The missing manifest metadata for the Maven bundles being captured elsewhere, merging this. Thanks for the cleanup.

PR #2645, opened 2016-10-28
Fix loci_tools bundle name to use underscore

sbesson:
The 5.3.0-m1 work introduced a bundle name change replacing the underscore by dash. This commit reverts this change and also unifies the module directory name to match the bundle name.


dgault:
The renaming all look fine and I couldnt find any other references. Using maven locally I built the loci_tools jar which was named correctly and tested it using the Fiji test cases normally used for release testing. Everything worked as expected.

Looks good to merge

PR #2622, opened 2016-10-18
Bio-Formats 5.3: components re-architecture

sbesson:
This branch re-integrates the first part of the Bio-Formats 5.3.0 work into the main development line.

### Build system

- drop embedded JARs and use the Maven Ant Tasks plugin to unify the dependency using the `pom.xml` #2599 
- drop deprecated osgi targets, OME Tools bundle and `ome-jxr` component #2603 #2604 #2605 
- remove PDF generation from the `docs-sphinx` target #2609
- suffix Javadoc zip bundle with version number #2611 
- migrate unit tests out of `test-suite` into `formats-bsd` #2606 #2607 
- simplify Travis build #2620
- store all versions in the top-level `pom.xml` #2637

### Component architecture

- remove `formats-common` component decoupled to https://github.com/ome/ome-common-java and consume `org.openmicroscopy:ome-common` artifact from Maven Central #2608 #2614 
- remove `ome-poi` component decoupled to https://github.com/ome/ome-poi and consume `org.openmicroscopy:ome-poi` artifact from Maven Central #2624 #2627 #2636
- remove `specification`, `xsd-fu` and `ome-xml` components decoupled to https://github.com/ome/ome-model and consume `org.openmicroscopy:{specification,xsd-fu,ome-xml}` artifacts from Maven Central #2631
- remove `mdbtools` component decoupled to https://github.com/ome/ome-mdbtools and consume `org.openmicroscopy:ome-mdbtools` artifacts from Maven Central #2633 #2639 
- developer documentation changes for the decoupled components #2638

PR #2646, opened 2016-10-30
Improve Ant jar and bundle target to honor maven.repo.local variable

sbesson:
## Background

See https://trello.com/c/MeIS9tyf/21-test-jobs-rearchitecture. The migration of the Ant build system to use `Maven Ant Tasks` in https://github.com/openmicroscopy/bioformats/pull/2622 greatly improved the build system and dependency management.

Prior to 5.3.0-m1, the Ant build was fully contained within the source code directory using embedded dependencies and generating all artifacts under `artifacts`. With the change above, the build logic now uses the Maven repository both for the project dependencies but also the individual components.

One downside of this approach has been reported in the CI environment where builds were previously implicitly relying on the build containment and are now failing due to race conditions and/or overwriting SNAPSHOT artifacts which could be used by other jobs.
## Changes

Similar to Maven, a property was introduced in https://github.com/openmicroscopy/bioformats/commit/f948bc1ad9aa6020c8d7c7d0878c087b2ce2707c allowing to override the local repository. However as it was only introduced for the dependency resolution, it made its usage incomplete.

This commit should restore the symmetry with the `mvn-init` task and use the repository defined by `maven.repo.local` if set for installing the component or bundle JAR.
## Testing

To test this PR, it should be possible to build Bio-Formats from the top-level directory using a custom repository and the Ant build system:

```
$ ant clean jars -Dmaven.repo.local=/tmp/bf_ant_repo
```
## Application

In the context of the CI testing, using this property should allow to differentiate the jobs which need to share artifacts using the default Maven repository under  `~/.m2` e.g. the build job and the integration/repository jobs from standalone job. An immediate candidate for the latter are the [repository subset jobs](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/) which are used to test PRs against subset of the data repository and should remain runnable at anytime without impacting the daily jobs like [the full repository job](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/) /cc @simleo

### DIFF
--- a/ant/java.xml
+++ b/ant/java.xml
@@ -176,7 +176,9 @@ your FindBugs installation's lib directory. E.g.:
       </manifest>
     </jar>
     <artifact:install file="${artifact.dir}/${component.jar}">
-        <pom refid="maven.pom"/>
+      <artifact:localRepository id="local.repository"
+        path="${maven.repo.local}"/>
+      <pom refid="maven.pom"/>
     </artifact:install>
   </target>
 
@@ -189,7 +191,9 @@ your FindBugs installation's lib directory. E.g.:
       </manifest>
     </jar>
     <artifact:install file="${artifact.dir}/${bundle.jar}">
-        <pom refid="maven.pom"/>
+      <pom refid="maven.pom"/>
+      <artifact:localRepository id="local.repository"
+        path="${maven.repo.local}"/>
     </artifact:install>
   </target>
 


rleigh-codelibre:
The testing instructions should probably read

```
$ ant clean jars -Dmaven.repo.local=/tmp/bf_ant_repo
```

Testing with the correct name shows it installing the jars into the correct local repo location.  Looks good to merge.

sbesson:
Thanks for the typo, corrected the description.

simleo:
Wiped out `~/.m2`, then ran the suggested command. All artifacts were correctly created in the custom dir and nothing under `~/.m2`. `ant test -Dmaven.repo.local=/tmp/bf_ant_repo` was successful and again did not create anything under `~/.m2`. Works as advertised.

sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/339/ as an example of application of this PR. Will review the various jobs and propose where to use this.

PR #2647, opened 2016-10-31
update developer docs

hflynn:
See https://trello.com/c/phPRO185/34-developer-documentation-updates

As well as reviewing the staging pages https://www.openmicroscopy.org/site/support/bio-formats5.3-staging/developers/overview.html etc, the tester needs to check that the decoupling of ome-xml means there are no post-import errors when importing the BF source into Eclipse - import 5.3.0-m1 to test.

PR #2650, opened 2016-11-01
ND2HandlerTest - Update package to match location

dgault:
I discovered while testing https://github.com/openmicroscopy/bioformats/pull/2647 that when importing a fresh copy of Bio-Formats into Eclipse that I would see an error due to ND2HandlerTest having a mismatch between its package and location. I have updated the package to match the file location.

To reproduce:
- In Eclipse, do File -> Import -> Existing Maven projects and import Bio-Formats without the PR, verify that an error is shown

To test:
- With the PR, reimport the project as before and verify that no error is present

### DIFF
--- a/components/formats-gpl/test/loci/formats/utests/ND2HandlerTest.java
+++ b/components/formats-gpl/test/loci/formats/utests/ND2HandlerTest.java
@@ -23,7 +23,7 @@
  * #L%
  */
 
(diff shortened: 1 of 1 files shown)


sbesson:
Tested with and without this PR. All happens as described when importing the project into Eclipse. With this PR included, there is no more error at import. Thanks.

PR #2652, opened 2016-11-02
Fix test-suite targets to use TestNG

sbesson:
Addresses https://trello.com/c/jGrPs6c9/16-fix-integration-tests

Following the build changes in https://github.com/openmicroscopy/bioformats/pull/2622, the `test-file-handles` and `test-openbytes-performance` targets of the `test-suite` component ended up broken. Historically these targets were relying on the `target-test-runner` Python script to generate a TestNG XML file from a test class, a series of target files and run the tests. With the removal of the JARs from the source code, reconstructing the CLASSPATH necessary to run the tests cannot be simply achieved in `target-test-runner`.

Instead this PR reduces the `target-test-runnner` scope to only generating the TestNG file and uses the Ant `testng` target to run the tests using the generated XML file. The single-threaded nature of the `FileHandlesTest` class is also captured in the `test-files-handles` target.

This PR should be tested via the https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-performance/ job wich runs both targets against a unified series of target files.

### DIFF
--- a/components/test-suite/build.xml
+++ b/components/test-suite/build.xml
@@ -26,6 +26,7 @@ Type "ant -p" for a list of targets.
   <target name="test-file-handles" depends="compile"
     description="run tests for leaking file handles">
     <exec executable="${basedir}/target-test-runner">
+      <!-- FileHandleTest is single-threaded -->
       <arg value="-t1"/>
       <arg value="-f${testng.filename}"/>
       <arg value="loci.tests.testng.FileHandleTest"/>
--- a/components/test-suite/build.xml
+++ b/components/test-suite/build.xml
@@ -11,6 +11,11 @@ Type "ant -p" for a list of targets.
   <property name="root.dir" location="../.."/>
   <import file="${root.dir}/ant/java.xml"/>
   <property file="build.properties"/>
+  <condition property="testng.threadCount" value="1">
+    <not>
+      <isset property="testng.threadCount"/>
+    </not>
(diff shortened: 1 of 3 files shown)


rleigh-codelibre:
@sbesson Ready to come out of breaking?

sbesson:
@rleigh-codelibre: almost. Talking with @melissalinkert, we downsized the work to be achieved on this front in the context of Bio-Formats 5.3.0. Once the configuration file containing the filesets to be tested has been reviewed so that we preserve the data coverage, I will update the description of this PR and move it to the daily build.

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-performance/24/ for the output of this PR. Removing the breaking comment for inclusion in the daily builds.

rleigh-codelibre:
Given the changes to `target_test_runner`, the `--hudson` option looks redundant,  You could retain the `--hudson` codepaths to generate `testng.xml` and drop the alternative codepaths, since it not longer has the ability to invoke testng itself.

sbesson:
@rleigh-codelibre makes sense. We already did break the scope of the `target-test-runner` script which is now limited to generating the TestNG file. Last commits should address your comments.

rleigh-codelibre:
The latest changes all look good, and the CI job is green.  This looks good to merge.

PR #2653, opened 2016-11-03
Fix test-metadata and move it to test-suite

simleo:
The `test-metadata` target in `components/formats-gpl/build.xml` was broken due to an obsolete TestNG option (`sourcedir`). This PR fixes the target and moves it (along with the test it refers to) to `test-suite`, since it's actually not a unit test and it's not formats-gpl-specific. Rather, it tests metadata levels on a single user-specified file:

```
ant -f components/test-suite/build.xml test-metadata -Dtestng.filename=/path/to/image.ext
```

### DIFF
--- a/components/formats-gpl/build.xml
+++ b/components/formats-gpl/build.xml
@@ -133,7 +133,7 @@ Type "ant -p" for a list of targets.
 
   <target name="test-metadata" depends="compile-tests"
     description="test metadata level support for a single file" if="doTests">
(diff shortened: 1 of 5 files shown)


dgault:
Moving this out of the unit tests seems like the right call to me as it really isn't a unit test.

The test runs as expected with the new ant target and Ive tested with both files that pass the tests as well as those which fail. The resulting test runs produced the correct outputs.

Looks good to merge.

PR #2654, opened 2016-11-04
Test build cleanup

sbesson:
Following the 5.3.0-m1 decoupling, this cleans up a couple of additional files in the top-level:

- setup.cfg is removed as all Python components have been migrated away
- the `test-build` script is reviewed to remove the obsolete `cpp` and `flake8` targets as well as the documentation
- the `.classpath-template` and `.gitignore` files are reviewed to removed decoupled/obsolete components and folders which should be ignored by the top-level `target`, `test-output`.... folders generated by the build system

### DIFF
--- a/tools/test-build
+++ b/tools/test-build
@@ -49,7 +49,7 @@ do
         ant)
             antbuild ;;
         all)
-            clean && maven_findbugs && antbuild;;
+            clean && maven && antbuild;;
         *)
             echo "Invalid argument: \"$arg\"" >&2
             exit 1
--- a/tools/test-build
+++ b/tools/test-build
@@ -17,32 +17,6 @@ maven()
     mvn
 }
 
-cpp()
-{
-    (
-        cd cpp
-        mkdir build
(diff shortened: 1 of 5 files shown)


tools/test-build
```diff
@@ -72,14 +46,10 @@ do
...
             maven ;;
-        cpp)
-            cpp ;;
-        flake8)
-            flake ;;
         ant)
             antbuild ;;
         all)
-            clean && maven_findbugs && sphinx && antbuild;;
+            clean && maven_findbugs && antbuild;;
```
rleigh-codelibre: Where is `maven_findbugs`?  I don't see a shell function by that name.  Does this need fixing too?

rleigh-codelibre: Thanks for fixing this.  This runs successfully.

rleigh-codelibre:
The changes here all look sensible.  The test-build script works correctly, including the `all` option.

This is good to merge once travis/jenkins are green.

PR #2658, opened 2016-11-08
DeltaVision: Improve accuracy of isThisType stream checks

rleigh-codelibre:
- Check that it's not a TIFF
- Check that the file size meets a minimum size

Testing: Check with the images from QA 17411 https://www.openmicroscopy.org/qa2/qa/feedback/17411/ which should now be detected as TIFF, not DeltaVision.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -42,6 +42,7 @@ import loci.formats.FormatTools;
 import loci.formats.MetadataTools;
 import loci.formats.meta.IMinMaxStore;
 import loci.formats.meta.MetadataStore;
+import loci.formats.tiff.TiffParser;
 
 import ome.xml.model.enums.Correction;
 import ome.xml.model.enums.Immersion;
@@ -161,6 +162,11 @@ public class DeltavisionReader extends FormatReader {
   /* @see loci.formats.IFormatReader#isThisType(RandomAccessInputStream) */
   @Override
   public boolean isThisType(RandomAccessInputStream stream) throws IOException {
+    // Extra sanity check; if it's a TIFF, return false immediately, to avoid
+    // false negatives.
+    if(new TiffParser(stream).isValidHeader()) {
+      return false;
+    }
     final int blockLen = 98;
     if (!FormatTools.validStream(stream, blockLen, true)) return false;
     stream.seek(96);
@@ -174,7 +180,7 @@ public class DeltavisionReader extends FormatReader {
     int x = stream.readInt();
     int y = stream.readInt();
     int count = stream.readInt();
-    return x > 0 && y > 0 && count > 0;
+    return x > 0 && y > 0 && count > 0 && (x * y * count > stream.length());
   }
 
   /* @see loci.formats.IFormatReader#getSeriesUsedFiles(boolean) */
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -180,7 +180,8 @@ public class DeltavisionReader extends FormatReader {
     int x = stream.readInt();
     int y = stream.readInt();
     int count = stream.readInt();
-    return x > 0 && y > 0 && count > 0 && (x * y * count > stream.length());
+    return x > 0 && y > 0 && count > 0 &&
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
```diff
@@ -174,7 +180,7 @@ public boolean isThisType(RandomAccessInputStream stream) throws IOException {
     int x = stream.readInt();
     int y = stream.readInt();
     int count = stream.readInt();
-    return x > 0 && y > 0 && count > 0;
+    return x > 0 && y > 0 && count > 0 && (x * y * count > stream.length());
```
melissalinkert: Shouldn't this be ```x * y * count < stream.length()```?  It would probably also be good to cast ```x```, ```y```, and ```count``` to long when multiplying, otherwise integer overflow could cause a false positive.

components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
```diff
@@ -161,6 +162,11 @@ public boolean isThisType(String name, boolean open) {
   /* @see loci.formats.IFormatReader#isThisType(RandomAccessInputStream) */
   @Override
   public boolean isThisType(RandomAccessInputStream stream) throws IOException {
+    // Extra sanity check; if it's a TIFF, return false immediately, to avoid
+    // false negatives.
+    if(new TiffParser(stream).isValidHeader()) {
+      return false;
+    }
```
melissalinkert: I think this will be redundant if the below check is fixed.  ```isValidHeader``` is checking the first four bytes; in the case of a valid header, these would be 0x49492a00 or 0x4d4d002a, which would be read as ```x``` below and should already fail the comparison with ```stream.length()```.

rleigh-codelibre: Unfortunately, this isn't the case.  The length check still passes with TIFF data, including the test data from 17411, so this is needed to be sure it's not TIFF.  The dv magic and length checks are simply too ambiguous on their own to guarantee a correct `false` return, unless there's something more definitive we can check for.

This check will at least cover most TIFF-based formats to prevent misbehaviour, though ideally it wouldn't be needed.

simleo:
Note that, without this PR, the behaviour is different with the two files provided. `_t20270.tif` fails to convert only if the `-stitch` option is used (the file is first detected as a TIFF, then as a DV after it goes through `FileStitcher`), while `_t20271.tif` always fails (immediately detected as DV).

Anyway, **with** this PR both files are correctly identified and converted, with or without `-stitch`.

mtbc:
This looks good to me as at least being an improvement. If I were to criticize the TIFF check it would be more that it reflects the particular sample data we have that triggered this issue rather than our more properly putting the more confident checks from other non-TIFF formats too ahead of the weak DV check here. Real shame we can't make more use of filename too though in this circumstance, for whatever reason.

components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
rleigh-codelibre: It also occurs to me that in reverse, a TIFF header could actually be valid DV data on occasion, so while this content-based check is likely more robust all around with this change (in terms of overall failure rates, by increasing the probability of a correct identification), but still far from perfect.

melissalinkert: I'm still not 100% convinced that this can't be replaced by a further check on ```x * y * count``` later on, which is negative in this case due to long overflow.  Perhaps wrapping each variable in a ```BigInteger``` instead of casting to ```long``` before multiplying would help and allow this to be removed?

sbesson:
@rleigh-codelibre is it worth trying to resolve the changes discussed above for 5.5.2 or should we create a card to fix this in a later release?

rleigh-codelibre:
I'll double-check tomorrow.

rleigh-codelibre:
Refreshing my memory of the discussion, I'm unsure that the suggsted size check addition is sufficiently robust-a valid TIFF could still end up being identified as DV. It all comes down to chance, which makes me uneasy about it.

sbesson:
@rleigh-codelibre can you open a configuration PR for adding QA 17411 to the data repository tests?

sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/573/. Merging as an improvement to a long-running detection issue.

PR #2659, opened 2016-11-14
Options API

simleo:
Adds a generic configuration object, `FormatOptions`, that allows to set and get arbitrary key/value pairs. While the actually stored mapping is `String => String`, several methods are provided to convert to and from commonly used types. `DefaultMetadataOptions` has been changed to extend `FormatOptions` (making it more flexible) and its original methods have been re-implemented accordingly. This also serves as a first usage example for `FormatOptions`.


sbesson:
As discussed in the [https://trello.com/c/eC6Peivz/50-options-api-priorities](card), one of the constraints of the PR is to maintain backwards-compatibility which effectively prevents us from adding new methods to the pubic`MetadataOptions` interface. 

Extending our default options implementation certainly preserves this contract and allows us to expoe the new key/value options mechanism internally. The casting is unfortunate but we will likely have to live with it until the next round of breaking changes. One thing that we could envision to simplify consumption would be to wrap this casting into helpers in the abstract `FormatHandler` class.

The extension proposed in this state could also be achieved by deriving an abstract base class implementing `MetadataOptions` and containing the key/value logic and letting `DefaultMetadataOptions` extend this abstract class e.g.

```
public abstract class AbstractMetadataOptions implements MetadataOptions {
  private Properties props;
...
}

public class DefaultMetadataOptions extends AbstractMetadataOptions {
...
}
```

Is there any reason to use one pattern rather than the other?

simleo:
@sbesson *at the moment* I don't think there would be any significant difference (we would still need the cast). Moving forward, however, I would rather not have the new k/v object be an implementation of `MetadataOptions`, since their relationship would be the opposite of what we should achieve once we are able to make breaking changes. The k/v object is the one that's most general, so anything that provides component-specific options such as the metadata level should be an extension / implementation / user of it, and not the other way around. Indeed, I think the cleanest way to go would be to have metadata level and validate be defined and manipulated directly by `IMetadataConfigurable` & children via alteration of the k/v object, and to eventually (after a period of deprecation) get rid of `MetadataOptions`.

Re: wrapping the cast in a helper in `FormatHandler`, that would just move the problem from the `MetadataOptions` stack to the `IFormatReader` one: in order to use the new helper, you would need to cast any `IFormatReader` you're holding to a `FormatReader` (same goes for writers).

rleigh-codelibre:
Looking at the changes:

- Why is `FormatsOptions` plural rather than the singular `FormatOptions`?  The latter would match `FormatHandler`, `FormatTools`, `FormatReader`, `FormatWriter` etc.  There are no existing plural instances.
- Could we make use of `Properties` defaults with the getters so that they can be handled automatically and without the caller having to have any knowledge of defaults?  Or could we do both?  The `Properties` API has a getter with and without a provided default; could we provide both sets of methods for the types we support?  Could we allow for passing in a Properties object with the defaults at construction time, rather than having it private?  And/or maintain the private default `Properties` alongside the properties and allow it to be updated with `setXXXDefault` methods? i.e.

```
private Properties defaultProps;
private Properties props;

public FormatsOptions() {
  defaultProps = new Properties();
  props = new Properties(defaultProps);
}
```

providing this works.

sbesson:
The rationale for the current design in https://github.com/openmicroscopy/bioformats/pull/2659#issuecomment-263254715 makes sense. I would agree with @rleigh-codelibre in https://github.com/openmicroscopy/bioformats/pull/2659#issuecomment-263278242 that we want to use `FormatOptions` for consistency with the rest of the API.

Re helpers, you are right about the casting to `FormatReader` for all classes implementing `IFormatReader` i.e. typically the `ReaderWrapper` stack. I still wonder if adding helpers might simplify the usage both in our concrete readers/writers and for any third-party readers extending `FormatReader`. This addition does not have to be decided as part of this PR but might be driven by examples.

The extension proposed in https://github.com/openmicroscopy/bioformats/pull/2659#issuecomment-263278242 makes sense a priori but would require extra time to implement and test across all use cases. In the scope of 5.3.0, we decided with @simleo and @joshmoore to limit the scope of the API addition to storing & retrieving key/value based options and ensuring these options are propagated down to all subreaders. Assuming backwards compatibility, we could explore the extension of this API to support default in a minor release.

@joshmoore @dgault @melissalinkert: any other thought on the current proposal?

simleo:
Renamed. We could also call it `Configuration` as they did in Hadoop if anyone thinks it's better.

simleo:
Re: auto-handling defaults: we can start enabling that right now at the `FormatOptions` level, but how to actually populate the defaults falls under the "where to store property names and defaults" item that was moved to the "Later" priority list on the card. Your proposal should be fairly straightforward to implement if we end up deciding to concentrate all property names and defaults in the `FormatOptions` class itself.

simleo:
I've moved `FormatOptions` to `loci.formats` since it applies to both readers and writers (to anything, actually). I think `MetadataOptions` & friends also belong there, but we can't move them now ofc.

simleo:
@rleigh-codelibre actually, assuming we can find a good solution to "where to store names and defaults", there's another problem: what to do when `get*(prop)` is called but the `prop` is not set and there is no default for it (e.g., `Memoizer`'s cache directory). I see two possible solutions:

 * return `null`. Since a primitive type cannot be null, we could change type-specific getters for primitive types to return the corresponding wrappers instead (e.g., `getLong` returns a `Long` object). **This is something we'd have to implement right now, before this PR is merged**.
 * Throw an exception. This also looks fine to me, but It feels more the Python than the Java way.

simleo:
Another thing we should to **before this PR is merged** is get rid of some type-specific {g,s}etters that we think would see no use, e.g., `short`, `char`. This was raised by @joshmoore but IIRC others agreed.

sbesson:
Discussed earlier. Closing in favor of #2679

PR #2663, opened 2016-11-17
Tiling API and Tiff Writer implementation

dgault:
Adds the following new API functions:
public int getTileSizeX()
public int getTileSizeY()
int setTileSizeX(int tileSize) throws FormatException
int setTileSizeY(int tileSize) throws FormatException 

The base implemenattion of these functions will default to full image height and width and return these values when you try to set to any other tiling size.

Any writer that uses the TiffWriter will have its own implementation which will allow any multiple of 16 to be used as a tile size. If a non multiple of 16 is provided then the tileSize will round to the nearest 16 and return this value to the caller.

For the TiffWriter, if a tile size has been set when saveBytes is called then the provided byte array will be written out in tiles of this size.

For testing the following combinations of Tiff images should be written out and verify that the tile sizes are correct using "tiffinfo -s" (should display for example '3072 Tiles, as opposed to '3072 strips') and verifying that the images display correctly:

- Single channel, multi time and z frames
- Single channel
- Multi channel interleaved image
- Multi channel non interleaved image
- Varying pixel formats
- Existing tiled image, verify that it can be resaved with different tile size
- Images with height and width which are not a multiple of 16
- Tiling with X and Y values which will overlap tile boundaries

Unit test contains framework for using mock writers to allow for unit testing of format writers withoutout outputting large filesets. I have included tests for basic TiffWriter functionality as well as the new API functions included for getting and setting tiling sizes. I have removed the incomplete tests for saveBytes which had been in progress, this functionality could be re-added later or covered with test-suite coverage.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -293,6 +293,38 @@ public abstract class FormatWriter extends FormatHandler
     this.sequential = sequential;
   }
 
+  /* @see IFormatWriter#getTileSizeX() */
+  @Override
+  public int getTileSizeX() {
+    int width = metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
+    return width;
+  }
+
+  /* @see IFormatWriter#setTileSizeX(int) */
+  @Override
+  public int setTileSizeX(int tileSize) throws FormatException {
+    int width = metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
+    if (tileSize < 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize > width) throw new FormatException("Tile size must not be > width - " + width);
+    return width;
+  }
+
+  /* @see IFormatWriter#getTileSizeY() */
+  @Override
+  public int getTileSizeY() {
+    int height = metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
+    return height;
+  }
+
+  /* @see IFormatWriter#setTileSizeY(int) */
+  @Override
+  public int setTileSizeY(int tileSize) throws FormatException {
+    int height = metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
+    if (tileSize < 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize > height) throw new FormatException("Tile size must not be > height - " + height);
+    return height;
+  }
+
   // -- IFormatHandler API methods --
 
   /**
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -343,7 +343,7 @@ public abstract class FormatWriter extends FormatHandler
     if (out != null) {
       out.close();
     }
-    out = new RandomAccessOutputStream(currentId);
+    out = createOutputStream();
 
     MetadataRetrieve r = getMetadataRetrieve();
     initialized = new boolean[r.getImageCount()][];
@@ -485,6 +485,10 @@ public abstract class FormatWriter extends FormatHandler
     int c = r.getPixelsSizeC(series).getValue().intValue();
     c /= r.getChannelSamplesPerPixel(series, 0).getValue().intValue();
     return z * c * t;
+  }
+
+  protected RandomAccessOutputStream createOutputStream() throws IOException {
+    return new RandomAccessOutputStream(currentId);
   }
 
 }
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -296,32 +296,30 @@ public abstract class FormatWriter extends FormatHandler
   /* @see IFormatWriter#getTileSizeX() */
   @Override
   public int getTileSizeX() {
-    int width = metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
-    return width;
+    return metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
   }
 
   /* @see IFormatWriter#setTileSizeX(int) */
   @Override
   public int setTileSizeX(int tileSize) throws FormatException {
     int width = metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
-    if (tileSize < 0) throw new FormatException("Tile size must be > 0.");
-    if (tileSize > width) throw new FormatException("Tile size must not be > width - " + width);
+    if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize > width) throw new FormatException("Tile width must be <= image width (" + width + ")");
     return width;
   }
 
   /* @see IFormatWriter#getTileSizeY() */
   @Override
   public int getTileSizeY() {
-    int height = metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
-    return height;
+    return metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
   }
 
   /* @see IFormatWriter#setTileSizeY(int) */
   @Override
   public int setTileSizeY(int tileSize) throws FormatException {
     int height = metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
-    if (tileSize < 0) throw new FormatException("Tile size must be > 0.");
-    if (tileSize > height) throw new FormatException("Tile size must not be > height - " + height);
+    if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize > height) throw new FormatException("Tile height must be <= image height (" + height + ")");
     return height;
   }
 
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -295,32 +295,38 @@ public abstract class FormatWriter extends FormatHandler
 
   /* @see IFormatWriter#getTileSizeX() */
   @Override
-  public int getTileSizeX() {
-    return metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
+  public int getTileSizeX() throws FormatException {
+    PositiveInteger width = metadataRetrieve.getPixelsSizeX(getSeries());
+    if (width == null) throw new FormatException("Pixels Size X must not be null when attempting to get tile size.");
+    return width.getValue();
   }
 
   /* @see IFormatWriter#setTileSizeX(int) */
   @Override
   public int setTileSizeX(int tileSize) throws FormatException {
-    int width = metadataRetrieve.getPixelsSizeX(getSeries()).getValue();
+    PositiveInteger width = metadataRetrieve.getPixelsSizeX(getSeries());
+    if (width == null) throw new FormatException("Pixels Size X must not be null when attempting to set tile size.");
     if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
-    if (tileSize > width) throw new FormatException("Tile width must be <= image width (" + width + ")");
-    return width;
+    if (tileSize > width.getValue()) throw new FormatException("Tile width must be <= image width (" + width.getValue() + ")");
+    return width.getValue();
   }
 
   /* @see IFormatWriter#getTileSizeY() */
   @Override
-  public int getTileSizeY() {
-    return metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
+  public int getTileSizeY() throws FormatException {
+    PositiveInteger height = metadataRetrieve.getPixelsSizeY(getSeries());
+    if (height == null) throw new FormatException("Pixels Size Y must not be null when attempting to get tile size.");
+    return height.getValue();
   }
 
   /* @see IFormatWriter#setTileSizeY(int) */
   @Override
   public int setTileSizeY(int tileSize) throws FormatException {
-    int height = metadataRetrieve.getPixelsSizeY(getSeries()).getValue();
+    PositiveInteger height = metadataRetrieve.getPixelsSizeY(getSeries());
+    if (height == null) throw new FormatException("Pixels Size Y must not be null when attempting to set tile size.");
     if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
-    if (tileSize > height) throw new FormatException("Tile height must be <= image height (" + height + ")");
-    return height;
+    if (tileSize > height.getValue()) throw new FormatException("Tile height must be <= image height (" + height.getValue() + ")");
+    return height.getValue();
   }
 
   // -- IFormatHandler API methods --
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -307,7 +307,6 @@ public abstract class FormatWriter extends FormatHandler
     PositiveInteger width = metadataRetrieve.getPixelsSizeX(getSeries());
     if (width == null) throw new FormatException("Pixels Size X must not be null when attempting to set tile size.");
     if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
-    if (tileSize > width.getValue()) throw new FormatException("Tile width must be <= image width (" + width.getValue() + ")");
     return width.getValue();
   }
 
@@ -325,7 +324,6 @@ public abstract class FormatWriter extends FormatHandler
     PositiveInteger height = metadataRetrieve.getPixelsSizeY(getSeries());
     if (height == null) throw new FormatException("Pixels Size Y must not be null when attempting to set tile size.");
     if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
-    if (tileSize > height.getValue()) throw new FormatException("Tile height must be <= image height (" + height.getValue() + ")");
     return height.getValue();
   }
 
--- a/components/formats-api/src/loci/formats/IFormatWriter.java
+++ b/components/formats-api/src/loci/formats/IFormatWriter.java
@@ -212,7 +212,8 @@ public interface IFormatWriter extends IFormatHandler {
   /**
    * Will attempt to set the tile width to the desired value and return the actual value which will be used
    * @param tileSize The tile width you wish to use
-   * @return The tile width which will actually be used, this may differ from the value attempted to set
+   * @return The tile width which will actually be used, this may differ from the value requested.
+   *         If the requested value is not supported the writer will return and use the closest appropriate value.
    * @throws FormatException Tile size must be greater than 0 and less than the image width
    */
   int setTileSizeX(int tileSize) throws FormatException;
@@ -228,7 +229,8 @@ public interface IFormatWriter extends IFormatHandler {
   /**
    * Will attempt to set the tile height to the desired value and return the actual value which will be used
    * @param tileSize The tile height you wish to use
-   * @return The tile height which will actually be used, this may differ from the value attempted to set
+   * @return The tile height which will actually be used, this may differ from the value requested.
+   *         If the requested value is not supported the writer will return and use the closest appropriate value.
    * @throws FormatException Tile size must be greater than 0 and less than the image height
    */
   int setTileSizeY(int tileSize) throws FormatException;
--- a/components/formats-api/src/loci/formats/IFormatWriter.java
+++ b/components/formats-api/src/loci/formats/IFormatWriter.java
@@ -205,14 +205,15 @@ public interface IFormatWriter extends IFormatHandler {
    * Retrieves the current tile width
    * Defaults to full image width if not supported
    * @return The current tile width being used
+   * @throws FormatException Image metadata including Pixels Size X must be set prior to calling getTileSizeX()
    */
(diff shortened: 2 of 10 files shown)


sbesson:
Are the new unit tests picked up by default by the Maven/Ant build or do they need to be registered in `components/formats-bsd/test/loci/formats/utests/testng.xml` ?

simleo:
I managed to trigger a `NullPointerException` by setting the tile size before  calling`setId`. Do we need a call to `FormatTools.assertId` in `FormatWriter.setTileSizeX` ?

dgault:
Good catch on the NullPointerException. It should be ok to call the tiling API before setId but it does require metadata to have been set. Its worth noting that there are actually a number of functions throughout our writers which will throw the same NullPointerException if called prior to setting metadata.

For the tiling functions I have updated them to throw an exception with at least a better description. I have also reworked the unit tests to provide testing of each of the possible formatException sceanrios for the tiling functions.

simleo:
> Its worth noting that there are actually a number of functions throughout our writers which will throw the same NullPointerException if called prior to setting metadata

I suspected as much. This has to be addressed in some way

components/formats-api/src/loci/formats/WriterWrapper.java
```diff
@@ -321,6 +325,30 @@ public String getCompression() {
   public void setWriteSequentially(boolean sequential) {
     writer.setWriteSequentially(sequential);
   }
+  
+  /* @see IFormatWriter#getTileSizeX() */
+  @Override
+  public int getTileSizeX() {
+    return writer.getTileSizeX();
```
sbesson: From https://travis-ci.org/openmicroscopy/bioformats/jobs/177082002

```
[INFO] -------------------------------------------------------------

[ERROR] /home/travis/build/openmicroscopy/bioformats/components/formats-api/src/loci/formats/WriterWrapper.java:[332,31] unreported exception loci.formats.FormatException; must be caught or declared to be thrown

[ERROR] /home/travis/build/openmicroscopy/bioformats/components/formats-api/src/loci/formats/WriterWrapper.java:[344,31] unreported exception loci.formats.FormatException; must be caught or declared to be thrown
```

components/formats-api/src/loci/formats/WriterWrapper.java
```diff
@@ -321,6 +325,30 @@ public String getCompression() {
...
+  /* @see IFormatWriter#setTileSizeX(int) */
+  @Override
+  public int setTileSizeX(int tileSize) throws FormatException {
+    return writer.setTileSizeX(tileSize);
+  }
+
+  /* @see IFormatWriter#getTileSizeY() */
+  @Override
+  public int getTileSizeY() {
+    return writer.getTileSizeY();
```
sbesson: Same comment as above

components/formats-api/src/loci/formats/FormatWriter.java
```diff
@@ -293,6 +293,42 @@ public void setWriteSequentially(boolean sequential) {
...
+    return width.getValue();
+  }
+
+  /* @see IFormatWriter#setTileSizeX(int) */
+  @Override
+  public int setTileSizeX(int tileSize) throws FormatException {
+    PositiveInteger width = metadataRetrieve.getPixelsSizeX(getSeries());
+    if (width == null) throw new FormatException("Pixels Size X must not be null when attempting to set tile size.");
+    if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize > width.getValue()) throw new FormatException("Tile width must be <= image width (" + width.getValue() + ")");
```
mtbc: Why? Surely we already have to deal with the last column of tiles being wider than what remains of an image?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -214,19 +226,53 @@ public void saveBytes(int no, byte[] buf, IFD ifd, int x, int y, int w, int h)
...
+      int nYTiles = h / tileSizeY;
+   
+      for (int yTileIndex=0; yTileIndex<nYTiles; yTileIndex++) {
+        for (int xTileIndex=0; xTileIndex<nXTiles; xTileIndex++) {
+          byte [] tileBuf = getTile(buf, xTileIndex, yTileIndex, x, y, w, h);
+          int tileX = x + (xTileIndex * tileSizeX);
+          int tileY = y + (yTileIndex * tileSizeY);
+         
+          // This operation is synchronized
+          synchronized (this) {
```
mtbc: What are the expectations regarding multi-threaded calling?

components/formats-api/src/loci/formats/FormatWriter.java
rleigh-codelibre: Also note that TIFF allows the tile size to be greater than the image size.  Overlapping the image bounds is fine.

rleigh-codelibre: Also, given the above, the constraint that PixelsSizeX must be set before the tile size could be lifted.

components/formats-api/src/loci/formats/IFormatWriter.java
```diff
@@ -201,4 +201,36 @@ void savePlane(int no, Object plane, Region tile)
...
+   * Defaults to full image width if not supported
+   * @return The current tile width being used
+   * @throws FormatException Image metadata including Pixels Size X must be set prior to calling getTileSizeX()
+   */
+  int getTileSizeX() throws FormatException;
+
+  /**
+   * Will attempt to set the tile width to the desired value and return the actual value which will be used
+   * @param tileSize The tile width you wish to use
+   * @return The tile width which will actually be used, this may differ from the value attempted to set
```
rleigh-codelibre: "attempted to set" could be "requested", and maybe also mention the reasons why the returned value might differ e.g. if the writer doesn't support it.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -82,6 +82,9 @@
    */
   private static final long BIG_TIFF_CUTOFF = (long) 1024 * 1024 * 3990;
 
+  /** TIFF tiles must be of a height and width divisible by 16. */
```
rleigh-codelibre: Why?  TIFF supports arbitrary tile sizes.  Is this a limitation in our TIFF writer?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -214,19 +226,53 @@ public void saveBytes(int no, byte[] buf, IFD ifd, int x, int y, int w, int h)
...
-        index = prepareToWriteImage(no, buf, ifd, x, y, w, h);
-        if (index == -1) {
-          return;
+    int imageWidth = retrieve.getPixelsSizeX(series).getValue().intValue();
+    int imageHeight = retrieve.getPixelsSizeY(series).getValue().intValue();
+    if (tileSizeX < imageWidth || tileSizeY < imageHeight) {
+      ifd.put(new Integer(IFD.TILE_WIDTH), new Long(tileSizeX));
+      ifd.put(new Integer(IFD.TILE_LENGTH), new Long(tileSizeY));
+    }
+    if (tileSizeX < w || tileSizeY < h) {
```
rleigh-codelibre: I'm unsure of the logic used to select "tile mode" here.  Maybe the tile size should be set to zero when tiling isn't in use.  This would be necessary to allow tile sizes greater than the image size?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -214,19 +226,53 @@ public void saveBytes(int no, byte[] buf, IFD ifd, int x, int y, int w, int h)
...
+      int nXTiles = w / tileSizeX;
+      int nYTiles = h / tileSizeY;
+   
+      for (int yTileIndex=0; yTileIndex<nYTiles; yTileIndex++) {
+        for (int xTileIndex=0; xTileIndex<nXTiles; xTileIndex++) {
+          byte [] tileBuf = getTile(buf, xTileIndex, yTileIndex, x, y, w, h);
+          int tileX = x + (xTileIndex * tileSizeX);
```
rleigh-codelibre: Are these `tileX` and `tileY` calculations correct?  What happens when x and y are not aligned to tile boundaries?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -499,4 +545,74 @@ else if (retrieve.getPixelsBinDataCount(series) == 0) {
...
+    if (tileSizeX == 0) {
+      tileSizeX = super.getTileSizeX();
+    }
+    return tileSizeX;
+  }
+
+  @Override
+  public int setTileSizeX(int tileSize) throws FormatException {
+    tileSizeX = super.setTileSizeX(tileSize);
+    tileSizeX = Math.round((float)tileSize/TILE_GRANULARITY) * TILE_GRANULARITY;
```
rleigh-codelibre: Is floating point needed here?  What's wrong with simple division and addition?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
dgault: This was based on the Tiff tag refrence for TileHeight and TileWidth - http://www.awaresystems.be/imaging/tiff/tifftags/tilewidth.html

mtbc: I think TIFF6 does indeed have the multiple-of-16 restriction.

rleigh-codelibre:
Regarding unit tests, it would be useful to test various tiling combinations to ensure the calculations in getTile are correct.  If you look at https://github.com/ome/ome-files-cpp/blob/master/test/ome-files/data/CMakeLists.txt we test all combinations of:

- Chunky and planar sample formats
- Image size x (32 43 64)
- Image size y (32 37 64)
- Tile size x (16 32 48 64)
- Tile size y (16 32 48 64)
- Strip sizes (01 02 05 14 32 60 64) [when not using tiles]

And then compare the output image pixels byte for byte with the input to ensure we correctly wrote the pixel data for each sample format/image size/tile size combination.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
rleigh-codelibre: Yes, you're right.  Please ignore this, I was thinking of something else which doesn't have the restriction.

rleigh-codelibre:
Regarding calling the tiling functions before setId, that's maybe OK for writing but not for reading.  Since the tiling parameters could be changed per plane (in theory) or per image (in practice), after would be fine, so long as they are only used by saveBytes.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
dgault: The floating point was used to round to the nearest value, if we were only every to round upwards (which I would be fine with) then it could be removed.

components/formats-api/src/loci/formats/ImageWriter.java
```diff
@@ -187,6 +187,26 @@ public IFormatWriter getWriter(Class<? extends IFormatWriter> c) {
     return w;
   }
   
+  @Override
+  public int getTileSizeX() {
+    return getWriter().getTileSizeX();
```
sbesson:
```
[ERROR] /home/travis/build/openmicroscopy/bioformats/components/formats-api/src/loci/formats/ImageWriter.java:[192,36] unreported exception loci.formats.FormatException; must be caught or declared to be thrown

[ERROR] /home/travis/build/openmicroscopy/bioformats/components/formats-api/src/loci/formats/ImageWriter.java:[202,36] unreported exception loci.formats.FormatException; must be caught or declared to be thrown
```

dgault:
Just to summarise based on the feedback so far.

Latest updates address some of the major points raised :
- Tile Width can now be greater than image width (in this scenario we are only writing out the partial tile rather than padding the tile)
- Using x and y parameters which overlap tile boundaries is supported

Regarding the use of width and height as a check, this is still in place as width and height are currently the default values. I will look into alternatives for this as well as some potentially redundant parameter passing.

Open questions:
- Currently if an invalid tile size is set we round to the nearest suitable value. This does have the issue that when setting a value less than 8 it will round down to 0 which is not desirable, should we have a special case to round up in this scenario or switch to rounding up by default?
- I am looking into adding multi-threaded testing at the moment but if there are any specific concerns on areas to focus on it would be good to know

mtbc:
Multi-threading: it'd be fantastic if the actually-probably-slow parts of the process do get multi-threaded (e.g., tile compression if I can call `openBytes` multi-threaded?) but I'd certainly understand if within this PR concurrency testing is limited more to avoiding errors / bad output rather than truly increasing performance.

melissalinkert:
Minor comment: could the use of ```java.awt.Rectangle``` please be replaced with ```loci.common.Region```?  Historically using AWT classes has caused some problems for users in a headless environment - ```Region``` should provide the same API and eliminate any potential worries there.

dgault:
Yeah Region provides the same functionality so its a better choice all round.

rleigh-codelibre:
I have tested the latest PR with this shell script:

```
#!/bin/sh

set -e

for image in *.png; do
  for planarconfig in chunky planar; do
    if [ "$planarconfig" = "chunky" ]; then
      interlace=plane
    else
      interlace=none
    fi
    tiffsource="${image%.png}-${planarconfig}.tiff"
    echo $image $tiffsource

    gm convert "$image" -interlace "$interlace" "$tiffsource"

    for tifftype in "big" "std"; do
      for compression in "none" "LZW"; do
        for tilex in 0 32 43 64; do
          for tiley in 0 32 43 64; do
            tiffdest="bfconvert-${tilex}-${tiley}-${tifftype}-${compression}-${tiffsource}"
            if [ "$tilex" != "0" -a "$tiley" != "0" ]; then
              tileopts="-tilex $tilex -tiley $tiley"
            else
              tileopts=""
            fi
            if [ "$tifftype" = "big" ]; then
              typeopts="-bigtiff"
            else
              typeopts=""
            fi
            if [ "$compression" = "LZW" ]; then
              compressopts="-compression LZW"
            else
              compressopts=""
            fi
            ../tools/bfconvert "$tiffsource" $typeopts $compressopts $tileopts -overwrite "$tiffdest"

            if [ "$tilex" != "0" -a "$tiley" != "0" ]; then
              if [ "$(tiffinfo "$tiffdest" | grep "Tile Width" | sed -e 's;.*Tile Width: \([0-9][0-9*]\).*;\1;')" != "$tilex" ]; then
                echo "FAIL: $tiffdest has incorrect tile width"
              fi
              if [ "$(tiffinfo "$tiffdest" | grep "Tile Length" | sed -e 's;.*Tile Length: \([0-9][0-9*]\).*;\1;')" != "$tiley" ]; then
                echo "FAIL: $tiffdest has incorrect tile length"
              fi
            else
              if tiffinfo "$tiffdest" | grep -q "Tile"; then
                echo "FAIL: $tiffdest has tile size when tiling should be disabled"
              fi
            fi

          done
        done
      done
    done
  done  
done
```

and these images:

test-g1
test-rgb1

This checks with different tiling parameters and image properties all work OK, but the script isn't perfect and needs a few further tweaks.

- The last IFD is incorrect:

```
% tiffinfo test/bfconvert-0-0-big-LZW-test-g1-chunky.tiff >/dev/null
TIFFFetchDirectory: Can not read TIFF directory count.
TIFFReadDirectory: Failed to read directory at offset 70775.
```

It's writing the last IFD with a non-zero offset, which is incorrect.

- It seems to be getting all the tiling correct for all the tile sizes, compression and bigtiff options tested above.  I can view them all OK.
- It needs testing for all the different pixel types, including BIT, to make sure the tiling calculations work for them.

I think that the different pixel types, plus all the parameters I tested in the script, should be in a unit test rather than integration tests.  We're testing the tiling logic works by writing out a sample plane, and then reading it back in and comparing the content.  The pixel data is just a greyscale or RGB gradient; trivial to generate for all the different pixel types, and the C++ tiling code tests in this way.  There are so many edge cases and variants, that I don't feel happy leaving this to integration tests which don't have full coverage--this code is critical and needs to be shown to be correct under all conditions.

dgault:
The new unit test runs through a combination of the following properties:
- Tiling Size
- little endian
- interleaved
- channel count
- series count
- compression
- pixel type

It will then write out and read back images, checking that the tile size has been set correctly and that the planes are the same as those written were possible.

The unit will be skipped by default due to long time it can take to run.

The following system properties have been added to configure its running:
testng.runWriterTilingTests - boolean, must be set to true to run tiling unit test otherwise test is skipped
testng.runSubsetTilingTests - integer, if the tests are being run then an int value representing the percentage of scenarios to be run can be used 

Still to be fixed - Incorrect offset in last IFD

sbesson:
Leaving @rleigh-codelibre to comment on the unit test logic/coverage, a few questions on how to operate this:
- could we merge the 2 configurations i.e. `testng.runSubsetTilingTests` being `0` by default which means effectively turning this off
- as shown by https://travis-ci.org/openmicroscopy/bioformats/builds/180332335, there is a potential for causing timeouts at the Travis level. Should we nevertheless consider having a very minimal subset run at this level? OR should these tests be delegated to our Jenkins builds exclusively?
- at the Jenkins level, which build will be running these tests and in which proportion? Adding it to the default build would work but might increase the time for generating artifacts. If that proved to be too costly, an alternative would be to extend https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-performance/ for running both long-running unit tests and integration/performance tests

dgault:
I have merged the two system properties as suggested
`testng.runWriterTilingTests` - 0 - tests skipped, also the default behaviour
`testng.runWriterTilingTests` - int value between 0 and 100 - The percentage of tests randomly selected 
`testng.runWriterTilingTests` - 100 - full test data used

rleigh-codelibre:
I'll carry on reviewing the latest commits tomorrow

sbesson:
With the last commit and `testng.runWriterTilingTests` set to 100, the [daily merge build](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/461/) ran 3034 FOrmats BSD unit tests with the build time increasing from 15min to 30 min. From my perspective assuming the coverage  is sufficient, the time increase in our CI builds (disabled by default for developers and in Travis) is an acceptable compromise. Would we need integration tests in addition or are these unit tests sufficient?

dgault:
These unit tests are pretty comprehensive in that they go through the full process of writing out and reading back in images as opposed to simply testing the saveBytes function independently. In that regard I don't know how much integration tests would add other than testing on 'real' data, in which case we would be reading in, writing out and then reading in again.

components/formats-bsd/test/loci/formats/utests/out/TiffWriterTest.java
```diff
@@ -89,6 +103,55 @@
...
+    int[] tileSize = {1, 32, 43, 64};
+    boolean[] booleanValue = {true, false};
+    int[] channelCount = {1, 3};
+    int[] seriesCount = {1, 5};
+    String[] compression = {COMPRESSION_UNCOMPRESSED, COMPRESSION_LZW, COMPRESSION_J2K, COMPRESSION_J2K_LOSSY, COMPRESSION_JPEG};
+    int compressionPixelTypeSize = (2 * pixelTypesOther.length) + pixelTypesOther.length - 1 + pixelTypesJ2K.length + 2;
+    int paramSize = tileSize.length * compressionPixelTypeSize * 4 * channelCount.length * seriesCount.length;
+    Object[][] data = new Object[paramSize][];
+    int index = 0;
+    for (int i = 0; i < tileSize.length; i++) {
```
rleigh-codelibre: You could probably use range-based for all these loops, to make it a bit simpler?

rleigh-codelibre:
Looking over all the last changes, it all looks good.  Just a few minor questions.  I'll look over in more detail on Monday as well.

sbesson:
Thanks @rleigh-codelibre. Merging this as we have most of the groundwork incl. API extension to support tiled writing, implementation for the TIFF writer and corresponding unit tests covering a wide range of use cases. Additional issues can be reported and addressed in separate PRs.

components/formats-bsd/build.xml
```diff
@@ -27,6 +27,8 @@ Type "ant -p" for a list of targets.
         <pathelement location="${test-classes.dir}"/>
         <pathelement location="${classes.dir}"/>
       </classpath>
+      <sysproperty key="testng.runWriterTilingTests" value="${testng.runWriterTilingTests}"/>
+    	<sysproperty key="testng.runWriterTilingTests" value="${testng.runWriterTilingTests}"/>
```
mtbc: some indent issue here

sbesson: Should have been fixed in 47a759d1959e4029fc77d8a2adc55b6f41999c24

components/formats-api/src/loci/formats/FormatWriter.java
mtbc: I'm indeed now running into this. Naively trying to set tile size in advance in the hope of not needing my tile-size-setting per-plane IFDs any more fails because I apparently also now need to set this pixels size thing somehow.

PR #2664, opened 2016-11-17
Bundle cleanup

melissalinkert:
See https://trello.com/c/i1XLwx3M/29-review-bundles

This adds version information to the manifest of Maven-built bundles and excludes ImageJ (```net.imagej:ij```) from bundles built with both Maven and Ant.

To test, build ```bioformats_package.jar``` and ```loci_tools.jar``` with both Ant (```ant clean jars tools```) and Maven (```mvn clean package install```).  Check the ```META-INF/MANIFEST.MF``` in all jars, and verify that all have matching information, including version number, git commit, and build date.  Also check that ImageJ is no longer included in any of the jars - each should be ~15MB smaller.

### DIFF
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -57,6 +57,7 @@
       <groupId>net.imagej</groupId>
       <artifactId>ij</artifactId>
       <version>${imagej1.version}</version>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -57,7 +57,6 @@
       <groupId>net.imagej</groupId>
       <artifactId>ij</artifactId>
       <version>${imagej1.version}</version>
-      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/components/bundles/bioformats_package/assembly.xml
+++ b/components/bundles/bioformats_package/assembly.xml
@@ -11,6 +11,7 @@
     <dependencySet>
       <excludes>
         <exclude>gov.nih.imagej:imagej</exclude>
+        <exclude>net.imagej:ij</exclude>
         <exclude>org.springframework:spring*</exclude>
         <exclude>aopalliance:aopalliance</exclude>
         <exclude>org.aspectj:aspectj*</exclude>
--- a/components/bundles/bioformats_package/pom.xml
+++ b/components/bundles/bioformats_package/pom.xml
@@ -71,7 +71,14 @@
           <archive>
             <manifest>
               <mainClass>loci.formats.gui.ImageViewer</mainClass>
+              <addClasspath>true</addClasspath>
+              <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
             </manifest>
+            <manifestEntries>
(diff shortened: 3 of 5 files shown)


sbesson:
Tested the bundle metadata by running the Maven build command. All build information is now populated as shown by `unzip -q -c ~/.m2/repository/ome/bioformats_package/5.3.0-SNAPSHOT/bioformats_package-5.3.0-SNAPSHOT.jar META-INF/MANIFEST.MF` or by testing the generated bundle in ImageJ: 

screen shot 2016-11-18 at 13 58 51

The differences between the Ant generated and the Maven generated bundles remain minimal and limited to META-INF content which is an improvement over 5.2.x. I'll review the content of the bundles later.

sbesson:
Using the latest version of the PR, there is no more Javadoc warning reported by the [daily CI job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/448/) during the build step.

Using `jar tf | sort -n` to review the content of the various Bio-Formats bundles, I compared the output of

- `bioformat_package` 5.3.0-SNAPSHOT generated by [Ant](https://github.com/openmicroscopy/bioformats/files/606932/bioformats_package_5.3.0_ant.txt) or [Maven](https://github.com/openmicroscopy/bioformats/files/606933/bioformats_package_5.3.0_maven.txt)
- `bioformats_package` 5.2.4 generated by [Ant](https://github.com/openmicroscopy/bioformats/files/606939/bioformats_package_5.2.4_ant.txt) or [Maven](https://github.com/openmicroscopy/bioformats/files/606937/bioformats_package_5.2.4_maven.txt) 
- `loci_tools` 5.3.0-SNAPSHOT generated by [Ant](https://github.com/openmicroscopy/bioformats/files/606942/loci_tools_5.3.0_ant.txt)

Looking at the diff between these text files, a couple of comments:
- the difference between the Ant and Maven generated bundles is now limited to meta-information as reported above :+1:
- the difference between `loci_tools` and `bioformats_packages` are the `log4j` and `slf4j-log4j12` packages :+1:
- the difference between the Maven generated `bioformats_package` 5.2.4 and `bioformats_package` 5.3.0-SNAPSHOT is that the `ij` package is no longer included in the bundle and the `jxrlib` related classes are included :+1:
- the difference between the Ant generated `bioformats_package` 5.2.4 and `bioformats_package` 5.3.0-SNAPSHOT is larger than the previous ones as many packages where missing in 5.2.4. In addition to the `jxrlib` classes, 5.2.4 was missing the `jhdf5` dependency but also `xml-apis` (`serializer` dependency) and `commons-logging` (`ome-poi` dependency).

For the last point, as we had a discrepancy in the 5.2.4 bundle, we could decide either way. I would assume the state of this PR (including all dependencies except `ij`) is correct. Is there any specific testing that should be conducted with the Ant-generated build in order to check this will have no impact on consumers?

melissalinkert:
It might not hurt to do a quick check with ImageJ and Matlab using the Ant-generated ```bioformats_package.jar```.  Opening one or two files (maybe JPEG-XR .czi and a CellH5 since those are the ones with affected dependencies?) and checking that logging is reasonable, OME-XML is generated and viewable, images can be shown, etc. is probably sufficient.

sbesson:
Tested the `bioformats_package.jar` using both ImageJ and the MATLB toolbox against a JPEG-XR compressed CZI and a CellH5 file. Both opened as expected with the right amount of logging when increasing verbosity, metadata is populated as expected and pixel data is correctly read. Thanks. Merging.

PR #2666, opened 2016-11-18
Test suite: don't exclude files that cause setId failures

simleo:
In the "remove duplicates" phase, `FormatReaderTestFactory` ignores `setId` failures, effectively preventing files that would fail the test right at `setId` from being tested at all.

With this PR:

1. If `setId` fails in the remove duplicate phase, the method breaks immediately (we have to catch, print error and rethrow because testng swallows exceptions thrown by factory methods)
2. If anything else goes wrong in the remove duplicate phase the file is kept in the list, so that it gets tested later

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -219,13 +219,16 @@ public class FormatReaderTestFactory {
     }
     Set<String> minimalFiles = new LinkedHashSet<String>();
     FileStitcher reader = new FileStitcher();
+    Set<String> failingIds = new LinkedHashSet<String>();
     while (!fileSet.isEmpty()) {
       String file = fileSet.iterator().next();
       try {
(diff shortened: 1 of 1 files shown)


components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
```diff
@@ -223,6 +223,11 @@ public String getProperty(String key) {
       String file = fileSet.iterator().next();
       try {
         reader.setId(file);
+      } catch (Exception e) {
+        LOGGER.error("setId({}) failed", file, e);
```
sbesson: Tested in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/354/console /cc @melissalinkert 

As discussed, assuming this does not introduce more instability, introducing a fail-fast mechanism for files failing `setId()` might make sense and save unnecessary CPU.
One RFE to would be to catch all files failing `setId`, captures them in a list and then throw a `RuntimeException`.

PR #2667, opened 2016-11-22
Update POM version numbers to 5.3.0-m2

snoopycrimecop:
Repository: openmicroscopy/bioformats
Already up-to-date.

Generated by build [BIOFORMATS-DEV-latest-version-bump#19](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-latest-version-bump/19/).

----
--no-rebase

### DIFF
--- a/components/autogen/pom.xml
+++ b/components/autogen/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>ome</groupId>
     <artifactId>pom-bio-formats</artifactId>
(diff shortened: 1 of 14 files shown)


sbesson:
Travis is red because the `bump_maven_version.py` script tried to bump `jxrlib-all` as well. This also means the `groupId` substitution logic of the script might be invalid depending on the use case. I'll open a PR manually instead and we can discuss whether this script should be redesigned or simply does not apply anymore.

PR #2669, opened 2016-11-22
m2 version history

hflynn:
See https://trello.com/c/KK3DJFbE/20-version-history-prs

I've maybe summarised too much and missed details that are useful for devs, happy to expand if so.

Staged at http://www.openmicroscopy.org/site/support/bio-formats5.3-staging/about/whats-new.html

### DIFF
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -58,7 +58,7 @@ Component architecture changes/decoupling:
 New features/API:
 
 * added support for JPEG-XR compressed CZI data (funded by a
-  `partnership between Glencoe Software and ZEISS <http://glencoesoftware.com/2016-08-30-glencoe-software-zeiss-partner-open-source-file-reader-whole-slide.html>`_), adding ``ome:jxrlib`` as a new dependency
+  `partnership between Glencoe Software and ZEISS <http://glencoesoftware.com/2016-08-30-glencoe-software-zeiss-partner-open-source-file-reader-whole-slide.html>`_), adding 'ome:jxrlib' as a new dependency
   of Bio-Formats
 
 5.2.4 (2016 October 18)
--- a/docs/sphinx/about/whats-new.txt
+++ b/docs/sphinx/about/whats-new.txt
@@ -1,6 +1,25 @@
 Version history
 ===============
 
+5.3.0-m2 (2016 November 23)
+---------------------------
+
+Fixes and clean up:
+
+* clean up of bundles, packages and build targets to reflect decoupling work
+* POM repositories clean up
+* updated developer documentation on building Bio-Formats
+* updated testing packages
+* updated build versioning from Maven
+
+Further decoupling:
+
+* removed stubs and metakit and replaced them with jars from Maven Central
+
+File format support:
+
+* added support for JPEG-XR compressed CZI data
+
 5.3.0-m1 (2016 October 27)
 --------------------------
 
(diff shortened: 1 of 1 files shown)


docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,25 @@
 Version history
 ===============
 
+5.3.0-m2 (2016 November 23)
```
sbesson: One question here: are we listing each milestone separately or integrating them together? I would favor the second option as developers will typically upgrade to 5.3.0 and want to list all changes.

docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,25 @@
...
+---------------------------
+
+Fixes and clean up:
+
+* clean up of bundles, packages and build targets to reflect decoupling work
+* POM repositories clean up
+* updated developer documentation on building Bio-Formats
+* updated testing packages
+* updated build versioning from Maven
```
sbesson: I think we could expand on this to explain that 1- we unified the versioning strategy, 2- we reviewed the meta information stored in the manifest of each JAR and 3- we are now making introspecting this meta-information in the `FormatTools` API to retrieve the version and revision number. Does that correctly represent your contribution @rleigh-codelibre ?

docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,25 @@
...
+
+Further decoupling:
+
+* removed stubs and metakit and replaced them with jars from Maven Central
```
sbesson: As we did previously, we might want to add extra information about the repository and `groupId/artifactId`

docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,25 @@
...
+
+File format support:
+
+* added support for JPEG-XR compressed CZI data
```
sbesson: I would minimally mention `ome:jxrlib` as a new dependency of Bio-Formats. Is it worth linking to the blog post/press announcement? @melissalinkert @chris-allan : anything in particular to mention on this page or should we generically review as part of the documentation of the CZI work?

docs/sphinx/about/whats-new.txt
hflynn: I assumed we'd amalgamate the list for the 5.3.0 release but I can do it as we go along if you prefer.

sbesson: Either way. The advantage of combining early is that we could cut some redundancy e.g. I wonder whether the "clean up..."  should be mentioned at all in a combined history.

hflynn:
Updated merging both milestones histories into one entry, keeping categories from m1 and adding in additional detail including the things noted by @sbesson and a link to the Glencoe press announcement. Hopefully better now!

sbesson:
Staged at https://www.openmicroscopy.org/site/support/bio-formats5.3-staging/about/whats-new.html and looks good to me. A couple of nitpicks unrelated to this PR that can be addressed post 5.3.0-m2:
- there is no `org.openmicroscopy:xsd-fu` artifact only `specification` and `ome-xml`
- should we include `ome:jxrlib` in the table of dependencies or rather in the components page?
- the quote format is inconsistent between the `org.openmicroscopy:*` components and `ome:jxrlib`

hflynn:
Quote format is related to this PR, I'll fix that now

dgault:
Having done a quick proof read it appears to read and display fine to me

hflynn:
Actually, do you want me to remove the xsd-fu artifact that doesn't exist now too @sbesson ?

hflynn:
2nd point not addressed here is covered by https://trello.com/c/HeoKWl5H/52-add-documentation-for-reading-czi-files-with-jpeg-xr

PR #2428, opened 2016-05-20
FileStitcher: avoid re-initializing underlying readers

simleo:
Keep individual readers from `ExternalSeries` open and don't re-initialize them in `initReader`. A test on `idr0010` data shows that, without this, individual readers can be reinitialized thousands of times if they are multi-series.

PR #2672, opened 2016-11-23
Options API: memoizer propagation

simleo:
Uses #2659 to get propagation of memoizing behaviour across a stack of wrappers and sub-readers. This is basically the same use case encountered in #2013, but here I've worked on `FilePatternReader` rather than `ScreenReader` because changes to the latter have not been ported to the mainline yet. Both readers rely on `FileStitcher` to group files together though, the main difference being that `ScreenReader` uses multiple stitchers whereas `FilePatternReader` uses only one.

These changes can be tested with a small ad-hoc program:

```java
import java.io.File;
import loci.formats.IFormatReader;
import loci.formats.Memoizer;

public final class TryPropagation {

  public static final long MIN_ELAPSED = 0L;
  public static final String DIR_PATH = "/tmp";

    String id = args[0];
    File dir = new File(DIR_PATH);
    IFormatReader reader = new Memoizer(new ImageReader(), MIN_ELAPSED, dir);
    reader.setId(id);
    for (int c = 0; c < reader.getSizeC(); c++) {
      reader.openBytes(reader.getIndex(0, c, 0));
    }
  }
}
```

You can get a minimal test setup starting from [multi-channel.ome.tiff](http://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/multi-channel.ome.tiff):

```
bfconvert multi-channel.ome.tif multi-channel-C%c.tif
echo 'multi-channel-C<0-2>.tif' >multi-channel.pattern
```

Then run the program with `multi-channel.pattern` as its argument. After running the program once you should get memo files (under `/tmp`) corresponding to the pattern file and to all individual image files:

```
.multi-channel-C0.tif.bfmemo
.multi-channel-C1.tif.bfmemo
.multi-channel-C2.tif.bfmemo
.multi-channel.pattern.bfmemo
```

Now run the program again. Looking at the logs (with level set to `DEBUG`) you should see that:

* In the first run, readers are initialized from scratch and saved to memo. Since `FileStitcher` does a bit of closing and reopening on its sub-readers (see also: #2428), you should also see memo loads from `.multi-channel-C{0,1,2}.tif.bfmemo`
* In the second run, *all* readers are always loaded from their memo files

--breaking

### DIFF
--- a/components/formats-api/src/loci/formats/ImageReader.java
+++ b/components/formats-api/src/loci/formats/ImageReader.java
@@ -830,6 +830,7 @@ public class ImageReader implements IFormatReader {
   @Override
   public void setId(String id) throws FormatException, IOException {
     IFormatReader currentReader = getReader(id);
+    currentReader.setMetadataOptions(getMetadataOptions());
     LOGGER.info("{} initializing {}",
       currentReader.getClass().getSimpleName(), id);
     currentReader.setId(id);
--- a/components/formats-bsd/src/loci/formats/FileStitcher.java
+++ b/components/formats-bsd/src/loci/formats/FileStitcher.java
@@ -49,7 +49,6 @@ import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.formats.in.DefaultMetadataOptions;
 import loci.formats.in.MetadataLevel;
-import loci.formats.in.MetadataOptions;
 import loci.formats.meta.MetadataStore;
 import loci.formats.Memoizer;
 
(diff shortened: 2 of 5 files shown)


sbesson:
@simleo there are pending conflicts so you would need to merge `origin/develop` into this branch. Two questions:
- are all the changes meant to be integrated or are the memoizer changes meant to provide an example of propagation? in the first case, we might want to isolate the 2 parts of the PR
- should we remove the `breaking` label?

simleo:
@sbesson ATM this contains three types of changes:

* `Memoizer` options propagation: I'm currently doing this across the board, so the changes here will be a subset of the upcoming propagation PR;
* Breaking changes to `getMemoFile` and `MemoizerTest`: these have to be thrown away as they break the current contract;
* Re-implementation of `Memoizer`'s configuration section according to the new API: I will open a new PR for this after the propagation work is done.

Taking into account all of the above, I'm closing this for now. We can still look at this as a usage example, although now we also have #2681 (I think that's even better as an example because it also shows how to get options from outside BF via the command line).

PR #2677, opened 2016-11-29
MetadataConverter: Deprecate and delegate to ome-xml implementation

rleigh-codelibre:
We have two copies of this code, one which appears to be manually copied into bioformats from ome-xml.  Use the ome-xml version, delegate this one to ome-xml and deprecate.  All internal uses use the ome-xml one, leaving this one purely for compatibility for any external users.

Testing: Check jobs are green.


dgault:
Deprecating this makes sense due to the duplication. Builds are green and the MetadataConverter can still be used for backwards compatibility.

Looks good to merge from my perspective.

sbesson:
Just two general impact questions. Unlike the other `Metadata*` API discussed yesterday, is it correct that  `MetadataConverter` was only used  internally via its static method so that no API method signature should be affected? Also is there any usage at the OMERO level that would need to be updated? /cc @mtbc

mtbc:
I'm not aware of any direct usage of `MetadataConverter` by OMERO, just implicitly via other Bio-Formats classes.

rleigh-codelibre:
Only used internally as far as I'm aware, but the change here is fully backward compatible for external users.  The only change there is moving to use the ome.xml.meta MetadataStore/Retrieve interfaces; since the loci.formats.meta equivalents extend these interfaces this is not breaking.  And since all the dropped methods were private, that's also non breaking.

sbesson:
Thanks for the extra information. My rationale for being extra careful is that this is effectively changing a the type of formal parameter in a public method which should a priori be considered as [breaking](http://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_classes_-_API_methods_and_constructors). However in this case, I think this change is safe since:
1. we are modifying a class rather than an interface
2. the class is `final` i.e. no-one should have extended it and would be broken
3. as said above, it uses a more general interface so it should work seamlessly for any existing code

PR #2678, opened 2016-11-29
MemoizerTest refactor

simleo:
Addresses several issues with `MemoizerTest`, mostly unused / redundant / duplicate code and the fact that temporary files were not being removed.

Since this PR contains whitespace changes, and to ensure no critical check has been inadvertently removed, it's best to review it one commit at a time, in order.

Functional checks include verifying that all tests pass and do not leave temporary files and directories around.

Some commit-specific additional notes:
 * b975b61b3d57677f4ae1aca0b122d9ce5f446a90: `testng.Assert{JUnit,}` have opposite conventions on tested vs expected argument ordering. Existing tests were using the `testng.Assert` convention.
 * 332f17bc5386865b769ef385ea8147c014907f82: uses `File.setWritable`'s return value to simplify the check and to take into account failures due to other reasons. If we want to be more explicit we can throw a `SkipException`.

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/MemoizerTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/MemoizerTest.java
@@ -204,15 +204,7 @@ public class MemoizerTest {
   @Test
   public void testGetMemoFilePermissionsDirectory() throws Exception {
     File directory = createTempDir();
-    directory.delete();
     Memoizer memoizer = new Memoizer(reader, 0, directory);
-
-    // Check non-existing memo directory returns null
-    assertNull(memoizer.getMemoFile(id));
-
-    // Create memoizer directory and memoizer reader
-    directory.mkdirs();
-    memoizer = new Memoizer(reader, 0, directory);
 
     // Check existing non-writeable memo directory returns null
     if (File.separator.equals("/")) {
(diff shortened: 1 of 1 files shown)


components/formats-bsd/test/loci/formats/utests/MemoizerTest.java
```diff
@@ -32,315 +32,201 @@
...
+      assertNull(memoizer.getMemoFile(id));
+    }
   }
 
   @Test
   public void testRelocate() throws Exception {
     // Create an in-place memo file
-    memoizer = new Memoizer(reader, 0);
+    Memoizer memoizer = new Memoizer(reader, 0);
     memoizer.setId(id);
```
simleo: I did not use `checkMemo` because it does the full cycle using the same id, while this test changes id between calls.

components/formats-bsd/test/loci/formats/utests/MemoizerTest.java
```diff
@@ -32,315 +32,201 @@
...
-  public void testSimple() throws Exception {
-      memoizer = new Memoizer(reader);
-      File f = memoizer.getMemoFile(id);
-      if (f != null && f.exists()) {
-        f.delete();
-      }
-      // At this point we're sure that there's no memo file.
-      reader.setId(id);
-      reader.close();
-      memoizer.setId(id);
```
simleo: The reason for removing that test was that it wasn't performing any explicit checks. I've just added a new one that performs the open/close on the wrapped reader, see if it looks useful.

simleo:
> add a test to document the special case of the constructor with directory=null

I think we already have that covered, by `testConstructorTimeElapsedNull` and `testConstructorReaderTimeElapsedNull`

sbesson:
You are right, I someway missed the `null` tests during my initial review. All looks good. Once validated by Travis and the daily builds, this should be ready to merge.

PR #2679, opened 2016-11-30
Options API: support for default values

simleo:
Follow-up to #2659. Adds single-argument getters and setters, which currently return `null` if the option is not set. To enable fallback to a set of defaults, we should just need to add:

```java
public FormatOptions(Properties defaults) {
  props = new Properties(defaults);
}
```

Although the defaults *would have to be expressed as strings*.

This is not included here because it requires substantial testing, but plugging it in at a later time should be non-breaking.

Note that even if we end up not doing that, these changes should still be useful by themselves (clients don't have to provide a default when none makes sense and can check if an option is unset).

Also removes support for `short`, `char` and `byte` (again, see #2659).

**Note:** there is no single-arg getter for `getEnum` (but clients can still check if the option is unset via `get()`). This is due to the fact that in this case the second arg is needed not just for its value but also for its type. If we want to enable this, we need to store the specific enum type when `setEnum` is called.


rleigh-codelibre:
All the latest changes look good to me.

sbesson:
Based on the latest review but also the history of #2659, merging this. Next steps will be to test the new API for specifying key/value pair options /cc @dgault @melissalinkert. With regard to documentation, is there  a place in the Sphinx developer documentation that would be worth reviewing and extending @simleo ?

simleo:
@seb I don't see any current section in the Sphinx docs that could be extended with info on this. The closest ones are [Reading files](http://www.openmicroscopy.org/site/support/bio-formats5.2/developers/file-reader.html) and [Writing files](http://www.openmicroscopy.org/site/support/bio-formats5.2/developers/file-writer.html) but options apply to both readers and writers. So perhaps a new section at the same level, i.e., under "Introduction to Bio-Formats". I think it's too early to write something though, since the docs would be about using the concrete feature owned by `FormatHandler` instances, and we're still ironing out the rough spots in that respect.

PR #2680, opened 2016-12-02
build: Unify specification and ome-xml versions

rleigh-codelibre:
Simplify the specification of these component versions.  Since they are coupled, use a single `ome-model.version` property for both.


sbesson:
Works as expected. Builds are green and documentation link work as expected. As both components are versioned together and released in locksteps, unifiying the version makes complete sense for all the use cases I can think of.

Issue #2634, opened 2016-10-25
BioFormats 5.2.4 CZIReader is broken for Tile Images


sebi06:
Hi,

I just tested BioFormats 5.2.4 with a 20x20 Tile Image I just acquired today. In former versions, such an image was recognized by BioFormats as 400 Image Series, no matter if the image in ZEN was with or without a pyramid.

Now BioFormats offers me to open 5 image series, which seem to correspond to the different pyramid levels. But all images except for the last image contain corrupted image information (see screenshot).

When I repeat the image acquisition without generating a pyramid, BF 5.2.4 does not recognize this image as a tile image any more (even it is exactly the same image). It now detects 400 image series again, which is the old (not nice) way to do it. But at least the image data are ok.

This is really nasty for the users, since the default way to acquire images in ZEN is of course with pyramid. Turning this off is a special option only.

**In fact with 5.2.4 most of the CZI images cannot be read an more! Any help is therefore really appreciated.**

Cheers, Sebi

Testdata: https://dl.dropboxusercontent.com/u/623476/CZI_Test_20x20Tile.czi

![bf5 2 4_20x20tile_broken](https://cloud.githubusercontent.com/assets/3833249/19687333/4ae81c5e-9ac5-11e6-8599-f32360730c7c.png)

dgault:
Hi Sebi,

Thank you for bringing this to our attention and providing test data.
I have been able to reproduce the issue as described and isolated the PR in which this functionality changed. 

I have added a card to our Trello inbox for managing this (https://trello.com/c/AldQjHV2/32-czireader-tiled-images-bug) and shall continue to debug to root cause the problem and work on a solution.

With Thanks,
David Gault

sebi06:
Hi David,

please let me know when you need more test CZI images from us. I have the feeling that the ongoing JPG-XR integration for reading CZI images seem to break some existing workflows for most of our users.

To be clear: Only a small percentage of our users actually use JPG-XR to compress their CZI images. Most of them use the normal uncompressed way. And 5.2.4 breaks some of their workflows that did run sooth so far.
This feature to read the pyramid levels is cool, but we have to be careful here not to introduce new issues in order to make the CZI JPG-XR work. People using JPG-XR are usuallay acquiring large tiles region (slide scanner), but the vast majority of CZIs are different and must also work with BioFormats / CZIReader.

melissalinkert:
[...]

Issue #2628, opened 2016-10-23
Avoid opening the image attachment for CZI files


sebi06:
Hi,

when I use the latest version of bioformats to open a CZI file from within Fiji, it always opens the image along with the image attachment. How can I avoid this?

Most CZI images do not use this attachment, so this should be an option, but not the default. Or is there already a way how to change this behavior?

Cheers, Sebi (from Zeiss)

melissalinkert:
Hi Sebi,

Each attachment image is stored as a separate Bio-Formats series, so there is already a way to prevent these images from being displayed.  When opening files in ImageJ/Fiji, the ```Open all series``` box should be unchecked in the ```Bio-Formats Import Options``` window.  In the resulting series chooser (the one with thumbnails), the attachment images should be shown at the very bottom, so just make sure they're unchecked.  If the files you're working with have a large number of series such that the thumbnail window is not shown, then choosing any combination of series except for the last one or two should work in general (depending on how many attachments there are).

If that doesn't help, then we'll need to know a little more about which options you have selected when opening the files.

sebi06:
Hi Melissa,

this sounds fine, but I still think there are some open questions here.

Imagine one acquires images from a wellplate, e.g. 96 images (2 Channels) = one per well. This used to be 96 series, which opened as a stack of 96 images usually, when the Open All option was checked. Now this will open the attachment as well and its dimension will be different from all the other image series. Therefore any automatic script that used to work will now have to deal with this one image inside the series, that is "different".

What also puzzle me is the fact that this approach does not really reflect the difference between

- 2 Wells with 9 Positions each --> 18 Series
- 2 Wells with 3x3 Tile Region each --> 18 Series? (by the way, the 2nd approach leads to a error --> separate issue)

Will the label image always be the last inside the whole series?

And currently this label image (BioFormats 5.2.4) returns an empty image with the wrong XY dimensions. This might be a bug.

From a user point of view the additional option to open an label images using a distinct way would be more more convenient in my opinion, because the user either wants to see the real image data or the attachment. To mix those together in one series fells strange to me.

sbesson:
Closing as per https://github.com/openmicroscopy/bioformats/issues/2634#issuecomment-266784419

PR #2681, opened 2016-12-02
Zeiss CZI: add options to disable autostitching and reading of attachment images

melissalinkert:
See https://github.com/openmicroscopy/bioformats/pull/2679, https://github.com/openmicroscopy/bioformats/issues/2632, https://github.com/openmicroscopy/bioformats/issues/2628, and https://trello.com/c/HeoKWl5H/52-add-documentation-for-reading-czi-files-with-jpeg-xr

This uses the new options API to add the ```zeissczi.allow.autostitch``` and ```zeissczi.include.attachments``` options in ```ZeissCZIReader```, both of which are expected to have boolean values and will default to ```true```.  The latter, when set to ```false```, should prevent attachment images (e.g. slide labels) from being read.  The former, when set to ```false```, should disable reading of all but the largest resolution in a pyramid, and will return that resolution as one series per image tile.

There are also changes to ```ImageInfo``` (== ```showinf```) and the ImageJ plugin configuration section to allow these options to be easily changed.  ```showinf``` can now accept command line arguments of the form ```-option key value``` to set any number of arbitrary key/value pairs.  In ImageJ, the ```Zeiss CZI``` entry in ```Plugins > Bio-Formats > Bio-Formats Plugin Configuration``` should now show two check boxes, one for each new option (both should be checked by default).

Things to test are:

* CI builds remain green - there should be no change to default behavior here
* ```showinf -nopix``` on any of the files in ```data_repo/curated/zeiss-czi/zeiss/jpeg-xr``` shows two more series than ```showinf -nopix -option zeissczi.include.attachments false``` on the same file
* ```showinf -nopix -option zeissczi.include.attachments false -option zeissczi.allow.autostitch false``` on any file whose configuration changed in https://github.com/openmicroscopy/data_repo_config/pull/154 or https://github.com/openmicroscopy/data_repo_config/pull/155 should result in a series count that matches the number of tiles reported by Zen (or slightly higher, in the case of files with multiple scenes).  Tile counts are shown in the ```Info``` tab in Zen, screenshots for which are available for most new/changed files.
* each combination of checked/unchecked in the two new boxes in ImageJ should produce series counts that match ```showinf``` with the relevant ```-option``` flags
* thumbnails in the series chooser window in ImageJ should look reasonable when the ```Automatically stitch tiled images``` box is unchecked (likely sufficient to spot-check a handful of files)
* tiles can be stitched with the Grid/Stitching plugin (per https://github.com/openmicroscopy/bioformats/issues/2634) when the ```Automatically stitch tiled images``` box is unchecked.  Spot checking a few smaller files is probably sufficient; ```curated/zeiss-czi/sebastian/Tiles*``` is a really small, but the FL datasets in ```curated/zeiss-czi/zeiss/jpeg-xr/``` are proper slides that may be small enough too (the BF ones are likely too big to fit into memory so I wouldn't suggest trying those).


simleo:
I have some changes, see melissalinkert#27. I have performed the following tests, with and without my changes:

Fiji
---

Opened `curated/zeiss-czi/qa-7771/CervantesNew_10EuroCentCoinSpain_Comb_8PO_OfflinePyramide.czi`

* with both options checked: 4 series
* with both options unchecked: 40 series

showinf
-------

```
% ./showinf -nopix /ome/data_repo/curated/zeiss-czi/zeiss/jpeg-xr/g_FL_1channel_noZ_compression85%.czi | grep "Series count"
Series count = 8
% ./showinf -nopix -option zeissczi.include.attachments false /ome/data_repo/curated/zeiss-czi/zeiss/jpeg-xr/g_FL_1channel_noZ_compression85%.czi | grep "Series count"
Series count = 6

% ./showinf -nopix /ome/data_repo/curated/zeiss-czi/qa-7771/CervantesNew_10EuroCentCoinSpain_Comb_8PO_OfflinePyramide.czi | grep "Series count"
Series count = 4
% ./showinf -nopix -option zeissczi.include.attachments false -option zeissczi.allow.autostitch false /ome/data_repo/curated/zeiss-czi/qa-7771/CervantesNew_10EuroCentCoinSpain_Comb_8PO_OfflinePyramide.czi | grep "Series count"
Series count = 40
```

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -88,6 +89,11 @@
 
   // -- Constants --
 
+  private static final String ALLOW_AUTOSTITCHING_KEY = "zeissczi.allow.autostitch";
```
sbesson: Semantics question: should we agree on some minimal naming schema for defined options? @joshmoore used `classname.key` in https://github.com/openmicroscopy/bioformats/pull/2013 although the latter referred directly to a field of the class.

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -962,48 +975,55 @@ else if (seriesCount > mosaics && mosaics > 1 && prestitched) {
...
-        {
-          if (!foundLabel) {
-            foundLabel = name.equals("Label");
-          }
-          if (!foundPreview) {
-            foundPreview = name.equals("SlidePreview");
-          }
-          segment.fillInData();
+    if (readAttachments) {
+      for (Segment segment : segments) {
```
sbesson: Is there a case for migrating the attachment reading to a private method?

melissalinkert:
Commits from https://github.com/melissalinkert/bioformats/pull/27 cherry-picked, with 0c54461 added to fix method name spelling and 533cd65 to address https://github.com/openmicroscopy/bioformats/pull/2681#discussion_r90904175.

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
melissalinkert: I was mostly modelling the names on https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java#L44.  I'm not especially attached to them, though, so happy to change if there is a better suggestion.

simleo:
What about changing `-option` to `-D` in `ImageInfo`? It seems to be a de facto standard for k/v options in Java apps.

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
rleigh-codelibre: `classname.key` is OK, though I think I'd prefer sometime more general.  Class names are internal implementation details.  For example, I'd prefer `tiff` over `TiffWriter`.  It's usable by both readers and writers, and also compatible with implementations in other languages which may have different class names.  And it precludes changing the implementation without breaking the option names.  And for an option like e.g. `jpeg.quality`, it can be used by the TIFF writer when using JPEG compression.  So I think it's more user-friendly to use accessible prefixes which are simpler, more flexible, and don't expose private implementation details.

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -1363,6 +1328,18 @@ else if (getZCTCoords(plane)[2] < timestamps.size()) {
     segments = null;
   }
 
+  // -- ZeissCZI-specific methods --
+
+  public boolean allowAutostitching() {
+    return ((DefaultMetadataOptions) getMetadataOptions()).getBoolean(
```
joshmoore: For me these are both breaking changes and generally dangerous. _Minimally_ these need to use an `instanceof`; otherwise these methods would throw if anyone has created their own implementation of the `MetadataOptions` class. See review comments for further thoughts.

simleo: You're right. We could change `setMetadataOptions` to *update* the object rather than reassign it, but any attempt to use a custom implementation of `MetadataOptions` would not work as expected (not on `FormatHandler` instances, anyway).

So, if we don't decide to go for one of your proposed refactorings of the options API, I will apply the `instanceof` fix.

joshmoore: The idea of updating objects is interesting, but we'd have to talk through what we want to support. The contract that is currently in place is that a user can pass the implementation of an interface. (And that gets at least _partially_ passed around.) Updating would violate that.

What's not working in `FormatHandler` at the moment?

melissalinkert: Is there anything you want me to actively change here at the moment, or is the next to-do to have a call to sort out API/inheritance changes?

sbesson: @melissalinkert: I think we want to sort this in a relevant call before taking any action. We can certainly update all usages of MetadataOptions in a go once we make the final decision on the API/inheritance pattern for 5.3.0.

bramalingam:
Half way through the ImageJ workflows today:
So far these are the issues noticed:

Opened image i_FL_3channel_2scenes_noZ_compression85%.czi
with the following options

attachment : true (default)
auto stitch : true (default)

series count : 10

attachment : false
auto stitch : true (default)

series count : 10 (Shouldn’t this be 8?) : should be without the label image and the macro image right?


Thumbnail loading issue:
Opened QA 10972 (https://github.com/openmicroscopy/data_repo_config/pull/154/files)
with the following options

attachment : false
auto stitch : false

Error loading thumbnail for series #7
Error loading thumbnail for series #13

series count : 20 (reproducible, doesn’t look like a memory issue from ImageJ)

Will complete the rest of the tests by noon tomorrow and update the comments.

Just a small question, I am currently picking files randomly from the folders and testing them, one or two files at the max for each workflow. Please let me know if you want me to repeat the workflows with more files as well.

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
sbesson: My 2c on the above:
- no strong argument against not using `classname` in favor of more generic ones, i.e. `zeissczi.xxxz works and also matches the convention in `LociPrefs`
- agreed with all recommendations of Simone, I would vote for lowercase across the board
- with regarding to length vs complexity, this will undoubtedly be a case by case decision. For instance, in this PR `zeissczi.include.attachments` could be simplified/shorterned as `zeissczi.include_attachments` or even `zeissczi.attachments`. On the other hand, if we think there might be similar options in the future, having a generic `zeissczi.include.xxx` makes complete sense
- a final point re maintenance. Like all other properties, these will effectively be public API and we will have to consider how to handle deprecation/upgrade.

sbesson:
@bramalingam for `i_FL_3channel_2scenes_noZ_compression85%.czi`, I can confirm that the count should be 10 with attachments included and 8 without (confirmed via the merge command-line tools). This might mean there is an bug with the options propagation.

sbesson:
Regarding https://github.com/openmicroscopy/bioformats/pull/2681#pullrequestreview-11572441, discussed with @joshmoore and @simleo this morning. Current proposal is to implement option 7 i.e. create a new implementation (`DynamicMetadataOptions` for now) containing the new key/value changes, use it internally (CL tools, ImageJ plugin) and keep driving it via use cases.

sbesson:
Excluding to get https://github.com/openmicroscopy/bioformats/pull/2685 through

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
sbesson: As discussed in the meeting, I would err on 1- not using classname, 2- using lowercase, 3- keeping names short while descriptive as much as possible.

In the case above, my preference goes with `zeissczi.autostitch` for the first option, for the second option, I don't have any good alternative to `zeissczi.include.attachments` to propose.

bramalingam:
Had a discussion with @melissalinkert , regarding the failure of the tile stitching functionality as well. There seems to be a general problem in the options being propagated from ImageJ to the reader, and thus will wait for the fixes for the time being.

simleo:
@bramalingam @melissalinkert options propagation is still WIP

rleigh-codelibre:
On 07/12/16 15:07, Sébastien Besson wrote:
> In the case above, my preference goes with |zeissczi.autostitch| for the
> first option, for the second option, I don't have any good alternative
> to |zeissczi.include.attachments| to propose.

zeissczi.attachments?  If other "things to include" are added later, we
can then move it to a further level?  It would not be hard to provide
compatibility mappings in the reader for renamed options.









The University of Dundee is a registered Scottish Charity, No: SC015096

melissalinkert:
Commits pushed to use #2685, change option keys to ```zeissczi.attachments``` and ```zeissczi.autostitch```, fix options propagation in ImageJ, and sort out one last issue with fused vs. tiled images.  That should address all outstanding comments, with the exception of:

1) from @simleo: https://github.com/openmicroscopy/bioformats/pull/2681#issuecomment-265105232
  * The standard ```-D``` syntax would mean we have something like ```-Doption.key=option_value```, which is fairly different from how other ```showinf``` flags are specified.  I don't strongly object to changing it if there is a consensus that it's a dealbreaker for this PR, but thought it better to err on the side of consistency with existing flags.

2) from @bramalingam:

> Thumbnail loading issue:
>Opened QA 10972 (https://github.com/openmicroscopy/data_repo_config/pull/154/files)
>with the following options
>
>attachment : false
>auto stitch : false
>
>Error loading thumbnail for series #7
>Error loading thumbnail for series #13
>
>series count : 20 (reproducible, doesn’t look like a memory issue from ImageJ)

  * Thumbnail loading errors for this file should no longer occur with the autostitch box unchecked, but are completely expected when the autostitch box is checked (as those are very large images).

simleo:
657fb1b4ddaa16491943e6bc5c2cf3a11a428cb0 should not be necessary if we merge #2690 first.

Apart from that, everything looks fine to me.

sbesson:
Thanks all, merging. Re https://github.com/openmicroscopy/bioformats/commit/657fb1b4ddaa16491943e6bc5c2cf3a11a428cb0, I propose we investigate the simplification post 5.3.0 in a patch release so that we can make sure we introduce no last minute regression.

PR #2682, opened 2016-12-05
OMEXMLReader: Handle BinData in non-Pixels elements such as Mask

rleigh-codelibre:
The reader made the assumption that `BinData` was only found inside `Pixels`.  However, it's also contained within `Mask`, and the reader was stripping out all `BinData` content when initially reading the OME-XML (it pre-processes it with a custom SAX parser before creating the `MetadataStore`).  This change makes the stripping selective, so it only occurs for `BinData` elements contained within Pixels.

This is related to https://github.com/ome/ome-model/pull/14 but is not coupled to it (it's useful for testing #14 with the sample in https://github.com/ome/ome-model/pull/13), but can be merged independently.  If you test these without this PR, you'll see that the ROI sample gets the `BinData` stripped out of the `Mask` ROI, and with this PR, it's correctly retained.

Testing: Check builds remain green.  Check that reading OME-XML with `Mask` `BinData` works with `showinf`.  The `specification/samples/2016-06/ROI.ome.xml` sample from https://github.com/ome/ome-model/pull/13 is useful for this purpose.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
@@ -347,7 +347,7 @@ public class OMEXMLReader extends FormatReader {
 
     @Override
     public void characters(char[] ch, int start, int length) {
-      if (currentQName.indexOf("BinData") < 0 || inPixels == false) {
+      if (!inPixels || currentQName.indexOf("BinData") < 0) {
         xmlBuffer.append(new String(ch, start, length));
       }
     }
@@ -373,7 +373,7 @@ public class OMEXMLReader extends FormatReader {
         inPixels = true;
       }
 
-      if (qName.indexOf("BinData") != -1 && inPixels == true) {
+      if (inPixels && qName.indexOf("BinData") != -1) {
         binData.add(
           new BinData(locator.getLineNumber(), locator.getColumnNumber()));
         String compress = attributes.getValue("Compression");
--- a/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
@@ -338,32 +338,61 @@ public class OMEXMLReader extends FormatReader {
     private final StringBuilder xmlBuffer;
     private String currentQName;
     private Locator locator;
+    private boolean inPixels;
 
     public OMEXMLHandler() {
       xmlBuffer = new StringBuilder();
+      inPixels = false;
     }
 
     @Override
     public void characters(char[] ch, int start, int length) {
-      if (currentQName.indexOf("BinData") < 0) {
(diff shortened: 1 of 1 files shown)


mtbc:
I assume that stripping one out but not the other actually makes sense? Does this reflect a more general difference in how the "normal" image data and the masks may be stored?

rleigh-codelibre:
This is an optimisation performed by the reader.  For all the Image data (BinData under Pixels), the reader stores the offset into the OME-XML.  This lets it read the pixel data directly without re-parsing the XML, and also allows the MetadataStore to not have BinData bloat within it, saving potentially vast amounts of memory.

However, this optimisation works only for pixel data returned via openBytes, which looks up the offset and reads the pixel data directly.  It does not work for BinData elsewhere in the model, such as inside Mask, where it strips out the data with no way to get it back.

The lack of uniformity in pixel data handling could be considered a model defect in its own right--why can't I store the Mask data in TiffData, for example?  But this PR specifically addresses the broken assumption that it's OK to strip out all BinData when in fact it's only safe to do so for Pixels BinData.

components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
```diff
@@ -338,67 +338,76 @@ else if (omexmlMeta.getPixelsBinDataCount(i) != 0) {
...
 
     public OMEXMLHandler() {
       xmlBuffer = new StringBuilder();
+      inPixels = false;
     }
 
     @Override
     public void characters(char[] ch, int start, int length) {
-      if (currentQName.indexOf("BinData") < 0) {
+      if (currentQName.indexOf("BinData") < 0 || inPixels == false) {
```
melissalinkert: ```!inPixels```?

components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
```diff
@@ -338,67 +338,76 @@ else if (omexmlMeta.getPixelsBinDataCount(i) != 0) {
...
       Attributes attributes)
     {
       currentQName = qName;
 
-      if (qName.indexOf("BinData") == -1) {
+      if (qName.indexOf("Pixels") != -1) {
+        inPixels = true;
+      }
+
+      if (qName.indexOf("BinData") != -1 && inPixels == true) {
```
melissalinkert: Just ```inPixels``` instead of ```inPixels == true```?

components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
mtbc: if fiddling anyway then maybe put it first given shortcut logic

rleigh-codelibre: Pushed fixes for these, thanks.

PR #2684, opened 2016-12-06
Add default tiff tile writing in MetadataOptions

dgault:
This PR adds the use of MetadataOptions to control if the writing of tiff based images using tiles should be the default behaviour.

Some points to note:
- The FormatOptions key used to set the default behaviour is made public via the TiffWriter
- The default tiling size selected is 256, any images smaller will default to full height and width

To test:
- Builds and tests should all be green
- New unit test added in this PR should run and pass

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -36,11 +36,13 @@ import java.io.IOException;
 import loci.common.RandomAccessInputStream;
 import loci.common.Region;
 import loci.formats.FormatException;
+import loci.formats.FormatOptions;
 import loci.formats.FormatTools;
 import loci.formats.FormatWriter;
 import loci.formats.ImageTools;
 import loci.formats.codec.CompressionType;
 import loci.formats.gui.AWTImageTools;
+import loci.formats.in.MetadataOptions;
 import loci.formats.meta.MetadataRetrieve;
 import loci.formats.tiff.IFD;
 import loci.formats.tiff.TiffCompression;
@@ -80,6 +82,11 @@ public class TiffWriter extends FormatWriter {
   /** TIFF tiles must be of a height and width divisible by 16. */
   private static final int TILE_GRANULARITY = 16;
 
+  /** Automatic tile writing off by default unless set in MetadataOptions using DEFAULT_TILE_WRITING_KEY */
+  public static final String DEFAULT_TILE_WRITING_KEY = "bioformats.tiff.defaultTileWriting";
+  private static final boolean TILE_WRITING_DEFAULT = false;
+  private static final int DEFAULT_TILE_SIZE = 256;
+  
   // -- Fields --
 
   /** Whether or not the output file is a BigTIFF file. */
@@ -545,7 +552,17 @@ public class TiffWriter extends FormatWriter {
   @Override
   public int getTileSizeX() throws FormatException {
     if (tileSizeX == 0) {
-      tileSizeX = super.getTileSizeX();
+      MetadataOptions options = getMetadataOptions();
+      boolean defaultTileWriting = TILE_WRITING_DEFAULT;
+      if (options instanceof FormatOptions) {
+        defaultTileWriting = ((FormatOptions) options).getBoolean(DEFAULT_TILE_WRITING_KEY, TILE_WRITING_DEFAULT); 
+      }
+      if (defaultTileWriting && DEFAULT_TILE_SIZE < super.getTileSizeX()) {
+        tileSizeX = DEFAULT_TILE_SIZE;
+      }
+      else {
+        tileSizeX = super.getTileSizeX();
+      }
     }
     return tileSizeX;
   }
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
The first commit (merging develop into itself) looks a bit redundant.  Could we drop that for cleanliness?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -80,6 +82,11 @@
   /** TIFF tiles must be of a height and width divisible by 16. */
   private static final int TILE_GRANULARITY = 16;
 
+  /** Automatic tile writing off by default unless set in MetadataOptions using DEFAULT_TILE_WRITING_KEY */
+  public static final String DEFAULT_TILE_WRITING_KEY = "bioformats.tiff.defaultTileWriting";
```
rleigh-codelibre: Do we have a policy for key naming?  This looks a little wordy.  `bioformats` is redundant since it's implicitly part of bioformats, as is `Writing`.  Would something like `tiff.tile = true` or `tiff.layout = tile|strip|scanline` be a bit more concise and to the point?  Also, why encode `default` in the name; isn't it implicitly a default just by virtue of being a default value?

components/formats-bsd/src/loci/formats/out/TiffWriter.java
```diff
@@ -545,7 +552,17 @@ else if (retrieve.getPixelsBinDataCount(series) == 0) {
   @Override
   public int getTileSizeX() throws FormatException {
     if (tileSizeX == 0) {
-      tileSizeX = super.getTileSizeX();
+      MetadataOptions options = getMetadataOptions();
+      boolean defaultTileWriting = TILE_WRITING_DEFAULT;
+      if (options instanceof FormatOptions) {
+        defaultTileWriting = ((FormatOptions) options).getBoolean(DEFAULT_TILE_WRITING_KEY, TILE_WRITING_DEFAULT);
```
rleigh-codelibre: Why don't we get a `TILE_SIZE_X` property here, and use that if non-null?  The "defaultTileWriting" logic and key doesn't seem to be conveying the needed information (the user-specified tile size).  The same applies to the Y tile size.

simleo:
I'm going to propose changes along the lines of melissalinkert#27, but first of all: why do we have a boolean option that controls falling or not falling back to a hardwired default size? Why not having an integer option that allows to set the tile size directly? I'm very confused.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
sbesson: See also https://github.com/openmicroscopy/bioformats/pull/2681#discussion_r90903910 for a related discussion. I am also in favor of dropping the `bioformats.` prefix at least when used directly by the API (it makes sense to use a prefix when exposing options in a more complex environment like Fiji/ImageJ where conflicts could happen). Do  we support more than two data layouts? If we ever think of doing it, I would also think something along the lines of `tiff.layout` might be the most flexible option.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
dgault: Im not sure I quite understand, the default tile writing is not the same as a user specified tile size. If the user has specified a tile size that will override any other logic and be used instead. The default is purely providing a default value and behaviour.

dgault:
It was originally decided that a boolean value would be used to allow a user to set if tiling should be used as the default mode. The new API functions we added for tiling allow the user to set and control the tile size directly, Im not sure how adding an option to directly set the size would make that process any easier other than duplicating the number of ways to set the values.

components/formats-bsd/src/loci/formats/out/TiffWriter.java
simleo: `{get,set}TileSize{X,Y}` should be reimplemented to manipulate properties in the options object.

simleo:
I think `{get,set}TileSize{X,Y}` should be reimplemented to manipulate properties in the options object. This would allow us to expose the options in `bfconvert`, similarly to what @melissalinkert did in #2681. The "no tiling at all" behaviour could be triggered by passing `0` or negative as the tile size.

rleigh-codelibre:
No tiling could also be conveyed by having the tile size properties set to `null`.  When we have the ability to introspect the default options, the default tile sizes could be deleted.

simleo:
Actually I would get rid of setters and corresponding fields for all these configurable items, because they greatly increase the change of leading to an inconsistent state. The getters are useful to allow retyping the long code needed to interact with the options object though (and they're harmless). Compare to #2681 and #2672 (now closed, but part of the `Memoizer` changes will be moved to a new PR).

sbesson:
One of the reason why we did not consider options for setting tile sizes is because this API has not been scoped and discussed when we originally worked though the new tiling API. The current proposal and the previous one implement the design discussed and agreed by everyone on the Formats team back in September.

While extending the use of options to set tile sizes etc could be considered, is it exactly how we want to use options and what are the pros and cons of using it as opposed to concrete methods? And what is the rationale for making this a priority this late in the development process? In particular, can you be more specific on the impact on our current timelines and the resources needed to review thi? Alternatively, would reviewing this with a concrete proposal and refactoring in an upcoming milestone be an option or would that put us in a complex place?

simleo:
> While extending the use of options to set tile sizes etc could be considered, is it exactly how we want to use options and what are the pros and cons of using it as opposed to concrete methods?

Setting `{X,Y}` tile size looks like the perfect option candidate to me. I've written above one of the pros, i.e., allowing to set these parameters by passing a CLI option to `bfconvert`. Also, taking advantage of propagation in cases where we have a stack of writers (I don't know if such a case exists yet, but it might exist in the future).

> And what is the rationale for making this a priority this late in the development process? In particular, can you be more specific on the impact on our current timelines and the resources needed to review thi? Alternatively, would reviewing this with a concrete proposal and refactoring in an upcoming milestone be an option or would that put us in a complex place?

I'm not saying it's a priority. One thing is having a tile size that is configurable by the Java programmer, and we already have that thanks to #2663. Another thing is exposing this to the external world and making it propagation-ready, as discussed above. I think that *once we decide* it's time to expose the tiling behavior via the options API, we might as well go the full distance and allow finer granularity, rather than only 256 vs sizeX (no tiling). If there's no time to do this now, I would postpone the change to a future milestone (if possible).

sbesson:
Closing to work on https://github.com/openmicroscopy/bioformats/pull/2685

PR #2685, opened 2016-12-07
Transitional MetadataOptions object

simleo:
See the discussion on #2681, in particular https://github.com/openmicroscopy/bioformats/pull/2681#pullrequestreview-11572441. This implements solution no. 7 from https://gist.github.com/joshmoore/8108f68055da27d4560951ad5ac921fc.

A note on the changes to `ImportProcess`: a reader always has an options instance, and the code that tried to work around null options was bugged anyway (it wasn't setting the metadata level on the newly created object). In theory someone could call `reader.setMetadataOptions(null)`, but we're not defending against this anywhere else (see other calls to `getMetadataOptions`).

**Remaining references to `DefaultMetadataOptions`:**

* `Image{Converter,Info}`: we can switch to the new options at a later time (`ImageInfo` needs to be changed in #2681).
* `{NRRD,Metamorph}Reader`: these can also be changed at a later time. It will probably happen during my ongoing work on options propagation.
* `{FormatReader,MetadataConfigurable}Test`: these should also be OK for now.

**Possible enhancements:**

* Add `public DynamicMetadataOptions getDynamicMetadataOptions() {return (DynamicMetadataOptions) metadataOptions}`?

  * This would be useless because that method is not part of the interface and we're doing non-breaking changes. So we would avoid casting the options object but we'd need to cast the reader/writer to a `FormatHandler`.

**Usage:**

I would use a try/catch pattern instead of checking via `instanceof`, i.e., do this:

```java
    try {
      return ((DynamicMetadataOptions) reader.getMetadataOptions()).get("k", DEFAULT);
    } catch (ClassCastException e) {
      return DEFAULT;
    }
```

Rather than this:

```java
    MetadataOptions opt = reader.getMetadataOptions();
    if (opt instanceof DynamicMetadataOptions) {
      return ((DynamicMetadataOptions) opt).get("k", DEFAULT);
    } else {
      return DEFAULT;
    }
```

It's one line shorter and you don't need to import `MetadataOptions`, although I'm not sure it's more readable, especially if the option name is long.


components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java
```diff
@@ -32,27 +32,21 @@
...
+public class DefaultMetadataOptions implements MetadataOptions {
 
-  public static final String READER_VALIDATE_KEY = "reader.validate.input";
-  public static final boolean READER_VALIDATE_DEFAULT = false;
+  private MetadataLevel metadataLevel;
+  private boolean validate;
 
   /**
    * Construct a new {@code DefaultMetadataOptions}. Set the metadata level
    * to {@link #METADATA_LEVEL_DEFAULT} and disable file validation.
```
sbesson: As per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-build/468/warnings40Result/ this needs to be updated as well.

components/formats-api/src/loci/formats/in/DynamicMetadataOptions.java
```diff
@@ -30,26 +30,71 @@
...
-package loci.formats;
+package loci.formats.in;
 
 import java.util.Properties;
 import java.io.File;
 
 
 /**
- * Generic configuration object for readers and writers.
+ * Configuration object for readers and writers.
```
sbesson: Once reviewed, it might be worth expanding on the description of this new implementation and its scope of usage.

components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java
simleo: Sorry. Just pushed the fix

components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
```diff
@@ -494,13 +492,8 @@ private void initializeFile() throws FormatException, IOException {
     baseReader.setMetadataFiltered(true);
     baseReader.setGroupFiles(!options.isUngroupFiles() || options.isGroupFiles());
     if(options != null && !options.showROIs()){
-        MetadataOptions mo = baseReader.getMetadataOptions();
-        if(mo == null){
```
sbesson: Although this should not affect our own reader stack which is largely extending `FormatHandler`, could the remove of this null pointer check affect e.g. any third party reader that would not define a default `MetadataOptions`?

sbesson: Briefly reviewed the various usages within ImportProcess and the reader will usually be a concrete ImageReader subreader created by `createBaseReader`, the OmeroReader or a wrapper for virtualstack which should delegate to the concrete reader. As all of these instances should be extending `FormatReader` so we definitely expect a default options object to be initialized by default the usage.
Setting `metadataOptions` to `null` is certainly a scenario that I do not expect our stack to handle and we would likely have to review how to deal with it with concrete use cases.

sbesson:
Last set of changes are in agreement with the decisions made earlier today for the scope of 5.3.0. All CI builds including the latest state of this PR passed. Merging in order to reintegrate the work in #2681.

PR #2690, opened 2016-12-08
ImageReader: ensure options reflect those of the actual reader

simleo:
This is a porting of https://github.com/openmicroscopy/bioformats/pull/2013/commits/f0d58c2cd9b8db921aa5886916e3f7ec8a42aa0e to the current state. Without this, a call to `reader.getMetadataOptions().set*` on an `ImageReader` instance won't be reflected in the actual reader (unless it's the first one in the list).

More specifically, **without** this PR, `testOptionsImplicit` is expected to fail, while `testOptionsExplicit` should pass because `ImageReader.setMetadataOptions` already assigns the provided instance to all readers. I've added the test anyway to prevent regressions on this front. **With** this PR, both tests should pass.

Note that I couldn't place the unit test under `formats-api` because it wouldn't work there (no concrete readers). Although this may look weird, we currently have several other unit tests in `formats-bsd` that use `ImageReader` (e.g., `EightBitLosslessJPEG2000Test`). Perhaps we should move `FakeReader` to `formats-api` as a default "dummy" reader.

### DIFF
--- a/components/formats-api/src/loci/formats/ImageReader.java
+++ b/components/formats-api/src/loci/formats/ImageReader.java
@@ -45,6 +45,7 @@ import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.formats.in.MetadataLevel;
 import loci.formats.in.MetadataOptions;
+import loci.formats.in.DynamicMetadataOptions;
 import loci.formats.meta.MetadataStore;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


dgault:
This makes sense and the units tests are passing with the PR included as expected. Only small note is that the readers aren't being closed again afterwards in the unit test.

I think having the test in formats-bsd is fine in this case as a concrete reader implementation is needed even if it is the FakeReader.

Should a similar change be made for ImageWriter or are focusing on readers specifically at this time?

sbesson:
Also agreed `FakeReader` is a concrete reader implementation for our own minimal test file format. Another proposal to separate the API test from the implementation would be to define a ` MockReader` class in the `formats-api` tests, create the `ImageReader` using the ClassList constructor. But this is outside the scope of this PR.

Re https://github.com/openmicroscopy/bioformats/pull/2690#issuecomment-265998485, there are definitely other aggregators/wrappers to review in the context of the options state/propagation. Having focused PRs for each type makes sense to me, so I would vote for integrating this into 5.3.0 and review the other APIs in upcoming 5.3.x releases.

simleo:
I'll open the PR for options propagation in reader wrappers (in practice only `FileStitcher`) pretty soon. I'm not sure if you also want to consider that one for 5.3.0.

sbesson:
@simleo: merging this. As we passed code freeze, the `FileStitcher` propagation can happen in 5.3.1 but feel free to open the PR.

simleo:
@sbesson OK, the stitcher PR is #2694

PR #2692, opened 2016-12-09
Tiling documentation and samples

dgault:
This PR adds documentation and samples detailing how to read and write files using the new tiling API functions added in https://github.com/openmicroscopy/bioformats/pull/2663

There are 2 sample files included, each showing a different way in which tiled files can be written:
- One simply reads and writes full plane images, allowing the writer to handle the tiling based on the tile sizes set
- The other specifically reads and writes individual tiles one at a time

These are included in a new tiling documentation page which can be found via the developers index page. For testing the documentation pages can be built using `ant docs-sphinx`


docs/sphinx/developers/tiling.txt
```diff
@@ -0,0 +1,129 @@
+Tiled reading and writing in Bio-Formats
+========================================
+
+This guide pertains to version 5.3 and later.
```
mtbc: Not sure this adds much given that it's part of the 5.3 docs bundle (and surely things might change "later") and tiling will get a mention in "what's new".

docs/sphinx/developers/tiling.txt
```diff
@@ -0,0 +1,129 @@
...
+
+Reading tiled images
+--------------------
+
+The reading of tiled images is straightforward and can be done in much the same way as reading a full image. In this case, to read an individual tile, we pass to the reader parameters for the x and y co-ordinates of the tile to read and the height and width of tile desired.
```
mtbc: no hyphen for "co-ordinate" elsewhere in docs

docs/sphinx/developers/tiling.txt
```diff
@@ -0,0 +1,129 @@
...
+
+::
+
+    byte[] tile = reader.openBytes(image, tileX, tileY, tileSizeX, tileSizeY);
+
+For Tiff based readers, if the image has specifically been written using tiles, then the tile width and height used can be found as below. These values can then be used with the above command to read the correct tiles individually.
```
mtbc: "TIFF-based" perhaps

mtbc: not sure what "specifically" adds here

docs/sphinx/developers/tiling.txt
```diff
@@ -0,0 +1,129 @@
...
+Each function takes in an integer parameter for the desired tile size. As not all tile sizes are supported, the image writer will round the requested value to the nearest supported tile size. The return value will contain the actual tiling size which will be used by the writer.
+
+To find out the tiling size currently being used at any point there are 2 further API functions to get the current tile size for a writer. If tiling is not being used or is not supported then the full image height and width will be returned.
+
+::
+
+	public int getTileSizeX() throws FormatException
+	public int getTileSizeY() throws FormatException 
+
+The tiling parameters for writers must be set after the image metadata is set. An example of initializing a writer for tiling is show below
```
mtbc: "shown"

docs/sphinx/developers/tiling.txt
```diff
@@ -0,0 +1,129 @@
...
+   :download:`SimpleTiledWriter.java <examples/SimpleTiledWriter.java>` - code from that class is
+   referenced here in part. You will need to have **bioformats_package.jar** in your 
+   Java CLASSPATH in order to compile SimpleTiledWriter.java.
+
+Reading and writing using tiling
+--------------------------------
+
+For the most efficient reading and writing of tiles you may instead wish to read in and write out the individual image tiles one at a time. 
+
+To do this you can setup the reader and writer as in the previous example above. In this case, when setting the tile height and width used it is important to store the return values which will be the valid tile size used by the writer.
```
mtbc: "set up"

docs/sphinx/developers/tiling.txt
mtbc: Sometimes before code you end with a period, sometimes not; I can't quite spot the pattern.

docs/sphinx/developers/tiling.txt
```diff
@@ -0,0 +1,129 @@
...
+
+            // Read tiles from the input file and write them to the output OME Tiff
+            byte[] buf = reader.openBytes(image, tileX, tileY, tileWidth, tileHeight);
+            writer.saveBytes(image, buf, tileX, tileY, tileWidth, tileHeight);
+          }
+        }
+
+Full working example code is provided in    
+   :download:`TiledReaderWriter.java <examples/TiledReaderWriter.java>` - code from that class is
+   referenced here in part. You will need to have **bioformats_package.jar** in your
```
mtbc: `:file:` markup

sbesson:
Can the 2 example classes can be added to `docs/sphinx/developers/examples/ExampleSuite.java` so that they get executed as part of the build?

dgault:
Updated to address the feedback and to add the 2 example files to the ExampleSuite now.

docs/sphinx/developers/examples/ExampleSuite.java
```diff
@@ -58,12 +58,18 @@ public static void main(String[] args) throws Exception {
...
     execute("FileConvert", new String[] {
       inputFile.getAbsolutePath(), convertedFile.getAbsolutePath()});
     execute("FileExport", new String[] {exportFile.getAbsolutePath()});
     execute("FileExportSPW", new String[] {exportSPWFile.getAbsolutePath()});
+    execute("SimpleTiledWriter", new String[] {
+        inputFile.getAbsolutePath(), simpleTiledFile.getAbsolutePath(), "256", "256"});
+    execute("TiledReaderWriter", new String[] {
+        inputFile.getAbsolutePath(), tiledFile.getAbsolutePath(), "256", "256"});
   }
 }
```
mtbc: nice to end with EOL

docs/sphinx/developers/examples/SimpleTiledWriter.java
```diff
@@ -0,0 +1,156 @@
...
+import java.io.IOException;
+import loci.common.services.ServiceFactory;
+import loci.formats.FormatException;
+import loci.formats.ImageReader;
+import loci.formats.meta.IMetadata;
+import loci.formats.out.OMETiffWriter;
+import loci.formats.services.OMEXMLService;
+
+/**
+ * Example class for reading a full image and use an OME Tiff writer to automatically write out the image in a tiled format,
```
rleigh-codelibre: Remaining 5.3 reference.  This is implicit since it's going to be released with 5.3 and the 5.3 documentation.

docs/sphinx/developers/examples/TiledReaderWriter.java
```diff
@@ -0,0 +1,183 @@
...
+import loci.common.services.ServiceFactory;
+import loci.formats.FormatException;
+import loci.formats.ImageReader;
+import loci.formats.meta.IMetadata;
+import loci.formats.out.OMETiffWriter;
+import loci.formats.services.OMEXMLService;
+
+/**
+ * Example class for reading and writing a file in a tiled OME Tiff format,
+ * using Bio-Formats version 5.3 or later.
```
rleigh-codelibre: Remaining 5.3 reference.

rleigh-codelibre:
In both examples:

- `tileSizeX` and `TileWidth` and `tileSizeY` and `tileHeight` are used as alternatives.  Why not use the same name throughout for consistency?
- `openBytes`: Why not create outside the loop, and pass in `buf[]` for efficiency in the loop, so the buffer can be reused?  That's good for the example to encourage efficient practice.
- `cleanup`: If the reader close throws, the writer close won't be run.  Put each in a separate `try` block for safety to avoid resource leaks?  Error messages contains `and` but should be `or`; won't matter if split into two blocks with individual messages.

All files:

- trailing whitespace in various places
- long lines which aren't autowrapped to 70 cols

Tiling example:

- While we compute `nXTiles` and `nYTiles`, would it make sense to add this to the tiling API to allow direct introspection of the tile counts for any image, using the set tile sizes? E.g. `getTileXCount` and `getTileYCount`?
- Reading and writing doesn't account for overlap with the image edge when the image size is not an exact multiple of the tile size?  Is that OK in this context, or does the overlapping tile needs its size computing?  Could we make this be tested with such an image size?

Documentation:

- Trailing whitespace and unwrapped lines here as well
- Code samples are directly copied.  However, they can be directly included from the sample code, along with Java syntax colouring.  See the C++ examples in ome-files-cpp for how to do this.  This means there's no code duplication, and the docs will always be up-to-date with the code samples, since they are using the code samples directly.
- Example of introspecting TIFF with IFD.TILE_LENGTH and IFD.TILE_WIDTH.  Shouldn't this be using getTileSizeX and getTileSizeY since the point of the tile API is to abstract reader-specific details such as this?  Could we remove this entirely to make it clear this is not recommended?

dgault:
Updates:
- Consistent use of tileSizeX instead of TileWidth, tileSizeY instead of tileHeight
- Use 2 seperate try/catch blocks for the cleanup step
- Reuse buffer for reading and writing in both examples
- Used literalinclude markup for the inline example code
- Removed '5.3 onwards' reference
- Removed any whitespace (though I could only find one or two spaces)
- Wrapped longer lines

Other points:
- The mention of the tile size IFD tags is due to the fact that the tile size information is stored there once the image is written. If the writer is still open and the user has access to it then they can use the API getters, however if they simply have an existing TIFF based image and want to know what tile size to use then you should read the information from the IFD tags as before.
- For tile sizes which overlap the image, the user will not have to worry about handling any padded data in this instance other than knowing that the number of tiles in the X and Y should account for these partial tiles. This should handled in the example.
- Any discussion of tiling API additions can be handled on a separate trello card

rleigh-codelibre:
@dgault If you have a TIFF image and need to know the tile sizes, why would you not use the TIFF reader and get it that way?  I don't understand why we would want people to ever use the low level TIFF code in a high-level example when we have an API for this now.  We used it before out of necessity, but I don't see why it would be strictly needed now.  You can't read and write a TIFF image concurrently, so getting this information from an existing TIFF could just use the reader API?

Regarding overlapping tile sizes, I'm unsure I understand what is happening here.  Is it legal for the tile size to overlap the image boundary when calling openBytes and saveBytes?  What happens in this case?  Is it clipped, or is it padded?  Does openBytes return a buffer allocated for the clipped region or the full region with padding?  The user will have to care about that if they want to use the data safely.  I think the specific behaviour needs clearly documenting here to avoid any doubt about what's happening.  As a user of the API, I must be certain about how to call this, and what I'm getting back.  The example doesn't show this, and since it's being passed straight back to saveBytes, the structure of the data in the buffer is not mentioned.  It's safe if I'm transferring the data immediately to saveBytes since they are hopefully symmetric, but an end user will likely want to do something else with the data.

dgault:
So which reader API call do you suggest be used instead?
If you want a more detailed example of an overlapping tiled scenario I will be happy to put one in on Monday.

sbesson:
Merging for 5.3.0 release as the major part of the tiling documentation is in place with examples tested by the build. Let me know if there is some amendment to the tiling documentation page following the last set of comments that would justify an emergency doc release or whether doc improvements can be handled via regular patch releases @rleigh-codelibre and @dgault

PR #2694, opened 2016-12-09
Options propagation: FileStitcher

simleo:
In most cases, there is no need to do anything to propagate options in reader wrappers. This is due to the fact that they do not have their own options instance, but rather refer to that of their wrapped reader.

`FileStitcher` is an exception though, since it uses several external readers to do its job, while the "standard" `ReaderWrapper` contract assumes there is only one wrapped reader, whose behaviour will be somehow modified by the wrapper. In practice, `FileStitcher` has two main states:

1. **before calling setId** the file pattern hasn't been seen yet, so `FileStitcher` does not have enough info to even decide *how many* subreaders it needs. Options consistency is guaranteed in this state, but the only subreader is a placeholder;

2. **after calling setId** concrete subreaders are in place, and they should have the same options as the `FileStitcher` itself. This PR ensures this by getting the options object from the placeholder reader and assign it to all external readers right after they are constructed.

Speaking of subreaders, I've also fixed a NPE that could be triggered by calling `getUnderlyingReaders` in state no. 1. I've chosen to fall back to the generic `ReaderWrapper` behaviour and return the placeholder, but returning `null` might also make sense (i.e., "there are no *real* subreaders"). Throwing an `IllegalStateException` (directly or via `assertId`) might also make sense.

However, the docs are not clear enough on what we mean by "underlying readers". For instance, `ImageReader` returns its list of "candidate" readers, while `FileStitcher` returns a list of readers that are actually being used as delegates to do its job. Finally, unless I've missed something, no concrete reader overrides the method, although it should make sense for any reader that uses one or more other readers (e.g., `ScanrReader`). Since `ImageReader`, in mos respects, acts as a "mutant" reader that becomes a concrete reader once `setId` has been called, perhaps it would be more appropriate to return `getReader().getUnderlyingReaders()`, at least after `setId`.

In the future, if we want to get rid of `FileStitcher`'s dual nature, we might make it something different from a `ReaderWrapper`. We could have a `ReaderAggregator` (or similar) abstract class that formalizes a one-to-many relationship. Of course, external readers would still be created after `setId`, so the best option for `getUnderlyingReaders` would probably be to return an empty array.

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
@@ -219,6 +219,7 @@ public class FileStitcherTest {
     assertNotNull(fs.getUnderlyingReaders());
     fs.setId("test_z<0-2>.fake");
     assertNotNull(fs.getUnderlyingReaders());
+    fs.close();
   }
 
   @Test
--- a/components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
@@ -65,6 +65,20 @@ public class FileStitcherTest {
   public static final String KEY = "test.option";
   public static final String VALUE = "foo";
 
+  public static void checkKV(IFormatReader r, String k, String expv) {
+    MetadataOptions rOpt = r.getMetadataOptions();
+    assertTrue(rOpt instanceof DynamicMetadataOptions);
+    String v = ((DynamicMetadataOptions) rOpt).get(k);
+    assertNotNull(v);
+    assertEquals(v, expv);
+  }
+
+  public static void checkKV(IFormatReader[] readers, String k, String expv) {  
+    for (IFormatReader r: readers) {
+      checkKV(r, k, expv);
+    }
+  }
+
   // expected core metadata for the final stitched image
   private static final int PIXEL_TYPE = FormatTools.UINT8;
   private static final int SIZE_X = 128;
@@ -229,13 +243,12 @@ public class FileStitcherTest {
     FileStitcher fs = new FileStitcher();
     fs.setMetadataOptions(opt);
     fs.setId("test_z<0-2>.fake");
-    for (IFormatReader r: fs.getUnderlyingReaders()) {
-      MetadataOptions rOpt = r.getMetadataOptions();
-      assertTrue(rOpt instanceof DynamicMetadataOptions);
-      String v = ((DynamicMetadataOptions) rOpt).get(KEY);
-      assertNotNull(v);
-      assertEquals(v, VALUE);
-    }
+    checkKV(fs.getUnderlyingReaders(), KEY, VALUE);
+    DynamicMetadataOptions newOpt = new DynamicMetadataOptions();
+    String newValue = VALUE + "_";
+    newOpt.set(KEY, newValue);
+    fs.setMetadataOptions(newOpt);
+    checkKV(fs.getUnderlyingReaders(), KEY, newValue);
     fs.close();
   }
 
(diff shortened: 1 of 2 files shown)


dgault:
With regards the NPE, it seems that like there are a large number of functions in the FileStitcher which would throw a similar NPE, with the few that do check for null handling it by calling the function on `reader` (which is `DimensionSwapper` and does not have `getUnderlyingReaders`, hence it would call the `ReaderWrapper` also). So in short, removing the NPE here and handling it with `ReaderWrapper` seems consistent with the rest of the FileStitcher behaviour, although theres probably a larger tidy up of potential NPE's needed at some point in the future. 

I agree also that the inconsistent behaviour of `getUnderlyingReaders` across readers is also something that could be better documented in future.

The setting of the `MetadataOptions` looks good and the unit tests cover the necessary scenarios and all pass as expected.

dgault:
With the last commit this PR looks good from my perspective.

components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
```diff
@@ -193,4 +212,43 @@ public void testStitch(Integer[] dims) throws IOException, FormatException {
     check(pattern, filenames.toArray(new String[filenames.size()]), dims);
   }
 
+  @Test
+  public void testUnderlyingReaders() throws IOException, FormatException {
+    FakeReader reader = new FakeReader();
+    FileStitcher fs = new FileStitcher(reader);
+    assertNotNull(fs.getUnderlyingReaders());
+    fs.setId("test_z<0-2>.fake");
+    assertNotNull(fs.getUnderlyingReaders());
```
sbesson: I assume the underlying readers should have been expanded from a single reader to 3 fake readers in that case? Should this also be covered in the unit test?

components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
```diff
@@ -193,4 +212,43 @@ public void testStitch(Integer[] dims) throws IOException, FormatException {
...
+  public void testOptionsImplicit(MetadataLevel level)
+      throws IOException, FormatException {
+    FileStitcher fs = new FileStitcher();
+    fs.getMetadataOptions().setMetadataLevel(level);
+    fs.setId("test_z<0-2>.fake");
+    for (IFormatReader r: fs.getUnderlyingReaders()) {
+      assertEquals(r.getMetadataOptions().getMetadataLevel(), level);
+    }
+    fs.close();
+  }
```
sbesson: How should the options be handled in the case of closing/reopening? Does it fundamentally fall back to whether options have been set explicitly/implicitly before the second call to `setId` or are the options from the initial `setId()` call still used?

components/formats-bsd/test/loci/formats/utests/FileStitcherTest.java
simleo: As discussed above, the `getUnderlyingReaders` protocol is not well defined, and in `FileStitcher` it also depends on the state, so I didn't feel like adding a more specific test here. In a way, unit tests are part of the docs. I can add a length check before and after `setId` if you want, but I think this would make the current behaviour more "official".

simleo:
@sbesson the additional changes to `FileStitcher` from #2013 are necessary if we want `setMetadataOptions` to also have an effect if called *after* `setId`, and I think the consensus is that we do want this to happen (as much as possible -- I'm not sure it would work with things like `Memoizer` wrapping anyway). Thanks for pointing this out, I will push the changes ASAP.

rleigh-codelibre:
The last commit changes look OK to me.  Is any specific testing needed for this one?

simleo:
@rleigh-codelibre in principle we could perform a real-world test by stitching together Zeiss CZI stacks (that's the only format where we currently have active options), but just checking that unit tests pass should suffice for now.

sbesson:
@simleo: tested https://github.com/openmicroscopy/bioformats/pull/2694#issuecomment-270378056 with a completely artificial fileset created by copying a CZI file as `test_c1.czi` and `test_c2.czi`. Then ran the following command

```
showinf -nopix -nometa -stitch -option zeissczi.autostitch false test_c1.czi 
```

With Bio-Formats 5.3.1, the command above results in an ``Exception in thread "main" java.lang.IllegalArgumentException: Invalid series: 8`` due to mismatching dimensionality. With this PR include, the reader initializes and creates a large number of series with sizeC=2.

PR #2696, opened 2016-12-13
ImageJ: preserve original pixel values when exporting signed data

melissalinkert:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-December/006286.html

This should resolve some strange behavior when exporting from an
ImagePlus that represents signed data.  There are two parts to the fix:

1. don't overwrite the MetadataStore's pixel type from int* to uint*
2. when filling the byte[] to pass to saveBytes, use the ImageProcessor's calibration table to determine the real pixel values

To test, use QA 17471 and the ImageJ macro at the end of the above thread (adjusting the file paths as needed).  Without this change, run the macro in ImageJ and verify that the two histograms have different min/max values but the same overall shape.  Verify that ```showinf -minmax -range 1 1``` on the original .dv shows that the pixel type is int16, and that ```showinf -minmax``` on the saved TIFF shows that the pixel type is uint16.  Min/max values in both cases should match the corresponding ImageJ histogram.

With this change, re-run the ImageJ macro and confirm that both histograms are identical.  ```showinf -minmax``` on the saved TIFF should now show pixel type int16, and the min/max values should match the ImageJ histogram.

It would probably also be a good idea to do the same test with one of our existing int8 samples (or a fake file), perhaps as well as one or two unsigned datasets just to double-check that nothing was broken by this change.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -435,11 +435,23 @@ public class Exporter {
                 store.setPixelsID(MetadataTools.createLSID("Pixels", 0), 0);
             }
 
-            // always reset the pixel type
+            // reset the pixel type, unless the only change is signedness
             // this prevents problems if the user changed the bit depth of the image
+            boolean applyCalibrationFunction = false;
             try {
-                store.setPixelsType(PixelType.fromString(
-                        FormatTools.getPixelTypeString(ptype)), 0);
+                int originalType = FormatTools.pixelTypeFromString(
+                  store.getPixelsType(0).toString());
+                if (ptype != originalType &&
+                  (!FormatTools.isSigned(originalType) ||
+                  FormatTools.getBytesPerPixel(originalType) !=
+                  FormatTools.getBytesPerPixel(ptype)))
+                {
+                  store.setPixelsType(PixelType.fromString(
+                    FormatTools.getPixelTypeString(ptype)), 0);
+                }
+                else if (FormatTools.isSigned(originalType)) {
+                  applyCalibrationFunction = true;
+                }
             }
             catch (EnumerationException e) { }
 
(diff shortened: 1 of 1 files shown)


dgault:
Testing using the sample files from QA 17471
Histogram before the PR has been applied showing different min and max values:
screen shot 2017-01-11 at 12 13 59

Histogram after the PR has been applied showing the correct min and max values:
screen shot 2017-01-11 at 12 12 16

The showinf details on the saved file after the PR has been applied also now show the correct int16 pixel type as opposed to uint16 before.

Other than generated samples the only other genuine sample I could find with int8 pixel type was an ICS file nuts_bolts. When rerunning the tests with this file file however the histogram remained different for the saved file even with the PR included.

melissalinkert:
I haven't yet successfully reproduced the differing histograms with ```curated/ics/melissa/nuts_bolts.ics```.  With this macro:

```
run("Bio-Formats Importer", "open=/home/melissa/windows_d/ics/melissa/nuts_bolts.ics autoscale color_mode=Default rois_import=[ROI manager] split_channels view=Hyperstack stack_order=XYCZT");
wait(1000); // imagej-voodoo
run("Bio-Formats Exporter", "save=/home/melissa/resaved_data.tif compression=Uncompressed");
run("Bio-Formats Importer", "open=/home/melissa/resaved_data.tif autoscale color_mode=Default rois_import=[ROI manager] view=Hyperstack stack_order=XYCZT");
selectWindow("nuts_bolts.ics - C=0");
run("Histogram");
selectWindow("resaved_data.tif");
run("Histogram");
selectWindow("nuts_bolts.ics - C=0");
resetMinAndMax();
```

I see these histograms (version window included for confirmation that the correct jar was installed):

image

dgault:
I reran the tests again using the file nuts_bolts.ics and the macro listed above above. For some reason I am seeing different results as before with different histograms. The jars used have been taken from the artifacts directory after running `ant clean jars`. Screenshots below:

screen shot 2017-01-30 at 16 12 11

screen shot 2017-01-30 at 16 17 04

Im not sure what is causing it though as the build works as expected with QA 17471. Is there anything obvious that I am doing wrong?

melissalinkert:
Could you please run ```md5sum``` on ```nuts_bolts.ics```?  There are two different ```nuts_bolts.ics``` files - the one in ```data_repo/curated/ics/melissa/``` (which should be int8), and the one in ```data_repo/curated/ics/nico/images/``` (which should be uint8).  If you have the uint8 one, that might explain the difference.

```
mlinkert@necromancer /ome/data_repo/curated/ics/nico/images $ md5sum nuts_bolts.ics
f7461d0b621c2cdaadcfca6818e1acf8  nuts_bolts.ics
mlinkert@necromancer /ome/data_repo/curated/ics/melissa $ md5sum nuts_bolts.ics
3e65c2b1bc68b4b72df85249e33be69c  nuts_bolts.ics
```

dgault:
Aha, that is exactly what it was, did not realise there were 2 different files. Retested with `curated/ics/melissa` and I am now seeing the correct histograms.

screen shot 2017-01-30 at 17 26 34

PR should be good to merge now.

PR #2699, opened 2016-12-15
TiffWriter - fix for incorrect last IFD and new unit test

dgault:
This is a bug fix for an issue reported in https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8179&p=17776#p17776

The lastIFD was being written incorrectly due to https://github.com/openmicroscopy/bioformats/pull/2663/commits/4d48e1144da4c1035ff3a79ddcd556bd2c46582d which was part of https://github.com/openmicroscopy/bioformats/pull/2663

With this PR all non tiled behaviour should be identical as to prior to 5.3.0

This PR also introduces a new unit test to help ensure that such an issue will be caught in future:
- The new saveBytes test is largely similar to the tiled writing unit tests which had been added previously
- A new system prop has been added to enable the test in the same way as for the tiling test
- The tiling and saveBytes tests have both then refactored to reduce duplication in the unit tests

**To reproduce the original issue:**
- See the forum thread and using the sample files mentioned follow the bfconvert step
- Verify that the image opens with 49 frames as opposed to 2 channels of 25 frames
- Also, run the new unit test provided while excluding commit https://github.com/openmicroscopy/bioformats/commit/cdae6d17e9bc52788556413ea85b6582c38c0de5. This should fail with incorrect image counts for saveBytes tests

**To verify:**
- With the PR included perform the same bfconvert and verify that it correctly has 2 channels of 25 frames
- Verify that the new unit test runs without error

Note: To run the new unit test you must set the system property as below:
`-Dtestng.runWriterSaveBytesTests=100`

### DIFF
--- a/components/formats-bsd/test/loci/formats/utests/out/TiffWriterTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/out/TiffWriterTest.java
@@ -501,28 +501,7 @@ public class TiffWriterTest {
 
     File tmp = File.createTempFile("tiffWriterTest_Tiling", ".tiff");
     tmp.deleteOnExit();
-    TiffWriter writer = new TiffWriter();
-    String pixelTypeString = FormatTools.getPixelTypeString(pixelType);
-    writer.setMetadataRetrieve(createMetadata(pixelTypeString, rgbChannels, seriesCount, littleEndian, sizeT));
-    writer.setCompression(compression);
-    writer.setInterleaved(interleaved);
-    writer.setTileSizeX(tileSize);
-    writer.setTileSizeY(tileSize);
-    writer.setId(tmp.getAbsolutePath());
-
-    int bytes = FormatTools.getBytesPerPixel(pixelType);
-    byte[] plane = getPlane(PLANE_WIDTH, PLANE_HEIGHT, bytes * rgbChannels);
-    Plane originalPlane = new Plane(plane, littleEndian,
-      !writer.isInterleaved(), rgbChannels, FormatTools.getPixelTypeString(pixelType));
-
-    for (int s=0; s<seriesCount; s++) {
-      writer.setSeries(s);
-      for (int t=0; t<sizeT; t++) {
-        writer.saveBytes(t, plane);
-      }
-    }
-
-    writer.close();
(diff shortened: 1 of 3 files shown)


components/formats-bsd/test/loci/formats/utests/out/TiffWriterTest.java
```diff
@@ -115,41 +116,61 @@
...
     int[] seriesCounts = {1, 5};
+    int[] timeCounts = {1};
     String[] compressions = {COMPRESSION_UNCOMPRESSED, COMPRESSION_LZW, COMPRESSION_J2K, COMPRESSION_J2K_LOSSY, COMPRESSION_JPEG};
+    return getData(tileSizes, channelCounts, seriesCounts, timeCounts, compressions);
+  }
+  
+  @DataProvider(name = "nonTiling")
+  public Object[][] createNonTiling() {
+    if (percentageOfTilingTests == 0) {
+      return new Object[][] {{0, false, false, 0, 0, null, 0}};
```
sbesson: The default should be updated to return 8 parameters to the test

dgault: Yep, just spotted this, should be corrected now

rleigh-codelibre:
This restores the expected behaviour for me using the user-provided data file.

Is there a default value for `testng.runWriterSaveBytesTests`?  I don't see one in `components/formats-bsd/build.xml`, where it seems to set it from a property by the same name; maybe set a default here?  I don't see it elsewhere either.  Same question for the tiling test value.

dgault:
Running the tests from https://github.com/openmicroscopy/bioformats/pull/2663#issuecomment-263026487 also show that without the PR tiffinfo shows:
`TIFFReadDirectory: Failed to read directory at offset 165485.`

With the PR tiffinfo is able to read the images correctly and the script runs correctly

sbesson:
Discussed  with @rleigh-codelibre and @dgault the last comment, currently tests are using `0` as the default while Jenkins sets the value to 100. The proposal moving forward is to try and set the default for both tests to a low value (`1`?) so that minimum test coverage happens unless otherwise specified.  In the meantime the extra unit tests are successfully https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/479 lists 500 extra unit tests corresponding to the addition of this PR. Merging to start testing.

PR #2701, opened 2016-12-15
Build Sphinx docs with maven

rleigh-codelibre:
As part of making all of bioformats build with maven, this is now made possible by borrowing the sphinx logic from ome-model:

- Add maven docs build, taken from ome-model with some amendments
- Make `Makefile` invokes `mvn` rather than `ant`
- Ant build updated to generalise some hardcoded behaviour and
  to use a separate build directory as for maven to allow
  correct conf.py generation for maven
- Ant build versioning fix (regex incompatible with maven version in place of git tag)
- Drop Ant manpage generation (there are no manpages)
- Drop remaining Ant PDF generation logic
- Invalid options converted to verbatim strings in the sphinx docs

This PR does not:

- Remove the ant logic, retained for compatibility due to issues invoking maven directly from ant when using the Jenkins Ant build step plugin (it's not guaranteed to be on the `PATH`), but this could be run on a suitable build node.  Can be made to delegate to maven, or drop entirely, as a future followup.

Testing:

```
cd docs/sphinx
make clean html linkcheck
ant clean html linkcheck
mvn clean
mvn
mvn -Dsphinx.builder=linkcheck
```

Or run from the top level with `ant docs-sphinx` or `mvn`.

The `Makefile` behaviour should be completely preserved, including the theming, so the build jobs should not require any changes.  We can switch over the jobs to use maven as a followup step.  Likewise the `_build` directory used by ant could be replaced with `target/sphinx`, but was retained for job compatibility when copying the built docs, but can also be switched over.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
   fast_finish: true
 
 before_install:
-  - if [[ $BUILD == 'ant' ]]; then pip install --user flake8 Sphinx==1.2.3; fi
+  - pip install --user -r requirements.txt
 
 script:
   - ./tools/test-build $BUILD
--- a/ant/toplevel.xml
+++ b/ant/toplevel.xml
@@ -100,8 +100,6 @@ Type "ant -p" for a list of targets.
     description="generate the Sphinx HTML documentation">
     <echo>----------=========== Sphinx HTML ===========----------</echo>
     <ant dir="docs/sphinx" target="html"/>
-    <echo>----------=========== Sphinx MAN ===========----------</echo>
-    <ant dir="docs/sphinx" target="man"/>
   </target>
 
   <target name="clean-docs-sphinx"
--- a/docs/sphinx/Makefile
+++ b/docs/sphinx/Makefile
@@ -1,36 +1,39 @@
 # Delegate all make targets directly to ant
 
-ifdef SPHINXOPTS
-ANT_SPHINXOPTS := -Dsphinx.opts="$(SPHINXOPTS)"
-endif
-
 ifdef BF_RELEASE
-ANT_BF_RELEASE := -Drelease.version="$(BF_RELEASE)"
+MVN_BF_RELEASE := -Drelease.version="$(BF_RELEASE)"
 endif
 
 ifdef RELEASE
-ANT_RELEASE := -Dsphinx.release="$(RELEASE)"
+MVN_RELEASE := -Dsphinx.release="$(RELEASE)"
 endif
 
 ifdef VERSION
-ANT_VERSION := -Dsphinx.version="$(VERSION)"
+MVN_VERSION := -Dsphinx.version="$(VERSION)"
 endif
 
 ifdef SOURCE_BRANCH
-ANT_SOURCE_BRANCH := -Dsphinx.source.branch="$(SOURCE_BRANCH)"
+MVN_SOURCE_BRANCH := -Dsphinx.source.branch="$(SOURCE_BRANCH)"
 endif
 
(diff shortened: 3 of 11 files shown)


rleigh-codelibre:
Note: The job changes required to build this are in configuration version "2016-12-16_13-19-18" in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-docs/jobConfigHistory/ and will need restoring when this is unexcluded.

sbesson:
Tested today from the HEAD of the branch. Locally the documentation builds as expected using Maven and the logic inherited from the work on https://github.com/ome/ome-model/pull/19. Conceptually, this PR fits with the overall effort of streamlining our build system and preparing the replacment of Ant in favor of Maven.

With regard to our upcoming timeline/roadmap, the immediate goals will be to deliver bug fixes for the functionalities introduced in 5.3.0 via quick patch releases. Modifying the underlying build system of the documentation would require extra as shown by https://github.com/openmicroscopy/bioformats/pull/2701#issuecomment-267595410. Also, similarly to what has been done for 5.3.0-m1, a minor version increment is certainly the most appropriate place to conduct, review and document changes to the build system for any component.

The proposal is to hold off on this changes for the upcoming 5.3.x releases. When we initiate the next set of changes for the following minor/major release work on Bio-Formats, starting with this migration is certainly appropriate. As shown above, some continuous integration work will be necessary as well as some documentation. It would be good to list and review all the places that will need modification before initiating the work. Happy to leave this PR opened but excluded or close it and reopen it when appropriate.

A few policy points that might also require discussion:

- requirements: capping vs uncapping the Sphinx version. At the moment, all other documentation components across the board cap it to 1.2.3. /cc @hflynn 
- artifact: what is the rationale for the the renaming even if this component is effectively unused at the moment?

rleigh-codelibre:
Regarding requirements, it's uncapped because this PR also makes a couple of tweaks to allow it to build with the latest sphinx release making the version restriction no longer necessary.

Regarding the artefact name, it's to align it with the ant build name used in `zipfileset`, we could also add the `bio-` prefix as well (though this would then mismatch with the other component names, which all start with `formats`).  `docs` is also not a grammatically correct contraction; `documentations` is not a word.

rleigh-codelibre:
Regarding 5.3.x releases, can't we start a 5.3 branch to allow work to proceed on 6.0.x independently?  The model PRs in particular, which can't be made for 5.3.x, need to be merged so we can work on the model and python bindings, and this is a blocker.

sbesson:
Understood regarding Sphinx and artifact renaming. The latter is likely not critical but we should treat any artifact renaming with care. No particular veto to uncapping Sphinx if everything is fixed although we should probably discuss and agree at the organization level as we have already been down this road before. /cc @hflynn 

My only constraint for 5.3.x is that at the moment we should preserve the existing workflow, documented branches and working infrastructure notably all the CI jobs on http://ci.openmicroscopy.org/. Happy to derive of a new integration branch for the extended metadata work as suggested in https://trello.com/c/rSHDJmDY/51-branches-and-infrastructure.

rleigh-codelibre:
@sbesson If you want to create an integration branch, that's great.  We'll need one for the model as well, unless we want to just use `master` and not make changes for the version used in 5.3.x.  On the other hand, that branch could just as easily be `develop` if we created a `5.3` or `dev_5_3` for backporting fixes--there's little practical difference other than the naming.

sbesson:
Closing for now and created an [investigation card](https://trello.com/c/B3J72U0V/108-maven-spinx-docs) to sort out some final design points (CI, linkcheck).

PR #2703, opened 2016-12-16
Version History 5.3.1

dgault:
Updated version history for 5.3.1 release.

Only PR included is https://github.com/openmicroscopy/bioformats/pull/2699#issuecomment-267576295


docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,14 @@
 Version history
 ===============
 
+5.3.1 (2016 December 16)
```
sbesson: I think the current date would be Monday (19)

docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,14 @@
...
+-----------------------
```
sbesson: Missing dashes

docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,14 @@
...
+
+Java bug fixes:
```
sbesson: Now we decoupled the low-level components, does this section still make sense /cc @hflynn ?

docs/sphinx/about/whats-new.txt
hflynn: I think your underlining here is one character too short

docs/sphinx/about/whats-new.txt
```diff
@@ -1,6 +1,14 @@
...
+
+* TIFF
+   - fixed invalid IFD values when writing TIFF files with ``loci.formats.TiffWriter``
```
sbesson: Might be worth precising (OME-)TIFF as both formats are affected. Do we need to precise the affected utilities like `bfconvert`?

docs/sphinx/about/whats-new.txt
hflynn: I'd be inclined to replace it with file format fixes (or other as appropriate down the line)

hflynn:
👍 looks good

PR #2706, opened 2016-12-21
ND2: discard chunk map if there are obviously missing chunk map entries

melissalinkert:
This will revert to the old scanning logic for finding offsets if the chunkmap has missing entries in the middle of the map, irrespective of the order in which entries are stored in the file.  For example, a file with only two entries of ```ImageDataSeq|0``` and ```ImageDataSeq|2``` would trigger this PR's changes, but a file with ```ImageDataSeq|0```, ```ImageDataSeq|2```, and ```ImageDataSeq|1``` would not.

A message is logged at ```WARN``` if the chunkmap is being discarded in this case, so it should be easy to see in repository test logs the full list of affected files.  At a minimum, ```nd2/benjamin/B2.nd2``` should use this change and pass all tests.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -423,6 +423,9 @@ public class NativeND2Reader extends FormatReader {
 
           long chunkMapEnd = chunkMapPosition + chunkMapLength;
 
+          int imageDataCount = 0;
+          int maxImageIndex = -1;
+
           while(in.getFilePointer() + 1 + 16 < chunkMapEnd) {
 
             char b = (char) in.readByte();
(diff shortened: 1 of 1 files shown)


sbesson:
After exclusing the 2 files from `curated/nd2/guillaume`, https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-repository-subset/173 is green with this PR, including the `B2.nd2` file. @melissalinkert: should we merge this PR and investigate the filesets above separately?

melissalinkert:
Yes, let's merge this.  I have work in progress on the other two datasets (see https://github.com/melissalinkert/bioformats/tree/nd2-offset-fix), but have not yet succeeded in getting something to fix those files without breaking others.  I will pick it back up after holidays.

PR #2707, opened 2016-12-21
Metamorph TIFF: wavelength population and dimension calculation fix

melissalinkert:
Backported from private @glencoesoftware PRs.  This updates ```MetamorphHandler``` to parse all stored channel names, and fixes ```MetamorphTiffReader``` to treat two planes with unique channel names and Z positions as just 2 channels, instead of 2 channels x 2 Z sections.  It also updates ```MetamorphTiffReader``` to set ```Channel.EmissionWavelength``` when appropriate.

I don't have real data that can be copied to data_repo, but an artificial dataset that exhibits the problem is now in ```data_repo/curated/metamorph/thomas/test.tif``` (generated from ```data_repo/curated/metamorph/thomas/t_test2001.tif```).  Without this change, ```showinf test.tif``` should show SizeZ = 2 and SizeC = 2.  ```tiffinfo test.tif | grep -c "TIFF Directory"``` should confirm that there are only two planes in the file.

With this change, ```showinf test.tif``` should show SizeZ = 1 and SizeC = 2.  Both channels are expected to be the same image due to how the file was generated.

The emission wavelength addition can be verified using ```showinf -nopix -omexml``` against the list of files in https://github.com/openmicroscopy/data_repo_config/pull/163.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphHandler.java
@@ -61,6 +61,7 @@ public class MetamorphHandler extends BaseHandler {
   private Length positionX, positionY;
   private Vector<Double> exposures;
   private String channelName;
+  private Vector<String> channelNames;
   private String stageLabel;
   private Double gain;
   private boolean dualCamera = false;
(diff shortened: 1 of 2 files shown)


dgault:
The code changes here look good and the sample file does show incorrect dimension sizes prior to this PR with correct values once the PR is applied.

Running  `showinf -nopix -omexml` against the sample file with this PR also produces xml which matches against the expected values in the data repo config PR https://github.com/openmicroscopy/data_repo_config/pull/163

PR looks good to merge

PR #2708, opened 2017-01-03
QuickTime: respect openBytes' tile Y coordinate

melissalinkert:
This fixes ```openBytes``` to return the correct tile when ```y``` is greater than 0 and only one channel is present.

To test, use ```data_repo/curated/quicktime/curtis/mri-stack-native.mov```.  Without this change, compare the images shown by ```showinf mri-stack-native.mov``` and ```showinf -crop 0,113,186,113 mri-stack-native.mov```.  The second test specifies coordinates for the bottom half of the original image, but will return the top half instead.

The same test with this change should show the bottom half of the original image as expected.

https://github.com/openmicroscopy/data_repo_config/pull/165 proposes adding the test file to the list covered by https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-performance, as that would have caught this problem.  Automated tests should otherwise be unaffected.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/NativeQTReader.java
+++ b/components/formats-bsd/src/loci/formats/in/NativeQTReader.java
@@ -260,7 +260,7 @@ public class NativeQTReader extends FormatReader {
         }
       }
       else {
-        System.arraycopy(t, row*srcRowLen + x*bpp*getSizeC(), buf,
+        System.arraycopy(t, (row + y)*srcRowLen + x*bpp*getSizeC(), buf,
           row*destRowLen, destRowLen);
       }
     }


dgault:
Code change looks good and the the `showinf` test was able to reproduce the initial issue and show the correction once the PR was applied. Other tests appear unaffected.

This looks good to merge alongside the associated data repo config addition PR.

sbesson:
See also https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-breaking-performance/76/ where tests failed without this PR on the Quicktime file as expected vs https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-performance/444/ where tests passed with this PR included.

PR #2709, opened 2017-01-03
NIFTI: fix up extended metadata parsing

melissalinkert:
This fixes a few problems with byte alignment when reading non-core metadata from NIFTI headers, most notably switching from ```readChar()``` to ```readByte()```.  This also changes unit handling to use the current API instead of multiplying by a factor of 1000 to get micrometers or seconds.

To test, verify that the changes match the NIFTI specification: https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields.  Without this change, ```showinf -nopix data_repo/curated/public/NIH/minimal.hdr``` should show ```Intent name: n1``` in the original metadata - this is suspicious because it is the magic string expected in the last four bytes of the file.  With this change, the same test should not show an ```Intent name``` key, as the field is empty.

Some configurations are affected (see https://github.com/openmicroscopy/data_repo_config/pull/166), but only for physical size units (due to using the stored units in 9c1741b) and image descriptions (due to the off-by-two fixed in 2bc9e8c).  Changed values in the affected files can be confirmed with ```nifti_tool -disp_hdr -infiles /path/to/file``` (from https://sourceforge.net/projects/niftilib/) and the source page for the public datasets (https://nifti.nimh.nih.gov/nifti-1/data).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -474,6 +474,7 @@ public class NiftiReader extends FormatReader {
     addGlobalMeta("Dimension 5", dim5);
     addGlobalMeta("Dimension 6", dim6);
     addGlobalMeta("Dimension 7", dim7);
+    addGlobalMeta("Dimension 8", dim8);
     addGlobalMeta("Intent #1", intent1);
     addGlobalMeta("Intent #2", intent2);
     addGlobalMeta("Intent #3", intent3);
@@ -501,6 +502,10 @@ public class NiftiReader extends FormatReader {
     addGlobalMeta("Quaternion x parameter", quaternionX);
     addGlobalMeta("Quaternion y parameter", quaternionY);
     addGlobalMeta("Quaternion z parameter", quaternionZ);
+    addGlobalMeta("Slice code", sliceCode);
+    addGlobalMeta("XYZT units", units);
+    addGlobalMeta("XYZ units", spatialUnits);
+    addGlobalMeta("Time units", timeUnits);
   }
 
 }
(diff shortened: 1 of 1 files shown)


sbesson:
Tested using various [public NIFti samples](http://downloads.openmicroscopy.org/images/NIfTI/NIH/) by comparing the output of `showinf -omexml` with Bio-Formats 5.3.2 and with the PR included, compared the output of the metadata key/value table with the output of `nifti_tool` e.g. for [newsirp_final_XML.nii](http://downloads.openmicroscopy.org/images/NIfTI/NIH/newsirp_final_XML.nii):

- [newsirp_final_XML_niftitool.txt](https://github.com/openmicroscopy/bioformats/files/714178/newsirp_final_XML_niftitool.txt)
- [newsirp_final_XML_5.3.2.txt](https://github.com/openmicroscopy/bioformats/files/714176/newsirp_final_XML_5.3.2.txt)
- [newsirp_final_XML_5.3.3-SNAPSHOT.txt](https://github.com/openmicroscopy/bioformats/files/714182/newsirp_final_XML_5.3.3-SNAPSHOT.txt)

For all tested samples with various units and dimensionality, the metadata with this PR included matches the expectations of the output of `nifti_tool`.

For confirmation, is the replacement `readChar` by `readByte` solely due to the different in size of `char` type between Java and C?

In the context of this PR but also for future debugging, could we also expose a few missing metadata values from the header as global metadata key/value pairs like `Dimension 8/dim8`, `Slice code` and `Units`?

melissalinkert:
> For confirmation, is the replacement readChar by readByte solely due to the different in size of char type between Java and C?

Yes, exactly.

> In the context of this PR but also for future debugging, could we also expose a few missing metadata values from the header as global metadata key/value pairs like Dimension 8/dim8, Slice code and Units?

Yes, I'll push an additional commit a bit later today.

sbesson:
All tests are passing against our [NIFTI samples](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/380/) with this PR included and the global metadata is correctly populated with the new fields. Looks good to merge.

PR #2717, opened 2017-01-08
TIFF writing tests: define non-zero default percentages

sbesson:
See https://trello.com/c/WZkp5y0e/9-increase-default-percentage-of-writing-tests

- the first commit fixes a bug found when running only non-tiling tests. Without this PR `{ant,mvn} test -Dtestng.runWriterSaveBytesTests=100 -Dtestng.runWriterTilingTests=0` should fail. With 209b3ca, the non-tiling tests should be configured using the correct variable and tests should run as expected.
- the second commit sets non-zero default values for the percentage of tiling and non-tiling tests. With this commit, a subset of the the TIFF writing tests should always be executed when running the default command.

To test this PR:

- check the Travis and Jenkins builds remain green and that  `{mvn,ant} test` includes tiling and non-tiling TIFF writing tests
- check the number of Formats BSD unit tests in [BIOFORMATS-DEV-merge-build](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/) is unchanged as the properties are overriden via the job configuration
- check the logic for setting the percentage of tiling/non-tiling tests via  `{mvn,ant} test -Dtestng.runWriterSaveBytesTest=xxx -Dtestng.runWriterTilingTests=yyy` remains functional

### DIFF
--- a/components/formats-bsd/build.xml
+++ b/components/formats-bsd/build.xml
@@ -12,6 +12,8 @@ Type "ant -p" for a list of targets.
   <property name="root.dir" location="../.."/>
   <import file="${root.dir}/ant/java.xml"/>
   <property file="build.properties"/>
+  <property name="testng.runWriterSaveBytesTests" value="50"/>
+  <property name="testng.runWriterTilingTests" value="2"/>
 
   <target name="test" depends="jar,compile-tests,test-long-running,
       test-no-ome-xml,test-no-lurawave,test-no-jai,test-no-exif,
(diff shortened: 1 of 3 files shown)


dgault:
The bug fix to allow for savesBytes to be configured independently of tiling tests is working correctly and can be configured individually. The ability to set parameters via either maven or ant is also working correctly.

Both Travis and Jenkins builds have remained green and stable.

The number of units tests being run as part of https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build remains constant at 3555 since all tests were enabled (previously 5043 with only tiling). These are the correct and expected values for the full test run with 100% of parameters used.
Total tests run: 3555, Failures: 0, Skips: 0

The tests with this PR included look as expected with the bug fix and builds remain stable. As such this PR looks good to merge.

PR #2719, opened 2017-01-09
Improvision TIFF: read channel colors, if present

melissalinkert:
To test, use ```data_repo/curated/improvision/karsten/Spiro.tiff``` and ```data_repo/curated/improvision/Vessels 32-bit Col.tif```.

Without this PR, ```showinf -nopix -omexml``` on each file should show that the ```Color``` attribute on ```Channel``` is not set.  With this PR, the same test should show a valid ```Color``` set for each ```Channel```.  ```Vessels 32-bit Col.tif``` should have a single channel set to ```-1``` (white), as it is an RGB image.  ```Spiro.tiff``` should have colors that match the channel names (```green``` and ```pink```); this can be confirmed by opening in ImageJ with the color mode set to ```Colorized```.

I would expect automated tests to be unaffected by this change.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
@@ -39,6 +39,7 @@ import loci.formats.MetadataTools;
 import loci.formats.meta.MetadataStore;
 import loci.formats.tiff.TiffParser;
 
+import ome.xml.model.primitives.Color;
 import ome.xml.model.primitives.PositiveFloat;
 
 import ome.units.quantity.Length;
@@ -57,6 +58,7 @@ public class ImprovisionTiffReader extends BaseTiffReader {
 
   // -- Fields --
 
+  private ArrayList<Color> channelColors = new ArrayList<Color>();
   private String[] cNames;
   private int pixelSizeT;
   private double pixelSizeX, pixelSizeY, pixelSizeZ;
@@ -101,6 +103,7 @@ public class ImprovisionTiffReader extends BaseTiffReader {
       readers = null;
       files = null;
       lastFile = 0;
+      channelColors.clear();
     }
   }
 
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
```diff
@@ -162,31 +165,57 @@ protected void initStandardMetadata() throws FormatException, IOException {
...
+          else if (key.equals("YCalibrationMicrons")) {
+            pixelSizeY = DataTools.parseDouble(value);
+          }
+          else if (key.equals("ZCalibrationMicrons")) {
+            pixelSizeZ = DataTools.parseDouble(value);
+          }
+          else if (key.equals("WhiteColour")) {
+            String[] rgb = value.split(",");
+            if (rgb.length < 3) {
+              continue;
```
rleigh-codelibre: This needs `channelColors.add(null);` before the `continue` or else all subsequent iterations will be off by one.

components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
```diff
@@ -356,6 +385,10 @@ protected void initMetadataStore() throws FormatException {
         if (cNames != null && i < cNames.length) {
           store.setChannelName(cNames[i], 0, i);
         }
+        int index = getIndex(0, i, 0);
+        if (index < channelColors.size()) {
```
rleigh-codelibre: `if (index < channelColors.size() && channelColors.get(index) != null`?

rleigh-codelibre:
The testing showed this worked correctly.  I've made one comment about the array indexing, but I'm not sure if that's an oversight or intentional.

melissalinkert:
The array indexing problem was an oversight, as the case of fewer than 3 colors has not yet been encountered in any actual files.  ac87564 should resolve it.

rleigh-codelibre:
Thanks, with the fix this looks good to merge to me.

components/formats-gpl/src/loci/formats/in/ImprovisionTiffReader.java
```diff
@@ -162,31 +165,58 @@ protected void initStandardMetadata() throws FormatException, IOException {
...
+          }
+          else if (key.equals("WhiteColour")) {
+            String[] rgb = value.split(",");
+            if (rgb.length < 3) {
+              channelColors.add(null);
+              continue;
+            }
+            int red = 255;
+            try {
+              red = Integer.parseInt(rgb[0]);
```
sbesson: Outside the scope of this PR but this code addition raises the question: are there are other places across our scope which could benefit from an extension of our DataTools API to support `DataTools.parseInt(value, default)`?

melissalinkert: As long as the default value remains optional, I think that would be a useful addition.

sbesson:
Changes makes sense and the automated tests have remained green for the last few runs. Merging.

PR #2721, opened 2017-01-10
Photoshop TIFF: turn off layer parsing

melissalinkert:
This should prevent extra layer series from being created when reading Photoshop TIFFs, and will make it easy to turn this feature into a proper option later on.

See https://trello.com/c/D8JDP7S8/582-idr0032-tifs-being-split-into-two-images

Without this PR, ```showinf -nopix -omexml``` on one of the files should confirm the presence of two series named ```Merged``` and ```BackgroundM```.  With this PR, the same test should show a single series whose name matches the relative file path.  Importing into OMERO should also result in a single Image whose name matches the relative file path.

/cc @eleanorwilliams

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/PhotoshopTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/PhotoshopTiffReader.java
@@ -66,6 +66,9 @@ public class PhotoshopTiffReader extends BaseTiffReader {
   private int[][] channelOrder;
   private String[] layerNames;
 
+  // this can be turned into an option in 5.3+
+  private boolean readLayers = false;
+
   // -- Constructor --
 
   /** Constructs a new Photoshop TIFF reader. */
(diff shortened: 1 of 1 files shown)


joshmoore:
closing for the moment. This invalidates bfmemo.

joshmoore:
The bfmemo issue makes this a blocker for demo32 so we have some time to discuss. My larger issue here was: is this the intended behavior for the mainline?

melissalinkert:
I think this would be worth considering for the mainline, but it's not critical.  I can move ```readLayers``` to be local to ```setId``` if that's helpful for IDR in the short term.

joshmoore:
@eleanorwilliams : then likely a question for you whether or not to attempt this solution or if we keep the orphaned images. Of course, we can always use one strategy for 0.3.2 and another for 0.3.3

eleanorwilliams:
If we want it out for 0.3.2 i think we should keep the orphaned images and then we can redo for 0.3.3.

PR #2722, opened 2017-01-13
TiffWriter - Exception when writing bigTiff multi series

dgault:
An issue was reported during the testing of Bio-Formats 5.3.0 tile writing that errors were encountered when using bigTiff. 

The unit test additions here show that an Exception was indeed being thrown when writing multi series bigTiff images. In fact the same issues can be seen when writing tiles or using the standard TiffWriter. The unit test currently does not write multi series for non tiled scenarios but this can easily be modified to  add that option if required.

**The error:**
The error that was being seen was a NullPointerException when calling saveBytes (without an IFD) and could occur when attempting to read any of the following tags from the IFD when writing the image: STRIP_OFFSETS, IFD.TILE_OFFSETS, IFD.STRIP_BYTE_COUNTS, IFD.TILE_BYTE_COUNTS

**Details of what is happening:**
When calling saveBytes without an IFD provided, TiffParser is used to generate one before calling saveBytes with this IFD. These tags all appear to be correctly present on the IFD and have an associated value. When the value is parsed it is of type OnDemandLongArray. The original type of these tags is primitive long array, but when TiffWriter generates the initial IFD by parsing the file it creates a type OnDemandLongArray. 

The OnDemandLongArray uses an input stream to the current file to read the values. In this case the stream relates to the file being written and was opened during the initial call to saveBytes in order to generate the IFD. As this stream is then closed during this original call, the result is that the IFD contains an OnDemandLongArray which has a RandomAccessInputStream with a null stream which has been closed.

The fix put in place is a rather simple high level solution of keeping the underlying implementation of using OnDemandLongArray as is, but keeping the input stream open and valid during the full saveBytes operation. There are other possible solutions to this at a lower level which could also be open for consideration if this is not seen as suitable (such as populating the IFD with a primitive array, or recognising the closed stream and reopening it).

**To reproduce:**
-  Apply the changes to TiffWriterTest unit test without the TiffWriter change https://github.com/openmicroscopy/bioformats/commit/9bcfae73290f24d9c6803749fd4608f7b9a86cd3
- Set the runWriterTilingTests parameter to run the full suite using -Dtestng.runWriterTilingTests=100
- Run the tests and verify that you can observe 1024 test failures, all should be when bigTiff parameter is set to true.

**Testing:**
- Ensure builds are green and unit tests are all passing
- Set the parameters to run the full suite using -Dtestng.runWriterTilingTests=100 and -Dtestng.runWriterSaveBytesTests=100
- Apply the PR and ensure that the TiffWriterTest unit test now passes fully

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -484,7 +484,8 @@ public class TiffWriter extends FormatWriter {
           tiffParserStream.close();
         }
       }
-    } else {
+    }
+    else {
       saveBytes(no, buf, ifd, x, y, w, h);
     }
   }
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -476,6 +476,7 @@ public class TiffWriter extends FormatWriter {
         if (no < ifdOffsets.length) {
           ifd = parser.getIFD(ifdOffsets[no]);
         }
+        saveBytes(no, buf, ifd, x, y, w, h);
       }
       finally {
         RandomAccessInputStream tiffParserStream = parser.getStream();
@@ -484,8 +485,6 @@ public class TiffWriter extends FormatWriter {
         }
       }
     }
-
-    saveBytes(no, buf, ifd, x, y, w, h);
   }
 
   /* @see loci.formats.IFormatWriter#canDoStacks(String) */
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -485,6 +485,9 @@ public class TiffWriter extends FormatWriter {
         }
       }
     }
+    else {
+      saveBytes(no, buf, ifd, x, y, w, h);
+    }
   }
 
   /* @see loci.formats.IFormatWriter#canDoStacks(String) */
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -484,8 +484,7 @@ public class TiffWriter extends FormatWriter {
           tiffParserStream.close();
         }
       }
-    }
-    else {
+    } else {
       saveBytes(no, buf, ifd, x, y, w, h);
     }
   }
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -421,143 +421,147 @@ public class TiffSaver {
     boolean isTiled = ifd.isTiled();
 
     RandomAccessInputStream in = null;
-    if (!sequentialWrite) {   
-      if (filename != null) {
-        in = new RandomAccessInputStream(filename);
+    try {
+      if (!sequentialWrite) {   
+        if (filename != null) {
+          in = new RandomAccessInputStream(filename);
+        }
+        else if (bytes != null) {
+          in = new RandomAccessInputStream(bytes);
+        }
+        else {
+          throw new IllegalArgumentException(
+              "Filename and bytes are null, cannot create new input stream!");
+        }
+        TiffParser parser = new TiffParser(in);
+        long[] ifdOffsets = parser.getIFDOffsets();
+        LOGGER.debug("IFD offsets: {}", Arrays.toString(ifdOffsets));
+        if (no < ifdOffsets.length) {
+          out.seek(ifdOffsets[no]);
+          LOGGER.debug("Reading IFD from {} in non-sequential write.",
+              ifdOffsets[no]);
+          ifd = parser.getIFD(ifdOffsets[no]);
+        }
       }
-      else if (bytes != null) {
-        in = new RandomAccessInputStream(bytes);
+  
+      // record strip byte counts and offsets
+  
+      List<Long> byteCounts = new ArrayList<Long>();
+      List<Long> offsets = new ArrayList<Long>();
+      long totalTiles = tilesPerRow * tilesPerColumn;
+  
+      if (!interleaved) {
+        totalTiles *= nChannels;
+      }
+  
+      if (ifd.containsKey(IFD.STRIP_BYTE_COUNTS) ||
+        ifd.containsKey(IFD.TILE_BYTE_COUNTS))
+      {
+        long[] ifdByteCounts = isTiled ?
+          ifd.getIFDLongArray(IFD.TILE_BYTE_COUNTS) : ifd.getStripByteCounts();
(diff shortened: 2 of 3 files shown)


dgault:
The maven unit tests appear to be taking much longer than they should do. I believe this is due to a bug which that was running the saveBytes tests with the same percentage of test cases as the tiling. I shall open a separate PR to address the unit tests.

sbesson:
@dgault: agreed on the time increase which likely caused Travis to time out i.e. TiffWriter tests have been exceeding the 10min Travis timeout. So is the expectation for the Jenkins jobs that the full testing time will double as well? This raises some concern about the scalability of these matrix unit tests as we might need to cover more scenarios/API parameters. Would it make sense to isolating the bigtiff testing using only one/a few tiling/non-tiling configurations?

rleigh-codelibre:
It looks and seems fine to me, and passes all the tests, but I'll defer to @melissalinkert for any subtle details I might have missed.

One stylistic nit: `} else {` might be nicer than having the `else {` on a separate line.

melissalinkert:
Changes read fine now to me as well, thanks.

dgault:
Closing this PR for now, will reopen once the PR to fix the Travis issues has been merged to avoid any conflicts

chris-allan:
Just a quick comment that while I do agree with and prefer `} else {` vs `else {` the `else {` style is in keeping with the rest of Bio-Formats. An additional nitpick, but I would revert 47ca846.

rleigh-codelibre:
The latest change looks absolutely fine to me, thanks.

sbesson:
I reproduced the exception using a modified version of @emilroz's gist [here](https://gist.github.com/sbesson/7ccf71072ae028b214dd9448a2b131da)

I could reproduce the initial NullPointerException using the following commands and Bio-Formats 5.3.2

```
javac -cp .:/path/to/bf_5.3.2/bioformats_package.jar TiledWriter.java
java -cp .:/path/to/bf_5.3.2/bioformats_package.jar TiledWriter
```

Unfortunately, with this morning merge artifacts including this PR, the NPE persists with the following stacktrace:

```
sbesson@ls30630:1c894ce866b78cee73e96ea4dd91decf (master) $ javac -cp /Users/sbesson/Downloads/bioformats_package.jar TiledWriter.java
sbesson@ls30630:1c894ce866b78cee73e96ea4dd91decf (master) $ java -cp .:/Users/sbesson/Downloads/bioformats_package.jar TiledWriter
...
13:59:57.401 [main] DEBUG loci.common.NIOByteBufferProvider - Using mapped byte buffer? false
13:59:57.657 [main] DEBUG loci.formats.tiff.TiffSaver - Attempting to write image.
13:59:57.678 [main] DEBUG loci.formats.tiff.TiffSaver - Compressed strip 1/1 length 131072
13:59:57.678 [main] DEBUG loci.formats.tiff.TiffSaver - Attempting to write image IFD.
13:59:57.680 [main] DEBUG loci.formats.tiff.TiffSaver - IFD offsets: [16, 131784, 263552, 395320]
13:59:57.680 [main] DEBUG loci.formats.tiff.TiffSaver - Reading IFD from 16 in non-sequential write.
	at loci.common.RandomAccessInputStream.getFilePointer(RandomAccessInputStream.java:198)
	at loci.formats.tiff.OnDemandLongArray.toArray(OnDemandLongArray.java:79)
	at loci.formats.tiff.IFD.getIFDLongArray(IFD.java:408)
	at loci.formats.tiff.TiffSaver.writeImageIFD(TiffSaver.java:464)
	at loci.formats.tiff.TiffSaver.writeImage(TiffSaver.java:397)
	at loci.formats.tiff.TiffSaver.writeImage(TiffSaver.java:273)
	at loci.formats.out.TiffWriter.saveBytes(TiffWriter.java:270)
	at loci.formats.out.OMETiffWriter.saveBytes(OMETiffWriter.java:195)
	at loci.formats.out.TiffWriter.prepareToWriteImage(TiffWriter.java:331)
	at loci.formats.out.TiffWriter.saveBytes(TiffWriter.java:263)
	at loci.formats.out.OMETiffWriter.saveBytes(OMETiffWriter.java:195)
	at loci.formats.out.OMETiffWriter.saveBytes(OMETiffWriter.java:185)
	at TiledWriter.main(TiledWriter.java:71)
```

dgault:
The tiled writer sample code is going through the OMETiffWriter (which is currently not tested via automated testing) and in turn taking a different path through the TiffWriter. The exception found here appears to be a long standing issue which has only come to light with more thorough testing which is currently taking place. 

In this case it is a similar issue in that the input stream used by TiffParser is required to be kept open until the offsets and byte counts can be converted back to and stored as primitive arrays. With the latest commit it should now run successfully also.

chris-allan:
Can we implement 6fd0949 in such a way that if there's an exception following `new TiffParser(in)` we don't end up leaking the file handle? Essentially a restoration of the try-finally block?

dgault:
The diff of the last commit looks a lot worse than it is, it is simply restoring the try catch structure for closing the input stream but a large portion of the writeImageIFD method is now included in the try block.

sbesson:
See https://github.com/openmicroscopy/bioformats/pull/2722/files?diff=unified&w=1 for a diff without whitespaces. Given the differences between OMETIFFWriter and TIffWriter, I would assume the next logical step might be to activate and systematically run OME-TIFF writing unit tests for. @dgault any feeling on how to do this? For instance, could we migrate most of the code in `TiffWriterTest` into an `AbstractTiffWriterTest` class and have `TiffWriterTest` and `OMETiffWriterTest` only overriding the format specific bits ?

sbesson:
Retested using the last revision of https://gist.github.com/sbesson/7ccf71072ae028b214dd9448a2b131da (script creating a set of tiled OME-TIFF from a 1240*1024 fake image with various pixel types, bigtiff and  sequential writing parameters). With Bio-Formats 5.3.2, I can reproduce the error reported above i.e. the writing operations fails when using bigtiff and no sequential writing (default). With this PR included, tiled writing completes for all combinations of BigTiff/sequential writing.

For all OME-TIFF generated with `setWriteSequentially(true)`, the pixels seem to be incorrectly written - see below the planes 1 and 2:
screen shot 2017-01-23 at 13 46 54
screen shot 2017-01-23 at 13 46 59

I could reproduce the same erroneous behavior with Bio-Formats 5.3.2 so this is not a regression due to this PR. I suggest we get this fix in and address the ongoing writing issues in a separate PR alongside the expansion of the unit tests (see https://github.com/openmicroscopy/bioformats/pull/2733). 

@chris-allan: are you okay with https://github.com/openmicroscopy/bioformats/pull/2722/commits/c11dbb3fa8756f6a8af7a49ca628553f4720dfb6 addressing your comment?

chris-allan:
Yessir, @sbesson. I had a separate conversation with @melissalinkert regarding the need to get that code refactored so that we are not trying to debug, indent or surround 100+ lines of code with try blocks. However, that refactoring need not be done in this PR.

PR #2723, opened 2017-01-16
Also copy the version history file as NEWS.rst into the MATLAB bundle

sbesson:
See https://trello.com/c/IFSFT27n/10-add-news-to-all-archives and https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8188

This addition should mirror the packaging of the Octave and command-line tools bundles by adding the Bio-Formats version history file as a NEWS. This should allow to simplify the identification of the toolbox version without using the API until we suffix the package name with a version number.

To test this PR, check all CI jobs remain green especially the ones testing MATLAB and that the bfmatlab.zip bundle now includes a NEWS.rst file with the version history.


mtbc:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/512/artifact/artifacts/bfmatlab.zip contains `bfmatlab/NEWS.rst` with version history. MATLAB CI is green across the board, safe to dive.

PR #2724, opened 2017-01-16
Bump to 0.3.1-SNAPSHOT (in IDR group)

joshmoore:
_Not_ having this PR caused open PRs built in the local devspace cache to be treated as the official 0.3.0 release.


joshmoore:
Built this morning on idr-ci cleanly. Merging.

PR #2725, opened 2017-01-16
Add minimal framework for testing the command line tools via TestNG

sbesson:
See https://trac.openmicroscopy.org/ome/ticket/10517

## Current status

Although widely used in our daily usage and y the community, we currently lack test coverage of our Bio-Formats command-line tools. This usually implies that some combinations of options are rarely if not never tested and might be subject to regression when proposing fixes or functionalities.

## Proposal

This commit declares `org.testng` as a dependency of the `bio-formats-tools` module like the other components and provides a first test class for `ImageConverter`. As command-line tools classes have private constructors and use `System.exit()`, the security manager is overriden at set up and restored at teardown while individual methods invoke `ImageConverter.main() `directly.

## Impact and limitations

The expected outcome of this proposal is to start adding unit tests for every modification of the command line utility classes. One aspect not covered by this framework are the bash and bat scripts and the set of options used for executing the utility functions. The packaging of the JVM call might require a separate testing step. For now, limiting this testing to to the standard daily/release testing procedure might suffice as the largest number of bugs arise from the Java classes themselves.

## Testing

Primarily, check that the new tests are executed and passing using both build systems and the various continuous integration platforms. As a complete proof of concept, it might be possible to use an existing utility bug (e.g. https://trello.com/c/6e2x7HMe/14-nd2-to-tiff-conversion-error) and write a corresponding failing unit test.

/cc @dgault @melissalinkert 

## References 

http://stackoverflow.com/questions/309396/java-how-to-test-methods-that-call-system-exit

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -196,7 +196,7 @@ public final class ImageConverter {
           catch (NumberFormatException e) { }
         }
         else if (!args[i].equals(CommandLineTools.NO_UPGRADE_CHECK)) {
-          LOGGER.error("Found unknown command flag: {}; exiting.", args[i]);
+          System.out.println("Found unknown command flag: " + args[i] + "; exiting.");
           return false;
         }
       }
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -196,7 +196,7 @@ public final class ImageConverter {
           catch (NumberFormatException e) { }
         }
         else if (!args[i].equals(CommandLineTools.NO_UPGRADE_CHECK)) {
-          System.out.println("Found unknown command flag: " + args[i] + "; exiting.");
+          LOGGER.error("Found unknown command flag: {}; exiting.", args[i]);
           return false;
         }
       }
--- a/components/bio-formats-tools/build.xml
+++ b/components/bio-formats-tools/build.xml
@@ -16,7 +16,6 @@ Type "ant -p" for a list of targets.
     <testng failureProperty="failedTest">
       <classpath refid="test.classpath"/>
       <classpath>
-        <pathelement location="${root.dir}/ant/"/><!-- logback.xml -->
         <pathelement location="${test-classes.dir}"/>
         <pathelement location="${classes.dir}"/>
       </classpath>
--- a/components/bio-formats-tools/build.xml
+++ b/components/bio-formats-tools/build.xml
@@ -12,4 +12,17 @@ Type "ant -p" for a list of targets.
   <import file="${root.dir}/ant/java.xml"/>
   <property file="build.properties"/>
 
+  <target name="test" depends="jar, compile-tests" description="run tests">
+    <testng failureProperty="failedTest">
+      <classpath refid="test.classpath"/>
+      <classpath>
+        <pathelement location="${root.dir}/ant/"/><!-- logback.xml -->
+        <pathelement location="${test-classes.dir}"/>
+        <pathelement location="${classes.dir}"/>
+      </classpath>
+      <xmlfileset file="${tests.dir}/testng.xml"/>
+      <jvmarg value="-mx${testng.memory}"/>
+    </testng>
+    <fail if="failedTest"/>
+  </target>
 </project>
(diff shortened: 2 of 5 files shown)


melissalinkert:
Definitely looks like a good starting point to me, and the use of security manager is pretty clever :-)

https://travis-ci.org/openmicroscopy/bioformats/jobs/192397542 shows that 6 tests are run by Maven, but when I run ```ant clean jars tools test``` locally, it looks like the new tests aren't run.  Does ```components/bio-formats-tools/build.xml``` need to be updated too?

sbesson:
:+1: pushed a commit updating the Ant build system. Also added a reference to the PR description for the testing solution.

dgault:
With the latest commit the tests are running successfully with `ant clean jars tools test`
```
[testng] Bio-Formats Tools Tests
[testng] Total tests run: 6, Failures: 0, Skips: 0
```

This solution appears to work well and does provide the means to be able to test the tools more robustly. From playing around with it, trying to extend or break the tests, the only drawback I could find was that the ExitException does not provide much in the way of detail on the exception message etc, however the output does show any warning or error messages which would be displayed to users. As such it should still provide a good basis for testing successful behaviour and providing an initial point for debugging failures.

Overall Im happy with the test suite and think it is certainly fit for purpose for future extension and greater test coverage.

sbesson:
@dgault: agreed on the `ExitException` limitation which someway matches the behavior of `System.exit(code)` where only an error code can be passed to the terminated JVM.
Initial thoughts to overcome this limitation would be to capture the stdout/stderr using `System.{setErr/setOut}` and process the output stream as part of the test themselves (similar to e.g. http://doc.pytest.org/en/latest/capture.html). Arguably this is something we might also need to implement to test a classes `ImageInfo`.

dgault:
Agreed for ImageInfo this might be needed for a thorough test. I did some quick testing with overriding the output and error streams and that looks like it would give us the flexibility to test the generated output info if needed. Happy to leave that to a separate PR though if its not being specifically used here:

```
  private final ByteArrayOutputStream outContent = new ByteArrayOutputStream();
  private final ByteArrayOutputStream errContent = new ByteArrayOutputStream();

  @BeforeMethod
  public void setUp() {
    System.setOut(new PrintStream(outContent));
    System.setErr(new PrintStream(errContent));
    System.setSecurityManager(new NoExitSecurityManager());
  }

  @AfterMethod
  public void tearDown() {
    System.setOut(null);
    System.setErr(null);
    System.setSecurityManager(null);
  }
```

joshmoore:
Two heads-up (odd phrase that) :
 * Could a test system have set these values? If so, perhaps reset to the previous setting.
 * Are there any issues with thread-safety with our current concurrent settings?

sbesson:
Pushed a set of commits including a proof of concept for the usage of stdout in the tests as suggested in https://github.com/openmicroscopy/bioformats/pull/2725#issuecomment-273251268. The system stdout, stderr and security manager are also reset to their initial values.

Re thread safety, https://github.com/openmicroscopy/bioformats/pull/2725/commits/9d395a843205bfc7021770eba42530c3e90a7386 also uses `{After/Before}Class` for setting the security manager which should allow individual tests to be executed parallely using `threadPoolSize`. That being said my initial multi-threaded tests were not successful and there are certainly a few places in the code that might need to be reviewed in the context of thread-safety.

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -196,7 +196,7 @@ else if (args[i].equals("-tiley")) {
           catch (NumberFormatException e) { }
         }
         else if (!args[i].equals(CommandLineTools.NO_UPGRADE_CHECK)) {
-          LOGGER.error("Found unknown command flag: {}; exiting.", args[i]);
+          System.out.println("Found unknown command flag: " + args[i] + "; exiting.");
```
dgault: For me the new test is able to run successfully and check the output from LOGGER without the need to us System.out directly. Can this be reverted or is there a reason its needed here?

sbesson: Sure I can look into that. The initial reason for making the change that is logback is not configured for the tests and the default formatter includes a timestamp which was breaking the assert. I'll review how to configure logback for the tests.

dgault:
With the latest commit the tests are all running and passing using both ant and maven.
The test framework here looks good to me and in a position for us to begin expanding it.

PR #2726, opened 2017-01-16
Bug fix and update default test percentages for TiffWriterTest

dgault:
This is a follow up to https://github.com/openmicroscopy/bioformats/pull/2717

There was a bug when attempting to set the 2 test properties (percentageOfTilingTests and percentageOfSaveBytesTests) individually. The value of percentageOfTilingTests was always using the correct setting. For percentageOfSaveBytesTests, if it was set to 0 then it would use the correct setting, otherwise it was using the value from percentageOfTilingTests.

This fix allows percentageOfSaveBytesTests to be set correctly independent of percentageOfTilingTests.

The second commit updates the default number of tests run, this should mean that the Travis unit tests or default ant and maven unit tests will not time out.

To test:
- Run the unit tests with -Dtestng.runWriterTilingTests=100 and -Dtestng.runWriterSaveBytesTests=10
- verify that only 10% of the saveBytes tests are run and the full tiling set is run
- Run the unit tests with -Dtestng.runWriterTilingTests=10 and -Dtestng.runWriterSaveBytesTests=100
- verify that only 10% of the tiling tests are run and the full saveBytes set is run

Note: saveBytes has 512 tests in its full run. saveBytesTiling has 2048 in its full run

### DIFF
--- a/components/formats-bsd/build.xml
+++ b/components/formats-bsd/build.xml
@@ -12,7 +12,7 @@ Type "ant -p" for a list of targets.
   <property name="root.dir" location="../.."/>
   <import file="${root.dir}/ant/java.xml"/>
   <property file="build.properties"/>
-  <property name="testng.runWriterSaveBytesTests" value="50"/>
+  <property name="testng.runWriterSaveBytesTests" value="2"/>
   <property name="testng.runWriterTilingTests" value="2"/>
 
   <target name="test" depends="jar,compile-tests,test-long-running,
--- a/components/formats-bsd/pom.xml
+++ b/components/formats-bsd/pom.xml
@@ -165,7 +165,7 @@
             <additionalClasspathElement>${basedir}/../../ant/</additionalClasspathElement>
           </additionalClasspathElements>
           <systemPropertyVariables>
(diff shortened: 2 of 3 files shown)


sbesson:
Tested:

- locally using `mvn test -Dtest=TiffWriter` with various combinations of the two parameters controlling the percentage of tiling and non-tiling tests. The number of executed tests was consistent with the total number of tests given in the description in all scenarios.
- in the [Travis Maven build](https://travis-ci.org/openmicroscopy/bioformats/jobs/192429419) the number of Formats BSD unit tests (1231) is unchanged
- in the [CI linux build](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/512) the total number of Formats BSD unit tests (3555) is unchanged and includes the full subset of non-tiling and tiling tests
- in the [Jenkins Windows build](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build-win/488/console), the total number of Formats BSD unit test (1845) is increased by 100 as a result of this PR and includes a subset of tiling and non-tiling tests

PR #2728, opened 2017-01-17
Support current timestamp format

michaelgoelzer:
added support for timestamps of lif files created with LAS AF 3.1 or
newer


michaelgoelzer:
split off from pull request #2625

sbesson:
See also https://trac.openmicroscopy.org/ome/ticket/12793 (and the corresponding QA file with a truncated TimeStampList)

melissalinkert:
Configuration PR for the file from ticket 12793: https://github.com/openmicroscopy/data_repo_config/pull/167

This looks fine to merge from my side.  The new ```translateSingleTimestamp``` methods could perhaps be consolidated a little to reduce some of the duplicated code, but that doesn't need to block this PR.

PR #2730, opened 2017-01-17
Fix support for Metamorph TIFF plates

melissalinkert:
This follows on from https://github.com/openmicroscopy/openmicroscopy/pull/4992 to address problems reported by @pwalczysko when working with Metamorph TIFF plates, specifically ```curated/metamorph/alessandro/Metamorph-TIFFs/```.

Without this PR, ```showinf -nopix -omexml Metamorph-TIFFs/001.tif``` should result in an ```OutOfMemoryError```.  Importing into OMERO is expected to fail similarly.  With this PR, the same ```showinf``` test should show generated OME-XML without throwing an exception or resulting in validation errors.

With this PR, importing ```Metamorph-TIFFs/001.tif``` into OMERO should result in a Plate with 2 wells.  None of the imported Images should end up under ```Orphaned Images```.  Note that a valid plate name is not present in the dataset, so ```setPlateName``` is not called on the ```MetadataStore```.  https://github.com/openmicroscopy/openmicroscopy/pull/4992 is therefore not required to test this PR, as the plate name in OMERO will be the same with or without it.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphTiffReader.java
@@ -47,6 +47,7 @@ import loci.formats.tiff.TiffParser;
 
 import ome.xml.model.enums.NamingConvention;
 import ome.xml.model.primitives.NonNegativeInteger;
+import ome.xml.model.primitives.PositiveInteger;
 import ome.xml.model.primitives.Timestamp;
 
 import ome.units.quantity.Length;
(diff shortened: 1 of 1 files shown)


pwalczysko:
Did not see this PR listed on Standup today ?

dgault:
Tested this with the instructions using `showinf -nopix -omexml Metamorph-TIFFs/001.tif` and can confirm that with the PR included I was able to successfully run the command without exception and the xml was displayed correctly. Without the PR applied I was unable to run showinf without error. Importing the image also worked without any issue.

The code changes also look good and makes sense. PR looks good to merge from my perspective.

sbesson:
Should we also configure the `curated/metamorph/alessandro` fileset to be included in the daily automated builds to catch regression early?

melissalinkert:
See https://github.com/openmicroscopy/data_repo_config/pull/169.  I expect this to be quite slow to test given the size and number of files, though, so probably worth keeping an eye on job run times.  Maybe it'll be OK though now that the ```imagestream``` time is reduced?

PR #2732, opened 2017-01-19
Leica-LIF: do not store missing timestamps in the OME metadata

sbesson:
Follow-up of https://github.com/openmicroscopy/bioformats/pull/2728

In the case of stopped acquisition, the number of timestamps stored in the LIF file can be lower than the number of planes. With this commit, timestamps are parsed as Double rather than double and only non-null timestamps are stored in the OME metadata.

To test this PR, use the sample file from https://trac.openmicroscopy.org/ome/ticket/12793. After opening the image using the command-line tools, the ImageJ plugin or another Bio-Formats client, the planes corresponding to the Z-sections of some images should not contain any `DeltaT` attribute with this PR included (as opposed to extreme negative values without). The corresponding planes should also be blank.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -141,7 +141,7 @@ public class LIFReader extends FormatReader {
   private Length[] posX, posY, posZ;
   private Double[] refractiveIndex;
   private List[] cutIns, cutOuts, filterModels;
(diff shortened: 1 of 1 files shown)


dgault:
The code here looks fine, however when testing using the sample file from https://trac.openmicroscopy.org/ome/ticket/12793  (QA-10917) I was unable to see any timestamps being read either with or without the PR. Is that the correct file that I am using ok?

PR #2735, opened 2017-01-23
Review the developers section (gradle, maven)

sbesson:
See https://trello.com/c/M7U9L70L/11-document-developer-changes-re-poms and https://github.com/openmicroscopy/bio-formats-examples/pull/4

This PR updates the two sections of the developers documentation mentioning the usage of Bio-Formats as a Java libary as follows:

- the usage of Bio-Formats as a library is removed from the page describing how to build Bio-Formats from source
- the outdated Maven & Gradle examples are removed in favor of the ones maintained and tested in https://github.com/openmicroscopy/bio-formats-examples
- the initial paragraph about using Bio-Formats as a library is rewritten and directly links to the https://github.com/openmicroscopy/bio-formats-examples as examples of consuming Bio-Formats using Gradle or Maven
- a note is added about the Unidata repository for the `netcdf` dependency


docs/sphinx/developers/java-library.txt
```diff
@@ -1,39 +1,25 @@
...
+Examples of getting started with Bio-Formats using Maven or Gradle are given
+in the https://github.com/openmicroscopy/bio-formats-examples repository.
+`OMERO <http://github.com/openmicroscopy/openmicroscopy>`_ uses Ivy to manage
+its Java dependencies including Bio-Formats.
+
+.. note::
+  In order to retrieve the NetCDF dependency of Bio-Formats, it is necessary
+  to configure the Unidata releases repository in addition to Maven Central and
+  the OME artifactory in your :file:`pom.xml`, :file:`build.gradle` or
+  :file:`ivy.xml` file.
```
dgault: Would it be appropriate to have an example of configuring the Unidata repo here?

sbesson: In the case of Maven and Gradle, https://github.com/openmicroscopy/bio-formats-examples/blob/master/pom.xml#L32 and https://github.com/openmicroscopy/bio-formats-examples/blob/master/build.gradle#L8 should include the declaration. I am a bit unsure about linking to a specific file line as nothing will break if there is a change but I could certainly link at the file URL. Thoughts?

dgault: I think even just linking to the example files URL would be fine, anyone configuring repos should be able to read and locate the correct lines ok.

dgault:
In general these changes look fine and the rewritten docs read well. The use of artifactory to generate the correct settings works well and is easy and straight forward for a user to follow so this seems like a good improvement over the hard coded sample code.

dgault:
The new URL links for the pom and gradle examples are both working and linking correctly.
Changes look good to merge.

PR #2737, opened 2017-01-24
ImageConverter - Invalid image index bug fix

dgault:
This is a bug in ImageConverter which saw `IllegalArgumentException` thrown with `Invalid image index` when valid parameters had been entered.

This is caused when the image converter command line tool was used with a parameter such as -z or -channel, but not -series. In the case of a specific channel or z plane being requested the writers metadata is configured to override the number of z planes or channels with the value 1. When -series was being used the behaviour was correct as the writer made a new metadata object from the readers values and overwrote them. Without -series however the readers metadata was being overwritten which then resulted in the Exception.

Trac ticket for issue: https://trac.openmicroscopy.org/ome/ticket/13316#ticket
Related forum thread: https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8182&p=17852#p17852

To reproduce:
- Using sample file QA-17475 run a conversation via the command line tools
- Use a -z, -channel or -timepoint option with a valid value (for this sample file -z works well for testing)
- Verify that an `IllegalArgumentException` is thrown

Testing:
- With the PR applied reproduce the test above and verify that the conversion is successful

### DIFF
--- a/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
+++ b/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
@@ -37,6 +37,8 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.security.Permission;
+import java.util.ArrayList;
+import java.util.Arrays;
 
 import loci.formats.IFormatReader;
 import loci.formats.ImageReader;
@@ -44,6 +46,7 @@ import loci.formats.ImageWriter;
 import loci.formats.FormatException;
 import loci.formats.tools.ImageConverter;
 
+import org.apache.commons.lang.ArrayUtils;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.BeforeMethod;
@@ -103,6 +106,12 @@ public class ImageConverterTest {
     return new Object[][] {{".ome.tiff"}, {".tif"}, {".ics"}};
   }
 
+  @DataProvider(name = "options")
+  public Object[][] createOptions() {
+    return new Object[][] {{"-z 2"}, {"-series 0 -z 2"}, {"-channel 1"}, 
+      {"-series 0 -channel 1"}, {"-series 0 -timepoint 3"}, {"-timepoint 3"}};
+  }
+
   public void checkImage() throws FormatException, IOException {
     IFormatReader r = new ImageReader();
     r.setId(outFile.getAbsolutePath());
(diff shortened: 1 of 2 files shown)


bramalingam:
Works as expected.Looks good to merge.

Compared:
BioFormats 5.3.2 : 
`[Nikon ND2] -> ./test/test20.tiff [Tagged Image File Format]
Exception in thread "main" java.lang.IllegalArgumentException: Invalid image index: 60/3
	at loci.formats.FormatTools.getZCTCoords(FormatTools.java:641)
	at loci.formats.FormatTools.getFilename(FormatTools.java:1124)
	at loci.formats.FormatTools.getFilename(FormatTools.java:1101)
	at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:537)
	at loci.formats.tools.ImageConverter.main(ImageConverter.java:880)`

And,
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/lastSuccessfulBuild/artifact/artifacts/

`[Nikon ND2] -> ./test/test20.tiff [Tagged Image File Format]
	Converted 3/3 planes (100%)`

sbesson:
@dgault is it possible to extend the coverage of `ImageConverterTest` to provide test cases for these options as well?

dgault:
I can certainly extend `ImageConverterTest` to cover some different option combinations. Do you want it as part of this PR or as a separate follow up?

bramalingam:
It would be great if we could have it as part of this PR itself. That way everything gets dealt with at the same time. (Functionality fix and the test associated with it). What do you feel @sbesson ?

dgault:
Updated with the unit tests. In writing the unit tests I additionally found a separate but similar bug when `-series` is being used. Running the tests without https://github.com/openmicroscopy/bioformats/pull/2737/commits/21e9d239c35202c65d3feb5c1fc848f4fb95e616 should result in 3 failures, but with the commit the 6 new tests should all pass.

components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
```diff
@@ -124,6 +133,25 @@ public void testDefault(String suffix) throws FormatException, IOException {
     }
   }
 
+  @Test(dataProvider = "options")
+  public void testOptions(String options) throws FormatException, IOException {
+    outFile = File.createTempFile("test", ".ome.tiff");
+    outFile.delete();
```
bramalingam: Could we add if `File.exists{ ... }` condition over here to make sure its easier to understand the logic here. Given that this logic has been used in other tests as well, this PR can go ahead in its current state and we could handle similar cases of this logic, in a separate PR.

bramalingam:
Works as expected. Looks good to merge. @dgault Could you please add a card for the File.delete logic alone, as discussed.

components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
sbesson: @bramalingam the original design comes from https://github.com/openmicroscopy/bioformats/pull/2725 where these unit tests where introduced. The idea is to generate a unique temporary file path per test without creating the file else `bfconvert` hangs.
The crudest solution I could find was to use `File.createTempFile` and then delete the file. There might be a more elegant way to do that using a third party dependency guava /cc @mtbc

bramalingam: Ahhh ok that makes sense.! Would be great if we could add a line of comment in such cases. It would help other developers to understand the reason behind this implementation as well (given that this is not intuitive while reading through the code). Happy to be corrected as well,just a suggestion.

mtbc: I'm not aware of anything great from Guava: but what's the requirement? For instance, should the File instance correspond to a writable path even if it's not written? Or could it be any old nonsense so long as it's unique?

sbesson: In this context, I definitely want a unique writable path as this should be passed as the output file of the converter class.

mtbc: Hmmm. If you're having to do a lot of them, don't know if you could have test setup create a unique *directory* then just create all the filenames in that by incrementing a counter or something. (Guava gives you `createTempDir`.)

PR #2738, opened 2017-01-27
TiffSaver - Enable tiled sequential writing

dgault:
This PR is a follow up to https://github.com/openmicroscopy/bioformats/pull/2722

To reproduce:
- Run a read write conversation with write `setWriteSequentially(true)` (for example you could modify one of the examples such as the new OverlappedTiledWriter https://github.com/dgault/bioformats/blob/fdac59c81b01f519b291714d38433b87fc2ba7a9/docs/sphinx/developers/examples/OverlappedTiledWriter.java)
- Verify that when opening the image that the image does not display correctly (tiles should all be the same)

To test:
- With the PR applied rerun the above test to verify that the image is converted successfully
- Ensure that the test works for different image types, eg various dimensions present, interleaved/non interleaved, different pixel types
- Ensure that builds are green and all unit tests are passing

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -80,6 +80,10 @@ public class TiffSaver {
   /** Whether or not to write BigTIFF data. */
   private boolean bigTiff = false;
   private boolean sequentialWrite = false;
+  
+  /** Store tile offsets and original file pointer when writing sequentially. */
+  private List<Long> sequentialTileOffsets;
+  private Long sequentialTileFilePointer;
 
   /** The codec options if set. */
   private CodecOptions options;
@@ -419,6 +423,7 @@ public class TiffSaver {
     int tilesPerColumn = (int) ifd.getTilesPerColumn();
     boolean interleaved = ifd.getPlanarConfiguration() == 1;
     boolean isTiled = ifd.isTiled();
+    long defaultByteCount = 0L;
 
     RandomAccessInputStream in = null;
     try {
@@ -443,6 +448,9 @@ public class TiffSaver {
           ifd = parser.getIFD(ifdOffsets[no]);
         }
       }
+      else if (isTiled) {
+        defaultByteCount = strips[0].length;
+      }
   
       // record strip byte counts and offsets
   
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
```diff
@@ -419,6 +423,7 @@ private void writeImageIFD(IFD ifd, int no, byte[][] strips,
     int tilesPerColumn = (int) ifd.getTilesPerColumn();
     boolean interleaved = ifd.getPlanarConfiguration() == 1;
     boolean isTiled = ifd.isTiled();
+    long defaultByteCount = 0L;
 
     RandomAccessInputStream in = null;
```
jburel: you may want to use ``ry-with-resources `` https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html
to avoid finally and make the code easier to read.
Also ``stripBuf`` does not seem to be closed.

jburel:
pr-2738
16-bit example with sequential on (left) and off (right)

jburel:
pr-2738-8bit
8bit image

jburel:
with PR
pr-2738

jburel:
with PR
pr-2738-8bit

jburel:
without the PR
pr-2738-no-fix

components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
jburel: other resources do not seem to be closed in ``writeIFD``
This could be fixed in another PR

sbesson:
@jburel: is this good to merge according to the last screenshots or is more testing time required?

jburel:
The writing issue is fixed.
The resources closure can (should) be fixed in a follow up PR. 
several input streams are not correctly handled

sbesson:
Thanks @jburel. Merging. Can you add a card to the Inbox of the [Bio-Formats 5.3.x board](https://trello.com/b/IBHfAIMP/bio-formats-5-3-x)?

dgault:
I've added a card to review the closure and handling of resources - https://trello.com/c/q6elgNFw/88-tiffsaver-review-resources

PR #2739, opened 2017-01-28
Leica LIF: add option to preserve pre-5.3.3 physical sizes

melissalinkert:
Follow up to https://github.com/openmicroscopy/bioformats/pull/2727.

This adds a new ```leicalif.old_physical_size``` boolean option to make physical X and Y size calculation backwards compatible with 5.3.2 and earlier.  By default, the option is set to ```false```.

To test, choose any Leica .lif file from ```data_repo/curated/leica-lif```.  Record the values of ```PhysicalSizeX``` and ```PhysicalSizeY``` shown in the OME-XML produced by ```showinf -nopix -omexml```, both with 5.3.2 and with develop not including this PR.

With this PR, ```showinf -nopix -omexml``` on the same file should result in ```PhysicalSizeX``` and ```PhysicalSizeY``` values that match the recorded output from develop.  With this PR and ```showinf -nopix -omexml -option leicalif.old_physical_size true```, ```PhysicalSizeX``` and ```PhysicalSizeY``` should match the recorded values from 5.3.2.

A corresponding ImageJ checkbox has also been added.  To test, install a ```bioformats_package.jar``` that includes this PR into ImageJ.  Open any series from any Leica .lif file, and note the physical image width and height at the top of the image window. Then click ```Plugins > Bio-Formats > Bio-Formats Plugin Configuration```, select ```Leica Image File Format``` from the list, then tick the ```Ensure physical sizes...``` box.  Close the configuration window and open the same series from the same file as before.  The physical image width and height should be slightly different.  You can also compare the pixel width and height shown in the image properties window (```Ctrl+Alt+p```), but those are typically rounded for display and may not 100% match what ```showinf``` reports.

I am not entirely convinced that the option name is clear, so happy to change it if anyone has a better idea.

PR #2740, opened 2017-01-28
Add documentation for new Leica LIF metadata option

melissalinkert:
See https://github.com/openmicroscopy/bioformats/pull/2739

If we decide to change the option name, this will need to be updated.

I do wonder if we need a separate page that covers the options in general - I didn't see anything existing apart from Javadocs.  I can start something here if desired, or add a 5.3.x card.

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -1212,7 +1212,20 @@ Commercial applications that support LIF include: \n
 \n
 * `Bitplane Imaris <http://www.bitplane.com/>`_ \n
 * `SVI Huygens <http://svi.nl/>`_ \n
-* `Amira <http://www.amira.com/>`_
+* `Amira <http://www.amira.com/>`_ \n
+\n
+Versions of Bio-Formats prior to 5.3.3 incorrectly calculated the physical \n
+pixel width and height.  The physical image width and height were divided by \n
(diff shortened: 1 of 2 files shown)


hflynn:
👍 Reads fine to me and renders correctly on staging.

We are probably approaching needing an additional page for this sort of info on top of marking things on the format pages - 'Format-specific additional options'? or similar under the user information section?

sbesson:
We certainly have at least 2 use cases already (`zeissczi` + `leicalif`) and the list of use cases might be growing. Working on some minimal documentation to the developers documentation in addition to the Javadoc certainly sense to me but I would not count this as a 5.3.3 blocker.

We might need some design/discussion on how to best represent these options for each format. I like the idea of a Format options section. It someway overlaps with the concept of options introspection which we have not formalized i.e. right now we have implicit naming convention to store the options keys and default values but no description. From an information point-of-view, is there anything else that would would be added to a table in addition to Name/Default/Description:

| Option name | Default | Description |
|----------------|----------|--------------|
| `leicalif.old_physical_size` | `true` |  Defines whether the pixel size... |

hflynn:
Agreed that extra page is not a blocker. We can debate user vs dev docs. I'd probably stick the file format in the table as well just because it's likely quicker for people to parse than having to read the whole option name.

PR #2741, opened 2017-01-31
ExampleSuite - clean existing output files

dgault:
In the ExampleSuite it was possible to see NullPointerExceptions when run multiple times without cleaning the directory. An example of this can be seen in https://travis-ci.org/openmicroscopy/bioformats/jobs/196902278

This fix will remove any existing output files if they already exist before beginning writing.
I opted for this approach rather than deleting files on exit as this is part of the ExampleSuite rather than a unit test and as such users may not wish to have output files removed immediately if they are using this as a base.

To reproduce:
- With the PR included run a clean Maven install using `mvn clean install`
- This should complete successfully without error
- Now rerun without cleaning using `mvn install`
- Verify that a NullPointerException is seen in the output

to testy:
- Rerun the above commands and verify that no exceptions are observed in the output

### DIFF
--- a/docs/sphinx/developers/examples/ExampleSuite.java
+++ b/docs/sphinx/developers/examples/ExampleSuite.java
@@ -31,6 +31,7 @@
  */
 
 import java.net.URL;
+import java.nio.file.Files;
 import java.io.File;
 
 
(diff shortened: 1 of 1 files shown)


sbesson:
All makes sense to me. I could reproduce the error as described and the deletion of the files prior to running the individual examples fixes the issue as confirmed either locally or by Travis build.

For the deletion strategy, the initial intent was certainly to preserve the generated files available after executing the examples. In that sense, the proposed approach using `Files.deleteIfExists` respects this contract. 

Merging to restore Travis and allow #2472 to be built.

PR #2744, opened 2017-02-01
Java9 General service loading

jburel:
This PR introduces a change required to use Bio-Formats with OMERO
when Java 1.9 is used
This should not change the behaviour with previous versions of Java.

In Java 9, ``javax.imageio.spi.ServiceRegistry`` checks if the requested service is an implementation of the Image I/O service provider interface. This module used the Image I/O service registry for loading external utility services like logging and caching. This is not an intended use and is now prohibited by the Java 9 runtime. javax.imageio.spi.ServiceRegistry uses 'java.util.ServiceLoader' internally.

Detailed of the error
https://trello.com/c/Eh7p5387/18-to-watch-java-1-9

Check that tests are green (especially svs)

### DIFF
--- a/components/forks/jai/src/com/sun/media/imageioimpl/common/ImageUtil.java
+++ b/components/forks/jai/src/com/sun/media/imageioimpl/common/ImageUtil.java
@@ -107,6 +107,8 @@ import java.util.Locale;
 
 //import javax.imageio.ImageTypeSpecifier;
 
+import java.util.ServiceLoader;
+
 import javax.imageio.IIOException;
 import javax.imageio.IIOImage;
 import javax.imageio.ImageReadParam;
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
Could we also open this against https://github.com/jai-imageio/jai-imageio-core ?

jburel:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-full-repository/429/ is green

jburel:
"Java 9" server now running at https://10.0.51.143:8443/
Without the PR, the server cannot start up

rleigh-codelibre:
Full repository is green, I was waiting on the windows tests to go green as well but they are still running.  I'm not really an expert on any of this service stuff, but it looks OK by the testing results so far.

sbesson:
@jburel was there any format-specific testing to do on this PR?

jburel:
I tested it with svs as suggested by @melissalinkert

sbesson:
Understood and thanks for the full-stack test including import. What do you reckon in terms of strategy? Is it worth merging and getting out immediately in a patch release of Bio-Formats or are we still too far from the Java 9 GA and we should rather leave this open for a couple of weeks/months until Java 9 stabilizes and we are certain of the changes to apply across or code base? /cc @mtbc

mtbc:
For these specific changes I don't see any reason to wait/hold.

sbesson:
Thanks both. Merging.

PR #2747, opened 2017-02-02
Pyramid TIFF: support multiple planes and OME-XML comments

melissalinkert:
Backports recent changes to ```PyramidTiffReader``` to allow multiple planes in each file and an OME-XML comment to determine Z/C/T dimensions and supply additional metadata.

To test, use the new ```data_repo/curated/pyramid-tiff/samples/multi-channel-4D-series.tiff``` file (see also https://github.com/openmicroscopy/data_repo_config/pull/173).  Verify that ```showinf``` detects the reader as ```PyramidTiffReader``` and that the Z, C, and T sizes and ordering matches the text in each plane.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/PyramidTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/PyramidTiffReader.java
@@ -168,7 +168,7 @@ public class PyramidTiffReader extends BaseTiffReader {
     int seriesCount = ifds.size() / nPlanes;
 
     String comment = ifds.get(0).getComment();
-    if (comment != null && comment.length() > 0) {
+    if (comment != null && comment.length() > 0 && comment.trim().startsWith("<")) {
       try {
         ServiceFactory factory = new ServiceFactory();
         OMEXMLService service = factory.getInstance(OMEXMLService.class);
--- a/components/formats-gpl/src/loci/formats/in/PyramidTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/PyramidTiffReader.java
@@ -26,18 +26,32 @@
 package loci.formats.in;
 
 import java.io.IOException;
+import java.util.List;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import loci.common.RandomAccessInputStream;
+import loci.common.services.DependencyException;
+import loci.common.services.ServiceException;
+import loci.common.services.ServiceFactory;
 import loci.formats.CoreMetadata;
 import loci.formats.FormatException;
 import loci.formats.FormatTools;
+import loci.formats.MissingLibraryException;
+import loci.formats.meta.IMetadata;
+import loci.formats.meta.MetadataConverter;
 import loci.formats.meta.MetadataStore;
+import loci.formats.services.OMEXMLService;
+import loci.formats.services.OMEXMLServiceImpl;
 import loci.formats.tiff.IFD;
 import loci.formats.tiff.PhotoInterp;
 import loci.formats.tiff.TiffParser;
+
+import ome.xml.meta.OMEXMLMetadataRoot;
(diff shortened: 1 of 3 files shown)


components/formats-gpl/src/loci/formats/in/PyramidTiffReader.java
```diff
@@ -52,6 +66,8 @@
 
   // -- Fields --
 
+  private IMetadata omexml = null;;
```
mtbc: double semicolon

dgault:
Without this PR it read using the OMETiffReader with the following planes:
	Plane #0 <=> Z 0, C 0, T 0
	Plane #50 <=> Z 0, C 1, T 3
	Plane #51 <=> Z 1, C 1, T 3
	Plane #52 <=> Z 2, C 1, T 3
	Plane #53 <=> Z 3, C 1, T 3
	Plane #54 <=> Z 4, C 1, T 3
	Plane #104 <=> Z 4, C 2, T 6

With the PR applied it correctly read with the PyramidTiffReader and the following planes:
	Plane #0 <=> Z 0, C 0, T 0
	Plane #50 <=> Z 0, C 1, T 3
	Plane #51 <=> Z 1, C 1, T 3
	Plane #52 <=> Z 2, C 1, T 3
	Plane #53 <=> Z 3, C 1, T 3
	Plane #54 <=> Z 4, C 1, T 3
	Plane #104 <=> Z 4, C 2, T 6

Dimension order = XYZCT

The dimensions haven't changed but the sizes and ordering are correct according to the text on the planes when the image is viewed.

The associated config PR is also showing the correct values and the build is green: https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/385/

dgault:
With that last commit this PR should be good to merge

rleigh-codelibre:
Regarding @sbesson's comments:

- For the C++ code, there is no ImageReader; `ome-files info` uses the OMETIFFReader exclusively, so I wouldn't be too concerned about the C++ detection here--until we implement ImageReader, there is no detection of the format *at all*.

Regarding OME-TIFF, supporting subresolutions there would be great, but I think that work should be done in the model (if needed) and the OME-TIFF reader and writer, treating this as a separate format unrelated to OME-TIFF.  As mentioned in the last meeting, there are several ways to do subresolutions, and for OME-TIFF using existing approaches like the [SUBIFD](http://www.awaresystems.be/imaging/tiff/tifftags/subifds.html) tag; using this would allow readers to continue to read OME-TIFFs without any subresolution support since it's both optional and backward compatible.  The Pyramid TIFF reader relies on specific ordering of IFDs and as such is not a drop-in solution.

That said, I don't think we would object to third-party formats embedding OME-XML--it was after all one of the goals of the data model, and so I don't think our reuse of OME-XML in a custom format is bad per se.  But I do think we need to be explicit about this being a totally separate file format which is completely unrelated to and incompatible with the OME-TIFF specification.  And I think at a minimum we should have a clear documented specification for the format, including exactly how the IFDs are laid out.  How exactly are multiple series and multiple dimensions and the subresolutions of each laid out?  Which IFDs are referred to by the TiffData blocks in the embedded OME-XML?

melissalinkert:
New design issue started here for discussion: https://github.com/openmicroscopy/design/issues/74
That outlines the plane ordering assumptions made here, and proposes some possible next steps both for making pyramids a formal part of OME-TIFF and making these hybrid pyramid TIFFs more obviously a distinct format.  If you all would prefer to discuss here, I can can close the issue and paste the text here - whatever is easier.

sbesson:
As this PR still requires some design and will not make it into the upcoming Bio-Formats 5.4.0, closing in favor of the [design issue](openmicroscopy/design#74) where we can pursue the discussion.

PR #2749, opened 2017-02-03
Documentation: OMERO extlink cleanup

sbesson:
Initially discussed with @jburel, the `omerodoc` extlink in the Bio-Formats documentation forces us to maintain a mapping with the version of a consumer application like OMERO. Since the decoupling of both components, this reverse mapping does not make sense e.g. a stable Bio-Formats series could be either consumed in either none or multiple stable series of OMERO.

This PR redirects all links to the OMERO documentation to the product page. In addition the unused Jenkins extlinks are also removed.

The documentation builds should remain green and the new links should point to the correct OMERO page.

### DIFF
--- a/docs/sphinx/Makefile
+++ b/docs/sphinx/Makefile
@@ -24,10 +24,6 @@ ifdef SOURCE_USER
 ANT_SOURCE_USER := -Dsphinx.source.user="$(SOURCE_USER)"
 endif
 
-ifdef OMERODOC_URI_JOB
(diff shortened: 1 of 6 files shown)


hflynn:
👍 Looks good to me - links work as expected on staging, build is green and I agree that there is no value in having docs links over this product page in this context. Good to merge from me.

hflynn:
I will look into a similar effort for the OMERO docs.

PR #2698, opened 2016-12-14
Read refractive index from MetaMorph headers (#1)

mabruce:
* Read refractive index from header

* Set refractive index in MetamorphReader

PR #2750, opened 2017-02-06
Metamorph: clean up instrument graph for datasets with multiple series

melissalinkert:
Backported from private PR.  For Metamorph datasets with multiple series, this should reduce size of the generated OME-XML/OMERO object graph by using a single ```Instrument``` for every ```Image```, instead of one ```Instrument``` per ```Image```.

I tested primarily with ```data_repo/curated/hms/HCT116.nd```, which has 32 series.  Without this PR, ```showinf -nopix -omexml HCT116.nd``` should show 32 ```Instrument```s, only the first of which has a ```Detector```.  Each ```Image``` should be linked to the ```Instrument``` with the same index (so only the first ```Image``` has a valid linked ```Detector```).

With this PR, the same test should show a single ```Instrument``` with a single ```Detector```, which is linked to every ```Image```.

It would also be good to perform the same pair of tests with ```data_repo/curated/qa-16907``` to verify that the changes from https://github.com/openmicroscopy/bioformats/pull/2698 haven't been broken by this.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -726,20 +726,20 @@ public class MetamorphReader extends BaseTiffReader {
 
     MetadataStore store = makeFilterMetadata();
     MetadataTools.populatePixels(store, this, true);
+
+    int nextObjective = 0;
+    String instrumentID = MetadataTools.createLSID("Instrument", 0);
     String detectorID = MetadataTools.createLSID("Detector", 0, 0);
+
+    store.setInstrumentID(instrumentID, 0);
+    store.setDetectorID(detectorID, 0, 0);
+    store.setDetectorType(getDetectorType("Other"), 0, 0);
 
     for (int i=0; i<getSeriesCount(); i++) {
       setSeries(i);
       handler = new MetamorphHandler(getSeriesMetadata());
 
-      String instrumentID = MetadataTools.createLSID("Instrument", i);
-      store.setInstrumentID(instrumentID, i);
       store.setImageInstrumentRef(instrumentID, i);
-
-      if (i == 0) {
-        store.setDetectorID(detectorID, 0, 0);
-        store.setDetectorType(getDetectorType("Other"), 0, 0);
-      }
 
       String comment = getFirstComment(i);
       if (comment != null && comment.startsWith("<MetaData>")) {
@@ -783,7 +783,7 @@ public class MetamorphReader extends BaseTiffReader {
       else {
         List<Double> zPositions = new ArrayList<Double>();
         final List<Double> uniqueZ = new ArrayList<Double>();
-    	  
+
         for (IFD ifd : ifds) {
           MetamorphHandler zPlaneHandler = new MetamorphHandler();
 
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/MetamorphReader.java
```diff
@@ -871,12 +874,13 @@ else if (firstSeriesChannels != null) {
 
           if ((int) wave[waveIndex] >= 1) {
             // link LightSource to Image
+            int laserIndex = i * getEffectiveSizeC() + c;
```
sbesson: As there is now only instrument/detector per fileset, should the DetectorSettings object setters above be updated in a similar manner by creating a `detectorSettingsIndex` using the series/channel and calling `setDetectorSettingsxxx(value, 0, settingsIndex)`?

melissalinkert: No, I don't think so.  The indexes on the ```setDetectorSettings*``` methods are for ```Image``` and ```Channel```, not ```Instrument```, so I think what is here will continue to work as expected.

sbesson: You're right (<a href="http://static.javadoc.io/org.openmicroscopy/ome-xml/5.4.0/ome/xml/meta/MetadataStore.html#setDetectorSettingsBinning-ome.xml.model.enums.Binning-int-int-">setDetectorSettingsBinning</a> and <a href="http://static.javadoc.io/org.openmicroscopy/ome-xml/5.4.0/ome/xml/meta/MetadataStore.html#setDetectorSettingsReadOutRate-ome.units.quantity.Frequency-int-int-">setDetectorSettingsReadOutRate</a> for external references). No need to change these callss

mtbc:
Couldn't find `data_repo/curated/qa-16907` so using `apache_repo/16907/`.

mtbc:
Outwith this PR but in metadata we have `showinf` show things like,
```
X position for position #1: ome.units.quantity.Length: value[-5276.3], unit[reference frame] stored as java.lang.Double
```
maybe these could be made prettier.

mtbc:
For the QA-16907 data we have the expected single,
```xml
  <Instrument ID="Instrument:0">
    <Detector ID="Detector:0:0" Type="Other"/>
    <Objective ID="Objective:0:0" LensNA="0.2"/>
  </Instrument>
```
and the image has,
```
    <InstrumentRef ID="Instrument:0"/>
    <ObjectiveSettings ID="Objective:0:0" RefractiveIndex="1.0"/>
```
so we don't seem to have broken #2698.

mtbc:
Using `data_repo/curated/metamorph/hms/HCT116.nd` (I guess the directories got rearranged a bit) I do indeed see the expected single instrument,
```xml
  <Instrument ID="Instrument:0">
    <Detector ID="Detector:0:0" Type="Other"/>
  </Instrument>
```
and each of the 32 images indeed has,
```xml
    <InstrumentRef ID="Instrument:0"/>
```

mtbc:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/387/ looks good. **Good to merge.**

PR #2751, opened 2017-02-06
Zeiss ZVI: set valid bits per pixel and channel colors

melissalinkert:
Backported from a private PR.  This ensures that the ```bitsPerPixel``` field in ```CoreMetadata``` is always set to match the final pixel type, and populates any channel colors that were parsed in the ```MetadataStore```.

The bits per pixel change should only affect ```uint16``` or ```int16``` files where the acquisition bit depth in the global metadata table is not a multiple of 8, and the RGB channel count is greater than 1.  A suitable test file would be ```data_repo/curated/zeiss-zvi/dan/adultNr1.zvi```, but there may be others.

Without this PR, confirm that ```showinf -nopix -omexml``` on the chosen file shows a pixel type of ```uint16``` and valid bits per pixel of 8.  The ```Color``` attribute on ```Channel``` should not be present in the OME-XML.  With this PR, the same test should show that the valid bits per pixel is 16, and that a ```Color``` is present on each ```Channel```.  The color value should match the color shown in Zen or AxioVision; in the case of ```adultNr1.zvi``` this is white (-1).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/BaseZeissReader.java
+++ b/components/formats-gpl/src/loci/formats/in/BaseZeissReader.java
@@ -347,7 +347,7 @@ public abstract class BaseZeissReader extends FormatReader {
           store.setChannelEmissionWavelength(emWavelength.get(c), s, i);
           store.setChannelExcitationWavelength(exWavelength.get(c), s, i);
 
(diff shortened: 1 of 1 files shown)


dgault:
Without the PR applied

Issue #2752, opened 2017-02-09
Bio-Formats Exporter throws NullPointerException


imagejan:
To reproduce, run the following macro in Fiji:

```
run("Mitosis (26MB, 5D stack)");
run("Bio-Formats Exporter", "save=[C:\\temp\\test.ics]");
```

the exception thrown is:

```
(Fiji Is Just) ImageJ 2.0.0-rc-56/1.51h; Java 1.8.0_112 [64-bit]; Windows 7 6.1; 331MB of 96000MB (<1%)
 
java.lang.NullPointerException
	at loci.plugins.out.Exporter.run(Exporter.java:443)
	at loci.plugins.LociExporter.run(LociExporter.java:75)
	at ij.plugin.filter.PlugInFilterRunner.processOneImage(PlugInFilterRunner.java:265)
	at ij.plugin.filter.PlugInFilterRunner.<init>(PlugInFilterRunner.java:114)
	at ij.IJ.runUserPlugIn(IJ.java:219)
	at ij.IJ.runPlugIn(IJ.java:181)
	at ij.Executer.runCommand(Executer.java:137)
	at ij.Executer.run(Executer.java:63)
	at ij.IJ.run(IJ.java:297)
	at ij.macro.Functions.doRun(Functions.java:601)
	at ij.macro.Functions.doFunction(Functions.java:96)
	at ij.macro.Interpreter.doStatement(Interpreter.java:230)
	at ij.macro.Interpreter.doStatements(Interpreter.java:218)
	at ij.macro.Interpreter.run(Interpreter.java:115)
	at ij.macro.Interpreter.run(Interpreter.java:85)
	at ij.macro.Interpreter.run(Interpreter.java:96)
	at ij.plugin.Macro_Runner.runMacro(Macro_Runner.java:155)
	at ij.IJ.runMacro(IJ.java:134)
	at ij.IJ.runMacro(IJ.java:123)
	at net.imagej.legacy.IJ1Helper$3.call(IJ1Helper.java:1039)
	at net.imagej.legacy.IJ1Helper$3.call(IJ1Helper.java:1035)
	at net.imagej.legacy.IJ1Helper.runMacroFriendly(IJ1Helper.java:986)
	at net.imagej.legacy.IJ1Helper.runMacro(IJ1Helper.java:1035)
	at net.imagej.legacy.plugin.IJ1MacroEngine.eval(IJ1MacroEngine.java:97)
[...]

PR #2754, opened 2017-02-09
ImageJ: fix NPE when check pixel type

melissalinkert:
This prevents exceptions when exporting images that were not opened via the Bio-Formats importer, and thus do not have a complete OMEXMLMetadata store.

See https://github.com/openmicroscopy/bioformats/issues/2752

Without this PR, running the macro listed on gh-2752 (on Windows) or the following on Mac/Linux should result in the ```NullPointerException``` noted on the issue:

```
run("Mitosis (26MB, 5D stack)");
run("Bio-Formats Exporter", "save=[/tmp/test.ics]");
```

With this PR (and after removing the output file), the same test should result in a completed export with no error messages.  This bug was introduced by https://github.com/openmicroscopy/bioformats/pull/2696

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -439,10 +439,14 @@ public class Exporter {
             // this prevents problems if the user changed the bit depth of the image
             boolean applyCalibrationFunction = false;
             try {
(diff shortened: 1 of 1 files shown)


dgault:
Running the macro with version 5.3.0 runs without error. Then running ti with version 5.3.3 produces the reported NullPointerException:

PR #2758, opened 2017-02-16
ScreenReader: exclude selected readers

simleo:
Adds support for a new `ExcludeReaders` entry in the `Plate` section that allows to exclude some readers from the list considered by `ImageReader`.

Example:

```
[Plate]
Name = Foo
Rows = 8
Columns = 12
Fields = 1
ExcludeReaders = loci.formats.in.Reader1,loci.formats.in.Reader2
```

This is useful in cases where a multi-file format is detected (e.g., due to the presence of specific TIFF tags) which would override the desired arrangement described in the file patterns.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ScreenReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ScreenReader.java
@@ -78,6 +78,7 @@ public class ScreenReader extends FormatReader {
   // -- Fields --
 
   private String[] plateMetadataFiles;
+  private String[] excludeReaders;
   private String[][] files;
   private String[] ordering;
   private int[][] axisTypes;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/ScreenReader.java
```diff
@@ -78,6 +78,7 @@
   // -- Fields --
 
   private String[] plateMetadataFiles;
+  private String[] excludeReaders;
```
joshmoore: This will likely invalidate the bfmemo cache and if so could not be introduced in a patch release.

simleo: Hmm, OK. Too bad because it appears to fix idr0025 problems described in https://github.com/IDR/idr-metadata/pull/211.

joshmoore: The solution in general makes sense, but the introduction of a new field can often cause problems. (Worth testing)

simleo:
Removed introduction of new field in 7176e83. I've tried generating a memo file without the PR and then reusing it with the PR and it worked.

joshmoore:
Merging for inclusion in the IDR/Bio-Formats 0.3.1 build. (Tested on idr0025 in test34)

PR #2759, opened 2017-02-17
merge RGB intensities using NDPIReaders

mstritt:
Here we use NDPIReaders and merge the RGB intensities instead of using ChannelSeparators as before.
Otherwise the images have been black for most  (non red) channels.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -129,13 +129,13 @@ public class NDPISReader extends FormatReader {
     // each channel is RGB data (with usually only one band used), thus we sum up the intensities
     if (readers[channel].isInterleaved()) {
       for (int i = 0; i < buf.length; i++) {
-        intens = bufRGB[i * 3] + bufRGB[i * 3 + 1] + bufRGB[i * 3 + 2];
+        intens = ((int)bufRGB[i * 3] & 0xff) + ((int)bufRGB[i * 3 + 1] & 0xff) + ((int)bufRGB[i * 3 + 2] & 0xff);
         buf[i] = (byte) (intens <= 255 ? intens : 255);   // clamp to byte
       }
     } else {    // not interleaved
       int offs = w*h;
       for (int i = 0; i < buf.length; i++) {
-        intens = bufRGB[i] + bufRGB[i + offs] + bufRGB[i + offs *2];
+        intens = ((int)bufRGB[i] & 0xff) + ((int)bufRGB[i + offs] & 0xff) + ((int)bufRGB[i + offs *2] & 0xff);
         buf[i] = (byte) (intens <= 255 ? intens : 255);   // clamp to byte
       }
     }
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -222,7 +222,7 @@ public class NDPISReader extends FormatReader {
       store.setChannelEmissionWavelength(new Length(wavelength, UNITS.NANOMETER),getSeries(), c);
 
       bandUsed[c] = 0;
-      if (readers[c].getSizeC()>=3) {
+      if (samplesPerPixel[c]>=3) {
         // define band used based on emission wavelength
         // wavelength = 0  Colour Image
         // 380 =< wavelength <= 490 Blue
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -122,9 +122,7 @@ public class NDPISReader extends FormatReader {
         }
       } else {    // not interleaved
         final int offs = w * h;
-        for (int i = 0; i < buf.length; i++) {
-          buf[i] = bufReader[i + offs * band];
-        }
+        System.arraycopy(bufReader,offs * band,buf,0,buf.length);
       }
       return buf;
     }
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -30,6 +30,9 @@ import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.formats.*;
 import loci.formats.meta.MetadataStore;
+import loci.formats.tiff.IFD;
+import ome.units.UNITS;
+import ome.units.quantity.Length;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -48,7 +51,10 @@ public class NDPISReader extends FormatReader {
 
   private String[] ndpiFiles;
   private NDPIReader[] readers;
-  private final static int CHANNELTAG = 65434;
+  private int[] bandUsed;
+  private int[] samplesPerPixel;
+  private final static int TAG_CHANNEL = 65434;
+  private final static int TAG_EMISSION_WAVELENGTH = 65451;
 
   // -- Constructor --
 
@@ -102,24 +108,26 @@ public class NDPISReader extends FormatReader {
     readers[channel].setId(ndpiFiles[channel]);
     readers[channel].setSeries(getSeries());
     readers[channel].setResolution(getResolution());
-    byte[] bufRGB = DataTools.allocate(w, h, 3); // w*h*RGB  - maybe this should be reused...
-    bufRGB = readers[channel].openBytes(0, bufRGB, x, y, w, h);
-    int intens;
-    // each channel is RGB data (with usually only one band used), thus we sum up the intensities
-    if (readers[channel].isInterleaved()) {
-      for (int i = 0; i < buf.length; i++) {
-        intens = ((int)bufRGB[i * 3] & 0xff) + ((int)bufRGB[i * 3 + 1] & 0xff) + ((int)bufRGB[i * 3 + 2] & 0xff);
-        buf[i] = (byte) (intens <= 255 ? intens : 255);   // clamp to byte
-      }
-    } else {    // not interleaved
-      int offs = w*h;
-      for (int i = 0; i < buf.length; i++) {
-        intens = ((int)bufRGB[i] & 0xff) + ((int)bufRGB[i + offs] & 0xff) + ((int)bufRGB[i + offs *2] & 0xff);
-        buf[i] = (byte) (intens <= 255 ? intens : 255);   // clamp to byte
-      }
-    }
+    int spp = samplesPerPixel[channel];
+    if (spp==1) return readers[channel].openBytes(0, buf, x, y, w, h);    // single band reader
 
-    return buf;
+    else {   // read intensity from used band (the other bands are close to zero, only jpeg artifacts and should be ignored)
+      byte[] bufReader = DataTools.allocate(w, h, spp); // w*h*RGB
+      bufReader = readers[channel].openBytes(0, bufReader, x, y, w, h);
+      int band = bandUsed[channel];
+      // each channel is RGB data (with usually only one band used), thus we sum up the intensities
+      if (readers[channel].isInterleaved()) {
+        for (int i = 0; i < buf.length; i++) {
+          buf[i] = bufReader[i * spp + band];
+        }
+      } else {    // not interleaved
+        final int offs = w * h;
+        for (int i = 0; i < buf.length; i++) {
+          buf[i] = bufReader[i + offs * band];
+        }
+      }
+      return buf;
+    }
   }
 
 
@@ -179,6 +187,7 @@ public class NDPISReader extends FormatReader {
       if (key.equals("NoImages")) {
         ndpiFiles = new String[Integer.parseInt(value)];
         readers = new NDPIReader[ndpiFiles.length];
+
       }
       else if (key.startsWith("Image")) {
         int index = Integer.parseInt(key.replaceAll("Image", ""));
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thank you for opening a pull request, @mstritt.

In an initial review, it seems like adding a ```loci.formats.ChannelMerger``` to the calling reader's stack (see http://www.openmicroscopy.org/site/support/bio-formats5.3/developers/file-reader.html#file-reading-extras) would have the same effect as the switch from ```ChannelSeparator``` to ```NDPIReader``` here.

Before we review more thoroughly, could you please describe a use case that requires this change, and if possible provide a sample NDPIS dataset that we can use to test it?  Files can be uploaded to http://qa.openmicroscopy.org.uk/qa/upload/

mstritt:
Dear  @melissalinkert ,
I uploaded a sample file (test3.ndpis). When importing to Omero I only see the FITC channel + overview image but DAPI and TRITC is skipped.
The new reader should solve this issue.

Using a ChannelMerger might be an option, too. However, be aware that the under-laying channels are RGB images, thus ChannelMerger.canMerge() might fail.
 Actually not just the channels have to be merged, but the R,G,B intensities of each channel have to be summed up and the result of this sum for each channel has to be merged.

bramalingam:
Hi,

Thank you for uploading your files to our QA server.

We are in the process of investigating them in better detail.
We will get back to you in the next couple of days with more information on the same.

Best,
Balaji

dgault:
The functionality on the sample file provided behaves as described. Tests on the existing sample files also pass : https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/392/

Some feedback I would have on some improvements for the code:
- The new openBytes could be removed as it is identical to the FormatReader method which it is overriding.

- When calculating the intensities there are a number of hard coded values and assumptions about the data, for example  `intens = ((int)bufRGB[i * 3] & 0xff) + ((int)bufRGB[i * 3 + 1] & 0xff) + ((int)bufRGB[i * 3 + 2] & 0xff); `. Perhaps using some of the utility functions such as `ImageTools.make24Bits` would allow for similar behaviour without having to assume the channelCount or bytesPerPixel. An example of this would be from the [autoscalePlane method in the ImageConverter](https://github.com/openmicroscopy/bioformats/blob/develop/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java#L829-L843)

mstritt:
- openBytes removed as proposed
- the the ImageTools.make24Bits: as far as I see it does not sum up and clamp to byte the values. plus we would have an additional memory allocation (new int[w*h]). ChannelCount should be fix 3 here.

sbesson:
Hi @mstritt, sorry for the delay. We are in the process of reviewing the latest changes in this PR and their impact and we'll comment here shortly. Two preliminary questions: can the sample file you uploaded (see https://github.com/openmicroscopy/bioformats/pull/2759#issuecomment-281015692) be made public or does it contain sensitive data? And/or can we paste screenshots in this public PR to discuss the various scenarios?

mstritt:
Hi @sbesson, you can make the sample and screenshots of it public, no problem.

sbesson:
Hi @mstritt,

First and foremost, thanks for granting permission to make these files public. The data is now available at [this URL](http://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/) and we will be using screenshots made using  this data in this comment.

To make sure, we all agree on the expectations and the current behavior, this is how the ndpis image looks using the acquisition software:

37c2ba20-05a6-11e7-80ee-05443b1ce8fc

Importing the fileset into a development OMERO 5.3 server gives the following in the Web client:

screen shot 2017-03-13 at 11 26 47

The NDPIs fileset is currently detected as a single image containing 3 channels with one Samples per Pixel each. This representation matches the state of our current data repository where in all our similar NDPIS samples, only the Red component of the first channel, the Blue component of the second channel and the Green component of the third channel have valid pixel values. This explains the current design originally implemented in  [this commit](https://github.com/openmicroscopy/bioformats/commit/b3236c1d8f2db34955c171aab6b16cf96d4f9b80).

These public samples provide a new category of raw data (is this a new version of the software or the format?) where the assumption that the 3 channels are pure Red, Green and Blue images does not hold anymore. The proposal behind this PR is to address the current limitation by summing the values of the R, G , B components for each channel i.e. create 3 grayscale channels.

With this PR included, the same fileset imported into an OMERO server looks like:

screen shot 2017-03-13 at 11 25 42

There are a few issues associated with this proposal:

- Unconditionally summing the RGB components of each channel still results in lost Information as it is not possible to come back to the original pixel values
- The simple sum conversion does not use any weighted values as usually done when converting RGB to grayscale. Moreover, this conversion has mutiple implementations varies depending on the methodology).
- Finally, as shown in the snapshots, the channel color would require to be mapped instead of using the classical RGB colors for rendering (which leads to the DAPI channel being red).

As they are many implementation of the RGB to grayscale conversion, the way Bio-Formats handles similar situations with multiple RGB channels (as done for other file formats e.g. Zeiss ZVI) is to create images with multiple channels with 3 samples per pixels each. In OMERO, this would imply the imported image would contain 9 logical channels. It is then the responsibility of the consumers or clients to consume these raw data channels and process them as they need for rendering or analysis.

The addition of the channel names metadata using the tag is completely orthogonal and looks sensible. Can we split this metadata addition into a separate PR for potential inclusion in the upcoming Bio-Formats 5.4.0? I propose we create an RFE ticket/card for handling the proper handling of these multi RGB channel images.

mstritt:
Dear @sbesson,
thanks for the detailed analysis. 
- indeed we have 9 logical channels here. However, usually only one on the R,G,B samples per channel is really used. E.g. DAPI will use blue.  (But yes, in theory the others could be used as well I guess.)
For this reason the simple summing up of R,G,B, is a pragmatic approach and works fine for all the sample
images I have.
- the rendering / OMERO interpretation should be treated independent of the reader. Of course it could be more than 3 channels. Orbit for instance reads the channel name and has a mapping channel name -> hue value.  For this reason the channel name is very important.
- I'm not sure, but other (weighted) RGB->Grayscale conversions might falsify the measured intensities.

sbesson:
Hi @mstritt 
 
> the rendering / OMERO interpretation should be treated independent of the reader

Completely agreed. In order to decouple the rendering, it is crucial that Bio-Formats faithfully returns the metadata and raw pixel data stored on disk.

> For this reason the simple summing up of R,G,B, is a pragmatic approach and works fine for all the sample images I have.

While it provides a workaround for the bug/limitation highlighted by this PR, my core impression is that the channel averaging leads to another version of the same limitation rather than a proper long-term fix. As shown by our [data repository pixel tests](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/198), the implementation of this PR is already breaking for the reading  of existing datasets. Exposing the community to multiple rounds of breaking changes is certainly something we want avoid as much as possible so I would rather try to define and move towards a long-term solution. Happy to read other's opinions as well /cc @melissalinkert @bramalingam

> I'm not sure, but other (weighted) RGB->Grayscale conversions might falsify the measured intensities.

There are definitely a wide range of color to grayscale conversion formulas which all rely on some assumption (see resources like [the Wikipedia page](https://en.wikipedia.org/wiki/Grayscale) or [this blog](http://www.tannerhelland.com/3643/grayscale-image-algorithm-vb6/). Averaging channels is one of the possible conversions where the weights are effectively identical for each component.

At the Bio-Formats level, an immediate action is certainly to document the assumptions behind the NDPIs reader as well as  the current limitations highlighted by this PR. /cc @hflynn.
Also from your description, I understand you would like to see this change propagated into OMERO eventually? We are currently freezing the code for the next minor & major releases of Bio-Formats 5.4.0 and OMERO 5.3.0. It would be good for us to record all the expectations if we want to plan and test in our upcoming milestones.

mstritt:
Hi @sbesson,
thanks again for your detailed analysis.  It's unclear to me how the breaking datasets have been generated - probably by an older version I guess. Of course it would be nice to have a reader which works for all NDPI(S) versions.
A mid-term Bioformats and Omero support for the (let's call it 'new') NDPIS files would be great and I'd highly appreciate this.

mstritt:
Hi @sbesson, 
after consulting the manufacturer the situation is as follows:
For all current scanners: each linked ndpi image is technically a RGB image.
Two kind of scanners:
- Type1: the intensities are written into one band based on the wavelength (e.g. DAPI goes into blue). The other bands are close to zero, except some jpeg compression artifacts which can be ignored.
- Type2: the intensities are copied into all bands (so 3 bands with the same values)

Conclusion: we have no 'logical channels' and it is ok to 'merge' the RGB images to one gray-scale image. 

I updated the reader so that it works under the mentioned assumptions. 
(I tried also to handle non RGB images (samplesPerPixel=1) but could not test it. Not sure where you got these images from - but the proposed reader should work with all newer NDPIS files.)

sbesson:
Hi @mstritt,

many thanks for driving the interaction with the manufacturer. The last comments about the expected behavior are extremely useful. To clarify the question about our existing NDPIS filesets, prior to this PR, we only had one, composed of 2 RGB NDPI files. Echoing your comment, I am not certain about how to handle non-RGB images or whether there is a valid use case.

The new proposed logic to properly detect the used band looks good. As shown by our [latest non-regression test](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/207/), this solution would also be fully backwards compatibility for NDPIS filesets which are the current reader assumption (i.e. using the red component for the  first channel, the greenchanne == G component) while fixing the generic behavior for arbitrary NDPIS filesets. This sounds perfectly reasonable as a proposed format fix.

We have a few comments specific to the implementation. Let us know when you are done with your latest round of changes and we can give feedback via code review. On our side, we will also update the configuration of our NDPIS filesets to match the additional metadata and pixel hashes populated by this PR and update our non-regression tests.

mstritt:
Hi @sbesson,
thanks for the feedback, I'm done so far, please go ahead with the code review.

components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -25,31 +25,36 @@
...
 import loci.common.DataTools;
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
-import loci.formats.ChannelSeparator;
-import loci.formats.CoreMetadata;
-import loci.formats.FormatException;
-import loci.formats.FormatReader;
-import loci.formats.FormatTools;
-import loci.formats.MetadataTools;
+import loci.formats.*;
```
mstritt: done

components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -25,31 +25,36 @@
...
 
   // -- Fields --
 
   private String[] ndpiFiles;
-  private ChannelSeparator[] readers;
+  private NDPIReader[] readers;
+  private int[] bandUsed;
+  private int[] samplesPerPixel;
+  private final static int TAG_CHANNEL = 65434;
+  private final static int TAG_EMISSION_WAVELENGTH = 65451;
```
mstritt: I see the point, so I'd recommend to include it for the next minor version inc.

components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -102,14 +108,27 @@ public int getOptimalTileHeight() {
...
+      if (readers[channel].isInterleaved()) {
+        for (int i = 0; i < buf.length; i++) {
+          buf[i] = bufReader[i * spp + band];
+        }
+      } else {    // not interleaved
+        final int offs = w * h;
+        System.arraycopy(bufReader,offs * band,buf,0,buf.length);
+      }
+      return buf;
+    }
```
mstritt: probably this would work, too. I'm not aware about the details of the ChannelSeparator.
(however, this solution seems to have less overhead / to be faster to me.)

components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -186,8 +206,35 @@ else if (key.startsWith("Image")) {
       ms.imageCount = ms.sizeC * ms.sizeZ * ms.sizeT;
     }
 
+    samplesPerPixel = new int[ndpiFiles.length];
+    bandUsed = new int[ndpiFiles.length];
     MetadataStore store = makeFilterMetadata();
+    for (int c=0; c<readers.length; c++) {     // populate channel names based on IFD entry
+      if (c>0) // 0 is already open
+        readers[c].setId(ndpiFiles[c]);
```
mstritt: ok

components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -186,8 +206,35 @@ else if (key.startsWith("Image")) {
...
+      IFD ifd = readers[c].getIFDs().get(0);
+      samplesPerPixel[c] = ifd.getSamplesPerPixel();
```
mstritt: RGBChannelCount is 1, but we have 3 samples per channel

components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -25,31 +25,36 @@
 
 package loci.formats.in;
 
-import java.io.IOException;
-import java.util.ArrayList;
```
mstritt: ok

components/formats-gpl/src/loci/formats/in/NDPISReader.java
sbesson: Thanks for the clarification. Including this change in a minor version increment certainly makes sense.

components/formats-gpl/src/loci/formats/in/NDPISReader.java
sbesson: For reference, the code is at https://github.com/openmicroscopy/bioformats/blob/v5.4.0/components/formats-bsd/src/loci/formats/ChannelSeparator.java#L164 (and uses `ImageTools.splitChannels()`). From a maintenance perspective, we still recommend to use/expand a built-in API as this decreases the number of implementations of the same functionality and it might cover additional use cases.

That being said, an overhead related to using the `ChannelSeparator` would be a legitimate concern that should be looked into. Do you have any initial metrics for the execution time of `openBytes` using both approaches?

bramalingam:
Works as expected. Looks good to merge.

Tested two folders,

/ome/data_repo/curated/hamamatsu/public/manuel,
screen shot 2017-04-25 at 18 56 13

The image got imported as expected.

and,
/ome/data_repo/curated/hamamatsu/philippe/fluorescence
checked for the existence of extra metadata in the Channels,

`ls29010:bftools-2 bramalingam$ ./showinf -nopix -omexml-only /Users/bramalingam/Desktop/ndpis/fluorescence/green\ redZ.ndpis `

and found,

`         <Channel EmissionWavelength="613.0" EmissionWavelengthUnit="nm" ID="Channel:0:0" Name="TRITC" SamplesPerPixel="1">
            <LightPath/>
         </Channel>
         <Channel EmissionWavelength="517.0" EmissionWavelengthUnit="nm" ID="Channel:0:1" Name="FITC" SamplesPerPixel="1">`

mstritt:
sorry for not having more time to investigate the ChannelSeparator approach. We're using the current solution here and it works fine. However, any solution is fine for me. Go ahead! ;-)

sbesson:
@mstritt: thanks. As shown in https://github.com/openmicroscopy/bioformats/pull/2759#issuecomment-297113941, the functional testing of this PR looks good. T additional channel metadata is populated and the correct components are read for each channel as expected. We have updated the configuration files for our Hamamatsu NDPIs samples to match the changes proposed in this PR and the [repository tests](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/513/) are all passing.

At this point, I do not see a reason to block this PR. I'll merge it and include the bug fixes and metadata features into the upcoming 5.5.0 release of Bio-Formats. We will likely open a follow-up PR to investigate the `ChannelSeparator` usage in `openBytes`. Thanks for your contribution!

PR #2764, opened 2017-02-20
CellVoyager: metadata fixes and general maintainability improvements

melissalinkert:
See https://trello.com/c/XA5EtcYa/101-fix-plate-naming-for-cellvoyager

CellVoyager datasets were previously untested in OMERO or by regular CI builds.  This adds a plate name as referenced on the card (in 6568998 specifically), but also fixes a number of issues that prevented automated testing and import into OMERO.

It may be easiest to review the diff with ```w=1```, as the final commit (b7de1e1) changes all of the many tabs to spaces for consistency with other readers.  I resisted making more extensive whitespace changes to fully bring the reader into line, but can do so here or elsewhere if desired.

https://github.com/openmicroscopy/data_repo_config/pull/177 adds all 3 known CellVoyager datasets to ```data_repo/curated/cellvoyager```.  As these are previously untested, it would be best to open each in ImageJ or with ```showinf```, checking images and OME-XML for any oddities.  Each should also be imported into OMERO to verify the plate name change, but note that the ```1_3_1_2_1``` dataset is not a plate (the other two datasets are plates).  For all datasets, the target file is ```MeasurementResult.xml```.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellVoyagerReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellVoyagerReader.java
@@ -149,6 +149,8 @@ public class CellVoyagerReader extends FormatReader
 		final AreaInfo area = well.areas.get( areaIndex );
 		final MinimalTiffReader tiffReader = new MinimalTiffReader();
 
+    imageFolder = new Location(currentId).getAbsoluteFile().getParentFile();
+    imageFolder = new Location(imageFolder, "Image");
 		for ( final FieldInfo field : area.fields )
 		{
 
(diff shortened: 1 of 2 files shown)


dgault:
The actual code changes here look good and the reader is much improved because of it.
 
Automated tests for the new files and config are all passing: https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/390/console

When opened with Fiji the metadata looks correct and the plate name is set (except for the 1_3_1_2_1 dataset which is expected).

Using OMERO I was able to import all 3 datasets with the PR applied while import failed previously without the PR.  (the 1_3_1_2_1 ends up as an orphaned image)

PR #2766, opened 2017-02-22
Formats-BSD: Initial warnings cleanup

dgault:
As part of this I have completed an initial pass of the formats-bsd component and reviewed any warnings.

- Remove unused imports
- Remove some unused variables
- Provide parameters for some generic types

The one exception is the last commit https://github.com/openmicroscopy/bioformats/commit/1d2c4387feb28827572ac492825171b7070cc17f which also provides a small fix to upgradeChecker as a success check was not being used previously.

To test:
All builds and tests should remain green

### DIFF
--- a/components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java
+++ b/components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java
@@ -201,7 +201,7 @@ public class JPEG2000Codec extends BaseCodec {
         out.close();
       }
       catch (IOException e) {
-        throw new FormatException("Could not compress JPEG-2000 data.", e);
+        throw new FormatException("Failed to close RandomAccessOutputStream.", e);
       }
     }
 
--- a/components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java
+++ b/components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java
@@ -38,7 +38,6 @@ import java.awt.image.DataBufferByte;
 import java.awt.image.DataBufferUShort;
 import java.awt.image.WritableRaster;
 import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 
 import loci.common.ByteArrayHandle;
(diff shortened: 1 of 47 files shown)


joshmoore:
Two general comments:

 * For the simple `a = open(); method(a); a.close();` sections I'd be for moving to https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html where we can
 * Some extra caution might be called for adding `<?>` to interfaces & public signatures, etc. cc: @mtbc

dgault:
Agreed that modifying the public signatures for the generics is probably in need of more thought and discussion rather than the intention of this PR which is to remove some basic warnings. Reverting the commit for those cases.

sbesson:
All looks good and included in the last round of merge builds successfully. @joshmoore: are you happy with the latest state of this PR?

components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java
```diff
@@ -190,14 +189,21 @@ else if (j2kOptions.bitsPerSample == 32) {
...
     }
     catch (ServiceException e) {
       throw new FormatException("Could not compress JPEG-2000 data.", e);
     }
+    finally {
+      try {
+        out.close();
+      }
+      catch (IOException e) {
+        throw new FormatException("Could not compress JPEG-2000 data.", e);
```
joshmoore: This exception seems counter-intuitive.

joshmoore:
One minor change request.

The non-`finally` [MemoryCacheImageInputStream.close()](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/javax/imageio/stream/MemoryCacheImageInputStream.java#MemoryCacheImageInputStream.close%28%29) calls are unfortunate, but that class isn't `Closeable`.

joshmoore:
Did closing the MCIIS instances causes a problem? Sorry if I caused confusion: I didn't mean that the close was bad, just that it was unfortunate we couldn't use the new resource closing that we've used elsewhere.

dgault:
Ah that was just a misunderstanding on my part then. There were no issues from closing the MCIIS instances as far as I'm aware. I shall re-include them.

rleigh-codelibre:
All looks reasonable to me.  Jobs are are green and seems fine to merge.

PR #2767, opened 2017-02-22
Formats-GPL: Initial warnings cleanup

dgault:
As part of this I have completed an initial pass of the formats-gpl component and reviewed any warnings.

- Remove unused imports
- Remove some unused variables
- Provide parameters for some generic types

To test:
All builds and tests should remain green


melissalinkert:
Will the removal of variables in ```CellH5Reader```, ```GatanReader```, ```LeicaHandler```, and ```OpenlabReader``` require memo files to be regenerated?  Definitely :+1: to cleaning those up, I just wonder if it requires a minor release?

sbesson:
Interesting thought raised in https://github.com/openmicroscopy/bioformats/pull/2767#issuecomment-282797131. A few comments:

- although we have no official enforcement re serialization as per out current [policy](https://www.openmicroscopy.org/site/support/bio-formats5.3/about/index.html#versioning-policy), incrementing the minor version number for such a code-wide cleanup is a sensible idea
- in terms of timelines, we should certainly aim at encapsulating the largest number of serialization breakages before we release OMERO 5.3.0 /cc @jburel 

I would suggest we review the state of this work on Wednesday afternoon and discuss our various options.

sbesson:
Discussed https://github.com/openmicroscopy/bioformats/pull/2767#issuecomment-282797131 today and we'll indeed call this a minor version increment. Anything else to raise @melissalinkert or are we good to go with this PR?

melissalinkert:
All fine with me.

PR #2769, opened 2017-02-27
Add missing Image-Instrument ref in GatanReader

sbesson:
Fixes https://trac.openmicroscopy.org/ome/ticket/12939

To test this PR in addition to the automated builds. open the file mentioned in the ticket with the command-line tools `showinf -omexml-only` and check the `Image` element has an `InstrumentRef` with this PR and none without.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -272,7 +272,9 @@ public class GatanReader extends FormatReader {
         }
       }
 
-      store.setInstrumentID(MetadataTools.createLSID("Instrument", 0), 0);
+      String instrument = MetadataTools.createLSID("Instrument", 0);
+      store.setInstrumentID(instrument, 0);
+      store.setImageInstrumentRef(instrument, 0);
 
       String objective = MetadataTools.createLSID("Objective", 0, 0);
       store.setObjectiveID(objective, 0, 0);


mtbc:
I wonder if some alternative underlying design of the metadata management would have made this bug impossible.

mtbc:
Without, then with, this PR,
```
 instrument |          name          
------------+------------------------
            | biosamplefullframe.dm3
         51 | biosamplefullframe.dm3
```

sbesson:
[Repository jobs](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/456/) are green with this included. Merging.

Re https://github.com/openmicroscopy/bioformats/pull/2769#issuecomment-282999982, the issue is obviously that the previous situation (unlinked `Instrument`) is perfectly legit as per our specification. My general feeling is that there is group of extra sanity checks that we should initially define before implementing them at the `model/metadata` level probably. Certainly something to bear in mind in the context the extended metadata work as a potential use of the `OMEModel` object.

mtbc:
As a point of interest: OMERO's graphs operations manage to keep these objects together by noting that the instrument is *indirectly* linked to the image via the settings. (I think maybe it's the [ObjectiveSettings](http://www.openmicroscopy.org/site/support/omero5.2/developers/Model/EveryObject.html#omero-model-class-objectivesettings) it uses.)

PR #2771, opened 2017-02-28
Zeiss CZI: make large uncompressed images more performant

melissalinkert:
This speeds up pixel data access when tiles from a large uncompressed image with no internal tiling are requested via openBytes.  Only the specific tile specified in the call to openBytes is read from disk, instead of the entire image being read and then copied.

See https://trello.com/c/YIaXQGvp/105-investigate-zeiss-czi-pyramid-generation-performance and https://github.com/openmicroscopy/data_repo_config/pull/179

Automated tests should pass using the configuration PR both with and without this PR - the actual pixel data should be the same regardless.

Performance testing was done by iterating over all tiles in all planes, using the optimal tile width and height reported by the reader as a guide.  Specifically, something like this:

```
reader.setId(filename);
int optimalX = reader.getOptimalTileWidth();
int optimalY = reader.getOptimalTileHeight();
int x = reader.getSizeX();
int y = reader.getSizeY();

for (int i=0; i<reader.getImageCount(); i++) {
   for (int yy=0; yy<y; yy+=optimalY) {
      for (int xx=0; xx<x; xx+=optimalX) {
         int w = (int) Math.min(optimalX, x - xx);
         int h = (int) Math.min(optimalY, y - yy);
         reader.openBytes(i, xx, yy, w, h);
      }
    }
}
```

For each change measured, the test code was run three consecutive times against ```data_repo/curated/zeiss-czi/michael-porter/EB01 - Airyscan Fast SR 20x_Airyscan Processing.czi```; the best of the three times was used for comparison and is reported here.

Without this PR, the test code ran in 111m22s.  With this PR, the test code ran in 4m38s.  With this PR and a now-discarded change to set the optimal tile size to 1024x1024, the test code ran in 53m34s.

In addition to verifying that builds pass and observed openBytes performance is consistent with the above, it would be good to spot check a few tiles in ImageJ to make sure that the tiles match up with the corresponding location in the full image (i.e. the file seeking arithmetic is all still correct).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -444,6 +444,9 @@ public class ZeissCZIReader extends FormatReader {
                 s.close();
               }
             }
+            else {
+              emptyTile = false;
+            }
             break;
           }
         }
@@ -3392,9 +3395,14 @@ public class ZeissCZIReader extends FormatReader {
       int bpp = FormatTools.getBytesPerPixel(getPixelType());
       if (directoryEntry.compression == UNCOMPRESSED) {
         if (buf == null) {
-          buf = new byte[tile.width * tile.height * bpp * getRGBChannelCount()];
+          buf = new byte[(int) dataSize];
         }
-        readPlane(s, tile.x, tile.y, tile.width, tile.height, buf);
+        if (tile != null) {
+          readPlane(s, tile.x, tile.y, tile.width, tile.height, buf);
+        }
+        else {
+          s.readFully(buf);
+        }
         return buf;
       }
 
@@ -3439,6 +3447,10 @@ public class ZeissCZIReader extends FormatReader {
         case 504: // camera-specific packed pixels
           data = decode12BitCamera(data, options.maxBytes);
           break;
+      }
+      if (buf != null && buf.length >= data.length) {
+        System.arraycopy(data, 0, buf, 0, data.length);
+        return buf;
       }
       return data;
     }
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -434,7 +434,7 @@ public class ZeissCZIReader extends FormatReader {
             else {
               rawData = new SubBlock(plane).readPixelData();
             }
-            if (rawData.length > buf.length) {
+            if (rawData.length > buf.length || pixels.size() > 0) {
               RandomAccessInputStream s = new RandomAccessInputStream(rawData);
               try {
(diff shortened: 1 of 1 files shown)


sbesson:
Is there also a case for reviewing these changes in the context of pyramid generation in OMERO or would the performance scripts in this PR description cover this workflow? /cc @kennethgillen @jburel

melissalinkert:
The above test code should largely simulate the calls to openBytes that occur during pyramid generation, but it couldn't hurt to test importing into OMERO as well (keeping in mind that the files are ~6 GB and ~24 GB).

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -425,14 +425,24 @@ else if (rawData.length == (realX + 1) * (realY + 1) * pixel) {
...
+            if (pixels.size() == 0) {
+              rawData = new SubBlock(plane).readPixelData(in, new Region(x, y, w, h), buf);
+            }
+            else {
+              rawData = new SubBlock(plane).readPixelData();
             }
-            finally {
-              s.close();
+            if (rawData.length > buf.length) {
+              RandomAccessInputStream s = new RandomAccessInputStream(rawData);
```
jburel: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html could be used here to avoid the ``try/finally``

sbesson:
See NPEs in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/391 and https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-performance/499/. Marking as `breaking` for investigation.

melissalinkert:
Subset job finally green: https://ci.openmicroscopy.org/view/Breaking/job/BIOFORMATS-DEV-breaking-repository-subset/189

Any veto to taking out of ```breaking```?

dgault:
Builds are all green and unit tests have been running successfully. The actual code changes all look fine to me also.

Opening sample images in ImageJ shows images displaying as before so the logic and arithmetic looks to still be correct.

**Without the PR:**
screen shot 2017-03-13 at 13 59 20

**With the PR:**
screen shot 2017-03-13 at 14 06 53

Running the performance test suggested against sample file `EB01 - Airyscan Fast SR 20x_Airyscan Processing.czi` also yielded some impressive improvements. Time measured was around the entire read loop.

**Without the PR:**
Total time elapsed: 602006
Total time elapsed: 600461
Total time elapsed: 598961

**With the PR:**
Total time elapsed: 27395
Total time elapsed: 27293
Total time elapsed: 26962

sbesson:
Also tested the pyramid generation for one of the corresponding sample CZI files. Without this PR, the pyramid generation time was measured to be 11941240ms and with this PR included it was reduced down to 3164184 ms. All looks good, merging.

PR #2772, opened 2017-03-01
Update broken NDP.view link

hflynn:
See https://ci.openmicroscopy.org/view/Docs/job/BIOFORMATS-DEV-merge-docs/616/warnings3Result/
This replaces the link to the old free NDP.view software with one to the new generation viewer NDP.view2 which is still free and claims to support the same file formats so is presumably a valid replacement.
Staged at https://www.openmicroscopy.org/site/support/bio-formats5.3-staging/formats/hamamatsu-ndpi.html and should make the merge build green again.


dgault:
Build is green and the new link works as expected.
I downloaded the new software and used it to open ndpi sample files so it is freely available and able to be used by users for viewing files as desired.

Looks good to merge.

PR #2778, opened 2017-03-08
Bump NetCDF dependency to 4.3.22

sbesson:
The NetCDF 4.3.19 dependency had an invalid POM hosted on the OME Artifactory. To resolve potential issues for downstream consumers, the dependency is upgraded to 4.3.22 which should be a patch release bump.

There should be no impact for application consumers. The Unidata artifactory is still mandatory in either a downstream pom.xml or gradle in order to retrieve NetCDF 4.3.x.

To test this PR, check the repository non-regression tests remain green. For manual testing, one might want to check a few HDF-based files using using `netcdf` esp. `ImarisHDF`, `MINC` and `Veeco`.

At the OMERO level, we might want to check this patch dependency bump causes no harm to the builds and integration tests when testing the 5.4.0-SNAPSHOT /cc @bramalingam @jburel


mtbc:
https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-breaking-full-repository/ remains red - we can look again on Monday.

sbesson:
Checked the OMERO integration by importing a few samples of the formats mentioned in the description via the command-line import - no issue detected /cc @bramalingam

PR #2779, opened 2017-03-08
Source files: headers cleanup

sbesson:
See https://trello.com/c/aOnYJJaJ/17-gpl-boilerplate-discrepancy

While reviewing the file headers, running `mvn license:update-file-header -DcanUpdateCopyright=true` across the code revealed some inconsistencies. This PR should bring the file headers in line with the  parameters of the License Maven plugin define in the top-level `pom.xml` for each component.

Summary of the main changes included in this PR that should be reviewed:

- the license `projectName` is `formats-api` is updated to match the component description and all license headers are updated accordingly (partly reverts the changes in https://github.com/openmicroscopy/bioformats/commit/4d84cd6f3e17bf05851087c388f0a49c66d97008)
- the copyright in the `IM3Reader` header is updated from GPL to BSD in agreement with the component license
- the copyright in the `CellH5Writer` header is updated from BSD to GPL in agreement with the component license
- missing headers are added to various recent files under `formats-bsd`

One remaining file which does not comply with the standard header is `components/formats-bsd/test/spec/AbstractTest.java` which is still licensed to Glencoe only unlike the rest of the codebase. I propose to unify this file using the same owner template as the rest of the code either in this PR or a follow-up /cc @melissalinkert @chris-allan @jrswedlow

### DIFF
--- a/components/autogen/src/MakeDatasetStructureTable.java
+++ b/components/autogen/src/MakeDatasetStructureTable.java
@@ -1,6 +1,6 @@
 /*
  * #%L
- * OME Bio-Formats package for reading and converting biological file formats.
+ * Bio-Formats autogen package for programmatically generating source code.
  * %%
  * Copyright (C) 2005 - 2016 Open Microscopy Environment:
  *   - Board of Regents of the University of Wisconsin-Madison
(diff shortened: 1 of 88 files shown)


melissalinkert:
All reads fine to me.

PR #2780, opened 2017-03-10
ImageConverterTest: create temporary directories using java.nio.file

sbesson:
See https://trello.com/c/NBMgI8Xz/113-change-the-file-creation-semantics-in-the-unit-tests

Also move the security manager logic to After and Before Method to allow individual test methods to be run via `mvn test` (see https://trello.com/c/WhRvUkyL/111-image-converter-tests-throw-if-run-individually)

To test this PR, check the Travis and Jenkins builds remain green and that the new logic is easy to understand. Also run individual methods as mentioned in the card above and check the individual tests now pass.

### DIFF
--- a/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
+++ b/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
@@ -36,6 +36,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.Permission;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
```diff
@@ -122,8 +129,7 @@ public void checkImage() throws FormatException, IOException {
 
   @Test(dataProvider = "suffixes")
   public void testDefault(String suffix) throws FormatException, IOException {
-    outFile = File.createTempFile("test", suffix);
-    outFile.delete();
+    outFile = tempDir.resolve("test." + suffix).toFile();
```
simleo: Extra dot in `"test."` (`suffix` already has a leading dot). Does not prevent the test from working fine though, so this change is not required.

simleo:
Travis and Jenkins builds OK, code looks good and all tests can now be run individually without throwing. Good to merge.

sbesson:
Extra commit pushed to remove unnecessary dot. I'll leave Travis to validate and then merge if that's okay @simleo.

PR #2781, opened 2017-03-12
Documentation: migration to rst as the file extension 

sbesson:
See https://github.com/openmicroscopy/bioformats/pull/2470 for the original version of this PR

The Model and OME Files documentation sets now use the standard `.rst` file extension for Sphinx files. This PR brings  the Bio-Formats documentation in line with this convention.

To test this PR:
- check the docs builds successfully incl. the documentation bundle
- check the Octave, MATLAB & Command Line tools bundles still include the history file
- check the autogen target (`gen-structure-table`, `gen-format-pages`, `gen-meta-support`) generates `.rst` files


dgault:
The docs build looks to be running successfully and from the latest builds the artifacts showed:
For Matlab - the NEWS.rst file was present
For Octave - the NEWS file was present but did not have the rst extension
For CL Tools - the NEWS.rst file was present

For local generation I ran the following:
ant -f components/autogen/build.xml gen-format-pages - build was successful and rst files generated for formats folder
ant -f components/autogen/build.xml gen-meta-support - build was successful and rst files generated for metadata folder
ant -f components/autogen/build.xml gen-structure-table - build was successful and rst files generated for the dataset-table

Tested the rst files by opening and viewing with restview. All pages tested appeared to display correctly.

The only thing that looked out of place was the lack of rst extension on the octave news file.

sbesson:
The semantics for the Octave package was introduced in https://github.com/openmicroscopy/bioformats/pull/1827. @carandraug any strong feeling about adding a `.rst` suffix to the `NEWS` file for consistency?

carandraug:
The `NEWS` file can be read manually by an user inspecting the release.  However, it is also identified by Octave's package manager which only installs files it recognise.  In Octave, the content of `NEWS` files are available via the `news` function so if the file is not installed, then `news bioformats` will not work.  So renaming the file for the Octave package would cause problems.

For what is worth, other recognised files are `ONEWS (for old news as per the GNU standards), `ChangeLog` (for all the commits), and `CITATION` (instructions for users that want to cite the package in publications and which would be displayed with `citation bioformats`).

sbesson:
Thanks @carandraug. As this is related to the client behavior, i suggest we keep things as they are @dgault?

dgault:
Agreed, keeping things as they are is fine with me. In that case the PR looks good to merge.

PR #2783, opened 2017-03-13
Bio-Formats Plugins: Warnings cleanup

dgault:
As part of this I have completed an initial pass of the bio-formats_plugins component and reviewed any warnings.

- Remove unused imports

**To test:**

- All builds and tests should remain green
- Run ImageJ tests from release process


sbesson:
All import cleanups make sense. As per https://github.com/openmicroscopy/bioformats/pull/2783/files?diff=unified&w=1, the only other change in this PR is the encapsulation of some reader/writers in try-with-resources block which looks sensible as well.

Also tried to reproduce a failing scenario by failing the Bio-Formats 5.3.4 Exporter (i.e. without this PR) and tracking potential leaks without success. This PR looks good to merge.

PR #2785, opened 2017-03-13
DICOM: add support for DICOMDIR files

melissalinkert:
This backports existing commits from two private pull requests.  See https://trello.com/c/ciKINMvW/18-dicom-support-for-directory-image

There are three things to test.  First is the physical size configuration change in 184c2f2 and https://github.com/openmicroscopy/data_repo_config/pull/183.  Most existing DICOM files are unaffected; the relevant specification is http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_10.7.html#sect_10.7.1.3

Second is the plane position parsing in eefed6c.  Without this PR, ```showinf -nopix -omexml curated/dicom/charles/img1582.dcm``` should not have any of the OME-XML ```Position*``` attributes set on ```Plane```.  With this PR, all ```Position*``` attributes should be set according to the ```0020,0032 Image Position (Patient) #1``` key in the original metadata.

Finally, a580739 and 6fecdde add remaining logic necessary to support ```DICOMDIR``` files, which allow multiple DICOM files in a single directory to be opened as a single dataset.  The only real example we have of this is QA 11401, which is configured in https://github.com/openmicroscopy/data_repo_config/pull/182.  Without this PR, each of the 3 DICOM files in the ```IMAGES``` directory under ```curated/dicom/qa-11401``` should open with ```showinf``` or ImageJ, but attempting to open the ```DICOMDIR``` file should result in an exception.  With this PR, opening the ```DICOMDIR``` file should result in 3 series (one per DICOM file) with all 4 files on the used files list.  The behavior when opening any of the individual DICOM files should remain unchanged.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -142,6 +142,9 @@ public class DicomReader extends FormatReader {
   private String date, time, imageType;
   private String pixelSizeX, pixelSizeY;
   private Double pixelSizeZ;
+  private List<Double> positionX = new ArrayList<Double>();
+  private List<Double> positionY = new ArrayList<Double>();
+  private List<Double> positionZ = new ArrayList<Double>();
 
   private Map<Integer, List<String>> fileList;
   private int imagesPerFile;
@@ -450,6 +453,9 @@ public class DicomReader extends FormatReader {
       originalSeries = 0;
       helper = null;
       companionFiles.clear();
+      positionX.clear();
+      positionY.clear();
+      positionZ.clear();
     }
   }
 
@@ -762,7 +768,7 @@ public class DicomReader extends FormatReader {
 
     // The metadata store we're working with.
     MetadataStore store = makeFilterMetadata();
-    MetadataTools.populatePixels(store, this);
+    MetadataTools.populatePixels(store, this, true);
 
     String stamp = null;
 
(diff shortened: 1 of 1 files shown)


sbesson:
See https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/397/ for a run of the repository subset against all DICOM samples with this PR included.

dgault:
The code changes themselves all look fine and make sense. Builds and tests are all running as expected. I carried out the following tests based on the PR description as well as some sanity checking by reading and opening existing samples to ensure they were unchanged. 

**Pixel Spacing**
Based on the provided spec (http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_10.7.html#sect_10.7.1.3) the updated physicalSize attributes look to be the correct reading. Tests with the modified config https://github.com/openmicroscopy/data_repo_config/pull/183 are passing ok.

**Plane Parsing**
Without the PR I cannot see any Position attribute on the OME-XML. With the PR included I can see:

```
      <XMLAnnotation ID="Annotation:114" Namespace="openmicroscopy.org/OriginalMetadata">
         <Value>
            <OriginalMetadata>
               <Key>Series 0 0020,0032 Image Position (Patient) #1</Key>
               <Value>0\0\3.699213</Value>
            </OriginalMetadata>
         </Value>
      </XMLAnnotation>

      .......

      <Plane PositionX="0.0" PositionXUnit="mm" PositionY="0.0" PositionYUnit="mm" PositionZ="3.699213" PositionZUnit="mm" TheC="0" TheT="0" TheZ="0"/>
```

The X, Y and Z positions look to be correct based on the original metadata.

**DICOMDIR support**
For https://github.com/openmicroscopy/bioformats/commit/a580739259d20cbac10885c9c95bcf93649e3cfb and https://github.com/openmicroscopy/bioformats/commit/6fecdde0267b307c4d79ddc3ae17d183a41cfbd7, without the PR I am unable to open the DICOMDIR file, with the PR the DICOMDIR file opens with 3 series containing the individual DICOM files which display correctly. The individual files also continue to open correctly on their own.

All testing passed as desired. PR looks like it should be good to merge.

Issue #2705, opened 2016-12-20
Problem with OME-XML validation


sebi06:
Hi guys,

I tried to validate an OME-TIFF using the bftools (5.3.1) but encountered the followwing problem.

1) Acquire CZI image and save to disk - OK
2) Open image an Fiji using BioFormats and save as OME-TIFF - OK
3) Create OME-XML using - OK

showinf -series 0 -no-upgrade -nometa -nopix -omexml-only -novalid -nocore Test.ome.tiff > Test.ome.xml

4) try to validate OME-XML --> but I could not get it to work, even with the no-upgrade option. What am I doing wrong here?

xmlvalid Test.ome.xml or xmlvalid -no-upgrade Test.ome.xml

First attempt:

Windows PowerShell
Copyright (C) 2013 Microsoft Corporation. Alle Rechte vorbehalten.

PS C:\Users\M1SRH\Documents> cd .\Software\BioFormats_Package\5.3.1
PS C:\Users\M1SRH\Documents\Software\BioFormats_Package\5.3.1> xmlvalid c:\Output\Test.ome.xml
Failed to compare version numbers
java.net.SocketTimeoutException: connect timed out
        at java.net.DualStackPlainSocketImpl.waitForConnect(Native Method) ~[na:1.8.0_111]
        at java.net.DualStackPlainSocketImpl.socketConnect(Unknown Source) ~[na:1.8.0_111]
        at java.net.AbstractPlainSocketImpl.doConnect(Unknown Source) ~[na:1.8.0_111]
        at java.net.AbstractPlainSocketImpl.connectToAddress(Unknown Source) ~[na:1.8.0_111]
        at java.net.AbstractPlainSocketImpl.connect(Unknown Source) ~[na:1.8.0_111]
        at java.net.PlainSocketImpl.connect(Unknown Source) ~[na:1.8.0_111]
        at java.net.SocksSocketImpl.connect(Unknown Source) ~[na:1.8.0_111]
        at java.net.Socket.connect(Unknown Source) ~[na:1.8.0_111]
        at sun.net.NetworkClient.doConnect(Unknown Source) ~[na:1.8.0_111]
        at sun.net.www.http.HttpClient.openServer(Unknown Source) ~[na:1.8.0_111]
        at sun.net.www.http.HttpClient.openServer(Unknown Source) ~[na:1.8.0_111]
        at sun.net.www.http.HttpClient.<init>(Unknown Source) ~[na:1.8.0_111]
[...]

PR #2788, opened 2017-03-15
Testing: Force Transformer to use internal implementation

dgault:
This PR is to be used for testing the effected attributes and transforms in relation to openmicroscopy/bioformats#2705

--breaking

### DIFF
--- a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
+++ b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
@@ -32,8 +32,11 @@
 
 package loci.formats.services;
 
+import java.io.File;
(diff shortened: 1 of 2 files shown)


sbesson:
As semi-expected - at least we are consistently testing them, the transform/specification unit tests are failing with this PR. For  the sake of the investigation you might want to disable them in the Maven build by commenting https://github.com/openmicroscopy/bioformats/blob/v5.3.4/components/formats-bsd/pom.xml#L162

sbesson:
Is there still ongoing investigation or should we close for now and reopen when we look into it?

dgault:
As this issue has not been widely reported and testing is not as straight forward as hoped I think it is fine to close this for now.

PR #2789, opened 2017-03-16
ScreenReader: ensure used files contain no duplicates

simleo:
Changes `getSeriesUsedFiles` to use a set rather than a list, to avoid duplicates. Note that, if calling `getUsedFiles`, to expose this bug it takes a mini-screen with one series (i.e., one well with one field). This is due to the fact that when there are multiple series, `FormatReader.getUsedFiles` merges partial lists into a set anyway.


joshmoore:
http://idr-testing.openmicroscopy.org/webclient/?show=plate-6101 looks reasonable. Cheers! cc: @eleanorwilliams

joshmoore:
Deployed as demo35

PR #2790, opened 2017-03-16
Add image acquisition tests as part of the testDeltaT

sbesson:
See https://trello.com/c/Uzb4YXli/116-andor-abd-vs-oiympus-fluoview-date-format

Although the `test-suite` configuration generation logic allows to store acquisition dates in the configuration files, these values are only used in the sanity check of `testSaneOMEXML`. In order to properly address issues like https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8238 together with non-regression tests, it is necessary to have acquisition dates tested as part of the `test-automated` target. Rather than creating another test, this PR proposes to extend the `testDeltaT` non-regression test and also compare read acquisition date with the values stored in the configuration files.

A companion configuration PR should adjust all the incorrect values. In additional to verifying the logic, the main review is to test the full repository job remains green with this PR included.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -1123,6 +1123,18 @@ public class FormatReaderTest {
     for (int i=0; i<reader.getSeriesCount(); i++) {
       config.setSeries(i);
 
+      // Test image acquisition date
+      String expectedDate = config.getDate();
+      String date = null;
+      if (retrieve.getImageAcquisitionDate(i) != null) {
+        date = retrieve.getImageAcquisitionDate(i).getValue();
+      }
+      if (expectedDate != null && date != null && !expectedDate.equals(date)) {
+        result(testName, false, "series " + i +
+          " (expected " + expectedDate + ", actual " + date + ")");
+        return;
+      }
+
       for (int p=0; p<reader.getImageCount(); p++) {
         Time deltaT = null;
         try {


sbesson:
Labelling as breaking and starting a full-repository job to review the state of our  data repository as some of the acquisition dates might be out of sync.

sbesson:
As per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-full-repository/55/, we have currently 335 failures related to AcquisitionDate being out of sync.

components/test-suite/src/loci/tests/testng/FormatReaderTest.java
```diff
@@ -1123,6 +1123,18 @@ public void testDeltaT() {
     for (int i=0; i<reader.getSeriesCount(); i++) {
       config.setSeries(i);
 
+      // Test image acquisition date
+      String expectedDate = config.getDate();
+      String date = null;
+      if (retrieve.getImageAcquisitionDate(i) != null) {
+        date = retrieve.getImageAcquisitionDate(i).getValue();
+      }
+      if (expectedDate != null && date != null && !expectedDate.equals(date)) {
```
dgault: Should we also fail in the instance when expectedDate is non null but actual date is null?

sbesson: That would certainly be a valuable addition and consistent with other tests like `testEmissionWavelengths`. The only reason for not adding it at this point is the size of the corresponding configuration PR fixing the mismatching configured values.

Can we handle this via a follow-up PR once this addition is included and first review the number of changes to do across the configured filesets?

dgault:
This seems like the sensible place to include this test and its good to have it part of the automated suite.
With the associated configuration PR all of the tests are now passing: https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/490/consoleFull

Leaving the additional null check for an additional PR is perfectly fine with me. Both this and the associated configuration PR should be good to merge.

sbesson:
Merging as per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository-win/101/ and https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/497/

PR #2804, opened 2017-03-30
Options doc

hflynn:
See https://trello.com/c/WQeSZG9U/14-options-documentation

This adds a new Options page under formats, crosslinks it from the relevant pages and adds a flag to the formats pages autogen to automatically stick a link to the new page on the appropriate format pages.

Not sure if we want to actually have the options underneath the notes, wasn't quite sure how to do that. And also do we want to remove/edit the existing options note under Leica LIF?

staged at https://www.openmicroscopy.org/site/support/bio-formats5.4-staging/formats/options.html etc

### DIFF
--- a/components/autogen/src/doc/FormatPage.vm
+++ b/components/autogen/src/doc/FormatPage.vm
@@ -87,13 +87,18 @@ Presence: |$presenceRating|
 
 Utility: |$utilityRating|
 
-#if (($privateSpecification) or ($notes))
+#if (($privateSpecification) or ($options) or ($notes))
 **Additional Information**
 #end
 
 #if ($privateSpecification)
 **Please note that while we have specification documents for this
 format, we are not able to distribute them to third parties.**
+#end
+
+#if ($options)
+Additional options are available for reading this format type, see
+doc link for information.
 #end
 
 #foreach ($note in $notes)
--- a/components/autogen/src/doc/FormatPage.vm
+++ b/components/autogen/src/doc/FormatPage.vm
@@ -99,8 +99,8 @@ format, we are not able to distribute them to third parties.**
 #if ($options)
 Additional options are available for reading this format type, see
 :doc:`options` for information.
-#end
 
+#end
 #foreach ($note in $notes)
 $note
 #end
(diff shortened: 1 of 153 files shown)


components/autogen/src/doc/FormatPage.vm
```diff
@@ -96,6 +96,11 @@ Utility: |$utilityRating|
 format, we are not able to distribute them to third parties.**
 #end
 
+#if ($options)
+Additional options are available for reading this format type, see
+:doc:`options` for information.
+#end
+
```
sbesson: Inverting lines 102 and lines 103 (i.e. putting the extra line in the conditional block) ad regenerating the format pages might fix the unnecessary blank lines additions for all formats.

dgault:
The staging looks good, everything displays and reads well, all the links are working correctly, the formats pages display correctly without unnecessary line breaks both when options are and are not included.

As for adding the format specific table on each page, I think as long as the we link to the options page there isn't a need. It would likely be a single line in most cases unless it grows substantially.

The existing Leica-LIF specific note probably should be edited to make note of the new option available.

One other small suggestion I would have, in the table, could the format name link to the format page?

hflynn:
@dgault hopefully sorted now. I'm still not entirely happy with the LIF page but 2 links to the new page seems less confusing than removing the autogenerated text for that one format.

docs/sphinx/formats/options.rst
```diff
@@ -0,0 +1,40 @@
...
+   * - Zeiss CZI
+     - ``zeissczi.autostitch``
+     - true
+     - Automatically stitch tiled images
+
+Usage
+-----
+
+Options can be used via the command line with
+:doc:`showinf -option </users/comlinetools/display>`, in ImageJ via the
```
hflynn: Done

docs/sphinx/formats/options.rst
```diff
@@ -0,0 +1,40 @@
+Additional Reader options
+=========================
+
+Some Readers have additional options which can be used to inform how
```
hflynn: Having checked, no, only when referring to a specific FormatReader so changed there & in page title

hflynn:
@sbesson this page isn't very extensive as yet so I certainly could add writer options here too - I could just split the page into Reader options and Writer options and have a separate table. What are the usage instructions for writer options?

hflynn:
Writer options added now too

dgault:
Yep this looks good on staging with the latest commits, good to merge from my perspective also

PR #2806, opened 2017-03-31
Memoizer: Add failIfMissing flag for perf (rebased onto metadata54)

aleksandra-tarkowska:
This is the same as gh-2435 but rebased onto metadata54.

----

If `failIfMissing` is set to true then no `setId` call
will be performed on the underlying reader. Instead, if
`loadMemo` returns null, a `FormatException` will be thrown
signaling that no cache is available.


aleksandra-tarkowska:
--rebased-from #2435

PR #2809, opened 2017-04-03
Add Hamamatsu public sample links

hflynn:
See https://trello.com/c/NlcprkHA/12-make-openslide-test-data-public

I've removed the Openslide links because the provenance for the files is listed in the readme anyway and we don't have it for any of the other samples on the format pages.

Staged at https://www.openmicroscopy.org/site/support/bio-formats5.4-staging/formats/hamamatsu-ndpi.html and https://www.openmicroscopy.org/site/support/bio-formats5.4-staging/formats/hamamatsu-vms.html

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -192,6 +192,7 @@ notes = .. seealso:: \n
 extensions = .svs
 owner = `Aperio <http://www.aperio.com/>`_
 bsd = no
+software = `OpenSlide <http://openslide.org>`_
 versions = 8.0, 8.2, 9.0
 weHave = * many SVS datasets\n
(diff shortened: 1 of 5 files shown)


sbesson:
:+1: now I think about it, I wonder if Openslide still be listed but instead as another instance of freely available software reading these file formats /cc @melissalinkert

melissalinkert:
Listing Openslide in the software section certainly makes sense to me.

hflynn:
No problem, I'll push another commit

hflynn:
Added OpenSlide to the other formats we supported listed on http://openslide.org except TIFF because they only read tiled TIFFs so I wasn't sure whether to add it there - can certainly do so if desired though.

PR #2810, opened 2017-04-04
Enable forward compatibility

bjoernthiel:
OBF is backward compatible, so the reader should be forward compatible.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -79,9 +79,6 @@ public class OBFReader extends FormatReader
   private static final String STACK_MAGIC_STRING = "OMAS_BF_STACK\n";
   private static final short MAGIC_NUMBER = (short) 0xFFFF;
 
-  private static final int FILE_VERSION = 2;
-  private static final int STACK_VERSION = 5;
-
   private static final int MAXIMAL_NUMBER_OF_DIMENSIONS = 15;
 
   private class Stack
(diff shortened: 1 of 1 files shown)


sbesson:
Hi @bjoernthiel, many thanks for the contribution. Could you expand a bit on the PR description to describe how to test the patch? In particular, do you have any corresponding sample files that we would also add to our data repository for be daily checked using our non-regression tests?

On a technical level, we are trying to avoid removing/adding private fields to the readers in patch releases to minimize serialization breakages. Could the two dropped `private static final` fields be marked as deprecated instead? We will clean them in the following minor version increment of Bio-Formats.

bjoernthiel:
The stack version and possibly the file version as well are are going to be increased in a future version of the OBF file format. Unfortunately I don't have example data yet. But as OBF is backward compatible these version checks are useless. 

Dropping the corresponding fields later is OK with me.

dgault:
Marking for inclusion in daily builds and tests
--include

snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

components/formats-bsd/src/loci/formats/in/OBFReader.java
```diff
@@ -135,9 +132,9 @@ private int getFileVersion(RandomAccessInputStream stream) throws IOException
   @Override
   public boolean isThisType(RandomAccessInputStream stream) throws IOException
   {
-    final int fileVersion = getFileVersion(stream);
+    final int fileVersion = getFileVersion(stream) ;
```
sbesson: Remove extra whitespace before the semicolon.

components/formats-bsd/src/loci/formats/in/OBFReader.java
```diff
@@ -135,9 +132,9 @@ private int getFileVersion(RandomAccessInputStream stream) throws IOException
...
 
-    return fileVersion >= 0 && fileVersion <= FILE_VERSION;
+    return fileVersion >= 0 ;
```
sbesson: Remove extra whitespace before the semi-colon

Issue #2775, opened 2017-03-04
Please add inputstream/outputstream as input/output options for IFormatReader in LociTools...


markehammons:
Hi All,

Right now it appears that when using locitools to read or write a new file, I need to give it a filename using the setId function. I would prefer to be able to send the output to a remote server, using an output stream. Another useful example is reading a gzipped file by passing in a gzipInputStream. Could you support the InputStream/OutputStream classes as inputs to an IFormatReader/IFormatWriter?

https://downloads.openmicroscopy.org/bio-formats/5.1.2/api/loci/formats/IFormatReader.html

dgault:
Hi,

One way that you could potentially achieve this functionality would be using `Location.mapFile` - https://github.com/ome/ome-common-java/blob/master/src/main/java/loci/common/Location.java#L219-L224

Typically an id is a filename on disk but it is possible to have this mapped to an IRandomAccess object instead.

A full example of reading and writing in this manner can be found here: https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/utils/ReadWriteInMemory.java

Hopefully that will be suitable for your needs but if you have any further questions please do let us know.

sbesson:
Assuming https://github.com/openmicroscopy/bioformats/issues/2775#issuecomment-284415638 is an acceptable workaroung for the time being, closing this since the in-memory example is now reviewed, tested and better exposed in the documentation with Bio-Formats 5.4.1 - see https://github.com/openmicroscopy/bioformats/pull/2812.

PR #2812, opened 2017-04-07
Add sphinx doc for in-memory read and write

dgault:
This PR follows on from the request in https://github.com/openmicroscopy/bioformats/issues/2775
Adding new documentation for an existing file available in utils. I have kept the file in its original location rather than moving it too the docs examples folders, this however is easily changed if its believed the examples folder is a better home for it.

To test:
- Ensure builds are green
- Test all links and markup is displaying correctly
- Ensure text is suitable for describing the use case for both reading and writing in-memory

### DIFF
--- a/components/formats-gpl/utils/ReadWriteInMemory.java
+++ b/components/formats-gpl/utils/ReadWriteInMemory.java
@@ -47,6 +47,7 @@ public class ReadWriteInMemory {
     }
     String path = args[0];
 
+    /* file-read-start */
     // read in entire file
     System.out.println("Reading file into memory from disk...");
     File inputFile = new File(path);
@@ -55,7 +56,9 @@ public class ReadWriteInMemory {
     byte[] inBytes = new byte[fileSize];
     in.readFully(inBytes);
     System.out.println(fileSize + " bytes read.");
+    /* file-read-end */
 
+    /* mapping-start */
     // determine input file suffix
     String fileName = inputFile.getName();
     int dot = fileName.lastIndexOf(".");
@@ -64,11 +67,13 @@ public class ReadWriteInMemory {
     // map input id string to input byte array
     String inId = "inBytes" + suffix;
     Location.mapFile(inId, new ByteArrayHandle(inBytes));
+    /* mapping-end */
 
     // read data from byte array using ImageReader
     System.out.println();
     System.out.println("Reading image data from memory...");
 
+    /* read-start */
     ServiceFactory factory = new ServiceFactory();
     OMEXMLService service = factory.getInstance(OMEXMLService.class);
     IMetadata omeMeta = service.createOMEXMLMetadata();
@@ -76,6 +81,8 @@ public class ReadWriteInMemory {
     ImageReader reader = new ImageReader();
     reader.setMetadataStore(omeMeta);
     reader.setId(inId);
+    /* read-end */
+
     int seriesCount = reader.getSeriesCount();
     int imageCount = reader.getImageCount();
     int sizeX = reader.getSizeX();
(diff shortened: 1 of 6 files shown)


hflynn:
Staged at https://www.openmicroscopy.org/site/support/bio-formats5.4-staging/developers/in-memory.html

docs/sphinx/developers/in-memory.rst
```diff
@@ -0,0 +1,82 @@
+In memory reading and writing in Bio-Formats
```
mtbc: hyphenate In-memory

docs/sphinx/developers/in-memory.rst
```diff
@@ -0,0 +1,82 @@
...
+To use a writer to output to memory rather than an output file a similar process 
+is required. First a mapping is created between a suitable output filename and 
+the in-memory data.
+
+.. literalinclude:: ../../../components/formats-gpl/utils/ReadWriteInMemory.java
+   :language: java
+   :start-after: out—mapping-start
+   :end-before: out—mapping-end
+
+The mapped filename can now be passed to initialise the writer as standard.
```
mtbc: initialize

docs/sphinx/developers/in-memory.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. literalinclude:: ../../../components/formats-gpl/utils/ReadWriteInMemory.java
+   :language: java
+   :start-after: write—init-start
+   :end-before: write—init-end
+
+The data can then be written to memory using the same read and write loop which would 
+normally be used to write a file disk.
```
mtbc: ... file *to* disk?

docs/sphinx/developers/in-memory.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. literalinclude:: ../../../components/formats-gpl/utils/ReadWriteInMemory.java
+   :language: java
+   :start-after: write-start
+   :end-before: write-end
+
+If it desired the data written to memory can then be flushed to disk and written to
```
mtbc: -it

docs/sphinx/developers/in-memory.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. literalinclude:: ../../../components/formats-gpl/utils/ReadWriteInMemory.java
+   :language: java
+   :start-after: flush-start
+   :end-before: flush-end
+
+.. seealso:: 
+   :source:`ReadWriteInMemory.java <components/formats-gpl/utils/ReadWriteInMemory.java>` - Full source code which is
+   referenced here in part. You will need to have :file:`bioformats_package.jar` in your 
+   Java CLASSPATH in order to compile :file:`ReadWriteInMemory.java`.
```
mtbc: would `:envvar:` or some other markup be appropriate?

sbesson:
If we expose these snippets into the developer documentation, my preference would be towards migrating it into `developers/examples` and add it to `ExampleSuite`  so that it can be tested as part of the Maven build.

More generally, while reviewing how to handle our duplicate examples collection (https://trello.com/c/COTJHwy0/97-developer-documentation-review-list-of-writing-examples), I would propose we could handle all similar examples as follows:

- maintain any existing classes under `formats-{gpl,bsd}/utils` as consumers might depend on them but mark them as deprecated as we migrate them
- add/copy/update all new examples directly under `docs/sphinx/developers/examples`, add them to `ExampleSuite` so they can be tested daily and link these from the Sphinx pages
- remove the deprecated `utils` classes in the next major version increment

Would this proposal make sense @dgault @bramalingam ?

dgault:
@sbesson I've pushed the changes needed to add this to the docs examples. However I did discover last night that we all ready have a mention of this class in the 'Use as a Java library' page - https://www.openmicroscopy.org/site/support/bio-formats5.4/developers/java-library.html#examples-of-usage

Perhaps that is all we need and this new doc is overkill?

hflynn:
It would be a shame to ditch this when you've taken the time to write it, I would just add a link to the Java library page referring devs to the new page for further info.

sbesson:
Overall reads fine and the new example is tested as part of the Travis Maven build

```
Executing ReadWriteInMemory
Reading file into memory from disk...
0 bytes read.

Reading image data from memory...
Series count: 1
First series:
	Image count = 1
	SizeX = 512
	SizeY = 512
	SizeZ = 1
	SizeC = 1
	SizeT = 1

Writing planes to destination in memory: .

Flushing image data to disk...
Success
```

One last grammatical comment (https://github.com/openmicroscopy/bioformats/pull/2812#discussion_r110609518) to address and this should be ready to merge.

components/formats-gpl/utils/ReadWriteInMemory.java
```diff
@@ -35,6 +35,8 @@
 
 /**
  * Tests the Bio-Formats I/O logic to and from byte arrays in memory.
+ * @deprecated Class has been moved to documentation examples and is available for download 
+ * from http://www.openmicroscopy.org/site/support/bio-formats/developers/in-memory.html
  */
 public class ReadWriteInMemory {
```
sbesson: Should the class also be annotated with `@Deprecated` in addition to the Javadoc?

PR #2813, opened 2017-04-10
Versioning clarifications

sbesson:
See https://trello.com/c/BC2QUyPW/9-versioning-policy-5-4-x-updates

This commit makes official rules which have been almost in place for the last patch/minor releases and should improve transparency for consumers.

- Precise the OMERO/Bio-Formats versioning decoupling point
- Use [semantic versioning](http://semver.org/) for the Bio-Formats API policy. This policy has been effectively in place since Bio-Formats 5.4.0.
- Increase the stringency of the serialization policy
- Add paragraph about the versioning of individual format changes

In terms of release, as per the usage of semver for the API, one expectation is that it should be possible to have more minor releases of Bio-Formats containing API backwards-compatible additions, format reader additions and/or backwards-incompatible or serialization breaking format reader changes.

### DIFF
--- a/docs/sphinx/about/index.rst
+++ b/docs/sphinx/about/index.rst
@@ -69,7 +69,7 @@ The following set of rules describe the current versioning policy using
 `RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`_.
 
 The Bio-Formats API follows strict `semantic versioning <http://semver.org>`_
-since Bio-Formats 5.3.0 i.e.:
+since Bio-Formats 5.4.0 i.e.:
 
 - The version number MUST take the form X.Y.Z where X, Y, and Z are
   non-negative integers, and MUST NOT contain leading zeroes. X is the major
--- a/docs/sphinx/about/index.rst
+++ b/docs/sphinx/about/index.rst
@@ -54,9 +54,9 @@ Bio-Formats versions
 --------------------
 
 Since Bio-Formats 5.1.3, Bio-Formats is decoupled from OMERO with its own
-release schedule than being updated whenever a new version of
+release schedule rather than being updated whenever a new version of
 :products_plone:`OMERO <omero>` is released.
-This change allows in more frequent releases to get fixes out to the
+This change allows for more frequent releases to get fixes out to the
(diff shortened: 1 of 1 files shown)


docs/sphinx/about/index.rst
```diff
@@ -50,52 +50,60 @@ include:
...
-Bio-Formats is now decoupled from OMERO with its own release schedule rather
-than being updated whenever a new version of :products_plone:`OMERO <omero>`
-is released.
-We expect this to result in more frequent releases to get fixes out to the
-community faster.
-
-See the :doc:`version history <whats-new>` for a list of major changes in
-each release.
+Since Bio-Formats 5.1.3, Bio-Formats is decoupled from OMERO with its own
+release schedule than being updated whenever a new version of
```
mtbc: "than" no longer fits without the "rather"

docs/sphinx/about/index.rst
```diff
@@ -50,52 +50,60 @@ include:
...
+:products_plone:`OMERO <omero>` is released.
+This change allows in more frequent releases to get fixes out to the
```
mtbc: "allows in" is odd

docs/sphinx/about/index.rst
```diff
@@ -50,52 +50,60 @@ include:
...
+
+For format reader fixes and additions, the policy should read as follows:
+
+- The minor version Y MUST be incremented when a new file format reader is
+  introduced.
+- The minor version Y MUST be incremented when a non backwards-compatible
+  format bug fix is introduced, e.g. a fix that modifies the core metadata
+  of existing files.
+- The patch version Z MUST be incremented if only backwards-compatible format
+  bug fixes are introducted
```
mtbc: "introduced", also needs terminal period

docs/sphinx/about/index.rst
hflynn: "allows for" would be correct

dgault:
This reads fine to me, it all makes sense and should be easily understood by users and developers who would be concerned with breaking changes. Looks good to merge from my perspective.

hflynn:
Looks good to me too

PR #2814, opened 2017-04-10
Add support for Olympus OIR

melissalinkert:
See https://trac.openmicroscopy.org/ome/ticket/13082 and https://github.com/openmicroscopy/data_repo_config/pull/186

This adds a new reader to handle Olympus .oir files and updates ```readers.txt``` accordingly.  No other readers or components should be affected.

As noted on the ticket, the Olympus plugin for ImageJ should be used for validation (http://imagej.net/OlympusImageJPlugin).  A screenshot and exported OME-TIFF from this plugin are present in the same ```data_repo``` directories as each valid .oir (see note on the configuration PR).

A few specific things to check apart from passing tests would be:

* metadata shown in screenshots is visible in the original metadata table and, where reasonable, the generated OME-XML
* channel colors are the same when opening an .oir with the Olympus plugin and Bio-Formats
* pixels hashes match and/or visual inspection of planes passes when comparing the .oir in Bio-Formats with the exported OME-TIFF
* memoization works, i.e. repeated calls to ```showinf -cache``` are successful

Adding to the breaking queue for now as discussed this morning.

/cc @sbesson, @chris-allan, @emilroz

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -449,11 +449,11 @@ public class OIRReader extends FormatReader {
       root = XMLTools.parseDOM(xml).getDocumentElement();
     }
     catch (ParserConfigurationException e) {
-      LOGGER.warn("Could not parse XML", e);
+      LOGGER.debug("Could not parse XML", e);
       return;
     }
     catch (SAXException e) {
-      LOGGER.warn("Could not parse XML", e);
+      LOGGER.debug("Could not parse XML", e);
       return;
     }
 
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -218,8 +218,10 @@ public class OIRReader extends FormatReader {
         }
       }
 
-      String laserId = MetadataTools.createLSID("LightSource", 0, ch.laserIndex);
-      store.setChannelLightSourceSettingsID(laserId, 0, c);
+      if (ch.laserIndex >= 0 && ch.laserIndex < lasers.size()) {
+        String laserId = MetadataTools.createLSID("LightSource", 0, ch.laserIndex);
+        store.setChannelLightSourceSettingsID(laserId, 0, c);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


mtbc:
`findbugs.xml` warns,
```xml
<FileStats path="loci/formats/in/OIRReader.java" bugCount="2" size="808" bugHash="0a01905a6822a8cedca31192e6128b3f"/>
```

melissalinkert:
Findbugs fixed and subset job now passing, see https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-breaking-repository-subset/243

components/formats-gpl/src/loci/formats/in/OIRReader.java
```diff
@@ -0,0 +1,1238 @@
...
+   */
+  @Override
+  public byte[] openBytes(int no, byte[] buf, int x, int y, int w, int h)
+    throws FormatException, IOException
+  {
+    FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
+
+    int nextPointer = 0;
+    int blocksPerChannel = pixelBlocks.size() / channels.size();
+
```
dgault: nextPointer and blocksPerChannel don't appear to be used. I can see value in having blocksPerChannel for future reference and development, is there any need for nextPointer?

dgault:
Subset job is green and testing spreadsheet has no issues reported on it.

Reviewing the actual code and the design, logic and readability of the code all look good to me.

Testing using the available sample files and the metadata from the screenshots looks consistent with that present in original metadata and OME XML. When opened the images display correctly when compared with the Olympus plugin and when exported to OME-TIFF they import and display as expected.

Running the command `showinf -cache` also generates a memo file as expected.

All of the testing for this reader looks good and it should be ready to merge.

components/formats-gpl/src/loci/formats/in/OIRReader.java
melissalinkert: Leaving both of those lines in was an oversight, so will certainly remove in the forthcoming follow-up PR.  It worries me a little that Findbugs didn't catch either of them, though, since it seems like both should have been flagged as dead local stores.

sbesson:
Barring the comments/concerns from https://github.com/openmicroscopy/bioformats/pull/2814#discussion_r112948901 which should be addressed in an upcoming PR, the current set of OIR files have been successfully tested on Windows and Linux by the automated test. Merging to allow the subsequent steps to be opened as follow-up PRs.

PR #2818, opened 2017-04-17
Fix Ant build.version property for non-SNAPSHOT versions

sbesson:
## Context

In https://github.com/openmicroscopy/bioformats/pull/2644, the hardcoding of the Git revision number into a properties file was replaced by the introspection of the `Implementation-Build` property in the JAR manifest.

While the [Build Number Plugin](http://www.mojohaus.org/buildnumber-maven-plugin/) is directly used for the Maven build, additional logic was added to the Ant build in `ant/global.xml` to parse the SCM revision.  Whilst testing the Ant-generated command-line tools, it was noticed that for non SNAPSHOT builds, the VCS revision is set to be equal to the release version:

```
$ ~/bftools-5.4.0/showinf -version
Version: 5.4.0
Build date: 20 March 2017
VCS revision: 5.4.0
```

## Changes

This commit fixes the behavior of the Ant calculation of the `build.version` property to use `git rev-parse --verify HEAD`  independently of whether a SNAPSHOT or a release version is built. This implementation is in agreement with the behavior of the Maven build number plugin and should fix the current discrepancy for the value of Implementation-Build stored in the JAR manifest.

## Testing

To test these changes, review the content of the `MANIFEST.MF` of JARs generated by either Ant or Maven for:
- a SNAPSHOT build e.g. the daily Jenkins CI builds 
- a release build, i.e. by running `./tools/bump_maven_version.py X.Y.Z` on top of this branch and generating the JARs

In all cases, the `Implementation-Build` should contain the full Git revision number e.g.

```
$ unzip -qc artifacts/bioformats_package.jar META-INF/MANIFEST.MF | grep Implementation-Build
Implementation-Build: 598e0a31688b92d5e54b9b6d880b73fe9be7face
```

### DIFF
--- a/ant/global.xml
+++ b/ant/global.xml
@@ -109,10 +109,9 @@ Type "ant -p" for a list of targets.
       </exec>
       <property name="build.version" value="${build.revision}"/>
     </try>
-    <catch>
-      <property name="build.version" value="${release.version}"/>
-    </catch>
+    <catch/>
   </trycatch>
+  <property name="build.version" value="${release.version}"/>
 
   <!-- Internal build targets -->
 
--- a/ant/global.xml
+++ b/ant/global.xml
@@ -97,22 +97,22 @@ Type "ant -p" for a list of targets.
     <isset property="isSnapshot"/>
     <then>
       <echo>isSnapshot = ${isSnapshot}</echo>
-      <trycatch>
-        <try>
-          <exec executable="git" outputproperty="build.revision"
-                failifexecutionfails="true" failonerror="true">
-            <arg value="rev-parse"/>
-            <arg value="--verify"/>
-            <arg value="HEAD"/>
-          </exec>
-          <property name="build.version" value="${build.revision}"/>
-        </try>
-        <catch>
-        </catch>
-      </trycatch>
     </then>
   </if>
-  <property name="build.version" value="${release.version}"/>
+  <trycatch>
+    <try>
+      <exec executable="git" outputproperty="build.revision"
+            failifexecutionfails="true" failonerror="true">
+        <arg value="rev-parse"/>
+        <arg value="--verify"/>
+        <arg value="HEAD"/>
+      </exec>
+      <property name="build.version" value="${build.revision}"/>
+    </try>
+    <catch>
+      <property name="build.version" value="${release.version}"/>
+    </catch>
+  </trycatch>
 
   <!-- Internal build targets -->
 
--- a/ant/global.xml
+++ b/ant/global.xml
@@ -111,7 +111,7 @@ Type "ant -p" for a list of targets.
     </try>
     <catch/>
   </trycatch>
-  <property name="build.version" value="${release.version}"/>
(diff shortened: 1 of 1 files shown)


sbesson:
As mentioned by @rleigh-codelibre who is going to do the testing, another use case to verify is to build from source release (i.e. no Git repository)

rleigh-codelibre:
`<property name="build.version" value="${build.revision}"/>` should always be outside the other logic--it's the fallback if unset.  It doesn't belong in the catch block since it's to be set in all cases if not set earlier.

Suggested change:
- A [add version suffix](https://github.com/rleigh-codelibre/bioformats/commit/fde64885342988d417c423d880e39023e23cdb8f), plus needs moving outside the catch.  This will make the build from the source release use the versioned tag as the build version.  For snapshots, it will be vnnn-SNAPSHOT which is wrong but no worse than the unprefixed version used right now which is equally invalid as a git version.
- B [alternative pr](https://github.com/rleigh-codelibre/bioformats/commit/a4ca2901dbbea47902377283ef2e322a74255fb5) -- add v suffix so that the build version becomes a versioned tag for the official release.  Behaviour in git tree unchanged.

For A the behaviour in the git tree is unchanged for snapshots, for releases you get a hash in the git tree, and a version tag in the source build
For B the behaviour in the git tree is unchanged for snapshots, for releases you get a version tag in the git tree and in the source build

So (B) is more consistent between git and source releases.  But (A) is more similar to maven.

Maven behaviour for reference: Implementation-Build in MANIFEST.MF set to hash when built from git.  Set to UNKNOWN when built from source release.

Either approach works for me, especially since we're going to be dropping the ant logic anyway.

sbesson:
Thanks for the testing and the source release discussion. My preference would go towards option A) and restoring the pre-5.3.0 behavior which was exposing the commit SHA1 in `FormatTools.VCS_REVISION` when possible.

On the other hand, the Maven behavior makes me wonder whether the fallback behavior should rather be to either not set the `build.version` property or set it to `''`. For source releases, the notion of VCS revision makes no sense and the manifest metadata is the only usage of the `build.version` property. In this case, the version number would still be stored in `Implementation-Version` and introspected via the `FormatTools` API.

rleigh-codelibre:
(A) would be fine with me.

ant/global.xml
```diff
@@ -109,10 +109,9 @@ Type "ant -p" for a list of targets.
       </exec>
       <property name="build.version" value="${build.revision}"/>
     </try>
-    <catch>
-      <property name="build.version" value="${release.version}"/>
-    </catch>
+    <catch/>
   </trycatch>
+  <property name="build.version" value="${release.version}"/>
```
rleigh-codelibre: Did you want to add a `v` suffix in there as well?

sbesson: Good point. I'll push an extra commit...

rleigh-codelibre:
This works correctly and is good to merge, thanks.

Issue #2820, opened 2017-04-20
ND2 files cannot be read and appears corrupt since version 5.0.3


SarenT:
Hi,

I am trying to open ND2 files from Nikon to save them as TIFFs. But it appears to be with some files, that the file cannot be read properly and looks corrupt (looks like noise only with weird unexpected shapes). I mainly use FiJi but because of this issue I have tried several version from the website. I have limited the issue to version 5.0.3 and above. Here is an example of the same file displayed completely differently:
![bio-formats error](https://cloud.githubusercontent.com/assets/2906565/25234689/2a9ef27a-25e3-11e7-9a03-c38164bdd235.png)

I can provice more information about the issue on demand. I am using Ubuntu 16.04 but the issue also occurs on Windows with ImageJ. In the syslog I see following errors:

[bio-formats_error.txt](https://github.com/openmicroscopy/bioformats/files/943427/bio-formats_error.txt)

SarenT:
Yes, first I tried out 5.4.1. Then I tried to find out where it started to not to open it correctly.
I tried with and without chunkmap and with and without native library. None of them works with the current release. I will upload the file ASAP. The file is created by NIS elements software.

SarenT:
I have uploaded the file as "sample.nd2" right now.

dgault:
Hi @SarenT, thank you for uploading the sample file. I was able to test and reproduce the issue you described. From further debugging this appears to be due to an incorrect PixelType being used after parsing the file. I have opened a PR https://github.com/openmicroscopy/bioformats/pull/2823 in order to test a potential fix for issue.

sbesson:
@SarenT: thanks again for contacting us, we expect David's fix to be part of the next release of Bio-Formats. In order to catch potential regressions, we will add the sample file you uploaded to us to our data repository for daily testing. One question, could this data be made public under the CC-BY license alongside other [public ND2 images](http://downloads.openmicroscopy.org/images/ND2/) or should it remain private? If needed, we can provide a location to review this file.

SarenT:
@sbesson: Sure, it is nice to hear that I could somehow contribute to the project. Could you pleace keep the file private for now. I need to ask other people, who has part in it (although this should not be any problem). Maybe tomorrow I can even provide another file, which can be made public without consent of others.

sbesson:
Closing this issue as https://github.com/openmicroscopy/bioformats/issues/2823 is merged

SarenT:
[...]

PR #2823, opened 2017-04-24
ND2Handler - Only use uiBpcInMemory and uiBpcSignificant

dgault:
This PR is a followup to https://github.com/openmicroscopy/bioformats/issues/2820 for testing

In the sample file related to the above Github issue we are setting bitsPerPixel as 14 but the PixelType is set to INT8. The original PixelType of the image would have been UINT16 but is overwritten by parsing "uiBpc" with a value of 8.

The possible tags which can be found in the ND2 file appear to be "uiBpc", "uiBpcInMemory" and "uiBpcSignificant". Originally we were parsing "uiBpcInMemory" but this was updated in https://github.com/openmicroscopy/bioformats/commit/63572fbc557f4e01514135e252d2464f2ac8193f

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Handler.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Handler.java
@@ -493,7 +493,7 @@ public class ND2Handler extends BaseHandler {
           LOGGER.warn("Could not set the pixel type", e);
         }
       }
-      else if (qName.equals("uiWidthBytes") || qName.startsWith("uiBpc")) {
+      else if (qName.equals("uiWidthBytes") || qName.startsWith("uiBpcInMemory") || qName.startsWith("uiBpcSignificant")) {
         int div = qName.equals("uiWidthBytes") ? ms0.sizeX : 8;
         if (div > 0) {
           int bits = Integer.parseInt(value);


sbesson:
As per https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-repository-subset/246/, this PR does not seem to introduce any regression for the ND2 files in our data repository.

components/formats-gpl/src/loci/formats/in/ND2Handler.java
```diff
@@ -493,7 +493,7 @@ else if (qName.equals("uiMaxDst")) {
           LOGGER.warn("Could not set the pixel type", e);
         }
       }
-      else if (qName.equals("uiWidthBytes") || qName.startsWith("uiBpc")) {
+      else if (qName.equals("uiWidthBytes") || qName.startsWith("uiBpcInMemory") || qName.startsWith("uiBpcSignificant")) {
```
rleigh-codelibre: Do these tags need to be parsed in order of priority, so we pick the most likely to be correct one first?  Is there any potential for non-determinism here in terms of the ordering during parsing which can lead to overwriting the value; should we check the value is set in the conditional to avoid overwriting?

Also, should uiBpcSignificant be handled separately and set bitsPerPixel in the core metadata (or call `setPixelsSignificantBits`) rather than set the pixel type?

dgault: My reluctance with prioritising them is that in this case the "uiBpc" value is incorrect for bitsPerPixel and PixelType and cant be used for this purpose no matter the order of the tags.

For `uiBpcSignificant`, it is used by `parseKeyAndValue` to set the bitsPerPixel in the core metadata. If both values are set then they should set the same PixelType though one will overwrite the other. Im unsure though if we can rely on both being set as there appear to be samples with only one or the other set in which case they would use either for PixelType.

sbesson: @rleigh-codelibre @dgault  did a quick audit of the metadata for the ND2  files we have in the data repository ([nd2_uiBpc.txt](https://github.com/openmicroscopy/bioformats/files/961570/nd2_uiBpc.txt)). As far as I can tell, when present `uiBpc` is either equal to `uiBpcSignificant` or lower in all cases and does not seem to be reliable for `CoreMetadata` as opposed to the two other keys.

For `uiBpcSignificant`, I think its value is already stored as `ms.bitsPerPixel` when available e.g. using [130606_JCC719_003.nd2](http://downloads.openmicroscopy.org/images/ND2/jonas/qa-7534/130606_JCC719_003.nd2), `showinf` gives:

```
Reading core metadata
filename = /ome/data_repo/curated/nd2/public/jonas/qa-7534/130606_JCC719_003.nd2
Series count = 1
Series #0 :
	Image count = 160
	RGB = false (1) 
	Interleaved = false
	Indexed = true (false color, 16-bit LUT: 3 x 65536)
	Width = 304
	Height = 232
	SizeZ = 8
	SizeT = 20
	SizeC = 1
	Thumbnail size = 128 x 97
	Endianness = intel (little)
	Pixel type = uint16
	Valid bits per pixel = 14
	Metadata complete = true
	Thumbnail series = false
```

Should the file provided in the original issue be added to our data repository, configured and included in the daily non-regression tests?

dgault: Thats correct. I can certainly add the file to the tests and open a config PR

PR #2824, opened 2017-04-24
Fix acquisition dates format for Andor ABD TIFF format

sbesson:
See https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8238 for more context

The FluoviewReader is used for both Olympus FluoView and Andor ABD TIFF formats given their similarity. As reported in the OME forums, the date format for Andor files is wrongly parsed. This commit uses the proper convention and should fix the issue for Andor files.

Parallely to the open-source thread, we contacted Andor and received confirmation this is the expected behavior. A companion PR will be opened to update the configuration for existing Andor ABD TIFF files in our data repository.


dgault:
Change looks good  in accordance with the new information we have received. The repository jobs are green with the configuration PR included and the updates to the config files all look correct.

Both this and the associated config PR are good to merge.

PR #2825, opened 2017-04-24
OIR follow-up fixes

melissalinkert:
See https://github.com/openmicroscopy/data_repo_config/pull/193 and https://github.com/openmicroscopy/bioformats/pull/2814#pullrequestreview-34300051

Without this PR, ```showinf``` or opening in ImageJ should result in an exception for all 6 new .oir files noted in the configuration PR above.  With this PR, the same testing procedure used for #2814 should pass and repository tests of existing data should remain green.

Be aware that the screenshots and OME-TIFF exports are a little less complete this time.  One of the files (noted on the configuration PR) could not be opened in the Olympus ImageJ plugin, and the ```SampleImagesWS``` files open but do not allow metadata to be displayed.

Also note that the window shown in ImageJ when ```Display metadata``` is checked should now order all global metadata keys before any series metadata keys, instead of aggregating all and sorting alphabetically.  This is due to a0e1960 and does not affect other formats; happy to discuss other/better options, but that seemed like the least invasive approach since we probably don't want to change the sorting strategy for other formats.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -258,6 +258,10 @@ public class OIRReader extends FormatReader {
       }
       long fp = in.getFilePointer();
       String xml = in.readString(length);
+      if (!xml.startsWith("<?xml")) {
+        in.seek(fp - 2);
+        continue;
+      }
       LOGGER.trace("xml = {}", xml);
       if (channels.size() == 0 || getSizeX() == 0 || getSizeY() == 0) {
         parseXML(xml, fp);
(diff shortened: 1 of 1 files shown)


melissalinkert:
62bce84 fixes latest batch of files received 26 April, and configured in https://github.com/openmicroscopy/data_repo_config/pull/193/commits/90b1699be85dbde86ca247df6d86d9fcd5e2f411

dgault:
The code changes here look to be fine and the previous unused variables have been removed.

Testing with the new files from https://github.com/openmicroscopy/data_repo_config/pull/193/commits/90b1699be85dbde86ca247df6d86d9fcd5e2f411

Without the PR I was unable to open any of the the new images with a variety of exceptions:
`java.lang.IllegalArgumentException: Comparison method violates its general contract!`
`java.lang.IllegalArgumentException: 0 must not be null and positive.`
`java.lang.ArrayIndexOutOfBoundsException: -1`

With the PR all of the files open and display with sensible looking images and metadata. Where available these match with the associated ome.tiff or clipboard images. When viewing the metadata the global metadata was ordered first as expected. This approach makes sense to me and works well.

The only file I failed to open was Map_A01, but it is marked as test=false so Im assuming this is expected.

Testing the memorization with `showinf -cache` works as expected also. Builds and tests are green showing no regression in the original files.

melissalinkert:
Thanks, @dgault.  ```Map_A01.oir``` is indeed expected to not open - apologies for not clarifying.  The associated clipboard TIFF shows that it also fails to open in Olympus' ImageJ plugin, so I don't think there is much we can do about it.

PR #2826, opened 2017-04-25
Add format and metadata pages for Olympus OIR

melissalinkert:
See https://trello.com/c/DTO7sNTA/2-olympus-oir

The ```notes``` section in ```format-pages.txt``` in particular could use careful approval - we had received written permission to mention this, but maybe there is a better way to phrase it.


hflynn:
Looks good to me.

The note reads fine actually - I'm sure we could reword it if we wanted to but I'm not convinced we need to.

PR #2827, opened 2017-04-27
Operetta: make sure TIFF files exist before reading

melissalinkert:
See https://trello.com/c/iXCflqLT/588-hipsci-davide-danovi-idr0034-kilpinen-hipsci and dataset in ```data_repo/from_IDR/operetta```.

I haven't tested importing into OMERO, but ```setId``` should work and planes can be read.  As with other readers, blank planes are returned when an expected file is missing.


joshmoore:
Merging based on testing in test37

PR #2413, opened 2016-05-10
Columbus export read

joshmoore:
From the original PR by @melissalinkert :

> The file to import is MeasurementIndex.ColumbusIDX.xml, though I'd expect a directory import to work as well (if slightly slower). Choosing any of the TIFFs in the `$timestamp` subdirectory will not result in a plate import.

> This should match up with the number of TIFF files - one file per field and well, for N total TIFF files. Physical pixel sizes, channel names, wavelengths, and timestamps should all be present. As noted in the reader, the channel color is present in the dataset but I'm unsure of how to parse it at the moment, so the colors shown on import are coming from the emission wavelengths.

The datasets of interest for the IDR are:

 * https://trello.com/c/1cXLMIan/226-bakal-lab-nf-kappa-b-activation-idr0019-sero-nfkappab
 * https://trello.com/c/hSnz0Al1/229-bakal-lab-idr0020-barr-chtog

PR #2830, opened 2017-04-27
Columbus export read (develop)

simleo:
* Rebases https://github.com/openmicroscopy/bioformats/pull/2413 onto develop
* Adds fixes for `ColumbusReader.{isThisType,initFile}`
* Adds a Columbus-specific exception to `FormatReaderTest.testIsThisType`

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
@@ -210,21 +210,12 @@ public class ColumbusReader extends FormatReader {
 
   /* @see loci.formats.FormatReader#initFile(String) */
   protected void initFile(String id) throws FormatException, IOException {
-    // make sure that we have the XML file and not a TIFF file
-
-    if (!checkSuffix(id, "xml")) {
-      Location parent = new Location(id).getAbsoluteFile().getParentFile();
-      Location xml = new Location(parent, XML_FILE);
-      if (!xml.exists()) {
-        throw new FormatException("Could not find XML file " +
-          xml.getAbsolutePath());
-      }
-      initFile(xml.getAbsolutePath());
-      return;
+    Location xml = findXML(id);
+    if (null == xml) {
+      throw new FormatException("Could not find " + XML_FILE);
     }
-    else {
-      super.initFile(id);
-    }
+    id = xml.getAbsolutePath();
+    super.initFile(id);
 
     Location parent = new Location(currentId).getAbsoluteFile().getParentFile();
 
--- a/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
@@ -322,14 +322,19 @@ public class ColumbusReader extends FormatReader {
     Timestamp date = new Timestamp(acquisitionDate);
     long timestampSeconds = date.asInstant().getMillis() / 1000;
     int nextWellSample = 0;
+
+    int prevRow = -1, prevCol = -1;
     for (int plane=0; plane<planes.size(); plane++) {
       Plane p = planes.get(plane);
 
-      if (plane % (getImageCount() * nFields) == 0) {
+      if (p.row != prevRow || p.col != prevCol) {
         nextWell++;
         store.setWellID(MetadataTools.createLSID("Well", 0, nextWell), 0, nextWell);
         store.setWellRow(new NonNegativeInteger(p.row), 0, nextWell);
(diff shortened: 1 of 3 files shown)


components/formats-gpl/src/loci/formats/in/ColumbusReader.java
```diff
@@ -0,0 +1,737 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+ * %%
+ * Copyright (C) 2005 - 2014 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
```
dgault: Header can be updated to the 2017 version

components/formats-gpl/src/loci/formats/in/ColumbusReader.java
```diff
@@ -0,0 +1,737 @@
...
+
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+import org.w3c.dom.Attr;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
```
dgault: A few unused imports could be removed:
import org.w3c.dom.Attr;
import org.w3c.dom.Document;
import ome.xml.model.primitives.PositiveFloat;

components/formats-gpl/src/loci/formats/in/ColumbusReader.java
```diff
@@ -0,0 +1,737 @@
...
+                }
+                store.setChannelColor(p.channelColor, p.series, p.channel);
+              }
+
+              for (int t=0; t<getSizeT(); t++) {
+                p = lookupPlane(row, col, field, t, c);
+                if (p != null) {
+                  p.series = wellSample;
+                  store.setPlaneDeltaT(new Time(p.deltaT - timestampSeconds, UNITS.S), p.series, getIndex(0, c, t));
+                }
```
dgault: UNITS.SECOND would be preferred here as we have since deprecated UNITS.S

components/formats-gpl/src/loci/formats/in/ColumbusReader.java
```diff
@@ -0,0 +1,737 @@
...
+          p.positionZ = correctUnits(new Double(value), unit);
+        }
+      }
+
+      planes.add(p);
+    }
+
+  }
+
+  private Double correctUnits(Double v, String unit) {
```
sbesson: Not sure how much this would impact the Columbus samples and hence IDR  but should the metadata be stored using the OME units rather than converting into micrometers? /cc @joshmoore

joshmoore: I'd assume the impact wouldn't be overly large of converting. Existing filesets most likely would ignore the modification and happily stay in micrometers.

sbesson: Had a look at the data the metadata is stored in meters, meaning the values are low enough that they are below `Constants.EPSILON` and would be converted anyways. Given the late hour and the low impact, it might be easier to keep the implementation as it is.

components/formats-gpl/src/loci/formats/in/ColumbusReader.java
```diff
@@ -0,0 +1,737 @@
...
+    if (unit.equals("m")) {
+      return v * 1000000;
+    }
+    else if (unit.equals("cm")) {
+      return v * 10000;
+    }
+    else if (unit.equals("nm")) {
+      return v /= 1000;
+    }
+    return v;
```
dgault: Would it be worth using `unit.toLowerCase()` here or will the unit always be lower case?

dgault:
All the builds and tests are passing and the sample files can be read and imported correctly.
Other than s few minor comments the coding of the reader looks good and is well designed.

components/formats-gpl/src/loci/formats/in/ColumbusReader.java
simleo: They are lowercase in the examples we have. I'm not sure if we know in advance that they will always be lowercase, but they should be in order to be valid units. /cc @melissalinkert

sbesson:
As per the last runs of the repository jobs on Linux and Windows, everything seems to work as expected. Merging this so that we can move forward with the format documentation prior to release.

PR #2831, opened 2017-04-27
NDPIS: switch back to using ChannelSeparator

melissalinkert:
See https://github.com/openmicroscopy/bioformats/pull/2759#issuecomment-297320194 and https://trello.com/c/aUkOf0QT/10-ndpis-follow-up

@sbesson did most of the work in c8a6218, I just added a couple of commits to fix failing tests.  There is also a corresponding configuration PR: https://github.com/openmicroscopy/data_repo_config/pull/196

Things to check are that:
1. builds continue to pass
2. code changes look sensible
3. changed pixel data in series 5 of both ```test3.ndpis``` makes sense

Point 3 is the only reason a configuration PR was necessary.  Without this PR, the first two channels of series 5 were blank (```showinf -series 5 -minmax``` or ImageJ can confirm), with only the third channel containing data matching series 5 of the corresponding .ndpi file.  With this PR, each channel contains data which matches series 5 of the corresponding .ndpi file.  ```showinf -series 5 -minmax``` or ImageJ should show a maximum pixel value of 1 for all channels.  ```green redZ.ndpis``` should exhibit the same behavior with this PR, but did not require a configuration change, likely because of the difference in channel count and ordering.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -202,7 +202,7 @@ public class NDPISReader extends FormatReader {
 
     bandUsed = new int[ndpiFiles.length];
     for (int c=0; c<readers.length; c++) {
-      // // populate channel names based on IFD entry
+      // populate channel names based on IFD entry
       TiffParser tp = new TiffParser(ndpiFiles[c]);
       IFD ifd = tp.getIFDs().get(0);
 
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -112,9 +112,8 @@ public class NDPISReader extends FormatReader {
     int[] zct = getZCTCoords(no);
     int channel = zct[1];
     readers[channel].setId(ndpiFiles[channel]);
-    readers[channel].setSeries(getSeries());
-    readers[channel].setResolution(getResolution());
-    int cIndex = (channel < readers[channel].getSizeC()) ? bandUsed[channel] : 0;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/NDPISReader.java
```diff
@@ -206,20 +198,22 @@ else if (key.startsWith("Image")) {
...
+    MetadataTools.populatePixels(store, this);
+
     bandUsed = new int[ndpiFiles.length];
-    MetadataStore store = makeFilterMetadata();
-    for (int c=0; c<readers.length; c++) {     // populate channel names based on IFD entry
-      readers[c].setId(ndpiFiles[c]);
-      IFD ifd = readers[c].getIFDs().get(0);
-      samplesPerPixel[c] = ifd.getSamplesPerPixel();
+    for (int c=0; c<readers.length; c++) {
+      // // populate channel names based on IFD entry
```
dgault: Extra `//`

dgault:
The code changes here look fine. Builds and tests are all green.

Testing with `test3.ndpis` and without the PR only one channel was populated on series 5. With the PR included all 3 channels are populated and look to be correct when compared to the NDPI files. The new MD5 sums on the corresponding config PR also appear to be correct.

sbesson:
If we are pushing an extra commit for this PR, happy to provide an amendment of c8a6218 with a more informative commit message.

sbesson:
With the last commit, https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/417/ works as expected. Leaving @dgault to sign off on this PR before merging.

dgault:
Retested with `test3.ndpis` and everything is working as expected. Looks good to merge.

PR #2832, opened 2017-04-28
Docs: add orthogonal plane example

melissalinkert:
This adds an example class for reading and writing XZ and YZ planes, written by @emilroz.

The main file reading/API description page seemed like the best place to link it, but can certainly move or add links elsewhere if anyone has a better idea.

### DIFF
--- a/docs/sphinx/developers/file-reader.rst
+++ b/docs/sphinx/developers/file-reader.rst
@@ -104,7 +104,7 @@ See :doc:`Bio-Formats metadata processing </about/index>` for more information o
   :source:`IFormatReader <components/formats-api/src/loci/formats/IFormatReader.java>`
     Source code of the ``loci.formats.IFormatReader`` interface
   :download:`OrthogonalReader.java <examples/OrthogonalReader.java>`
-    Example of reading XY, XZ, and YZ image planes from a file
+    Example of reading XZ and YZ image planes from a file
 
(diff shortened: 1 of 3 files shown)


melissalinkert:
Updated to remove argparse4j dependency and add a case to ```ExampleSuite```.  We may want to consider a dependency on argparse4j in the future (when we have less pressing issues), so a "maybe later" card is now here: https://trello.com/c/MK3TAs02/467-update-command-line-tools-examples-to-use-argparse4j

docs/sphinx/developers/file-reader.rst
```diff
@@ -103,6 +103,8 @@ See :doc:`Bio-Formats metadata processing </about/index>` for more information o
 .. seealso::
   :source:`IFormatReader <components/formats-api/src/loci/formats/IFormatReader.java>`
     Source code of the ``loci.formats.IFormatReader`` interface
+  :download:`OrthogonalReader.java <examples/OrthogonalReader.java>`
+    Example of reading XY, XZ, and YZ image planes from a file
```
sbesson: Worth specifying that the example is reading and writing and only applies to XZ and YZ.

melissalinkert:
Last commits should address all comments above.

sbesson:
Thanks. All looks good. Merging.

PR #2833, opened 2017-05-03
Bump to 0.3.4-SNAPSHOT

joshmoore:
Unlikely to be needed until post POTN


joshmoore:
Merging this so that the branch is in a SNAPSHOT state even though further commits are not expected.

PR #2837, opened 2017-05-05
Docs: update section on adding format/reader docs

simleo:
Adds entries for `mif` and `pyramid` to `format-documentation.rst`.


sbesson:
Reads well to me. Thanks for keeping the doc in sync.

PR #2842, opened 2017-05-08
CZI: fix pyramid resolution indexing for pyramids of different depths

melissalinkert:
This prevents an infinite loop when a resolution in one of the pyramids
is missing, i.e. the pyramids are of different sizes/depths.

See https://trac.openmicroscopy.org/ome/ticket/13320 and
https://trello.com/c/S60YCf0H/95-fiji-zeiss-czi-axioscan-fails-to-open

To test, use the files from QA 17539 and QA 17711 (https://github.com/openmicroscopy/data_repo_config/pull/198).  Without this PR, ```showinf -nopix``` should enter an infinite loop around line 972.  With this PR, ```showinf -nopix``` should complete and ```showinf -nopix -noflat``` should show that the number of pyramids matches the number of scenes shown in ZEN.  The individual images should also match what is shown in ZEN, except for some of the larger resolutions which will show black pixels to indicate a missing tile instead of the white pixels displayed in ZEN.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -340,7 +340,10 @@ public class ZeissCZIReader extends FormatReader {
     boolean emptyTile = true;
     try {
       int minTileX = Integer.MAX_VALUE, minTileY = Integer.MAX_VALUE;
-      int baseResolution = (maxResolution + 1) * (currentIndex / (maxResolution + 1));
+      int baseResolution = currentIndex;
+      while (baseResolution > 0 && core.get(baseResolution - 1).sizeX > core.get(baseResolution).sizeX) {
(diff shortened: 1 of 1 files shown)


dgault:
The code changes here make sense and look to be fine. 

Builds and tests are passing, the existing CZI tests remain unaffected and the new configs from https://github.com/openmicroscopy/data_repo_config/pull/198 are passing - https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/422/

I tested with QA-17539 and QA-17711 via the command line and without the PR `showinf -nopix` entered an infite loop as expected. With the PR the command completed and using `showinf -nopix -noflat` appears to show the correct number of pyramids. When opening the images in FIJI with the PR applied and comparing to Zen the images import and display correctly (taking into account the inverse missing tile color).

This PR and the associated config look good to merge.

PR #2843, opened 2017-05-08
ImageInfo: fix image dimensions when the '-autoscale' option is used

melissalinkert:
See https://trello.com/c/kbzOM2NF/298-incorrect-dimensions-used-for-autoscaling-in-imageinfo

As noted on the card, the original exception can be reproduced without this PR using ```showinf -series 0 -autoscale -crop 0,0,512,512 data_repo/curated/hamamatsu/public/openslide/CMU-3.ndpi```, or any other use of ```-autoscale``` combined with ```-crop```.  With the PR, the same command should result in an image of the size specified by ```-crop```.  Note that ```-crop``` does not apply if ```-thumbs``` is also used, which is why there is a special case.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -855,9 +855,16 @@ public class ImageInfo {
             pix = DataTools.normalizeDoubles((double[]) pix);
           }
         }
-        images[i - start] = AWTImageTools.makeImage(ImageTools.make24Bits(pix,
-          sizeX, sizeY, reader.isInterleaved(), false, min, max),
-          sizeX, sizeY, FormatTools.isSigned(pixelType));
+        if (thumbs) {
+          images[i - start] = AWTImageTools.makeImage(ImageTools.make24Bits(pix,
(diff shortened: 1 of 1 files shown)


dgault:
The code changes look fine. I tested using the sample file from http://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/openslide/CMU-3/

Without the PR I see `NegativeArraySizeException` as reported in https://trello.com/c/kbzOM2NF/9-incorrect-dimensions-used-for-autoscaling-in-imageinfo
`Exception in thread "main" java.lang.NegativeArraySizeException
	at loci.formats.ImageTools.make24Bits(ImageTools.java:110)
	at loci.formats.ImageTools.make24Bits(ImageTools.java:75)
	at loci.formats.tools.ImageInfo.readPixels(ImageInfo.java:858)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1044)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1110)`

With the PR applied no exception is thrown and showinf completes with the correct metadata shown and an image the size of the crop region is displayed.

Using -crop without -autoscale also continues to behave correctly.

PR #2844, opened 2017-05-09
Autogen: gen-meta-support fix

sbesson:
See https://trello.com/c/GTMfjd5R/34-fix-bio-formats-autogen-jobs

Following the decoupling of the model components in Bio-Formats 5.3.0, the assumption that the generated ome-xml sources are available in the source tree is no longer correct. A long term solution might be some API to introspect the model elements.

As Bio-Formats is now following semantic versioning and the list of elements should only change for major version bumps, this PR tackles the issue by hardcoding the list of elements. This list will need to be reviewed during model development. 

Additionally this PR uses the logback.xml in the `ant` folder similarly to the other components to reduce the amount of logging generated by the `autogen` targets by default.

### DIFF
--- a/components/autogen/src/gen-meta-support.py
+++ b/components/autogen/src/gen-meta-support.py
@@ -69,14 +69,34 @@ def is_file(f, ftype=".java"):
 
 def get_xml_elements():
     """List all XML elements from the model"""
-    elements = []
-    modelDir = join(
-        componentsDir, 'ome-xml', 'build', 'src', 'ome', 'xml', 'model')
-    for f in sorted(listdir(modelDir)):
-        if not is_file(join(modelDir, f)):
-            continue
-        elements.append(basename(f).rstrip('.java'))
-    return elements
+
+    # Since Bio-Formats 5.3.0, the ome-xml component is decoupled from
+    # Bio-Formats. This function returns a hard-coded list of the OME-XML
+    # elements which should be updated for each model change
+    return [
+        'AffineTransform', 'Annotation', 'AnnotationRef', 'Arc',
+        'BasicAnnotation', 'BinDat', 'BinaryFile', 'BinaryOnly',
+        'BooleanAnnotation', 'Channel', 'ChannelRef', 'CommentAnnotation',
+        'Dataset', 'DatasetRef', 'Detector', 'DetectorSettings', 'Dichroic',
+        'DichroicRef', 'DoubleAnnotation', 'Ellipse', 'EmissionFilterRef',
+        'ExcitationFilterRef', 'Experiment', 'ExperimentRef', 'Experimenter',
+        'ExperimenterGroup', 'ExperimenterGroupRef', 'ExperimenterRef',
+        'External', 'Filament', 'FileAnnotation', 'Filter', 'FilterRef',
+        'FilterSet', 'FilterSetRef', 'Folder', 'FolderRef',
+        'GenericExcitationSource', 'Image', 'ImageRef', 'ImagingEnvironment',
+        'Instrument', 'InstrumentRef', 'Label', 'Laser', 'Leader',
+        'LightEmittingDiode', 'LightPath', 'LightSource',
+        'LightSourceSettings', 'Line', 'ListAnnotation', 'LongAnnotation',
+        'ManufacturerSpec', 'MapAnnotation', 'Mask', 'MetadataOnly',
+        'MicrobeamManipulation', 'MicrobeamManipulationRef', 'Microscope',
+        'NumericAnnotation', 'OME', 'Objective', 'ObjectiveSettings',
+        'Pixels', 'Plane', 'Plate', 'PlateAcquisition', 'PlateRef', 'Point',
+        'Polygon', 'Polyline', 'Project', 'ProjectRef', 'Pump', 'ROI',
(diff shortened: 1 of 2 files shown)


components/autogen/src/gen-meta-support.py
```diff
@@ -69,14 +69,34 @@ def is_file(f, ftype=".java"):
...
+        'ManufacturerSpec', 'MapAnnotation', 'Mask', 'MetadataOnly',
+        'MicrobeamManipulation', 'MicrobeamManipulationRef', 'Microscope',
+        'NumericAnnotation', 'OME', 'Objective', 'ObjectiveSettings',
+        'Pixels', 'Plane', 'Plate', 'PlateAcquisition', 'PlateRef', 'Point',
+        'Polygon', 'Polyline', 'Project', 'ProjectRef', 'Pump', 'ROI',
+        'ROIRef', 'Reagent', 'ReagentRef', 'Rectangle', 'Reference', 'Rights',
+        'Screen', 'Settings', 'Shape', 'StageLabel', 'StructuredAnnotations',
+        'TagAnnotation', 'TermAnnotation', 'TextAnnotation', 'TiffData',
+        'TimestampAnnotation', 'TransmittanceRange', 'TypeAnnotation', 'UUID',
+        'Union', 'Well', 'WellSample', 'WellSampleRef', 'XMLAnnotation']
```
rleigh-codelibre: Can't we get this directly from the ome-xml jar specified in the pom?

sbesson: As stated in the PR description, the robust mid-term solution probably involves introspection of the ome-xml JARs indeed. I did not find a straightforward way to achieve this without redesigning the full metadata generation workflow. Thus I limited the scope of this PR to fixing this component target and allow the generated metadata documentation pages to be kept up-to-date with the code again.

sbesson: Ah missed the edition of https://github.com/openmicroscopy/bioformats/pull/2844#discussion_r115682851. Relying on the local Maven repository path could work as suggested. The hard part is certainly to retrieve the path for a single dependency, is there a way to do that?

simleo:
In the hardcoded list, `'BinDat'` needs to be changed to `'BinData'`. The old version had a bug that caused both `'BinData'` and `'TiffData'` to be truncated in this way. The fix would be:

```
-        elements.append(basename(f).rstrip('.java'))
+        elements.append(splitext(basename(f))[0])
```

But we're not going to use that anymore anyway.

After the above correction, this PR is good to merge, assuming we don't want to go for Roger's solution immediately. Otherwise, we might close this and open a card for a future implementation of Roger's solution.

sbesson:
b7ff0999670c37020975b4fbcf0d76a5f084bd31 partly implements the suggestion of https://github.com/openmicroscopy/bioformats/pull/2844#discussion_r115682851 by introspecting the `ome-xml` JAR. There is still some level of hardcoding in the dependency but this should provide a more robust solution which will account for and test bump of the `ome-model` dependency.

components/autogen/src/gen-meta-support.py
```diff
@@ -71,32 +79,16 @@ def get_xml_elements():
...
-        'MicrobeamManipulation', 'MicrobeamManipulationRef', 'Microscope',
-        'NumericAnnotation', 'OME', 'Objective', 'ObjectiveSettings',
-        'Pixels', 'Plane', 'Plate', 'PlateAcquisition', 'PlateRef', 'Point',
-        'Polygon', 'Polyline', 'Project', 'ProjectRef', 'Pump', 'ROI',
-        'ROIRef', 'Reagent', 'ReagentRef', 'Rectangle', 'Reference', 'Rights',
-        'Screen', 'Settings', 'Shape', 'StageLabel', 'StructuredAnnotations',
-        'TagAnnotation', 'TermAnnotation', 'TextAnnotation', 'TiffData',
-        'TimestampAnnotation', 'TransmittanceRange', 'TypeAnnotation', 'UUID',
-        'Union', 'Well', 'WellSample', 'WellSampleRef', 'XMLAnnotation']
+    # Bio-Formats. This function returns introspect the OME-XML JAR under
```
simleo: "returns introspect" -> "introspects"

components/autogen/src/gen-meta-support.py
```diff
@@ -56,6 +58,12 @@
 # ImagingEnvironment.Temperature = Partial see ticket #167 for details
 
 """
+OMEXML_VERSION = sys.argv[1]
+OMEXML_PATH = join(
+    expanduser("~"), ".m2", "repository", "org", "openmicroscopy", "ome-xml",
```
rleigh-codelibre: This should likely be using the `maven.repo.local` ant property.  Couldn't the path to the jar be passed directly from ant?

sbesson:
Thanks for the feedback. Given the current bandwidth and priorities, I would propose we get this in as a first step to fix the currently broken logic and capture the hardcoding improvements separately.

I do not recall this script was working directly on the `MetadataStore.java` class  but only grepping all the `MetadataStore` setter calls within each concrete reader to list all the supported OME-XML elements. Which line were you referring to?

rleigh-codelibre:
Sorry, misread it, I was looking at `MetadataTools` which is fine.  Though ideally it should be migrated to ome-xml.

Issue #2847, opened 2017-05-10
Extracting sub images from a TIFF file results in an error


felixveysseyre:
Hello,

I am using _bfconvert_ executable to extract images from a TIFF file:

> bfconvert -bigtiff -series 0 "input.tiff" ".../output.%z.%t.%c.tiff"

The process fails quickly with the following error:
> TiffDelegateReader initializing input.tiff
> Reading IFDs
> Populating metadata
> Checking comment style
> Populating OME metadata
> [Tagged Image File Format] -> output.%z.%t.%c.tiff [Tagged Image File Format]
> Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
> 	at java.util.ArrayList.rangeCheck(ArrayList.java:635)
> 	at java.util.ArrayList.get(ArrayList.java:411)
> 	at ome.xml.model.Pixels.getChannel(Pixels.java:628)
> 	at ome.xml.meta.OMEXMLMetadataImpl.getChannelName(OMEXMLMetadataImpl.java:1211)
> 	at loci.formats.FormatTools.getFilename(FormatTools.java:1138)
> 	at loci.formats.FormatTools.getFilename(FormatTools.java:1101)
> 	at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:541)
> 	at loci.formats.tools.ImageConverter.main(ImageConverter.java:884)

I am using the latest version of BFTools:
> $ ~/Softwares/bftools/bfconvert -version
> Version: 5.5.0
> Build date: 8 May 2017
> VCS revision: d65f35a268ce2b23e13713d58ba24c4a18f3520c

I would like to add the image to this issue but it contains some non anonymized information. What is the correct procedure ?

Thanks for your time !

melissalinkert:
Thank you for reporting this problem, @felixveysseyre.  The best way to send the image file is to upload it to http://qa.openmicroscopy.org.uk/qa/upload/.  The OME team can then see it for testing, but it will not become publicly available without your permission.

felixveysseyre:
@melissalinkert Thanks for your help !

I just uploaded the image, the id was 17718.

dgault:
Thanks @felixveysseyre for submitting the sample file. I was able to test and reproduce the same error that you reported. From debugging the issue I believe I have found the root cause and should have a PR open shortly to put a fix in place.

felixveysseyre:
@dgault Thanks a lot for your time !

[...]

PR #2848, opened 2017-05-12
FormatTools - Chanel IndexOutOfBoundsException

dgault:
This PR is in relation to https://github.com/openmicroscopy/bioformats/issues/2847
An associated sample file can be found in QA-17718

The IndexOutOfBoundsException is thrown as the file has 4 samples per pixel but 1 effective channel. All of the other utility methods take a reader as an input parameter and use the following for dimensions:
`    int zSize = reader.getSizeZ();
    int cSize = reader.getEffectiveSizeC();
    int tSize = reader.getSizeT();`

In this case MetadataRetrieve is available but not the reader, as such getEffectiveSizeC or getImageCount are not able to be used. The `int sizeC = retrieve.getPixelsSizeC(series).getValue();` will return 4 which will then result in the out of bounds exception.

I have updated this to use `int sizeC = retrieve.getChannelCount(series);` which should be equivalent of `getEffectiveSizeC()`

To reproduce:
- Use bfconvert with the sample file provided and verify that an IndexOutOfBoundsException is thrown (a standard `bfconvert input output` is suitable)

To test:
- Ensure all existing tests remain green
- With the the PR applied retest the sample file and verify no errors are thrown
- Sanity test other samples to verify that the behaviour remains unchanged

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1118,7 +1118,7 @@ public final class FormatTools {
     filename = filename.replaceAll(SERIES_NAME, imageName);
 
     DimensionOrder order = retrieve.getPixelsDimensionOrder(series);
-    int sizeC = retrieve.getPixelsSizeC(series).getValue();
+    int sizeC = retrieve.getChannelCount(series);
     int sizeT = retrieve.getPixelsSizeT(series).getValue();
     int sizeZ = retrieve.getPixelsSizeZ(series).getValue();
     int[] coordinates = FormatTools.getZCTCoords(order.getValue(), sizeZ, sizeC, sizeT, sizeZ*sizeC*sizeT, image);


melissalinkert:
The code change makes complete sense.  Without this PR:

```
$ bfconvert input.tiff output_%z_%t_%c.tiff
input.tiff
TiffDelegateReader initializing input.tiff
Reading IFDs
Populating metadata
Checking comment style
Populating OME metadata
[Tagged Image File Format] -> output_%z_%t_%c.tiff [Tagged Image File Format]
	Converted 1/37 planes (2%)
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
	at ome.xml.model.Pixels.getChannel(Pixels.java:628)
	at ome.xml.meta.OMEXMLMetadataImpl.getChannelName(OMEXMLMetadataImpl.java:1211)
	at loci.formats.FormatTools.getFilename(FormatTools.java:1138)
	at loci.formats.FormatTools.getFilename(FormatTools.java:1101)
	at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:541)
	at loci.formats.tools.ImageConverter.main(ImageConverter.java:884)
```
With this PR, the same command produces 37 TIFFs with no exceptions.  With and without this PR, ```bfconvert -separate input.tiff output_%z_%t_%c.tiff``` correctly produces 148 TIFFs, so this has not affected behavior for the case when ```getPixelsSizeC``` is equal to ```getChannelCount```.

Assuming builds remain green, this looks good to merge to me.

PR #2849, opened 2017-05-15
Fix CellH5Reader.openBytes() for non-zero y values

sbesson:
See https://trello.com/c/Rlg1SdFD/12-cellh5-broken-tile-loading

While testing the new iviewer app on a CellH5 file, a bug was discovered related to opening of subsets of CellH5 files, namely `openBytes(r, no, x, y, w, h)` for `y>0`. 

This commit fixes this behavior by adjusting the origin of the block array to account for the vertical origin in the private `getImageData` method.

In addition to the automated testing (which should not be affected as we are not testing any non (0,0) based tiles, it might be useful to check that plane loading either using the full planes or tiles now works as expected. https://gist.github.com/sbesson/2d3ea28313dfdb2988df8b3b368a5432 shows the testing script used in MATLAB against a sample `.h5` data but can be converted in Java or other language.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellH5Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellH5Reader.java
@@ -337,7 +337,7 @@ public class CellH5Reader extends FormatReader {
 
     int elementSize = jhdf.getElementSize(CellH5PathsToImageData.get(series));
 
-    int[] arrayOrigin = new int[] {channel, time, zslice, 0, 0};
+    int[] arrayOrigin = new int[] {channel, time, zslice, y, 0};
     int[] arrayDimension = new int[] {1, 1, 1, height, width};
 
     MDIntArray test = jhdf.readIntBlockArray(CellH5PathsToImageData.get(series),


melissalinkert:
Would it be worth adding one of the .ch5 files to https://github.com/openmicroscopy/data_repo_config/blob/master/ci/performance_files.txt, since https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-performance is meant to catch issues like this?

melissalinkert:
After brief discussion with @sbesson, see https://github.com/openmicroscopy/data_repo_config/pull/201

simleo:
Tested with the [MATLAB script](https://gist.github.com/sbesson/2d3ea28313dfdb2988df8b3b368a5432) on `curated/cellh5/samples/full/0038.ch5`. Without this PR (libs built from develop) I got four zeros, with it I got four ones.

:+1:

sbesson:
In addition to the testing in https://github.com/openmicroscopy/bioformats/pull/2849#issuecomment-301769434, a sample CellH5 file was added to the performance tests to catch future regressions. This [build log](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-performance/215/) shows the failing`openBytes` tests without this PR while the same tests are passing in [this morning build](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-performance/576) with this PR included.

PR #2857, opened 2017-05-26
Metadata ratings update

melissalinkert:
See https://trello.com/c/WmLyfC5N/61-rfe-update-format-metadata-support-ratings

This adds a new ```gen-metadata-ratings``` target in the ```autogen``` component, which will calculate the correct metadata rating for each format and update ```format-pages.txt``` accordingly.  The developer documentation has also been updated to reference this target and the algorithm for setting the metadata rating.

7db9f62 is the result of running ```ant gen-metadata-ratings gen-format-pages```, so all format pages should be up to date with the new rating policy.


joshmoore:
There seem to be a handful of 2+ jumps (very good --> fair). Do they (especially the downgrades) match expectations?

sbesson:
In addition to the specific question about the rating changes mentioned in https://github.com/openmicroscopy/bioformats/pull/2857#issuecomment-304593474, another  thought while reading this PR is: if the metadata ratings are now automatically generated, should we completely drop them from the `format-pages.txt` source file and combine the new target with the existing `gen-format-pages`?

melissalinkert:
Few commits pushed to refine the rating algorithm, which reduces the number and severity of the changes.  There are still some +/- 2 changes, but most seem to be for graphics formats (JPEG-2000, JPEG, BMP) or legitimate corrections (e.g. Imacon, Nikon NEF).  If there are any that are specifically worrying, though, I'll keep working on it.

It would certainly make sense to drop the metadata rating from ```format-pages.txt``` in the long term, I just thought this might be easier to review as a first step.  Happy to make that change here once the new ratings are approved, or as a follow-up PR.

sbesson:
Looking only at the 2+ changes, we have

```
AIM -3
AVI -3
Axon +2
MRI -2
CellH5 +2
Cellomics +3
FITS +2
GIF -2
Imacon -2
ImagePro Sequence -3
ImagePro Workspace -3
JPEG-200 +4
JPEG +2
Leo +2
Li-COR +2
MIAS +2
Minolat -2
Nef -3
ND2 -2
Quicktime -2
SimplePCI +2
BMP -3
Woolz +2
```

Generally, I certainly think exposing some rules both ath the generation component and in the documentation itself goes towards an increased transparency and sets a valuable basis of discussion for the future.

No veto to getting this merged in its current state and refining in the future. The `autogen` component should probably be reviewed (Mavenized?) to reduce the number of targets but in the meantime, we can certainly add `gen-metadata-ratings` to the jobs to make sure this gets executed.

PR #2855, opened 2017-05-26
NDPISReader: make sure TiffParser streams are closed

sbesson:
Performance tests highlighted some file handle leaks in the NDPISReader initialization step. This commit should ensure all resources are closed by created the RandomAccessInputStream with a try-with-resources block before creating each TiffParser.

PR #2859, opened 2017-06-03
Fix VeecoReader openBytes for tiles where y is larger than 0

sbesson:
VeecoReader.openBytes needs to flip the image retrieved using the HDF service vertically to match the expected origin in Bio-Formats. This commit should fix the implementation to work for any version of (y,w) when retrieving a tile and fix the openbytes tests.


dgault:
The previous logic here was certainly incorrect and the adjusted logic looks good. I tested manually with the milena and the images open and display correctly, I then manually tested some imports while using cropped regions and the displayed sections opened and displayed correctly with data from the correct regions.

sbesson:
Merging as per https://github.com/openmicroscopy/bioformats/pull/2855#issuecomment-306738096

Issue #2613, opened 2016-10-13
MetamorphReader reads data from unrelated tif files


imagejan:
When there are additional files in a folder containing a Metamorph dataset (`nd` file + several `stk` files), it can happen that the Reader gets confused by tif images having a similar name. I uploaded `nd_issue.zip` via http://qa.openmicroscopy.org.uk/qa/upload/ to reproduce the issue (sorry, I must have missed how to keep track of the ticket number there).

![openndfailing](https://cloud.githubusercontent.com/assets/2033938/19345629/80850226-913f-11e6-90ce-a219f257077d.png)

With the tif file present, opening the nd file in Fiji results in entirely white preview thumbs in the series dialog, and in some artifacts in the opened image, apparently the pixel intensity scaling is incorrect and gets wrapped around the UnsignedShort limit.

Renaming the `.tif` file to `.tmp` restores the correct, expected behavior.

![openndcorrect](https://cloud.githubusercontent.com/assets/2033938/19345732/18a6880e-9140-11e6-84ca-628cf4e536db.png)

I assume there's something wrong in how the reader is looking for the STK files here:

https://github.com/openmicroscopy/bioformats/blob/fff8e47a3a3490343005b7b26d0158b5075ef01e/components/formats-gpl/src/loci/formats/in/MetamorphReader.java#L1345-L1406

In particular, how the name prefix and suffix is generated:

```
String prefix = name.substring(0, name.indexOf('_'));
String suffix = name.substring(name.indexOf('_'));
```

dgault:
Hi Jan,

Thanks for providing the sample fileset for testing this issue. I have been able to reproduce the problem just as you described and have spent time debugging through the code to locate the cause. 

It appears that the reader is able to locate the stk files successfully with and without the tif present but it also sees the tif file as part of the fileset. Was the tif supposed to be part of the same fileset with the nd and stk files? Or was it simply a similarly named file in the same location?

imagejan:
Hi David,

In the real case that I used to compile the above example dataset, we have a series of (similarly named) nd datasets that are being stitched using Fiji and the Grid/Collection Stitching plugin (using bioformats in the backend). For some of the datasets, we write multi-channel tiff files with Fiji for each tile (=set of one stk file per channel) to improve stitching quality and avoid having to stitch the channels separately one after each other.

So, in short, the tif file(s) were not supposed to belong to the dataset, but were derived from one of the other nd/stk datasets in the same folder.

Does that make sense?

[...]

PR #2860, opened 2017-06-05
MetaMorph ND file matching

dgault:
This PR is being used as a test for an issue raised in https://github.com/openmicroscopy/bioformats/issues/2613

Previously the reader matched associated files based on the first underscore, from the ND documentation we have it only mentions associated files as having the possibility of `_w#wavename, _s#, _t#`. Testing this on breaking against all existing sample data to see if we can change the current behaviour.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -1361,6 +1361,8 @@ public class MetamorphReader extends BaseTiffReader {
     String parent = l.getParent();
 
     if (name.indexOf('_') > 0) {
+      String pattern = "_[swt]\\d+";
+      System.out.println(name.replaceAll(pattern, ""));
       String prefix = name.substring(0, name.indexOf('_'));
       String suffix = name.substring(name.indexOf('_'));
 


dgault:
Removing from breaking as https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-breaking-repository-subset/265/ is green

imagejan:
IIUC, this change wipes out all occurrences of `_w#`, `_s#` and `_t#` and then still checks for the *first* occurrence of an underscore to determine prefix and suffix.

The problem with filenames containing one or multiple underscores in their basename persists: for example, `day01_myTest3_slide1.nd` would result in the prefix `day01` being tested, and unrelated stk or tif files from e.g. `day01_myTest1_slide1.nd` being falsely picked up.

components/formats-gpl/src/loci/formats/in/MetamorphReader.java
```diff
@@ -1361,6 +1361,8 @@ private String getRealSTKFile(Location l) {
     String parent = l.getParent();
 
     if (name.indexOf('_') > 0) {
+      String pattern = "_[swt]\\d+";
+      System.out.println(name.replaceAll(pattern, ""));
```
sbesson: Can this `System.out.println` be removed (or replaced using logging at a debug level minimally) as it clutters the console of the automated builds?

dgault: Yeah sorry, it shouldn't have been left in

PR #2861, opened 2017-06-06
Bump to 0.4.0

joshmoore:
Bump in order to release that state as of ome2017.

cc: @sbesson

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>idr</groupId>
     <artifactId>pom-bio-formats</artifactId>
-    <version>0.4.0-SNAPSHOT</version>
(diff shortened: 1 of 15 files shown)


components/formats-gpl/pom.xml
```diff
@@ -59,7 +59,7 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>ome</groupId>
+      <groupId>idr</groupId>
```
sbesson: This needs to be reverted (and maybe accounted for in the script `groupId` functionality)

sbesson:
Latest change makes sense to me and a7bbc22 should prevent the extra commits in the future (although there might be a cleaner way to do so). Good to see that Travis is green on this branch to pick up these failures.

PR #2863, opened 2017-06-08
Fix NPE in OMEXMLServiceImpl

simleo:
`OMEXMLServiceImpl.transformToLatestVersion` is currently not handling `null` results from `getOMEXMLVersion`. This leads to an unhandled `NullPointerException` propagating to upper layers. For instance, in `OMETiffReader` the NPE is thrown [here](https://github.com/openmicroscopy/bioformats/blob/0e53c11bea06021d941007fe14531104e9881b3c/components/formats-bsd/src/loci/formats/in/OMETiffReader.java#L487) but the block expects to catch a `ServiceException`.

With this PR, the `null` value is handled and a `ServiceException` is thrown. New unit tests have been added to test both this specific scenario and `transformToLatestVersion`'s "normal" functionality.

### DIFF
--- a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
+++ b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
@@ -198,7 +198,7 @@ public class OMEXMLServiceImpl extends AbstractService implements OMEXMLService
   public String transformToLatestVersion(String xml) throws ServiceException {
     String version = getOMEXMLVersion(xml);
     if (null == version) {
-      throw new ServiceException("Could not get OME XML version");
+      throw new ServiceException("Could not get OME-XML version");
     }
     if (version.equals(getLatestVersion())) return xml;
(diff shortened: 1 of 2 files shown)


components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
```diff
@@ -197,6 +197,9 @@ public String getLatestVersion() {
   @Override
   public String transformToLatestVersion(String xml) throws ServiceException {
     String version = getOMEXMLVersion(xml);
+    if (null == version) {
+      throw new ServiceException("Could not get OME XML version");
```
mtbc: we usually hyphenate OME-XML

dgault:
Without the PR there is certainly the possibility of `getOMEXMLVersion` returning `null` and throwing a `NullPointerException`. Throwing a `ServiceException` in this case makes more sense and is better handled.

The new unit tests look fine to test either scenario and are running successfully. Looks good to merge.

PR #2858, opened 2017-06-02
FV1000Reader: fix "---" filter models

stelfrich:
Under some circumstances, filter names in OIBs/OIFs are `"---"` instead of `""`. This triggers parsing of the transmittance range cut-in and cut-out values from the name which raises an ArrayIndexOutOfBoundsException in [FV1000Reader.java#L1103](https://github.com/openmicroscopy/bioformats/blob/9088cd5a056d742066d2befdf19a83c71ec76e5d/components/formats-gpl/src/loci/formats/in/FV1000Reader.java#L1103).

PR #2865, opened 2017-06-09
Fv1000: channel metadata population refactoring

sbesson:
Following https://github.com/openmicroscopy/bioformats/pull/2858, this PR reviews the channel metadata population section of the `FV1000Reader` to use `DataTools.parseDouble`when appropriate and move the filter checking  logic to the `ChannelData` setters.

With this PR included, the automated tests should keep passing for all `oib/oif` samples.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
@@ -536,10 +536,9 @@ public class FV1000Reader extends FormatReader {
       IniTable guiChannel = f.getTable("GUI Channel " + index + " Parameters");
       while (guiChannel != null) {
         ChannelData channel = new ChannelData();
-        String gain = guiChannel.get("AnalogPMTGain");
-        if (gain != null) channel.gain = new Double(gain);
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Overall, the code looks fine, but the functions `getBarrierFilter`, `setEmissionFilter` and `setExcitationFilter` look redundant.  Since they are just one line functions, what's the advantage of these functions over simply calling `getFilter` inline three times?  The latter would overall be simpler and more readable.

sbesson:
@rleigh-codelibre thanks for the feedback, concretely, you propose something along the lines of:

```
- channel.setBarrierFilter(guiChannel.get("BF Name"));
+ channel.barrierFilter = channel.getFilter(guiChannel.get("BF Name"));
```

making `ChannelData.getFilter(String)` public and removing the setters. Is that correct? If so, no objection to moving to this representation if it improves clarity.

rleigh-codelibre:
That's exactly what I meant.  No objection to keeping the setters if they have value, I just wasn't sure what purpose they were serving, so I think it would be clearer with the change if that's OK.

dgault:
With the latest commit this looks fine. Builds and tests are green so it is good to merge.

PR #2867, opened 2017-06-10
Allow companion file to wrap plain TIFFs

simleo:
See https://trello.com/c/wBeP0xmA/4-companion-fileset-relaxation-investigation and the "External Companion File" design proposal. With this PR, when `initFile` is called on a companion file, it sets the metadata file as the companion itself and moves forward, instead of recursively calling `initFile` on the first TIFF referenced by the companion.

See data_repo_config PR no. 211 for a test case.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -470,15 +470,8 @@ public class OMETiffReader extends FormatReader {
       companion = true;
     }
     else {
-      RandomAccessInputStream ras = new RandomAccessInputStream(fileName, 16);
-      try {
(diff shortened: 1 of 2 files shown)


simleo:
First stab on the write side: https://gist.github.com/simleo/7445f1fcdf969fdcc2fd9c8634199895. Creates an XML companion associated to a single plain TIFF file, without modifying it and without writing any binary data. The resulting dataset (companion + original plain TIFF) should be readable with this PR and yield the same data and metadata as the original file.

sbesson:
Closing for now as this is not ready to merge and will probably require a full review of the companion specification. We can re-use this work as a starting point next time when we need to generate metadata-rich datasets from a set of existing TIFFs.

PR #2868, opened 2017-06-13
ND2 scanline pad fix

sbesson:
See https://trello.com/c/TaqXNGco/152-nd2-pixel-offset.

From the Nikon ND2 samples in our repository, it seems like having an odd size both alongside the X dimension and the C dimension is necessary and sufficient to have the scanline padded.

This PR simplifies the `NativeND2Reader.getScanlinePad()` method to use this simpler calculation. To test this PR, check the automated tests keep passing against all ND2 samples. Additionally, test the sample file provided in the [Fiji forum thread](http://forum.imagej.net/t/importing-nd2-files-results-in-horizontal-lines-for-channels/5607) and check that without this PR, the image display with some horizontal lines while with this PR, it matches the expectation from NIS Elements.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -2326,15 +2326,9 @@ public class NativeND2Reader extends FormatReader {
   }
 
   private int getScanlinePad() {
-    int scanlinePad = isJPEG ? 0 : getSizeX() % 2;
-    if (scanlinePad == 1) {
-      if (split && !isLossless && ((nXFields % 2) != 0 ||
-        (nXFields == 0 && (getSizeC() >= 4 || getSizeC() == 2))))
-      {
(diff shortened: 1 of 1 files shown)


dgault:
Testing the sample file from the forum thread shows that the scanlines are present without this PR but with the PR applied the file is read and displayed correctly. I additionally tested a number of other nd2 files from our repo and all continued to read and display correctly also. 

The subset jobs are green and the new config file also passes testing - https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/444/

I have been trying to locate the addition of this logic originally and although the `getScanlinePad` function was added back in 2014 the logic itself was being used dating further back still.

Based on the sample files and information that we have this relaxation of the `getScanlinePad` rules appears to work as expected. Once the merge full job is back to normal this should be good to merge.

dgault:
Daily builds and tests are all back to normal and previous issues were unrelated to this PR.

PR #2871, opened 2017-06-14
DICOM: do not allow core metadata to be changed in isThisType

melissalinkert:
This adds a new getNextTag(RandomAccessInputStream, boolean) signature,
which allows the caller to control whether or not the core metadata can
be altered.

See https://trello.com/c/apvieSPU/154-bug-dicom-reader and https://github.com/openmicroscopy/data_repo_config/pull/212

Without this change and with https://github.com/openmicroscopy/data_repo_config/pull/212, the subset job will fail on ```curated/dicom/public/samples``` as described in the config PR comments.  With this change, the subset job and full repository jobs should pass.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -191,7 +191,7 @@ public class DicomReader extends FormatReader {
     stream.seek(0);
 
     try {
-      int tag = getNextTag(stream);
+      int tag = getNextTag(stream, false);
       return TYPES.containsKey(tag);
     }
     catch (NullPointerException e) { }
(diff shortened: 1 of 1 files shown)


sbesson:
https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/442 looks good. Will review the proposed changes into detail tomorrow.

melissalinkert:
```addFileToList()``` is only called in the context of ```setId``` (via ```scanDirectory``` and ```makeFileList```), so I think using ```getNextTag()``` there is safe since it won't affect the behavior datasets other than the one being initialized.

sbesson:
Thanks for the precision. Makes sense to me. Merging.

PR #2874, opened 2017-06-15
CZI: store bezier ROIs as Polygons

melissalinkert:
Resolves point 7 of http://trac.openmicroscopy.org/ome/ticket/12935.  See also https://trello.com/c/y6KpD3sU/4-czi-tickets

To test, use ```data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 blue LSM 800 -May2015/ConfocalCh1T15Z1Bleaching2ROIS/ConfocalCh1T15Z1Bleaching2ROIS.czi```.  Opening in ZEN should show two ROI shapes in red: one rectangle and one closed contour.  Without this PR, ```showinf -nopix -omexml``` should only show the rectangle shape being parsed; opening in ImageJ with ```Display ROIs``` checked can also confirm this.

With this PR, the same test should show an additional polygon shape which approximates the shape shown in ZEN.  As far as I can see, the points defined in the .czi file are the control points for a set of Bezier curves, which is why the shapes will not exactly match.  I'm not sure if logging a warning would be useful in this case, but it would be easy enough to add instead of or in addition to ffbfbb1.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -2444,6 +2444,12 @@ public class ZeissCZIReader extends FormatReader {
           shape = populatePolylines(closedPolylines, i, shape, true);
         }
 
+        NodeList beziers =
+          getGrandchildren(layer, "Elements", "Bezier");
+        if (beziers != null) {
+          shape = populatePolylines(beziers, i, shape, true);
+        }
+
         NodeList rectRoi = getGrandchildren(layer, "Elements", "RectRoi");
         if (rectRoi != null) {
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
Checked with ZEN and in imagej with and without this PR.  The spline is imported as a polyline as described.

We already take this approach with ZVI/Zeiss-TIFF, but AxioVision uses a (IMO) better algorithm here which is better suited to fitting shapes in an image and for conversion to polyline.  The CZI/ZEN algorithm does odd things with the tension with no visible additional control points to intuit the behaviour, and the control points aren't always at the apex of a curve.  This makes the shapes slightly less representative of the overall ROI shape when converted to polyline.  Probably the best we can do right now though.

Looks absolutely fine to merge to me.

PR #2875, opened 2017-06-15
Kodak .bip: enforce stricter file type checking

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13321.  See also https://trello.com/c/tcGhzBQl/96-bug-kodakreader

To test, use the .bip file from QA 17540.  Without this change, ```showinf``` should throw an ```IllegalArgumentException``` as noted in the ticket.  The file is not a Kodak .bip file, however, as it only contains unrelated text data (this can be confirmed by opening in a text editor).  With this change, ```showinf``` should correctly throw ```UnknownFormatException```.


rleigh-codelibre:
The change looks fine, and testing shows the behaviour described.  Please merge.

PR #2876, opened 2017-06-19
Andor SIF: fix support for images that were cropped

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13313

The ```Pixel number``` line contains the dimensions of the originally
acquired image, which may not match what is actually stored in the file.
The line following ```Pixel number``` contains the bounding box of the stored image.

To test, use the file from QA 17457.  Without this change, ```showinf``` should throw an exception as indicated in the ticket.  With this change, the file should initialize without exception and result in 500 timepoints of 181x181 image planes.  As this is floating point data, it may be easier to check the actual image data in ImageJ instead of ```showinf```.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SIFReader.java
@@ -121,9 +121,6 @@ public class SIFReader extends FormatReader {
           m.imageCount = getSizeZ() * getSizeT() * getSizeC();
           timestamp = new double[getImageCount()];
           endLine = lineNumber;
-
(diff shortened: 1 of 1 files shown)


dgault:
Without the PR, importing the file from QA-17457 results in the following exception:
`java.lang.IllegalArgumentException: Negative position`

With the PR applied there is no exception and the file imports and displays correctly. Running the subset job and the andor and andor-sif tests all continue to pass with the PR included - https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/445/

PR #2877, opened 2017-06-19
build: Update maven plugin versions

rleigh-codelibre:
Needed as a first step to build with Java 9.  This is a routine update of the maven plugins to new versions.  Older versions of specific plugins don't work with Java 9.

This doesn't add Java 9 support; it makes it possible to start building with Java 9.  Additional fixes will be needed in a followup PR to complete the support.

If you run `mvn versions:display-plugin-updates` you'll see that the plugins are all up to date.

### DIFF
--- a/docs/sphinx/pom.xml
+++ b/docs/sphinx/pom.xml
@@ -59,7 +59,6 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>exec-maven-plugin</artifactId>
-        <version>1.2.1</version>
         <executions>
           <execution>
             <phase>test</phase>
--- a/docs/sphinx/pom.xml
+++ b/docs/sphinx/pom.xml
@@ -57,6 +57,10 @@
     </resources>      
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+      </plugin>
+      <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>exec-maven-plugin</artifactId>
         <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
     <ome-poi.version>5.3.1</ome-poi.version>
     <ome-mdbtools.version>5.3.0</ome-mdbtools.version>
     <jxrlib.version>0.2.1</jxrlib.version>
+    <xalan.version>2.7.2</xalan.version>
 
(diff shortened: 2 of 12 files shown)


docs/sphinx/pom.xml
```diff
@@ -59,7 +59,6 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>exec-maven-plugin</artifactId>
-        <version>1.2.1</version>
```
melissalinkert: Should this be set to ```1.6.0``` as in the root POM?

rleigh-codelibre: I'm no maven expert, so may be wrong, but I *think* by removing the version it inherits the 1.6.0 version from the root pom.  That was certainly what the `mvn versions:display-plugin-updates` output seemed to indicate.

pom.xml
```diff
@@ -237,7 +237,7 @@
           <artifactId>maven-javadoc-plugin</artifactId>
           <!-- NB: The same version declaration and configuration block also
                appears in the <reporting> section, and must be kept in sync. -->
```
mtbc: looks like this still needs doing

docs/sphinx/pom.xml
mtbc: That is the impression I get from `mvn help:effective-pom`.

Note that `components/formats-{api,bsd}/pom.xml` also make use of a `[1.2.1,)` version specifier.

pom.xml
rleigh-codelibre: Fixed in a new commit.

docs/sphinx/pom.xml
rleigh-codelibre: These are inside the eclipse m2 lifecycle-mapping plugin, which I've moved out into a eclipse-specific build profile.  I've updated the versions.

I'm unsure this is used by anyone on the team.  It's only been set up for formats-api and formats-bsd, which makes me think this is maybe scifio-related.  It looks incomplete though.

If I delete all this me2 lifescycle stuff, it still imports into eclipse and builds without trouble, so I'm unsure what specific value this adds.

rleigh-codelibre:
@mtbc Should all be fixed up now.

There's a warning from the version:display-plugin-updates plugin due to [this issue](https://stackoverflow.com/questions/29344809/versionsdisplay-plugin-updates-does-not-understand-maven-enforcer-plugin), which is harmless.  We'll need to wait on a fixed plugin to remove the warning.

sbesson:
As this is conflicting with #2883 which has priority, closing this PR for now. We can reconsider it  when we start work on Java9 across the board.

PR #2878, opened 2017-06-19
Bio-Rad Gel: fix width of pixel data offset field

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13094

The ```skip``` field used to calculate the pixel data offset is 4 bytes wide.  Previously submitted files only required the lower 2 bytes, though, so it had been assumed to be 2 bytes wide.

To test, use the file from QA 16880.  Without this change, ```showinf``` should result in an exception as indicated in the ticket.  With this change, ```showinf -autoscale -minmax``` and/or ImageJ should display a reasonable-looking gel image without throwing an exception.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/BioRadGelReader.java
+++ b/components/formats-gpl/src/loci/formats/in/BioRadGelReader.java
@@ -90,6 +90,9 @@ public class BioRadGelReader extends FormatReader {
       // offsets were determined by trial and error, and may not be 100% correct
       if (diff < 0) {
         in.seek(0x379d1);
+        if (in.getFilePointer() + planeSize > in.length()) {
+          in.seek(PIXEL_OFFSET + 62);
+        }
       }
       else if (diff == 0) {
         in.seek(PIXEL_OFFSET);
@@ -141,6 +144,7 @@ public class BioRadGelReader extends FormatReader {
 
     boolean codeFound = false;
     int skip = 0;
+    long baseFP = 0;
 
     while (!codeFound) {
       short code = in.readShort();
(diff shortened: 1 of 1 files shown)


mtbc:
Reading failed rather: [debug.txt](https://github.com/openmicroscopy/bioformats/files/1101542/debug.txt)

melissalinkert:
@mtbc: that's definitely the exception I'd expect to see without this PR, but if you're still seeing it with this PR then that's a bigger problem.  Could you please confirm that ```showinf -version``` shows a commit that would include this PR?  If it does, and you performed the build locally, you might try removing cached dependency files (```find bioformats -name *-dependencies.xml -exec rm {} \;```) and rebuilding to see if that helps.

mtbc:
```
$ showinf -version
Version: 5.5.3-SNAPSHOT
Build date: 26 June 2017
VCS revision: 233a0ada8ff2df12b5c3bf4787eacad322b4e351
$ git log 233a0ada8ff2df12b5c3bf4787eacad322b4e351 | grep 13094
    Fixes https://trac.openmicroscopy.org/ome/ticket/13094
```

mtbc:
Last time I just used `ant clean`. I'll try with `git clean -dfx`...

mtbc:
Ha, that works, now I see a gel. Maybe `ant clean` needs to clean more? Thank you for the hint about cached files.

sbesson:
@mtbc did you have a chance to capture an exact sequence of calls allowing to reproduce? I think I have encountered similar issues before I could not reproduce using sucessive runs `ant clean jars/tools` after switching branches.

mtbc:
I didn't, I'm afraid, but will look out for the issue reappearing.

PR #2879, opened 2017-06-20
MINC metadata parsing improvements

melissalinkert:
See https://trello.com/c/9iQdOahX/52-qa-17215-minc-metadata

d3b69ef addresses point 1 on the card; 52050fc addresses point 3, and fa6bab1 addresses point 2.  Only d3b69ef affects the configurations of existing MINC files.

To test, use the file from QA 17215 as indicated on the card.  Without this change, ```showinf -omexml``` should show a uint16 image with ```PhysicalSizeX```, ```PhysicalSizeY```, and ```PhysicalSizeZ``` reported in micrometers and no ```Plane``` elements.  With this change, the pixel type should be int16, all physical sizes should be reported in millimeters, and there should be one ```Plane``` and associated ```Position{X,Y,Z}``` for each image.  The original metadata and output of ```mincheader``` and ```mincinfo``` referenced on the card should confirm that the new metadata is correct.

The pixel type change in fa6bab1 is arguably breaking, though it doesn't affect any previously-configured files (all are unsigned).  If anyone is uncomfortable with having that change in a patch release, I can revert and open a separate PR for the next minor release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MINCReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MINCReader.java
@@ -226,9 +226,9 @@ public class MINCReader extends FormatReader {
       throw new FormatException(e);
     }
 
-    Double physicalX = null;
-    Double physicalY = null;
-    Double physicalZ = null;
+    Length physicalX = null;
+    Length physicalY = null;
+    Length physicalZ = null;
 
     if (isMINC2) {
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
The code looks fine, and the tests all pass as indicated.  Please merge.

PR #2880, opened 2017-06-23
CZI: populate Pixels.TimeIncrement

melissalinkert:
ZEN often shows timestamps calculated from the T increment, instead of the actual timestamps that are stored with each plane.  We now store the increment in Pixels.TimeIncrement, so that it is easy to work with either timestamp interpretation.

Resolves point 8 of http://trac.openmicroscopy.org/ome/ticket/12935

The new ```timeIncrement``` variable is not needed after ```setId```, so it is ```transient``` to be non-breaking.

To test, use ```data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 blue LSM 800 -May2015/ConfocalCh1T15Z1Bleaching2ROIs/ConfocalCh1T15Z1Bleaching2ROIs.czi```.  Open in ZEN, then mouse over the ```Time``` slider and observe the ```dt``` value in the mouse-over label at various timepoints:

clipboard

Compare with the value of ```TimeIncrement``` on ```Pixels``` as reported by ```showinf -nopix -omexml``` with this change.  ```dt``` at the second timepoint should match ```TimeIncrement```, and subsequent values of ```dt``` should be a multiple of ```TimeIncrement```.  Without this change, ```TimeIncrement``` is not populated.  With or without this change, ```showinf -nopix -omexml``` should report the same values for ```DeltaT``` on ```Plane```.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -142,6 +142,7 @@ public class ZeissCZIReader extends FormatReader {
   private String userInstitution;
   private String temperature, airPressure, humidity, co2Percent;
   private String correctionCollar, medium, refractiveIndex;
+  private transient Time timeIncrement;
 
   private String zoom;
   private String gain;
@@ -504,6 +505,7 @@ public class ZeissCZIReader extends FormatReader {
       positionsZ = null;
       zoom = null;
       gain = null;
+      timeIncrement = null;
 
       channels.clear();
       binnings.clear();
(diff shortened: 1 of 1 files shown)


dgault:
The code here looks to be fine and the builds and tests are all green. Marking the field as transient to prevent serialization breakage is valid in this instance.

With this PR the TimeIncrement and TimeIncrementUnit appear on the Pixels object. With the recommended sample file the values are as follows:
`TimeIncrement="1.9651124000000002" TimeIncrementUnit="s"`

This matches the value as seen in Zen. The dt in Zen then increases by a multiple of TimeIncrement, whereas the DeltaT timestamps on the Plane objects look somewhat off. I assume this is the 'questionable timestamps' mentioned on the Trac ticket and is expected.

melissalinkert:
Yes, the discrepancy between ```DeltaT``` and the ```dt``` in ZEN is expected.  The ```DeltaT``` values are coming from the timestamps stored separately for each plane, which I'd guess are the actual plane acquisition times (with ```dt``` being the expected time step).

sbesson:
Discussed today at the Formats meeting with @dgault and @melissalinkert, all looks good here. Merging before the last CZI PR gets opened for 5.5.3

PR #2882, opened 2017-06-29
Docs: update CellSens VSI page to include spec list

melissalinkert:
We have two specification documents for Olympus' CellSens .vsi format, both of which were missing from the format page.  See ```data_repo/curated/specs/VSI File Format v1.6.pdf``` and ```data_repo/curated/specs/XV VSI File Format v1.3_confidential.pdf``` (both private).

Discussed with @sbesson in connection with https://list.nih.gov/cgi-bin/wa.exe?A2=IMAGEJ;2b3255.1706


sbesson:
Thanks. Makes sense to add this information and renders well on the staging documentation page. Would it justify increasing the openness rating to Good as well?

melissalinkert:
Commit pushed to update openness and metadata ratings (via ```gen-metadata-ratings```).

PR #2883, opened 2017-06-29
Use new ome-codecs jar

rleigh-codelibre:
Major changes:

- Remove components/forks/jai
- Replace components/formats-(api|bsd)/codecs classes with wrappers.

Testing: Check builds and full repo tests.  Note: jobs won't pass until ome-codecs is available.


sbesson:
As suggested by [these failures](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-breaking-matlab/799/), there is probably still an issue with the dependency resolution with `ome-common:5.3.1` being resolved rather than `ome-common:5.3.2` with the service addition.

From [the Maven documentation](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html), "Maven 2.0 only supports using the "nearest definition" which means that it will use the version of the closest dependency to your project in the tree of dependencies. You can always guarantee a version by declaring it explicitly in your project's POM. Note that if two dependency versions are at the same depth in the dependency tree, until Maven 2.0.8 it was not defined which one would win, but since Maven 2.0.9 it's the order in the declaration that counts: the first declaration wins."

So, we are down to a few options:
- reorganizing the pom to make sure the order is  so that `ome-common:5.3.2` is always the first declaration in the Maven sense
- explicitly declare `ome-common` as the first dependency of each `pom.xml`
- re-release all intermediate components to consume the latest patch release of `ome-common`

sbesson:
A few first round of comments:

- with the latest commit, some of failures above were fixed :+1:
- the POM reorganization sound sensible and should allow us to bump low-level components according to the Maven dependency mechanism
- as mentioned verbally, the first commit has so many changes it puts a lot of burden on the reviewer. Could we break this into separate commits (at least separate the one removing the fork classes) and compromise on having each commit individually buildable?
- the migration of the `TargaRLECodec` to `formats-gpl` means a relicensing as per the definition of these components, can this be reverted
- `loci.codecs.CodecOptions` is functionally equivalent `ome.codecs.CodecOptions`. I am unclear on the impact of modifying the signature of the `IFormatWriter` interface to use the latter. Could that impact a consumer class?
- the contract of `ome.codecs.Codec` interface is different from  `loci.formats.codec.Codec` as per the `CodecException`. All inheritance changes to the `loci.formats.codec` classes are thus breaking and should be reverted.
- the internal changes to the concrete readers to use the  low-level codec classes look sensible and functional

rleigh-codelibre:
I'll look at all these today.

Regarding licensing.  It's perfectly fine to include a BSD-licenced source file in a GPL work.  The file remains BSD licensed; the work as a whole remains GPL.  The Linux kernel, for example, includes many BSD source files, but is GPL.

rleigh-codelibre:
I've split up and rebased the branch to remove the conflicts.  The removals are now in separate commits.

Regarding the `CodecOptions` interface stuff.  It's allowed to introduce a new parent class without a compatibility break.  See [here](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html); last bullet point in the introductory section.  Changing this is where the wrappers become unmanageable, just like for Metadata/Model classes, where we can't use the ome-model API directly because we can't return ome-model types, only the wrapped types, which makes using it impossible.  In this case it's about the types we pass in; the `CodecOptions` classes in ome-codecs can't be passed via the writer API if we revert the change.

Similarly with `loci.formats.codec.Codec` inheriting `loci.formats.codec.Codec`.  Yes, it's breaking; decoupling necessitated introducing `CodecException`.  We can remove the inheritance, but I want to be very clear about the impact of that: this means we retain every last bit of codecs functionality inside bio-formats, making the decoupling largely pointless.  I can make every reader and writer use the wrapped codecs, which means nothing in bioformats will actually use the new jar other than the wrappers.  Nothing in the codebase right now uses the wrappers; they are purely for compatibility for anyone who happened to be using them.  Every reader and writer using the new jar wraps and rethrows `FormatException`, so all existing users should be safe.  Anyone using the codec wrappers directly would need to catch `CodecException`, but this will be a tiny minority of our userbase.

This is where our strategy is, I think, misguided.  The clean and pragmatic solution here is to use the new library properly, make a breaking change and document this in the release notes.  Done in a single step.  Pretty much all of the changes in ths PR are not going to be noticed by end users unless you're using the codec classes directly, i.e. not using the writer API, and in this case you need to use a different package name or catch `CodecException`.  We have the same problems added as for the model API wrappers; we can't continue to accumulate all these wrappers indefinitely.  Internally, we can't even make use of our own jars without tortuous and convoluted workarounds.  We need to do a breaking release to clean it all up, because it is getting increasingly unmanageable.  While laudable, this "compatibility" has little practical benefit to our end users; most will never know or care about the changes at this level, and those that do can make the needed changes in just a few minutes.  I'd like to strongly suggest we consider what the real cost/benefit of the current strategy is, and to maybe rethink it.

sbesson:
While planning this decoupling/integration, we agreed that we would not break the Bio-Formats API even though this choice comes at some manageable cost. Changing this policy has implications for the project and deadlines and so cannot be changed without a re-assessment. Can we still meet the 5.6.0-m1 deadline without a breaking API change?

rleigh-codelibre:
Possibly with more work and more extensive wrapping.  Note that it's simply not possible to wrap enumerations.  You can't inherit them, wrap them, substitute one for another.  In the current PR, they are copied, which makes two identical, but separate and *incompatible* enumeration types; but only the ome-codecs ones are actually used.  The other is there to retain API compatibility only; it's not for actual use.  We can possibly write wrapper methods which forward the bioformats enum values to the ome-codecs enum values, but it will mean more work to achieve that.

The reason I am bringing this up is because this all has a cost.  Quite a large cost.  And I don't feel that the small benefits gained by doing this make it a worthwhile exercise.  The benefits here are marginal at best.

This level of wrapping is where the scifio work ran into trouble.  Because what is being asked requires extraordinary measures to achieve, if in fact it is achievable.  Java is quite limited when it comes to this type of wrapping.  Retaining compatibility at the bioformats level as you are asking requires a hard compatibility break in the future when the wrappers are removed, because using the wrappers means it's not possible to use the ome-codecs API *at all*.  In this PR, I've made it possible to use ome-codecs with the bioformats writer API directly, providing an upgrade path.  The caveat is needing to catch `CodecException`, which I have mitigated to the maximum extent possible.  That minor change is the tradeoff this PR makes to mitigate the severity of future breakage, as well as providing a sane upgrade path and deprecation schedule.  None of these are possible with a zero-breakage approach; we'll end up in the situation we have with the metadatastore classes, where it's not possible to deprecate them or switch over to the new implementations.

rleigh-codelibre:
I have put some followup work here: https://github.com/rleigh-codelibre/bioformats/tree/ome-codecs-wip

This fully wraps and delegates to all `Codec` implementations using a `WrappedCodec` class, which all delegated codecs derive from.

All enumerations are used directly from ome-codecs.  They aren't exposed in public APIs.  The existing enumerations are left in place for compatibility.

The main sticking point is `CodecOptions`.   It's a concrete class in both ome-codecs and in formats-api.  Here the latter extends the former.  Which works fine, until you get to derived instances for Huffman, JPEG2000 etc.  We don't have multiple inheritance, but the underlying codecs need an instance of the ome-codecs classes, so we derive from these directly and relax the setters to use ome.codecs.CodecOptions in the writer classes.  This works.  But... the `TiffCompression` and related classes have getters which return `loci.formats.codec.CodecOptions`.  Supporting this requires a deep copy of all the options.  This branch does a copy of `CodecOptions` when returning; this clips off the derived CodecOptions' class content.  Does a two-way deep copy of the concrete type instances in both components would be more work, but is getting a bit overly-complex and fragile, particularly when there are zero callers of this method.

I've worked on getting formats-bsd building this afternoon; some failing unit tests indicate some additional work is needed.  If anyone would like to suggest a better approach, please feel free.

joshmoore:
It might help to look at a few example usages, but I suspect that parallel hierarchies will need to be maintained and that the `loci` classes can have copy-constructors and conversion methods for getting an `ome.codec` version if useful. (Note: I also assume leaving `@Deprecated` on all of these classes is appropriate for the minor release)

mtbc:
> requires a hard compatibility break in the future

Absolutely: the question is perhaps more of *when*. :+1: for the idea of `@Deprecated` annotations; furthermore, if we can promptly decide when, then `@deprecated` Javadoc may be able to warn now when the break will happen.

joshmoore:
First read looks good. Ignoring the JAI removal:

```
git diff --stat=200,200 origin/develop roger/ome-codecs | grep -v jai
 .classpath-template                                                                                    |    2 -
 ant/toplevel.xml                                                                                       |   45 +-
 components/autogen/pom.xml                                                                             |   15 +
 components/bio-formats-plugins/pom.xml                                                                 |   19 +-
 components/bio-formats-tools/pom.xml                                                                   |   11 +-
 components/bundles/bioformats_package/build.properties                                                 |    1 -
 components/formats-api/pom.xml                                                                         |   12 +
 components/formats-bsd/build.xml                                                                       |   52 +-
 components/formats-bsd/pom.xml                                                                         |   21 +-
 components/formats-bsd/src/loci/formats/codec/Base64Codec.java                                         |  164 +---
 components/formats-bsd/src/loci/formats/codec/BitBuffer.java                                           |  230 +-----
 components/formats-bsd/src/loci/formats/codec/BitWriter.java                                           |  155 +---
 components/formats-bsd/src/loci/formats/codec/ByteVector.java                                          |   57 +-
 components/formats-bsd/src/loci/formats/codec/HuffmanCodec.java                                        |  155 +---
 components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java                                       |  297 +------
 components/formats-bsd/src/loci/formats/codec/JPEGCodec.java                                           |  120 +--
 components/formats-bsd/src/loci/formats/codec/JPEGTileDecoder.java                                     |  348 +--------
 components/formats-bsd/src/loci/formats/codec/LZOCodec.java                                            |  144 +---
 components/formats-bsd/src/loci/formats/codec/LZWCodec.java                                            |  335 +-------
 components/formats-bsd/src/loci/formats/codec/LosslessJPEGCodec.java                                   |  304 +-------
 components/formats-bsd/src/loci/formats/codec/LuraWaveCodec.java                                       |  111 +--
 components/formats-bsd/src/loci/formats/codec/MJPBCodec.java                                           |  307 +-------
 components/formats-bsd/src/loci/formats/codec/MSRLECodec.java                                          |   94 +--
 components/formats-bsd/src/loci/formats/codec/MSVideoCodec.java                                        |  235 +-----
 components/formats-bsd/src/loci/formats/codec/PackbitsCodec.java                                       |   52 +-
 components/formats-bsd/src/loci/formats/codec/PassthroughCodec.java                                    |   31 +-
 components/formats-bsd/src/loci/formats/codec/QTRLECodec.java                                          |  159 +---
 components/formats-bsd/src/loci/formats/codec/RPZACodec.java                                           |  185 +----
 components/formats-bsd/src/loci/formats/codec/WrappedCodec.java                                        |  203 +++++
 components/formats-bsd/src/loci/formats/codec/ZlibCodec.java                                           |   41 +-
 components/formats-bsd/src/loci/formats/codec/package.html                                             |   35 -
 components/formats-bsd/test/loci/formats/utests/JAIIIOServiceTest.java                                 |  298 --------
 components/formats-bsd/test/loci/formats/utests/LuraWaveServiceTest.java                               |   67 --
 components/formats-bsd/test/loci/formats/utests/MissingJAIIIOServiceTest.java                          |   62 --
 components/formats-bsd/test/loci/formats/utests/MissingLuraWaveServiceTest.java                        |   62 --
 components/formats-bsd/test/loci/formats/utests/testng-no-lurawave.xml                                 |   40 -
 components/formats-bsd/test/loci/formats/utests/testng.xml                                             |    2 -
 components/formats-bsd/test/loci/formats/utests/tiff/TiffCompressionDecompressTest.java                |    8 +-
 components/formats-gpl/pom.xml                                                                         |   25 +-
 components/test-suite/pom.xml                                                                          |   19 +-
 docs/sphinx/developers/components.rst                                                                  |    4 +-
 docs/sphinx/users/fiji/index.rst                                                                       |    4 +-
 excludebugs.xml                                                                                        |    4 +
 pom.xml                                                                                                |    5 +-
 tools/test-build                                                                                       |    1 -
 399 files changed, 375 insertions(+), 142666 deletions(-)
```

says to me that no changes were needed to consume the wrapped codecs (I didn't see any missing constructors or such).

Can you explain the intent though of the check in `components/formats-bsd/test/loci/formats/utests/tiff/TiffCompressionDecompressTest.java`:

```
     catch (FormatException e) {
-      if (!(e.getCause() instanceof ZipException)) {
+      if (!(e.getCause() instanceof ome.codecs.CodecException &&
+            e.getCause().getCause() instanceof ZipException) &&
+          !(e.getCause().getCause() instanceof ZipException)) {
         fail("Unexpected exception: " + e);
```

rleigh-codelibre:
The check is is making sure that the thrown exception is a FormatException and that the wrapped exception is a ZipException, and no other type.

With the delegation, the FormatException is now a wrapped CodecException which is itself wrapping ZipException.  I left the original check in place in case there were codecs in use which were unwrapped which could also throw ZipException.

I could probably remove the latter check; I left it in for safety.

joshmoore:
> With the delegation, the FormatException is now a wrapped CodecException

Or alternatively unwrap the cause and pass it to `FormatException`?

Relatedly I noticed:

```
$ git diff origin/develop roger/ome-codecs | grep -E ^[+] | grep public | grep extends | grep -v WrappedCodec
+public class BitBuffer extends ome.codecs.BitBuffer {
+public class BitWriter extends ome.codecs.BitWriter{
+public class ByteVector extends ome.codecs.ByteVector {
+public class JPEGTileDecoder extends ome.codecs.JPEGTileDecoder {
```

Could these also use delegation rather than subclassing?

rleigh-codelibre:
The cause could be unwrapped, certainly.  But what we have here is not in violation of the API contract, and would removing the intermediate exception hide problems if this ever needs to be debugged?

For the other classes, these are subclassed because they are simple classes where subclassing is straightforward and does not present any complications or compatibility concerns.  Is there any reason to delegate here, when we don't have to?

sbesson:
As an additional input to this discussion, we have at least one specific use case (see https://github.com/openmicroscopy/bioformats/pull/2889#discussion_r125682297) where extending some of the low-level classes, minimally `ome.codecs.JPEGTileDecoder` to contract the Closeable interface would be valuable. Does this affect the delegation vs subclassing decisions in any way?

joshmoore:
> what we have here is not in violation of the API contract, and would removing the intermediate exception hide problems if this ever needs to be debugged?

Not really a violation, no, unless we had said in the Javadoc "check the cause for X", but on the one hand, it's hard to know whether or not a consumer is relying on something like that and on the other, what benefit is there to the consumer to find a `CodecException` when s/he essentially doesn't even know they exist? I'd think if we have the values from:

 * `getMessage()`,
 * `getCause()`, and
 * `getStackTrace()`

set on the `FormatException` then the user need never see the CodecException (until it's formally added to the API)

rleigh-codelibre:
@sbesson I don't think adding an interface would be a problem so long as there's an implementation provided; it's then at most a compatible addition.  Something for this PR, or to do later?

joshmoore:
> so long as there's an implementation provided

`Closeable` is a marker interface for the JVM itself permitting `try-with-resources` statement.

> Is there any reason to delegate here, when we don't have to?

The biggest win of delegating that I see is the ability to keep changing the underlying behavior, i.e. making breaking changes to ome-codec while it's pre-1.0, without impacting end-users.

> there will be some performance overhead with all the CodecOptions wrapping and conversion

I'd think _very_ minimal.

rleigh-codelibre:
I have pushed changes to make all the classes wrap and delegate to the ome.codecs implementations.

The thrown exception cause hasn't been changed yet.  I'm unsure about this one.  We might not even have a cause for the wrapped CodecException since it's used to substitute for FormatException and is thrown directly without a cause in several of the codecs which could be used by the TIFF code.  In the testcase, we're digging into the cause as part of a unit test to explicitly verify the behaviour; it's unlikely that an end user would do so, and if they did they should be prepared for it to vary for different codecs in any case and they should be respecting the exception specification which does not guarantee that certain causes will or will not be present.

joshmoore:
> We might not even have a cause for the wrapped CodecException since it's used to substitute for FormatException and is thrown directly without a cause in several of the codecs which could be used by the TIFF code

You mean it might not be necessary to have the test for `formatExc.getCause() is CodecException`? If so, I think you're right. But I can also see the case for just erasing `CodecExceptions` existence since it is not (yet) of benefit to the consumer.

rleigh-codelibre:
In this specific test case, we are checking that a specific exception is not thrown and we can't change that without the test being broken.

In the general case for all of the FormatException instances being thrown by the wrappers, which have CodecException as a cause, we can't replace that cause because there is no underlying cause to replace it with in many cases.  While we could copy out the stack trace and message, this seems to be obfuscating and confusing things for little tangible benefit.  It would also behave inconsistently depending upon whether CodecException had an underlying cause or not.

joshmoore:
> In this specific test case, we are checking that a specific exception is not thrown and we can't change that without the test being broken.

Right. And additionally, I don't see it as the place of this PR to change the test. The behavior should be adapted until the test is green.

> It would also behave inconsistently depending upon whether CodecException had an underlying cause or not.

I disagree. You optionally copy the cause. A single copy function in `WrappedCodec` can implement the required check:

 * if cause & message use `new FormatException(msg, cause)`
 * if just cause, use `new FormatException(Cause)`
 * if just message, use `new FormatException(msg)`

> While we could copy out the stack trace and message, this seems to be obfuscating and confusing things for little tangible benefit.

I disagree. We're adapting to go along with the delegation pattern. That's the goal. The consumer shouldn't need to care about `CodecException` yet.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-DEV-merge-push#1062](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/1062/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/1062/consoleText) for more details.
Possible conflicts:
  - PR #2877 rleigh-codelibre 'build: Update maven plugin versions'
    - components/forks/jai/pom.xml

sbesson:
Closed #2877 and restarted the daily build to include the latest state of this work in the repository job.

docs/sphinx/developers/components.rst
```diff
@@ -90,9 +90,7 @@ not be any code here, just build system files.
 
 .. _forks-jai:
 
-:source:`forks/jai (JAI Image I/O Tools) <components/forks/jai>`:
-
-`Ant: jar-jai`
+`OME JAI (deprecated) <https://github.com/ome/ome-jai>`_:
```
sbesson: The OME codecs components should also be added to this page alongside the other upstream OME dependencies

docs/sphinx/users/fiji/index.rst
```diff
@@ -38,7 +38,7 @@ an earlier version using this guide if you need to.
 4) Remove bio-formats_plugins.jar from the 'plugins' sub-folder
 5) Remove all of the .jars from the 'jars/bio-formats' sub-folder:
 
-   - jai_imageio.jar
+   - ome-jai.jar
```
sbesson: `ome-codecs.jar` should probably be listed here

With decoupling, performing this operation manually will become increasingly burdensome. Instead it should be possible to use https://github.com/openmicroscopy/bio-formats-fiji directly. Updating these instructions to consume this repo is outside the scope of this PR.

docs/sphinx/users/fiji/index.rst
```diff
@@ -57,7 +57,7 @@ an earlier version using this guide if you need to.
    http://downloads.openmicroscopy.org/bio-formats/) and place them in the
    'jars/bio-formats' sub-folder:
 
-   - jai_imageio.jar
+   - ome-jai.jar
```
sbesson: same as above.

excludebugs.xml
```diff
@@ -22,6 +22,10 @@
     <Class name="~loci\.formats\.ome\..*"/>
     <Bug pattern="NM_SAME_SIMPLE_NAME_AS_SUPERCLASS"/>
   </Match>
+  <Match>
+    <Class name="~loci\.formats\.codec\..*"/>
+    <Bug pattern="NM_SAME_SIMPLE_NAME_AS_SUPERCLASS"/>
```
rleigh-codelibre: I don't think so.

components/formats-bsd/src/loci/formats/codec/WrappedCodec.java
```diff
@@ -0,0 +1,221 @@
...
+
+import loci.common.RandomAccessInputStream;
+import loci.formats.FormatException;
+
+/**
+ * Implements encoding (compress) and decoding (decompress) methods
+ * for Wrapped.  This code was adapted from the Jakarta Commons Codec source,
+ * http://jakarta.apache.org/commons
+ *
+ * @author Melissa Linkert melissa at glencoesoftware.com
```
rleigh-codelibre: Done.

docs/sphinx/developers/components.rst
rleigh-codelibre: Done.

docs/sphinx/developers/components.rst
```diff
@@ -195,6 +194,14 @@ In addition to I/O, there are several classes to assist in working with XML
 (:common_javadoc:`DebugTools <loci/common/DebugTools.html>`), and byte arithmetic
 (:common_javadoc:`DataTools <loci/common/DataTools.html>`).
 
+.. _ome-codecs:
+
+`OME Codecs <https://github.com/ome/ome-codecs>`_:
+
+Provides classes for encoding and decoding compressed data for a
+variety of compression formats.  There are no dependencies on other
+components.
```
melissalinkert: Perhaps rephrase the last sentence, as ome-codecs does depend upon ome-common?

PR #2885, opened 2017-06-30
Style docs updates

dgault:
Porting the style updates to sphinx docs from https://github.com/openmicroscopy/ome-internal/pull/353/files


sbesson:
Thanks @dgault. With this PR, the doc artifacts generated withby the [Ant job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/679) have the new OME styling. Unfortunately, the changes in this PR seem to be incompatible with the usage of our usage of `-Dhtml_theme` to apply the Plone-like Sphinx theme - see [this failure](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-docs/756/).

The underlying problem is that some the `html_theme_options` defined here are specific to the chosen HTML theme and cannot be easily overriden (see https://github.com/sphinx-doc/sphinx/issues/3884). 

Marking this PR as breaking for now. If no workaround is possible, this implies that applying the new OME style will unconditionally force us to drop the alternate Sphinx theme and migrate to the new documentation layout to be discussed/agreed on Tuesday. If this is the case, I would propose to take the time to do this in 5.6.0-m1 rather than a patch release.

sbesson:
@dgault can you also cherry-pick the commit from https://github.com/openmicroscopy/ome-internal/pull/355 ?

sbesson:
Except for the minor scaling adjustment mentioned in https://github.com/openmicroscopy/bioformats/pull/2885#issuecomment-313497664, the output of the documentation build looks in agreement with the new theme. Once the last commit is pushed this should be good to merge.

Issue #2132, opened 2015-12-03
Loading large jpeg files fails


felixveysseyre:
Hello,

(This issue is a copy of a topic opened on OME forum : https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=7957)

I am currently trying to convert a very large JPEG image (65500px x 53090px) with bfconvert.

When I run the executable with this image, I got the following stack trace :
```
../bftools/bfconvert -bigtiff -series 0 "/.../.../x-3.jpeg" "/.../.../TIFF/%z.%t.tiff" 2>&1
/.../.../x-3.jpeg
JPEGReader initializing /.../.../x-3.jpeg
Populating metadata
Exception in thread "main" java.lang.IllegalArgumentException: width*height > Integer.MAX_VALUE!
   at javax.imageio.ImageReader.getDestination(ImageReader.java:2823)
   at com.sun.imageio.plugins.jpeg.JPEGImageReader.readInternal(JPEGImageReader.java:1046)
   at com.sun.imageio.plugins.jpeg.JPEGImageReader.read(JPEGImageReader.java:1014)
   at javax.imageio.ImageIO.read(ImageIO.java:1422)
   at javax.imageio.ImageIO.read(ImageIO.java:1326)
   at loci.formats.in.ImageIOReader.initImage(ImageIOReader.java:150)
   at loci.formats.in.ImageIOReader.initFile(ImageIOReader.java:122)
   at loci.formats.in.JPEGReader$DefaultJPEGReader.initFile(JPEGReader.java:190)
   at loci.formats.FormatReader.setId(FormatReader.java:1426)
   at loci.formats.DelegateReader.setId(DelegateReader.java:290)
   at loci.formats.in.JPEGReader.setId(JPEGReader.java:89)
   at loci.formats.ImageReader.setId(ImageReader.java:835)
   at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:367)
   at loci.formats.tools.ImageConverter.main(ImageConverter.java:874)
```

I am currently using this version of bfconvert:
```
Version: 5.1.4
VCS revision: 05840624ab3d1d1dca14d1ccfebabcb61c42ec27
Build date: 4 September 2015
```

Do you advise any other solution to convert this image to a TIFF file ?

Regards,
Félix

melissalinkert:
[...]

PR #2889, opened 2017-07-05
JPEG: support images with more than Integer.MAX_VALUE pixels

melissalinkert:
See https://trello.com/c/M8PidMnv/16-use-jpegturboservice-to-read-large-jpeg-images

There are two new files in ```data_repo/curated/jpeg/big-images```, both of which were created by copying the raw bytes from the first strip in the first IFD of an existing .ndpi file.

With and without this change, attempting to open ```smithd_20111115_2.ndpi.jpg``` (from ```data_repo/curated/hamamatsu/david-smith/smithd_20111115_2.ndpi```) should throw a ```javax.imageio.IIOException``` as noted on the card.  This is because the image dimensions exceed 65500 x 65500 (the maximum allowed by JPEG), and so are set to 0.

Without this change, ```showinf -nopix``` on ```PB_9217_Retina_04 - 2010-07-16 19.04.48.ndpi.jpg``` (from ```data_repo/curated/hamamatsu/harvey/PB_9217_Retina_04 - 2010-07-16 19.04.48.ndpi```) should throw an ```IllegalArgumentException``` as noted in https://github.com/openmicroscopy/bioformats/issues/2132.  With this change, ```showinf -crop 0,0,512,512``` or similar should log an ```IllegalArgumentException``` but then proceed to open the requested tile.  The detected size X and Y should confirm that there more than ```Integer.MAX_VALUE``` pixels.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
@@ -106,20 +106,20 @@ public class TileJPEGReader extends FormatReader {
 
     in = new RandomAccessInputStream(id);
     JPEGTileDecoder decoder = new JPEGTileDecoder();
-    decoder.initialize(in, 0, 1, 0);
+    int[] dimensions = decoder.preprocess(in);
+    decoder.close();
+    decoder = null;
 
     CoreMetadata m = core.get(0);
 
     m.interleaved = true;
     m.littleEndian = false;
 
-    m.sizeX = decoder.getWidth();
-    m.sizeY = decoder.getHeight();
+    m.sizeX = dimensions[0];
+    m.sizeY = dimensions[1];
     m.sizeZ = 1;
     m.sizeT = 1;
 
-    decoder.close();
-    decoder = null;
     reopenFile();
 
     m.sizeC = 3;
--- a/components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
@@ -47,7 +47,7 @@ import loci.formats.services.JPEGTurboService;
 import loci.formats.services.JPEGTurboServiceImpl;
 
 /**
- * Reader for decoding JPEG images using java.awt.Toolkit.
+ * Reader for decoding JPEG images using java.awt.Toolkit and TurboJPEG.
  * This reader is useful for reading very large JPEG images, as it supports
  * tile-based access.
  *
@@ -58,8 +58,7 @@ public class TileJPEGReader extends FormatReader {
 
   // -- Fields --
 
-  private JPEGTileDecoder decoder;
-  private JPEGTurboService service;
+  private transient JPEGTurboService service;
 
   // -- Constructor --
 
@@ -82,18 +81,6 @@ public class TileJPEGReader extends FormatReader {
     if (service != null) {
       service.getTile(buf, x, y, w, h);
     }
-    else {
-      int c = getRGBChannelCount();
-
-      for (int ty=y; ty<y+h; ty++) {
-        byte[] scanline = decoder.getScanline(ty);
-        if (scanline == null) {
-          decoder.initialize(currentId, 0);
-          scanline = decoder.getScanline(ty);
-        }
-        System.arraycopy(scanline, c * x, buf, (ty - y) * c * w, c * w);
-      }
-    }
 
     return buf;
   }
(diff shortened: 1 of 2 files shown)


components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
```diff
@@ -122,34 +105,24 @@ public void initFile(String id) throws FormatException, IOException {
     super.initFile(id);
 
     in = new RandomAccessInputStream(id);
-    decoder = new JPEGTileDecoder();
-    decoder.initialize(in, 0, 1, 0);
+    JPEGTileDecoder decoder = new JPEGTileDecoder();
+    int[] dimensions = decoder.preprocess(in);
+    decoder.close();
```
joshmoore: One vote for JTD being made Closeable and this happening within a `with` block.

melissalinkert: Makes sense, I just don't want to change ```JPEGTileDecoder``` here since it is removed in https://github.com/openmicroscopy/bioformats/pull/2883.  Happy to open a follow-up card to be addressed once ome-codecs settles down.

sbesson: The integration with the work in #2883 makes complete sense to drive the new low-level codec component. Would the following step here be to review all decoders which should be made `Closeable`, release version `0.2.0` and propagate upstream?

joshmoore: Sounds good to me.

components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
```diff
@@ -58,8 +58,7 @@
 
   // -- Fields --
 
-  private JPEGTileDecoder decoder;
-  private JPEGTurboService service;
+  private transient JPEGTurboService service;
```
mtbc: Does this invalidate old memoization caches?

components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
```diff
@@ -161,4 +134,20 @@ public void initFile(String id) throws FormatException, IOException {
     MetadataTools.populatePixels(store, this);
   }
 
+  @Override
+  public void reopenFile() throws IOException {
+    if (in != null) {
+      in.close();
+    }
+    in = new RandomAccessInputStream(currentId);
+    in.seek(0);
```
mtbc: It's outwith this PR, but if opening a file for input might start with the file position anywhere other than the beginning, it might be good to add that warning to the constructor's Javadoc.

components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
dgault: Yeah unfortunately it does, carried out a test locally and it results in:
`2017-07-11 11:24:40,034 [main] WARN  loci.formats.Memoizer - deleting invalid memo file: /Sample Images/jpg/.8kx8k.jpg.bfmemo
com.esotericsoftware.kryo.KryoException: Unable to find class: pg`

melissalinkert: Does this block merging, or is it acceptable given that this is meant for a minor release?

mtbc: If judged acceptable then we might want to think if there were any other memo breakages that we'd held that could also now go in.

joshmoore: As a data point in deciding, I was just forced to break bfmemo compatibility in a _patch_ release. Using three boxes, the estimated time to recalculate all the memo files is roughly 100 hours.

rleigh-codelibre:
I wasn't sure what additional review was needed.  I tested it and it works according to the description, and the changes look fine.

The only minor comments I have are that if `reopenFile` fails to initialise the service, `service` will remain non-`null`, and subsequent uses might have interesting failure modes.  Maybe it should be set to `null` on failure?

There's also a check for `service != null` in `openBytes`.  This will silently return the buffer if it is ever null, rather than failing with an NPE.  Why are we checking for non-null here rather than failing appropriately should this condition ever occur (which it shouldn't)?  Would it be appropriate to simply remove the check?

melissalinkert:
53b8ecd changes the ```service``` checks per the above comment.  Listing for final review tomorrow.

rleigh-codelibre:
Thanks, the latest changes look fine to me.

PR #2897, opened 2017-07-14
CZI: add basic support for PALM data

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12469, see also https://trello.com/c/y6KpD3sU/18-czi-tickets

I would expect tests to remain green and memo files to be unaffected.  The following are new or previously untested files:

```
data_repo/curated/zeiss-czi/qa-11007/SCC47 EGF647 LAMP1488 - LAMP1 only 3D with 405 2_PALM - KAthrin.czi
data_repo/curated/zeiss-czi/qa-11007/SCC47 EGF647 LAMP1488 - EGF only 3D with 405_PALM.czi
data_repo/curated/zeiss-czi/qa-17616/abeta647-dilution1milliontimesnanomolar.czi
data_repo/curated/zeiss-czi/qa-9410/cav_PALM_drift_correct_Convert to Image.czi
data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 black LSM 880 -May 2015/Image 5_PALM_verrechnet.czi
data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 black LSM 880 -May 2015/PALM_OnlineVerrechnet.czi
data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 black LSM 880 -May 2015/Palm_Source.czi
```
and existing files with updated configuration:

```
data_repo/curated/zeiss-czi/zeiss/dvds/lsm-disk1/CZI-Example Data LSM/PALM/Palm_mitDrift.czi
```

For each, compare in ZEN vs. ```showinf``` or ImageJ with this PR.  With the exception of ```Palm_Source.czi```, I'd expect ZEN to show either one or two channels with the same XY size.  The first channel is stored at a higher resolution, but downsampled for display; if you mouse over the image, you can see that pixel values are not displayed.  The ```Histo``` tab can be used to confirm pixel min/max values, otherwise it's a matter of checking whether the image looks roughly the same in ```showinf```/ImageJ.  The second channel, if present, is stored and displayed at the same (lower) resolution; the pixel values will be shown if you mouse over the image in ZEN.

```showinf```/ImageJ should show the two channels as being split into separate series due to the difference in XY size.  The images and pixel min/max values should look the same as in ZEN, and basic metadata (physical pixel size, channel names, etc.) should match.

```Palm_Source.czi``` is a special case in that Bio-Formats detects a single channel with 100 timepoints, where ZEN shows 2 channels as described above.  All 100 timepoints should show data, and that image count is consistent with the size of the file.  The thumbnail in the right-hand panel of ZEN should be consistent with the images returned by Bio-Formats.  My understanding is that ZEN Lite does not have true PALM support (only ZEN black does), so in my opinion ZEN is not definitive in this case and what Bio-Formats does is reasonable.  If there is a usable installation of ZEN black, that be more definitive, else happy to hear other thoughts on how this file should be handled.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -174,6 +174,8 @@ public class ZeissCZIReader extends FormatReader {
   private int[] tileHeight;
   private int scaleFactor;
 
+  private transient boolean isPALM = false;
+
   // -- Constructor --
 
   /** Constructs a new Zeiss .czi reader. */
@@ -535,6 +537,7 @@ public class ZeissCZIReader extends FormatReader {
       tileWidth = null;
       tileHeight = null;
       scaleFactor = 0;
+      isPALM = false;
     }
   }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java [hunk 1]
```diff
@@ -174,6 +174,8 @@
   private int[] tileHeight;
   private int scaleFactor;
 
+  private transient boolean isPALM = false;
```
dgault: The transient usage here is fine as the variable is only ever used during `initFile` which means that it wont affect memo file usage.

Unless there are plans to make further use of this variable later I would be tempted to make it local to the `initFile` method would remove any confusion regarding whether or not it affects serialisation. Either approach works fine though.

melissalinkert: Good point - cleaned up in 8252660

dgault:
The actual code changes here look good. From the details provided it certainly seems that the new changes are all behaving as expected. The builds and tests are also green with the associated config PR.

From carrying out testing with the sample files listed I had the following results:

The below files show as single channel in Zen and open as a single series with Bio-Formats, the images are empty with 0 pixel count, though this is consistent in both Bio-Formats and Zen:
```
data_repo/curated/zeiss-czi/qa-11007/SCC47 EGF647 LAMP1488 - LAMP1 only 3D with 405 2_PALM - KAthrin.czi
data_repo/curated/zeiss-czi/qa-11007/SCC47 EGF647 LAMP1488 - EGF only 3D with 405_PALM.czi
```

The below images open as described with 2 channels in Zen and as 2 series in Bio-Formats. All the metadata and pixel values look to be aligned between the 2:
```
data_repo/curated/zeiss-czi/qa-17616/abeta647-dilution1milliontimesnanomolar.czi
data_repo/curated/zeiss-czi/qa-9410/cav_PALM_drift_correct_Convert to Image.czi
data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 black LSM 880 -May 2015/PALM_OnlineVerrechnet.czi
```

The following image only displayed as a single channel in Zen and as single series in Bio-Formats. The pixel values and metadata matched up correctly:
`data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 black LSM 880 -May 2015/PALM_OnlineVerrechnet.czi`

As mentioned in the PR description `data_repo/curated/zeiss-czi/zeiss/ZEN 2.1 black LSM 880 -May 2015/Palm_Source.czi` opened with 2 channels in Zen but a series of 100 images in Bio-Formats. I was able to open the file with Zen Black, here it also displays as a single channel with 100 time points and looks and displays as Bio-Formats does.

In summary the display and metadata of the files in Bio-Formats was comparable as when viewing the files using Zen (Blue or Black).

PR #2908, opened 2017-07-25
Fix broken QTFF link

dgault:
Attempting to fix link in https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-latest-docs/395/warnings3Result/

PR #2910, opened 2017-07-26
Add missing SIF format to docs

dgault:
While testing a PR in Bio-Formats 5.5.3 (https://github.com/openmicroscopy/bioformats/pull/2876) I noticed that we were missing SIF on the formats docs pages.

I have tried to base the ratings on the comments from https://trello.com/c/WmLyfC5N/7-rfe-update-format-metadata-support-ratings

The reader itself is very basic and doesn't support many metadata fields.

### DIFF
--- a/components/autogen/src/format-pages.txt
+++ b/components/autogen/src/format-pages.txt
@@ -160,6 +160,7 @@ owner = `Andor Technology <http://www.andor.com/>`_
 developer = Andor Bioimaging Department
 bsd = no
 weHave = * a small number of Andor SIF datasets
+weWant = * an Andor SIF specification document
 pixelsRating = Good
 metadataRating = Fair
 opennessRating = Fair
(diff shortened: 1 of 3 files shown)


rleigh-codelibre:
Looks absolutely fine and good to merge to me.

sbesson:
Looks good. As the individual format pages are auto-generated using the `autogen` module, the metadata for the new format page should be registered incomponents/autogen/src/format-pages.txt` otherwise this addition will be reverted by the auto-generated documentation job.  The changes from https://github.com/openmicroscopy/bioformats/pull/2908 could also be ported to `format-pages.txt`at the same time

sbesson:
:+1: for the last sets of commits. Looking at [the generated commit](https://github.com/snoopycrimecop/bioformats/tree/develop/merge/autogen), the main missing item is the line about the specification document. Otherwise I assume the generated `Developer/Owner` section matches the other Andor format pages and is the correct one?

dgault:
Added the missing weWant section, the Dev/Owner section matches the other Andor format pages

sbesson:
:+1: the diff between `format-pages.txt` and the generated format pages will be cleared when during the documentation auto-generation step for Bio-Formats 5.6.0.

PR #2914, opened 2017-08-01
Handle null returned from getParentFile

dgault:
This PR is in conjunction with https://github.com/ome/ome-common-java/pull/12

To test:


mtbc:
Ready for review?

sbesson:
With ome/ome-common-java#12 requiring some work to minimize the impact on client code, would  using the return value of `Location.getParent()` to perform the non-null check be a possibility to address this issue in the interim?

sbesson:
Closing for now. Requires a bit more investigation to address this issue while minimizing the impact on the code/readers.

PR #2918, opened 2017-08-07
ant: remove bundles build files during 'clean'

melissalinkert:
This prevents the Maven *-dependencies.xml files from being retained in ```components/bundles/*/build```, as that can cause the bundle jars to include a confusing mix of dependencies.

To reproduce the original problem:

 ```
$ git checkout v5.4.0 # no Olympus .oir support
$ find . -name *-dependencies.xml -exec rm {} \; # clean cached dependencies so that build is really 5.4.0
$ ant clean jars tools
$ showinf -nopix olympus-oir/olympus/40X_4CH_XY.oir # should throw UnknownFormatException as expected
$ git checkout develop # has Olympus .oir support since v5.5.0
$ ant clean jars tools # don't explicitly clean cached dependencies
$ showinf -nopix olympus-oir/olympus/40X_4CH_XY.oir # incorrectly throws UnknownFormatException
```

Repeating the above test, replacing ```develop``` with this PR's branch or the merge branch should result in the .oir file being initialized in the last step, indicating that the 5.4.0 dependency files were automatically removed.

### DIFF
--- a/ant/toplevel.xml
+++ b/ant/toplevel.xml
@@ -55,7 +55,8 @@ Type "ant -p" for a list of targets.
       clean-bio-formats-tools,
       clean-tests,
       clean-turbojpeg,
-      clean-docs-sphinx"
+      clean-docs-sphinx,
+      clean-bundles"
     description="remove build files for all components">
     <delete dir="${artifact.dir}"/>
     <delete dir="build"/>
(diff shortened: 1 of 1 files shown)


sbesson:
Thank you. The caching issue had been reported by a few of us but we never got the time to  pinpoint the root of the problem. 

Testing following the instructions of the PR and I can confirm that the daily build including this PR now properly cleans the cached dependencies for the bundles. This means `git checkout && ant clean tools` should be sufficient to have a clean build without requiring the extra `git clean -dfx` step.

PR #2922, opened 2017-08-08
Metamorph: Correct DeltaT across channels

dgault:
Issue was originally raised in forum thread https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8325
A sample set of files was also provided in QA-17824

In short each plane should have a unique delta T value, however each channel has the same timepoints as the that of the first channel. This is due to the fact that for the given sample file set, the channels are split across files but when determining which file to read in order to populate metadata the channel index is not taken into account.

Opening this PR to test against other filesets in the data repo

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -946,7 +946,7 @@ public class MetamorphReader extends BaseTiffReader {
         Double expTime = exposureTime;
         Double xmlZPosition = null;
 
-        int fileIndex = getIndex(0, 0, coords[2]) / getSizeZ();
+        int fileIndex = getIndex(0, coords[1], coords[2]) / getSizeZ();
         if (fileIndex >= 0) {
           String file = stks == null ? currentId : stks[i][fileIndex];
           if (file != null) {


rleigh-codelibre:
This certainly looks fine, testing shows unique and reasonable deltaT values.  Looks good to merge when the full repo test is green.

sbesson:
As shown by https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/463/ this will probably require a configuration change for one of our filesets. @dgault I propose to exclude this PR for now to get the repository jobs back to green and come back to this in a patch release assume it does not break serialization.

dgault:
Associated QA PR opened: https://github.com/openmicroscopy/data_repo_config/pull/231
I also tested the serialisation by saving a memo file without the  PR and verifying that it loaded from memo with the PR applied.

Removing from breaking.

sbesson:
Tested with the QA provided with the initial forum post. Everything worked as expected and the planetimestamps in the second channel are properly offset from the plane timestamps  in the first channel.

The configuration file changes for our internal QA file also shows the same offset behavior. In addition, the values of the Zposition values are also slightly offset between the various channels.

Discussed the latter point with @dgault, while the Plane dimension indices is conserved to express the mapping between planes, this change also provides the correct representation for the plane acquisition metadata (position , timestamp).

Also checked these changes do not break serialization from 5.6.0. Once validated by the full repository tests, this PR should be good to merge.

PR #2925, opened 2017-08-09
TiffSaver: Don't write new IFD for each tile

rleigh-codelibre:
This removes the addition of an unused (and unreferenced) IFD for each call to `TiffSaver.writeImage()` in `TiffSaver.writeIFDStrips()`.  This fixes the problem noted [here](https://trello.com/c/YBsFrE0Q/171-performance-problem-writing-tiles).

Testing: With and without this PR, try bfconverting a large file to a TIFF or OME-TIFF with different tiling parameters.  Without the PR, small tile sizes e.g. 64×64 will cause the file size to bloat dramatically.  This will not occur with the PR merged.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -943,7 +943,9 @@ public class TiffSaver {
     else if (isTiled) {
       fp = sequentialTileFilePointer;
     }
-    writeIFD(ifd, 0);
+    if (fp == out.getFilePointer()) { // Create IFD only if at the end of file
+      writeIFD(ifd, 0);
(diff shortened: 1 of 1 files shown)


dgault:
Ive tested this both using the performance tests that first discovered the problem and also against sample data from the repo. 

Writing the files produces significantly smaller file sizes for tiled data with a large quantity of tiles. With the PR applied the file size remains relatively constant no matter the volume of tiles being written.

Testing against converted sample data shows that the output files are both valid and open with the correct pixel and metadata. 

Testing against converted sample data without tiles also shows that the output images are valid and display correctly with all their metadata.

sbesson:
Thanks. Merging for inclusion in 5.6.0.

PR #2926, opened 2017-08-10
Imaris - incorrect dimensions

dgault:
This issue was first raised on twitter https://twitter.com/alexcarisey/status/894714414909104128
A set of QA files has been submitted QA-17835

The issue is caused when the core metadata for the first series is setup during the parseAttributes function. This will identify any additional resolutions to increase the series count. The metadata for these additional series are then setup after the call to parseAttributes. 

What is occuring is that after identifying an additional resolution in parseAttributes it is then continuing to parse those attributes and overwriting the metadata for the first series.

This change means that once a resolution is identified its metadata does not overwrite that of the first series.

To test:
- Import the ims files in QA-17835 and verify that the dimensions of the image are correct and match those in the associated tiff

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
@@ -486,6 +486,14 @@ public class ImarisHDFReader extends FormatReader {
       if (value == null) continue;
       value = value.trim();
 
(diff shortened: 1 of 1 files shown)


dgault:
Closing in favour of community opened PR https://github.com/openmicroscopy/bioformats/pull/2928

PR #2928, opened 2017-08-10
Fix resolution problems in Imaris HDF Reader 

dinelia:
Currently there are two problems regarding correctly showing all the resolutions.

1) The first (biggest) resolution has wrong dimensions.
  Fixed reading biggest resolution dimensions, to make sure they are from "DataSet/ResolutionLevel_0". 
  
2) When the resolution metadata is not in ascending order of resolutions, the resolution number is wrongly determined. 
     Fixed the resolution count in method parseAttributes(). Now it is correct also for files where groups are not sorted ascendingly by resolution indices.

Another commit in this PR refers to pixel type reading for .ims files (in method initFile). 
The current way reads (part of) pixel data, to see what type it is. While this sounds Ok when pixel data is actually needed (in openBytes()), if one just needs the type, netcdf provides .getDataType() for both Attributes and Variables.

  Modified the way pixel type is determined in method initFile(String), eliminating the need of reading and interpreting pixel data. 
  Added method getVariableDataType(String) in netcdf Service files.

All modifications were tested with Bioformats 5.5.3, current version of Fiji, trying alternatively 3 versions of netcdf jar: netcdf-4.3.22.jar (current version in Fiji), netcdfAll-4.6.10.jar and netcdfAll-5.0.0-SNAPSHOT.jar(built from sources)
Working with the current netcdf version of Fiji has its known limitations when opening .ims files. (It doesn't open one of the files in the below mentioned QA)

Imaris test files with relevant resolution problems were uploaded in QA 17838.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
@@ -303,19 +303,25 @@ public class ImarisHDFReader extends FormatReader {
     ms0.thumbnail = false;
     ms0.dimensionOrder = "XYZCT";
 
-    // determine pixel type - this isn't stored in the metadata, so we need
-    // to check the pixels themselves
-
+    //pixel type is stored in file as "DATATYPE" field of variable "Data"
     int type = -1;
-
-    Object pix = getImageData(0, 0, 0, 1, 1);
-    if (pix instanceof byte[][]) type = FormatTools.UINT8;
-    else if (pix instanceof short[][]) type = FormatTools.UINT16;
-    else if (pix instanceof int[][]) type = FormatTools.UINT32;
-    else if (pix instanceof float[][]) type = FormatTools.FLOAT;
-    else if (pix instanceof double[][]) type = FormatTools.DOUBLE;
+    String path = "/DataSet/ResolutionLevel_0/TimePoint_0/Channel_0/Data";
+    String dataType;
+    
+    try {
+      dataType = netcdf.getVariableDataType(path);
+    }
+    catch (ServiceException e) {
+      throw new FormatException(e);
+    }
+    
+    if (dataType.endsWith("byte")) type = FormatTools.UINT8;
+    else if (dataType.endsWith("short")) type = FormatTools.UINT16;
+    else if (dataType.endsWith("int")) type = FormatTools.UINT32;
+    else if (dataType.equals("float")) type = FormatTools.FLOAT;
+    else if (dataType.equals("double")) type = FormatTools.DOUBLE;
     else {
-      throw new FormatException("Unknown pixel type: " + pix);
+      throw new FormatException("Unknown pixel type: " + dataType);
     }
 
     for (int i=0; i<core.size(); i++) {
--- a/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
@@ -303,25 +303,19 @@ public class ImarisHDFReader extends FormatReader {
     ms0.thumbnail = false;
     ms0.dimensionOrder = "XYZCT";
 
-    //pixel type is stored in file as "DATATYPE" field of variable "Data"
+    // determine pixel type - this isn't stored in the metadata, so we need
+    // to check the pixels themselves
+
     int type = -1;
-    String path = "/DataSet/ResolutionLevel_0/TimePoint_0/Channel_0/Data";
-    String dataType;
-    
-    try {
-      dataType = netcdf.getVariableDataType(path);
-    }
-    catch (ServiceException e) {
-      throw new FormatException(e);
-    }
-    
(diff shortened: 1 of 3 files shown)


components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
```diff
@@ -303,19 +303,25 @@ protected void initFile(String id) throws FormatException, IOException {
...
+    String dataType;
+    
+    try {
+      dataType = netcdf.getVariableDataType(path);
+    }
+    catch (ServiceException e) {
+      throw new FormatException(e);
+    }
+    
+    if (dataType.endsWith("byte")) type = FormatTools.UINT8;
```
rleigh-codelibre: Do we need to cater for signed variants INT8/INT16/INT32 for this file format as well?

rleigh-codelibre:
Overall this looks great.  I was able to open and view the fourth series of `17838/retina_large.ims`, and the metadata looks correct.  One oddity is that the z size is only changed in the forth series and is the same in all previous series'.  Is this intentional and correct?  It just seems a little inconsistent.  Is this series part of the pyramid, or a separate thumbnail?

I wasn't able to open `17838/2016-07-26_17.11.03_Sample Prefix_5P298C2_Protocol.ims` with ImageJ and current bioformats.  I got this error:

```
Exception in thread "main" java.io.IOException: java.lang.IllegalStateException: MessageDataspace: unknown version= 0
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:425)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:392)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:379)
        at ucar.nc2.NetcdfFile.open(NetcdfFile.java:367)
        at loci.formats.services.NetCDFServiceImpl.init(NetCDFServiceImpl.java:326)
        at loci.formats.services.NetCDFServiceImpl.setFile(NetCDFServiceImpl.java:100)
        at loci.formats.in.ImarisHDFReader.initFile(ImarisHDFReader.java:246)
```

This is the error you were seeing with netCDF?  I see this with and without this PR, so it's certainly not a regression.

While this might use a second look over by someone more familar with HDF5 and netCDF, it looks fine to me and good to merge.

dinelia:
Thank you for the feedback.
I uploaded some more files in QA 17839.
The first 2 are more examples of bioformats not opening all the resolutions. (One has the same Z across all resolutions, the other has different Z for the last resolution.) The file 17838/retina_large.ims was obtained by 3D resampling a file with smaller dimensions, but the files in the new upload were generated from the beginning with the current dimensions. It should be OK that some resolutions also modify Z, not only X and Y.
The third file in the upload in QA 17839 has a wrong compression format (not supported anymore by .ims file format and also not recognized by netcdf). It corrrectly doesn't open. It may be useful for testing the triggering of the service exception (and format exception) related to the commit about reading pixel datatype. 
I get the same error as you when trying to open 17838/2016-07-26_17.11.03_Sample Prefix_5P298C2_Protocol.ims with current version of netCDF jar in Fiji.

dgault:
This PR attempts to address two main issues as far as I can see. The first is a bug regarding incorrect dimensions which we have seen reported by users (https://trello.com/c/aYTZfYdJ/172-imaris-v9-incorrect-image-sizes) and the changes made here using `DataSet/ResolutionLevel_0` look to good.

The second issue addressed here is that of the data types. Unfortunately as there is an addition to the NetCDFService API with `getVariableDataType` this would not be eligible for a minor version release. 

There are 2 possible approaches for this:
- We split these 2 issues into separate PR's and schedule them for different releases
- If possible the use of the `getVariableDataType` API addition is removed for the minor release and then refactored at a later date.

It is also worth noting that we are currently reviewing the NetCDF dependency to see if we can upgrade to a more recent version. Although at the moment this is causing issues with OMERO which are under investigation: https://trello.com/c/VrpWk1Fj/119-netcdf-review

dinelia:
The first issue (the bug) is more urgent for users, so either approach is good if the bug fix gets merged sooner. How can I help?

We are very happy about the NetCDF upgrade initiative.

dgault:
@dinelia I think the best option would be to split this into 2 PR's. If we keep this PR open for the first bug and force push the second commit away (https://github.com/openmicroscopy/bioformats/pull/2928/commits/00de4a997115084294de5a4bdc9befdf776653ba) then we will get it included for an upcoming minor release.

The pixel type reading changes can then be opened as a second PR for follow up after.

dinelia:
I reverted the problematic commit.

sbesson:
Thanks @dinelia. With  only the last commit, I can confirm the resolution issues are solved for the uploaded sample files as well as the other [reported by the community](https://trello.com/c/aYTZfYdJ/172-imaris-v9-incorrect-image-sizes). The submitted QA files were also added to our data repository for inclusion in the daily non-regression tests. Merging for inclusion in the upcoming minor release of Bio-Formats.

PR #2935, opened 2017-08-16
Add OMERO pyramid format page

melissalinkert:
See https://trello.com/c/lKU1Tf8M/5-move-omero-pyramid-description-to-a-new-format-page

This adds a new generated format page for the internal OMERO pyramid format.  This page is referenced from the existing general pyramid/resolution API page.

The format page references a separate new page which contains specific implementation details of the OMERO pyramid format, i.e. the beginnings of a spec.  It was easier to write and proofread using two separate pages (so that the spec is not encoded in ```components/autogen/src/format-pages.txt```), but if there is a preference for a single page I can change it once the content is approved.


docs/sphinx/developers/omero-pyramid.rst
```diff
@@ -0,0 +1,22 @@
...
+====================================
+
+For files that contain very large images and are not in a format that supports pyramids, OMERO will generate its own
+image pyramid to improve visualization performance.  Bio-Formats can read these generated pyramids, but cannot
+currently write them outside of OMERO.  For details of how to read image pyramids with Bio-Formats, see :doc:`wsi`
+
+The OMERO pyramid format is a :doc:`TIFF </formats/tiff>` file containing JPEG-2000 compressed image tiles.  All resolutions for a tile
+are encoded in the same JPEG-2000 stream, using the "decompression levels" feature of JPEG-2000.
+As a result, only data types supported by the JPEG-2000 standard (``uint8`` and ``uint16``) are supported.
+Images with pixel type ``uint32``, ``float`` (32 bit floating point), or ``double`` (64 bit floating point) cannot be converted to
```
mtbc: hyphenate 32-bit, 64-bit

components/autogen/src/format-pages.txt
```diff
@@ -1732,6 +1732,22 @@ presenceRating = Fair
...
 reader = SISReader
 
+[OMERO Pyramid]
+developer = `Open Microscopy Environment <http://www.openmicroscopy.org/>`_
+bsd = yes
+versions = 1.0.0
+weHave = * a :doc:`specification document </developers/omero-pyramid>`\n
+* several OMERO Pyramid datasets\n
+* the ability to produce additional datasets
+pixelsRating = Outstanding
```
rleigh-codelibre: Regarding the "outstanding" rating, what's the justification for this?

- I understand it can't represent floating point for pixels?
- It supports almost zero metadata
- It's really only useful in the context of OMERO

So would it make sense to use some more realistic assessment of the pixels, metadata and utility?

sbesson:
Barring the discussion on the ratings (see https://github.com/openmicroscopy/bioformats/pull/2935#discussion_r134690191), separating the format specification from the format information  page makes sense to me.

The developers documentation still sounds a bit out of place as this format spec is effectively consumed by Bio-Formats. Trying to think of alternate locations, the ones I can think of are:

- the `ome-model` Sphinx documentation alongside the other official OME file formats
- as a dedidcated resource document e.g. under https://downloads.openmicroscopy.org/resources/omero/pyramid-format/1.0.0

Also mentioned today during the OMERO 5.4 call. Once this spec is available as publicly, it can be cross-referenced from the relevant places page in the OMERO documentation as a follow-up task.

sbesson:
Discussed at the weekly formats meeting, I convinced myself that the most appropriate location for this spec remains the Model documentation. Instead of creating of a new ephemerous developer page for which we will have to maintain redirects, the proposal is to keep the  description as a section of the existing WSI page and use this as a reference in the format page. Following 5.7.0, we can start working on integrating this specification page formally in ome-model, bump the version in Bio-Formats and start refering to the underlying documentation page.

melissalinkert:
Last commit removes the new page under ```developers``` and moves the pyramid description back to the WSI page as discussed.

docs/sphinx/developers/index.rst
```diff
@@ -40,6 +40,7 @@ Using Bio-Formats as a Java library
     export2
     tiling
     wsi
+    omero-pyramid
```
sbesson: This line needs to be removed as it is failing Travis

PR #2936, opened 2017-08-16
TIFF: add support for FillOrder == 2

melissalinkert:
See https://trac.openmicroscopy.org/ome/ticket/4166

This updates ```TiffParser``` to use ```Integer.reverse(int)``` to reverse the bits in each byte of a tile when the ```FillOrder``` tag is set to 2.  The reversal is performed immediately after the tile buffer is read from the file, before any decompression or unpacking takes place.  An extra check is added to the tile reading logic, so that the special case of returning a single raw tile is never used for ```FillOrder == 2```.

This is a much simpler approach than either of the fixes proposed on the ticket, as it does not require any additional API or changes to other components.  It should be safe for a patch release, but is not urgent and can be deferred if needed.

To test, use the ```data_repo/curated/tiff/john/C100915-MA_AXIS_LEH_IMAGE_KIND_FLAT_ICE_3.tif``` file.  Without this change, ```showinf``` should result in an image with vertical black and white pinstripes.  With this change, ```showinf``` should show a smooth gradient, matching the results of ImageMagick's ```display``` command.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -882,7 +882,7 @@ public class TiffParser {
       photoInterp != PhotoInterp.WHITE_IS_ZERO &&
       photoInterp != PhotoInterp.CMYK && photoInterp != PhotoInterp.Y_CB_CR &&
       compression == TiffCompression.UNCOMPRESSED &&
-      ifd.getIFDIntValue(IFD.FILL_ORDER) == 1 &&
+      ifd.getIFDIntValue(IFD.FILL_ORDER) != 2 &&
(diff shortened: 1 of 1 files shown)


sbesson:
This change seems to have affected the repository tests - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/620. Marking as breaking for now.

melissalinkert:
Test failure should be fixed with a722516

sbesson:
Removing the breaking label for inclusion into the daily builds.

mtbc:
Interestingly ImageJ 1.51i also has the pinstripe problem.

mtbc:
Local testing goes just as expected without and with this PR.

PR #2943, opened 2017-08-28
Plugins: Missing macro doc

dgault:
Issue was discovered while investigating https://www.openmicroscopy.org/community/posting.php?mode=reply&f=13&t=8353

In ImageJ when you open the Bio-Formats Macro doc via:
`Plugins -> Bio-Formats -> Bio-Formats Macro Extensions`

The following function is missing:
`Ext.getSeriesMetadataValue(field, value)`

PR adds the missing function

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/macro/LociFunctions.java
+++ b/components/bio-formats-plugins/src/loci/plugins/macro/LociFunctions.java
@@ -725,6 +725,8 @@ public class LociFunctions extends MacroFunctions {
       IJ.log("-- dimension order.");
       IJ.log("Ext.getMetadataValue(field, value)");
       IJ.log("-- Obtains the specified metadata field's value.");
+      IJ.log("Ext.getSeriesMetadataValue(field, value)");
+      IJ.log("-- Obtains the specified series metadata field's value.");
       IJ.log("Ext.getSeriesName(seriesName)");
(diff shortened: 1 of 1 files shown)


components/bio-formats-plugins/src/loci/plugins/macro/LociFunctions.java
```diff
@@ -725,6 +725,8 @@ public void run(String arg) {
       IJ.log("-- dimension order.");
       IJ.log("Ext.getMetadataValue(field, value)");
       IJ.log("-- Obtains the specified metadata field's value.");
+      IJ.log("Ext.getSeriesMetadataValue(field, value)");
```
mtbc: So is the value argument some kind of reference for where to put the gotten value?

mtbc:
macro-log

components/bio-formats-plugins/src/loci/plugins/macro/LociFunctions.java
dgault: Yeah, essentially `value` will store the returned value after the call, so a very simple example would be:

`run("Bio-Formats Macro Extensions");
id = File.openDialog("Choose a file");
Ext.setId(id);
Ext.setSeries(0);
Ext.getSeriesMetadataValue("myMetadataField", value);
print("Returned value = " + value);`

PR #2944, opened 2017-08-28
Docs: Add new plugins options page

dgault:
This is a follow up to https://github.com/openmicroscopy/bioformats/pull/2891

Adds a new docs page for configuring of the Bio-Formats plugin, namely the slice label pattern.

### DIFF
--- a/docs/sphinx/users/imagej/options.rst
+++ b/docs/sphinx/users/imagej/options.rst
@@ -12,17 +12,15 @@ Configuring the slice label pattern
 The slice label is the text displayed at the top of the image window in ImageJ.
 The values displayed here can be modified by configuring the slice label pattern.
 
-The list of available parameters for configuration is as follows:
-::
+The list of available parameters for configuration is as follows::
+  :%s: series index
+  :%n: series name
+  :%c: channel index
+  :%w: channel name
+  :%z: Z index
+  :%t: T index
+  :%A: acquisition timestamp
 
-    %s - series index
-    %n - series name
-    %c - channel index
-    %w - channel name
-    %z - Z index
-    %t - T index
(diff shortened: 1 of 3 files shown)


sbesson:
The new page also needs to be added to the `toctree` of the top-level `users/index.rst` file.

docs/sphinx/users/imagej/options.rst
```diff
@@ -0,0 +1,28 @@
...
+configuration of features for the plugin such as upgrade checking or modifying 
+the slice label pattern.
+
+Configuring the slice label pattern
+-----------------------------------
+
+The slice label is the text displayed at the top of the image window in ImageJ.
+The values displayed here can be modified by configuring the slice label pattern.
+
+The list of available parameters for configuration is as follows:
```
rleigh-codelibre: You can just use `follows::` here and then you don't need `::` on the next line.

docs/sphinx/users/imagej/options.rst
```diff
@@ -0,0 +1,28 @@
...
+::
+
+    %s - series index
```
rleigh-codelibre: This is fine, but you could use a definition list here, i.e.

```
%n
  series name
%c
  channel index
etc.
```

rleigh-codelibre:
Overall this looks fine; some small Sphinx markup improvements noted, but they aren't required.

docs/sphinx/users/imagej/options.rst
dgault: Ive updated and opted for a field list which seemed to look better from a formatting perspective but I can swap to a definition or option list if that is more the correct implementation.

PR #2950, opened 2017-09-04
clobbering parsed OME XML with poor guesses fixed

bjoernthiel:


dgault:
Hi @bjoernthiel, thank you for taking the time to contribute a PR.

Would you be able to provide a description of the proposed change along with some testing scenarios or instructions. If you have any particular sample files related to the PR you can also upload them to http://qa.openmicroscopy.org.uk/qa/upload/

bjoernthiel:
@melissalinkert can you please comment on this

melissalinkert:
Thank you for updating the description, @bjoernthiel.  Unfortunately, this change as it is will cause the file referenced in https://github.com/openmicroscopy/bioformats/pull/2068#issuecomment-154984783 to fail tests.  This is due to the fact that the OME-XML stored in the file does not contain a required ```LittleEndian``` attribute, which was corrected by the call to ```MetadataTools.populatePixels(ome, this, false, false)```.

As @dgault requested earlier, could you please upload a file that requires this fix?

bjoernthiel:
I see. It's too bad that one cant see the tests fail. 
So @melissalinkert please tell me if something else is wrong with the OME-XML stored in that file.

bjoernthiel:
To fix missing `BigEndian = false` and to really overcome `MetadataTools.populatePixels` I had to patch the code a little bit more.

dgault:
@bjoernthiel the logs containing the test fails you are looking for can be found at https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/495/

From the log the 4 failures are as below:
```
   [testng] [2017-09-15 06:30:57,090] [pool-1-thread-4] 	DimensionOrder: FAILED (Series 0 (got XYZTC, expected XYZCT))
   [testng] [2017-09-15 06:30:57,265] [pool-1-thread-4] 	PhysicalSizeX: FAILED (Series 0 (expected ome.units.quantity.Length: value[3.3333333869999996E-6], unit[mm] stored as java.lang.Double, actual ome.units.quantity.Length: value[3.333333387E-9], unit[m] stored as java.lang.Double))
   [testng] [2017-09-15 06:30:57,265] [pool-1-thread-4] 	PhysicalSizeY: FAILED (Series 0 (expected ome.units.quantity.Length: value[3.3333333869999996E-6], unit[mm] stored as java.lang.Double, actual ome.units.quantity.Length: value[3.333333387E-9], unit[m] stored as java.lang.Double))
[testng] [2017-09-15 06:30:57,659] [pool-1-thread-4] 	testValidXML: FAILED ()
[testng] java.lang.NullPointerException: null
   [testng] 	at loci.tests.testng.FormatReaderTest.testSaneOMEXML(FormatReaderTest.java:614) ~[classes/:na]
```

bjoernthiel:
@melissalinkert @dgault 
The first one is not an error. It is the intended change: The real value `XYZTC` is no longer clobbered with the poor guess `XYZCT`. The second and third one are simple rounding errors. The forth one may come from the XML parser itself and needs further investigation: 

> Invalid content was found starting with element 'Plane'

The empty element `<ome:MetadataOnly/>` should have been read just before.

sbesson:
Hi @bjoernthiel, thanks for the update and clarifying the expectations in terms of metadata changes. We'll capture these and update our internal configuration once the remaining tests are fixed and this PR is ready for formal review. With regard to the remaining XML issue, `CellVoyagerReader` would probably be another reader worth looking at since it also constructs an OME-XML metadata from a parsed OME-XML string. Let us know if you need more input for the investigation.

As an aside, although we have not performed any code review yet, the current diff contains a lot of whitespace changes and variable renaming. At minimum, could the commit(s) related to the bug fix described in the PR be isolated from the rest? More generally, it is yet unclear that these stylistic changes are necessary and they make the these changes harder to review and track in the future.

bjoernthiel:
@sbesson @dgault
Thanks for looking on my PR and pointing me to the `CellVoyagerReader`. Concerning the fourth failure mentioned above I don't know what to do next. Can I reproduce that somehow?

bjoernthiel:
@sbesson @melissalinkert
Concerning the 

> lot of whitespace changes and variable renaming

I have to admit that there is indeed one simple white space change (line 133) that is superfluous and one variable renaming (line 225 `series` -> `image`) that is not strictly necessary here but only prevents us from bad surprise in the future as it recovers the base class member `series`. If you wish I can revert these. 
To help you with the code review please consider the following notes:

- A class member `file_version` has been added to overcome the local variable `fileVersion` and to make the `initStack` call shorter. 
- A class member `ome_meta_data` has been added. It is filled with the parsed (and patched) OME XML and put into the base class member `metadataStore` afterwards. That replaces the local variable `ome`. 
- Corresponding to `ome_meta_data` the local variable `obf` has been renamed to `meta_data` to make the code more readable to me (line 71) and others.

sbesson:
@bjoernthiel: to reproduce the failure, you might need to use to the file you uploaded in https://github.com/openmicroscopy/bioformats/pull/2068#issuecomment-154984783 as this is the one which failed the automated tests. Do you still access to it?

There might be two ways to reproduce:
- either by using the Bio-Formats [tests](https://docs.openmicroscopy.org/bio-formats/5.7.1/developers/commit-testing.html) i.e. by generating a configuration file for the file and running the `test-automated` target - this is what gets executed by the CI job
- or from the command-line tools, `showinf -nopix -omexml Measurement1.msr` should throw errors while validating the OME-XML

Thanks for the detailed notes in https://github.com/openmicroscopy/bioformats/pull/2950#issuecomment-331091538, could these be turned into the description of individual commits?

bjoernthiel:
@sbesson @melissalinkert 
Thanks for showing me the way to reproduce the failure via `showinf -nopix -omexml Measurement1.msr`. 
It is `service.convertMetadata` that processes everything but the `<MetadataOnly/>` element. In case that is intended (and not a bug) I have a workaround for my code. 
Please advise.

sbesson:
@bjoernthiel sorry for the delay. It is true the the[MetadataConverter](https://github.com/ome/ome-model/blob/v5.5.3/ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java) does not include the conversion of Pixels MetadataOnly of TiffData elements.

This absence of this conversion is something we will have to discuss internally as modifying this behavior would certainly have implications for existing client code. With the existing API, either the `MetadataTools.populatePixels()` or the `OMEXMLService.addMetadataOnly()` might be the appropriate API for setting these elements in the reader.

bjoernthiel:
@sbesson I've set up a workaround now (sorry for the delay).

dgault:
Hi @bjoernthiel, the following build was run against the Impsector repo with the latest changes applied: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/499/

The latest commit has fixed the issue with the MetadataOnly element, however the OME-XML test still fails because the Channel elements have no SamplesPerPixel:

`   [testng] [2017-10-17 11:43:16,938] [pool-1-thread-3] 
   [testng] java.lang.NullPointerException: null
   [testng] 	at loci.tests.testng.FormatReaderTest.testSaneOMEXML(FormatReaderTest.java:614) ~[classes/:na]`

The line in the FormatReaderTest which fails is:
`        int samplesPerPixel = retrieve.getChannelSamplesPerPixel(i, 0).getValue();`

Technically the OME-XML created here is valid as SamplesPerPixel is optional in the schema, hence it passes for testValidXML. But within Bio-Formats the expectation of readers is a bit stricter and as such fields such as SamplesPerPixel are populated for all readers.

Correcting this should be as straight forward as setting the ChannelSamplesPerPixel on MetadataStore:
`setChannelSamplesPerPixel`

bjoernthiel:
@dgault @sbesson @melissalinkert 
Is that _bit stricter expectation_ of Bio-Formats documented somewhere so I can patch all missing items at once?

sbesson:
@bjoernthiel: the complete list is not available as such in the technical documentation although the  call to `MetadaTools.populateMetadata()` used in most readers and recommended in the [new reader page](https://docs.openmicroscopy.org/bio-formats/5.7.1/developers/reader-guide.html) populates the mandatory metadata  elements to make the readers compliant with the tests.

Briefly,
- the [MetadataTools.verifyMinimumPopulated](https://github.com/openmicroscopy/bioformats/blob/v5.7.1/components/formats-api/src/loci/formats/MetadataTools.java#L336) checks the set elements allowing to comply with the [OME model minimum specification](http://docs.openmicroscopy.org/ome-model/5.5.7/specifications/minimum.html)
- the [FormatReadetTest.testSaneOMEXML](https://github.com/openmicroscopy/bioformats/blob/v5.7.1/components/test-suite/src/loci/tests/testng/FormatReaderTest.java#L594) performs additional checks either for the presence of some elements like Channel.SamplesPerPixel (which is essential for any SizeC-related calculation) or the validity of the optional elements like the Image.AcquisitionDate.

bjoernthiel:
@dgault @sbesson
Thanks for pointing me to the right direction (and sorry for the delay).

bjoernthiel:
@dgault @sbesson @melissalinkert 
Can someone please give this another try and run the tests?

dgault:
Hi @bjoernthiel, just as an update to the progress of this PR. I have been taking a more in depth look review of the changes here in hope of including the PR in the next patch release Bio-Formats 5.7.3. 

I have opened a configuration PR https://github.com/openmicroscopy/data_repo_config/pull/247 which will update the DimensionOrder of the failed sample file. From debugging into the sample file I can confirm that the OME-XML being read does contain the updated dimension order and this change is correct.

The 2 other failed values PhysicalSizeX and PhysicalSizeY are also being read correctly from the file as `3.333333387E-9`, however when running the test these values are not propagating through correctly. I will continue to investigate as to why these tests are failing and will update you shortly with the results.

bjoernthiel:
Thanks @dgault

dgault:
Following on from the above comment on the Physical Size attributes. The value stored in the config is `3.333333387E-9` which matches the value read from the file. However the test suite parses this value from the config using `FormatTools.getPhysicalSizeX`, which will round the value using `Constants.EPSILON`. The `convertMetadata` method now used when parsing the file simply copies the value and does not use the FormatTools method. Prior to this PR the file was parsed using the original `FormatTools.getPhysicalSizeX` which would have stored the `3.3333333869999996E-6` value.

In short the issue is in the difference between the `convertMetadata` and `FormatTools.getPhysicalSizeX`. Although not an ideal solution I have tested locally by setting the physical sizes using `FormatTools` after the `convertMetadata` call and this will be enough to ensure that tests pass and the correct values are being used.

I shall push the changes that I have made to this PR and that should allow a full test run to complete and the PR ready to merge.

bjoernthiel:
Thanks @dgault for further investigating this. 
But, I'm afraid, although this one test case would be fixed for dimension X and Y I'm not convinced: 
What happens to the other physical dimensions Z T ... ? 
@sbesson @melissalinkert 
How does the OMEXMLReader handle PhysicalSizeX value[3.333333387E-9], unit[m]? Is there a test?

dgault:
Hi @bjoernthiel, you raise a very good point, this is not an issue limited to this reader and your initial solution was absolutely the correct approach and should have been expected to pass tests without issue. 

This has infact highlighted a wider issue regarding inconsistency between readers and the way the test suite reads configuration files. As you pointed out this is also present in OMEXMLReader for example. We have been discussing internally the best long term solution to this and I have opened a separate PR to test the impact of modifying how our test suite reads configurations files (https://github.com/openmicroscopy/bioformats/pull/3020). 

We will keep you up to date with our progress on this and we must thank you for your patience and help in identifying these problems.

dgault:
With an update to the test suite configuration (https://github.com/openmicroscopy/bioformats/pull/3020) the last commit to force the setting of PhysicalSizes using FormatTools should no longer be needed.

Reverting the commit https://github.com/openmicroscopy/bioformats/pull/2950/commits/9274d06edae8d6413645a39e663e4fd76f98f0af

bjoernthiel:
@dgault @sbesson
Sounds good. Many thanks. When is the 5.8.0 milestone going to reached?

sbesson:
Hi @bjoernthiel and happy new year. We are currently completing the work for the upcoming Bio-Formats 5.7.3 release.
Following this, we will most likely switch development towards the next minor release of Bio-Formats and start integrating the open contributions which are labelled as 5.8.0 candidates. There is no formal date as of today but we would expect a minor release of Bio-Formats will be available by the first quarter of 2018.

bjoernthiel:
@dgault @sbesson @melissalinkert 
Is there anything I can help with to get this into the next patch release?

sbesson:
@bjoernthiel the main issue with regard to its inclusion is that the nature of the proposed changes breaks the serialization. As such, the proposed format fixes are suitable for a minor release but not a patch release (see [this section](https://docs.openmicroscopy.org/bio-formats/5.7.3/about/index.html#versioning-policy) for more details about the Bio-Formats versioning policy). With Bio-Formats 5.7.3 [released yesterday](http://www.openmicroscopy.org/2018/01/11/bio-formats-5-7-3.html), we will start going through our backlog of open PRs for final inclusion and get back to your on Monday.

dgault:
with all of the builds and tests passing this PR looks good for inclusion in the next release 5.8.0. 

While reviewing this PR I had noticed that the OBFReader currently has a mix of different formatting styles, particularly around braces. As it affects the whole reader and not just the code touched in this PR I am happy to merge this PR and I will open a follow up PR to standardise the formatting.

PR #2953, opened 2017-09-04
Reference format options page from user/developer pages

melissalinkert:
The format options page can be difficult to find if you don't already know it exists.  This adds links to the options page from the ```users``` and ```developers``` index pages, so that it's hopefully more accessible to newcomers.

### DIFF
--- a/docs/sphinx/developers/index.rst
+++ b/docs/sphinx/developers/index.rst
@@ -46,6 +46,7 @@ Using Bio-Formats as a Java library
     matlab-dev
     python-dev
     non-java-code
+    /formats/options
 
 Using Bio-Formats as a native C++ library
 =========================================
--- a/docs/sphinx/developers/index.rst
+++ b/docs/sphinx/developers/index.rst
@@ -46,7 +46,9 @@ Using Bio-Formats as a Java library
     matlab-dev
     python-dev
     non-java-code
-    /formats/options
+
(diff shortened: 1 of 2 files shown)


dgault:
Looking through the docs and I agree that the options are not obvious unless you are aware of them already and are prepared to look for them.

Adding them to the index pages fits and it makes sense to have it on both the users and developers pages.

Looks good to merge.

hflynn:
N.B. adding this page to the index by including it in multiple TOCtrees does break the previous/next navigation - once you are on the options page, there's no 'Next topic' in the sidebar. 

That isn't necessarily to say don't do it this way, just a heads up, but if you want to preserve the sidebar navigation, the link needs to be added as a `:doc:` link in text or a see also box.

melissalinkert:
Sorry, didn't realize that would break sidebar navigation.  All are switched to ```seealso``` now, so hopefully that works better.

hflynn:
I realize now it's not strictly broken, there's no Next because you are skipping straight to the last file but that still has the same functional result that you can't use the Next navigation to move through the rest of the user and dev docs... So provided @dgault is happy, I prefer the see also boxes. Also handily highlights the content for even better findability.

sbesson:
Looks good to me. @dgault are you fine with the latest changes?

PR #2954, opened 2017-09-05
Leica LIF: attach companion metadata files if present

melissalinkert:
Resolves http://trac.openmicroscopy.org/ome/ticket/4568

To test, use ```data_repo/curated/leica-lif/alex/2010-11-29-test-for-OME.lif```.  Without this change, ```showinf -nopix``` should show just one used file (the .lif itself).  With this change, ```LASAF_CP.xsl```, ```LeicaLogo.jpg```, and ```Pos002_S001.xml``` should also appear on the used files list.

The extra files are a result of exporting the XML metadata for a single series in the .lif file using Leica's LAS AF software.  The .xml file does not include any additional metadata that isn't already in the .lif, but it is nicely formatted and easier to read (and references the .xsl/.jpg).  None of the files are parsed; they are just attached as companion files.  ```showinf -nopix -trace``` can now be used to show the XML that is in the .lif for comparison.

The automated data tests are also updated to not use the companion files for setId/type checking.

Since there isn't any different metadata stored in the companion files, I don't feel 100% that we need this feature, but it might be useful to someone so opening for review/discussion.  If any major objections, feel free to close and I'll change the ticket to ```wontfix```.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -44,6 +44,7 @@ import javax.xml.parsers.ParserConfigurationException;
 
 import loci.common.DataTools;
 import loci.common.DateTools;
+import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.common.services.DependencyException;
 import loci.common.services.ServiceException;
@@ -97,6 +98,9 @@ public class LIFReader extends FormatReader {
 
   /** The encoding used in this file.*/
   private static final String ENCODING = "ISO-8859-1";
+
+  private static final String LOGO_FILE = "LeicaLogo.jpg";
+  private static final String STYLESHEET_FILE = "LASAF_CIP.xsl";
 
   private static final ImmutableMap<String, Integer> CHANNEL_PRIORITIES =
     createChannelPriorities();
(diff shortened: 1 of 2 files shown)


dgault:
The code changes here look good and builds and tests have been green.

Testing with the recommended sample file shows that the 3 new files are being correctly read and added to the used files list. Opening and reading the file shows that the displayed image data and metadata remains unchanged and is correct :
`Used files:
	/Users/dgault/Documents/Sample Images/leica-lif/alex/2010-11-29-test-for-OME.lif
	/Users/dgault/Documents/Sample Images/leica-lif/alex/Pos002_S001.xml
	/Users/dgault/Documents/Sample Images/leica-lif/alex/LeicaLogo.jpg
	/Users/dgault/Documents/Sample Images/leica-lif/alex/LASAF_CIP.xsl`

sbesson:
Sorry for the late comment:

- from a metadata standpoint, this addition is largely harmless as shown by the tests :+1:
- the only change is the addition of new files to the fileset :+1:

From an OMERO-centric perspective, my feeling is that being inclusive on the fileset is probably better than being restrictive in case some extra information needs to be parsed out of companion files in the future. /cc @jburel @joshmoore 

Two more general questions:

- we have a few Leica LIF samples containing companion XML files named after the master LIF file rather than the internal series name. Should these be considered for inclusion in the fileset as well?
- if we are not completely sure of the benefits feature, would it be worth defensively implementing as an opt-in or opt-out option?

melissalinkert:
> we have a few Leica LIF samples containing companion XML files named after the master LIF file rather than the internal series name. Should these be considered for inclusion in the fileset as well?

If I remember correctly, those files were created by logging the XML contained in each LIF to a file using our own tools - not Leica's software.  I'd lean toward not including them, but it's an easy enough change so happy to do it if anyone feels strongly that they should be included.

> if we are not completely sure of the benefits feature, would it be worth defensively implementing as an opt-in or opt-out option?

Certainly makes sense to be able to enable/disable; I would imagine there are other readers that could make use of a "don't look for companion files" flag too.  Would it be more generally useful to have an API method or a somehow higher-level option rather than something for just this reader?

PR #2956, opened 2017-09-06
Nikon ND2: Z dimension and position fixes

melissalinkert:
Ported from a private PR.  The original test case requiring these changes could not be shared, so a similar test case was manufactured by copying ```data_repo/curated/nd2/christian/CW002_3_rfp.nd2``` to ```data_repo/curated/nd2/christian/missing-dimensions-test.nd2``` and slightly modifying the ```ImageTextInfo``` block to remove Z dimension information.

To test 6174d49, use ```data_repo/curated/nd2/christian/missing-dimensions-test.nd2```.  Without this PR, ```showinf -nopix``` should report 60 series.  With this PR, the same test should show 61 Z sections, matching ```data_repo/curated/nd2/christian/CW002_3_rfp.nd2```.

To test 36c24de, use ```data_repo/curated/nd2/hallie/RNAi Plate 4.nd2``` and ```data_repo/curated/nd2/qa-17785/spc2493.nd2```.  Using ```showinf -nopix -omexml```, compare the ```PositionZ``` values on each ```Plane``` with and without this PR.  Without this PR, the positions should match the ```*PiezoZ Drive``` column in the ```Recorded Data``` tab of the image properties window in NIS Elements.  With this PR, the positions should match the ```Z Coord``` column in the same tab.  See also screenshots on forthcoming configuration PR.

This shouldn't break memo files, but I don't know if it qualifies as breaking due to the configuration change.  It can certainly wait for 5.8.0 if needed; I don't think this will resolve any of the other open issues with .nd2 files, so it's not critical.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -1854,6 +1854,7 @@ public class NativeND2Reader extends FormatReader {
    */
   private void iterateIn(RandomAccessInputStream in, Long stop) {
     Object value; // We don't know if attribute will be int, double, string....
+    Double zHigh = null, zLow = null;
 
     try {
       Integer currentColor = null;
@@ -1971,6 +1972,12 @@ public class NativeND2Reader extends FormatReader {
         else if (name.equals("dZStep")) {
           trueSizeZ = new Double(value.toString());
         }
+        else if (name.equals("dZHigh")) {
+          zHigh = DataTools.parseDouble(value.toString());
+        }
+        else if (name.equals("dZLow")) {
+          zLow = DataTools.parseDouble(value.toString());
+        }
         else if (name.equals("dPosX")) {
           positionCount++;
         }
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -1983,6 +1992,10 @@ else if (name.equals("dPosX")) {
     catch (Exception e) {
       LOGGER.debug("", e);
     }
+
+    if (zHigh != null && zLow != null && trueSizeZ != null) {
```
dgault: For safety I would check that trueSizeZ > 0 so as to avoid a divide by 0 error

dgault:
Other than a single small comment the code looks good and the builds and tests are green

Without the PR:
Opening `data_repo/curated/nd2/christian/missing-dimensions-test.nd2` gives 60 series with a single plane
For `data_repo/curated/nd2/qa-17785/spc2493.nd2` the PositionZ value is 50
For `data_repo/curated/nd2/hallie/RNAi Plate 4.nd2` the PositionZ value is 50.735

With the PR applied:
Opening `data_repo/curated/nd2/christian/missing-dimensions-test.nd2` gives a single series with 61 Z planes
For `data_repo/curated/nd2/qa-17785/spc2493.nd2` the PositionZ value ranges from 2678.025 to 2678.075 which matches the values shown in NIS Elements
For `data_repo/curated/nd2/hallie/RNAi Plate 4.nd2` the PositionZ value ranges from 3400 to 3450 which matches the values shown in NIS Elements

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
melissalinkert: Fixed in 7d8d556

dgault:
With the last commit this PR looks good to merge

PR #2957, opened 2017-09-07
TIFF: support >4GB ImageJ TIFFs with multiple channels

melissalinkert:
See https://trello.com/c/dmISsgn6/118-4gb-imagej-tiffs-not-supported

Without this PR, >4GB ImageJ TIFFs were only supported if multiple Z sections and/or timepoints were present.  This changes the ImageJ comment parser to check the IFD count against the total expected image count, not just the expected Z and T counts.

To test, use ```data_repo/curated/tiff/samples/melissa/imagej-4gb/4gb-imagej.tif``` (created by opening the ```.fake``` in the same directory in ImageJ, then selecting ```File > Save As > Tiff...```).  Without this PR, ```showinf -nopix``` should report a single plane.  With this PR, the same test should report 2050 channels, matching the ```.fake``` file.  All other dimensions and pixel data should also be the same between the two files.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/TiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/TiffReader.java
@@ -255,6 +255,7 @@ public class TiffReader extends BaseTiffReader {
 
     int z = 1, t = 1;
     int c = getSizeC();
+    int images = 1;
 
     CoreMetadata m = core.get(0);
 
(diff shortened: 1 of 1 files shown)


mtbc:
The TIFF has,
```
Indexed = false (false color)
Endianness = motorola (big)
```
which are different from the fake. Is this okay? The code changes look sensible and the channel count is as expected.

melissalinkert:
Yes, that's fine.  ImageJ writes big-endian TIFFs by default (see https://imagej.nih.gov/ij/docs/guide/146-26.html#sub:Save[s]), independent of the input format.  The dimension order doesn't particularly matter in this case as only ```SizeC``` is greater than 1.

sbesson:
We are already handling ImageJ TIFF files as a special case of `TiffReader`, this bug fix makes sense and causes no regression in our automated tests. Merging for upcoming 5.7.1 release.
As discussed earlier,it might be useful to update the TIFF format page and document the support for these types of TIFF files in a separate PR.

PR #2959, opened 2017-09-15
Add reader for IonpathMIBI TIFF

rfinck:
Ionpath's technology, Multiplexed Ion Beam Imaging (MIBI), uses secondary ion mass spectrometry (SIMS) to image tissue sections. We would like to get a reader into bioformats for the use of our beta customers. I've suggested creating a SIMS domain since there are multiple other SIMS Imaging modalities out there.

I have tested this with loci.formats.tools.ImageInfo, and have a sample image I am happy to provide. I saw that one can upload images to http://qa.openmicroscopy.org.uk/qa/upload/ but I was not sure how to tag it as being of this type, which seems necessary in order to use it for testing.

Thanks in advance!

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -331,6 +331,9 @@ public final class FormatTools {
 
   /** Identifies the astronomy domain. */
   public static final String ASTRONOMY_DOMAIN = "Astronomy";
+  
+  /** Identifies the secondary ion mass spectrometry domain. */
+  public static final String SIMS_DOMAIN = "Secondary Ion Mass Spectrometry (SIMS)";
 
   /**
    * Identifies the graphics domain.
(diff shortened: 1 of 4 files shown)


dgault:
Hi @rfinck,

Thank you for taking the time to contribute to the project. Any sample files or formats specs that you have can be uploaded to http://qa.openmicroscopy.org.uk/qa/upload/, you should be able to add a comment to the QA but even if not we should be able to identify them.

Currently Travis is failing the build for Java 7 (https://travis-ci.org/openmicroscopy/bioformats/jobs/275713578), this is due to the below error:
`java.util.HashMap.Entry is not public in java.util.HashMap; cannot be accessed from outside package`

This can be fixed by using `Map.Entry` rather `HashMap.Entry`

mtbc:
Sample file provided in QA 17933.

rfinck:
Thanks @dgault and @mtbc. I fixed the HashMap issue and as noted uploaded a sample file and spec.

dgault:
Thanks @rfinck, the builds are now looking good and having a spec document is a great help. I have carried out some initial testing and the sample image looks as though both the pixel and metadata is being parsed and displayed correctly.

I will include the PR as part of our main build and test suite and carry out a more comprehensive review of it in the coming days.

rleigh-codelibre:
Note that followup tasks for this PR should include a corresponding update to the OME Files imaging domains.

We might also look into any needed model changes to better represent this imaging modality, such as any specific "light source" and "detector" requirements.  Light source is obviously a misnomer for this modality.  The "generic" light source might be useful for attaching instrument-specific setup relating to secondary ion generation e.g. voltages, gas pressure.

sbesson:
@rfinck: sorry for the delay. We are converging towards our next minor release of Bio-Formats including new readers and are  considering this PR for final inclusion. Barring a couple of changes regarding coding styles that we can handle as a follow-up, a couple of more general questions:

- a sample file and a specification has been initially provided to support this new reader - see https://github.com/openmicroscopy/bioformats/pull/2959#issuecomment-329794208. Have you currently frozen your version 1.0 of the IonPath MIBI specification? If so, do you have a document for the final version and/or representative samples?

- on the API domain, as you noticed, we have few coverage of mass spectrometry imaging formats. Although the addition makes sense from a code standpoint, it has other implications on our side to maintain our enumeration in sync across the board including in our reference C++ implementation.
As an alternative, would you be fine with dropping this API addition from this PR and using the `UNKNOWN_DOMAIN` like other readers at the moment? We would then review our list of imaging domain and include new terms like the one you are proposing which could be introduced in a future minor release.

Thanks again for your contribution and let us know what you think of the questions above.

rfinck:
@sbesson 

- We do have a final version of the spec and some sample data we can provide. I will want to double-check the code here and make sure none of the small changes since this draft was submitted warrant any updates. I can do that later this week.

- I'm fine with using `UNKNOWN_DOMAIN` for now if that helps get this into the next release.

sbesson:
Thanks @rfinck, please let us know when you have a final version of the specification and sample data and we will include them in our data repository for testing. In the meantime, integrating this for the imminent 5.8.0 and I will open a follow-up PR reverting the `FormatTools` changes as discussed above and adjusting the code style to comply with [Bio-Formats formatting](http://docs.openmicroscopy.org/bio-formats/5.7.3/developers/code-formatting.html). A final PR will also be required to add a dedicated documentation page - see  https://docs.openmicroscopy.org/bio-formats/5.7.3/developers/format-documentation.html

PR #2965, opened 2017-09-20
GIF: fix behavior of openBytes when reading planes out of order

melissalinkert:
See https://trello.com/c/abmCVITh/149-animated-gif-in-omero and
https://trac.openmicroscopy.org/ome/ticket/12962

To test, use the file from QA 11167.  Without this change, compare ```showinf -range 3 3``` (or any range of planes not including 0) against the corresponding planes shown by just ```showinf```.  The plane shown by ```showinf -range 3 3``` should be obviously different, though ```showinf -range 0 0``` will match the first plane in ```showinf```.

With this PR, the same tests should result in all planes matching.

This should be safe for a patch release, as it should not break memoization.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/GIFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/GIFReader.java
@@ -34,6 +34,8 @@ package loci.formats.in;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.Vector;
 
 import loci.common.RandomAccessInputStream;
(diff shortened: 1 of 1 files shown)


mtbc:
Do we tend to remove these `transient`s when breaking memoization for other reasons?

joshmoore:
For me, the `transient`s are not _just_ a mitigating factor but provide de-/serialization functionality.

sbesson:
Successfully tested by https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository-win/187/ and https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/655/.

PR #2966, opened 2017-09-21
ND2: Incorrect channel size

dgault:
This PR is being used to test a fix for https://trello.com/c/RyXt2BAg/177-nd2-incorrect-channel-count

Essentially a single channel single plane image is being mistakenly read as a 3 channel rgb image.

When there is only a single image offset the availableBytes is based on `in.length` which is not an accurate measure of the true image size. As such for single plane images they can be mistakenly modified from single channel to rgb.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -1192,7 +1192,7 @@ public class NativeND2Reader extends FormatReader {
           availableBytes -= 4096;
         }
       }
-      if (planeSize > 0 &&
+      if (planeSize > 0 && imageOffsets.size() > 1 && 
         availableBytes > DataTools.safeMultiply64(planeSize, 3))
       {
         if (availableBytes < DataTools.safeMultiply64(planeSize, 6)) {
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -1192,7 +1192,7 @@ public class NativeND2Reader extends FormatReader {
           availableBytes -= 4096;
         }
       }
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -1192,7 +1192,7 @@ else if (numSeries == 1 && positionCount > 1) {
           availableBytes -= 4096;
         }
       }
-      if (planeSize > 0 &&
+      if (planeSize > 0 && imageOffsets.size() > 1 &&
```
mtbc: remove trailing space

MaximeMaW:
Hi @dgault,

I confirm that the PR solves the issue mentioned on the [Trello card](https://trello.com/c/RyXt2BAg/177-nd2-incorrect-channel-count). Looking forward to seeing it included in BF!

Thanks a lot for investigating!

sbesson:
Sorry for the delay in reviewing. As shown by https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/497, this change does not affect any of our existing ND2 sample files. Could the initial file be configured for addition into our data repository?

sbesson:
Hi @MaximeMaW, thanks for providing us some early feedback. From our side, all looks good and we will include this bug fix as part of the upcoming patch 5.7.2 release of Bio-Formats.  Thanks also for providing the failing sample file initially. We will add it to our curated data repository against which each Bio-Formats change is tested. Can the `BF007.nd2` file be made public at https://downloads.openmicroscopy.org/images/  under a CC-BY license or should it remain private?

MaximeMaW:
Hi @sbesson. Thank you so much for your feedback, and it will be so useful to the lab to have this bug fixed! Feel free to make the file public under a CC-BY.

PR #2968, opened 2017-09-28
bio-formats-plugins: LibraryChecker: Use safe version check

rleigh-codelibre:
`java.version` version comparison is broken with Java 9.  Use a safe comparison using functionality already built into ImageJ.

Testing: Run ImageJ with a build containing the fixes in this PR.  If you can open the importer dialogue, this is working correctly (with Java 7 or 8).  It should also open with Java 9 (with latest daily imagej build), but there may be other Java 9 issues, so working with Java 9 isn't strictly required.


dgault:
With Java 9 the version should just return 9 rather than 1.9, which does indeed break the existing functionality. The built in isJava17 method checks for versions of 1.7 or higher and does include both 1.9 or 9.

Testing the functionality works as expected and the import dialogue is able to be opened and used correctly. PR is good to merge.

Issue #2946, opened 2017-08-28
Bioformats issues with Z.1 lightsheet files.


shohpfizer:
Hi,

I am currently trying to use bioformats (and python bioformats) to read some CZI files generated by a Z.1 lightsheet. The series is a tiled imaging of a single channel and was done with a dual illumination. The series were separated by illumination, thus, 'file.czi' = left illumination, 'file(1).czi' = right illumination, 'file(2).czi' - left again. (For each stack: XYZTC = 1400,1400,350,1,1)

Now we would like to open these files separately in python/bioformats and extract the image data. For some reason it works well with the files from one side, but gives me an empty (zeros) image for the other illumination. [Zeiss ZEN is displaying both cases correctly.]

Has anyone encountered this issue, and if yes, is there any good fix? I would be happy to just extract the image contents of both illumination images reliably.

Below a screenshow of the file listing, their shapes at a certain z position, and the maximum value of that slice - note that the channel switches from 2 (uneven numbered tile) to 3 (even numbered tile) , and in the 3 channel case, there is no valid image data (max value of zero). Oddly enough, the masterfile (not shown) displays only 2 channels and still has a valid max pixel value.

![image](https://user-images.githubusercontent.com/29429835/29788437-8fffd322-8c00-11e7-8542-e69183dedcae.png)

Has anyone encountered this issue and knows how to fix that?

Thank you and best regards.

dgault:
Hi @shohpfizer,

In order for us to test and debug this issue would it be possible to upload a sample fileset to https://www.openmicroscopy.org/qa2/qa/upload/?

David Gault

shohpfizer:
Hi @dgault , since the filelimit was too small, I uploaded it to Box. Please let me know if you got my email.
Best!

dgault:
Hi @shohpfizer, I was able to test the provided sample files today and from debugging through the code the issue appears related to handling of the Illuminations dimension metadata. 

For the second set of files the metadata is parsed to show `dimension=I, start=1, size=1, startCoordinate=1.0, storedSize=1`. In this case Bio-Formats has used the start index (0-based) and the size to determined that there is a mismatch and has incorrectly updated the number of Illuminations. This in turn has caused the channel count to increase and the indexing when searching for the planes is also incorrect.

I believe I should be able to put a fix in place to better handle this scenario when the files are being opened individually. Once I have a PR opened I will link it to this issue.

shohpfizer:
[...]

PR #2969, opened 2017-09-28
CZI - Incorrect illuminations

dgault:
This PR is a follow up to https://github.com/openmicroscopy/bioformats/issues/2946

The issue occurring is that we have a series that is a tiled imaging of a single channel and was done with a dual illumination. The series were separated by illumination, thus, 'file.czi' = left illumination, 'file(1).czi' = right illumination, 'file(2).czi' - left again. When opening the files separately one half of the files opens correctly while the other opens with an incorrect illumination and channel count.

From the debugging the problem was that the reader was parsing an illumination dimension start value of 1 rather than 0 and incorrectly modifying the number of illuminations and channels.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1635,6 +1635,9 @@ public class ZeissCZIReader extends FormatReader {
             if (dimension.start >= illuminations) {
               illuminations = dimension.start + 1;
             }
+            if (dimension.size < illuminations) {
+              illuminations = dimension.size;
+            }
             break;
           case 'B':
             if (dimension.start >= acquisitions) {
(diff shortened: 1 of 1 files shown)


sbesson:
The last few runs of the [full repository job](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository) failed with a single failure on an existing CZI file:

```
   [testng] [2017-10-02 01:20:27,568] [pool-1-thread-9] Initializing /ome/data_repo/curated/zeiss-czi/qa-10301/141016_OP43adult_Worm3_DAPI_head.czi: 
   [testng] [2017-10-02 01:20:43,342] [pool-1-thread-9] 	SizeC: FAILED (Series 0 (expected 2, actual 1))
```

Is this an unwanted side-effect of this PR or an expected metadata change?

sbesson:
Discussed with @dgault. Labelling as breaking as more investigation is required.

sbesson:
Still additional investigation required to get a fix into place that does not break other datasets. Closing this as we are not planning any work on it at the moment.

PR #2970, opened 2017-10-06
JPEG-XR: allow non-interleaved data to be returned

melissalinkert:
As with several other codecs, ```JPEGXRCodec``` now uses ```options.interleaved``` to determine how pixel data should be returned.  This is useful for any future readers that might use ```JPEGXRCodec``` and expect non-interleaved images.  ```ZeissCZIReader``` expects interleaved images, but is updated to supply more parameters via ```CodecOptions```.  f00b70, 797dee8, and 6dc10da were backported from private PRs; the remaining commits are minor cleanup.

It looks like ```JPEGXRCodec``` was never migrated to https://github.com/ome/ome-codecs, which is why I haven't opened a PR there as well.

There should be no difference in functionality from the CZI perspective, so testing is a matter of code review and verifying that builds remain green.

### DIFF
--- a/components/formats-gpl/src/loci/formats/codec/JPEGXRCodec.java
+++ b/components/formats-gpl/src/loci/formats/codec/JPEGXRCodec.java
@@ -31,6 +31,7 @@ import loci.common.RandomAccessInputStream;
 import loci.common.services.DependencyException;
 import loci.common.services.ServiceFactory;
 import loci.formats.FormatException;
+import loci.formats.ImageTools;
 import loci.formats.MissingLibraryException;
 import loci.formats.UnsupportedCompressionException;
 import loci.formats.services.JPEGXRService;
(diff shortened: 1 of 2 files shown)


components/formats-gpl/src/loci/formats/codec/JPEGXRCodec.java
```diff
@@ -76,7 +81,33 @@
...
+    int c = uncompressed.length / (pixels * bpp);
+
+    if (c == 1) {
+      return uncompressed;
+    }
+
+    byte[] deinterleaved = new byte[uncompressed.length];
+
+    for (int p=0; p<pixels; p++) {
+      for (int channel=0; channel<c; channel++) {
```
dgault: Its a very minor nitpick but as someone reading the calculations below my initial thinking was to assume that `channel` would be the total and `c` would be the index, similar to the pixels and bpp loops. Theres nothing wrong with it as is but from readability it might be nicer to swap the variables eg:

`for (int c=0; c<channels; c++) {`

dgault:
The code and the calculations all look correct and builds are green. 

Completely out of the scope for this PR but as an aside thought, the interleaving calculation seems like it would be a good fit for one of the utility classes someday as it used in various different places throughout the BF stack.

components/formats-gpl/src/loci/formats/codec/JPEGXRCodec.java
melissalinkert: Updated in 5405b5b

melissalinkert:
Follow-up card for interleaving review: https://trello.com/c/q87zxrtw/193-review-locations-where-channel-interleaving-deinterleaving-is-performed

dgault:
With the last commit the PR looks good to merge

PR #2972, opened 2017-10-09
Add java 7 reminders & fix downloads

hflynn:
See https://trello.com/c/WnTLpmC6/4-make-bio-formats-prereqs-obvious

This adds reminders that BF requires Java 7 or above. 

While I was editing, i noticed the hardcoded 5.6 versioned downloads and API links - I've changed these to use the current version number so we don't have to keep updating them and also amended the various download links on pages to point at the latest downloads page on the new website where they are referring to latest or new versions (this does mean we have a slight mish-mash of 'this version' and 'latest version' links so I'm happy to change all of them to match if we can agree a preference between the two).

### DIFF
--- a/docs/sphinx/users/octave/index.rst
+++ b/docs/sphinx/users/octave/index.rst
@@ -33,7 +33,7 @@ Installation
    system-wide or user installation respectively).
 #. Add `bioformats_package.jar` to Octave's *static* javaclasspath (see
    `Octave's documentation <https://www.gnu.org/software/octave/doc/interpreter/Making-Java-Classes-Available.html>`_).
-#. Download the Octave package from the :downloads:`downloads page <>`.
+#. Download :downloads:`Octave package <artifacts/>`.
 #. Start octave and install the package with::
 
       >> pkg install path-to-bioformats-octave-version.tar.gz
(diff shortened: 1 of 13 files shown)


hflynn:
@sbesson tells me my versioned downloads will only work for release builds so I've updated to hardcoded 5.7 links for now and we'll need to fix them properly on a separate PR with an If statement to revert to just using latest if the version is a SNAPSHOT once Seb's figured out how to get it working.

docs/sphinx/users/octave/index.rst
```diff
@@ -33,7 +33,7 @@ Installation
    system-wide or user installation respectively).
 #. Add `bioformats_package.jar` to Octave's *static* javaclasspath (see
    `Octave's documentation <https://www.gnu.org/software/octave/doc/interpreter/Making-Java-Classes-Available.html>`_).
-#. Download the Octave package from the :downloads:`downloads page <>`.
+#. Download :downloads:`Octave package <artifacts/>`.
```
mtbc: I think we can retain "the".

dgault:
All the links are working as expected and the pages are displaying correctly. Looks good to merge.

PR #2974, opened 2017-10-10
Deltavision: parse more key/value pairs from the log file

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13237

To test, use the dataset in ```data_repo/curated/deltavision/michael/ticket-13237/```.  This was created by copying ```data_repo/curated/deltavision/michael/mtools/CTR_centrosome_01_R3D.dv``` to ```01_R3D.dv``` and placing ```01_R3D.dv.log``` from the ticket attachment in the same directory.

Comparing ```showinf -nopix 01_R3D.dv``` with and without this change should show that there are more original metadata key/value pairs with this change.  The log file lines noted on the ticket should be represented by original metadata with this change.  There should be no difference in pixel data or image dimensions.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -990,7 +990,7 @@ public class DeltavisionReader extends FormatReader {
 
     for (String line : lines) {
       int colon = line.indexOf(':');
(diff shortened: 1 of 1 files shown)


dgault:
Merging the PR as all additions to the global meta look good and remaining metadata is unchanged.

PR #2977, opened 2017-10-12
docs: sphinx: Correct jai-imageio link

rleigh-codelibre:
Correct link and reference new upstream source repository.

### DIFF
--- a/docs/sphinx/developers/components.rst
+++ b/docs/sphinx/developers/components.rst
@@ -92,13 +92,13 @@ not be any code here, just build system files.
 
 `OME JAI (deprecated) <https://github.com/ome/ome-jai>`_:
 
-This is a fork of JAI ImageIO [no longer maintained; the most active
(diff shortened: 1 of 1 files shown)


ant/logback.xml
```diff
@@ -5,7 +5,7 @@
       <pattern>%d [%t] %-5p %c - %m%n</pattern>
     </encoder>
   </appender>
-  <root level="WARN">
+  <root level="DEBUG">
```
mtbc: Could you explain more about this change?

rleigh-codelibre: An unintentional addition from a dirty tree; I've got rid of it.

hflynn:
We don't in general use square brackets for docs, no, unless it's quoted code featuring them obviously

sbesson:
Only action left here is to replace square brackets by curved ones, is that correct @hflynn ?

hflynn:
Yes please.

hflynn:
👍

dgault:
Link is working and page displays correctly, PR is good to merge

PR #2979, opened 2017-10-16
ZeissCZIReader: Pad scene numbers to order lexicographically

rleigh-codelibre:
See https://trello.com/c/52uaFv3n/194-czi-scene-ordering for original bug report and sample files.

This PR ensures that an appropriate amount of zero padding is used for scene numbers.

Testing:

```
./tools/showinf -noflat -nopix -omexml 18097/2017_10_13__4052-1_offline_fuse_jpegXR_lossy85.czi > new1-noflat.log
./tools/showinf -noflat -nopix -omexml -novalid 18097/2017_10_13__4064_none_JpegXR_lossy85percentdefault.czi > new2-noflat.log
```

Then look for `Scene #` in the output.  Without this PR, scene numbers will be the minimum number of digits.  With the PR, scene numbers will be left padded with zeros for the maximum number of digits used by all scenes.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1233,7 +1233,7 @@ public class ZeissCZIReader extends FormatReader {
           store.setImageName("", i);
         }
         else {
-          store.setImageName("Scene #" + i, i);
+          store.setImageName(String.format("Scene #%0" + ((int) Math.log10(positions - extraImages.size()) + 1) + "d", i), i);
         }
       }
       else if (extraIndex == 0) {
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1233,7 +1233,7 @@ public class ZeissCZIReader extends FormatReader {
           store.setImageName("", i);
         }
         else {
-          store.setImageName(String.format("Scene #%0" + ((int) Math.log10(positions - extraImages.size()) + 1) + "d", i), i);
+          store.setImageName(String.format("scene position #%0" + ((int) Math.log10(positions - extraImages.size()) + 1) + "d", i), i);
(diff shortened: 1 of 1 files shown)


sbesson:
See https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-repository-subset/498 for the execution of this PR against the repository subset. A few failures that might require more code adjustements and/or configuration changes. Marking as breaking for now as this requires investigation.

A few questions with regard to the impact of these changes:

- what is the behavior of the ZEN software with regard to the nomenclature of individual scenes?
- how can we convince ourselves this change in the `ImageName` metadata will not break any assumption in client code?
- related to the two questions above, should we make the scene name padding behavior controlled via option?

rleigh-codelibre:
@sbesson ZEN doesn't have this issue because it's a bioformats-specific suffix added due to our data model lacking support for scenes and positions.  We even collapse multiple scenes together, making the position numbers for subsequent scenes result in no way to know which scene a given position came from.  And it likely interacts badly with the `label image` and `macro image` for all but the final scene since it will treat them as regular positions.

All changes to metadata could potentially break client code.  This one is no different in that respect.  However, the naming is already broken.  `Scene #` is incorrect because these aren't scene numbers, they are *positions within a single scene* when there is a single scene (as in these examples).  `Position #` is almost as bad, because they aren't solely positions if there are multiple scenes.  I've changed the label to `scene position` to match the casing of the other labels and also to add position to the label name.  Not ideal, but better than the existing convention.

We could make the padding behaviour controlled by an option, but I think there are limits to the practicality of using such options.  We already pad all the other indices; this one was an exception.  Guaranteeing the stability of every last bit of metadata is hard, but in this case given the brokenness I'm not convinced it's needed, particularly if we need to make further changes.

sbesson:
Independently on the discussion above, this PR is still affecting the pixel automated tests for a few Zeiss CZI filesets - see https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-repository-subset/293/console.

PR #2983, opened 2017-10-26
Add support for Avizo binary files

### LINKED ISSUE #2982: Avizo binary files imported as text (Fiji/ImageJ)
Avizo (AmiraMesh) binary files are imported in Fiji/ImageJ as text files. It seems to be a simple problem associated with the new header of the software because manually changing the first line of the file header from something like

`# Avizo BINARY-LITTLE-ENDIAN 2.1`

to

`# AmiraMesh BINARY-LITTLE-ENDIAN 2.1`

makes it work.

dgault:
This is a follow up to https://github.com/openmicroscopy/bioformats/issues/2982

Both of the following should be recognised as a Binary file and read:
```
# Avizo BINARY-LITTLE-ENDIAN 2.1
# AmiraMesh BINARY-LITTLE-ENDIAN 2.1
```

To test:
- Ensure that sample files in the repo are green with both variations of the header above

Fixes #2982

### DIFF
--- a/components/formats-bsd/src/loci/formats/tools/AmiraParameters.java
+++ b/components/formats-bsd/src/loci/formats/tools/AmiraParameters.java
@@ -87,7 +87,7 @@ public class AmiraParameters {
     throws FormatException, IOException
   {
     String firstLine = inputStream.readLine();
-    Matcher amiraMeshDef = Pattern.compile("#\\s+AmiraMesh.*?" +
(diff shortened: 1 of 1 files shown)


melissalinkert:
The same fix might be needed in https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/AmiraReader.java#L275, at least for consistency.

PR #2991, opened 2017-11-10
Disable snapshots for the unidata-releases Nexus repository

sbesson:
Following upstream deployment changes on the http://artifacts.unidata.ucar.edu repository, the Ant build of Bio-Formats is seemngly broken during the `mvn-ant-tasks:dependencies` build step of the `formats-gpl` component. The Maven build seems to be unaffected by these changes.

This commit tries to fix these issues by disabling the snapshots retrieval from the Unidata repository. In practice, there should be no harm in doing so as the only dependency we are consuming is the netcdf and we should not use SNAPSHOTS with the unidata-releases repository.

To test this PR:
- [Travis builds](https://travis-ci.org/openmicroscopy/bioformats/builds/300148474) also affected by these changes should be green with these changes
- the Bio-Formats Jenkins jobs on the [OME CI](https://ci.openmicroscopy.org/) should also turn green
- it might be worth testing the impact of this PR locally using a clean Maven repository both via the Ant and the Maven build system i.e.

   ```
   ant clean jars -Dmaven.repo.local=/tmp/repo1/.m2/repository
   mvn clean install -Dmaven.repo.local=/tmp/repo2/.m2/repository
   ```

PR #2992, opened 2017-11-13
Review all repositories to use HTTPS rather than HTTP

sbesson:
Follow-up of https://github.com/openmicroscopy/bioformats/pull/2991

Reproducing the last reported failures with an Ant-based build from a clean Maven repository locally and looking at the details of the faulty artifacts, I realized there was a permanent redirect from http to https when retrieving dependencies from http://artifacts.unidata.ucar.edu/.

This PR review the URLs of all `repositories` and `snapshotRepositories` declared in the various  `pom.xml` (Central, Unidata & OME Artifactory) to use HTTPS consistently. This change is expected to fix the issues raised in https://github.com/openmicroscopy/bioformats/pull/2991#issuecomment-344055075 and https://github.com/openmicroscopy/bioformats/pull/2991#issuecomment-344056024.

Similarly to https://github.com/openmicroscopy/bioformats/pull/2991, both Travis and Jenkins should remain green with these changes. Additionally, both Ant-based and Maven-based builds using a clean Maven repository (either cleaning `~/.m2` or passing `-Dmaven.repo.local`) should pass with this change.
 

NB: as an aside, https://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/BuildDependencies.html is now recommending https://artifacts.unidata.ucar.edu/repository/unidata-all/ rather than unidata-releases and we could consider updating the URL to match this recommendation.

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -356,17 +356,17 @@
     <repository>
       <id>central</id>
       <name>Central Repository</name>
-      <url>http://repo.maven.apache.org/maven2</url>
+      <url>https://repo.maven.apache.org/maven2</url>
     </repository>
     <repository>
       <id>unidata.releases</id>
-      <url>http://artifacts.unidata.ucar.edu/content/repositories/unidata-releases</url>
+      <url>https://artifacts.unidata.ucar.edu/content/repositories/unidata-releases</url>
       <snapshots><enabled>false</enabled></snapshots>
     </repository>
     <repository>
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
This worked fine for me, once I removed the old m2 with invalid content.

dgault:
Testing with a clean build and repo these changes build correctly for both ant and maven builds. the individual formats-gpl component also builds fine on its own. Merging changes.

PR #2996, opened 2017-11-15
Detect and fix Findbugs' SBSC_USE_STRINGBUFFER_CONCATENATION

melissalinkert:
See https://trello.com/c/5E3itceq/179-findbugs-sbscusestringbufferconcatenation

7913cae alone updates the Findbugs configuration to detect ```SBSC_USE_STRINGBUFFER_CONCATENATION``` (http://findbugs.sourceforge.net/bugDescriptions.html#SBSC_USE_STRINGBUFFER_CONCATENATION).  This is done by changing the reporting level to ```medium``` but filtering out all ```medium``` priority bugs apart from ```SBSC_USE_STRINGBUFFER_CONCATENATION```.  Using ```medium``` priority with no filtering results in thousands of additional bugs being detected.

The remaining commits address bugs reported by ```ant -Dfindbugs.home=/path/to/findbugs-3.0.0 findbugs``` with 7913cae included in the build.  Actual functionality should not be affected, so I would expect builds to remain green.

I would also expect this to be safe for a patch release, but it's very low priority and can be closed/excluded if too distracting.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
@@ -207,7 +207,7 @@ public class FilePatternDialog extends ImporterDialog {
       }
     }
     else if (useRanges) {
-      StringBuffer pattern = new StringBuffer();
+      StringBuilder pattern = new StringBuilder();
       for (int i=0; i<counts.length; i++) {
         BigInteger first = new BigInteger(firsts[i]);
         BigInteger fileCount = new BigInteger(counts[i]);
--- a/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
@@ -207,7 +207,7 @@ public class FilePatternDialog extends ImporterDialog {
       }
     }
     else if (useRanges) {
-      String pattern = "";
+      StringBuffer pattern = new StringBuffer();
       for (int i=0; i<counts.length; i++) {
         BigInteger first = new BigInteger(firsts[i]);
         BigInteger fileCount = new BigInteger(counts[i]);
@@ -219,24 +219,24 @@ public class FilePatternDialog extends ImporterDialog {
         // (number of images - 1) * axis increment + axis first image
         fileCount = fileCount.subtract(BigInteger.ONE).multiply(increment).add(first);
 
-        pattern += fp.getPrefix(i);
-        pattern += '<';
+        pattern.append(fp.getPrefix(i));
+        pattern.append('<');
         int firstPadding = paddingZeros[i] - first.toString().length() + 1;
         for (int zero=0; zero<firstPadding; zero++) {
-          pattern += '0';
+          pattern.append('0');
         }
-        pattern += first;
-        pattern += '-';
+        pattern.append(first);
+        pattern.append('-');
         int lastPadding = paddingZeros[i] - fileCount.toString().length() + 1;
         for (int zero=0; zero<lastPadding; zero++) {
-          pattern += '0';
+          pattern.append('0');
         }
-        pattern += fileCount;
-        pattern += ':';
-        pattern += increment;
-        pattern += '>';
+        pattern.append(fileCount);
+        pattern.append(':');
+        pattern.append(increment);
+        pattern.append('>');
       }
-      id = pattern + fp.getSuffix();
+      id = pattern.toString() + fp.getSuffix();
     }
 
     options.setId(id);
--- a/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
@@ -236,7 +236,8 @@ public class FilePatternDialog extends ImporterDialog {
         pattern.append(increment);
         pattern.append('>');
       }
-      id = pattern.toString() + fp.getSuffix();
+      pattern.append(fp.getSuffix());
+      id = pattern.toString();
     }
 
     options.setId(id);
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -531,13 +531,14 @@ public class MicromanagerReader extends FormatReader {
               (key != null && value == null))
             {
               if (value == null && (propType == null || !propType.equals("PropType"))) {
-                value = token;
+                StringBuffer sb = new StringBuffer(token);
 
                 while (q + 1 < tokens.length && tokens[q + 1].trim().length() > 0) {
-                  value += ':';
-                  value += tokens[q + 1];
+                  sb.append(':');
+                  sb.append(tokens[q + 1]);
                   q++;
                 }
+                value = sb.toString();
               }
               if (!value.equals("PropVal")) {
                 parseKeyAndValue(key, value, digits, plane + i, 1);
@@ -776,7 +777,8 @@ public class MicromanagerReader extends FormatReader {
           token.indexOf("\"", dash)));
 
         token = st.nextToken().trim();
-        String key = "", value = "";
+        String key = "";
+        StringBuffer valueBuffer = new StringBuffer();
         boolean valueArray = false;
         int nestedCount = 0;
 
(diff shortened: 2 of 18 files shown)


components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
```diff
@@ -219,24 +219,24 @@ else if (useRanges) {
...
-        pattern += ':';
-        pattern += increment;
-        pattern += '>';
+        pattern.append(fileCount);
+        pattern.append(':');
+        pattern.append(increment);
+        pattern.append('>');
       }
-      id = pattern + fp.getSuffix();
+      id = pattern.toString() + fp.getSuffix();
```
joshmoore:
```
pattern.append(fp.getSuffix())
pattern.toString()
```
?

joshmoore:
> This is done by changing the reporting level to medium but filtering out all medium priority bugs apart from SBSC_USE_STRINGBUFFER_CONCATENATION

Ah, neat idea. Did you see any other clear candidates for next inclusions?

Only real question I have here is if this shouldn't be `StringBuilder` since `StringBuffer` is synchronized.

melissalinkert:
> Ah, neat idea. Did you see any other clear candidates for next inclusions?

Big ones to deal with next would be:

http://findbugs.sourceforge.net/bugDescriptions.html#URF_UNREAD_FIELD
http://findbugs.sourceforge.net/bugDescriptions.html#URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD
http://findbugs.sourceforge.net/bugDescriptions.html#RR_NOT_CHECKED
http://findbugs.sourceforge.net/bugDescriptions.html#SR_NOT_CHECKED
http://findbugs.sourceforge.net/bugDescriptions.html#DLS_DEAD_LOCAL_STORE
http://findbugs.sourceforge.net/bugDescriptions.html#MS_PKGPROTECT

Or more generally, anything with a category other than ```STYLE``` or ```PERFORMANCE```, then anything with category ```STYLE```, then anything (else) with category ```PERFORMANCE``` (e.g. example 4 of http://findbugs.sourceforge.net/manual/filter.html#d0e2255 for category-based exclusion). 

> Only real question I have here is if this shouldn't be ```StringBuilder``` since ```StringBuffer``` is synchronized.

No strong preference here - happy to change to ```StringBuilder``` unless anyone else feels otherwise.

components/bio-formats-plugins/src/loci/plugins/in/FilePatternDialog.java
melissalinkert: Fixed in 1e8b72d

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -2412,9 +2412,10 @@ private void parseText(String textString, int offsetCount, boolean useDimensions
 
       lines = textString.split(" ");
       for (int i=0; i<lines.length; i++) {
-        String key = lines[i++];
-        while (!key.endsWith(":") && key.indexOf('_') < 0 && i < lines.length) {
-          key += " " + lines[i++];
+        StringBuffer sb = new StringBuffer(lines[i++]);
+        while (sb.charAt(sb.length() - 1) != ':' && sb.indexOf("_") < 0 && i < lines.length) {
```
sbesson: This seems to have broken the `setId` call on some nd2 datasets - see https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/703/console

melissalinkert:
Both requested changes have been addressed now.

components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
```diff
@@ -798,18 +800,18 @@ else if (token.trim().startsWith("}")) {
...
+              valueBuffer.append(token.trim().replaceAll("\"", ""));
               token = st.nextToken().trim();
               continue;
             }
           }
           else {
             int colon = token.indexOf(':');
             key = token.substring(1, colon).trim();
-            value = token.substring(colon + 1, token.length() - 1).trim();
+            valueBuffer.delete(0, valueBuffer.length());
```
mtbc: `valueBuffer.setLength(0)` may be clearer.

components/formats-bsd/src/loci/formats/tools/AmiraParameters.java
```diff
@@ -454,49 +454,60 @@ public static String entryToString(Object object, String indent)
...
+      StringBuilder result = new StringBuilder("\"");
       int offset = 0;
       for (;;) {
         int nextOffset = string.indexOf('"', offset + 1);
         if (nextOffset < 0) {
           break;
         }
         if (nextOffset > offset + 1) {
-          result += string.substring(offset, nextOffset);
+          result.append(string, offset, nextOffset);
```
mtbc: I'm doubtful about `nextOffset`: is this code path getting tested?

components/formats-gpl/src/loci/formats/in/BaseZeissReader.java
```diff
@@ -2120,36 +2120,79 @@ public static Charset get(int value) {
...
+      s.append("\n    ");
+      for (double point : points) {
+        s.append(point);
+        s.append(", ");
+      }
+      s.append("\n");
+      return s.toString();
+    }
+
+    private void appendHexValue(StringBuilder s, String key, int i) {
```
mtbc: looks like these can be `static`

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -2412,9 +2412,10 @@ private void parseText(String textString, int offsetCount, boolean useDimensions
 
       lines = textString.split(" ");
       for (int i=0; i<lines.length; i++) {
-        String key = lines[i++];
-        while (!key.endsWith(":") && key.indexOf('_') < 0 && i < lines.length) {
-          key += " " + lines[i++];
+        StringBuilder sb = new StringBuilder(lines[i++]);
+        while ((sb.length() == 0 || sb.charAt(sb.length() - 1) != ':') && sb.indexOf("_") < 0 && i < lines.length) {
```
mtbc: `sb.isEmpty()` may read more clearly

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -2423,18 +2424,21 @@ private void parseText(String textString, int offsetCount, boolean useDimensions
         if (i >= lines.length) {
           break;
         }
+        String key = sb.toString();
 
-        String value = lines[i++];
+        sb.delete(0, sb.length());
```
mtbc: maybe `setLength(0)`

components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
```diff
@@ -3949,13 +3949,15 @@ public String toString() {
         ", pyramidType = " + pyramidType + ", dimensionCount = " + dimensionCount;
       if (dimensionCount > 0) {
         s += ", dimensions = [";
```
mtbc: move a line below so you can use `append`

mtbc:
A few suggestions but no blockers, so take or leave as you please.

components/formats-gpl/src/loci/formats/in/BaseZeissReader.java
melissalinkert: The inner class is non-static, so making these methods static resulted in a compile error.

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
melissalinkert: ```StringBuilder``` doesn't seem to have an ```isEmpty``` method, so leaving this as-is.

melissalinkert:
Thanks, @mtbc.  Suggestions implemented in 2f06100, apart from the couple of notes above.

components/formats-gpl/src/loci/formats/in/BaseZeissReader.java
mtbc: Aha, thank you for trying it; sorry for the noise!

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
mtbc: Indeed not! (Sorry, recovering from a cold, clearly still more addled than I'd realized.)

components/formats-gpl/src/loci/formats/in/BaseZeissReader.java
joshmoore: NB: Longer-term :+1: for making the inner classes static since the reverse could play havoc with at least previous garbage collection implementations.

PR #2997, opened 2017-11-15
MRC: read the ISPG field and store in original metadata

melissalinkert:
See https://trac.openmicroscopy.org/ome/ticket/8688

To test, choose any file from ```data_repo/curated/mrc/``` and verify that ```showinf -nopix``` with this change shows ```ISPG``` and ```Is data cube``` entries in the original metadata table.  The location of the ISPG field was determined from http://www.ccp4.ac.uk/html/maplib.html#description

I would expect builds to remain green and memo files to be unaffected, so this should be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MRCReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MRCReader.java
@@ -274,7 +274,9 @@ public class MRCReader extends FormatReader {
       }
     }
 
-    in.skipBytes(4);
+    int ispg = in.readInt();
+    addGlobalMeta("ISPG", ispg);
+    addGlobalMeta("Is data cube", ispg == 1);
 
     extHeaderSize = in.readInt();
 


dgault:
Builds and tests are green and the changes have no effect on memo files - https://ci.openmicroscopy.org/view/Bio-Formats/job/BIOFORMATS-DEV-merge-repository-subset/513/

The code changes look good, 4 bytes that were previously skipped are now being read. From the spec on http://www.ccp4.ac.uk/html/maplib.html#description this matches up with the field `Space group number`. From specs it also reads that `ispg should be 0 for an image stack, 1 for a volume`, so the new `Is data cube` entry also appears to be correct.

Without this PR there are no entries in the metadata for ISPG. With this PR the global metadata shows entries for `ISPG` and `Is data cube`. The remaining metadata is unaffected.

PR looks good to merge.

PR #2999, opened 2017-11-16
ant: Remove broken utils-formats-api target

rleigh-codelibre:
While looking at other targets, I noticed this one was broken because this component no longer has any utilties at all.  This PR removes the unused and broken target.

Testing: Check builds remain green.  There are no remaining uses of this target, so nothing should break.


dgault:
Without the PR trying to run the target utils-formats-api does indeed fail reporting a missing utils directory. The formats-api component does not have any utils therefore the target is no longer in use and should be removed. 

With the PR applied all builds and tests run successfully and remaining targets are building as expected. PR is good to merge.

PR #3000, opened 2017-11-17
appveyor: Add appveyor configuration

rleigh-codelibre:
This is a Windows equivalent to the existing travis support: a top-level `.appveyor.yml` and a `tools/test-build.bat` script.  A test matrix tests the maven and ant builds with Java 1.7 and 1.8 (1.7/ant excluded because current Ant releases don't work with Java 1.7 anymore).  `.m2` and python virtualenv are cached between runs (though I'm not entirely certain the venv caching is working between runs); the m2 caching is.

While I'm sure we can improve this further, this is an initial working example of how we could improve our CI testing by building routinely on Windows for every open PR.

This can be tested by enabling Appveyor support for the snoopy bioformats git branch and then running the merge-push job.  See https://ci.appveyor.com/project/rleigh-codelibre/bioformats/build/1.0.19 for an example.  Also note that this archives the built zip/tar/jar files for later download, including bftools and documentation.  Appveyor may be enabled for individual developer accounts as well, as for travis.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,46 @@
+environment:
+  AV_PROJECTS: 'c:\projects'
+  AV_BF_M2: 'c:\projects\m2'
+  AV_BF_PYTHON: 'c:\projects\python'
+  AV_BF_SOURCE: 'c:\projects\bioformats'
+
+  matrix:
+    - java: 1.8
+      build: maven
+    - java: 1.7
+      build: maven
+    - java: 1.8
+      build: ant
+
+cache:
+  - '%AV_BF_M2% -> appveyor.yml'
+  - '%AV_BF_PYTHON% -> appveyor.yml'
+
+os: 'Visual Studio 2015'
+clone_folder: '%AV_BF_SOURCE%'
+clone_depth: 5
+platform: x64
+
+init:
+  - git config --global core.autocrlf input
+  - if [%build%] == [ant] appveyor-retry cinst -y ant
+  - refreshenv
+  - 'if [%java%] == [1.7] set "JAVA_HOME=C:\Program Files\Java\jdk1.7.0"'
+  - 'if [%java%] == [1.8] set "JAVA_HOME=C:\Program Files\Java\jdk1.8.0"'
+  - PATH=%JAVA_HOME%\bin;%PATH%
+  - 'if NOT EXIST "c:\projects\%java%\%build%\" mkdir "c:\projects\%java%\%build%\"'
+  - 'if NOT EXIST "%AV_BF_PYTHON%\" C:\Python36-x64\python -m pip install virtualenv'
+  - 'if NOT EXIST "%AV_BF_PYTHON%\" C:\Python36-x64\python -m virtualenv %AV_BF_PYTHON%'
+  - PATH=%AV_BF_PYTHON%;%AV_BF_PYTHON%\scripts;%PATH%
+  - python -m pip install sphinx
+  - 'if [%build%] == [maven] set "MAVEN_OPTS=-Dmaven.repo.local=%AV_BF_M2%"'
+  - 'if [%build%] == [ant] set "ANT_OPTS=-Dmaven.repo.local=%AV_BF_M2%"'
+
+build_script:
+  - tools\test-build %build%
+
+artifacts:
+  - path: 'components\**\*.zip'
+  - path: 'components\**\*.jar'
+  - path: 'components\**\*.tar.*'
+  - path: 'artifacts\*'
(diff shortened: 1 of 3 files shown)


joshmoore:
Thanks, @rleigh-codelibre 

https://ci.appveyor.com/project/snoopycrimecop/bioformats/build/1.0.1 manually launched (in-progress) Currently, only `develop/merge/daily` is activated. A random build version can be used rather than these incremental ones.

joshmoore:
Comparing travis & appveyor:

|             | travis        | appveyor           |
|-------------|---------------|--------------------|
| BUILD       | {maven,ant}   | {mavent,ant}       |
| JDKs        | oracle8,open7 | {1.7/1.8, 1.8}     |
| Maven Steps | mvn           | mvn install        |
| Ant Steps   | ...           | ... + tools + test |

Can you explain the differences in JDKs and the various steps?

From my point-of-view, getting this in at any time should be fine, especially if others want to start testing it locally on their branches. Before we activate this such that it starts blocking merges, we should:

 * unify the above (or at least all agree on the differences)
 * make sure that we feel comfortable that this is going to be green
 * haven chosen the proper point to activate the build (i.e. post 5.7.2)

cc: @sbesson @dgault

rleigh-codelibre:
JVM versions are the same bar ant/1.7 being deliberately omitted; that's different due to Ant 1.10 not working with Java7 as noted in the description.  AppVeyor only provides Oracle JDKs, so these are what are used.

The default maven target is `install` so `mvn` and `mvn install` are equivalent in practice.  We could remove the `install` or keep it explicit, or make travis equally explicit.

Ant `tools` was added so that `bftools` could be provided as an artefact for daily testing.  But it also tests that the target works; I don't know why it's missing from the travis build (do you remember @sbesson ?).  The test target was added initially to make sure it worked, but can be removed to save build time if needed; I'd suggest it should be added for travis as well though since it should really be running the tests every time.  Unless we consider running them via maven to be sufficient; I'm not certain it really is though.

As discussed with @joshmoore there are some problems with storing artefacts; there's a 1GiB limit which we exceeded after three builds.  I've asked [here](https://help.appveyor.com/discussions/problems/9986-artifact-and-cache-space-utilisation) for some information regarding artefact retention policies and space accounting.  If we have to manually delete old builds if will be rather painful to manage, so we would likely need to either turn off artefact saving entirely or manually upload them to our systems.

joshmoore:
> JVM versions are the same bar ant/1.7 being deliberately omitted; that's different due to Ant 1.10 
not working with Java7 as noted in the description.

Ok. So we this is the default version that comes with VS 2015?

> AppVeyor only provides Oracle JDKs, so these are what are used.

Understood. Might be worth a note.

> The default maven target is install so mvn and mvn install are equivalent in practice. We could remove the install or keep it explicit, or make travis equally explicit.

For anyone looking at keeping these two files in sync, having them identical is just that much easier.

> Ant tools was added so that bftools could be provided as an artefact for daily testing. But it also tests that the target works;

Naturally. But again, having the same level for each would be preferable.

> I don't know why it's missing from the travis build (do you remember @sbesson ?). The test target was added initially to make sure it worked, but can be removed to save build time if needed; I'd suggest it should be added for travis as well though since it should really be running the tests every time. Unless we consider running them via maven to be sufficient; I'm not certain it really is though.

I'm not sure of the impact.

> As discussed with @joshmoore there are some problems with storing artefacts; there's a 1GiB limit which we exceeded after three builds. I've asked here for some information regarding artefact retention policies and space accounting. If we have to manually delete old builds if will be rather painful to manage, so we would likely need to either turn off artefact saving entirely or manually upload them to our systems.

Keep us posted.

rleigh-codelibre:
Ant isn't available at all in the provided images.  We install it from chocolatey in the appveyor.yml.  We could download an older 1.9.x version by hand if we really wanted to test building with Java 1.7.  I kept it simple for the initial work.

joshmoore:
Ah, parsing the line noise:

```
+  - if [%build%] == [ant] appveyor-retry cinst -y ant
```

https://chocolatey.org/packages/ant seems to have a 1.9.9 available. I defer to @dgault and @sbesson but from my point-of-view, having the develop branch consistent is quite valuable.

rleigh-codelibre:
@joshmoore Revised to fix the above and align the two scripts.  It's green in https://ci.appveyor.com/project/snoopycrimecop/bioformats/build/1.0.6

rleigh-codelibre:
See the reply from appveyor in https://help.appveyor.com/discussions/problems/9986-artifact-and-cache-space-utilisation -- they will be trialling configurable artifact retention policies and said they would contact us for testing that once it's available.  They also increased our quota from 1GB to 10GB, so we could potentially use that, though it would likely be used up in a few days without a automatic deletion policy.

joshmoore:
Thanks, @rleigh-codelibre ! Let's get this in as is, get all open PRs merged, and then set `Required` for appveyor. If in that time we've got a artifacts solution, we can start rolling it out. _Very exciting._

PR #3009, opened 2017-11-24
ND2 - Support multipoint global metadata

dgault:
This PR is in relation to https://trello.com/c/obHnFk58/12-nd2-multipoint-point-name-missing

Previously the values for the multiple positions were correctly being read but as each set used the same names, when added to the global metadata the key values were duplicate and as a result only the last set of values was stored. This PR resolves this by storing each set by adding an index to the key.

Bio-Formats already identified when multiple positions were present and keeps a count. I am reusing this value as the index. The slight change is that dPosX was being used to keep this count. In instances were an associated name is not present this is still the same behaviour, however if a name is present (dPosName) this will appear before the X value and as such will be used for the index instead.

To Test:
- Using the sample file from QA-17936, read the file and view the global metadata. 
- Without this PR there should be a single set of entries with the key dPos (dPosName, dPosX, dPosY, dPosZ etc)
- With this PR included there should be multiple entries of these position sets with an index beside them. The values should match up with those in the screenshot of forum thread https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8368&p=18538&sid=247d39d589db9edc0815d8868720803a#p18538

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -123,6 +123,7 @@ public class NativeND2Reader extends FormatReader {
 
   private boolean textData = false;
   private Double refractiveIndex = null;
+  private boolean hasMultipointNames = false;
 
   // -- Constructor --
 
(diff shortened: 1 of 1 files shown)


sbesson:
7855097 seems to have no negative impact on the existing ND2 sample files  - see [log](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/509/console).

Having done some minimal functional testing, comparing the core dimensions before after this PR:
- NIS elements 4.11.0 detects the ND2 files as 4 positions with 3 timepoints and 2 channels each

screen shot 2017-11-27 at 21 07 59

- Bio-Formats 5.7.2 detects 6 series with 2 timepoints and 2 channels each
- with this PR, Bio-Formats detects only 1 series with 12 timepoints and 2 channels each

dgault:
Moving to 5.8 as further investigation required into the incorrect number of timepoints

PR #3011, opened 2017-11-28
Add ScreenReader as an external reader

joshmoore:
In preparation for migrating ScreenReader from a specialized
branch to a separate repository, the upcoming class needs
registering in readers.txt.

**NB:** likely to migrate this to a `clisplit` merge branch unless it's generally decided to be tolerable to have a non-extant reader listed in the mainline.

Part of https://trello.com/c/bFTepbx5/6-idr-rebasing-scoping


joshmoore:
In preparation for extracting `ScreenReader.java` here is a list of all classes which were edited in a commit with `ScreenReader` (bioformats]$ git log --stat -S ScreenReader):

 * components/formats-bsd/src/loci/formats/itk/ITKBridgePipes.java (spurious)
 * .../loci/formats/utests/ScreenDetectionTest.java (removed)
 * .../test-suite/src/loci/tests/testng/FormatReaderTest.java (in commit message: 07db7067739d8238aed6ed8503ffaa57397d73b8)
```

**Comment by sbesson** [comment]
Happy to get this tested in `cisplit` first although this might be a suitable candidate for `develop` as another step to get rid of the `metadata` branches. On the nonextance,  the approach here is effectively no different from the one used with `SlideBook6Reader` albeit for different reasons.

Unclear whether the order of readers should be considered if you want to preserve serialization of existing files (although the decoupling job might introduce serialization breakages in all cases).

**Comment by joshmoore** [comment]
Trying to build the result of 

```
git filter-branch --index-filter '
                        git read-tree --empty
                        git reset $GIT_COMMIT -- components/formats-gpl/src/loci/formats/in/ScreenReader.java
                ' \
        -- --all -- components/formats-gpl/src/loci/formats/in/ScreenReader.java
```

followed by 

```
git filter-branch -f --prune-empty --subdirectory-filter components/formats-gpl/
```

now.

joshmoore:
Doh. Apparently, I had already done this: https://github.com/joshmoore/screen-reader/blob/master/pom.xml

sbesson:
Interestingly, the semantics of the `ScreenReader` in develop makes this be selected as the default reader for a lot of data (see the failures from [build 717](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATS-DEV-merge-full-repository/717/consoleText)). As well as making `ScreenReader` external, can this PR also remove `ScreenReader.java` from the codebase ?

joshmoore:
Alternatively, I can use a different package in the `screen-reader` repo. I'll move this PR to `next_infra` and then we can go from there.

joshmoore:
I've now removed `ScreenReader` here since it was shadowing the version from the screen-reader repo. Happy to _not_ do that and instead rename in screen-reader, but since this version was not activated in `readers.txt` deleting it may be more useful regardless. If we're going to rename the screen-reader class, we'll need to decide on a new, decoupled package name.

sbesson:
As this reader has never been activated and is extracted to its own repository, I do not see a reason not to remove it completely. Adding to tomorrow's agenda for a sign off from the @openmicroscopy/formats team.

sbesson:
Coming back to this. Do we want to switch the base branch to `develop` for potential inclusion in 5.8.0?

joshmoore:
Certainly doesn't _hurt_ to have the base changed the base.

sbesson:
Given the last runs of the repository tests, I see no negative impact of this PR. Assuming no veto by EOB, will include it for 5.8.0.

PR #3012, opened 2017-11-29
ScreenReader enhancements (rebased onto develop)

joshmoore:
This is the same as gh-2389 but rebased onto develop.

----

...ScreenReader items removed...

Also adds a new `overrideAxisTypes` method to `FileStitcher`, which allows to externally set the axis types before `setId` is called. In this case, `FileStitcher` uses the new `AxisGuesser(int[], ...)` (see 19ef8e25d6a8244575a78d031c8715331da6e337) that sidesteps axis types guessing (i.e., the `AxisGuesser` is only used for `getAdjustedOrder`).

Note that, without `overrideAxisTypes`, `ScreenReader` would have to call `FileStitcher.setAxisTypes` _after_ the call to `setId`, so `AxisGuesser` would perform its guessing work for nothing. After that, `ScreenReader` would have to call `reader.setId` again to make the wrapping `DimensionSwapper` pick up the changes, with an additional performance hit.

### DIFF
--- a/components/formats-bsd/src/loci/formats/AxisGuesser.java
+++ b/components/formats-bsd/src/loci/formats/AxisGuesser.java
@@ -103,6 +103,24 @@ public class AxisGuesser {
   /** Whether the guesser is confident that all axis types are correct. */
   protected boolean certain;
 
+  // -- Helpers --
+
+  private boolean swapZT(int sizeZ, int sizeT, boolean haveZ, boolean haveT) {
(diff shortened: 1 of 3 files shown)


joshmoore:
--rebased-from #2389

joshmoore:
Pushed a cherry-pick of 4541321c8793b2e7039355e53ed7a51b5410c638 from gh-2428 here to try to deal with performance issues.

PR #3013, opened 2017-11-30
Add first version of Dockerfile for running the tests standalone

joshmoore:
- use the jdk:8 image as the base
- builds Bio-Formats using the Ant build system
- defines the test-automated target as the entry-point
- defines expected paths for the image data and configuration files

With the assumptions above, an image can be built and tested as follows
$ docker build -t bf .
$ docker run -v /path/to/data:/opt/data -v /path/to/config:/opt/config bf

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,5 +16,7 @@ USER bf
 WORKDIR /opt/bioformats
 RUN /opt/ant/bin/ant clean jars tools
 
+ENV TZ "Europe/London"
+
 WORKDIR /opt/bioformats/components/test-suite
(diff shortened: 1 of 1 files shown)


bramalingam:
Tested the docker build locally (building step alone). Builds successfully and checked the travis builds. Everything looks fine according to me.

sbesson:
Thanks @bramalingam, merging this as step 1 in the intent of being consumed internally via the CI infrastructure. We can probably decide later the semantics about the entry points and whether the built image should be pushed to Hub.

PR #3014, opened 2017-12-01
Add SUPPORT.md

hflynn:
See https://trello.com/c/Gt8CIWJ0/598-support-files

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ Bio-Formats is a standalone Java library for reading and writing life sciences
 image file formats. It is capable of parsing both pixels and metadata for a
 large number of formats, as well as writing to several formats.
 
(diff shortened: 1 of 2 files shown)


hflynn:
Wasn't sure if we wanted to say anything about opening GH issues or not?

joshmoore:
re: issues -- certainly a number of people use them. Having the policy spelled out probably makes sense, but I don't know if it must happen here & now.

dgault:
This reads fine to me and the pages display correctly. All the links are correct and working and the link to  `SUPPORT.md` is also working. I agree that info on using GitHub issues probably could be clarified but I don't think its necessary to include with this PR and the links referenced in the PR are a good reflection of our policy.

PR #3019, opened 2017-12-10
Add setFilePatternIds method to ImporterOptions

ctrueden:
When coupled with `setGroupFiles(true)`, this lets you pass a file pattern directly to the `ImporterOptions` as an id, similarly to what you can do with the interactive `FilePatternDialog`.

See:

* http://lists.openmicroscopy.org.uk/pipermail/ome-users/2017-December/006824.html
* http://forum.imagej.net/t/group-images-with-bioformats-in-java-code/7893

The following code now works in my tests:

```java
final String id = "/Users/curtis/data/mri-stack/slice00<05-24>_mri-stack.tif";

ImporterOptions options = new ImporterOptions();
options.setUsingPatternIds(true);
options.setGroupFiles(true);
options.setId(id);

ImagePlus[] imps = BF.openImagePlus(options);
for (final ImagePlus imp : imps)
  imp.show();
```

(The TIFFs were created by splitting the MRI Stack sample and saving each as a separate slice.)

There is still some weirdness: the title bar reads `slice00<01-27>_mri-stack.tif` even though we used a narrower range; I guess a file pattern is still computed from the first file of the above pattern, even if it is not ultimately used. This same behavior can be seen when specifying the above pattern via the UI dialogs.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java
@@ -48,6 +48,7 @@ import loci.formats.ChannelFiller;
 import loci.formats.ChannelSeparator;
 import loci.formats.ClassList;
 import loci.formats.DimensionSwapper;
+import loci.formats.FilePattern;
 import loci.formats.FileStitcher;
 import loci.formats.FormatException;
 import loci.formats.FormatTools;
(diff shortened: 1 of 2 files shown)


dgault:
This seem likes a nice option to add as currently there is no nice way of doing this directly from the Bio-Formats plugin. 

The FilePattern dialog is able to able to handle this using the existing options but due to the sequencing of the import process this wont work from the Java example scenario listed. I was also playing around with other  possibilities similar to https://github.com/openmicroscopy/bioformats/blob/develop/components/bio-formats-plugins/utils/Mass_Importer.java but there is no simple way of doing this currently and reworking the import process is significantly more work.

This is a nicer solution and the naming of the options also matches up with the existing naming conventions for setting patterns.

For testing I used a small Java plugin similar to that listed above and a range of different patterns based on the examples in https://docs.openmicroscopy.org/bio-formats/5.7.2/formats/pattern-file.html. All tests passed as expected with the correct selection of files being opened.

The issue with the title bar is also present when using the standard FilePatternDialog so is not specifically related to this PR and is something which can be followed up on seperately.

dgault:
With all the file pattern testing passing this PR is ready for inclusion in the 5.8.0 release.

PR #3020, opened 2017-12-11
Configuration: Use raw physicalSize values not FormatTools

dgault:
Using FormatTools for the physicalSize values was applying rounding which was inconsistent with the values stored in the readers under certain scenarios. 

Other FormatTools functions such as getTime and getWavelength do not apply any such rounding and as such should not have any impact.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -262,7 +262,8 @@ public class Configuration {
     String physicalSize = currentTable.get(PHYSICAL_SIZE_X);
     String sizeXUnits = currentTable.get(PHYSICAL_SIZE_X_UNIT);
     try {
-      return physicalSize == null ? null : UnitsLength.create(new Double(physicalSize), UnitsLength.fromString(sizeXUnits)); 
+      UnitsLength xUnits = sizeXUnits == null ? UnitsLength.MICROMETER : UnitsLength.fromString(sizeXUnits);
+      return physicalSize == null ? null : UnitsLength.create(new Double(physicalSize), xUnits); 
     }
     catch (NumberFormatException e) { }
     catch (EnumerationException e) { }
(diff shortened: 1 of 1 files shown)


components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -260,32 +262,33 @@ public Length getPhysicalSizeX() {
     String physicalSize = currentTable.get(PHYSICAL_SIZE_X);
     String sizeXUnits = currentTable.get(PHYSICAL_SIZE_X_UNIT);
     try {
-      return physicalSize == null ? null : FormatTools.getPhysicalSize(new Double(physicalSize), sizeXUnits);
-    }
-    catch (NumberFormatException e) {
-      return null;
+      return physicalSize == null ? null : UnitsLength.create(new Double(physicalSize), UnitsLength.fromString(sizeXUnits));
```
sbesson: The `FormatTools.getPhysicalSize()` also includes the knowledge of the default unit for the physical sizes (`UNITS.MICROMETER`). Here, if `PHYSICAL_SIZE_X_UNIT` is not specified in the configuration table e.g. for old configuration files, the getter will return `null` and cause the tests to fail - see https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-breaking-repository-subset/292/console.

To reduce the number of false positives, the unit length creation and`EnumerationException` should probably be handled separately. Eventually we might want to enforce the unit specification in the configuration files but at this is probably too much for the scope of this PR.

sbesson:
:+1: have you run this against a small representative subset of file formats ? I will try to move this to the daily queue once we have https://github.com/openmicroscopy/bioformats/pull/3013 sorted.

sbesson:
With minimal configuration adjustments, these changes have been [successfully tested](https://web-proxy.openmicroscopy.org/next-ci/job/BIOFORMATS-test-repo/76/) against the full data repository. Let's discuss the next steps when coming back from the break.

sbesson:
Discussed this work and its impact on Wednesday at the formats meeting. The changes proposed here make sense and allow to test for small physical size values e.g. as stored in OME-XML. This change is directly applicable to the testing of sample datasets for various supported file formats like Imspector OBF - see https://github.com/openmicroscopy/bioformats/pull/2950. Thanks @dgault for the investigation, merging.

PR #3031, opened 2018-01-05
OME-TIFF: try harder to preserve metadata when files are ungrouped

melissalinkert:
This should only affect multi-file OME-TIFF datasets, and only when file grouping is disabled (e.g. via the ```-nogroup``` option to ```showinf```). 

To test, use the ```tubhiswt-4D``` dataset from http://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/

Without this change, ```showinf -nogroup -omexml``` with ```tubhiswt_C0_TP0.ome.tif``` and ```tubhiswt_C1_TP0.ome.tif``` should show 10 timepoints and 1 Z section, with OME-XML that does not represent everything in the file's OME-XML comment.  ```showinf -nopix tubhiswt_C0_TP0.ome.tif``` should show that the entire dataset's dimensions are 43 timepoints, 10 Z sections, and 2 channels; each file contains all Z sections for a single timepoint and channel.

With this change, ```showinf -nogroup -omexml``` with ```tubhiswt_C0_TP0.ome.tif``` and ```tubhiswt_C1_TP0.ome.tif``` should correctly show 1 channel, 1 timepoint, and 10 Z sections.  The OME-XML should largely match what is in the file, apart from ```Pixels.SizeZ```, ```Pixels.SizeC``` and ```Pixels.SizeT```.

I would expect all tests to remain green.  I think this should be safe for a patch release, but it is not critical for 5.7.3.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -66,6 +66,11 @@ import loci.formats.tiff.PhotoInterp;
 import loci.formats.tiff.TiffIFDEntry;
 import loci.formats.tiff.TiffParser;
 
+import ome.xml.meta.OMEXMLMetadataRoot;
+import ome.xml.model.Channel;
+import ome.xml.model.Image;
+import ome.xml.model.Pixels;
+import ome.xml.model.Plane;
 import ome.xml.model.primitives.NonNegativeInteger;
 import ome.xml.model.primitives.PositiveInteger;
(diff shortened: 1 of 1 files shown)


dgault:
The code changes here look to be good and the logic used for multi file ungrouped datesets makes sense.

The PR was tested using the `tubhiswt-4D` dataset.  Without the PR the file opens and shows 10 timepoints and 1 Z. This does not match with the OME-XML. With the PR included the file opens and shows 1 timepoint and 10 Z sections as expected.

PR is good to merge.

PR #3033, opened 2018-01-09
NRRD: add support for GZIP pixel stream contained within a .nrrd file

melissalinkert:
This allows for the case when the file itself is uncompressed, but the pixels block after the header is a single GZIP stream.  The implementation is modeled after a similar feature in ```ICSReader```.

To test, use ```data_repo/curated/nrrd/samples/gzip/dt-helix.nrrd.gz```, which was generated from ```data_repo/curated/nrrd/samples/dt-helix.*```.  Without this change, ```showinf -separate -normalize dt-helix.nrrd.gz``` should throw:

```
Exception in thread "main" loci.formats.UnsupportedCompressionException: Unsupported encoding: gzip
	at loci.formats.in.NRRDReader.openBytes(NRRDReader.java:175)
	at loci.formats.FormatReader.openBytes(FormatReader.java:886)
	at loci.formats.ImageReader.openBytes(ImageReader.java:441)
	at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:334)
	at loci.formats.tools.ImageInfo.readPixels(ImageInfo.java:818)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1051)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1117)
```

With this change, ```showinf -separate -normalize dt-helix.nrrd.gz``` should read all planes without an exception, and the pixels and metadata should match what is reported for ```data_repo/curated/nrrd/samples/dt-helix.nhdr```.

I would expect builds to remain green and memo files to be unaffected.  I think this would be safe for a patch release, but is not critical for 5.7.3.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/NRRDReader.java
+++ b/components/formats-bsd/src/loci/formats/in/NRRDReader.java
@@ -33,7 +33,10 @@
 package loci.formats.in;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
 import java.io.IOException;
+import java.util.zip.GZIPInputStream;
 
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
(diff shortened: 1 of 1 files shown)


chris-allan:
This was mistakenly merged by me. @sbesson is going to force push this away from `HEAD` of `develop` so the PR will have to be opened again.

Apologies to all.

PR #3036, opened 2018-01-10
VSI: Add support for Lossless JPEG

dgault:
This PR is in relation to https://www.openmicroscopy.org/community/viewtopic.php?t=8425&p=18953
The associated Trello card is https://trello.com/c/fUhH0Cxo/141-cellsens-vsi-npe-in-openbytes

Sample files can be found in QA-18870 and QA-17728

The issue was caused by an unknown compression type of value 5. From the user feedback this has been identified as Lossless JPEG. This PR adds support using the existing LosslessJPEG codec.

To test:
- All builds and tests should remain green
- With this PR opening QA-18870 should result in a NullPointerException
- With this PR opening QA-18870 should display the images correctly without error

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -609,6 +609,7 @@ public class CellSensReader extends FormatReader {
       metadataIndex = -1;
       previousTag = 0;
       expectETS = false;
+      pyramids.clear();
     }
   }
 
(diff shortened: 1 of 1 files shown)


sbesson:
With the last commit, all automated tests have been successfully running against the `cellsens` folder and the failures reported in https://github.com/openmicroscopy/bioformats/pull/3036#issuecomment-359514291 when testing multiple CellSens filesets while recycling the `ImageReader` are gone. Merging.

PR #3038, opened 2018-01-10
CZI: preserve DisplaySetting metadata in the original metadata table

melissalinkert:
Only one set of ```DisplaySetting``` data is preserved, so that datasets with
one set of settings per plane or tile do not cause the original metadata
table to be too big.  Most of these values are available elsewhere, but the optional display gamma value is only stored under ```DisplaySetting```.

See https://github.com/openmicroscopy/bioformats/commit/169ae85ec4cb8282ad03310714f44a0eaf51ced8 for the original commit disabling ```DisplaySetting``` population.

To test, use pretty much any dataset from ```data_repo/curated/zeiss-czi```; I used ```greg/Experiment-519.czi``` and ```silke/White_574.czi```.  Without this change, ```showinf -nopix``` should not show any original metadata keys beginning with ```DisplaySetting```.  With this change, the same test should show several keys beginning with ```DisplaySetting|Channel```.  The number of ```DisplaySetting``` keys should not vary with the number of planes or tiles.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -3120,9 +3120,6 @@ public class ZeissCZIReader extends FormatReader {
 
   private void populateOriginalMetadata(Element root, Deque<String> nameStack) {
     String name = root.getNodeName();
-    if (name.equals("DisplaySetting")) {
-      return;
-    }
     nameStack.push(name);
 
     final StringBuilder key = new StringBuilder();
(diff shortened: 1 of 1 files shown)


dgault:
The code changes look fine to me and the repo jobs have been green with this PR included.

I tested the PR against a number of files from `curated/zeiss-czi ` and it behaved as expected with `DisplaySetting` added to global metadata with the number of entries not being affected by the number of planes. An example of the generated metadata is as below:

```
DisplaySetting|Channel|ChannelUnit|ChannelType = Unspecified
DisplaySetting|Channel|ChannelUnit|FactorI = 1
DisplaySetting|Channel|ChannelUnit|OffsetI = 0
DisplaySetting|Channel|ChannelUnit|UnitI = Unknown
DisplaySetting|Channel|Color = #FFFFFF
DisplaySetting|Channel|ColorMode = Color
DisplaySetting|Channel|Gamma = 1
DisplaySetting|Channel|High = 0.080436966807329133
DisplaySetting|Channel|Id = 11940272116114987513732794182069686880
DisplaySetting|Channel|IsSelected = true
DisplaySetting|Channel|Low = 0.0022888532845043367
DisplaySetting|Channel|Name = Cam1
```
PR is good to merge.

PR #3044, opened 2018-01-15
OBFReader: Use Java variant of K&R braces and remove white space

dgault:
As a follow up to https://github.com/openmicroscopy/bioformats/pull/2950, this PR stanardises the formatting in the OBFReader to bring it inline with other readers and https://docs.openmicroscopy.org/bio-formats/5.7.3/developers/code-formatting.html

There are no functional changes in this PR, only formatting.

To test:
- Builds and tests should remain green


sbesson:
Diff makes sense and does not seem to impact the automated tests against the Imspector files in our repositories. I have been looking for a more systematic way to review/enforce our formatting style. https://github.com/google/google-java-format could be a candidate but the [Google Java style](https://google.github.io/styleguide/javaguide.html) is stricter than our current Bio-Formats formatting.

PR #3048, opened 2018-01-19
MinMaxCalculator: switch from series to core index based indexing

melissalinkert:
Backported from a private PR.  This allows min and max values to be calculated for each core index instead of each series, so that the values are not dependent on ```hasFlattenedResolutions()```.

f4a5ad2 is the test case, and 0d8a155 is the fix.  With f4a5ad2 alone, running unit tests should show one failure in ```MinMaxCalculatorTest.testMultipleResolutions```.  With both commits, all unit tests should pass.

### DIFF
--- a/components/formats-bsd/src/loci/formats/MinMaxCalculator.java
+++ b/components/formats-bsd/src/loci/formats/MinMaxCalculator.java
@@ -113,7 +113,7 @@ public class MinMaxCalculator extends ReaderWrapper {
       throw new FormatException("Invalid channel index: " + theC);
     }
 
-    int series = getSeries();
+    int series = getCoreIndex();
 
     // check that all planes have been read
     if (minMaxDone == null || minMaxDone[series] < getImageCount()) {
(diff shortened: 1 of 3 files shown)


dgault:
The new unit test included here looks sensible and adds coverage for multi resolution. The behaviour to 
flattened resolutions appears to be unaffected by these changes and this has no effect on the repo tests and builds which remain green.

With only the test included there is a failure:
```
FAILED: testMultipleResolutions
java.lang.AssertionError: expected:<-1.0> but was:<-2.0>
	at org.testng.AssertJUnit.fail(AssertJUnit.java:59)
	at org.testng.AssertJUnit.failNotEquals(AssertJUnit.java:364)
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:80)
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:88)
```

With the changes to MinMacCalculator also included all of the tests pass as is the correct behaviour:
```
PASSED: testMultipleResolutions
PASSED: testValidMinMax
PASSED: testValidMinMaxBothHalvesLowerFirst
PASSED: testValidMinMaxBothHalvesUpperFirst
PASSED: testValidMinMaxDoesntRecalculateOnFullPlane
PASSED: testValidMinMaxFirstHalf
PASSED: testValidMinMaxNoOutOfBufferInspection
PASSED: testValidMinMaxSecondHalf
PASSED: testValidOpenBytes

===============================================
    Default test
    Tests run: 9, Failures: 0, Skips: 0
===============================================
```

PR #3049, opened 2018-01-19
Zeiss CZI: use Z step when Z positions are not enumerated

melissalinkert:
Backported from private PRs.  If a Z step is present, then the values for ```Plane.PositionZ``` are calculated by using the first plane's position and the Z step.  This prevents ```Plane.PositionZ``` from being set to the same value for all Z slices in a dataset.  This is typically a problem in datasets with fused tiles.

To test, use the files whose configurations are updated in https://github.com/openmicroscopy/data_repo_config/pull/258.  Open each in ZEN, and note the relative Z position and units for each plane by mousing over the ```Z-Position``` slider in the bottom left portion of the window.  ```showinf -omexml -nopix``` should show absolute positions in ```PositionZ``` on ```Plane```, with the same spacing and units as the relative Z positions in ZEN.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1339,9 +1339,19 @@ public class ZeissCZIReader extends FormatReader {
         }
         else if (positionsZ != null && i < positionsZ.length) {
           int zIndex = getZCTCoords(plane)[0];
-          double step = zIndex * zStep.value().doubleValue();
-          Length pos = new Length(step + positionsZ[i].value(zStep.unit()).doubleValue(), zStep.unit());
-          store.setPlanePositionZ(pos, i, plane);
+          if (positionsZ[i] != null) {
+            if (zStep != null) {
+              double value = positionsZ[i].value(zStep.unit()).doubleValue();
+              if (zStep != null) {
+                value += zIndex * zStep.value().doubleValue();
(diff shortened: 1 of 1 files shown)


dgault:
The builds and tests have been green with this PR and its associated config PR included. The code to calculate the update psoitions list looks good. The only question I would have is the use of both UNITS.METRE and UNITS.MICROM, though none of the files in our repo use METRE.

The PR has now been tested with all of the filesets from https://github.com/openmicroscopy/data_repo_config/pull/258 and the updated position values appear to be correct.

melissalinkert:
I think I have a private sample file that demonstrates why ```UNITS.METRE``` is correct, but it's unlikely I'll be able to make a shareable sample before EOB today (and if I'm wrong, this will need another commit anyway).  For some reason I thought a few of the files in https://github.com/openmicroscopy/data_repo_config/pull/258 would show this, but all seem to pass when replacing ```UNITS.METRE``` with ```UNITS.MICROM```.

sbesson:
Discussed with @melissalinkert and @dgault. Since none of the samples we currently have in the OME data repository is affected by the unit question raised in https://github.com/openmicroscopy/bioformats/pull/3049#issuecomment-366933482, this change is probably safe to integrate as it is for 5.8.0.
Having some shareable sample allowing to reproduce the behavior would also be useful at some point. If we receive bug reports together with supporting sample files, we can review this code path as necessary.

PR #3050, opened 2018-01-19
ImagePro: add support for .ips files

melissalinkert:
See https://trac.openmicroscopy.org/ome/ticket/4180

The .ips file is an optional binary metadata file that supplies channel names and a list of ImagePro TIFF files to be grouped together.  There is one test dataset in ```data_repo/curated/imagepro/alex```.  With this change, ```showinf -omexml``` on the .ips file should show 2 series (positions), 2 channels, 3 timepoints, and 10 Z sections (consistent with the TIFF file names).  The OME-XML should be valid and should contain non-null channel names for both channels in each series.  All TIFF files should appear on the used file list.

The behavior of ```showinf``` on one of the TIFF files should not be affected by this PR, as the reader will not attempt to find an .ips file if a TIFF file was passed to ```setId```.  This could be changed if desired, but would slow down initialization for datasets without an .ips file (which is the majority of what we have).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SEQReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SEQReader.java
@@ -27,10 +27,13 @@ package loci.formats.in;
 
 import java.io.IOException;
 
+import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.formats.CoreMetadata;
 import loci.formats.FormatException;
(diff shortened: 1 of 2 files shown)


sbesson:
Excluding from the daily builds for the scope of 5.9.0 as the changes in this PR would break memo files which is outside the contract of the release.

PR #3051, opened 2018-01-19
InCell: parse min and max values from .xdce file

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/13247

If minimum and maximum pixel values are present in an InCell .xdce file, then they will be parsed and stored in the reader's ```MetadataStore``` if it is an instance of ```IMinMaxStore``` (expected for OMERO imports).  This should reduce min/max calculation time and thereby improve import performance.

With this PR, ```showinf -trace -minmax``` on an .xdce that defines min/max values should show ```TRACE``` logging indicating that min and max values are parsed, and these should match the min and max values reported by the ```-minmax``` flag.  These values should also match the minimum and maximum reported by ```showinf -minmax``` without this PR (i.e. the values have not changed).

I was testing with the ```data_repo/curated/incell/public/```; some of the older plates (e.g. ```data_repo/curated/incell/michael```) do not define minimum and maximum pixel values, so will not show this feature.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -42,6 +42,7 @@ import loci.formats.FormatException;
 import loci.formats.FormatReader;
 import loci.formats.FormatTools;
 import loci.formats.MetadataTools;
+import loci.formats.meta.IMinMaxStore;
 import loci.formats.meta.MetadataStore;
 
 import ome.xml.model.enums.Binning;
@@ -633,6 +634,7 @@ public class InCellReader extends FormatReader {
         int field = getFieldFromSeries(i);
         int timepoint = oneTimepointPerSeries ?
           i % channelsPerTimepoint.size() : 0;
+        Double[][] minMax = new Double[getEffectiveSizeC()][2];
         for (int time=0; time<getSizeT(); time++) {
           if (!oneTimepointPerSeries) timepoint = time;
           int c = channelsPerTimepoint.get(timepoint).intValue();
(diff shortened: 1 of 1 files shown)


sbesson:
Tested this change from the OMERO perspective using one of the [public Incell2000 datasets](http://downloads.openmicroscopy.org/images/HCS/INCELL2000/INMAC384-DAPI-CM-eGFP_59223_1/).

Importing the plate into a 5.4 server without this PR included, the full import takes ~17m of which 11min are spent in the pixeldata process (see [Incell2000_import.log](https://github.com/openmicroscopy/bioformats/files/2079657/Incell2000_import.log))

| Step              | Time        | Notes                        |
|-------------------|-------------|------------------------------|
| Detection/Parsing | few seconds | multi-file                   |
| Upload/transfer   | 3min30s     | in-place (ln_s), no checksum |
| Registration      | 30s         | no checksum                  |
| Pixeldata         | 11min       | recalculating min/max        |
| Thumbnails        | 2min        |                              |

Tried running the same process into another OMERO server with this PR included and the metrics are pretty much unchanged i.e. no performance improvement.

Discussing with @melissalinkert , the reason for the absence of performance improvement is probably outside the scope of this PR. In short, the internal Memoizer logic uses a store which does not implement `IMinMaxStore` so it is likely that the min/max values are lost during memoization and cannot be re-used in later import steps. At the moment, only 2 readers retrieve the channel min/max (Deltavision and BioRad). However, as we start serious work on import performances, this is probably a problem we will have to solve in order to enable substantial gain for some of our file formats. /cc @joshmoore @jburel 

Depending on the cost of doing the above, an intermediate workaround might be to think of exporting the minmax values in a format that could be digested by the `omero-cli-render` plugin e.g. as a hook of `showinf`:

```
$ showinf -minmax INMAC384-DAPI-CM-eGFP_59223_1.xdce --export minmax=render.yml
$ screen_id= $(bin/omero import --transfer=ln_s --skip=checksum --skip=minmax INMAC384-DAPI-CM-eGFP_59223_1.xdce)
$ bin/omero render set $screen_id render.yml
```

sbesson:
No regression and the extra parsing make sense. We can review this if we get to tackle  https://trello.com/c/QdCI9tHU/407-memoizer-does-not-pass-along-min-max-values for Bio-Formats 6.0.0.

PR #3052, opened 2018-01-22
formats-bsd: tiff: Handle OnDemandLongArray in IFD.printIFD()

rleigh-codelibre:
Trello: https://trello.com/c/ylVCzWR0/24-crash-with-trace-log-level

This missing case broke TRACE log output.

Testing:
- set `level=TRACE` in `tools/logback.xml`
- run showinf on a tiled TIFF
- will throw when it hits the offsets/byte count fields calling `printIFD` in `TiffParser`.
- will stop throwing and correctly dump the IFD values with this patch applied


dgault:
With this PR showinf does not report any errors when log level is set to TRACE. The offsets and byte counts are displayed correctly as expected. PR is good to merge.

PR #3053, opened 2018-01-22
formats-bsd: TiffWriter: Don't clobber the tile size between series

rleigh-codelibre:
Trello: https://trello.com/c/CQo5glGl/23-tiling-size-bug-creates-broken-tiff-files

The tile size fields were being overwritten with the series image height and width.  With multi-series images, this leads to subsequent series having incorrect tile sizes.  Never overwrite the set size.

This is quite a nasty bug, since it can create corrupt TIFF files with tile sizes which aren't a multiple of 16.

Testing:
- Try running `bfconvert` with the SCN file from QA#17671.
- Check the tile sizes with `tiffinfo | less`, they should all be 512x512 but everything after the first IFD will have the image width and height from the previous series as the tile width and height
- Apply this patch and rebuild.  Testing should now show 512x512 for all IFDs


dgault:
The code changes here all look fine. PR was tested using QA#17671 as suggested. Without the PR included the tilesizes for the first series are correct and subsequent ifds contain display incorrect sizes. The tilesizes with the PR included are all correct and read as 512x512.

PR #3054, opened 2018-01-22
tools: ImageInfo: Report tile size in core metadata

rleigh-codelibre:
Trello: https://trello.com/c/v6sJ0mQ2/25-showinf-report-tile-sizes-for-each-series

Useful for checking the tile size for any reader to ensure the tile sizes are correct.

Testing:
- run showinf with a tiled tiff and an untiled tiff.  Both should report the "optimal" tile size values for each series.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -573,6 +573,8 @@ public class ImageInfo {
       Modulo moduloT = reader.getModuloT();
       int thumbSizeX = reader.getThumbSizeX();
       int thumbSizeY = reader.getThumbSizeY();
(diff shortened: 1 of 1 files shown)


mtbc:
Goodness, for `8kx8k.jpg` we have `Tile size = 8000 x 8000` suggested!

mtbc:
Outwith this PR, should tile size become part of the `ome/data_repo/config` metadata that's checked from `.bioformats` files or is that a wholly separate thing?

rleigh-codelibre:
I think it's something which we could add to the expected test results; it couldn't hurt and would check codepaths which we currently aren't testing.  It might catch regressions in some readers.

sbesson:
Collected the proposed extension of the testing framework as RFE. Merging as this as an improvement that can be released with 5.8.0

PR #3055, opened 2018-01-22
tools: ImageConverter: Add -noflat option

rleigh-codelibre:
Trello: https://trello.com/c/1fuP0veq/26-bfconvert-add-noflat-option

This is to match the `showinf` behaviour using the `-noflat` option already present there.  This will enable the creation of OME-TIFF without subresolutions, or with flat subresolutions (the default).

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -108,6 +108,7 @@ public final class ImageConverter {
   private int saveTileWidth = 0, saveTileHeight = 0;
   private boolean validate = false;
   private boolean zeroPadding = false;
+  private boolean flat = true;
 
   private IFormatReader reader;
   private MinMaxCalculator minMax;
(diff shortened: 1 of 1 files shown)


sbesson:
Tested using the merge command line tools and a SVS file (3 series, the first one containing 3 pyramidal levels). Without any option, `bfconvert 450CD3_2016-04-12.tif out.ome.tiff` creates 5 series while `bfconvert  -noflat 450CD3_2016-04-12.tif out.ome.tiff` creates 3 series as described

Primary question is with regard to the expectations of this new utility API and the backwards compatibility. As none of the writers  support the writing of subresolutions, the current behavior of the API is effectively to write the  highest level of resolution for each non-flattened series. 
How would this behavior be modified when some writers implement subresolution support? Would the `bfconvert -noflat` contract remain unchanged and another API would be required to choose the layout of the target image? Or would the layout become conditional to the level of support offered by the writer i.e. write subresolutions if available?

rleigh-codelibre:
Once we have added subresolution support to our writers, `-noflat` will then be creating the subresolutions as well.  It's basically going to mirror the behaviour of showinf for writing.

rleigh-codelibre:
Rebased onto `east`.

PR #1944, opened 2015-08-20
SimplePCI .cxd: adjust SizeX if extra bytes are stored in the planes

melissalinkert:
Fixes https://trac.openmicroscopy.org/ome/ticket/12886.

To test, use the file from QA 11073.  Without this change, SizeX should be 397, and the images should look incorrect as noted in the file attached to the ticket.  With this change, SizeX should be 398, and the images be an obvious split field of view with moving artifacts (you will likely need to open in ImageJ to see this, as it's 16-bit data).

PR #3056, opened 2018-01-25
PCI: Don't modify the width when binning is used

dgault:
This issue has been raised a few times, see Trello card https://trello.com/c/MQYsukMf/132-pci-reader-incorrect-width

In a previous PR it was necessary to adjust the width if extra bytes were available https://github.com/openmicroscopy/bioformats/pull/1944, however in instances were binning is used there appears to extra bytes available but no need to adjust the size.

This is not an ideal solution, it would much better if we had an understanding of what exactly the extra bytes are being used for. However if it provides a solution for users then it may be worth including.

Sample QA files can be found in QA-17696 and QA-19182

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/PCIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/PCIReader.java
@@ -387,8 +387,8 @@ public class PCIReader extends FormatReader {
     String file = imageFiles.get(0);
     RandomAccessInputStream s = poi.getDocumentStream(file);
     TiffParser tp = new TiffParser(s);
-    // don't correct the image width if it's stored as a TIFF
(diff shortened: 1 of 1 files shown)


sbesson:
This change seems to impact one of the existing PCI sample files (see [log](https://web-proxy.openmicroscopy.org/next-ci/job/BIOFORMATS-test-folder/6331/console) both in terms of pixel data and core metadata. Is that expected?

dgault:
All failures are related to qa-11073, I will need to test to see if the changes to this file are correct. The fact that the width modification is by a single pixel might make it difficult.

dgault:
Looking at the file and the adjusted width appears to be the correct one. Marking the PR as breaking for further investigation

--breaking

PR #3063, opened 2018-02-06
Volocity: expand Image names to include the stack parent name(s)

melissalinkert:
Backport of a private PR.  Volocity allows images to be arranged into separate directories within the .mvd2 database.  This updates the image names to preserve the directory names (if any), so that the original hierarchy is not lost.

To test, check the datasets with changed configurations as noted in the forthcoming configuration PR.  The only changes should be to some image names; in each case, the new name should end with ```/``` plus the old name.

With both PRs included, builds should continue to pass and memo files should be unaffected.


melissalinkert:
2293c88 fixes the Findbugs warning flagged in https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-build/903/

dgault:
Tested with the files listed in the config PR, all the files open as expected, the only metadata changes are to the names which also look to be correct.

PR #3064, opened 2018-02-06
Inveon: attempt to find renamed data files

melissalinkert:
Backported from a private PR.  The data file name stored in the header is not updated if the header and
data files are renamed. If the stored data file name does not exist,
then a file whose name is similar to the header file's name will be used
(if such a file is present).

To test, use the dataset in ```data_repo/curated/inveon/samples/rename-test```.  Without this PR, ```showinf batch1_v1.ct.img.hdr``` should throw a ```FileNotFoundException```.  With this PR, the same test should display images and metadata matching ```showinf data_repo/curated/inveon/samples/batch1_v1.ct.img.hdr```.  The used file list should contain two files, both of which exist.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InveonReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InveonReader.java
@@ -27,6 +27,7 @@ package loci.formats.in;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.ArrayList;
 
 import loci.common.DataTools;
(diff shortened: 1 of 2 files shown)


dgault:
Builds and test have been green with this PR included. The code changes make sense and look fine.

Tested using the sample files at `data_repo/curated/inveon/samples/rename-test`. Without this PR showinf produces a `FileNotFoundException` as described. With the PR included the file is read without error and displays correctly.

PR #3066, opened 2018-02-07
Fake: allow per-plane ExposureTime/Position{X,Y,Z} keys in the INI file

melissalinkert:
Backported from a private PR.

The INI file can now optionally contain one table per series, named "series_n"
(where n is the 0-based series index).  Each table can describe the
exposure time, X position, Y position, and Z position for any subset of
the planes in the series, using the following keys (where i is the
0-based plane index):

```
ExposureTime_i
ExposureTimeUnit_i
PositionX_i
PositionXUnit_i
PositionY_i
PositionYUnit_i
PositionZ_i
PositionZUnit_i
```

This corresponds to the keys used by the ```.bioformats``` files used for data
testing.

To test, use the new example in ```data_repo/curated/fake/samples/ini```.  ```showinf -nopix -omexml multiSeries-metadata.fake``` with this change should show exposure times and plane positions populated according to the key/value pairs in ```multiSeries-metadata.fake.ini``` (each plane/series will be a little different).  The new configuration file can also be compared to ```multiSeries-metadata.fake.ini```, as the exposure time and plane position key/values should be nearly identical.

This should be safe for a patch release if needed; memo files and builds should be unaffected.

Semi-related ```FakeReader``` cards:

https://trello.com/c/t6McetbQ/261-unify-key-value-parsing-between-fake-files-and-bioformats
https://trello.com/c/l2bTBOik/265-fake-series-of-different-sizes

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -1138,7 +1138,7 @@ public class FakeReader extends FormatReader {
 
       if (exposureTime != null) {
         try {
-          Double v = new Double(exposureTime);
+          Double v = Double.valueOf(exposureTime);
           Time exposure = FormatTools.getTime(v, exposureTimeUnit);
           if (exposure != null) {
             store.setPlaneExposureTime(exposure, newSeries, i);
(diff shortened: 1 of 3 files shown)


sbesson:
I have not reviewed the code details yet. Generally adding support for per-series metadata in our generated test images sounds useful. Two general comments:

- the [documentation](http://docs.openmicroscopy.org/bio-formats/5.8.0/developers/generating-test-images.html) should probably be expanded to document the new keys (we might need some additional column to mention the keys are INI file only)
- for testing, the current approach is to add representative sample fake files to the curated data repository. This addition could be supplemented or even completely superseded by an extension of the `FakeReaderTest` class to cover the new semantics. Thoughts?

components/formats-bsd/src/loci/formats/in/FakeReader.java
```diff
@@ -1113,6 +1125,110 @@ private void fillRegions(MetadataStore store, int imageIndex) {
...
+    int s = getSeries();
+    setSeries(newSeries);
+
+    for (int i=0; i<getImageCount(); i++) {
+      String exposureTime = table.get("ExposureTime_" + i);
+      String exposureTimeUnit = table.get("ExposureTimeUnit_" + i);
+
+      if (exposureTime != null) {
+        try {
+          Double v = new Double(exposureTime);
```
mtbc: `Double.valueOf`

components/formats-bsd/src/loci/formats/in/FakeReader.java
```diff
@@ -1113,6 +1125,110 @@ private void fillRegions(MetadataStore store, int imageIndex) {
...
+          if (exposure != null) {
+            store.setPlaneExposureTime(exposure, newSeries, i);
+          }
+        }
+        catch (NumberFormatException e) {
+          LOGGER.trace("Could not parse ExposureTime for series #" + s + " plane #" + i, e);
+        }
+      }
+
+      String positionX = table.get("PositionX_" + i);
```
jburel: We have 3 time more or less the same code block
It will be better to have a common method e.g. ``convert(String position, String positionUnit, int axis)`` where axis will indicate the method to use ``0= setPlanePositionX 1= setPlanePositionY, 2= setPlanePositionZ``

mtbc: Maybe with a TODO that with Java 8 just pass the method reference!

melissalinkert:
Comments from @mtbc and @jburel addressed in 1bf3055.  I will plan to add documentation and unit tests before Monday.

melissalinkert:
Last 3 commits address remaining code review comments, update the fake file documentation page, and add a unit test for the new keys.

dgault:
The updated documentation reads and displays well. The included sample also looks correct and contains suitable examples. The new unit test also runs successfully and contains a suitable test case for this scenario. All of the previous comments have been addressed and this PR is now ready to merge.

PR #3069, opened 2018-02-09
Add links to public NRRD samples

hflynn:
See https://trello.com/c/aKhTXS7n/31-nrrd-public-samples-and-documentation


sbesson:
Makes sense. I realize there is some duplication between the official samples and our own public samples but this is true for other formats as well. Arguably, all official samples should be part of our public data repository and the provenance would be captured per dataset.

PR #3065, opened 2018-02-07
MicroManagerReader: Minor fix to handle metadata block closing

bramalingam:
ref: https://trello.com/c/dqTlLuXf/637-serge-pelet-single-cell-data#comment-5a783cf80e00f887aa369c3f

Testing Instructions:
1) Test via Fiji, Command line tools and check if the file opens appropriately via Bio-Formats.
Without this PR you should get an exception (as shown in the trello card).
2) Check if the subset job listed below is green.

@sbesson

PR #3071, opened 2018-02-15
MicroManagerReader: Minor fix to handle metadata block closing (rebased onto metadata54)

bramalingam:
This is the same as gh-3065 but rebased onto metadata54.

----

ref: https://trello.com/c/dqTlLuXf/637-serge-pelet-single-cell-data#comment-5a783cf80e00f887aa369c3f

Testing Instructions:
1) Test via Fiji, Command line tools and check if the file opens appropriately via Bio-Formats.
Without this PR you should get an exception (as shown in the trello card).
2) Check if the subset job listed below is green.

@sbesson


bramalingam:
--rebased-from #3065

sbesson:
Successfully tested via https://trello.com/c/dqTlLuXf/637-serge-pelet-single-cell-data-idr0040-aymoz-singlecell

PR #3073, opened 2018-02-15
Ionpath MIBI reader fixes

sbesson:
Follow-up of https://github.com/openmicroscopy/bioformats/pull/2959, this PR:

- removes the imaging domain API and uses `UNKNOWN_DOMAIN` for the new reader - see https://github.com/openmicroscopy/bioformats/pull/2959#issuecomment-365058850 for the discussion
- updates the new reader to comply with the Bio-Formats formatting style (performed using https://github.com/google/google-java-format/)

This PR should not affect the existing data repository tests using the IonPath MIBI sample file. For code review, it might be useful to use https://github.com/openmicroscopy/bioformats/pull/3073/files?diff=split&w=1 to ignore the whitespace changes.

I have not included a format documentation page. This will need to be done separately, probably after the inclusion of https://github.com/openmicroscopy/bioformats/pull/3072 to minimize conflicts.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -331,9 +331,6 @@ public final class FormatTools {
 
   /** Identifies the astronomy domain. */
   public static final String ASTRONOMY_DOMAIN = "Astronomy";
-  
-  /** Identifies the secondary ion mass spectrometry domain. */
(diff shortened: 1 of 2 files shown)


dgault:
Using `UNKNOWN_DOMAIN` rather than the API additions is inline with other similar usages that we have and makes sense. The formatting changes look good. Merging for 5.8.0

PR #3084, opened 2018-02-26
Flex: read image offsets when a single tile is used

melissalinkert:
Backported from a private PR.

If every image is contained in a single tile, we likely can't reliably calculate the image offsets.  Reading all of the offsets is a little slower, but is more accurate.

To test, use the three single-file datasets referenced in the config PR:

```
data_repo/curated/flex/qa-7469/002002.flex
data_repo/curated/flex/lee/RLM1 SSN3 300308 008015000.flex
data_repo/curated/flex/tony/001001000_objects.flex
```

Open each in ImageJ with and without this PR, and note the obvious differences for ```qa-7649``` and ```tony``` especially.  With this PR, ```showinf -trace``` on each file and series should log the scaling factor and offset to the first strip for each plane.  The logged offsets should match what ```tiffdump``` reports for the ```StripOffsets``` tags, and the logged factor should match the plane's factor in the original metadata table.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FlexReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FlexReader.java
@@ -1307,10 +1307,16 @@ public class FlexReader extends FormatReader {
             compressed =
               firstIFD.getCompression() != TiffCompression.UNCOMPRESSED;
 
-            if (compressed || firstIFD.getStripOffsets()[0] == 16) {
+            if (compressed || firstIFD.getStripOffsets()[0] == 16 ||
+              firstIFD.getStripOffsets().length == 1)
+            {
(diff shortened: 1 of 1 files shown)


dgault:
Looking at a sample of the recent flex repo jobs compared to a set without the PR included shows a slight increase in the run time of the job but it is still well within an acceptable range.

Without the PR testing `qa-7469/002002.flex` and `tony/001001000_objects.flex` both displayed noticeable corruption on the first plane. With the PR included this corruption was no longer present. Using `showinf -trace` on each of the suggested files shows offsets which match that from the `StripOffsets` tags.

Overall the testing looks good and the PR is ready to merge.

PR #3086, opened 2018-02-27
Cellomics: fix indexing when the field counts are variable

melissalinkert:
Backported from a private PR.

This fixes file grouping and well/field indexing when the field count is not consistent for every well.

To test, use the example plate in ```data_repo/curated/cellomics/samples/different-field-count```, which was constructed by copying/renaming ```data_repo/curated/cellomics/omar/FB03f00d0.C01``` as described in the readme.

Without this PR, ```showinf -nopix -omexml plate_B03f00d0.C01``` should have a single series representing one well and field.  The used file list should contain only ```plate_B03f00d0.C01```.  With this PR, the same test should detect all 3 files in the plate, and represent them as 3 series (1 for well B03 and 2 for well C03).

One additional configuration change was necessary to change the well name from A00 to A01 for single file plates where the plate name cannot be detected by the regex; only the files in ```data_repo/curated/cellomics/omar``` are impacted.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
@@ -210,10 +210,6 @@ public class CellomicsReader extends FormatReader {
     wellRows = uniqueRows.size();
     wellColumns = uniqueCols.size();
 
-    if (fields * wellRows * wellColumns > files.length) {
-      files = new String[] {id};
-    }
-
     Arrays.sort(files, new Comparator<String>() {
         @Override
         public int compare(String f1, String f2) {
@@ -250,7 +246,6 @@ public class CellomicsReader extends FormatReader {
 
         }
     });
-
 
     core.clear();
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
This may also be enough to fix https://trac.openmicroscopy.org/ome/ticket/13230

sbesson:
Looking at the local branch I had started to investigate ticket 13230, I converged on the same fix as the one here so agreeing with the statement in https://github.com/openmicroscopy/bioformats/pull/3086#issuecomment-369630887. My primary concern was to assess the impact on existing data i.e. would this force all Cellomics plates to be reimported in an OMERO server

mtbc:
> Without this PR, showinf -nopix -omexml plate_B03f00d0.C01 should have a single series representing one well and field. The used file list should contain only plate_B03f00d0.C01. With this PR, the same test should detect all 3 files in the plate, and represent them as 3 series (1 for well B03 and 2 for well C03).

This accurately matches what I see in OMERO for the plate imported without a patched BF in OMERO and with.

mtbc:
If I import the plate into OMERO without this PR and then view it from one with this PR then it retains its previous behavior, probably as expected.

dgault:
Without this PR opening `plate_B03f00d0.C01` results in a single series and only1 file in the used files list. 

With the PR opening `plate_B03f00d0.C01 `results in all 3 files being included and represented as 3 separate series with the names `Well B03, Field #00`, `Well C03, Field #00` and `Well C03, Field #01`.

Overall this PR looks good and both this and the associated config PR are ready to merge.

PR #3087, opened 2018-02-27
ImageJ exporter: stop plugin when unsupported pixel type is found

melissalinkert:
See https://trello.com/c/FDmPiqW5/225-imagej-plugin-too-many-dialog-boxes-when-attempting-to-export-images-with-an-unsupported-pixel-type

To test, use the macro code on the Trello card.  Without this PR, an initial unsupported pixel type error message is shown, followed by one message per plane in the ```ImagePlus``` being exported (as described in the card).  With this PR, a single unsupported pixel type error is shown and then the exporter plugin finishes immediately.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -609,6 +609,7 @@ public class Exporter {
             if (notSupportedType) {
                 IJ.error("Pixel type (" + FormatTools.getPixelTypeString(thisType) +
                         ") not supported by this format.");
+                return;
             }
 
             if (codecs != null && codecs.length > 1) {


jburel:
Maybe I am missing something
Without this PR, the pixels type message is shown but only once
With this PR, no message.

melissalinkert:
Sorry, that's my mistake.  On re-testing without this PR, the problematic behavior only occurs when manually opening the image and exporting.

Non-macro testing instructions then are:

1. select ```File > Open Samples > T1 Head (2.4M, 16-bits)``` and wait for images to appear
2. select ```Plugins > Bio-Formats > Bio-Formats Exporter```
3. set ```Files of Type``` to ```Audio Video Interleave (.avi)``` in the file chooser, and save with any name to any writable directory
4. uncheck all boxes in the ```Bio-Formats Exporter - Multiple Files``` window and click OK
 
Without this PR, the above steps should result in 130 error messages.  With this PR, there should only be one error message.  Running a macro that performs the same steps:

```
run("T1 Head (2.4M, 16-bits)");
run("Bio-Formats Exporter", "save=[/replace/with/valid/export/path.avi]");
```

should also result in one error message with this PR (I still see it after re-testing with this PR).

jburel:
Re-tested both macro and non-macro, in both cases only one dialog pops up
Without the PR, for non-macro test, the dialog pops up for every frame.
It works as expected

PR #3091, opened 2018-03-01
TIFF tile index fix

rleigh-codelibre:
[Trello](https://trello.com/c/jzajo8mX/226-tile-writing-logic-bug)

This change removes what appears to be an unused parameter.  `TiffWriter` has a `prepareToWriteImage` method which has a [plane] `no` parameter.  However, when writing a tiled image, this was being called with the `tileNo` [the tile number on a plane; actually the tile number within the area being written].  This inconsistent calling behaviour meant that `prepareToWriteImage` could not tell whether a plane number or a tile number had been passed.  Looking at the usage of `no` in this function, the `initialized[][]` array indexing looks broken when called with a tile number, because it's sized by the plane count for each series, and it looks like it will potentially misbehave when writing tiles--there's some logic to prevent an index out of bounds error, but it's misusing the array despite that.  `no` is also used for computing the return value.  The return value looks like it's unused; it's checked for equality to -1, but since I can't see any way that a negative value could ever be returned, this looks completely pointless as well.  Edit: since the returned number is the IFD index to pass to `TiffSaver.writeImage`, using a tile index is actively dangerous since it could potentially access the wrong IFD.

/cc @dgault

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -243,7 +243,7 @@ public class TiffWriter extends FormatWriter {
           synchronized (this) {
             // This operation is synchronized against the TIFF saver.
             synchronized (tiffSaver) {
-              index = prepareToWriteImage(tileNo++, tileBuf, ifd, tileParams.x, tileParams.y, tileParams.width, tileParams.height);
+              index = prepareToWriteImage(no, tileNo++, tileBuf, ifd, tileParams.x, tileParams.y, tileParams.width, tileParams.height);
               if (index == -1) {
                 return;
               }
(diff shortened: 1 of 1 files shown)


dgault:

Agreed that the use of `initialized[][] ` is incorrect with the tile no and the check was producing the expected result. The updated behaviour should be correct, checking against the plane value.

I have carried out a range of manual testing using a small TiffWriting program across a range of tiled and untiled images as well as converting existing files from our public samples to Tiff. Additionally the previous performance test suites and benchmarks have all been run with these changes. Also I have been 'dogfooding' with these changes for a couple weeks without any issues.

The only area that I have run into problems with is 2 channel tiled writing, although that is an issue which is addressed by a seperate PR in https://github.com/openmicroscopy/bioformats/pull/3057

Overall this change looks to be correct and behaving as expected. PR is good to merge.

PR #3089, opened 2018-03-01
MicroManager: use PositionName key/value for storing the ImageName

sbesson:
See https://trello.com/c/5KT3Bvie/1-idr0040-aymoz-singlecell

Currently the Micro-Manager reader uses one of the TIFF filename for populating the ImageName metadata. For typical replica experiments across multiple positions (Pos0, Pos1,...), this results in duplicate image names. This commit now parses the PositionName key/value in the metadata JSON and use it if available to fill the ImageName.

This PR will probably affect the metadata of existing filesets, require a configuration change in which case it might  be more suitable for a minor release increment.

/cc @dominikl

PR #3092, opened 2018-03-01
MicroManager: use PositionName key/value for storing the ImageName (rebased onto metadata54)

sbesson:
This is the same as gh-3089 but rebased onto metadata54.

----

See https://trello.com/c/5KT3Bvie/1-idr0040-aymoz-singlecell

Currently the Micro-Manager reader uses one of the TIFF filename for populating the ImageName metadata. For typical replica experiments across multiple positions (Pos0, Pos1,...), this results in duplicate image names. This commit now parses the PositionName key/value in the metadata JSON and use it if available to fill the ImageName.

This PR will probably affect the metadata of existing filesets, require a configuration change in which case it might  be more suitable for a minor release increment.

/cc @dominikl

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -317,9 +317,13 @@ public class MicromanagerReader extends FormatReader {
         }
       }
 
-      if (positions.size() > 1) {
-        Location parent = new Location(p.metadataFile).getParentFile();
-        store.setImageName(parent.getName(), i);
+      if (p.name != null) {
+        store.setImageName(p.name, i);
+      } else {
+        if (positions.size() > 1) {
+          Location parent = new Location(p.metadataFile).getParentFile();
+          store.setImageName(parent.getName(), i);
+        }
       }
 
       if (getMetadataOptions().getMetadataLevel() != MetadataLevel.MINIMUM) {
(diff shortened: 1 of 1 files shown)


sbesson:
--rebased-from #3089

sbesson:
With this PR included, the updated image names for the Micro-Manager filesets was not preserved when importing the images into OMERO  - see https://trello.com/c/5KT3Bvie/1-idr0040-aymoz-singlecell#comment-5a99443815517e95b51ccdc8. Here, the problem is related in the OMERO import process which overrides the name of the image to use either an externally specified value or a default value based on the filename of `getCurrentFile()`. This behavior and its problems had been already raised and described in greater details in https://github.com/openmicroscopy/design/issues/57.

In OMERO 5.3.0, part of this metadata loss has been fixed in the case of High-Content Screening datasets where the plate name is preserved and used. For non SPW data, the `ImageName` populated by Bio-Formats remains lost on import. To modify the behavior for Images, https://github.com/openmicroscopy/openmicroscopy/pull/4992/commits/94575d51d788872c15e6abac2a887d29814c47ef probably contains most of the required logic but the breaking impact requires discussion.

Immediately, populating the `name` column in bulk imports as was done in https://github.com/IDR/idr-metadata/pull/297/commits/8a5a7cdebda7cc4d1891ebfab6cdabe6d582492e is our best option. Closing this PR against `metadata54` but leaving the original PR opened against `develop` for consideration as the metadata improvement the Micro-Manager datasets remains a feature for the consumers of the Bio-Formats API.

PR #3081, opened 2018-02-22
Remove sphinx documentation

rleigh-codelibre:
Moved to https://github.com/ome/bio-formats-documentation

This is a test for the east-ci, using the decoupled repository.

PR #3093, opened 2018-03-01
Remove decoupled components

rleigh-codelibre:
Additional work on top of https://github.com/openmicroscopy/bioformats/pull/3081

Remove all components integrated into bio-formats-build, which leaves formats-(api|bsd|gpl).

Note: expected to fail travis/appveyor testing, because it requires an unreleased ome-turbojpeg jar.


rleigh-codelibre:
Green in https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge/31/

joshmoore:
Assuming that we want to get out some set of these repositories (e.g. docs) before we release all of them, then likely we will need to revert part of this PR or close it while "phase 1" which is 5.8-compatible is handled.

rleigh-codelibre:
@joshmoore Would we not do 5.8-compatible changes directly on develop?  Isn't `east` completely separate?

joshmoore:
Ultimately, but I was thinking largely to get `latest` green. i.e. I can't merge this and then expect to have an `east`-based full repo build make use of the test-suite.

rleigh-codelibre:
@joshmoore test-suite removal reverted temporarily.

PR #3097, opened 2018-03-06
AVI: fix padding for RGB images

melissalinkert:
d5fce6a backported from a private PR; 1feb70c added to fix issues with the last plane of the new test file.

To test, use the new ```data_repo/curated/avi/sample-avis/rgb-odd-width/HyperStack.avi```.  As noted in the readme, this is generated by ImageJ and each plane should contain its own 4 digit 1-based index.  Without this PR, ```showinf HyperStack.avi``` should result in each plane being skewed.  With this PR, the same test should show all 4 planes correctly aligned.

The impact of d5fce6a can be assessed from existing tests, but as 1feb70c affects the last plane it would be a good idea to check the last plane of at least a few existing AVIs (RGB or otherwise).  Assuming tests pass, I would expect this to be safe for a patch release.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/AVIReader.java
+++ b/components/formats-bsd/src/loci/formats/in/AVIReader.java
@@ -195,7 +195,7 @@ public class AVIReader extends FormatReader {
     }
 
     long fileOff = offsets.get(no).longValue();
-    long end = no < offsets.size() - 1 ? offsets.get(no + 1) : in.length();
+    long end = no < offsets.size() - 1 ? offsets.get(no + 1) : fileOff + lengths.get(no);
     long maxBytes = end - fileOff;
     in.seek(fileOff);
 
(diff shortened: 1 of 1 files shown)


jburel:
Use ``data_repo/curated/avi/sample-avis/rgb-odd-width/HyperStack.avi``
With this PR the planes are correctly displayed
I have also look at avi under ``test_images_good`` and the last plane was correct
Note that it is not possible to open any ``.avi`` under ``data_repo/curated/avi/sample-avis/`` with or without this PR. Is that expected?

melissalinkert:
Yes, it is expected that the four files in ```data_repo/curated/avi/sample-avis``` will not open.  All have testing disabled (https://github.com/openmicroscopy/data_repo_config/blob/master/curated/avi/sample-avis/.bioformats) and use the unsupported Indeo 4.1 codec. https://trac.openmicroscopy.org/ome/ticket/4092 is the relevant ticket.

jburel:
@melissalinkert Thanks for the clarification
Everything works as expected

PR #3100, opened 2018-03-13
NIFTI: use the nDimensions field to read additional dimensions

melissalinkert:
See https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/dim.html for documentation of the dimension fields.

Fixes https://trac.openmicroscopy.org/ome/ticket/13291.  To test, use the file from QA 17334 (now in ```data_repo/curated/nifti/qa-17334```).  Without this PR, ```showinf -nopix``` should show SizeZ=43 and SizeT=SizeC=1.  With this PR, the same command should show SizeZ=43, SizeT=1, and SizeC=3.  This should match the dimensions described in the original bug report:

http://lists.openmicroscopy.org.uk/pipermail/ome-users/2016-September/006158.html

As the linked specification indicates, the fifth dimension is defined, but the sixth and seventh dimensions are undefined and uncommon.  This PR packs dimensions 5, 6, and 7 (if present) into SizeC for the moment, but we don't have any test data with 6 or 7 dimensions.  If/when we have examples of higher dimension data, it may make sense to use Modulo as needed.  Logging has been added to assist with detecting and debugging such data.

Tests should continue to pass, and this should be safe for a patch release (but is not especially urgent as the original bug report is 18 months old).

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -248,8 +248,22 @@ public class NiftiReader extends FormatReader {
     m.sizeY = in.readShort();
     m.sizeZ = in.readShort();
     m.sizeT = in.readShort();
+    m.sizeC = 1;
+    int[] extraDims = new int[3];
+    extraDims[0] = in.readShort();
+    extraDims[1] = in.readShort();
+    extraDims[2] = in.readShort();
+    if (nDimensions > 4) {
+      LOGGER.debug("nDimensions = {}", nDimensions);
 
-    in.skipBytes(20);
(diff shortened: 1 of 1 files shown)


dgault:
The code changes here look fine and the logic for handling rgb looks correct.

Tested using `data_repo/curated/nifti/qa-17334`. Without the PR shows a single channel image with SizeZ=43 and SizeT=SizeC=1.

With the PR included shows dimensions of SizeZ=43, SizeT=1, and SizeC=3. The images on all 3 channels display and look correct throughout the stack.

No other images from the data repo appear to be affected by the change and builds and tests are passing. The new configuration looks good and is also passing the repo tests. Merging for inclusion in 5.8.2.

PR #3101, opened 2018-03-14
TiffParser: Don't add SUBIFDS to main IFD list

rleigh-codelibre:
This is a test to determine which readers are reliant upon this behaviour, so that we can parse images with SUBIFDS without this behaviour.


rleigh-codelibre:
Failing formats:

- `canon` e.g. `kinor/2013_03_08_0808.CR2`
- `nikon` e.g. `samples/RAW_NIKON_D200_SRGB.NEF`

So any changes to `TiffParser` will need these readers updating to match.

Closing this PR.

joshmoore:
Even if our subclasses can be updated to deal with this change, migrating the logic to a method that can be overwritten (see 3102) or a config option (2899 etc) would likely be worth the investment in case this is a surprise for downstream consumers.

PR #3102, opened 2018-03-14
Have Memoizer.getMemoFile use an isWritableDirectory helper.

mtbc:
Introduces an overridable method that can be used to work around https://trello.com/c/q5L39XPZ/5-bioformatscache-is-expecting-read-write until Memoizer officially supports read-only access to its memos. Should not change Memoizer's behavior in any material way.

Currently open only for comment, **not for merging**.

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -755,7 +755,7 @@ public class Memoizer extends ReaderWrapper {
     return service;
   }
   protected Slf4JStopWatch stopWatch() {
-      return new Slf4JStopWatch(LOGGER, Slf4JStopWatch.DEBUG_LEVEL);
+    return new Slf4JStopWatch(LOGGER, Slf4JStopWatch.DEBUG_LEVEL);
   }
 
   /**
@@ -816,7 +816,7 @@ public class Memoizer extends ReaderWrapper {
    * @return if the given {@link File} is indeed a writable directory
    */
   protected boolean isWritableDirectory(File writeDirectory) {
-      return writeDirectory.canWrite() && writeDirectory.isDirectory();
+    return writeDirectory.canWrite() && writeDirectory.isDirectory();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


joshmoore:
Assuming you want to override this to return `False` won't this lead to `mkdirs()` being called?

mtbc:
If the memo already exists then I expect `mkdirs()` to return without doing anything.

components/formats-bsd/src/loci/formats/Memoizer.java
```diff
@@ -810,6 +810,15 @@ public File getMemoFile(String id) {
     return new File(p, "." + n + ".bfmemo");
   }
 
+  /**
+   * Test if the given {@link File} instance is a writable directory.
+   * @param writeDirectory a possible writable directory
+   * @return if the given {@link File} is indeed a writable directory
+   */
+  protected boolean isWritableDirectory(File writeDirectory) {
+      return writeDirectory.canWrite() && writeDirectory.isDirectory();
```
mtbc: ah, have to fix this indent Bio-Formats style!

mtbc:
Currently planning to have OMERO *copy* memo files from read-only to read-write before attempting to use to avoid also having to adjust `loadMemo()`, etc. regarding complicating issues such as the handling of invalid memo files and logging of such issues that are less significant in the presence of fallbacks.

joshmoore:
> copy memo files from read-only to read-write

You mean copy from a read-only filesystem to some temporary/internal read-write filesystem on the read-only server?

mtbc:
I do, yes: you correctly unpackaged my confusing phrasing!

mtbc:
The copying approach is working well in initial local testing of OMERO with this PR.

mtbc:
Force-pushed to remove some irrelevant recent history.

joshmoore:
Considering:

 * the red-ness of this PR
 * the red-ness of https://github.com/openmicroscopy/openmicroscopy/pull/5649
 * the fact that this branch is sub-1
 * the state of the east devspace

I'm merging and re-leasing this proactively. It's a simple enough process and then the jars will be available on artifactory for us to consume. If we need to roll this back, let's do that as a follow on version.

Thanks, @mtbc

mtbc:
Note that in its minimal approach this PR falls short of enabling testing for a *valid* memo file.

PR #2916, opened 2017-08-04
Fix file extension missing bug

bramalingam:
ref: https://trello.com/c/m7LqcpdM/167-czi-reader-fails-when-the-extension-is-missing

Testing instructions:
1) Using bftools please open an CZI file from /data_repo/curated/zeiss-czi/ (Should open without any issues)
2) please copy the file locally and remove the ".czi" extension and try opening the file again using bftools (Should open without any issues)

Without this PR, you should get the following error for the file without the ".czi" extension in bftools (5.5.3),
`Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1967)
	at loci.formats.in.ZeissCZIReader.initFile(ZeissCZIReader.java:583)
	at loci.formats.FormatReader.setId(FormatReader.java:1397)
	at loci.formats.ImageReader.setId(ImageReader.java:839)
	at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:650)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1031)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1117)`

As always code review please : 
@melissalinkert @sbesson

PR #3111, opened 2018-03-23
Zeiss CZI: fix handling of files with no extension

melissalinkert:
See https://trello.com/c/m7LqcpdM/167-czi-reader-fails-when-the-extension-is-missing and #2916.

As noted in the card and original PR, use an existing .czi file from ```data_repo/curated/zeiss-czi/``` to test.  I used ```data_repo/curated/zeiss-czi/aaron/101post.czi```, but you may wish to use something different.  Copy the file to a different directory and remove the extension. e.g.:

```
$ cp data_repo/curated/zeiss-czi/aaron/101post.czi ~/101post
```

Without this PR, ```showinf ~/101post``` should throw ```StringIndexOutOfBoundsException``` as noted in the original bug report: https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8336

With this PR, the same test should result in successful initialization and images being displayed.  Apart from file names, there should be no difference between ```showinf ~/101post``` and ```showinf data_repo/curated/zeiss-czi/aaron/101post.czi```.

Tests should continue to pass without any configuration change, and I would expect this to be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -583,7 +583,8 @@ public class ZeissCZIReader extends FormatReader {
     parser = XMLTools.createBuilder();
 
     // switch to the master file if this is part of a multi-file dataset
-    String base = id.substring(0, id.lastIndexOf("."));
+    int lastDot = id.lastIndexOf(".");
+    String base = lastDot < 0 ? id : id.substring(0, lastDot);
     if (base.endsWith(")") && isGroupFiles()) {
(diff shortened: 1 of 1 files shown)


dgault:
The code changes make sense and look good.

Without the PR renaming an existing sample file to remove the extension results in `StringIndexOutOfBoundsException` as described:
`java.lang.StringIndexOutOfBoundsException: String index out of range: -1`

With the PR included the extensionless file is able to be opened and displayed as normal.

Builds and tests have all been passing with this PR included. There is no update to configurations required and the PR is suitable for a patch release. Merging for inclusion in 5.8.2

PR #3112, opened 2018-03-26
Temporarily ignore links

hflynn:
Likely this is a temporary issue but will hopefully get the build green for now


hflynn:
CBIA software page currently has a grey box with no live links for the I3D library we're linking to but they've also moved their software page URL so likely an edit to the https://docs.openmicroscopy.org/bio-formats/5.8.1/users/i3dcore/index.html page will be needed. I've emailed someone to find out the status of the package though so I can hopefully fix everything at once or alternatively, remove the page if the tool is no longer available.

hflynn:
Added https://trello.com/c/lvf099f3/234-i3dcore-cbia-3d-library-status so this doesn't get lost

hflynn:
Closing for now so @rleigh-codelibre can test #3117 - will possibly just reopen against the new repo after that.

PR #3113, opened 2018-03-26
ImageReader: log an ERROR-level message when finding a reader for an empty file

melissalinkert:
This won't prevent initialization from being attempted, but alerts the user that the file may be corrupt before any calls to isThisType(...).  Fake files and files initialized via a specific format reader
(e.g. TiffDelegateReader) will not show this message.

See https://trello.com/c/AjjsuGr3/54-change-error-msg-for-empty-file

An exception could be thrown instead of logging at ```ERROR```, but I decided to start with the lower impact option, minimally so that this is sure to be safe for a patch release.

To test, use the steps noted in the card:

```
$ touch empty.tif
$ showinf -nopix -debug empty.tif
```

Without this PR, a ```FormatException``` is logged and ```MissingLibraryException``` is thrown without any immediate indication that the file is empty.  The output suggests that installing JAI will allow the file to be read.  With this PR, the same exceptions are logged and thrown, but there should be an ```ERROR```-level message indicating that the file is empty.

Fake files should be unaffected; with this PR ```touch empty.fake && showinf -nopix empty.fake``` should not display the empty file message.

### DIFF
--- a/components/formats-api/src/loci/formats/ImageReader.java
+++ b/components/formats-api/src/loci/formats/ImageReader.java
@@ -175,6 +175,9 @@ public class ImageReader implements IFormatReader {
    // e.g., for files, this will throw an exception if the file is missing.
    if (!fake && !omero) {
      Location.checkValidId(id);
+     if (new Location(id).length() == 0) {
+       LOGGER.error("File has length 0 and may be corrupt");
+     }
    }
 
     if (!id.equals(currentId)) {


dgault:
Logging an `ERROR` message in this use case at least provides some appropriate feedback to the user and is a suitable improvement for a patch release. In future an exception may be a longer term fix in this scenario.

Testing using an empty tif as suggested and without the PR I see a number of different exceptions thrown:

```
loci.formats.FormatException: Invalid TIFF file: empty.tif
	at loci.formats.in.MinimalTiffReader.initFile(MinimalTiffReader.java:441) ~[bioformats_package.jar:5.8.1]
...
java.lang.reflect.InvocationTargetException: null
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.7.0_80]
...
Caused by: java.io.EOFException: null
	at com.sun.media.jai.codec.SeekableStream.readUnsignedShort(SeekableStream.java:505) ~[jai_codec.jar:1.1.2]
...
loci.formats.FormatException: loci.common.ReflectException: Cannot execute method: getNumPages
	at loci.formats.in.TiffJAIReader.initFile(TiffJAIReader.java:149) ~[bioformats_package.jar:5.8.1]
```

With the PR included the new `ERROR` message is additionally logged. 
The behaviour of .fake files is unaffected by this PR and do not display the `ERROR` message.

PR #3114, opened 2018-03-27
JPEG: use DefaultJPEGReader for big images with no restart markers

melissalinkert:
```TileJPEGReader```/```JPEGTurboServiceImpl``` rely upon restart markers to read tiles without opening the whole image.  If restart markers aren't present, then ```DefaultJPEGReader```/ImageIO does a better job of decoding.

See https://trello.com/c/sRIQUQrr/206-jpg-issue-when-reading-full-slide-dec and https://trello.com/c/BzUB5Fuw/196-jpegreader-turbojpeg-failure-memory-usage-pixel-data

To test, use the two files in ```data_repo/curated/jpeg/human-embryology-2018-01-04``` and ```data_repo/curated/jpeg/darwinjob```.  Without this PR, ```showinf``` on each should show a mostly gray image as shown in one of the original bug reports: https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8393&p=18693&sid=098fb8ec2af6a4eff1dfc6562fa41fb1#p18693

With this PR, ```showinf``` should result in an image that matches other reference software (e.g. Mac OS Preview, Windows Picture Viewer).  Both images are large, so ```BF_MAX_MEM``` may need to be increased; I was testing with 4096m, but that may not be the lower bound.

Working with these files into OMERO may still be difficult even after this PR is merged, as the whole image needs to be decoded in order to read a tile of any size.  This may be another case where pre-generating a pyramid would be useful.

I don't think the addition of ```LOGGER``` to ```JPEGTurboServiceImpl``` will cause a problem for a patch release, but happy to rework or defer to 5.9.0 if any concerns.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/JPEGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/JPEGReader.java
@@ -139,9 +139,23 @@ public class JPEGReader extends DelegateReader {
     if (getSizeX() > MAX_SIZE && getSizeY() > MAX_SIZE &&
       !legacyReaderInitialized)
     {
+      // this is a large image, so try to open with TileJPEGReader first
+      // TileJPEGReader requires restart markers to be present; if this file
+      // doesn't contain restarts then TileJPEGReader will throw IOException
+      // and we'll try again with DefaultJPEGReader
       close();
       useLegacy = true;
-      super.setId(id);
+      try {
+        super.setId(id);
+      }
+      catch (IOException e) {
+        // this case usually requires a lot of memory as it's a big image
+        // that requires the whole image to be opened for any size tile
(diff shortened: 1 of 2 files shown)


dgault:
Tested using `data_repo/inbox/human-embryology-2018-01-04` and `data_repo/curated/jpeg/darwinjob`. Without this PR reading the images with showinf displayed a grey image as reported in https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8393&p=18693&sid=098fb8ec2af6a4eff1dfc6562fa41fb1#p18693.

With the PR included the images are read and display correctly. Automated tests with the associated configuration files are completing successfully and builds are green. Merging the PR for inclusion in 5.8.2

PR #3115, opened 2018-03-28
Have Memoizer.getMemoFile use an isWritableDirectory helper. (rebased from metadata54)

mtbc:
Introduces an overridable method that can be used to work around https://trello.com/c/q5L39XPZ/5-bioformatscache-is-expecting-read-write until Memoizer officially supports read-only access to its memos. Should not change Memoizer's behavior in any material way.

--rebased-from #3102

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -755,7 +755,7 @@ public class Memoizer extends ReaderWrapper {
     return service;
   }
   protected Slf4JStopWatch stopWatch() {
-      return new Slf4JStopWatch(LOGGER, Slf4JStopWatch.DEBUG_LEVEL);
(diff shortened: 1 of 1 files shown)


joshmoore:
Based on https://semver.org/#spec-item-1, I'd say that that depends on whether we treat, either globally or specifically for `Memoizer`, the protected methods as "public API". Don't mind either way, but due to the practical considerations you mention, I'd ask that we do one of the following two things:

 * (a) consider the protected methods *not* sufficiently part (currently) of the public API and include this in 5.8.x, *or*
 * (b) *do* consider them part of the public API, requiring a matching minor bump, but then we slot 5.9 for a quick release and moving large considerations to 5.10+.

sbesson:
:+1: to choosing between options (a) and (b) and having this included for the upcoming release. @dgault @melissalinkert any preference from your side?

PR #3118, opened 2018-03-29
Prevent NPE in Bio-Formats ImageJ macro extensions

sbesson:
See http://forum.imagej.net/t/macro-error-when-using-planetimings-txt-from-bioformats/10080/ for a description of the issue

## Changes

This commit adds a null check for the retrieved exposure time before trying to retrieve the value in seconds in the `Ext.getPlaneTimingExposureTime` macro extension like other `Time` and `Length` quantities retrieved by this API.

## Testing

Running the script provided in the [ImageJ forum post](http://forum.imagej.net/t/macro-error-when-using-planetimings-txt-from-bioformats/10080) above should fail with the current version of Bio-Formats (5.8.1) when using an image with `Plane` elements but no `Plane.ExposureTime` e.g. https://downloads.openmicroscopy.org/images/DICOM/samples/CR-MONO1-10-chest.dcm 
With this PR included, the macro should execute successfully.


jburel:
Tested with and without the PR
* NPE without the PR
* Fixed with the PR

PR #3119, opened 2018-04-03
Add extended subresolution support

rleigh-codelibre:
Intended for improved subresolution support in readers, for adding SUBIFDS support to TIFF and OME-TIFF readers.

- Add `MetadataList` and `CoreMetadataList` classes and unit tests
- Add `SubResolutionFormatReader` class as an alternative to `FormatReader`, using `CoreMetadataList`
- Deprecate core-index-related methods in `IFormatReader`, which are exposing internal implementation details of `FormatReader`
- Readers for sub-resolution formats and all TIFF formats use SubResolutionFormatReader
- Note the subclassing of `CoreMetadata` in `SVSReader` to store reader-specific subresolution-specific metadata directly in the `CoreMetadataList` container

`CoreMetadataList` replaces the `ArrayList<CoreMetadata>` in `FormatReader`, in a new `SubResolutionFormatReader` class (to avoid any API breaks in `FormatReader`).  The difference is that it's an array of arrays, representing series and resolution, rather than a flat list.  This makes lookups O(1) rather than O(n) due to avoiding the need to do a full traversal or series of hops over the list; it's now always two array lookups.  It implements the bits of `ArrayList` needed to be as compatible as possible with the previous implementation, so it's a drop-in replacement in many (but not all) aspects.  `SubResolutionFormatReader` is otherwise a complete duplication of the existing `FormatReader`.

`MetadataList` is a generalisation of `CoreMetadataList` to be used for storing arbitrary metadata by individual readers such as IFD mappings, or whatever, for series and subresolutions.  Currently used as the underlying implementation of `CoreMetadataList`, but will be used by the TIFF and other readers in the future.  This could be used to simplify some of the existing readers providing subresolutions, as well as to provide new capabilities, such as SUBIFDS support in the TIFF readers.  If you don't subclass `CoreMetadata`, this allows separate containers for different types of reader-specific metadata.

Testing:

- Ensure all builds are green and all full repository tests are green (reader regressions are possible)
- Pay special attention to these formats (check that the series and resolutions are in the same order with both flat and nonflat viewing modes, and that the pixel data is OK):
  - AFI (SVS-derived)
  - Imaris HDF
  - Leica-SCN
  - OMERO Pyramid TIFF
  - SVS

(the full repo test should verify all this, but it wouldn't hurt to check over a random sampling).

PR #3122, opened 2018-04-05
test-suite: Add testng.allow-missing property

rleigh-codelibre:
This allows unconfigured files to be skipped, which previously caused the build to fail; now it ignores these files and succeeds.

Sample output for cellomics from https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/6667/console (from https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/49/flowGraphTable/), which was previously failing:

```
[testng] [2018-04-05 22:28:34,319] [main] testng.allow-missing = true
...
[testng] [2018-04-05 22:28:53,117] [main] /data/samples/different-field-count/plate_B03f00d0.C01 not configured (skipping).
   [testng] [2018-04-05 22:28:53,118] [main] /data/samples/different-field-count/plate_C03f00d0.C01 not configured (skipping).
   [testng] [2018-04-05 22:28:53,118] [main] /data/samples/different-field-count/plate_C03f01d0.C01 not configured (skipping).
```

i.e. the unconfigured files are being deliberately ignored due to settings testng.allow-missing=true which can also be seen in the log.

Testing: check test-repo job is green.

### DIFF
--- a/components/test-suite/build.xml
+++ b/components/test-suite/build.xml
@@ -110,6 +110,7 @@ Type "ant -p" for a list of targets.
       <sysproperty key="testng.omexmlDirectory" value="${testng.omexmlDirectory}"/>
       <sysproperty key="testng.configDirectory" value="${testng.configDirectory}"/>
       <sysproperty key="testng.configSuffix" value="${testng.configSuffix}"/>
+      <sysproperty key="testng.allow-missing" value="${testng.allow-missing}"/>
       <sysproperty key="testng.multiplier" value="${testng.multiplier}"/>
       <sysproperty key="lurawave.license" value="${lurawave.license}"/>
       <sysproperty key="testng.in-memory" value="${testng.in-memory}"/>
@@ -139,6 +140,7 @@ Type "ant -p" for a list of targets.
       <sysproperty key="testng.directory" value="${testng.directory}"/>
       <sysproperty key="testng.omexmlDirectory" value="${testng.omexmlDirectory}"/>
       <sysproperty key="testng.configDirectory" value="${testng.configDirectory}"/>
+      <sysproperty key="testng.allow-missing" value="${testng.allow-missing}"/>
       <sysproperty key="testng.multiplier" value="${testng.multiplier}"/>
       <sysproperty key="lurawave.license" value="${lurawave.license}"/>
       <sysproperty key="testng.in-memory" value="${testng.in-memory}"/>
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
@sbesson With this change, we now have most tests passing, but there are still failures in:

- incell (invalid series)
- flex (NPE, pixels hashes, sizeC/Z, seriesCount)
- jpeg (pixels hashes)
- nifti (sizeC)

Looks like the data and/or configuration are not correct?  I don't think they are related to this PR?

rleigh-codelibre:
https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/49/flowGraphTable/

rleigh-codelibre:
@sbesson https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/7015/console is still failing in the same way, unfortunately.

rleigh-codelibre:
@sbesson The new run is looking much better, we're down to two failures now:

- incell https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/7015/console - invalid series
- tiff https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/7100/consoleFull - missing files
waiting on the final format to complete (cellsens).

sbesson:
Ah good progress indeed. So I think the summary is the following:

- the API works for excluding all the sample files passing `setId` but not configured for testing
- the two remaining failures corresponds to the case where the sample files are failing the initial `setId`

I think the two options to handle these remaining cases would be:

1.  test API option: make https://github.com/rleigh-codelibre/bioformats/blob/c5bd71d52ca8493e73858586eff834b7de3823f7/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java#L254 conditional to the existence of a configuration file if `testng.allow-missing` is true i.e. re-use the logic in https://github.com/rleigh-codelibre/bioformats/blob/c5bd71d52ca8493e73858586eff834b7de3823f7/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java#L284 during the initial detection phase.

2. configuration repo option: create a branch to exclude the sample files expected to fail `setId`

rleigh-codelibre:
@sbesson The latest commit should address this.  I'll re-run once it's green.

sbesson:
:+1: Main concern about aa75f85 is that with `testng.allow-missing=true`, any files failing `setId` would be unconditionally skipped. Concretely this could results in a lot of false negatives when a PR breaks file initialization. Once we get a green repository build from `east` with the above, we should work on refining the conditions for skipping tests before re-integrating #3119.

rleigh-codelibre:
@sbesson If the logic in the last commit makes sense for you, this should handle setId failure with configuration present vs not present.

rleigh-codelibre:
Green in https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/53/ and 54

rleigh-codelibre:
@sbesson All green for the last three builds; could we merge this please?

joshmoore:
Wow. Thanks, guys. I should disappear for a week more often. Merging into east to unlock gh-3119. We may want to (briefly) re-review with everyone when it comes time for re-integration. @rleigh-codelibre, I assume you have any related changes for https://github.com/ome/bio-formats-test in hand.

PR #3129, opened 2018-04-23
FileStitcher: do not use detected patterns that include missing files

melissalinkert:
See https://trello.com/c/hTxcpCtl/37-full-repository-skipped-filesets

When detecting a pattern based upon a list of file names, ignore any patterns that include file names not in the list.  To test, use the TIFF files in ```data_repo/curated/cellr/kees/7Z5GR81B_DocumentFiles```.  Without this change, ```showinf -nopix -stitch``` on any of them should throw a ```FileNotFoundException```.  With this change, there should be no exception and a single file should be on the used files list.

Unit tests should continue to pass with this change, but there is a chance that this will impact https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-repo/ and require additional fixes.

### DIFF
--- a/components/formats-bsd/src/loci/formats/FileStitcher.java
+++ b/components/formats-bsd/src/loci/formats/FileStitcher.java
@@ -33,6 +33,7 @@
 package loci.formats;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


sbesson:
With this change included, the repository tests are still green so no obvious regression. While more robustness in the detection of the file stitching is certainly useful, I still struggle to understand what causes a wrong pattern to be detected and issue a `FileNotFoundException` in the first place. 

With

```
[sbesson@ome-c6100-1 ~]$ ls test1
test_T02.tif  test_T03.tif  test_T04.tif  test_T06.tif  test_T07.tif
```

executing `showinf -nopix -stitch` on any input file with Bio-Formats 5.8.2 returns a single file pattern as expected.

melissalinkert:
The files need to have variation in more than one digit, as this causes multiple numerical blocks to be detected.  The file names shouldn't have consistent values in the second digit for all values of the first digit.

As a minimal example, with these names:

```
$ ls
t03.tif  t04.tif  t12.tif  t13.tif
```

the patterns previously detected are ```t<0-1><3-4>.tif``` and ```t<0-1><2-3>.tif```, which should result in a ```FileNotFoundException``` with 5.8.2.  Renaming ```t12.tif``` to ```t14.tif``` would result in ```t<0-1><3-4>.tif```, which should pick up all 4 files without throwing an exception (with or without this PR).

dgault:
A message on the users list today (http://lists.openmicroscopy.org.uk/pipermail/ome-users/2018-June/007075.html) was resulting in the FileNotFoundException when using patterns. I checked to see if the FileStitcher changes here would have any impact in this particular scenario but it still produced the same issue. In this particular case though that is probably the expected result.

dgault:
i do not see this having an negative consequences and fixes the FileNotFoundException with a sensible fallback behaviour. Merging for 5.9.0

PR #3130, opened 2018-04-24
switch modifier order to "static final" in readers

mtbc:
Matches Oracle's recommendations and convention. Should not change behavior at all.


rleigh-codelibre:
Builds are green, and looks fine.  Please merge.

PR #3131, opened 2018-04-24
Sync 5.8.2 changes with east

rleigh-codelibre:
Testing: Check builds are green.  All the changes were reviewed for 5.8.2, so there's not much to check here other than basic sanity.


sbesson:
Successfully tested in https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/ from what I can see, is there anything else we want to consider here before merging?

rleigh-codelibre:
Nothing I can think of, no.  If the test repo job is green, the test config is good, and that was the only thing I think.

PR #3132, opened 2018-04-25
Zeiss CZI: fix big image edge case when tiling is present but a pyramid is not

melissalinkert:
See https://trello.com/c/3AHJeUPc/236-czi-compressed-pyramidal-error

To test, use the file from QA 20612.  Without this PR, opening the file using ImageJ or ```showinf``` will result in 14 series, most of which look corrupted (as noted on the card).

With this PR, ImageJ or ```showinf``` should report 3 series - one full resolution image, one label, and one macro/overview.  There is no stored pyramid, so there should be no subresolutions detected.  The image dimensions and pixel data should match what is shown in ZEN.

Builds should remain green with this change.  This should be safe for a patch release, so is low priority for 5.9.0 as it could be deferred to 5.9.1.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -890,7 +890,7 @@ public class ZeissCZIReader extends FormatReader {
         plane.pixelTypeIndex = pixelTypes.indexOf(plane.directoryEntry.pixelType);
       }
 
-      if (seriesCount * pixelTypes.size() > 1) {
+      if (core.size() * pixelTypes.size() > 1) {
         core.clear();
         for (int j=0; j<pixelTypes.size(); j++) {
           for (int i=0; i<seriesCount; i++) {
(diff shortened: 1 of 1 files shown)


dgault:
Builds and tests are green with this PR included and the associated config PR. 

Without this PR QA-20612 opens with 14 series and a number of the images appear to be corrupted. This behaviour is the same with or without the autostitching option.

With the PR included and autostitching switched off I see the same behaviour as in the original tests. When autostitching is enabled (which is the default) QA-20612 opens with 3 series, the image, label and macro image. All 3 open and display correctly, the metadata looks correct and there are no subresolutions.

The test scenario with `zeissczi.autostitch` is working correctly. @melissalinkert is it expected that this will still fail without the autostitch option?

melissalinkert:
The tiles that are stitched together are not of uniform size, so it doesn't surprise me that turning off ```zeissczi.autostitch``` still results in incorrect tiles.  Non-uniform tiles aren't uncommon when the tiles represent pieces of a slide instead of stage positions (as is the case for QA 20612); QA 17740 has the same problem for sure, and there may be others.

I can see both sides of supporting/fixing this case - we do allow ```zeissczi.autostitch``` to be disabled, so it seems broken from a user perspective, but it's also kind of the wrong way to work with slide data (other WSI readers don't expose tiles as stored in the file).  If we do want to support it, I think it merits a separate PR and a better way to test for regressions with ```zeissczi.autostitch``` disabled.

dgault:
Agreed, this PR definitely provides a fix for the mode in which most users should be using the data and as such is good for inclusion with 5.9

PR #3138, opened 2018-05-02
Maven updates

rleigh-codelibre:
This PR cherry-picks the `east` maven update commits for use on `develop`.

Testing: check jobs remain green.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,9 +7,6 @@ environment:
   matrix:
     - java: 9
       build: maven
-    - java: 9
-      build: ant
-      ant_version: 1.10.1
     - java: 1.8
       build: maven
     - java: 1.8
@@ -17,9 +14,6 @@ environment:
       ant_version: 1.10.1
     - java: 1.7
       build: maven
-    - java: 1.7
-      build: ant
-      ant_version: 1.9.9
 
 cache:
   - '%AV_BF_M2% -> appveyor.yml'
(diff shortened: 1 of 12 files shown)


.appveyor.yml
```diff
@@ -5,6 +5,11 @@ environment:
 
 # Note that only Oracle JDK is provided.
   matrix:
+    - java: 9
```
sbesson: Main concerns about adding the Appveyor  matrix is: are we going to reach the limits of our usage and start to block PR checks once merged into`develop`? Do we have metrics about usage, wait & completion times?

rleigh-codelibre: I have dropped two ant builds from both travis and appveyor so that it only builds with the primary supported jdk (8).

joshmoore:
I'm unclear what the goal of opening this against develop at the moment is. My understanding from our discussion yesterday was that we were going to focus on getting `east` into a state that it was ready for merging into `develop`. Closing for the moment.

rleigh-codelibre:
This *is* getting it ready for merging, by collecting together the maven parts of east and getting them onto develop.  This reduces `east` to the subresolution-specific changes only so that it's easy to review the delta between them.  That is to say, it is taking the non-subresolution-related changes out of the picture.

joshmoore:
Appreciate the effort, but as we discussed, I don't think this is how we see the reintegration working. If we decide to go this route, we can always reopen. For the moment, let's assume that the review will take place via a single larger merge, rather than a rebase strategy.

PR #3139, opened 2018-05-02
Remove autogen pom

rleigh-codelibre:
This one slipped through the cracks when merging the 5.8.2 changes into `east`.


dgault:
Builds are green, pom is good to remove

PR #3140, opened 2018-05-02
Support for unconfigured tests

rleigh-codelibre:
This picks the test-suite changes from `east` onto `develop`.

Testing: Check full-repo test continues to pass.  Since it doesn't use the new `testng.allow-missing` option (and there are no missing files), there should be zero change in behaviour.


rleigh-codelibre:
@sbesson We'll actually need this.  Not all of the test data files are readable on develop (by the testsuite) due to being too large (`CMU*`) or causing faults (`Leica-Fluorescence`).  In consequence, we will need to ignore these.

PR #3141, opened 2018-05-02
OBF: add basic FLIM support

melissalinkert:
See https://trac.openmicroscopy.org/ome/ticket/13196.

To test, use the file from QA 17111.  Without this PR, ```showinf -nopix``` should report 43 series, each with ```SizeX``` and ```SizeZ``` similar but ```SizeY``` noticeably different.  With this PR, the same command should show 43 series, each of which has ```SizeX``` equivalent to ```SizeY```.  This should be consistent with the ticket description.  It is probably also worth checking the 256x256 series in ImageJ to verify that the images look plausible for FLIM data.

Builds should continue to pass without any configuration changes for existing datasets.  This might be safe for a patch release, but is definitely low priority either way.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -156,6 +156,7 @@ public class OBFReader extends FormatReader {
       for (String key = readString(); key.length() > 0; key = readString()) {
         if (key.equals("ome_xml")) {
           final String ome_xml = readString();
+          LOGGER.trace("OME-xml = {}", ome_xml);
 
           try {
             ServiceFactory factory = new ServiceFactory();
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -221,7 +221,7 @@ public class OBFReader extends FormatReader {
         CoreMetadata meta_data = core.get(image);
 
         final String name = meta_data.seriesMetadata.get("Name").toString();
-        metadataStore.setImageName(name, image);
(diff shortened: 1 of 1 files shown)


dgault:

Tested with the file in `QA-17111`. Without this PR the file shows 43 series, many of which have odd heights such 256x1. The series do not appear to display correctly either.

With this PR included the series have equal X and Y values and opening each series shows that the images are displaying properly.

dgault:
Hi @bjoernthiel, thank you for helping to identify the issue. Using the sample file you attached I was able to reproduce the problem and it appears the problem is due to the dimensions being incorrectly set at https://github.com/ome/bioformats/pull/3141/files#diff-5720da964f9ddb1ef5d8d17f099f4d2fR457

This should be something which we can fix for the next patch release. In the sample file provided the problem occurred when the dimensions for the second image are read as `[259, 40, 250, 1, 1]` and the labels as `[ExpControl X, ExpControl Z, ExpControl Y, None, None]`. As such the code attempted to set the dimensions as below with Z and Y inverted:
X: 259
Z: 40
Y: 250

Can you confirm that the labels being parsed are correct and in the right order?

bjoernthiel:
Yes, the labels are [ExpControl X, ExpControl Z, ExpControl Y, ...] and the dimensions are [259, 40, 250, ...].

bjoernthiel:
@dgault @melissalinkert 
Is there something else I can help with?

dgault:
@bjoernthiel, I will try and get a PR opened in the next few days with the aim of having a fix in place for the next patch release 6.3.1

The sample file provided should be sufficient for testing purposes but if there are any additional files which can be used for testing that would also be useful.

PR #3142, opened 2018-05-03
build: Never run upgrade checker when running unit tests

rleigh-codelibre:
This results in unnecessary hits and biased statistics.  Add `bioformats_can_do_upgrade_check=false` to all test invocations.

The main culprits are the bio-formats-tools unit tests.  However, this PR disables them comprehensively to avoid any potential for future mis-reporting.

Testing: check CI builds are green.


joshmoore:
Several days of passing builds, and no issues voiced recently. General :thumbsup:

PR #3144, opened 2018-05-03
ci: Only test ant build with JDK8

rleigh-codelibre:
The last commit from https://github.com/openmicroscopy/bioformats/pull/3138 but for `east`.  This will speed up the CI wait times, particularly for appveyor by 15-20 minutes per run.

The test matrix size for travis and appveyor will reduce from 6 to 4.

Testing: check the builds are green and that the matrix size has decreased.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,9 +7,6 @@ environment:
   matrix:
     - java: 9
       build: maven
-    - java: 9
-      build: ant
-      ant_version: 1.10.1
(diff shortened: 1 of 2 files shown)


joshmoore:
Both builds are green and match "build maven on {jdk7, jdk8, jdk9} and build ant on jdk8" :+1:

screen shot 2018-05-03 at 13 49 00
screen shot 2018-05-03 at 13 48 51

PR #3146, opened 2018-05-07
test-suite: Fixes from east

rleigh-codelibre:
A few selected fixed from https://github.com/openmicroscopy/bioformats/pull/3145 so that the large images are tested the same as on `east`.


rleigh-codelibre:
These aren't needed to pass the tests with the develop-specific test configuration, so closing.

PR #3150, opened 2018-05-15
Adding new KLBReader

dgault:
This reader is for the Keller Lab Block file type and is based on https://bitbucket.org/fernandoamat/keller-lab-block-filetype/overview

Details of the image header and image IO can be found at https://bitbucket.org/fernandoamat/keller-lab-block-filetype/src/8e3c39b75022/src/?at=master

Testing is in progress, further testing required on the following:
- Multi channel data
- Data with multiple  in all 5 dimensions
- ROI's

Future Work:
- Read associated XML and add as an Annotation

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/KLBReader.java
+++ b/components/formats-gpl/src/loci/formats/in/KLBReader.java
@@ -1,0 +1,346 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+ * %%
+ * Copyright (C) 2005 - 2018 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the 
+ * License, or (at your option) any later version.
+ * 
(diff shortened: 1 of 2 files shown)


components/formats-gpl/src/loci/formats/in/KLBReader.java
```diff
@@ -0,0 +1,346 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+ * %%
+ * Copyright (C) 2005 - 2018 Open Microscopy Environment:
```
jburel: no need for range

components/formats-gpl/src/loci/formats/in/KLBReader.java
```diff
@@ -0,0 +1,346 @@
...
+ * Reader for Keller Lab Block (KLB) files.
+ */
+public class KLBReader extends FormatReader {
+
+  // -- Constants --
+  private static final int KLB_DATA_DIMS = 5; //images have at most 5 dimensions: x,y,z, c, t
+  private static final int KLB_METADATA_SIZE = 256; //number of bytes in metadata
+  private static final int KLB_DEFAULT_HEADER_VERSION = 2;
+
+  private static final int UINT8_TYPE = 0;
```
mtbc: Seeing as we're introducing new code here would it be okay to use enums for this kind of thing?

mtbc:
Be sure to avoid https://trello.com/c/lJtRpXGZ/515-bug-used-files-getters. :smiley:

PR #3153, opened 2018-05-28
TiffWriter: Add support for DEFLATE (zlib) compression

rleigh-codelibre:
See https://trello.com/c/O5LkXtx0/29-imagewriter-compression-behaviour

Testing: Run `bfconvert` with and without the `-compression zlib` option, then run `tiffinfo` on both files.  You will see `Compression Scheme: AdobeDeflate` with zlib compression, and the file size should be reduced.


dgault:
Tested using `bfconvert`, converting the same file both with and without compression produces 2 images both of which can be read by Bio-Formats and display correctly. The file size of the compressed file is considerably smaller. Using tiffinfo on the compressed file shows `Compression Scheme: AdobeDeflate` while the uncompressed file shows none `Compression Scheme: None` as expected.

The new compression option appears to be working as expected and is good to merge.

PR #3154, opened 2018-05-28
ImageWriter: Get compression types from the writer in use

rleigh-codelibre:
See https://trello.com/c/O5LkXtx0/29-imagewriter-compression-behaviour

Testing: Try `bfconvert -compression zlib` with and without both this PR and https://github.com/openmicroscopy/bioformats/pull/3153

- without both PRs, zlib compression will be allowed, but won't be used
- with https://github.com/openmicroscopy/bioformats/pull/3153 zlib compression will be allowed and will be used
- with this PR but without https://github.com/openmicroscopy/bioformats/pull/3153 zlib compression will be disallowed (tests it fails correctly if a codec is not supported by the chosen writer, but is supported by another writer and ends up in the combined codec list)

The combined codec list is the cause of the problem, and I'm unsure why it's required.  We don't have the same behaviour for any other features like e.g. pixel types.

### DIFF
--- a/components/formats-api/src/loci/formats/ImageWriter.java
+++ b/components/formats-api/src/loci/formats/ImageWriter.java
@@ -96,12 +96,6 @@ public class ImageWriter implements IFormatWriter {
    */
   private String[] suffixes;
 
-  /**
-   * Compression types for all file format writers.
-   * Populated the first time getCompressionTypes() is called.
-   */
-  protected String[] compressionTypes;
-
   /** Name of current file. */
   protected String currentId;
 
@@ -366,19 +360,7 @@ public class ImageWriter implements IFormatWriter {
   /* @see IFormatWriter#getCompressionTypes() */
   @Override
   public String[] getCompressionTypes() {
-    if (compressionTypes == null) {
-      HashSet<String> set = new HashSet<String>();
-      for (int i=0; i<writers.length; i++) {
-        String[] s = writers[i].getCompressionTypes();
-        if (s != null) {
-          for (int j=0; j<s.length; j++) set.add(s[j]);
-        }
-      }
-      compressionTypes = new String[set.size()];
-      set.toArray(compressionTypes);
-      Arrays.sort(compressionTypes);
-    }
-    return compressionTypes;
+    return getWriter().getCompressionTypes();
   }
 
   /* @see IFormatWriter#getPixelTypes() */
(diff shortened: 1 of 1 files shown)


sbesson:
Looking briefly at this, I agree that there is some inconsistency between the way different properties are returned. Some attributes (pixel type, tile sizes) use a delegation strategy to fetch the property from the underlying writer while others (suffixes, compression types) aggregate all properties.

My assumption is that these APIs were implemented in this way with client code use cases in mind i.e. in order to introspect the available extensions and compressions types when writing data. Obviously a caveat specific is that using flat lists does not fully represent the relationship between formats and compressions i.e. ideally such a method should return a map.

The proposed change here align the changes with the other APIs of `ImageWriter`. However, we can already see it will impact our own code as the MATLAB helper functions were using the implemented contract of `getCompressionTypes` as shown by the [test failures](https://ci.openmicroscopy.org/view/Failing/job/BIOFORMATs-DEV-merge-matlab/971/). I would also expect the ImageJ plugin to also be impact and more generally the impact of this change should be considered as breaking from a behavioral perspective for any downstream consumer code.

I see a few ways to discuss the next steps.
1. either consider the API contract is to return the compression types of the initialized writer (or of the first writer in the list if not initialized). Actions might be:
  - clarify the Javadoc and/or the technical documentation to express the contract of `getCompressionTypes()`
  - possibly create a `getAllCompressionTypes()` or `getCompressionTypes(boolean)` method if maintaining a high-level API returning the combined properties
  - decide on the behavior of other properties like `getSuffixes`
2. either consider the current implementation is the expected API i.e. `ImageWriter` is not a strict`WriterWrapper` and the API contract depends on the property. Actions might be:
 - clarify the Javadoc to differentiate the types where the delegate property vs a combined list of properties is returned
 - review our codebase  and update to use `ImageWriter.getWriter().getCompressionTypes()` instead of `ImageWriter.getCompressionTypes` when necessary
3. if we considered 1. is the correctlong-term solution but implementing it in a minor release is too breaking, we could try to maintain some level of  backwards compatibility in the interim by enabling the new behavior via a metadata option e.g. `imagewriter.compression.strategy={combined,delegate}` and update our internal code to consume the new behavior. The old behavior could then be dropped in an upcoming breaking release.

This will probably be worth a discussion at a Formats meeting:

PR #3156, opened 2018-05-31
Warn if passing metadata with dimensionOrder in bfsave

jarmond:
Silent dropping of the dimensionOrder parameter stumped me for some time while trying to combine files.

### DIFF
--- a/components/formats-gpl/matlab/bfsave.m
+++ b/components/formats-gpl/matlab/bfsave.m
@@ -5,7 +5,9 @@ function bfsave(varargin)
 %    specified by outputPath.
 %
 %    bfsave(I, outputPath, dimensionOrder) specifies the dimension order of
-%    the input matrix. Default valuse is XYZCT.
+%    the input matrix. This value will be ignored if an OME-XML metadata
+%    object is also passed to bfsave via the metadata key/value parameter.
+%    Default value is XYZCT.
 %
 %    bfsave(I, outputPath, 'Compression', compression) specifies the
 %    compression to use when writing the OME-TIFF file.
--- a/components/formats-gpl/matlab/bfsave.m
+++ b/components/formats-gpl/matlab/bfsave.m
@@ -73,6 +73,9 @@ if isempty(ip.Results.metadata)
         ip.Results.dimensionOrder);
 else
     metadata = ip.Results.metadata;
+    if ~ismember('dimensionOrder', ip.UsingDefaults)
+        warning('''dimensionOrders'' is ignored if passing ''metadata''');
+    end
 end
 
 % Create ImageWriter


components/formats-gpl/matlab/bfsave.m
```diff
@@ -73,6 +73,9 @@ function bfsave(varargin)
         ip.Results.dimensionOrder);
 else
     metadata = ip.Results.metadata;
+    if ~ismember('dimensionOrder', ip.UsingDefaults)
+        warning('''dimensionOrders'' is ignored if passing ''metadata''');
```
dgault: Rather than simply adding the warning perhaps we should instead set the value which is being ignored. Something similar to below:

`
% Set dimension order
dimensionOrderEnumHandler = javaObject('ome.xml.model.enums.handlers.DimensionOrderEnumHandler');
dimensionOrder = dimensionOrderEnumHandler.getEnumeration(ip.Results.dimensionOrder);
metadata.setPixelsDimensionOrder(dimensionOrder, 0);
`

sbesson: Might be worth a discussion about the expected contract of this API i.e. considering the following situations

```
>> I = zeros(200,200,10,3,3);
>> m = createMinimalOMEXMLMetadata(I, 'XYZCT');
>> bfsave(I, '/tmp/test.ome.tiff', 'XYCZT')
>> bfsave(I, '/tmp/test.ome.tiff', 'metadata', m)
>> bfsave(I, '/tmp/test.ome.tiff', 'XYCZT', 'metadata', m)
```

The expectations from the two first calls in terms of dimension order are obvious. For the third case, we minimally need to agree on the precedence order. Currently the `dimensionOrder` from `metadata` takes precedence and 2f7ef63 preserves this statu quo while adding some warning for the consumer. We might want to update the function help as well.
https://github.com/openmicroscopy/bioformats/pull/3156#discussion_r193109325 would propose to give precedence to the `dimensionOrder` parameter. My main worry would be the potential for unexpected behavior change for existing client code. 

Thoughts/suggestions welcome.

jarmond: I guess my thought is that `dimensionOrder` is there for the convenience of not having to create a `metadata`. But if you have gone to the trouble of creating a `metadata` then you ought to be willing to complete that `metadata` as necessary because most likely you want to record more than only `dimensionOrder`, prior to calling `bfsave`.

sbesson: @jarmond thanks for commenting and sorry for the long silence here. We have discussed with @dgault and concluded that not changing the (currently undocumented) `dimensionOrder` contract is probably the least breaking solution here and your comments support the fact it is not completely invalid from a consumer perspective.

For the upcoming release of 5.9.0, documenting this behavior and adding some warnings as you propose here is certainly a very sensible step forward. I will try and open a PR against the technical Sphinx documentation. 

In addition to the warning added here, would you be okay to push a commit updating the function help (https://github.com/openmicroscopy/bioformats/pull/3156/files#diff-3c484902b71dc82dcdba91e8603ddff7R7) to also mention that dimensionOrder will be ignored if metadata is passed?

sbesson:
Thanks @jarmond for helping us to clarify this behavior. Including for the upcoming release of Bio-Formats.

PR #3157, opened 2018-05-31
ND2: prevent integer overflow when reading a tile from a large image

melissalinkert:
Backported from a private PR.

The original use case was a 12589 x 48981 uint16 .nd2 file.  Reading tiles from approximately the bottom half of the image resulted in an integer overflow when calculating the number of bytes to skip before reading.  This changes all of the relevant ```skipBytes``` calls (which only takes an ```int``` ) to ```seek``` calls, and uses ```long``` multiplication as needed to prevent overflows.

Discussed with @sbesson yesterday, and agreed that a test file would not be necessary given the limited scope of this change (i.e. code review and passing builds are sufficient).  Also /cc @mtbc for review as suggested by @sbesson.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -294,12 +294,15 @@ public class NativeND2Reader extends FormatReader {
       int rowLength = getSizeX() * pixel + scanlinePad * bpp;
       int destLength = w * pixel;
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
```skipBytes(long)``` certainly makes sense to me as a long-term solution.

dgault:
The code change to use `seek` rather than `skipbytes` looks good and all builds and tests are green. Behaviour of the reader remains unchanged other than preventing the mentioned integer overflow errors.

PR #3160, opened 2018-06-07
Fix Metamorph RGB series channel count

jmuhlich:
In ND files referencing RGB images, sizeC was getting set to 1 rather than 3.
(The individual STK/TIFF files are properly handled). This change multiplies
the sizeC by 3 for RGB images, and replaces a later use of sizeC with
effectiveSizeC.

The specific failure I was encountering was that the pixel buffer would be too small in calls to openBytes, by a factor of 3. `showinf` for example crashes on such images. I uploaded an example failing image to the Glencoe ftp server as `jmuhlich_metamorph_rgb`. Note that I manually edited the .nd file down from an enormous 450-series slide scan to just a single series. With this patch, this image is handled correctly.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -679,6 +679,9 @@ public class MetamorphReader extends BaseTiffReader {
       ms0.sizeC = cc;
       ms0.sizeT = tc;
       ms0.imageCount = getSizeZ() * getSizeC() * getSizeT();
+      if (isRGB()) {
+        ms0.sizeC *= 3;
+      }
       ms0.dimensionOrder = "XYZCT";
 
       if (stks != null && stks.length > 1) {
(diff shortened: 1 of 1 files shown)


sbesson:
Thanks @jmuhlich, including this in the daily builds. We will get in touch with Glencoe to retrieve the sample files and add it to the QA repository for testing the changes. Out of curiosity, can the sample files be made [public](http://downloads.openmicroscopy.org/images/)?

jmuhlich:
I've asked the image creator if the image can be made public. I'm actually putting together a 4-tile version from the same dataset, as this was actually a `.scan` file not `.nd`, so it could serve as a test for both this PR and #3161 .

melissalinkert:
Dataset now moved from FTP server to ```data_repo/inbox/jmuhlich_metamorph_rgb/``` for testing.

jmuhlich:
By the way, is it possible to upload to OMERO.qa without having Flash installed? Or should I just keep using the Glencoe FTP server?

mtbc:
No non-Flash upload yet I'm afraid but it's certainly on our to-do list! Glencoe FTP remains the trusty fallback.

dgault:
The changes here make sense and the code looks good. 

Builds are green with this PR included and existing files in the repo are unaffected by this change: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/575

Testing the uploaded sample and without the PR results in an exception for incorrect buffer size:
`loci.formats.cache.CacheException: loci.formats.FormatException: Buffer too small (got 3207360, expected 9622080).`

With the PR included the sample file opens and displays correctly. I have opened a PR to add a config for the sample file https://github.com/openmicroscopy/data_repo_config/pull/300 which is also green in the tests: https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-repository-subset/576/consoleFull

PR is good to merge.

PR #3161, opened 2018-06-07
Add .scan extension to MetamorphReader

jmuhlich:
The .scan extension seems equivalent to .nd by visual inspection. This is
reinforced by this Molecular Devices help page:

http://mdc.custhelp.com/app/answers/detail/a_id/19368/~/supported-file-formats

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -96,6 +96,7 @@ public class MetamorphReader extends BaseTiffReader {
   public static final String LONG_DATE_FORMAT = "dd/MM/yyyy HH:mm:ss";
 
   public static final String[] ND_SUFFIX = {"nd", "scan"};
+  public static final String[] PARENT_INDEX_SUFFIX = {"nd", "scan", "htd"};
   public static final String[] STK_SUFFIX = {"stk", "tif", "tiff"};
 
   public static final Pattern WELL_COORDS = Pattern.compile(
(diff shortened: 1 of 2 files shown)


sbesson:
Thanks @jmuhlich for this contribution. Do you have some representative sample of the Metamorph  Scan Slide format that could be shared with us for inclusion in the QA repository?

jmuhlich:
Oops, I missed one thing -- adding `.scan` to the list of extensions checked in `isThisType`. Doing that now, also refactoring it a bit to reduce the redundancy.

jmuhlich:
Yes I have an example, see #3160.

melissalinkert:
Thanks, @jmuhlich.  Dataset now moved from FTP to ```data_repo/inbox/metamorph_scan_rgb/``` so @sbesson and @dgault can access.

components/formats-gpl/src/loci/formats/in/MetamorphReader.java
```diff
@@ -190,17 +191,13 @@ public boolean isThisType(String name, boolean open) {
...
-          (new Location(parent, baseName + ".nd").exists() ||
-          new Location(parent, baseName + ".ND").exists()))
-        {
-          return true;
-        }
-        if (checkSuffix(name, suffixes) &&
-          (new Location(parent, baseName + ".htd").exists() ||
-          new Location(parent, baseName + ".HTD").exists()))
-        {
-          return false;
```
dgault: Previously the behaviour in `isThisType` was to return `false` for the suffix htd, but now HTD has been added to `PARENT_INDEX_SUFFIX` it is returning `true` which is allowing it to identify with files which should be claimed by `CellWorxReader`.

jmuhlich: Oops, I just misread that as `return true` in the original code and overzealously refactored it. Fixed in latest commit.

dgault:
All of the builds and tests are now passing with the sample data configured and included (https://github.com/openmicroscopy/data_repo_config/pull/327)

With the latest commits the code looks good and manual testing with the sample image provided results in the image opening and displaying as expected. All of the metadata also appears to be sensible and correcr. The changes also have no impact on existing datasets. PR is ready to merge.

PR #3163, opened 2018-06-11
GelReader: Don't allow overflow when reading unsigned integer values

rleigh-codelibre:
[Trello](https://trello.com/c/WeajJ80u/242-gel-incorrect-pixel-data)

The use of DataTools.bytesToShort was causing values to wrap around when reading unsigned values which were not representable in a signed short.  Use bytesToInt to avoid this problem.

Testing: Try reading the GEL file from https://www.openmicroscopy.org/qa2/qa/feedback/21519/ with and without this patch.  It should display properly with this patch applied.

Note: may require the GEL configuration test data regenerating for images which were previously being read incorrectly.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GelReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GelReader.java
@@ -121,7 +121,7 @@ public class GelReader extends BaseTiffReader {
       int originalBytes = ifd.getBitsPerSample()[0] / 8;
 
       for (int i=0; i<tmp.length/4; i++) {
-        long value = DataTools.bytesToShort(tmp, i*originalBytes,
+        long value = DataTools.bytesToInt(tmp, i*originalBytes,
           originalBytes, isLittleEndian());
         long square = value * value;
         float pixel = square * scale;


mtbc:
Re. config. regeneration: Unfortunately neither of the full repo jobs are looking good right now: one's disabled, the other's failing: Maybe it suffices to run the subset job? Over `gel/` rather than `biorad-gel/`?

sbesson:
NB: https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/14049/console confirms that some configuration files need to be regenerated to match the changes proposed in thsi PR as suggested in the description.

sbesson:
NB: https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/14049/console confirms that some configuration files need to be regenerated to match the changes proposed in this PR as suggested in its description.

mtbc:
For QA-21519 this PR changes the previous

before

to the new

after

which fits well with the original complaint that

> The bands which are brightest turn into black ...

in that with this PR the bands are no longer weirdly outlined.

rleigh-codelibre:
Configuration update in https://github.com/openmicroscopy/data_repo_config/pull/302

Might also be worth checking the existing six gel samples visually to check the before and after images are not horribly mangled.

sbesson:
File copied under the curated QA repository so that it can be picked up by tomorrow's repository job.

pwalczysko:
> to check the before and after images are not horribly mangled.


@sbesson what are ``before`` images ? (``after`` images are the ones with this PR included ?)

rleigh-codelibre:
@pwalczysko Before means the how the images appear without this PR, and after means how the images appear with this PR merged.

pwalczysko:
mported curated/gel into west ci (https://web-proxy.openmicroscopy.org/west-web/webclient/, user: gel). The appearance of the images in the old web viewer and Preview is as expected. No iviewer is installed.

imported the ``before`` images importing ``/ome/curated/gel/`` on the cli of eel latest. See user-11 http://web-dev-latest.openmicroscopy.org/webclient/?show=dataset-2852

Except for the inversion of the ``after`` images vs ``before`` images, and the lack of the bordering effect in the ``after`` images as noticed by @mtbc , which are both intended as I understand from the conversation here, there is a perfect match of features and intensities between ``before`` and ``after``.

Ready to merge FMPOV

Issue #3166, opened 2018-06-13
Errors importing MetaMorph time series with ImageJ or bfconvert


Cumol:
Here are the first 10 frames of an image series captured with MetaFlour (a MetaMorph software):

https://www.dropbox.com/s/2xvxycbwvm9p85k/test-series.zip?dl=0

Using the Fiji BioFormats Plugin Importer with the "Group files with simlar names" option ticket gives me the following error:

```
(Fiji Is Just) ImageJ 2.0.0-rc-67/1.52c; Java 1.8.0_171 [64-bit]; Mac OS X 10.11.6; 297MB of 6222MB (4%)
 
java.lang.NullPointerException
	at loci.formats.in.MetamorphTiffReader.getField(MetamorphTiffReader.java:568)
	at loci.formats.in.MetamorphTiffReader.initFile(MetamorphTiffReader.java:243)
	at loci.formats.FormatReader.setId(FormatReader.java:1397)
	at loci.plugins.in.ImportProcess.initializeFile(ImportProcess.java:499)
	at loci.plugins.in.ImportProcess.execute(ImportProcess.java:142)
	at loci.plugins.in.Importer.showDialogs(Importer.java:140)
	at loci.plugins.in.Importer.run(Importer.java:76)
	at loci.plugins.LociImporter.run(LociImporter.java:78)
	at ij.IJ.runUserPlugIn(IJ.java:222)
	at ij.IJ.runPlugIn(IJ.java:186)
	at ij.Executer.runCommand(Executer.java:137)
	at ij.Executer.run(Executer.java:66)
	at java.lang.Thread.run(Thread.java:748)
```

The same happens when using the bfconvert tool. Either on the whole folder or a singe file:

```
./bfconvert -debug ~/Desktop/test/23.001 23.001.tif
```

Gives this:

```
No update needed
/Users/MuadWork/Desktop/test/23.001
Could not find loci.formats.in.SlideBook6Reader
java.lang.ClassNotFoundException: loci.formats.in.SlideBook6Reader
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381) ~[na:1.8.0_171]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424) ~[na:1.8.0_171]
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349) ~[na:1.8.0_171]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357) ~[na:1.8.0_171]
[...]

PR #3167, opened 2018-06-14
MetaMorph - Prevent NullPointerException

dgault:
This is a fix for the NullPointerException and a divide by zero exception as reported in https://github.com/openmicroscopy/bioformats/issues/3166

To test:
- Open the provided sample file in https://github.com/openmicroscopy/bioformats/issues/3166 with Bio-Formats
- Without the PR an exception is thrown, with the PR the file should open and display without error

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphTiffReader.java
@@ -73,9 +73,9 @@ public class MetamorphTiffReader extends BaseTiffReader {
   // -- Fields --
 
   private String[] files;
-  private int wellCount = 0;
-  private int fieldRowCount = 0;
-  private int fieldColumnCount = 0;
+  private int wellCount = 1;
+  private int fieldRowCount = 1;
+  private int fieldColumnCount = 1;
   private boolean dualCamera = false;
 
   // -- Constructor --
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/MetamorphTiffReader.java
```diff
@@ -73,9 +73,9 @@
   // -- Fields --
 
   private String[] files;
-  private int wellCount = 0;
-  private int fieldRowCount = 0;
-  private int fieldColumnCount = 0;
+  private int wellCount = 1;
```
sbesson: I assume there is no downside to changing the default values here since these are private and not introspectable. The only thing I could think of would be some logic in the reader that would rely on the initialization state of these variables i.e.`(if wellCount == 0)`.

dgault: There are no checks currently to ensure the values are initialised. If they haven't been set correctly then you actually end up with other issues such as 0 image count and divide by zero issues. I can see how it might be beneficial to log a warning in the case when we haven't initialised them and are falling back to the default though.

dgault:
Config has been added https://github.com/openmicroscopy/data_repo_config/pull/305

sbesson:
Merging as per https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-repo/111/

PR #3169, opened 2018-06-15
OIR - Prevent infinite loop

dgault:
This is a follow up to a thread on the forums: https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8522&sid=c0ae5e3d8131cd28aaf123a570cd3ca6

From debugging the file is caught in a while loop at https://github.com/dgault/bioformats/blob/47d5e46ed52f13c1388f7315545bebe684bdcc0e/components/formats-gpl/src/loci/formats/in/OIRReader.java#L615

Each iteration of the loop the file pointer remains unchanged and the length read at https://github.com/dgault/bioformats/blob/47d5e46ed52f13c1388f7315545bebe684bdcc0e/components/formats-gpl/src/loci/formats/in/OIRReader.java#L618 is 0

To test:
- Ensure that all builds and tests remain green
- Without the PR test using QA-21538 and see that importing the file results in an infinite loop
- With the PR reproduce the above test and ensure that the image opens and displays correctly

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -616,7 +616,7 @@ public class OIRReader extends FormatReader {
       s.findString("<?xml");
       s.seek(s.getFilePointer() - 9);
       int length = s.readInt();
-      if (length < 0 || length + s.getFilePointer() > s.length()) {
+      if (length <= 0 || length + s.getFilePointer() > s.length()) {
         break;
       }
       long fp = s.getFilePointer();


sbesson:
The change looks good and https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/14978/ still passes with it included against the existing sets of OIR files. Should the QA file be also configured and added to the curated repository for non-regression testing?

melissalinkert:
https://trello.com/c/qYQEbBm8/49-very-slow-oir-file-hanging (QA 21508) seems to be fixed by this as well.

sbesson:
:+1: worth configuring and adding the two datasets to the curated repository?

sbesson:
Looks good. Importing the submitted file into a server with this version of Bio-Formats resulted into a successful import followed by a successful pyramid generation (meaning that all pixels were correctly read). Merging for 5.9.0.

PR #3171, opened 2018-06-17
Build updates

rleigh-codelibre:
See https://github.com/ome/bio-formats-build/pull/22


sbesson:
Merging to simplify the merging of `5.9.0` into `east`

PR #3173, opened 2018-06-18
Leica LIF: only prioritize C in dimension order if channels are RGB

melissalinkert:
See https://trello.com/c/Zp2rgbqf/246-leica-lif-dimension-problems

To test, use ```data_repo/curated/leica-lif/pascal/test.lif```.  Without this change, verify that the dimension order (```XYCZT```) is incorrect by comparing what LAS X shows against each channel as opened in ImageJ.
I would expect C=0, Z=0, T=0 and C=1, Z=0, T=0 in ImageJ to be from the same channel according to LAS X.

With this change, the same comparison should show that C=0 and C=1 are now distinct channels.  The dimension order should now be reported as ```XYZCT```.  Builds should remain green, and this should be safe for a patch release if needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -2159,7 +2159,7 @@ public class LIFReader extends FormatReader {
     Arrays.sort(bytes);
     ms.dimensionOrder = "XY";
     if (getRGBChannelCount() == 1 || getRGBChannelCount() == getSizeC()) {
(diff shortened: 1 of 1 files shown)


dgault:

Tested using `data_repo/curated/leica-lif/pascal/test.lif`. Without this PR both channels appear to be identical when viewed with the second channel appearing in the second half of the z stacks for each. The dimension order is XYCZT.

With the PR included both channels display correctly and are different. The dimension order is XYZCT.

PR #3174, opened 2018-06-20
Remove Java 7 from Appveyor matrix

sbesson:
See https://central.sonatype.org/articles/2018/May/04/discontinue-support-for-tlsv11-and-below/ for more details

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,7 +32,7 @@ init:
   - 'if [%java%] == [9] set "JAVA_HOME=C:\Program Files\Java\jdk9"'
   - PATH=%JAVA_HOME%\bin;%PATH%
   - 'if [%build%] == [maven] set "MAVEN_OPTS=-Dmaven.repo.local=%AV_BF_M2%"'
-  - 'if [%build%] == [maven] and [%java%] == [1.7] set "MAVEN_OPTS=-Dmaven.repo.local=%AV_BF_M2% -Dhttps.protocols=TLSv1,TLSv1.1,TLSv1.2"'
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
The change looks fine and all builds are green, so it looks fine to merge.

However, it might also be worth considering an alternative solution: dropping the Java 7 build, since later supported versions of Java are unaffected by this change.

If you retain this approach, you might also wish to make it conditional on both ` [%build%] == [maven]` and `[%java%] == [1.7]` so that you're only setting `http.protocols` for Java 7 builds.

sbesson:
Thanks @rleigh-codelibre for the quick feedback, I was just reviewing how many repositories would need to be similarly patched - https://github.com/search?q=user%3Aome+user%3Aopenmicroscopy+MAVEN_OPTS&type=Code.

Dropping the Java 7 component of the Appveyor matrix would certainly be an option to consider not only for reducing the logic of the YML file but also to reduce the current execution times.

In the meantime I will try and push an extra conditional to set the `http.protocols` for Java 7 as you suggested so that we can review and discuss the pros and cons of both approaches

rleigh-codelibre:
@sbesson It will be a subset the repositories referenced in https://github.com/ome/bio-formats-build/pull/22 (they don't all have appveyor enabled, but they might have appveyor configurations to update even then).

sbesson:
After some thoughts, removing Java 7 from the Appveyor matrix fully as suggested in https://github.com/openmicroscopy/bioformats/pull/3174#issuecomment-398883327. In addition to the problem highlighted here, we definitely have limited capacity on free Appveyor and it is probably even sufficient to test LTS Java on Windows.

Once this is green, proposing to get this is in and port the changes to other repos as necessary like https://github.com/ome/bio-formats-documentation/ to reduce the false positives for new PRs failures.

rleigh-codelibre:
Looks good to me, thanks.

sbesson:
Got separate 👍 from @joshmoore, @melissalinkert and @dgault, merging.

PR #3175, opened 2018-06-21
Clarify ordering expectation in get*UsedFiles Javadocs

melissalinkert:
See https://trello.com/c/zgKbVcun/32-add-documentation-for-getusedfiles-ordering

This is a documentation change only, so a review for clarity/correctness should be sufficient.  I would not expect builds to be impacted.

Definitely not urgent, so feel free to push to a later patch release as needed.


sbesson:
:+1: Clarifying the expectations from the reader interface is definitely worth having int 5.9.0.

PR #3176, opened 2018-06-22
Fix download URLs in bf.bat and bf.sh

melissalinkert:
See https://trello.com/c/gcxA7ZMJ/514-tools-downloads-link-falls-out-of-date

The new URLs remove an explicit version number reference, and should currently point to the 5.8.2 download directory.

This shouldn't affect builds or behavior other than changing what is printed when using ```showinf``` or ```bfconvert``` cannot find valid Bio-Formats jars.  Definitely not urgent, and should be safe for a patch release.

### DIFF
--- a/tools/bf.bat
+++ b/tools/bf.bat
@@ -59,7 +59,7 @@ if exist "%BF_JAR_DIR%\bioformats_package.jar" (
   echo Required JAR libraries not found. Please download:
   echo   bioformats_package.jar
   echo from:
-  echo   http://downloads.openmicroscopy.org/latest/bio-formats5.1
+  echo   https://downloads.openmicroscopy.org/latest/bio-formats/artifacts/
   echo and place in the same directory as the command line tools.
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
The URLs look fine.

Also will require updating in https://github.com/ome/bio-formats-bundles/tree/master/bftools/scripts .  Hopefully we can adopt the decoupled version soon.

melissalinkert:
https://github.com/ome/bio-formats-bundles/commit/63bedfce284004fd5c5fbefacd3bde542e2edb48 removed the URLs from ```bf.bat``` and ```bf.sh```, so presumably that means a PR in ```bio-formats-bundles``` isn't necessary?

rleigh-codelibre:
@melissalinkert Ah, sorry, you're right.  Because it relies upon always having the jar in place as a result of the maven assembly, the instructions are no longer necessary there (or here, really).

PR #3178, opened 2018-07-02
Correcting image pixel metadata parsing for LEOReader

dmankus:
Attempting to correct the issue highlighted here: https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8535

Currently my testing has this compiling and running fine as an ImageJ plugin, though something goes wrong, as the actual metadata values in ImageJ end up as defaults without the parsed values. Probably has something to do with assigning the variables correctly, as I'm unsure I've done that right.

Any guidance on how to troubleshoot from here would be appreciated.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LEOReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LEOReader.java
@@ -96,23 +96,23 @@ public class LEOReader extends BaseTiffReader {
       for (int line=36; line<lines.length; line++) {
         if (lines[line].equals("AP_IMAGE_PIXEL_SIZE")) {
           // pixel size is stored in nm, converted now to micrometers
-          double xSize = Double.parseDouble(lines[line++].split("\\s+=\\s+")[1].replace(" nm","e+03"));
+          double xSize = Double.parseDouble(lines[++line].split("\\s+=\\s+")[1].replace(" nm","e-03"));
         }
         else if (lines[line].equals("AP_WD")) {
           //working distance stored in mm, converting now to micrometers
-          double workingDistance = Double.parseDouble(lines[line++].split("\\s+=\\s+")[1].replace(" mm","e-03"));
+          double workingDistance = Double.parseDouble(lines[++line].split("\\s+=\\s+")[1].replace(" mm","e+03"));
         }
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/LEOReader.java
```diff
@@ -88,25 +90,32 @@ protected void initStandardMetadata() throws FormatException, IOException {
...
-      }
-    }
-
     if (getMetadataOptions().getMetadataLevel() != MetadataLevel.MINIMUM) {
-      // physical sizes stored in meters
-      xSize = Double.parseDouble(lines[3]) * 1000000;
+      for (int line=36; line<lines.length; line++) {
+        if (lines[line].equals("AP_IMAGE_PIXEL_SIZE")) {
+          // pixel size is stored in nm, converted now to micrometers
+          double xSize = Double.parseDouble(lines[++line].split("\\s+=\\s+")[1].replace(" nm","e-03"));
```
dgault: As the variables xSize etc have already been declared as instance variables then you can simply use:
`xSize = Double.parseDouble(lines[++line].split("\\s+=\\s+")[1].replace(" nm","e-03"));`

Declaring it again creates a local variable which is not used outside of the block and as such results in the default values still appearing. Making that small change sees the values being parsed and populated correctly with the sample files from https://keybase.pub/dmankus/

dgault:
Thanks @dmankus, with the latest changes I have tested using the sample files provided in https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8535

The metadata values appear to be populated correctly and match the values annotated on the images. I have opened a PR to add these sample files to the test suite. We currently only have 1 our sample file configured for this format and its core metadata is unaffected by this PR.

dgault:
Ive updated the PR by pushing a small change to restrict the scope of the `eht` and `filament` variables. This is done to prevent breaking serialization of the memo cache files and thus enable us to include these changes much sooner in an upcoming patch release.

dgault:
Ive pushed a further change to fix failed tests in https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/23184/console

This will maintain the original behaviour in instances when the AP_IMAGE_PIXEL_SIZE is not present.

Issue #3120, opened 2018-04-04
DeltaVisionReader doesn't support files with more than 5 channels


jwerbin:
Hi,

When opening .dv files with more than 5 wavelengths DeltavisionReader throws an exception.

I think the culprit is in DeltavisionReader.initExtraMetadata()
https://github.com/openmicroscopy/bioformats/blob/dcefe24cf8c9191fed22dc497b4dbd38b202b925/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java#L596-L597

When trying to load a 6 wavelength file using the ImageJ plugin
> java.lang.reflect.InvocationTargetException
>	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
>	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
>	at java.lang.reflect.Method.invoke(Method.java:597)
>	at ij.Command.runPlugIn(Command.java:148)
>	at ij.Command.runCommand(Command.java:97)
>	at ij.Executer.run(Executer.java:64)
>	at java.lang.Thread.run(Thread.java:662)
> Caused by: java.lang.ArrayIndexOutOfBoundsException: 5
>	at loci.formats.in.DeltavisionReader.initExtraMetadata(DeltavisionReader.java:744)
>	at loci.formats.in.DeltavisionReader.initFile(DeltavisionReader.java:272)
>	at loci.formats.FormatReader.setId(FormatReader.java:1320)
>	at loci.plugins.in.ImportProcess.initializeFile(ImportProcess.java:482)
>	at loci.plugins.in.ImportProcess.execute(ImportProcess.java:146)
>	at loci.plugins.in.Importer.showDialogs(Importer.java:141)
>	at loci.plugins.in.Importer.run(Importer.java:79)
>	at loci.plugins.LociImporter.run(LociImporter.java:81)
>	at ij.IJ.runUserPlugIn(IJ.java:185)
>	at ij.IJ.runPlugIn(IJ.java:152)
>	at ij.IJ.runPlugIn(IJ.java:141)
>	at HandleExtraFileTypes.openImage(HandleExtraFileTypes.java:405)
>	at HandleExtraFileTypes.run(HandleExtraFileTypes.java:57)
>	at ij.IJ.runUserPlugIn(IJ.java:185)
[...]

PR #3181, opened 2018-07-04
Deltavision: update to detect up to 12 channels

melissalinkert:
See https://trello.com/c/nik6LmGE/256-deltavision-with-more-than-5-channels and #3120.

To test, use the .rcpnl file from QA 21574.  Without this PR, ```showinf RC_SevenChannelScan.rcpnl``` should throw an ```ArrayIndexOutOfBoundsException``` as noted in #3120.  With this PR, the same test should not throw an exception, and should result in 7 channels being read.  A valid and plausible wavelength (in nm) should be present in original metadata and OME-XML for all 7 channels.  The TIFF file in QA 21574 should match the first image when flipped horizontally (expected since the DV reader flips planes to account for the origin being in the lower-left corner).

I think we had established that adding new ```static final``` variables doesn't impact memo files, but it might be a good idea to double-check anyway.  If it's a problem, I can remove ```NEW_TYPE```and ```MAX_CHANNELS``` for now so that this can be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -78,6 +78,9 @@ public class DeltavisionReader extends FormatReader {
   private static final short LITTLE_ENDIAN = -16224;
   private static final int HEADER_LENGTH = 1024;
 
+  private static final int NEW_TYPE = 100;
+  private static final int MAX_CHANNELS = 12;
+
   private static final String[] IMAGE_TYPES = new String[] {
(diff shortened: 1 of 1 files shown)


dgault:
Code looks good and builds and tests have been green with this PR and associated config PR included. 

Tested using the QA 21574 file suggested, without the PR included an `ArrayIndexOutOfBoundsException`  is thrown. With the PR included the file read and wavelength metadata is present for the 7 available channels.

The static final variables are fine and do not impact memo files. Furthermore the PR has no impact on existing older type DV files.

PR #3184, opened 2018-07-10
Check series count vs OME-XML Image count

melissalinkert:
Also adds a ```testUnflattenedSaneOMEXML``` to check for usable OME-XML when
resolution flattening is disabled.  See https://trello.com/c/j6HbJibe/237-tests-check-image-count-matches-series-count

To test, verify that build failures are not introduced.  To check that the test will fail when appropriate, modify one of the readers to create one OME-XML ```Image``` per resolution when resolutions are not flattened (e.g. around https://github.com/openmicroscopy/bioformats/blob/east/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java#L360).  ```test-automated``` on the affected reader's folder(s) should then fail.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -608,66 +608,39 @@ public class FormatReaderTest {
       result(testName, true);
       return;
     }
+    String msg = checkOMEXML(reader);
+    result(testName, msg == null, msg);
+  }
+
+  @Test(groups = {"all", "xml", "automated"})
+  public void testUnflattenedSaneOMEXML() {
+    String testName = "testUnflattenedSaneOMEXML";
+    if (!initFile()) result(testName, false, "initFile");
+    if (!config.hasValidXML()) {
+      LOGGER.debug("Skipping valid XML test");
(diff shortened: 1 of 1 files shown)


components/test-suite/src/loci/tests/testng/FormatReaderTest.java
```diff
@@ -2724,4 +2685,94 @@ private static void result(String testName, boolean success, String msg) {
...
+      }
+    }
+    catch (Throwable t) {
+      LOGGER.info("", t);
+      msg = t.getMessage();
+    }
+    return msg;
+  }
+
+  private IFormatReader setupUnflattenedReader() throws FormatException, IOException {
```
sbesson: Only question here is whether this could be turned into `IFormatReader setupReader(boolean flattened)` that could be reused by `setupReader()`

melissalinkert: Definitely makes sense.  Would you prefer to have that done here, or as a follow-up PR?

sbesson: Follow-up PR is fine. Merging to reduce the number of open contributions

PR #3185, opened 2018-07-12
Allow .fake files to represent a pyramid

melissalinkert:
See https://trello.com/c/hvjosNYD/18-fake-resolutions

The ```resolutions``` key defines the total number of resolutions, so must be a positive integer.  By default, it is 1.  The ```resolutionScale``` key defines how ```SizeX``` and ```SizeY``` are calculated for each resolution; the sizes are divided by ```resolutionScale``` for each successively smaller resolution.  It must be greater than 1, and is 2 by default.

To test, try ```showinf``` with a few pyramids as indicated in the Javadoc.  Verify that unit tests continue to pass and that the new ```testPyramid*``` tests are being run.


sbesson:
Follow-up captured as https://trello.com/c/coh3IxDi/34-fake-resolution-examples-and-doc.

PR #3187, opened 2018-07-25
OIR: prevent current file from being overridden if it has the .oir extension

melissalinkert:
See https://trello.com/c/VzvQbYBf/14-incorrect-handling-of-similarly-named-oir-files

To test, use the files from QA 21577.  Without this PR, ```showinf -nopix R13L-S2-5_X20_ZOOM1_Z.oir``` should show that ```R13L-S2-5_X20_ZOOM1_Z_CONTROL.oir``` is in the used file list (despite being a separate dataset).  With this PR, ```showinf -nopix``` on either file should show only the selected file in the used file list.

Builds should remain green, and memo files should be unaffected so this should be safe for a patch release.  There is a workaround, though, so could certainly be deferred until after 5.9.1 if needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -310,9 +310,21 @@ public class OIRReader extends FormatReader {
           catch (NumberFormatException e) {
           }
         }
-        else if (file.startsWith(prefix) && checkSuffix(file, "oir")) {
-          current = new Location(parent, file);
-          currentId = current.getAbsolutePath();
+      }
+
+      // tried to initialize using one of the companion files
+      // reset the current file to the main .oir file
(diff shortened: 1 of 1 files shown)


sbesson:
The change is successfully tested against the submitted sample files by the [repository tests](https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/22899/console). Happy to discuss tomorrow but the fix looks small and simple enough and fixes an existing issue reported by consumers without asking them to rename or move their files in a different folder so my vote would be for integrating into 5.9.1.

dgault:

Tested using QA 21577 as suggested, without the PR reading `R13L-S2-5_X20_ZOOM1_Z.oir` shows `R13L-S2-5_X20_ZOOM1_Z_CONTROL.oir` in the used files list. With the PR included reading either file only shows that given file in the file list.

PR is good to merge.

PR #3189, opened 2018-07-30
Initial pyramid OME-TIFF writer

melissalinkert:
See https://trello.com/c/MRzY025E/2-java-writer-implementation

```loci.formats.out.PyramidOMETiffWriter``` is the new writer for pyramid OME-TIFFs.  It will be used whenever an OME-TIFF extension is specified and resolution metadata is supplied to the writer.  ```PyramidOMETiffWriter``` extends ```OMETiffWriter```; the only major differences are the addition of ```NEW_SUBFILE_TYPE``` and ```SUB_IFD``` tags during ```saveBytes```, and the ```SUB_IFD``` tag and IFD offset overwriting during ```close()```.  Note that when multiple planes per resolution are present, the assumption is that every plane in the resolution is written before switching resolutions (i.e. R0P0, R0P1, ... R0Pn, R1P0, R1P1, ...).  If the spec differs, this will need to be updated in the writer.

Stand-alone examples are ```components/formats-gpl/utils/MinimumPyramidWriter.java``` (writing a pyramid when the pixels/metadata are not supplied by a reader) and ```components/formats-gpl/utils/WritePyramid.java``` (writing a pyramid by reading a set of files, one per resolution).  Unit tests are in ```components/formats-bsd/src/loci/formats/utests/tiff/PyramidTest.java```.

When a source file is recognized as a pyramid, ```bfconvert -noflat input output.ome.tiff``` (using BigTIFF if needed) should result in an OME-TIFF that ```showinf -nopix -noflat``` recognizes as a pyramid.  Be aware that some formats with pyramids are not recognized as such on ```east``` (independent of this PR), so check ```showinf -nopix -noflat``` before trying a conversion (NDPI for example won't work, but SVS and Leica SCN should be OK).  ```showinf -nopix``` on the resulting output file will now show a series count matching the resolution count (plus any extra series), instead of just the full resolution(s).  This is consistent with how flattened resolutions are handled in other readers, and was the only real reader change necessary.

Testing pyramid writing exposed a few bugs in ```TiffSaver```, particularly when overwriting IFD values/offsets in BigTIFF files or when a tag value is not inlined and requires multiple bytes per array entry.

The main thing that needs discussion here is whether the API for supplying resolution metadata is what we actually want to support.  I picked an "extend MetadataStore" strategy partly to reduce the burden on the caller in the conversion case, but mostly just needed to get *something* in place to move on to the actual writer.  An alternative (as referenced on the Trello card) would be to have ```setSizeX``` and ```setSizeY``` in ```IFormatWriter``` and just remove the ```IPyramidStore```/```OMEPyramidStore```/```MetadataTools```/etc. changes here.  Either way, enforcement of resolution ordering needs to be added - it's a minor change, but was waiting for the spec PR to make sure I get it right.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -503,95 +503,99 @@ public final class ImageConverter {
     long timeLastLogged = System.currentTimeMillis();
     for (int q=first; q<last; q++) {
       reader.setSeries(q);
-      firstTile = true;
+      for (int res=0; res<reader.getResolutionCount(); res++) {
+        reader.setResolution(res);
+        firstTile = true;
 
-      if (!dimensionsSet) {
-        width = reader.getSizeX();
-        height = reader.getSizeY();
-      }
-
-      int writerSeries = series == -1 ? q : 0;
-      writer.setSeries(writerSeries);
-      writer.setInterleaved(reader.isInterleaved() && !autoscale);
-      writer.setValidBitsPerPixel(reader.getBitsPerPixel());
-      int numImages = writer.canDoStacks() ? reader.getImageCount() : 1;
-
-      int startPlane = (int) Math.max(0, firstPlane);
-      int endPlane = (int) Math.min(numImages, lastPlane);
-      numImages = endPlane - startPlane;
-
-      if (channel >= 0) {
-        numImages /= reader.getEffectiveSizeC();
-      }
-      if (zSection >= 0) {
-        numImages /= reader.getSizeZ();
-      }
-      if (timepoint >= 0) {
-        numImages /= reader.getSizeT();
-      }
-
-      total += numImages;
-
-      int count = 0;
-      for (int i=startPlane; i<endPlane; i++) {
-        int[] coords = reader.getZCTCoords(i);
-
-        if ((zSection >= 0 && coords[0] != zSection) || (channel >= 0 &&
-          coords[1] != channel) || (timepoint >= 0 && coords[2] != timepoint))
-        {
-          continue;
+        if (!dimensionsSet) {
+          width = reader.getSizeX();
+          height = reader.getSizeY();
         }
 
-        String outputName = FormatTools.getFilename(q, i, reader, out, zeroPadding);
-        if (outputName.equals(FormatTools.getTileFilename(0, 0, 0, outputName))) {
-          writer.setId(outputName);
-          if (compression != null) writer.setCompression(compression);
+        int writerSeries = series == -1 ? q : 0;
+        writer.setSeries(writerSeries);
+        writer.setResolution(res);
+        writer.setInterleaved(reader.isInterleaved() && !autoscale);
+        writer.setValidBitsPerPixel(reader.getBitsPerPixel());
+        int numImages = writer.canDoStacks() ? reader.getImageCount() : 1;
+
+        int startPlane = (int) Math.max(0, firstPlane);
+        int endPlane = (int) Math.min(numImages, lastPlane);
+        numImages = endPlane - startPlane;
+
+        if (channel >= 0) {
+          numImages /= reader.getEffectiveSizeC();
         }
-        else {
-          int tileNum = outputName.indexOf(FormatTools.TILE_NUM);
-          int tileX = outputName.indexOf(FormatTools.TILE_X);
-          int tileY = outputName.indexOf(FormatTools.TILE_Y);
-          if (tileNum < 0 && (tileX < 0 || tileY < 0)) {
-            throw new FormatException("Invalid file name pattern; " +
-              FormatTools.TILE_NUM + " or both of " + FormatTools.TILE_X +
-              " and " + FormatTools.TILE_Y + " must be specified.");
+        if (zSection >= 0) {
+          numImages /= reader.getSizeZ();
+        }
+        if (timepoint >= 0) {
+          numImages /= reader.getSizeT();
+        }
+
+        total += numImages;
+
+        int count = 0;
+        for (int i=startPlane; i<endPlane; i++) {
+          int[] coords = reader.getZCTCoords(i);
+
+          if ((zSection >= 0 && coords[0] != zSection) || (channel >= 0 &&
+            coords[1] != channel) || (timepoint >= 0 && coords[2] != timepoint))
+          {
+            continue;
           }
-        }
 
-        int outputIndex = 0;
-        if (nextOutputIndex.containsKey(outputName)) {
-          outputIndex = nextOutputIndex.get(outputName);
-        }
-
-        long s = System.currentTimeMillis();
-        long m = convertPlane(writer, i, outputIndex, outputName);
-        long e = System.currentTimeMillis();
-        read += m - s;
-        write += e - m;
-
-        nextOutputIndex.put(outputName, outputIndex + 1);
-        if (i == endPlane - 1) {
-          nextOutputIndex.remove(outputName);
-        }
-
-        // log number of planes processed every second or so
-        if (count == numImages - 1 || (e - timeLastLogged) / 1000 > 0) {
-          int current = (count - startPlane) + 1;
-          int percent = 100 * current / numImages;
-          StringBuilder sb = new StringBuilder();
-          sb.append("\t");
-          int numSeries = last - first;
-          if (numSeries > 1) {
-            sb.append("Series ");
-            sb.append(q);
-            sb.append(": converted ");
+          String outputName = FormatTools.getFilename(q, i, reader, out, zeroPadding);
+          if (outputName.equals(FormatTools.getTileFilename(0, 0, 0, outputName))) {
+            writer.setId(outputName);
+            if (compression != null) writer.setCompression(compression);
           }
-          else sb.append("Converted ");
-          LOGGER.info(sb.toString() + "{}/{} planes ({}%)",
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by melissalinkert** [comment]
Two example files finally uploaded to ```data_repo/inbox/new-pyramid-ome-tiff/```.  These are the results of ```bfconvert -noflat``` using 6708c7b on ```data_repo/curated/svs/public/77917.svs``` and ```data_repo/curated/leica-scn/public/openslide/Leica-1/Leica-1.scn``` (the latter for comparison with ```data_repo/curated/ome-tiff/sub-resolution/Leica-1/Leica-1-subifds.ome.tiff```).

**Comment by rleigh-codelibre** [comment]
@melissalinkert Would it be possible to also put these example files somewhere public so I could use them for testing OME Files C++?

Some initial comments on the Java implementation:

- For the annotation namespace `openmicroscopy.org/PyramidResolution`, why the use of `Pyramid` when the rest of the API is using `resolution`.  Would e.g. `ResolutionLevels` or similar be more in keeping?  The same applies to `IPyramidHandler` and other naming with pyramid rather than resolution.
- These annotations are only ever going to be used when writing, so it would make sense to delete them before writing out the OME-XML to avoid complications when converting metadata in the future, where they might be used in favour of newly-added annotations
- The `Resolution` helper class contains `x` and `y` members, but it might be worth adding `z` in there as well to be fully future proof so that the API can work with `z` reductions without any future breakage.  Just set `z` to the full SizeZ for now.
- The `IPyramidStore` and `OMEPyramidStore` additions seem overly complex.  What would be wrong with adding a couple of helper methods to `MetadataTools` to get and set these annotations?  I'm not sure I see why any methods need adding at the `MetadataStore` level when it's simply adding an annotation.  This wouldn't scale gracefully as you add more annotations in the future.
- The PyramidResolution ID is `Annotation:Resolution:i`, where `i` is the image index.  However, why isn't the annotation being linked to the image so that it doesn't rely on the indexing this way?  If you link to the image, the annotation is trivially fetchable as a linked annotation and with zero ambiguity.  It also caters for sparse annotations where not all series have sub-resolutions.
- Why is there a separate writer for Pyramid OME-TIFF?  Why not have a single writer and base it on whether or not there are annotations present on a per-series basis?  It would need only implement the new `IPyramidHandler` interface.

Regards,
Roger

**Comment by sbesson** [comment]
Thanks for the early feedback @rleigh-codelibre.

Agreed this is the first working Java implementation of the pyramidal OME-TIFF writer and there are certainly several aspects to consider and review in terms of API. Our immediate next step will be to carry out some extensive functional testing of this Java implementation using our entire stack.

As this work will probably involve the generation of OME-TIFF pyramids for a variety of formats, we will look into making the generated data available when applicable.

**Comment by rleigh-codelibre** [comment]
@sbesson I'm working on the C++ writer implementation right now.  These files would be very useful for testing interoperability, and having access to them this week would be useful if you could arrange that.

**Comment by melissalinkert** [comment]
@sbesson / @chris-allan / @emilroz : as discussed Tuesday (and noted in https://trello.com/c/MRzY025E/2-java-writer-implementation), 0f10d8d adds a first draft of a simpler API for specifying resolution information.  ```components/formats-gpl/utils/AlternativeMinimumPyramidWriter.java``` is the corresponding example; it should produce the same results as ```MinimumPyramidWriter```, so is an easy way to compare APIs.  The examples work and unit tests continue to pass, so I wouldn't expect an impact on testing, but feel free to revert if needed for tomorrow.

File: components/formats-api/src/loci/formats/Resolution.java [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+  // -- Fields --
+
+  public PositiveInteger sizeX;
+  public PositiveInteger sizeY;
+  public int index;
+
+  // -- Constructors --
+
+  public Resolution(int index, int x, int y) {
+    this.index = index;
```
rleigh-codelibre: Is the index needed at this level, or would it not be better for this to be represented at a higher level, for example the index of this object in a container?

rleigh-codelibre:
Regarding addition and removal of the pyramid annotations, please see [qi-pyramid-write](https://github.com/rleigh-codelibre/ome-files-cpp/tree/qi-pyramid-write) for an example of this using a few helper functions added to `MetadataTools`.  This allows for easy addition of annotations to one series at a time, or all series at once, and likewise for querying and for removal.  These can be combined to allow editing.  And the function to remove all annotations can be used by the writer to drop the annotations before writing out the OME-XML metadata.  This is also making use of annotation references from Image to link the image to a specific pyramid map annotation, which allows a sparse representation.  This is also more efficient to query because it requires a linear scan through the annotation references for a single image element, rather than for every structured annotation in the whole model.

sbesson:
Briefly tested the conversion with this PR today from the [Bio-Formats image](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-image/). I was not able to generate the pyramidal OME-TIFF as described in https://github.com/openmicroscopy/bioformats/pull/3189#issuecomment-409066844 since the `bfconvert` operation failed with :

```
Exception in thread "main" java.lang.IllegalArgumentException
	at java.nio.Buffer.limit(Buffer.java:275)
	at loci.common.NIOFileHandle.buffer(NIOFileHandle.java:609)
```

Reverting 0f10d8d and/or removing the `-noflat` option led to the same error. Re-executed the command in `-debug` mode so that we can identify the source of the problem with @melissalinkert.

melissalinkert:
I haven't been able to reproduce the above error on my local machines, but can at least consistently reproduce on necromancer using plain ```bfconvert 77917.svs 77917.ome.tiff```, even with develop alone and nothing from east (including this PR).  Still debugging, but I think we've at least ruled out it being an issue introduced by this PR.

pwalczysko:
So regarding the bit from the PR header

>should result in an OME-TIFF that showinf -nopix -noflat recognizes as a pyramid

Maybe the bit of the output of the ``showinf`` which identifies the ome.tiff created using this PR is (see below)


```
Series #0 :
	Resolutions = 5
		sizeX[0] = 96999
		sizeX[1] = 24249
		sizeX[2] = 6062
		sizeX[3] = 3031
		sizeX[4] = 1024
```

???

melissalinkert:
@pwalczysko: yes, that output indicates that series 0 is a pyramid with 5 resolutions.

pwalczysko:
I was capturing the results of my up-till-now testing in https://trello.com/c/ZHyYCXv6/32-ome-tiff-pyramids-functional-testing#comment-5b69bbf0df291987fa8b0ddb. Basically, I concur with @melissalinkert 's reported success on building ome.tiff from an svs and Leica files as per https://github.com/openmicroscopy/bioformats/pull/3189#issuecomment-409066844 The only problems encountered are rather higher level (``bfconvert``) level RFEs, specified on the same Trello card.

@sbesson is telling me that https://github.com/openmicroscopy/bioformats/pull/3189#issuecomment-410805187 is being taken care of in a separate PR.

Is there anything else what would be useful at this stage of the testing for this PR ?

melissalinkert:
@rleigh-codelibre: We appreciate the thorough review. However fundamentally this PR is at the working prototype stage as we are still sorting out final API decisions internally.  Your suggestions are taken under advisement, but do expect the API to change from what's here, and don't rely upon implementation specifics such as the PyramidResolution MapAnnotation until the final API is announced.  We will of course notify the community via all usual channels once this is finalized and released.

rleigh-codelibre:
@melissalinkert If you would like me to join in any of these discussions, I'd be happy to do so.  I would like to keep https://github.com/ome/ome-files-cpp/pull/100 as closely aligned with your Java implementation as possible, and can update it to match changes here as they are made.  I hope that the proposed changes there might also be useful for potential improvements to this PR as well, most of which were already detailed above.

Regards,
Roger

sbesson:
Looking at the latest results of the [repository jobs](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/28903/console) with this PR included,

- [ome-xml](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/28903/console) seems to be related to the new `OMEPyramidStore` introduced in the PR
- [zeiss-tiff](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/28975/) and [zeiss-czi](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/28980/) are known to have some flaky results and should be handled/captured separately
- [mrc](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/29008/) and [avi](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/29037/) are failing due to open 5.9 PRs (although this indicates issues with the `testng.allow-missing` parameter)
- [nd2](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/29043/console) seem to indicate an issue with a codec PR
- [quicktime](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/28949/) needs investigation

Out of these, the main concern related to this PR is the `ome-xml` - do you want to handle this as part of this PR @melissalinkert? As discussed yesterday, the priority is certainly to keep driving the functional testing of the conversion/reading across a set of pyramidal file formats and then explore alternative APIs in a second step. Happy to have this merged into an integration branch as the first intermediate version of the pyramid writing API.

melissalinkert:
eb6959b should fix ```ome-xml```; everything else can be handled separately (as discussed on codec PRs/cards).

PR #3192, opened 2018-08-09
Type detection and physical size unit fixes for MRC

melissalinkert:
Backported from a private PR.

To test, use ```data_repo/curated/mrc/samples/agar-exttype/golgi-modified-AGAR.mrc```.  As noted in the readme, this is a copy of ```data_repo/curated/mrc/samples/golgi.mrc``` with 10 bytes modified to illustrate the problem (since the original file written by FEI's EPU software cannot be shared).

Without this PR, ```showinf -nopix -omexml golgi-modified-AGAR.mrc``` should detect the file as being Deltavision and not MRC.  The ```DeltaT```, ```ExposureTime```, and physical size/position values in the OME-XML are highly improbable, which is good indication that the type detection is wrong.

With this PR, the same test should detect the file as MRC, with more reasonable physical sizes of 1um x 1um.  The extended header type (```AGAR```) and size (```0```) should be logged.

The extended header type is used to determine the physical size units, which seems valid based upon:

http://www.ccpem.ac.uk/mrc_format/mrc2014.php#note8
http://www.ccpem.ac.uk/downloads/EPU_user_manual_AppendixCfordistribution.pdf
http://blake.bcm.edu/eman2/doxygen_html/mrcio_8h_source.html#l00236

FEI software can apparently use the ```AGAR``` (old style) or ```FEI1``` (new style) extended type.  Only ```AGAR``` is handled, since we don't have an example of ```FEI1```.  Per the linked EMAN2 documentation, older MRC files generated by FEI software use micrometers for units instead of meters as noted in the specification.

Builds should remain green and memo files should not be affected.  This should be safe for a patch release, but certainly not 5.9.1.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -170,7 +170,7 @@ public class DeltavisionReader extends FormatReader {
     if(new TiffParser(stream).isValidHeader()) {
       return false;
     }
-    final int blockLen = 98;
+    final int blockLen = 212;
     if (!FormatTools.validStream(stream, blockLen, true)) return false;
     stream.seek(96);
     int magic = stream.readShort() & 0xffff;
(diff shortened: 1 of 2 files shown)


melissalinkert:
I haven't been able to find an EMDB sample that reproduces this, but that's probably not surprising given the EMDB distribution format: ftp://ftp.ebi.ac.uk/pub/databases/emdb/doc/Map-format/current/EMDB_map_format.pdf

PR #3193, opened 2018-08-09
Columbus: Add support for PlaneID

dgault:
See https://trello.com/c/1tAKDsuU/261-planeid-support-for-columbus

This PR adds support for the PlaneID field and is in relation to an issue originally raised in forum thread https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8550&p=19850&sid=65eaffd409d3f3b6e7506cf70614c778#p19850

A sample dataset is available in QA-21609. The partial fileset contains 9 planes, each with 3 channels. Currently only a single 3 channel image is opened.

All of the tiff files are picked up by the reader and the plane data is all correctly parsed. However the PlaneID attribute is not used. I have modified the reader to treat PlaneID the same as FieldID, this results in the correct number of series being used. A better longer term fix would be to include this as a separate attribute on plane rather than reusing field.

To read the dataset a MeasurementIndex.ColumbusIDX.xml must be added and the reader has also been modified to handle the partial dataset:

To test:
- Without the PR only the first plane is read
- With the PR all of the planes should be read correctly

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
@@ -28,6 +28,7 @@ package loci.formats.in;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
@@ -478,6 +479,7 @@ public class ColumbusReader extends FormatReader {
       Element image = (Element) images.item(i);
       Plane p = new Plane();
 
+      Location fileLoc = null;
       NodeList children = image.getChildNodes();
       for (int q=0; q<children.getLength(); q++) {
         Node child = children.item(q);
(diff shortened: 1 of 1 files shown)


sbesson:
Unfortunately, as shown by https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/23615/console, the proposed changes cause some regression on our current `columbus` samples (single plates from the idr0019 and idr0019 datasets).

Compared to the current ground truth in IDR (https://idr.openmicroscopy.org/webclient/?show=well-1024940)

screen shot 2018-08-12 at 09 27 01

importing one of the `idr0019` plate with this change into a testing server reveals that only the first field is detected:

screen shot 2018-08-12 at 09 26 05

Excluding for now and we can discuss with @dgault on return the next steps in order to get this fix integrated while staying compatible with other representative datasets.

sbesson:
NB: this is another PR which could benefit from the automated tests enhancements discussed in https://trello.com/c/c9PwJFcK/284-add-configuration-keys-for-hcs-metadata to make sure no regression is introduced in terms of the `SPW` representation.

sbesson:
Rethinking about it, I suspect this metadata field might be similar to the Operetta/Harmony export format where `FieldID` specifies the index of the field of view while `PlaneID` specifies the Z-index.

In all cases, we need a complete representative sample in order to thoroughly test this change and QA 21690 was only a subset. Closing as such and we can revisit if some data requires this change.

PR #3172, opened 2018-06-18
Update the ImageJ API link for Javadoc crosslinks

sbesson:
http://rsbweb.nih.gov/ij/developer/api/ is now redirecting to https://imagej.nih.gov/ij/developer/api/.

The daily Bio-Formats CI job has been unstable recently failing on the first link. Trying to switch to the redirected HTTPS link to see if it helps stabilizing things on our side. /cc @dgault

PR #3197, opened 2018-08-13
Use imagej.net URL for ImageJ Javadoc

sbesson:
See https://github.com/openmicroscopy/bioformats/pull/3172#issuecomment-412448699

A previous switch to imagej.net had been conflating with other issues related to the CI environment. As  the connection to imagej.nih.gov seems to be frequently flaky in the morning, trying to switch again to imagej.net as the canonical location for the ImageJ API.


dgault:
The links look good, if the builds stay green then lets use imagej.net

Issue #3200, opened 2018-08-16
BioFormats crashes when trying to open second slice - bioformats or image problem?


bgriffen:
The images in question can be found [here](https://drive.google.com/drive/folders/19ETt4fdci-H2lkHOflP-L-XvJm9KLtW7?usp=sharing); any work as an example here. 

I recently was applying the [BigStitcher](https://github.com/PreibischLab/BigStitcher) plugin to stitch some tiles together but because that module uses BioFormats, it Fiji was encountering an opening error. The identification of this issue occurred [here](https://github.com/PreibischLab/BigStitcher/issues/16).

The output I get on my (up-to-date) Fiji is as follows:

```
(Fiji Is Just) ImageJ 2.0.0-rc-68/1.52e; Java 1.8.0_66 [64-bit]; Linux 4.15.0-30-generic; 117MB of 230000MB (<1%)
 
java.lang.ArrayIndexOutOfBoundsException
	at java.lang.System.arraycopy(Native Method)
	at loci.formats.tiff.TiffParser.getSamples(TiffParser.java:1087)
	at loci.formats.tiff.TiffParser.getSamples(TiffParser.java:834)
	at loci.formats.in.MinimalTiffReader.openBytes(MinimalTiffReader.java:306)
	at loci.formats.DelegateReader.openBytes(DelegateReader.java:227)
	at loci.formats.ChannelFiller.openBytes(ChannelFiller.java:156)
	at loci.formats.ChannelSeparator.openBytes(ChannelSeparator.java:227)
	at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:348)
	at loci.formats.DimensionSwapper.openBytes(DimensionSwapper.java:249)
	at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:348)
	at loci.formats.MinMaxCalculator.openBytes(MinMaxCalculator.java:269)
	at loci.formats.MinMaxCalculator.openBytes(MinMaxCalculator.java:260)
	at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:334)
	at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:334)
	at loci.plugins.util.ImageProcessorReader.openProcessors(ImageProcessorReader.java:186)
	at loci.plugins.in.ImagePlusReader.readProcessors(ImagePlusReader.java:422)
	at loci.plugins.in.ImagePlusReader.readPlanes(ImagePlusReader.java:387)
[...]

PR #3202, opened 2018-08-20
TIFF: split IFDs into separate series if the dimensions or pixel type mismatch

melissalinkert:
Backported from a private PR.  See #3200.

To test, use the files in ```data_repo/curated/tiff/bgriffen``` (same as linked from #3200).  Without this PR, ```showinf``` on any file should throw an ```ArrayIndexOutOfBoundsException``` and ```SizeT``` should be 2.  With this PR, the same test should show 2 series each with ```SizeT``` of 1.  The image for each series should be readable without an exception.

```tiffinfo``` on each file will confirm that there are two IFDs, both of which have a subfile type of 0 (i.e. neither is flagged as a thumbnail).  There doesn't appear to be any other metadata in the files to indicate that the second image should be ignored, even though it's pretty clearly just a thumbnail.

It's possible that this may introduce some build failures, as impact is difficult to assess given how many readers rely upon ```MinimalTiffReader```.  This can certainly be excluded after a first test run to reduce noise.  Memo files shouldn't be affected, but this is arguably a breaking core metadata change.

https://trac.openmicroscopy.org/ome/ticket/13144 is related, but not completely covered by this PR.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
@@ -97,6 +97,12 @@ public class MinimalTiffReader extends FormatReader {
   /** Merge SubIFDs into the main IFD list. */
   protected transient boolean mergeSubIFDs = false;
 
+  /**
+   * Whether or not IFDs with different dimensions can be
+   * split into separate series.
+   */
+  protected transient boolean canSeparateSeries = true;
+
   /** Number of JPEG 2000 resolution levels. */
   private Integer resolutionLevels;
 
@@ -490,7 +496,7 @@ public class MinimalTiffReader extends FormatReader {
             (int) ifd.getImageLength() != y ||
             ifd.getPixelType() != type)
           {
-            separateSeries = true;
+            separateSeries = canSeparateSeries;
           }
         }
       }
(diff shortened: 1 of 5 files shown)


sbesson:
As suggested by the PR description the changes to `MinimalTiffReader` have some impact on derived formats including [canon](https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/24935/), [hamamatsu](https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/24848/), [nikon](https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/24901/) and [pyramid](https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-folder/24913/). @melissalinkert let us know whether a fix can be easily pushed or excluding/closing this PR is safer for now to review 5.9.2 contributions.

melissalinkert:
Excluding for now - will iterate on the test failures in the background and can re-include after 5.9.2.

melissalinkert:
Last two commits here plus https://github.com/openmicroscopy/data_repo_config/pull/318 should address the issues flagged by https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-repo/161/, so this could be re-included when convenient.

melissalinkert:
Re-included and listed for review - can exclude again if there are any further test failures.

melissalinkert:
One last commit pushed to address the ```hamamatsu``` and ```cellsens``` failures in https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-repo/181/

dgault:
All builds and tests are now green and derived formats appear to be unaffected with the latest commits.

Testing using the linked dataset results in an `ArrayIndexOutOfBoundsException ` without the PR. 
With the PR the series are split and read without exception, the metadata and pixel data also look correct.

PR #3208, opened 2018-08-27
Fix configuration series table retrieval for pyramid detection

melissalinkert:
See https://github.com/openmicroscopy/data_repo_config/pull/321 and https://trello.com/c/q5mPwXHU/37-review-broken-pyramid-readers

The series table names should match the unflattened series/resolution
indexes.  The table data (SizeX, SizeY, physical sizes, ...) corresponds
to the metadata from a reader with flattened resolutions.  This updates
Configuration to allow the flattened-ness to be specified when setting
the active series or getting the series count.  By default, flattened is
true.  The unflattened pixels tests now check that the unflattened
reader's series count matches the configuration's unflattened series
count.

All of this means that readers that correctly report pyramids will have
passing tests, and readers that do not correctly report pyramids will
fail testUnflattened*PixelsHashes.

Assuming no errors in the config PR above, I'd expect to see no test failures for SVS files, and 2 failures per dataset for NDPI files and anything else that should be a pyramid but isn't currently.

### DIFF
--- a/components/test-suite/build.xml
+++ b/components/test-suite/build.xml
@@ -243,6 +243,7 @@ Type "ant -p" for a list of targets.
       <sysproperty key="testng.configDirectory" value="${testng.configDirectory}"/>
       <sysproperty key="testng.cacheDirectory" value="${testng.cacheDirectory}"/>
       <sysproperty key="testng.configSuffix" value="${testng.configSuffix}"/>
+      <sysproperty key="testng.allow-missing" value="${testng.allow-missing}"/>
       <sysproperty key="testng.multiplier" value="${testng.multiplier}"/>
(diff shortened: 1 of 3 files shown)


melissalinkert:
b1f79d3 should fix the long running ```fake``` and ```columbus``` tests that have been causing https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo to run for multiple days.

f16f4ee should fix the test failures due to unconfigured files (since ```BIOFORMATS-test-repo``` already uses ```-Dtestng.allow-missing=true```).

sbesson:
https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/ has been green for the last few days. Merging.

PR #3217, opened 2018-09-07
ZeissTIFFReader cleanup

melissalinkert:
See https://trello.com/c/gJ2nQw9v/29-zeiss-tiff-distributed-failures and https://trello.com/c/FQlFqn62/30-axiovision-duplicate-paths-in-usedfiles

This removes a bunch of unused code from ```ZeissTIFFReader```, replaces use of ```CaseInsensitiveLocation``` with ```Location```, and fixes up the resulting file detection issues.

To test, use ```data_repo/curated/zeiss-tiff/qa-21570```.  Without this PR, ```showinf -nopix 250618_J1_diff_D3-2-0002.tif``` should show a used file list with ```250618_J1_diff_D3-2-0002.tif``` listed twice (as described in https://trello.com/c/FQlFqn62/30-axiovision-duplicate-paths-in-usedfiles).  With this PR, the same test should show a used file list with no duplicate entries.

Builds should remain green, and reverting the change to single-thread tests in https://trello.com/c/gJ2nQw9v/29-zeiss-tiff-distributed-failures should also result in consistently green builds.  I would expect this to impact memo files due to the removed ```slices```, ```timepoints```, and ```channels``` fields, so this is not safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissTIFFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissTIFFReader.java
@@ -535,86 +535,12 @@ public class ZeissTIFFReader extends BaseZeissReader {
   }
   // -- Helpers --
 
-  class Channel
-  {
-    int id;
-
-    Channel(int id)
-    {
-      this.id = id;
-    }
-  }
-
-  HashMap<Integer,Channel> channels = new HashMap<Integer,Channel>();
-
-  Channel getChannel (int id)
-  {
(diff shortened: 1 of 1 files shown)


sbesson:
Reverted the CI workaround to test the `zeiss-tiff` folder single-threaded. It is probabl worth waiting for a couple of builds until we confirm multi-threaded tests successfully pass against sample files of this format.

PR #3218, opened 2018-09-07
Restrict use case for TiffJAIReader in TiffDelegateReader

melissalinkert:
See https://trello.com/c/tEwJO09D/263-misleading-jai-error-message-on-corrupt-tiff

```TiffJAIReader``` should now only be used if ```setLegacy(true)``` is explicitly called or ```openBytes``` throws an ```UnsupportedCompressionException```.  See also https://trac.openmicroscopy.org/ome/ticket/4090

To test, use a corrupt TIFF (e.g. ```/ome/team/mlinkert/corrupt.tiff```) and a TIFF with unsupported compression (e.g. ```data_repo/curated/tiff/public/libtiff/fax2d.tif```).  Without this PR, ```showinf -debug corrupt.tiff``` will throw an exception indicating that JAI is missing.  ```xxd corrupt.tiff``` will confirm that there are only 4 bytes in the file (TIFF endian marker and magic bytes), so there is no way it could be initialized by any TIFF reader.  ```showinf fax2d.tif``` will throw ```UnsupportedCompressionException``` with no indication that JAI could be used to read the file.

With this PR, ```showinf corrupt.tiff``` will throw an ```EOFException``` with no mention of JAI.  ```showinf fax2d.tif``` will throw an exception indicating that JAI is missing during ```openBytes```.  ```showinf -debug fax2d.tif``` will also log the original ```UnsupportedCompressionException```.

Builds should remain green; I would not expect this to affect memo files.  Definitely low priority, though, since this is less about fixing a bug and more about making it less irritating to identify corrupt TIFF issues.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/TiffDelegateReader.java
+++ b/components/formats-bsd/src/loci/formats/in/TiffDelegateReader.java
@@ -32,7 +32,13 @@
 
 package loci.formats.in;
 
+import java.io.IOException;
+import java.util.ArrayList;
+
+import loci.formats.CoreMetadata;
 import loci.formats.DelegateReader;
+import loci.formats.FormatException;
+import loci.formats.UnsupportedCompressionException;
 
 /**
  * TiffDelegateReader is a file format reader for TIFF files.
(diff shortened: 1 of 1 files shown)


mtbc:
:+1: This PR seems good to me and behavior matches the description. Just have to get https://web-proxy.openmicroscopy.org/west-ci/job/BIOFORMATS-test-repo/ green now.

PR #3219, opened 2018-09-10
Add new KLB Reader

dgault:
This reader is for the Keller Lab Block file type and is based on https://bitbucket.org/fernandoamat/keller-lab-block-filetype/overview

Details of the image header and image IO can be found at https://bitbucket.org/fernandoamat/keller-lab-block-filetype/src/8e3c39b75022/src/?at=master

This PR is a follow up to the prototype reader which was being tested in https://github.com/openmicroscopy/bioformats/pull/3150

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/KLBReader.java
+++ b/components/formats-gpl/src/loci/formats/in/KLBReader.java
@@ -255,15 +255,18 @@ public class KLBReader extends FormatReader {
 
           // Location within the block for required XY plane
           int inputOffset = (coordBlock[2] % dims_blockSize[2]) * blockRowSize * blockSizeAux[1];
-          if (coordBlock[1] < y && blockSizeAux[1] != dims_blockSize[1] && blockSizeAux[0] != dims_blockSize[0]) inputOffset += blockSizeAux[0] * (dims_blockSize[1] - blockSizeAux[1]) * bytesPerPixel;
+          if (coordBlock[0] < x && coordBlock[1] < y && blockSizeAux[1] != dims_blockSize[1] && blockSizeAux[0] != dims_blockSize[0]) inputOffset += ((dims_blockSize[0] * (dims_blockSize[1] - blockSizeAux[1])) + (x - coordBlock[0])) * bytesPerPixel;
+          // Partial block at the start of x tile
           else if (coordBlock[0] < x && blockSizeAux[0] != dims_blockSize[0]) inputOffset += (dims_blockSize[0] - blockSizeAux[0]) * bytesPerPixel;
+          // Partial block at the start of y tile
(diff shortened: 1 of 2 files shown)


dgault:
Yeah currently the spec describes the reading of any single KLB file but we dont have a description of the overall file/folder structure for larger datasets.

I agree that long term going forward having a more generic handling of potential filesets would be much more desirable.

I can open a config for say the first 10 or 15 timepoints in the dataset we have as a start to the repo tests.

components/formats-gpl/src/loci/formats/in/KLBReader.java
```diff
@@ -0,0 +1,577 @@
...
+    ByteBuffer bb = ByteBuffer.wrap(b).order(ByteOrder.LITTLE_ENDIAN);
+    return bb.getLong();
+  }
+
+  private PositiveFloat readFloat32() throws IOException {
+    byte[] b = new byte[4];
+    in.read(b, 0, 4);
+    ByteBuffer bb = ByteBuffer.wrap(b).order(ByteOrder.LITTLE_ENDIAN);
+    return new PositiveFloat((double) bb.getFloat());
+  }
```
melissalinkert: Is there a reason not to use the ```readInt```, ```readLong```, and ```readFloat``` methods in ```RandomAccessInputStream``` here?

components/formats-gpl/src/loci/formats/in/KLBReader.java
```diff
@@ -0,0 +1,577 @@
...
+    }
+    String imageID = MetadataTools.createLSID("Image", 0);
+    store.setImageID(imageID, 0);
+    imageID = MetadataTools.createLSID("Image", 1);
+    store.setImageID(imageID, 1);
+    imageID = MetadataTools.createLSID("Image", 2);
+    store.setImageID(imageID, 2);
+    imageID = MetadataTools.createLSID("Image", 3);
+    store.setImageID(imageID, 3);
+    MetadataTools.populatePixels(store, this);
```
melissalinkert: ```populatePixels``` automatically calls ```setImageID``` for each series, so I would expect that lines 386-393 could be removed.

components/formats-gpl/src/loci/formats/in/KLBReader.java
dgault: Yeah, this is because they are using the opposite endianness, I had originally tried to use RandomAccessInputStream for it

dgault:
The last commit implements some of the suggested changes:
- Overrides close to reset values
- Allows for single files or files in the structure of the idr fileset
- For the idr structure the projections will not be the first series
- Allows for easier addition of extra projections or series

Even with those changes it still remains rather heavily specialised for the idr dataset, but as that is our only use case thus far I think its going to be hard to avoid that

components/formats-gpl/src/loci/formats/in/KLBReader.java
melissalinkert: Using https://github.com/ome/ome-common-java/blob/master/src/main/java/loci/common/RandomAccessInputStream.java#L256 as needed should handle that case; it's safe to swap endianness multiple times on an open stream.  See for example https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/GatanReader.java#L683

PR #3221, opened 2018-09-11
FormatReaderTest: refactor reader setup methods

melissalinkert:
See https://github.com/openmicroscopy/bioformats/pull/3184#pullrequestreview-153797483

This just cleans up some duplicate code and documents the ```setupReader(boolean, boolean)``` method that actually constructs a reader.  There should be no change in test behavior, and all builds should continue to pass.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -2699,10 +2699,7 @@ public class FormatReaderTest {
       try {
         ir.setId(id);
       }
-      catch (FormatException e) {
-        LOGGER.error("Could not initialize " + id, e);
-      }
-      catch (IOException e) {
+      catch (FormatException | IOException e) {
         LOGGER.error("Could not initialize " + id, e);
       }
     }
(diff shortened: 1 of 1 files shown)


components/test-suite/src/loci/tests/testng/FormatReaderTest.java
```diff
@@ -2661,20 +2661,55 @@ public void saveFileScanList() {
...
+    ir.setMetadataFiltered(true);
+
+    if (initialize) {
+      try {
+        ir.setId(id);
+      }
+      catch (FormatException e) {
+        LOGGER.error("Could not initialize " + id, e);
+      }
+      catch (IOException e) {
```
mtbc: These days can use multi-catch if you like.

sbesson:
https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/219/ is now passing with this change. Do you want to push an extra commit to  address https://github.com/openmicroscopy/bioformats/pull/3221#discussion_r216933801? Otherwise LGTM

melissalinkert:
Commit pushed; see also https://trello.com/c/kePyWW3S/524-review-places-where-multi-catch-could-be-used (for whenever we have too much time on our hands).

sbesson:
The single failure from https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-repo/220/ is unrelated to this PR. Changes looks good. Merging

PR #3224, opened 2018-09-17
Initial CV7000 reader

melissalinkert:
Backported from a bunch of private PRs.  See https://trello.com/c/b0nFSUzZ/238-cv7000

To test, use 3 plates in ```/uod/idr/repos/curated/cv7000/glencoe/```.  For each, the ```.wpi``` file is the one to initialize (this is the main metadata file).  The ```DW 0001``` plate has 370 wells - standard 16x24 with D07, E12, H14, H21, I14, I24, J09, J18, K18, L17, L18, M02, P10, and P20 missing.  ```DW 0002``` is a single row (22 wells, H13 and H14 missing) and ```DW 0003``` is a single column (16 wells).  All three are single field, single T, single Z, and 4 channels (```DAPI, 488, 594, mitotracker```).  There are many more plates that cover a wider range of dimensions which could be transferred to ```data_repo```, but this requires coordination due to https://trello.com/c/AJf3bdxa/83-the-slow-death-of-ome-datarepo

For each, confirm that ```showinf -nopix -omexml``` shows dimensions that match the description above.  It should be possible to open each series (well), either via ```showinf``` or ImageJ.  The images are 16 bit, and not particularly biologically interesting (all were acquired specifically for testing).

Memo files will be impacted; may conflict with #3219 and/or #3151.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -127,16 +127,14 @@ public class CV7000Reader extends FormatReader {
   /* @see loci.formats.IFormatReader#getUsedFiles(boolean) */
   @Override
   public String[] getUsedFiles(boolean noPixels) {
-    if (noPixels) {
-      HashSet<String> files = new HashSet<String>();
-      for (String file : allFiles) {
-        if (!checkSuffix(file, "tif")) {
-          files.add(file);
-        }
+    ArrayList<String> files = new ArrayList<String>();
+    files.add(new Location(currentId).getAbsolutePath());
+    for (String file : allFiles) {
+      if (!files.contains(file) && (!noPixels || !checkSuffix(file, "tif"))) {
+        files.add(file);
       }
-      return files.toArray(new String[files.size()]);
     }
-    return allFiles;
+    return files.toArray(new String[files.size()]);
   }
 
   /* @see loci.formats.IFormatReader#getSeriesUsedFiles(boolean) */
@@ -145,7 +143,7 @@ public class CV7000Reader extends FormatReader {
     FormatTools.assertId(currentId, true, 1);
 
     HashSet<String> files = new HashSet<String>();
-    files.add(currentId);
+    files.add(new Location(currentId).getAbsolutePath());
     files.add(measurementPath);
     if (detailPath != null) {
       files.add(detailPath);
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -150,7 +150,9 @@ public class CV7000Reader extends FormatReader {
     }
     if (!noPixels && channels != null) {
       for (Channel c : channels) {
-        if (c != null && c.correctionFile != null) {
+        if (c != null && c.correctionFile != null &&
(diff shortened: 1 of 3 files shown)


components/formats-gpl/src/loci/formats/in/CV7000Reader.java
```diff
@@ -0,0 +1,900 @@
...
+            }
+          }
+        }
+
+        for (int p=0; p<getImageCount(); p++) {
+          Plane plane = lookupPlane(i, p);
+          if (plane == null) {
+            continue;
+          }
+          store.setPlanePositionX(FormatTools.createLength(plane.xpos, UNITS.REFERENCEFRAME), i, p);
```
sbesson: Similarly to #3244, we might want to fill the corresponding `WellSample.Position` metadata with the same values when possible. This can happen as a follow-up PR.

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
```diff
@@ -0,0 +1,900 @@
...
+                break;
+              }
+            }
+            if (channel == null) {
+              continue;
+            }
+
+            if (c == 0) {
+              store.setPixelsPhysicalSizeX(FormatTools.getPhysicalSizeX(channel.xSize), i);
+              store.setPixelsPhysicalSizeY(FormatTools.getPhysicalSizeY(channel.xSize), i);
```
mtbc: Should this be `channel.ySize`?

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
```diff
@@ -0,0 +1,900 @@
...
+        LightSource l = new LightSource();
+        l.name = attributes.getValue("bts:Name");
+        l.type = attributes.getValue("bts:Type");
+
+        String wavelength = attributes.getValue("bts:WaveLength");
+        String power = attributes.getValue("bts:Power");
+
+        if (wavelength != null) {
+          try {
+            l.wavelength = new Double(wavelength);
```
mtbc: Could we use `Double.parseDouble` for these?

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
melissalinkert: Good catch - fixed in e81f793.

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
melissalinkert: a49b94b switches all usage of ```new Double``` and ```Double.parseDouble``` to ```DataTools.parseDouble```.

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
melissalinkert: Noted for follow-up PR: https://trello.com/c/b0nFSUzZ/49-cv7000#comment-5bc498c81ae2b885ae77c111

sbesson:
Thanks @melissalinkert. Once we sort the build issue and this PR is successfully tested against the existing `cv7000` datasets, happy to see this merged as a first iteration on this reader.

sbesson:
Merging as per https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/37870/console.

PR #3228, opened 2018-09-20
Relax companion

sbesson:
See https://github.com/openmicroscopy/bioformats/pull/2867 for the original PR

This PR re-opens the investigation of the OME-TIFF reader change necessary to support relaxed multi-file companion filesets composed of:

- a top-level master companion file containing the OME-XML metadata i.e. `<filename>.companion.ome`
- multiple TIFF files referenced from the top-level companion file via the `TiffData` element

The main change from the current behavior is that the back-referencing to the companion file from each individual TIFF is no longer mandatory.


components/test-suite/.ant-targets-build.xml
```diff
@@ -0,0 +1,64 @@
+bundle
```
rleigh-codelibre: What's this file used for?  It's not in the original PR, and it doesn't appear to be used in this one either.

sbesson: Unexpected addition while fixing conflicts on my commit cherry-pick. Thanks for spotting promptly.

dgault:
Tested this with a variation of the tubhistwt-4D public sample data and it appears to working well.

Sample data was generated by doing the following:
- Converted each of the ome.tif files to .tif using bfconvert -nogroup
- Exported the OME-XML of the original to a companion file
- Edited the companion to change the ome.tif files to .tif

Looks good to merge

PR #3229, opened 2018-09-20
Micromanager: better handling of very large *metadata.txt files

melissalinkert:
Backported from a private PR.  Originally part of the same PR as https://github.com/ome/ome-common-java/pull/24 (prior to repo split).

Original use case was larger (~7000 plane) datasets acquired using the diSPIM plugin (https://micro-manager.org/wiki/ASIdiSPIM_Plugin).  I wasn't able to find a way to run a diSPIM acquisition without a connected instrument, so the closest test data is a 5000 T x 5 Z x 2 C multi-D acquisition with the relevant ```SPIMmode``` key added manually.

Complete dataset yet to be uploaded, so not ready for final testing.  Once uploaded, ```showinf -nopix 1.4-large-acq-take-2_1_MMStack_Pos0_metadata.txt``` without this PR or https://github.com/ome/ome-common-java/pull/24 will require ```BF_MAX_MEM``` to be at least 1200 MB.  With a build including both PRs, only 900 MB should be required.  As ```1.4-large-acq-take-2_1_MMStack_Pos0_metadata.txt``` is 195 MB, a warning should be logged indicating that the file may require additional memory.

Tests should continue to pass with existing Micromanager data; memo files will be affected.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -36,7 +36,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.StringTokenizer;
 import java.util.Vector;
 
 import loci.common.DataTools;
@@ -100,6 +99,9 @@ public class MicromanagerReader extends FormatReader {
   private MinimalTiffReader tiffReader;
 
   private Vector<Position> positions;
+  private int start = 0;
+
+  private boolean spim = false;
 
   // -- Constructor --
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
Dataset now under ```curated```, see https://github.com/openmicroscopy/data_repo_config/pull/333

dgault:
Builds and tests have been unaffected by this PR. Using the suggested test files without this PR required `BF_MAX_MEM` of 1200 MB. With this PR included it required 900 MB and a warning was logged with regard the requirement for extra memory.

PR is good to merge.

sbesson:
Does this fix require ome/ome-common-java#24? In which case should we merge the PR, cut a patch release of `org.openmicroscopy:ome-common` cut and bump it in this repo?

PR #3230, opened 2018-09-21
ZVI: reuse stream for sequential calls to openBytes on the same plane

melissalinkert:
See https://trello.com/c/rgUGN9Ll/187-zvi-windows-performance-issuefiji

This reuses each plane's ```RandomAccessInputStream``` when possible, so that the stream is not reopened many times when the same plane (or tiles from the same plane) are being read.  Opening streams is more expensive than usual for ZVI files, as it typically requires assembling many non-contiguous blocks from the internal filesystem-like POI structure.

To test, use the ```test.zvi``` file from QA 21711.  Without this PR, open the file in ImageJ and verify that it takes several minutes for images to be displayed.  Restart ImageJ and open the file again; the performance should be similar (i.e. OS and disk-level caching have a negligible effect).

With this PR, the same test should result in a noticeably faster opening time.  ```showinf -minmax test.zvi``` should be similarly faster.  On my machine, the original time was >15 minutes, and the new time was 2 minutes.

All tests should continue to pass; memo files should not be affected.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissZVIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissZVIReader.java
@@ -65,6 +65,8 @@ public class ZeissZVIReader extends BaseZeissReader {
 
   protected transient POIService poi;
   protected String[] files;
+  protected transient RandomAccessInputStream currentPlane;
+  protected transient int currentPlaneIndex = -1;
 
   // -- Constructor --
 
@@ -123,21 +125,27 @@ public class ZeissZVIReader extends BaseZeissReader {
       return buf;
     }
 
-    RandomAccessInputStream s = poi.getDocumentStream(imageFiles[index]);
-    s.seek(offsets[index]);
+    if (currentPlane == null || currentPlaneIndex != index) {
+      if (currentPlane != null) {
+        currentPlane.close();
+      }
+      currentPlane = poi.getDocumentStream(imageFiles[index]);
+      currentPlaneIndex = index;
+    }
+    currentPlane.seek(offsets[index]);
 
     int len = w * pixel;
     int row = getSizeX() * pixel;
 
     if (isJPEG) {
-      byte[] t = new JPEGCodec().decompress(s, options);
+      byte[] t = new JPEGCodec().decompress(currentPlane, options);
 
       for (int yy=0; yy<h; yy++) {
         System.arraycopy(t, (yy + y) * row + x * pixel, buf, yy*len, len);
       }
     }
     else if (isZlib) {
-      byte[] t = new ZlibCodec().decompress(s, options);
+      byte[] t = new ZlibCodec().decompress(currentPlane, options);
       for (int yy=0; yy<h; yy++) {
         int src = (yy + y) * row + x * pixel;
         int dest = yy * len;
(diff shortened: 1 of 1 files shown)


strife-cloud:
Hi @melissalinkert (it was not possible to comment on the trello bioformat 6.x.x)

So I'm working with Alban Favre (the person who trace the bug) and he wanted to know if it's possible to get the patch so he can resume his project.
So I don't know exactly how I have to do this but if I understand correctly, I have to clone the following project : 
https://github.com/openmicroscopy/bioformats
Get the jar en replace the old jar with the newest with fix bug, right?

Thanks for you help,

dgault:
Tested with a number of ZVI sample files both opening in ImageJ and using `showinf -minmax`, with the PR included there is a very noticeable performance improvement.

The code itself behaves as before but without the overhead of creating a new stream multiple times for the same plane. The streams are also properly cleaned up upon close of the reader. PR is good to merge.

strife-cloud:
Hi David,

I was wondering when the update will be up because I'm using it and I'm
currently blocked because of the bug

Have a nice day,
Danh

Le lun. 29 oct. 2018 à 17:05, David Gault <notifications@github.com> a
écrit :

sbesson:
Hi @strife-cloud. This patch will be included in a development milestone on the Bio-Formats 6 roadmap. This will not be a production release however we will be distributing it to a few beta testers to get some feedback on our ongoing OME-TIFF pyramidal work.

Are you consuming Bio-Formats via ImageJ or Fiji? In the former case, we can paste links to the JARs to download. In the latter case, we have been considering starting to distribute these development versions via the Bio-Formats update site.

strife-cloud:
Hi Sebastien,

I'm currently using Fiji so should I wait that these development versions
being available on the site?

Have a nice day,
Danh


Le jeu. 22 nov. 2018 à 10:02, Sébastien Besson <notifications@github.com> a
écrit :

sbesson:
@strife-cloud see [our blog post](https://blog.openmicroscopy.org/file-formats/community/2018/11/29/ometiffpyramid/) and [our announcement on the image.sc forum](https://forum.image.sc/t/bio-formats-6-update/21223) for an update of Bio-Formats 6. We expect to start pushing milestones to the update site soon but have not done so so far. 
Before we do so, if you want to give the latest milestone a first try, you can [download the individual JARs](https://downloads.openmicroscopy.org/bio-formats/6.0.0-m3/) and manually update your distribution as described in [this page](https://docs.openmicroscopy.org/bio-formats/6.0.0-m3/users/fiji/index.html#manual-upgrade).

strife-cloud:
Hi Sebastien,

I'm reviving you if it is possible to get development version, it's kind of
important so I can advance on my project.

Have a nice day,
Danh

sbesson:
@Alban554 sorry for the delay. You are right some of these JARs are not part of the Bio-Formats downloads but low-level dependencies. As an alternative, could you try replacing all the JARs above and put the `bioformats_package.jar` available from  https://downloads.openmicroscopy.org/bio-formats/6.0.0-m3/artifacts/ instead?

sbesson:
@Alban554 @strife-cloud with the release of [Bio-Formats 6.0.0](https://www.openmicroscopy.org/2019/02/18/bio-formats-6-0-0.html), the Bio-Formats version distributed via the Java-8 update site  has been updated and should now include the fix for this particular issue. Please try updating your Fiji distribution and let us know if the problem persists by opening a new issue.

PR #3231, opened 2018-09-24
CZI original metadata reduction

melissalinkert:
Backported from a private PR.

To test, use any pyramid .czi file, e.g. ```curated/zeiss-czi/zeiss/jpeg-xr/a)BF-1scene-nozstack-compression85%.czi```.  Without this PR, examine the original metadata for series 0 with ```showinf -nopix```.  There should be hundreds of keys such as ```Positions|Series 03| #041: 0```, one per tile in each pyramid resolution.  All values for a given series should be equal, and the table for a single series contains the keys for every series.

With this PR, the same test should show a single key in the series original metadata table; nothing else should change.

74d5dac fixes a bug in ```CoreMetadata``` that causes the original metadata tables to be the same for every series.  21f4525 follows on from https://github.com/openmicroscopy/bioformats/pull/2600#issuecomment-252793637 and only sets a single position value in the original metadata table when reading a pyramid file.  Together the two commits should shrink the size of the combined original metadata tables, which reduces ```setId``` time particularly when a non-```DummyMetadata``` is used (e.g. OMERO import).

Memo files should be unaffected and tests should continue to pass.

### DIFF
--- a/components/formats-api/src/loci/formats/CoreMetadata.java
+++ b/components/formats-api/src/loci/formats/CoreMetadata.java
@@ -168,7 +168,8 @@ public class CoreMetadata implements Cloneable {
     indexed = r.isIndexed();
     falseColor = r.isFalseColor();
     metadataComplete = r.isMetadataComplete();
-    seriesMetadata = r.getSeriesMetadata();
+    seriesMetadata = new Hashtable<String, Object>();
+    seriesMetadata.putAll(r.getSeriesMetadata());
     thumbnail = r.isThumbnailSeries();
     resolutionCount = r.getResolutionCount();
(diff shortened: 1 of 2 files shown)


dgault:
The fix in https://github.com/openmicroscopy/bioformats/commit/74d5dac78d198f90b753472a7b838ee9c8737773 to prevent duplication across series is pretty important and will impact a lot of readers. The new behaviour is correct and would be good to have in as part of 6.0.0

For the CZI pyramid changes, removing the duplication for each tile makes sense in this case. Testing with the recommended sample shows that with this PR only a single entry is tored as opposed to the original list of values.

Issue #2811, opened 2017-04-05
SCN to JPEG transform is failling


YOUNGING:
I'm using bfconvert command to transform Leica SCN images, the original image is like this (using ImageScope to view)
![image](https://cloud.githubusercontent.com/assets/12063725/24700650/643aa00c-1a2a-11e7-8fa4-f54c1f6873bb.png)

but the output image is not what I'm expecting. All the white background become pink, and the tissue become black and blur.

![image](https://cloud.githubusercontent.com/assets/12063725/24700445/c7a8aac2-1a29-11e7-83c9-adaf12bcd6c9.png)

Can anyone tell me what is the problem here? Thanks a lot.

dgault:
Hi @YOUNGING,

I was able to run some tests on existing sample files we have and could not reproduce the issue with the latest Bio-Formats 5.4.0. 

Can you tell us which version you are currently using `bfconvert -version`
Also would it be possible to upload a sample file which produces this issue to the link below:
https://www.openmicroscopy.org/qa2/qa/upload/

YOUNGING:
Thanks for your help. I update the bftools  to the latest version (5.4.0), but the issue is still there. I have uploaded one sample file to that link. The filename is 132223-12_022a1963-b32e-467f-8242-b51a5295951f.SCN.

dgault:
Thank you for submitting the sample file (QA-17671), I was able to reproduce the issue with the file provided.

This looks like it is potentially a bug in the LeicaSCNReader, comparing the imported image to that when using ImageScope the metadata being read appears to match the expected values however the pixel counts are higher than expected and are being capped at 255 (max value for uint8 data). 

This will require further investigation and debugging on our part.

YOUNGING:
@dgault 

I notice that Bio-formats have openslide as a component. Is it possible to use openslide as the default reader for SCN files?

I'm new to Java. If there is anything else I can do to accelerate the debug process, please tell me.

dgault:
@YOUNGING  Apologies for the delayed response, I was off at the end of last week. I don't believe the issue is related to  the `dissimilar main images` and we are not directly using openslide for this reader.

The main thing that would be of help to us would be to have more sample files. Have you noticed this issue occurring with a large number of images? Have you noticed any particular settings which cause images to display this problem?

YOUNGING:
I have more than 100 images of this format. They all display this problem.

dgault:
[...]

PR #3240, opened 2018-10-02
Leica SCN: improvements for Versa datasets

melissalinkert:
Backported from recent private PR.  Coincidentally addresses at least part of #2811.

To test, use the file from QA 17671.  Without this PR, ```showinf -nopix -noflat``` should show 2 pyramids and no extra images.  ```showinf``` on any of the images should result in a weird pinkish image as documented in #2811.

With this PR, ```showinf -nopix -noflat``` should show 2 pyramids and an extra label image in series 2.  ```showinf``` on any of the images should show normal looking slide images with a white instead of pink background.  Leica/Aperio ImageScope can be used to verify that the images are valid.  Note that the slides are rotated 90 degrees from how they are presented in ImageScope; I don't see any obvious rotation metadata, so the reader doesn't deviate from the X and Y dimensions stored in the file.

Tests on all other SCN files should continue to pass.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
@@ -452,7 +452,17 @@ public class LeicaSCNReader extends BaseTiffReader {
           store.setPlanePositionY(offsetY, pos, q);
         }
 
-        store.setImageName(i.name + " (R" + subresolution + ")", pos);
+        if (hasFlattenedResolutions()) {
+          store.setImageName(i.name + " (R" + subresolution + ")", pos);
+        }
+        else {
+          if (ms.resolutionCount > 1) {
+            store.setImageName("", pos);
+          }
+          else {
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
@melissalinkert If it's not in the XML metadata, the rotation isn't in the `ORIENTATION` TIFF tag is it?  Or a separate EXIF IFD?

melissalinkert:
Yes, I checked for the presence of an ```ORIENTATION``` tag and ```EXIF``` IFD; neither are present in this case.

sbesson:

dgault:
The code changes look good and the PR builds and passes tests. It does not appear to impact existing datasets.

Tested with QA-17671, without the PR there are 2 pyramids (each with 6 resolutions) and no label. Opening one of the images shows the pinkish hue as described.

With the PR included the 2 pyramids are present along with a label image, displaying the images shows a normal white slide as expected.

PR is good to merge.

PR #3243, opened 2018-10-09
ChannelSeparator: copy all channel metadata instead of just names

melissalinkert:
Backported from unmerged private PR.  Uses https://github.com/ome/ome-model/pull/88.

To test, use a file with RGB data and more OME-XML Channel metadata than just a name.  I used ```curated/zeiss-czi/silke/White_574.czi```, but brightfield datasets in most of the WSI formats would work too.  Without this change, ```showinf -nopix -noflat -omexml -separate``` should show one ```Channel``` per original RGB component.  The names of all ```Channel```s in an ```Image``` should be the same, but only the first will have things like wavelengths, detector references, etc.

With this PR, ```showinf -nopix -noflat -omexml -separate``` should show that all ```Channel```s in an ```Image``` are identical apart from the ID.

This could impact OMERO imports, so it probably makes sense to also test a few imports of RGB data to make sure that nothing goes wrong.  I would expect builds to remain green, and I don't think this will affect memo files.

### DIFF
--- a/components/formats-bsd/src/loci/formats/ChannelSeparator.java
+++ b/components/formats-bsd/src/loci/formats/ChannelSeparator.java
@@ -321,7 +321,7 @@ public class ChannelSeparator extends ReaderWrapper {
             pixelsPopulated = true;
           }
           for (int i=1; i<rgbChannels; i++) {
-            MetadataConverter.convertChannelMetadata(retrieve, s, cIndex,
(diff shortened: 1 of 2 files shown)


sbesson:
@melissalinkert can you also bump the version of `ome-model` used so that this passes Travis and gets included into the daily builds?

PR #3244, opened 2018-10-11
OperettaReader: also populate WellSample positions if the metadata exists

sbesson:
See https://trello.com/c/WrSJs57r/197-improve-operetta-harmony-metadata#comment-5bbf50de8977243c9f33da8d

Currently this metadata is stored at the individual plane level. Similarly to other HCS reader, we are storing the X and Y positions of the first plane at the well sample level as well if this information exists.

This PR should not cause regression in the automated tests. The outcome of the population can be tested:
- either by looking at the OME-XML when reading a sample Operetta plate and verifying the `WellSample` elements have their positions populated
- or by importing a sample plate into an OMERO server with this PR and checking a grid is displayed in the Web client using the relative positions of the individual well samples.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -619,7 +619,7 @@ public class OperettaReader extends FormatReader {
         else if ("PositionY".equals(currentName)) {
           // position stored in meters
           final double meters = Double.parseDouble(value);
-          activePlane.positionY = new Length(meters, UNITS.METRE);
+          activePlane.positionY = new Length(-meters, UNITS.METRE);
         }
         else if ("AbsPositionZ".equals(currentName)) {
           // position stored in meters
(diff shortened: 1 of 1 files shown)


sbesson:
Commit 6318f9d comes from an ongoing investigation with @olatarkowska on a dataset exported by the Harmony software. Opening the dataset with `TileStitcher` with the ImageJ plugin seems to indicate the Y-axis parsed from the XML file is inverted compared to the expectation (leading to tiles organized in the wrong vertical order)

I am expecting this last commit will have some breaking effects on our existing `Operetta` sample datasets, we should review whether the behavior above applies to all datasets or only some of them.

melissalinkert:
I think we can generalize the position inversions by parsing the ```<OrientationMatrix>[[1,0,0],[0,-1,0],[0,0,1]]</OrientationMatrix>``` from each ```Image``` in the ```Index.idx.xml```.  This only seems to present in Harmony-type datasets; the non-Harmony ```curated/perkinelmer-operetta/hms/``` dataset at least don't have this value, so wouldn't be affected.

sbesson:

sbesson:
Closing in favor of #3257

PR #3246, opened 2018-10-11
Speed up TiffSaver.writeImage with bulk array writes

popiel:
The writeImage method in TiffSaver does a byte-by-byte copy from the input buffer to the strip buffers, often in non-sequential order (because of interleaving).  I replace this with bulk array copies if the buffers are determined to be of equivalent widths and bit depths.

This work was motivated by the desire to speed up image format conversions for image files in excess of 50GB. In conjunction with additional changes in ome-codecs and ome-common-java, I have achieved a geometric to linear time asymptotic improvement, plus another ~30% speed improvement of just the linear time cost. The patch for ome-codecs and ome-common-java have already been submitted.

Separately, I am in the process of re-working writeIFD to consolidate buffering and file writes, but that isn't fully baked, yet.  Additional pull requests may be forthcoming when the baking is done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tools/*.jar
 *.xpr
 *.class
 *.iml
+.*.swp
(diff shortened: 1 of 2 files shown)


sbesson:

PR #3247, opened 2018-10-12
buildFiles IOException when listing remote files

manics:
Proposed changes in loci.common.Location to support remote files may result in IOExceptions
- See https://github.com/ome/ome-common-java/pull/33

### DIFF
--- a/components/formats-bsd/src/loci/formats/FilePattern.java
+++ b/components/formats-bsd/src/loci/formats/FilePattern.java
@@ -34,6 +34,7 @@ package loci.formats;
 
 import java.io.File;
 import java.math.BigInteger;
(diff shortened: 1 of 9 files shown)


manics:
And now travis fails because this isn't using the modified ome-common-java library:
```
[ERROR] /home/travis/build/openmicroscopy/bioformats/components/formats-gpl/src/loci/formats/in/BDReader.java:[123,5] exception java.io.IOException is never thrown in body of corresponding try statement
```

manics:
https://github.com/ome/ome-common-java/pull/33/commits/6258c23964a00d4d4c46fe918865eaa08cbcf334 throws `UncheckedIOException` to avoid changing the API.

PR #3249, opened 2018-10-18
Add link to reader-guide documentation

manics:
`readers.txt` says
> Please do not edit unless you know what you are doing (see reader-guide.txt).

It's not clear where `reader-guide.txt` is. My best guess is https://docs.openmicroscopy.org/latest/bio-formats/developers/reader-guide.html but that still doesn't explain how readers.txt works. Is there another document?


melissalinkert:
That's the correct document.  The eighth point under https://docs.openmicroscopy.org/bio-formats/5.9.2/developers/reader-guide.html#other-useful-things briefly describes ```readers.txt```.

sbesson:

PR #3256, opened 2018-10-23
KLB: Allow for flexible channel naming

dgault:
This is a follow up to the original KLB PR which addresses the issue discussed in https://github.com/IDR/idr0044-mcdole/issues/1


components/formats-gpl/src/loci/formats/in/KLBReader.java
```diff
@@ -314,12 +316,17 @@ protected void initFile(String id) throws FormatException, IOException {
...
-          if (channelNum == sizeC) {
-            sizeC = channelNum + 1;
+          if (!channels.contains(channelNum)) {
+            channels.add(channelNum);
           }
         }
       }
 
+      if (channels.size() > 0) {
+        sizeC =channels.size();
```
mtbc: missing space

PR #3258, opened 2018-10-31
Data test updates

melissalinkert:
Fixes recent test issues noted in https://trello.com/c/9MJdfI8N/48-add-test-for-duplicate-used-files, https://trello.com/c/0MB0a98B/66-fix-test-config-to-correctly-generate-new-pyramid-configurations, and https://github.com/openmicroscopy/data_repo_config/pull/331

93a830f should make ```test-config``` work as expected with pyramid files in all formats.  To test, run ```test-config``` and then ```test-automated``` on a directory containing pyramid files in any format (I used a local copy of ```curated/svs/alexandra```) and verify that the tests fail.  Remove the generated ```.bioformats``` file(s), then run the same commands again and verify that the tests pass.

0019a42 and 9286409 from @sbesson add unit tests for ```ConfigurationTree.relocate*``` and e869843 fixes ```ConfigurationTree```, which should prevent the weird test failures that occurred when running ```test-automated``` when ```.bioformats``` files were present in the ```testng.directory```.  ```cd components/test-suite && mvn``` with just 0019a42 and 9286409 should result in 2 test failures.  The same test with the whole PR should result in no test failures.

09d3252 updates ```testSaneUsedFiles``` to fail if there are any duplicate entries in the used files list.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -38,6 +38,7 @@ import loci.common.IniParser;
 import loci.common.IniTable;
 import loci.common.IniWriter;
 import loci.common.Location;
+import loci.formats.FileStitcher;
 import loci.formats.FormatException;
 import loci.formats.FormatTools;
 import loci.formats.IFormatReader;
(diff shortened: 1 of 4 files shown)


sbesson:
Obviously, 0019a42  assumes UNIX-based systems which explains the false positive failures in Appveyor. Either we want to skip these tests on Windows systems for now or we might want to make them system-agnostic.

sbesson:
Re the duplicate tests, this PR certainly demonstrates its effectiveness as https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/38458/ reported duplicate files for one of the Harmony datasets that can be confirmed by running `showinf`.

dgault:
For the `Configuration` changes the use of unflattenedReader matches that of the tests and testing with a couple of different pyramid filesets shows that the generation of config looks good and the `test-automated` runs and passes as expected. Without the `Configuration` changes the tests fail the `testUnflattenedPixelsHashes` and `testUnflattenedSubimagePixelsHashes` tests.

Testing against non pyramid sample files does not show any changes to the generated config files, which is as expected.

PR #3259, opened 2018-11-01
ImageConverter: Remove plates when series is chosen

joshmoore:
Prevents warnings of the form:

```
ome.xml.model.WellSample@557caf28 reference to Image:1 missing from object hierarchy.
```

Files converted without this change cannot be imported into OMERO.


joshmoore:
Note: the test addition here would benefit from having `WARN` states fail the run.

PR #3260, opened 2018-11-07
Fix some plugin OME URLs.

mtbc:
The `libraries.txt` file was pointing to some broken legacy URLs. I guess that link checking doesn't cover these.


dgault:
Good spot, the new links are all working and correct. PR is good to merge.

PR #3261, opened 2018-11-12
Fixes for writing TIFF files in memory

melissalinkert:
Backported from a private PR.

b65d8c5 adds a unit test that will save TIFFs in memory with various combinations of endianness, compression, and dimensions.  The test setup is identical to the existing test for writing TIFFs to disk.
With b65d8c5 alone, ```ant test``` or ```mvn``` in ```components/formats-bsd``` should fail.

0c14653 fixes all test failures.  The change to ```TiffWriter``` resolves an issue with writing a single little endian plane; the remaining changes are required for writing multiple planes in memory.  With both commits, ```ant test``` or ```mvn``` in ```components/formats-bsd``` should pass.

Note that only a subset of the tests are run by default; setting ```testng.runWriterSaveBytesTests``` to 100 will run everything.  I did do this locally and all tests passed, but it might not be a bad idea to double-check during review.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/TiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/TiffWriter.java
@@ -307,6 +307,7 @@ public class TiffWriter extends FormatWriter {
         initialized[series][no] = true;
 
         RandomAccessInputStream tmp = createInputStream();
+        tmp.order(littleEndian);
         if (tmp.length() == 0) {
           synchronized (this) {
             // write TIFF header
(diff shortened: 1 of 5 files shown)


dgault:
Running the TiffWriterTests with https://github.com/openmicroscopy/bioformats/commit/b65d8c545c710b7b4ee8d126fcf9417628271f68 but without the remainder of the PR results in a number of failures for testSaveBytesInMemory:
`java.io.EOFException: Attempting to read beyond end of file.`

Running it with the full PR results in all tests passing as expected. PR is good to merge.

PR #3262, opened 2018-11-12
OME-TIFF: rework how missing planes in TiffData are handled

melissalinkert:
Backported from a private PR.

Without this PR, if any plane was not covered by a ```TiffData```, then all ```TiffData``` information was thrown out and the plane count for the current ```Image``` was reset to the IFD count of the current file.  That's not unreasonable for single file datasets, but when multiple files are present this can be problematic.  ece2e8d describes the original use case as MicroManager-acquired data, where the datasets may contain multiple files that are split at 4GB so a single file does not neatly map to a single ```Image```, Z section, timepoint, etc.

A simpler example that demonstrates the problem is in ```curated/ome-tiff/samples/missing-planes```.  This was generated via ```bfconvert -option ometiff.companion test.companion.ome test&sizeZ=10.fake test-Z&z.ome.tiff```; the ```TiffData``` corresponding to ```test-Z5.ome.tiff``` (and ```test-Z5.ome.tiff``` itself) were then removed by hand.

Without this PR, ```showinf test-Z0.ome.tiff``` should result in a single plane.  With this PR, the same test should result in 10 planes, with Z=5 blank.

I would expect all tests to continue to pass without config changes.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1321,6 +1321,7 @@ public class OMETiffReader extends SubResolutionFormatReader {
         }
         core.add(s, c);
       }
+      rs.close();
     }
     core.reorder();
 
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1004,14 +1004,27 @@ public class OMETiffReader extends SubResolutionFormatReader {
       OMETiffCoreMetadata m = (OMETiffCoreMetadata) core.get(s, 0);
       info[s] = planes;
       try {
-        RandomAccessInputStream testFile = new RandomAccessInputStream(info[s][0].id, 16);
+        RandomAccessInputStream testFile = null;
+        if (info[s][0].id != null) {
+          testFile = new RandomAccessInputStream(info[s][0].id, 16);
+        }
+        if (info[s][0].reader == null) {
+          info[s][0].reader = new MinimalTiffReader();
+        }
         String firstFile = info[s][0].id;
-        if (!info[s][0].reader.isThisType(testFile)) {
+        if (firstFile == null ||
+          (testFile != null && !info[s][0].reader.isThisType(testFile)))
+        {
           LOGGER.warn("{} is not a valid OME-TIFF", info[s][0].id);
           info[s][0].id = currentId;
           info[s][0].exists = false;
+          if (info[s][0].ifd < 0) {
+            info[s][0].ifd = 0;
+          }
         }
-        testFile.close();
+        if (testFile != null) {
+          testFile.close();
+        }
         for (int plane=1; plane<info[s].length; plane++) {
           if (info[s][plane].id == null || info[s][plane].reader == null) {
             continue;
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1003,8 +1003,8 @@ public class OMETiffReader extends SubResolutionFormatReader {
       // populate core metadata
       OMETiffCoreMetadata m = (OMETiffCoreMetadata) core.get(s, 0);
       info[s] = planes;
+      RandomAccessInputStream testFile = null;
       try {
-        RandomAccessInputStream testFile = null;
         if (info[s][0].id != null) {
           testFile = new RandomAccessInputStream(info[s][0].id, 16);
         }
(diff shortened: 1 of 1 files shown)


melissalinkert:
1be0176 should fix the test failure in https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/39997

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
```diff
@@ -933,7 +937,7 @@ else if (t == 0) {
           planes[no].ifd = ifd + q;
           planes[no].certain = true;
           planes[no].exists = exists;
-          LOGGER.debug("      Plane[{}]: file={}, IFD={}",
+          LOGGER.trace("      Plane[{}]: file={}, IFD={}",
```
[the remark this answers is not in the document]
joshmoore: Hi @rleigh-codelibre. Independent of the decision on the severity level for the condition you're talking about, I'd warn *against* raising many lines of logging at INFO or higher. But happy to see a single summary of the problem at a higher level with "turn on tracing for X for more information".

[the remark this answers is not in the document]
melissalinkert: The logging here is not indicative of invalid input data.  It is logging of internal state that happens every time a ```TiffData``` is parsed, even for 100% valid data.  ```showinf -nopix -debug``` without this PR on anything in http://downloads.openmicroscopy.org/images/OME-TIFF can confirm this.  For datasets with many planes, this results in log files that are so large as to be nearly unusable, as explained in the relevant commit message: https://github.com/openmicroscopy/bioformats/pull/3262/commits/f2b667c388bf3fbfc40484e333425a193da45dea

I strongly disagree that the level here should be changed to ```INFO```, but if further review indicates that ```DEBUG``` is still appropriate then I would not be opposed to reverting f2b667c.

The logging of actually invalid data occurs on lines 977 and 984, and is set to ```WARN```.

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
```diff
@@ -964,45 +968,67 @@ else if (t == 0) {
...
+        try {
+          planes = new OMETiffPlane[r.getImageCount()];
+          for (int plane=0; plane<planes.length; plane++) {
+            planes[plane] = new OMETiffPlane();
+            planes[plane].id = currentId;
+            planes[plane].reader = r;
+            planes[plane].ifd = plane;
           }
+          num = planes.length;
+        }
```
rleigh-codelibre: Might be helpful to rename `num` to `planeCount` for clarity.

Also, this fallback (including before this PR) doesn't cater for the case where the plane count in the TIFF file is less than the image count dictated by the OME metadata.  In this case, we're liable to get an ArrayOutOfBounds exception when we try to access a valid plane index outside the `planes` sizes.  Maybe if `num != r.getImageCount()` we should bail out properly.  And also bail when the series count is greater than 1, because we have no hope of mapping higher series to the correct IFD.  At some level, when the TiffData is invalid or missing, we should not try to work around the brokenness, but error properly.

components/formats-bsd/src/loci/formats/in/OMETiffReader.java
```diff
@@ -1308,6 +1339,7 @@ private void addSubResolutions() throws IOException, FormatException {
         }
         core.add(s, c);
       }
+      rs.close();
```
dgault: This line is currently conflicting with changes in https://github.com/openmicroscopy/bioformats/pull/3276

dgault: This line currently conflicts with changes added in https://github.com/openmicroscopy/bioformats/pull/3276/files#diff-b6bbe95ea3520ea325f6313eb1cc47a7R1270

melissalinkert: I'm happy for this PR to be temporarily excluded in favor of #3276, if that makes things easier.

jburel: @melissalinkert I am going to exclude temporarily this PR to test the build

melissalinkert:
Conflicts fixed now that #3277 is merged.  Should be ready to re-include when there is review bandwidth.

dgault:
This handling of missing TiffData certainly provides a better solution to what is currently present.

I have tested the PR by reproducing the steps to generate `curated/ome-tiff/samples/missing-planes

Without` this PR reading `test-Z0.ome.tiff` results in just a single plane.
With this PR it results in 10 planes and the 5th Z stack is blank as expected.

The repo tests have also been green with this PR included and this is good to merge.

PR #3264, opened 2018-11-12
ImageConverter: usage extension

sbesson:
See https://trello.com/c/ko5vYYZ7/35-bfconvert-help-review

This PR primarily reviews the usage of `bfconvert` to list the available list of extension/formats as well as compression schemes for relevant file formats.

Additionally, the upgrade check is skipped if the usage is displayed e.g. when calling `bfconvert` without argument and the utility usage is printed directly to to the stdout independently of the logger level (see https://trello.com/c/FHWeDznP/53-cli-systemout-vs-logging). Similar changes are applied to the other utility classes (`showinf`, `gen-cache`, `tiffcomment`)

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -222,7 +222,7 @@ public final class ImageConverter {
   }
 
   /* Create a table of the extensions per output file format */
-  private String listExtensions() {
+  private static String listExtensions() {
       IFormatWriter[] writers = new ImageWriter().getWriters();
       String[] suffixes;
       StringBuilder sb = new StringBuilder();
@@ -239,7 +239,7 @@ public final class ImageConverter {
   }
 
   /* Create a table of the compression tables per output file format */
-  private String listCompressions() {
+  private static String listCompressions() {
       IFormatWriter[] writers = new ImageWriter().getWriters();
       String[] compressionTypes;
       StringBuilder sb = new StringBuilder();
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -233,7 +233,7 @@ public final class ImageConverter {
         for (int j=1; j<suffixes.length; j++) {
             sb.append(", ." + suffixes[j]);
         }
-        sb.append("\n");
+        sb.append('\n');
       }
       return sb.toString();
   }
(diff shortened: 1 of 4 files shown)


components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -221,6 +221,42 @@ else if (!args[i].equals(CommandLineTools.NO_UPGRADE_CHECK)) {
     return true;
   }
 
+  /* Create a table of the extensions per output file format */
+  private String listExtensions() {
```
mtbc: These could probably be `static`?

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -221,6 +221,42 @@ else if (!args[i].equals(CommandLineTools.NO_UPGRADE_CHECK)) {
...
+      StringBuilder sb = new StringBuilder();
+      for (int i=0; i<writers.length; i++) {
+        compressionTypes = writers[i].getCompressionTypes();
+        if (compressionTypes != null) {
+          sb.append(" * "  + writers[i].getFormat() + ": ");
+          sb.append(compressionTypes[0]);
+          for (int j=1; j<compressionTypes.length; j++) {
+              sb.append(", " + compressionTypes[j]);
+          }
+          sb.append("\n");
```
mtbc: If just adding a single character use the `char` method.

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -262,6 +298,16 @@ private void printUsage() {
       "     -padded: filename indexes for series, z, c and t will be zero padded",
       "     -option: add the specified key/value pair to the options list",
       "",
+      "The extension of the output file specifies the file format to use",
+      "for the conversion. The list of available formats and extensions is:",
+      "",
+      listExtensions(),
+      "",
```
mtbc: Don't need these blanks after the `list*` because already `println`'ing strings that end with `\n`.

mtbc:
* Could we sort the lists of formats and maybe the lists of extensions and compressions? They look kind of random.
* OME-TIFF gets listed twice.

sbesson:
Last commit addresses https://github.com/openmicroscopy/bioformats/pull/3264#issuecomment-438995344 by creating sorted maps of format/extension and format/compressions and factoring out the string building code.

The duplication of OME-TIFF comes from the addition of a new`PyramidOMETiffWriter` concrete class which is due for API review at a later milestone (https://trello.com/c/UOqS3lTH/61-pyramidstore-review). In the interim, the usafe of a sorted map should hide the internal details for consumers of the utility since the format/extension/suffixes are the same between the two classes.

mtbc:
"JPEG-2000 Lossy" is interesting: I suppose users have to quote that option.

sbesson:
Thanks @mtbc. Merging as more work might be coming on the `bfconvert` usage.

PR #3265, opened 2018-11-13
Fake: populate WellSample positions using Plane data (if present)

melissalinkert:
See https://trello.com/c/dQfCgPyX/79-fakereader-wellsamplepositions

To test, use a .fake file with a .fake.ini that specifies plane positions and SPW data.  For example:

```
$ touch test.fake
$ cat test.fake.ini
plates=2
plateRows=1
plateCols=3
fields=2

[series_0]
PositionX_0 = -0.5
PositionXUnit_0 = mm

[series_1]
PositionX_0 = 0.25
PositionXUnit_0 = mm

[series_2]
[series_3]
[series_4]

[series_5]
PositionY_0 = 100
PositionYUnit_0 = mm

[series_6]
PositionX_0 = 15
PositionXUnit_0 = nm
```

Verify that ```showinf -nopix -omexml test.fake``` populates both ```Plane``` and ```WellSample``` positions, and that the series tables map to the correct ```Plate```/```Well```/```WellSample``` (the ```WellSample``` ```Index``` should match the series index).  Default positions are set via https://github.com/ome/ome-model/blob/master/specification/src/main/java/ome/specification/XMLMockObjects.java#L1030, so there should not be any missing ```WellSample``` positions.

I considered opening a corresponding PR against ome-model to have ```XMLMockObjects``` set more realistic positions automatically, but didn't want to risk breaking anything else that uses ```XMLMockObjects``` without opening for discussion first.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -206,6 +206,13 @@ public class FakeReader extends FormatReader {
 
   private OMEXMLService omeXmlService;
 
+  private transient int screens = 0;
+  private transient int plates = 0;
+  private transient int plateRows = 0;
+  private transient int plateCols = 0;
+  private transient int fields = 0;
+  private transient int plateAcqs = 0;
+
   /**
    * Read byte-encoded metadata from the given plane.
    * @see FakeReader#readSpecialPixels(byte[], int, boolean, int, boolean)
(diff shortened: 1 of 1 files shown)


sbesson:
I suspect we can expand the `FakeReader` unit tests similarly to https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/test/loci/formats/utests/FakeReaderTest.java#L368.

Re XMLMockObjects, no particular objection to adding more flexibility at this level as well. Primary question is rather about how the API can be updated without appending even more arguments to the existing signature.

joshmoore:
> ...is rather about how the API can be updated without appending even more arguments to the existing signature.

Builder pattern.

dgault:
Using the sample ini provided (modified to reduce the series count to 6 rather than 7) results in the Plates, Wells and Well Samples being populated. The positions are correctly populated for the appropriate WellSample with default values appearing for the remainder. The Well Sample Index also matches correctly with the series index.

PR #3266, opened 2018-11-13
Add example and bfconvert options for generating pyramid resolutions

melissalinkert:
See https://trello.com/c/spl6WleR/74-ome-tiff-pyramid-testing-phase-ii and https://trello.com/c/cl9gX372/76-pyramid-option

```bfconvert``` now has two additional options: ```-pyramid-scale``` and ```-pyramid-resolutions```, both of which take an int parameter.  Both default to 1, so setting one but not the other is a no-op.  The new downsampling API is used to take each plane from the input file and scale by the specified value to create a pyramid with the specified number of resolutions.  For example, if the input file has a single 8000x8000 plane, then ```bfconvert -pyramid-scale 4 -pyramid-resolutions 4 input-file output.ome.tiff``` should result in a pyramid with resolutions 8000x8000, 2000x2000, 500x500, and 125x125.

Similarly, ```GeneratePyramidResolutions``` shows a minimal example for using the downsampling and resolution APIs to create a pyramid when the input file only contains the full resolution.

Any larger image without a stored pyramid can be used for testing.  ```curated/gatan/emdb```, ```curated/jpeg/big-images```, and ```curated/mrc/samples/1160.mrc``` are all smaller examples (and includes some floating point data).


melissalinkert:
Example class now moved to https://github.com/ome/bio-formats-examples/pull/40

PR #3267, opened 2018-11-13
LIF: unify double parsing to use DataTools.parseDouble(...)

melissalinkert:
See https://trello.com/c/Fig1abYV/239-lif-parsedouble-exception

All usage of ```new Double``` and ```Double.parseDouble``` in ```LIFReader``` should now be replaced by ```DataTools.parseDouble```.

```showinf curated/leica-lif/qa-21092/1-GBP1-CaaX+LLoMe.lif``` without this PR should throw ```NumberFormatException``` as noted in the Trello card.  The same test with this PR should result in successful reading.

I would expect all tests to pass without configuration changes; memo files should be unaffected.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -1310,10 +1310,10 @@ public class LIFReader extends FormatReader {
 
         String v = detector.getAttribute("Gain");
         Double gain =
-          v == null || v.trim().isEmpty() ? null : new Double(v.trim());
+          v == null || v.trim().isEmpty() ? null : DataTools.parseDouble(v.trim());
         v = detector.getAttribute("Offset");
         Double offset =
-          v == null || v.trim().isEmpty() ? null : new Double(v.trim());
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/formats-gpl/src/loci/formats/in/LIFReader.java [hunk 1]
```diff
@@ -2040,7 +2077,7 @@ private void translateImageNodes(Element imageNode, int i)
       if (StringUtils.isBlank(v)) {
         physicalLen = 0d;
       } else {
-        physicalLen = new Double(v.trim());
+        physicalLen = DataTools.parseDouble(v.trim());
```
sbesson: This seem to have broken https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/40072/ with some length values being off by a factor 100 .

melissalinkert: The breakage actually seems to be correct, and is a result of https://github.com/ome/ome-common-java/pull/30. In the process of updating configurations on https://github.com/openmicroscopy/data_repo_config/pull/342 with supporting screenshots from LAS X.

PR #3268, opened 2018-11-16
Allow xmlvalid utility to perform offline validation

sbesson:
This mirrors the behavior implemented in OMEXMLServiceImpl allowing to consume the xsd schemas available within the specification JAR when validating and XML.

An alternative approach would be to declare `OMEXMLServiceImpl.SCHEMA_CLASSPATH_READER` public and consume it in `XMLToolsValidate`.

To test this PR, try to run  `./xmlvalid` against any of the OME-XML samples at https://downloads.openmicroscopy.org/images/OME-XML/2016-06/. Offline, the command should fail without this PR and success with this PR. Online, the command should take a few seconds without this PR and return much faster with this PR due to the usage of the local schema in the specification JAR.

/cc @mtbc

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/XMLValidate.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/XMLValidate.java
@@ -35,8 +35,11 @@ package loci.formats.tools;
 import java.io.BufferedReader;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringReader;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import loci.common.Constants;
 import loci.common.xml.XMLTools;
@@ -46,6 +49,26 @@ import loci.formats.tiff.TiffParser;
  * Attempts to validate the given XML files.
  */
 public class XMLValidate {
+
+  /* The pattern of system ID URLs for OME-XML schema definitions. */
+  private static final Pattern SCHEMA_URL_PATTERN = Pattern.compile(
+    "http://www.openmicroscopy.org/Schemas/" +
+    "\\p{Alpha}+/(\\w+-\\w+)/(\\p{Alpha}+)\\.xsd");
+
+  private static final XMLTools.SchemaReader SCHEMA_CLASSPATH_READER =
+    new XMLTools.SchemaReader() {
+    @Override
+    public InputStream getSchemaAsStream(String url) {
+      final Matcher matcher = SCHEMA_URL_PATTERN.matcher(url);
+      if (matcher.matches()) {
+        /* from specification.jar */
+        return getClass().getResourceAsStream("/released-schema/" +
+          matcher.group(1) + "/" + matcher.group(2) + ".xsd");
+      } else {
+        return null;
+      }
+    }
+  };
 
   public static void process(String label, BufferedReader in)
     throws IOException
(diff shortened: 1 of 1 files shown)


mtbc:
With this PR https://trello.com/c/6zoPk0TG/525-ome-xml-at-under-40-bytes-sec still takes ages. Is that expected?

mtbc:
(Possibly the problem is the version check? Except I can't get `xmlvalid` to tell me its command-line options.)

sbesson:
You can try `./xmlvalid -no-upgrade` to skip the upgrade check. There is larger epic to expose the options across all CL tools (probably via `argparse4j`).

Re, https://github.com/openmicroscopy/bioformats/pull/3268#issuecomment-441988172, the biggest danger is the detection of false positives (e.g. when validating a file containing new enums using an old Bio-Formats). Arguably this is already what would happen with `showinf -validate` which uses the local schema and one goal of this PR is unifying the tools to use the same strategy. If the comment above remains a concern, the two options I can think of are:

- implement some high-level: offline/online validation flag
- implement some internal logic to compare the minor version of the schema with the one published upstream and only use the local schema if they match

The first option is the simplest but puts the burden on the consumer. The second one is probably more fine-grained but requires more time.

mtbc:
`java.io.FileNotFoundException: -no-upgrade` -- it doesn't understand that option.

mtbc:
Experimenting with and without `-no-upgrade` with `showinf` suggests that at least for that it doesn't make much different in timing anyway. Given https://github.com/openmicroscopy/bioformats/pull/3268#issuecomment-442008681 I'm thus not convinced that this PR is having the intended effect for `xmlvalid`.

A third option might be: if validation with local schema fails *then* try fetching live one.

rleigh-codelibre:
Just FYI, for OME Files we always use the local schema files via local XML Catalog and a custom EntityResolver.  The remote schema files are never used.

While there is an argument to be made that the remote schema files might be more up-to-date, the fact is that all of the model API is generated against the local schema, so by using the local schema you ensure that you are always validating against the specific schema the model and metadata store code was generated from, including all of the enumeration values.  If it's a missing major schema version then it will use the remote URI in any case.

jburel:
What is the status of this PR?
#3273 (memory leak and new methods) cannot be included in the build due to conflict

sbesson:
I need to review it as per the comments above. Excluding for now

mtbc:
It also takes ages with `time-series.ome.xml`, far longer than `showinf -no-upgrade`.

sbesson:
Some additional information. The fix provided here and my initial testing were actually done offline. Under these conditions it turns out the functionality works as expected i.e. `xmvalid` returned in a timely manner rather than throwing an exception while trying to resolve the schema remotely.

I can reproduce https://github.com/openmicroscopy/bioformats/pull/3268#issuecomment-443137577 when coming back offline. Trying to profile the calls, it looks despite the local schemabeing passed , the bottleneck still happens in [schemaFactory.newSchema](https://github.com/ome/ome-common-java/blob/v5.3.7/src/main/java/loci/common/xml/XMLTools.java#L874):

screen shot 2018-12-12 at 11 40 13

So far this change seems to resolve one of the 2 related issues (complete offline validation) but fails to provide the  performance improvements when running the validation oneline.

mtbc:
This PR could be useful then, we don't enough test/support offline users. (E.g., if I'd tried during my hour on the bus this morning!)

PR #3272, opened 2018-11-20
ci: appveyor: Test with JDK11

rleigh-codelibre:
Update to match Travis, by using JDK11 rather than 10.


sbesson:
Both Appveyor and Travis are green and unifying the test environments make sense. Thanks, merging.

PR #3273, opened 2018-11-22
Validate xml

jburel:
Add methods to validate files
I have not modified the returned value of the ``process`` method

Background https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8623&p=20270#p20270

PR #3274, opened 2018-11-22
Close TiffParser

jburel:
Implement closeable
A full review of the usage of TiffParser should happen when the switch to Java 8 is done
so we can take advantage of ``try-with-resources`` instead of ``try/finally``

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -32,6 +32,7 @@
 
 package loci.formats.tiff;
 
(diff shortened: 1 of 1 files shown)


sbesson:
:+1 for catching this. I do not think the review needs to be tied to the version of Java since `try-with-resources` is available in Java 7+  and already used in a couple of places.

jburel:
you are correct.
When this is merged we can start taking advantage and start closing resources
e.g. https://github.com/openmicroscopy/bioformats/pull/3273

jburel:
Closing this PR for now
This will be discussed at the next format meeting and we will use the card for discussion

PR #3278, opened 2018-11-26
CV7000: fix channel indexing when the stored indexes do not match the count per field

melissalinkert:
An arbitrary number of channels can be defined in the ```MeasurementDetail.mrf``` file, and only a subset may be referenced by each field.  This changes how channel counting and indexing is done to look
at the total number of unique channels in a field, instead of the total number of unique channels across all fields/wells.

See https://trello.com/c/b0nFSUzZ/49-cv7000

To test, use the ```Test-plate-1.wpi``` referenced on the card (configuration PR forthcoming).  Without this change, ```showinf -nopix Test-plate-1.wpi``` should show 10 channels.  This is because 10 channels are defined in the corresponding ```MeasurementDetail.mrf```, and the specific subset used varies by field.

With this change, the same test should show 5 channels in each series, which matches the number of unique channel identifiers in each field (and the acquisition description supplied in the ```*.md``` file).  Viewing a subset of the series should show that the channels are not mixed up (ImageJ may be easier for this since the images are 16 bit).  ```showinf -omexml -nopix``` should also show that ```WellSample.PositionX``` and ```WellSample.PositionY``` are populated, as noted in https://github.com/openmicroscopy/bioformats/pull/3224#discussion_r224841439

Memo files may be affected due to the addition of ```channelIndex``` in the ```Plane``` inner class.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -289,41 +289,47 @@ public class CV7000Reader extends FormatReader {
     }
 
     String firstFile = null;
-    int minSizeZ = Integer.MAX_VALUE, maxSizeZ = 0;
-    int minSizeC = Integer.MAX_VALUE, maxSizeC = 0;
-    int minSizeT = Integer.MAX_VALUE, maxSizeT = 0;
+    HashMap<Integer, MinMax> minMax = new HashMap<Integer, MinMax>();
+
     fields = 0;
     HashSet<Integer> uniqueWells = new HashSet<Integer>();
 
     for (Plane p : planeData) {
       if (p != null) {
+        int wellIndex = p.row * plate.getPlateColumns() + p.column;
+        if (!minMax.containsKey(wellIndex)) {
+          minMax.put(wellIndex, new MinMax());
+        }
+        MinMax m = minMax.get(wellIndex);
+
         if (p.file != null && firstFile == null) {
           firstFile = p.file;
         }
 
-        if (p.timepoint > maxSizeT) {
-          maxSizeT = p.timepoint;
+        if (p.timepoint > m.maxT) {
+          m.maxT = p.timepoint;
         }
-        if (p.timepoint < minSizeT) {
-          minSizeT = p.timepoint;
+        if (p.timepoint < m.minT) {
+          m.minT = p.timepoint;
         }
-        if (p.z > maxSizeZ) {
-          maxSizeZ = p.z;
+        if (p.z > m.maxZ) {
+          m.maxZ = p.z;
         }
-        if (p.z < minSizeZ) {
-          minSizeZ = p.z;
+        if (p.z < m.minZ) {
+          m.minZ = p.z;
         }
-        if (p.channelIndex > maxSizeC) {
-          maxSizeC = p.channelIndex;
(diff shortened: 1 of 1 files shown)


sbesson:
Imported the sample CV7000 dataset into a testing OMERO server with this PR included:

In total one plate contains 6 wells (B06 to B08 and E07 to E09) was imported matching the description of the datasets:

screen shot 2018-11-30 at 16 59 15

For each well except E07, all fields have positional information as shown by the plate grid. For all fields, individual images are detected as 9 Z-sections, 7 timepoints and 5 channels.

screen shot 2018-11-30 at 16 59 20

For the E row, images are detected with  9 Z-sections, 7 timepoints and 5 channels with the channel being different from the ones of row B (matching the mapping of each TIFF and the 10 channels declared MeasurementDetail.mrf)

screen shot 2018-11-30 at 16 59 25

Looking at the individual images, thumbnails appear as black. For Well E07 field 2, only the first Z section contains data (matching the submitter information about the fact these images are projections ) and only the timepoints 5 to 7 contain real data. All remaining planes are black. For wells on the row B only timepoints 1 to 4 contain real data

screen shot 2018-11-30 at 16 59 53

Summarizing briefly:

- the logic assigning channel per field seems to work as expected
- the WellSample position population logic works as expected for one well sample
- it looks like Z and T dimensions might fall under the same constraints as channels i.e. it is not possible to assume they will be conserved across all wells/fields and might need to be computed per field

melissalinkert:
Thanks for the thorough testing.  If I understand correctly, then:

* all fields in all wells in row B should have SizeT = 4
* all fields in all wells of row E should have SizeT = 3
* all fields in all wells of row E should have SizeZ = 1

with everything else unchanged.  My only slight concern would be if someone were performing some sort of analysis on a B well and an E well, where it might matter that the timepoint indexes don't overlap and everything in E is later than everything in B (but then maybe setting AcquisitionDate/DeltaT is enough?).

In any case, probably won't have time to make real progress on this until the week after next, so feel free to exclude if anything else is blocked.

sbesson:
With the last commits, the dimensions appear to match https://github.com/openmicroscopy/bioformats/pull/3278#issuecomment-443366386. This is also in agreement with the description accompanying the dataset that specifies that both sets of wells belong to different timelines and one of them is maximally projected while the other isn't.

screen shot 2019-01-14 at 14 46 57
screen shot 2019-01-14 at 14 47 04
screen shot 2019-01-14 at 14 47 22
screen shot 2019-01-14 at 14 47 17

I share the concern about the ability to compare the content of wells within a single plate. The supplied datasets clearly provided an example where this assumption is not valid. In order to inform the end-user, there are at least two pieces of metadata:

- `bts:TimelineIndex` specified the timeline for each well/image
- as suggested in https://github.com/openmicroscopy/bioformats/pull/3278#issuecomment-443366386, the combination of `AcquisitionDate/DeltaT` might be sufficient and more generic.
At the moment, the latter metadata is not parsed by the reader. It looks like populating would require parsing `bts:Time` for each TIFF/plane.

I would not count this is a blocker. These cahnges already allow to import this variant of the Yokogawa format in a representation that match the description of the submitters. Improving the temporal metadata population can be captured/actioned as a follow-up.

melissalinkert:
Card added for timestamp parsing: https://trello.com/c/pr1pJw7o/308-cv7000-parse-timestamp-information-for-all-planes

PR #3250, opened 2018-10-19
URL Pattern Reader

manics:
This is a version of FilePatternReader intended for importing remote files into OMERO as links. OMERO should only see this file, it should not see anything listed inside the `.urlpattern` file.

PR #3279, opened 2018-11-28
Refactor FilePatternReader into a generic WrappedReader

manics:
These commits are extracted from https://github.com/openmicroscopy/bioformats/pull/3250 and are intended to support the development of external wrapped readers by moving most of the helper methods out of `FilePatternReader` and into a new `WrappedReader` class.

These commits should result in no changes to the behaviour of `FilePatternReader`. Several arguable improvements could be made to `FilePatternReader` by taking advantage of `WrappedReader`:
- Pass `IMetadataConfigurable` methods to the helper (it is unclear whether this was intentionally omitted from the current `FilePatternReader` class)
- Construct the helper object in `initFile` and call `callDeferredSetters` instead of creating the helper in the class constructor. This would make the class more robust to other future wrappers since at present `!c.equals(FilePatternReader.class)` is required to prevent infinitely nested `FilePatternReader`.

See for example https://gitlab.com/openmicroscopy/incubator/bioformats-urlreader

### DIFF
--- a/components/formats-api/src/loci/formats/WrappedReader.java
+++ b/components/formats-api/src/loci/formats/WrappedReader.java
@@ -32,20 +32,35 @@
 
 package loci.formats;
 
-import loci.formats.in.MetadataLevel;
-import loci.formats.in.MetadataOptions;
-import loci.formats.meta.MetadataStore;
-
 import java.io.IOException;
 import java.util.Hashtable;
 import java.util.List;
-import java.util.Set;
 
 /**
  * Helper class for readers which wrap other readers.
  *
- * getHelper() must return a valid object once the constructor has completed unless the reader is
- * closed. returning null will cause strange errors.
+ * Some methods can only be called before initFile/setId, others can only be called after.
+ * This class assumes the helper object is only available after initFile() has been called.
+ *
+ * Instead the inherited FormatReader and IMetadataConfigurable methods are kept since the properties
+ * they set are inherited. These properties can be set on the helper class in initFile() by calling
+ * callDeferredSetters().
+ *
+ * The following setters and corresponding getters are not passed to the helper.
+ * If you initialise the helper in the constructor instead of at runtime during initFile()
+ * you may wish to override them.
+ *
+ * - getSupportedMetadataLevels
+ * - setMetadataOptions, getMetadataOptions
+ * - setGroupFiles, isGroupFiles
+ * - setNormalized, isNormalized
+ * - setOriginalMetadataPopulated, isOriginalMetadataPopulated
+ * - setMetadataFiltered, isMetadataFiltered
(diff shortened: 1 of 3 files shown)


components/formats-api/src/loci/formats/WrappedReader.java
```diff
@@ -0,0 +1,473 @@
...
+
+/**
+ * Helper class for readers which wrap other readers.
+ *
+ * Some methods can only be called before initFile/setId, others can only be called after.
+ * This class assumes the helper object is only available after initFile() has been called.
+ *
+ * Instead the inherited FormatReader and IMetadataConfigurable methods are kept since the properties
+ * they set are inherited. These properties can be set on the helper class in initFile() by calling
+ * callDeferredSetters().
```
joshmoore: It's a MUST that subclasses call this? Is it more likely to be required than to not be?

components/formats-api/src/loci/formats/WrappedReader.java
```diff
@@ -0,0 +1,473 @@
...
+ *
+ * The following setters and corresponding getters are not passed to the helper.
+ * If you initialise the helper in the constructor instead of at runtime during initFile()
+ * you may wish to override them.
```
joshmoore: You would override them to do the forwarding? Does callDeferredSetters help? Would a flag to detect where the helper was set help?

components/formats-api/src/loci/formats/WrappedReader.java
manics: This is equivalent to asking "Can a reader choose to ignore one or more of the fields set by the methods in `callDeferredSetters()`?". I've no idea what the technical answer is. Most things seem to work without calling them, the exception being `setMetadataStore()` which if not called will results in a failed OMERO import.

components/formats-api/src/loci/formats/WrappedReader.java
manics: It depends what the helper does. If the helper doesn't change its state anything between the construction of `WrappedReader` and `WrappedReader.initFile()` then either would work, in which case you might as well create the helper in `initFile()` (see second suggested improvement in the PR description). If it does change it's state then there's nothing we can do, it's down to the implementer to handle everything.

components/formats-api/src/loci/formats/WrappedReader.java
```diff
@@ -0,0 +1,473 @@
+/*
+ * #%L
+ * BSD implementations of Bio-Formats readers and writers
+ * %%
+ * Copyright (C) 2005 - 2017 Open Microscopy Environment:
```
sbesson: Can you fix the top-level the boilerplate to be`Top-level reader and writer APIs` and have `2018` as  the Copyright?

manics:
Updated

PR #3281, opened 2018-11-28
ImageJ exporter: skip grayscale LUTs and allow non-gray LUTs to be omitted

melissalinkert:
See https://trello.com/c/Qk6NBnPs/92-imagej-ome-tiff-writing-performance

cb82e22 fixes the immediate issue by omitting lookup table saving for any plane that has a default grayscale lookup table.  6e57191 adds an option to always omit lookup tables, which is ```false``` by default.

To test, run this macro:
```
run("Mitosis (26MB, 5D stack)");
// resulting file should be 35 MB
run("Bio-Formats Exporter", "save=macro-no-lut-no-save.ome.tiff do_not_save_lookup_tables compression=Uncompressed");

// resulting file should be 35 MB
// not disabling LUTs doesn't matter here since every plane has the default grayscale LUT
run("Bio-Formats Exporter", "save=macro-no-lut-save.ome.tiff compression=Uncompressed");

// explicitly change the LUT for every plane
run("Thallium");
// resulting file should be 35 MB since LUT saving is explicitly disabled
run("Bio-Formats Exporter", "save=macro-thallium-no-save.ome.tiff do_not_save_lookup_tables compression=Uncompressed");

// resulting file should be 226 MB (LUTs are preserved)
// should be backwards compatible with old macros when a non-default LUT is present
run("Bio-Formats Exporter", "save=macro-thallium-save.ome.tiff compression=Uncompressed");
close();
```

and verify that the resulting file sizes match expected sizes in the comments and that write times are generally faster except in the very last case.  That said, I'm not sure that I'm 100% sold on 6e57191 (is it too complicated for users to understand? should the default behavior be different?), so happy to revert, rework or split into a separate PR if anyone has strong feelings.

I didn't do anything here with respect to more efficiently storing duplicate LUTs in a TIFF as discussed in https://openmicroscopy.slack.com/archives/C0K5EED3R/p1543417107942200, but that could be a follow-up PR (it is just a little trickier, and may conflict with other TIFF-focused PRs).

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -140,6 +140,7 @@ public class Exporter {
         Boolean padded = null;
         Boolean saveRoi = null;
         String compression = null;
+        Boolean noLookupTables = null;
 
         Boolean windowless = Boolean.FALSE;
         if (plugin.arg != null) {
@@ -170,6 +171,8 @@ public class Exporter {
             if (w != null) {
                 windowless = Boolean.valueOf(w);
             }
+            String lut = Macro.getValue(plugin.arg, "skip_luts", "false");
+            noLookupTables = Boolean.valueOf(lut);
             plugin.arg = null;
         }
         if (outfile == null) {
@@ -301,6 +304,7 @@ public class Exporter {
             if (splitC == null) splitC = Boolean.FALSE;
             if (splitT == null) splitT = Boolean.FALSE;
             if (padded == null) padded = Boolean.FALSE;
+            if (noLookupTables == null) noLookupTables = false;
         }
         if (splitZ == null || splitC == null || splitT == null) {
             // ask if we want to export multiple files
@@ -311,12 +315,22 @@ public class Exporter {
             multiFile.addCheckbox("Write_each_timepoint to a separate file", false);
             multiFile.addCheckbox("Write_each_channel to a separate file", false);
             multiFile.addCheckbox("Use zero padding for filename indexes", false);
+            // prompt for lookup tables here instead of in compression options window,
+            // since the compression window won't be shown for formats with
+            // only one compression type
+            //
+            // checkbox is "Do not save..." instead of "Save..." because macro
+            // recording/parsing only looks for checkboxes that are enabled
+            // this preserves backwards compatibility with existing macros,
+            // so that non-default lookup tables do not suddenly disappear
+            multiFile.addCheckbox("Do_not_save_lookup_tables", false);
             multiFile.showDialog();
 
             splitZ = multiFile.getNextBoolean();
             splitT = multiFile.getNextBoolean();
             splitC = multiFile.getNextBoolean();
             padded = multiFile.getNextBoolean();
+            noLookupTables = multiFile.getNextBoolean();
             if (multiFile.wasCanceled()) return;
         }
 
@@ -792,7 +806,7 @@ public class Exporter {
                     // saving a LUT for every plane can cause performance issues,
                     // especially for 16 bit data
                     // see https://trello.com/c/Qk6NBnPs/92-imagej-ome-tiff-writing-performance
-                    if (!proc.isDefaultLut()) {
+                    if (!proc.isDefaultLut() && (noLookupTables == null || !noLookupTables)) {
                         if (luts[currentChannel] != null) {
                             // expand to 16-bit LUT if necessary
 
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -46,12 +46,14 @@ import ij.process.ImageProcessor;
 import ij.process.LUT;
 import ij.process.ShortProcessor;
 
+import java.awt.Checkbox;
 import java.awt.Image;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


sbesson:
Initial testing suggests that the size of the generated file is definitely reduced and in agreement with the expectations from a regular ImageJ TIFF export:

```
sbesson@ls30630:Desktop $ ls -alh *.tif*
-rw-r--r--  1 sbesson  1133848969    38M  3 Dec 17:38 out.ome.tiff
-rw-r--r--  1 sbesson  1133848969    38M  3 Dec 17:39 out.tiff
-rw-r--r--  1 sbesson  1133848969    39M  3 Dec 17:37 test.ome.tiff
-rw-r--r--@ 1 sbesson  1133848969    38M  3 Dec 17:38 test.tif
```

As a matter of fact, the ImageJ OME-TIFF export time is also reduced and more inline with the conversion times of the Bio-Formats command-line tools although the overall writing performance is still a couple of times slower than the ImageJ TIFF export. I will perform more testing tomorrow and report more precise metrics.

melissalinkert:
Might also be worth checking what the ImageJ TIFF export files look like.  We've seen in the past TIFFs from ImageJ that use a single IFD, and assume that all pixels are written contiguously.  That's obviously going to be faster than writing one IFD per plane (with the various offset calculations that go along with that), but probably not something we should duplicate.

melissalinkert:
A couple of quick tests suggest that writing larger tiles (instead of a single row of pixels) will further reduce the write time, as discussed separately.  I'd propose to do that as a follow-up PR though once this and #3284 are in.  Card added: https://trello.com/c/3KFAZQue/97-add-tiling-to-imagej-exporter

sbesson:
Sorry for the delay, I have been the running the following macro largely inspired by the description:


```
run("Mitosis (26MB, 5D stack)");

start = getTime();
saveAs("Tiff", "/tmp/mitosis-default.tiff");
print("ImageJ Save as TIFF", (getTime()-start)/1000);

start = getTime();
run("Bio-Formats Exporter", "save=/tmp/mitosis-nolut.ome.tiff do_not_save_lookup_tables compression=Uncompressed");
print("Bio-Formats save as TIFF without LUT", (getTime()-start)/1000);

start = getTime();
run("Bio-Formats Exporter", "save=/tmp/mitosis-default.ome.tiff compression=Uncompressed");
print("Bio-Formats save as TIFF with LUT", (getTime()-start)/1000);

run("Thallium");

start = getTime();
run("Bio-Formats Exporter", "save=/tmp/mitosis-thallium-nolut.ome.tiff do_not_save_lookup_tables compression=Uncompressed");
print("Bio-Formats save as TIFF without LUT", (getTime()-start)/1000);

start = getTime();
run("Bio-Formats Exporter", "save=/tmp/mitosis-thallium-default.ome.tiff compression=Uncompressed");
print("Bio-Formats save as TIFF with LUT:" + (getTime()-start)/1000);
close();
```

This was run on Imagej with three versions of Bio-Formats:
- with the last stable version of Bio-Formats (5.9.2)
- with the last milestone release of Bio-Formast 6 (6.0.0-m3)
- with the daily builds including this PR as well as other ongoing performance improvements

|                                    | Bio-Formats 5.9.2 | Bio-Formats 6.0.0-m3 | Bio-Formats 6.0.0-SNAPSHOT |
|------------------------------------|-------------------|----------------------|----------------------------|
| Save as TIFF (ImageJ)              | 0.149 (33M)       | 0.17s (33M)          | 0.124s (33M)               |
| Save as OME-TIFF greyscale, no LUT | 13.791s (225M)    | 10.227s (226M)       | 4.35s (34M)                |
| Save as OME-TIFF greyscale, no LUT | 12.88s (225M)     | 10.181s (226M)       | 4.248s (34M)               |
| Save as OME-TIFF thallium, no LUT  | 12.908s (225M)    | 10.246s (226M)       | 4.26s (34M)                |
| Save as OME-TIFF thallium, no LUT  | 12.953s (225M)    | 10.311s (226M)       | 9.715s (226M)              |
| bfconvert                          | 9.334s (34 M)     | 6.363s (34M)         | 5.168s (34M)               |

In terms of functional teting
- the output file size is now consistent with the default TIFF export except when LUTs are present and saved at export time :+1:
- the writing time for this particular sample is improved by a factor with the various performance improvements :+1:
- looking at other immediate potential improvements, the structure of the TIFF files is very similar: 510 IFDs in all cases. The most notable different is the number of Rows/Strip: 196 for the Save as TIFF vs 1 for the OME-TIFF export. Agreed with @melissalinkert, we might want to wait until all `TiffSaver` related PR are included but this looks like the next low-cost performance improvement we can look at.

In terms of usability, a couple of comments:

- I was definitely confused by the double negation (`Do not` + box unchecked) and more generally I am concerned by adding more options to the succession of 3 export windows.

screen shot 2018-12-11 at 22 07 57

- I could see value in enabling the extra LUT flexibility via the macro language as a starting point. More important is probably to make sure we agree default behavior
- unrelated to this PR, I realized there has been some behavior changes in the exporter UI when selecting the output file since Bio-Formats 5.9.2. When clicking `Save as > OME-TIFF` the default export format is no longer `OME-TIFF` as expected and filling `<filename>` without the `.ome.tif[f]` extension results in an exception

screen shot 2018-12-11 at 22 13 53

melissalinkert:
31c6189 removes the ```Do not save lookup tables``` checkbox from the UI at least for now, as discussed separately.  The option still exists when using a macro, so the macro code above should continue to work as before.

sbesson:
Retested both the graphical export workflow as well as the Macro workflow. Everything works as expected. Merging for the next milestone.

PR #3283, opened 2018-11-29
Formats plugins review

jburel:
Make sure all resources are always closed


dgault:
The changes to these 3 files look good and have no negative impact on builds. PR is good to merge.

PR #3284, opened 2018-11-29
Update default tile size to 0

dgault:
This PR is in response to https://trello.com/c/U9gw1uKc/89-tiling-api-tile-size-image-size

The request was to allow the full image size to be a valid tile size.

Without this PR:
- By default tileSizeX is the image width and tileSizeY is the image height
- These default tile sizes indicate that tiling is not being used
- A tileSize of 0 is invalid and throws an exception

With this PR:
- The default tile sizes are now 0
- A tileSizeX of the image width and tileSizeY of the height are valid sizes and tiling will be used
- A tileSize of 0 is valid and indicates that tiling is not being used

I don't believe there is any need to update the tiling developers page at https://github.com/ome/bio-formats-documentation/blob/9bc5d24a4b3a328f36ca8fd4b668bdc731cffa7e/sphinx/developers/tiling.rst

### DIFF
--- a/components/formats-api/src/loci/formats/FormatWriter.java
+++ b/components/formats-api/src/loci/formats/FormatWriter.java
@@ -304,28 +304,28 @@ public abstract class FormatWriter extends FormatHandler
   /* @see IFormatWriter#getTileSizeX() */
   @Override
   public int getTileSizeX() throws FormatException {
-    return getSizeX();
+    return 0;
   }
 
   /* @see IFormatWriter#setTileSizeX(int) */
   @Override
   public int setTileSizeX(int tileSize) throws FormatException {
     int width = getSizeX();
-    if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize < 0) throw new FormatException("Tile size must be >= 0. Setting a tile size of 0 will disable tiling");
     return width;
   }
 
   /* @see IFormatWriter#getTileSizeY() */
   @Override
   public int getTileSizeY() throws FormatException {
-    return getSizeY();
+    return 0;
   }
 
   /* @see IFormatWriter#setTileSizeY(int) */
   @Override
   public int setTileSizeY(int tileSize) throws FormatException {
     int height = getSizeY();
-    if (tileSize <= 0) throw new FormatException("Tile size must be > 0.");
+    if (tileSize < 0) throw new FormatException("Tile size must be >= 0. Setting a tile size of 0 will disable tiling");
     return height;
   }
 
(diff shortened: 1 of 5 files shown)


rleigh-codelibre:
Have you considered use of something like `Optional<int>` to provide a clean and standard way to communicate the difference between an unset and set or defaulted value, without the need for "magic" values like zero or the image size.

The image size is bad as the tile size because it doesn't necessarily reflect the intent of the user or the writer.  They can't know for sure if tiling is enabled because not all image sizes are valid tile sizes.

Zero is bad as the tile size, because it will result in a divide by zero if the caller doesn't check every time.  Especially since dividing the image size by the tile size is going to be a common operation.  As is getting the modulus.

Something like an `Optional` would clearly indicate if tiling is disabled (value not present) or enabled (value present indicating the tile size).

dgault:
The other alternative would be to simply add in the API methods originally discussed, `boolean isTilingEnabled()` and `setTilingEnabled(boolean tiling)`. That would accomplish the same thing, giving the user a method to check rather than relaying on a particular value.

rleigh-codelibre:
That's certainly a possibility,  However, it would still permit incorrect usage of the `getTileSize` methods if you ignored these methods, for example in existing callers, while using `Optional` makes it impossible to use incorrectly by being explicit and self-describing.

melissalinkert:
Changes all make sense upon reading.  I think it does make sense to have an easier way of checking that tiling is disabled than testing ```== 0```, but I don't think that needs to block this PR since the priority is to improve TIFF size/writing performance.  Will test more thoroughly and double-check examples tomorrow.

melissalinkert:
Few more thoughts after testing a bit more:

* Javadoc for all 4 methods should be updated in ```IFormatWriter```
* Calling ```setTileSizeX(0)``` and ```setTileSizeY(0)``` explicitly doesn't disable tiling, I'd guess due to https://github.com/openmicroscopy/bioformats/pull/3284/files#diff-8fd83f088ec5c5707bcdb304467c4505R541 (a PR is open against this branch to add a relevant test)
* More complicated tiling examples probably need to be updated to check for 0 (they don't throw an exception now, but I think that's only due to the above issue):
  - https://github.com/ome/bio-formats-examples/blob/master/src/main/java/OverlappedTiledWriter.java
  - https://github.com/ome/bio-formats-examples/blob/master/src/main/java/TiledReaderWriter.java

sbesson:
Does one of the TIFF writing unit tests explicitly test the case where tile size == image size i.e. the main limitation of the previous API that this PR is proposing to tackle?

dgault:
I shall modify the tests now to ensure its covered

PR #3285, opened 2018-12-04
MicroManager: Store channel colors consistent with metadata

dgault:
This issue was raised in forum thread https://www.openmicroscopy.org/community/viewtopic.php?f=13&t=8641&p=20331#p20331

A sample file is available in QA-27124

The thread covers 2 issues:
- The values stored in the metadata when exported from MicroManager differ from those expected. This will need resolution for the MicroManager team
- When opening the OME-TIFF the colors are read from the OME-XML and display correctly based on the stored values. When opening via the metadata file the MicroManagerReader is used and the colors are not stored resulting in different behaviour.

This PR covers the second scenario and should result in the same colors being populated and displayed when opening via either the ome.tiff or the metadata file.

Note the change to remove line 753 was due to it unnecessarily removing an extra character. I am unsure if this will have any unforeseen knock on effect on other metadata being parsed.

To test:
- Without this PR using QA-27124, open with Bio-Formats using the ome.tiff file and then the metadata file. Verify that when opening the ome.tiff that one of the channels appears as Magenta. The metadata file when opened will not have the magenta coloring. 
- Without this PR using QA-27124, use `showinf -omexml -nopix` and verify that with the ome.tiff the channels have color properties populated. With the metadata file the channels will have no color associated with them in the OME-XML
- With the PR rerun the above tests and verify that both files display the same with the Magenta color, and that the OME-XML for both contains color properties for the channels
- Ensure builds and tests are green

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -55,6 +55,7 @@ import loci.formats.meta.MetadataStore;
 import loci.formats.tiff.IFD;
 import loci.formats.tiff.IFDList;
 import loci.formats.tiff.TiffParser;
+import ome.xml.model.primitives.Color;
 import ome.xml.model.primitives.Timestamp;
 
 import org.xml.sax.Attributes;
@@ -370,6 +371,9 @@ public class MicromanagerReader extends FormatReader {
 
         for (int c=0; c<p.channels.length; c++) {
           store.setChannelName(p.channels[c], i, c);
+        }
+        for (int c=0; c<p.channelColors.length; c++) {
+          store.setChannelColor(Color.valueOf(p.channelColors[c]), i, c);
         }
 
         Length sizeX = FormatTools.getPhysicalSizeX(p.pixelSize);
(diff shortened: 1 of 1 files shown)


dgault:
I cant find any min max values in the metadata so I don't believe its related to the parsing. Im not sure what the difference might be.

sbesson:
I think I finally found to the root of this issue. The min/max settings in this case were calculated from the mid-T plane. This led me to realize the number of timepoints is inconsistent between the 2 imported image:
- the OME-TIFF reader detects 439 timepoints across 2 channels 
- the Micro-Manager reader detects 1000 timepoints across 2 channels
All planes for the last timepoints of the Micro-Manager image are black. This matches the fact there are only 878 IFDs in the TIFF file.

Testing previous versions, it looks like the dimensions have been incorrected detected at least since version 5.3.0 so I do not think this is a regression but probably a new bug. I do not know whether this has already been recorded independentyl. Happy to create a card otherwise /cc @melissalinkert 

Since the original bug tackled by this PR (the mismatched channel colors) is addressed, I will merge it now as an incremental step towards consistent behavior.

melissalinkert:
```MicromanagerReader``` has some support for omitting timepoints that weren't actually acquired, but it looks like this needs to be generalized for the case when pixels are stored in OME-TIFF (not one file per plane).  See https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java#L1099

Card added: https://trello.com/c/viXSC6OF/300-micromanager-update-timepoint-cropping-to-handle-ome-tiff-pixels-files

PR #3288, opened 2018-12-11
Gatan DM3/DM4: adjust endianness and record byte count for long values

melissalinkert:
Backported from private PR.  See https://trello.com/c/kTL4I4HP/282-gatan-dm4-missing-metadata

This change fixes reading of long and long array values. Endianness is now swapped as needed, consistent with how other multi-byte values are read. The width of long values is also now correctly defined, so that long array values are all read from the correct file offset.

To test, use the file from QA 21761.  ```showinf -nopix -debug``` without this change should show log messages during tag parsing similar to ```unknown type```.  The original metadata table should contain ~60 key/value pairs.

The same test with this change should not show any ```unknown type``` messages during tag parsing, and the original metadata table should be substantially larger (> 200 key/value pairs).

Not a high priority for the next milestone (or 6.0.0 in general).  I wouldn't expect this to cause any test failures or impact memo files, so would be safe for a patch release if needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -707,7 +707,10 @@ public class GatanReader extends FormatReader {
         return in.readByte();
       case UNKNOWN:
       case UNKNOWN2:
-        return in.readLong();
+        if (adjustEndianness) in.order(!in.isLittleEndian());
+        long l = in.readLong();
+        if (adjustEndianness) in.order(!in.isLittleEndian());
+        return l;
     }
     return 0;
   }
@@ -727,6 +730,9 @@ public class GatanReader extends FormatReader {
       case UBYTE:
       case CHAR:
         return 1;
+      case UNKNOWN:
+      case UNKNOWN2:
+        return 8;
     }
     return 0;
   }


chris-allan:
Mistakenly merged by me. Should be double checked against publicly available Gatan data. Tests were green so this PR did not introduce any regressions but we should be 100% we want it in.

/cc @sbesson, @dgault, @joshmoore

sbesson:
Retested by importing the sample DM4 image.

Without this PR
screen shot 2018-12-17 at 16 05 47

With this PR
screen shot 2018-12-17 at 16 07 06

Pixel reading is unchanged as confirmed by the automated tests against all DM3/DM4 samples.
The amount of original metadata increased as described. Most of the values are found identical with a couple of differences read more sensible (`Active Size (pixels)`). For this sample file at least, the OME `Voltage` element is correctly populated.
The only change I was not 100% sure about  is the `Magnification` incorrectly set to `0` without this PR and now set to `115000`. There are three magnification key/value pairs in the metadata table and I assume `Indicated Magnification` is stil the correct one to parse?

screen shot 2018-12-17 at 16 11 59

Barring this minor comment, everything looks good

melissalinkert:
I think ```Indicated Magnification``` is still correct and has a valid value here, based upon the ```mag``` value from https://microscopies.med.univ-tours.fr/?page_id=260 as noted in the original forum post (https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8592).  A data sheet for the Tecnai G2 F30 (best guess of the same/comparable instrument) also suggests that magnifications up to 1000000x can be valid:

http://emcfiles.missouri.edu/pdf/2007_07_TecnaiF2_30_ds.pdf

PR #3137, opened 2018-05-01
Improve AFI and SVS metadata

melissalinkert:
Backported from a private PR.

This updates ```AFIReader``` to more closely mirror ```SVSReader```, updates image names to use the same format as https://github.com/openmicroscopy/bioformats/pull/3094, and fixes exposure time and channel name metadata.

SVS tests should continue to pass with no configuration changes.  AFI files require updated configuration; the main things to check using any ```.afi``` from ```curated/afi/samples``` are:

* ```showinf -nopix file.afi``` and ```showinf -nopix -noflat file.afi``` show the same series counts and dimensions as ```showinf -nopix file_xyz.svs``` and ```showinf -nopix -noflat file_xyz.svs```
* ```showinf -nopix -omexml file.afi``` shows plausible exposure times - less than 1 second with this change, compared with hundreds of seconds with develop
* ```showinf -nopix -omexml file.afi``` shows channel names that correspond to the names of the grouped .svs files
* ```showinf -nopix -noflat file.afi``` shows image names that clearly indicate the label and macro series as in https://github.com/openmicroscopy/bioformats/pull/3094
* ```showinf -nopix file.afi``` shows 3 entries for each key in the original metadata table, one per channel

073ecff, 30434e2, and 1344477 add logic for handling pyramid resolutions and thumbnails with mismatched pixel types.  If there is a mismatch within resolutions of a single SVS, then anything not matching the full resolution image is removed.  If there is a mismatch between the whole pyramid in two SVS files (e.g. one SVS has all 16-bit images, and one has all 8-bit images), then the narrower pixels are expanded.  Often this indicates inconsistent post-processing, but I don't have a good example for testing as these cases are expected to be very rare in practice.  I could add ```WARN``` or ```ERROR``` level logging if that would help, and/or an option to turn off that logic.

This does require a minor release as memo files will be affected.  I don't have any other planned AFI/SVS changes though, so it's not blocking anything and can be medium/low priority for 5.9.0.

PR #3289, opened 2018-12-17
SVS: add support for multiple Z sections

melissalinkert:
Backported from a private PR.  See https://trello.com/c/zewt9Ke4/50-svs-z-stack-support

To test, use ```/uod/idr/repos/curated/svs/converted/z-stack/PHAS 01.svs```.  As noted in the corresponding readme, this was converted from an existing .ndpi file using Aperio ImageScope (a screenshot of the .svs in ImageScope is in the same directory).  With this PR, ```showinf -noflat PHAS\ 01.svs``` should show 3 resolutions with ```SizeZ = 4``` plus 2 extra series for the label and macro images.

When multiple Z sections are present, the expected IFD ordering is:

- largest resolution (r = 0), Z= 0
- extra ignored image (see #3137)
- r = 1, Z = 0
...
- r = pyramid depth - 1, Z = 0
- r = 0, Z = 1
...
- r = pyramid depth - 1, Z = 1
...
- r = pyramid depth - 1, Z = SizeZ - 1
- label image
- macro image

I would expect this change to affect memo files, but all tests should continue to pass.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SVSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SVSReader.java
@@ -27,6 +27,8 @@ package loci.formats.in;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


dgault:
With the PR included and using the test image instructed (`/uod/idr/repos/curated/svs/converted/z-stack/PHAS 01.svs`), results in the image being read and displaying correctly. 

There are 3 resolutions plus the macro image and label as expected. Each resolution has 4Z, the images at each Z look very similar but the pixel values do differ so each plane is being read correctly. Looking into the IFD ordering also matches that described in the PR description.

The SVS tests have been passing for a while now so this PR and the associated config PR are good to merge.

PR #3292, opened 2018-12-18
ND2: fix chunk map handling when CustomData blocks are between ImageDataSeqs

melissalinkert:
Backported from a private PR.  See https://trello.com/c/dDcSZtur/296-indexoutofboundsexception

To test, use ```inbox/from-guilhem-chenon/faultyND2.nd2```.  Without this change, ```showinf -nopix``` should result in an ```IndexOutOfBoundsException``` as noted in the card.

With this change, the same test should show 7 series each with 5 channels.  Nikon's NIS Elements viewer shows 7 timepoints each with 5 channels, so this fixes the plane detection issue and reduces the problem with ```faultyND2.nd2``` to the one reported in https://trello.com/c/LEjj3O3A/281-nd2-timepoints-incorrectly-detected-as-series.  No configuration PR (yet), since the dimensions are still wrong.

I would not expect tests to fail or memo files to be impacted.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -678,7 +678,7 @@ public class NativeND2Reader extends SubResolutionFormatReader {
               }
 
               if(!entry.name.startsWith("ImageDataSeq")) {
-                break;
+                continue;
               }
 
               if(lastImage!=null) {


sbesson:
At least one failure reported in  https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/45847/console

```
   [testng] [2018-12-19 05:39:57,924] [pool-1-thread-9] Initializing /data/public/aryeh/qa-9507/control003.nd2: 
   ...
   [testng] [2018-12-19 05:40:09,339] [pool-1-thread-9] 	ExposureTimes: FAILED (Series 4 plane 0 channel 0 (got ome.units.quantity.Time: value[0.4], unit[s] stored as java.lang.Double, expected ome.units.quantity.Time: value[0.3], unit[s] stored as java.lang.Double))
```

Looking at the [original file](https://downloads.openmicroscopy.org/images/ND2/aryeh/qa-9507/), I am a bit unclear what the expected ground truth should be.

melissalinkert:
From looking at NIS Elements, it looks like channel 0 should have a 300 ms exposure time, and channels 1 and 2 should have 400 ms exposure times.

Screenshots of the failing file in NIS Elements (posting here since the .nd2 itself is public):

image
image
image

Both the configured exposure times and the exposure times with this PR are then incorrect.  Captured as https://trello.com/c/UnDryDi9/113-fix-exposure-times-for-control003nd2.  I'd propose to open a PR in addition to this one (that shouldn't conflict) before review picks up again in the new year.

Issue #3127, opened 2018-04-18
Physical size units issue with Fiji


ehrenfeu:
Hi all,

I'm experiencing an issue with the units stored by BF when exporting an image from (an up-to-date) Fiji. When setting the "Unit of length" to `nm` or [`nanometre`](https://en.wikipedia.org/wiki/Nanometre) (which are correctly recognized by Fiji), then exporting and re-importing it with Bio-Formats, the `PhysicalSizeXUnit` is reported as `µm` (or `microns`) - while being inconsistent with the "Image Properties" dialog of Fiji (the latter still displaying `nm`).

Checking the OME-XML generated by the BF exporter reveals that the physical unit stored there is actually `µm` whereas it should be rather `nm`.

The following IJ1 macro demonstrates this issue:

```javascript
TMPNAME = "/tmp/physical_units_test.ome.tif";

function print_calibration() {
    getPixelSize(unit, pw, ph, pd);
    print("Pixel size (" + getTitle() + "):  x=" + pw + unit + " y=" + ph + unit);
}

run("Close All");
print("\\Clear");

newImage("orig", "8-bit white", 30, 20, 1);
run("Properties...", "unit=nm pixel_width=100 pixel_height=100");
run("Bio-Formats Exporter", "save=" + TMPNAME);
print_calibration();

run("Bio-Formats Importer", "open=" + TMPNAME);
rename("reloaded");
print_calibration();

run("Properties...", "unit=nm pixel_width=100 pixel_height=100");
rename("recalibrated");
print_calibration();
```

Resulting output is:
```
Pixel size (orig):  x=100nm y=100nm
Pixel size (reloaded):  x=100microns y=100microns
Pixel size (recalibrated):  x=100nm y=100nm
```

Any hints?

Cheers,
Niko

dgault:
Hi @ehrenfeu,

I can certainly reproduce the same behaviour as you are seeing. From some initial debugging it looks as though the Exporter in the Bio-Formats plugin is specifically choosing not to apply the calibration. Quickly testing an override of this has resulted in a `NullPointerException` so I shall have to continue debugging this further. I will take a more detailed look into this tomorrow and update with any progress.

melissalinkert:
Thanks again for reporting this, @ehrenfeu.  Closing in favor of proposed fix in https://github.com/openmicroscopy/bioformats/pull/3295; happy to continue the discussion there.

PR #3295, opened 2018-12-19
ImageJ export: preserve physical size units

melissalinkert:
See https://github.com/openmicroscopy/bioformats/issues/3127

Both the physical sizes and units are now taken from the ImagePlus' Calibration.  This should have no impact on the common case where physical sizes are unchanged after import, but should fix the case when units are changed after import.

To test, make sure that the original bug is fixed by running the macro in https://github.com/openmicroscopy/bioformats/issues/3127 with this change.  ```showinf -omexml``` on the resulting file should also confirm that the correct physical sizes are stored.  Also test with line 12 of the macro commented out (the first ```run("Properties..." ...```) and uncommented but with a unit that obviously cannot be mapped to an OME unit (I tried ```unit=december```).  Opening files using the Bio-Formats importer plugin and then exporting again should also continue to work as before.

Not sure if there are any other use cases to consider, or if/how this differs from the NPE-producing test mentioned in https://github.com/openmicroscopy/bioformats/issues/3127#issuecomment-382781009.
ImageJ allows the specified units to be free-form, so I do wonder if we should be doing any extra checking/logging in the case where a unit string cannot be mapped to real unit enum (as in the ```unit=december``` test above).

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -505,10 +505,10 @@ public class Exporter {
 
             Calibration cal = imp.getCalibration();
 
-            store.setPixelsPhysicalSizeX(FormatTools.getPhysicalSizeX(cal.pixelWidth), 0);
-            store.setPixelsPhysicalSizeY(FormatTools.getPhysicalSizeY(cal.pixelHeight), 0);
-            store.setPixelsPhysicalSizeZ(FormatTools.getPhysicalSizeZ(cal.pixelDepth), 0);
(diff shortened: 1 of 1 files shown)


dgault:
Tested using the macro from https://github.com/openmicroscopy/bioformats/issues/3127

The resulting log shows:
```
Pixel size (orig):  x=100nm y=100nm
Pixel size (reloaded):  x=0.1microns y=0.1microns
```

It still uses microns for the reloaded image but the value has been correctly scaled. When using showinf to view the OME-XML the value is correct and in `nm` as expected.

This PR is working as expected and good to merge.

PR #3296, opened 2018-12-20
Fix tests to fail when a configured file throws UnknownFormatException

melissalinkert:
Prior to this change, any file for which ```isThisType``` did not return true in at least one reader would have been removed from the list of tests.  This meant that regressions affecting ```isThisType``` (especially on non-TIFF readers) were not necessarily detected.

I encountered this issue while in the process of backporting a private PR affecting ```MRCReader```, but it's fairly easy to test even with fake files.  To set up (without this change):

```
$ mkdir $TEST
$ touch $TEST/test.fake
$ ant -Dtestng.configDirectory=$TEST -Dtestng.directory=$TEST test-config
$ touch $TEST/unreadable-file.txt
$ ant -Dtestng.configDirectory=$TEST -Dtestng.directory=$TEST test-automated # this should pass
# now apply https://gist.github.com/melissalinkert/16afa151faa5f076b7f7ffce85597403 and rebuild
$ ant -Dtestng.configDirectory=$TEST -Dtestng.directory=$TEST test-automated # this should still pass, with 0 files tested
```

With this change and https://gist.github.com/melissalinkert/16afa151faa5f076b7f7ffce85597403:

```
$ ant -Dtestng.configDirectory=$TEST -Dtestng.directory=$TEST test-automated
```

should fail - only ```test.fake``` is tested, and ```setId``` fails with ```UnknownFormatException```.  With this change only (revert https://gist.github.com/melissalinkert/16afa151faa5f076b7f7ffce85597403):

```
$ ant -Dtestng.configDirectory=$TEST -Dtestng.directory=$TEST test-automated
```

should pass with only ```test.fake``` tested.  ```unreadable-file.txt``` is not tested because it throws ```UnknownFormatException``` and is not configured with ```test=true```, which means that we hopefully won't get additional false negatives from various extra text files that are scattered throughout the repo.

I'd hope that this wouldn't cause any additional test failures, but will keep an eye on tomorrow's build and note any changes here.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTestFactory.java
@@ -37,6 +37,7 @@ import java.util.HashMap;
 
 import loci.common.DataTools;
 import loci.formats.FileStitcher;
+import loci.formats.UnknownFormatException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


melissalinkert:
Looks like only 3 files fail with this change, and all are Deltavision: https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/46070/console

I was initially concerned that https://github.com/openmicroscopy/bioformats/pull/3297 broke those files, but ```UnknownFormatException``` is thrown with develop alone (this PR just exposes the problem).  The fix is very similar to https://github.com/openmicroscopy/bioformats/pull/3297/commits/7cdaac054c48ca3e923930348937b1f42b9c97cc, but in ```DeltavisionReader``` instead of ```MRCReader```, so I will add a commit to https://github.com/openmicroscopy/bioformats/pull/3297 so that it's easier to review together.

dgault:
With the fake example listed and https://gist.github.com/melissalinkert/16afa151faa5f076b7f7ffce85597403 applied the tests always pass without this PR included.

With the PR included the test fails as expected. Given that the 3 Deltavision fixes are in place and no further failures have been caused by this PR then this is ready to merge once MRC PR is also merged.

PR #3297, opened 2018-12-20
MRC: fix endian detection in isThisType for old style headers

melissalinkert:
4b771c5 backported from a private PR; 7cdaac0 added to fix test failures.   See the definition of offsets 196-216 in https://bio3d.colorado.edu/imod/doc/mrc_format.txt.  New style headers have an explicit endianness marker, but old style headers do not.

To test, use the file ```mrc/gh-3297/old-header/```, which was hand-created from an existing sample file (see the readme).  Without this change, ```showinf -nopix``` should throw ```UnknownFormatException```.  With this change, ```showinf -normalize``` should result in the file being detected as MRC, with sensible looking pixel data displayed.

While ```mrc/gh-3297/old-header``` has a little-endian example with an old style header, ```mrc/samples/golgi.mrc``` is a big-endian example with an old style header - so we do have to handle both cases.

This is the change that prompted #3296.  I would not expect any test failures, and this should be safe for a patch release if needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -173,6 +173,7 @@ public class DeltavisionReader extends FormatReader {
     final int blockLen = 212;
     if (!FormatTools.validStream(stream, blockLen, true)) return false;
     stream.seek(96);
+    stream.order(true);
     int magic = stream.readShort() & 0xffff;
     boolean valid = magic == DV_MAGIC_BYTES_1 || magic == DV_MAGIC_BYTES_2;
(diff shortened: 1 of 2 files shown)


dgault:
The new check for MRC endianness matches that from the spec in https://bio3d.colorado.edu/imod/doc/mrc_format.txt

Without this PR and using the new sample file results in `UnknownFormatException `. With this PR and using `showinf -normalize` results in the file being read correctly.

The tests have been passing with the new sample file included and we have coverage of both use cases for the old style header.

The Deltavision change also results in tests passing with the PR from https://github.com/openmicroscopy/bioformats/pull/3296

PR #3298, opened 2018-12-20
ND2: fix handling of duplicate and incomplete exposure time lists

melissalinkert:
See https://trello.com/c/UnDryDi9/113-fix-exposure-times-for-control003nd2

https://github.com/openmicroscopy/data_repo_config/pull/353 updates the configuration for public QA 9507 to match the expected ground truth as recorded in https://github.com/openmicroscopy/bioformats/pull/3292#issuecomment-448717313

Exposure times are defined multiple times in multiple places.  This reduces the duplicates so that consecutive times correspond to consecutive channels, and handles the case when the exposure time list is larger than the ```SizeC``` for a single series, but smaller than the sum of ```SizeC``` for all series.

Not sure if this will cause any other test failures, but will keep an eye on tomorrow's build.  If there are additional test failures, it's possible that the configurations would need to change instead of this PR.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -2119,10 +2119,16 @@ public class NativeND2Reader extends SubResolutionFormatReader {
     }
 
     // populate PlaneTiming and StagePosition data
-    if (handler != null && handler.getExposureTimes().size() > 0 &&
-      (exposureTime.size() == 0 || exposureTime.size() % getSizeC() != 0))
-    {
-      exposureTime = handler.getExposureTimes();
+    if (handler != null && handler.getExposureTimes().size() > 0) {
+      if (exposureTime.size() == 0 || handler.getExposureTimes().size() == 1 || exposureTime.size() % getSizeC() != 0) {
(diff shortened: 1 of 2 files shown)


melissalinkert:
Not unexpectedly, this introduces several more test failures: https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/46163/console

I will add screenshots from NIS Elements to https://github.com/openmicroscopy/data_repo_config/pull/353, since most of the failing files are not public.

components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -2138,9 +2168,14 @@ else if (tsT.size() == getSizeZ()) {
           }
 
           int index = i * getSizeC() + coords[1];
-          if (exposureTime.size() == getSizeC()) {
+          if (exposureTime.size() >= getSizeC() && exposureTime.size() < getSizeC() * getSeriesCount()) {
             index = coords[1];
           }
+          else if (exposureTime.size() == 1) {
+            // if exposure times are the same for all channels,
+            // then sometimes only one value is stored
```
mtbc: :+1: for the comment, that's helpful here.

PR #3299, opened 2018-12-20
Zeiss CZI: handle invalid JPEG-XR tiles

melissalinkert:
See QA 27169.

The code in the QA issue indicates that several tiles in the lower right corner of the largest image (the file contains a pyramid) cannot be read.  Checking the actual compressed bytes in each tile indicates that a JPEG-XR stream is present, but invalid.  One example is the tile whose upper left corner is (23216, 38012).  Opening the file in ZEN reliably produces this error the first time the mouse is moved over one of the affected tiles:

zen-screenshot

Subsequent mouse-overs show pixel values of 0 for the last channel.

With this change, the test code in QA and reading affected tiles via ```showinf -crop``` should all work without an exception being thrown.  The wrapped ```ome.jxrlib.FormatError``` should be logged at ```WARN```, though, so it's clear that decodes were not successful.

The QA report mentions making the sample file public, but I have not yet done this as there is no contact information for the submitter.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -3709,7 +3709,13 @@ public class ZeissCZIReader extends FormatReader {
           options.height = directoryEntry.dimensionEntries[1].storedSize;
           options.maxBytes = options.width * options.height *
             getRGBChannelCount() * bytesPerPixel;
-          data = new JPEGXRCodec().decompress(data, options);
+          try {
+            data = new JPEGXRCodec().decompress(data, options);
+          }
+          catch (FormatException e) {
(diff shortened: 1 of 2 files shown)


mstritt:
thanks a lot for taking care of this.
The images was provided by Gebhard Stopper, Uni Saarland.

dgault:
Tested using QA-27169 and `showinf -crop` with tile positions listed.

Without the PR results in the Exception being thrown and the image is unable to be read:
`Exception in thread "main" ome.jxrlib.FormatError: ERROR: Could not get image bytes: -1`

With the PR included the image opens and metadata is displayed. The Exception is handled and the new message is displayed:
`Could not decompress block; some pixels may be 0`
The values for the first 2 channels at the given position are correct and the third channel has 0 value for the invalid tile as expected.

PR is good to merge.

PR #3301, opened 2019-01-05
Simplified bfGetPlane.m, avoids a data copy

crisluengo:
The original code contains a transpose operation `'` for unsigned integer images (i.e. most image files). This transpose incurs a data copy that is not necessary. This commit avoids this data copy by using the Java method `makeDataArray2D` also for unsigned int images, the resulting 2D array is reshaped to a vector to call `typecast`, and then reshaped back (reshaping does not copy data).


dgault:
Thank you for the contribution @crisluengo 

Having carried out some initial testing the changes here keep the data unchanged from the original code but as you say does avoid the need for the transpose and the extra copy. I don't believe there is ever a scenario when `makeDataArray2D` is not available but `makeDataArray` is, as such the change looks to be safe from any breaking behaviour.

Adding this for inclusion with the upcoming Bio-Formats 6.0.0

PR #3303, opened 2019-01-14
Solve error for big .fli files

rharkes:
Big files gave an integer overflow on line 191.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
@@ -188,7 +188,7 @@ public class LiFlimReader extends FormatReader {
       }
     }
     else {
-      in.seek(dataOffset + bytesPerPlane * no);
+      in.seek(dataOffset + (long) bytesPerPlane * (long) no);
 
       int thisSeries = getSeries();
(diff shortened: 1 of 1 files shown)


sbesson:
Thanks @rharkes. Including the changes for the daily CI builds. We do not have any representative `.fli` file that shows such an integer overflow. Would you have one you would be able to share with us so that we could add it to our non-regression tests?

rharkes:
@sbesson Yes I have one, but it is 2.3 GB in size, since the problem does not occur below 2GB.

melissalinkert:
Thank you for uploading sample files, @rharkes.

@openmicroscopy/formats, files copied to ```inbox/gh-3303/``` for testing.

dgault:
In this particular case the Integer overflow is being caused by the large number of timepoints as opposed to the bytes per plane. As such the change makes sense and other instances of plane size do not need to be updated.

I tested using the large sample file provided.

Without the PR opening the image and reading one of the later timepoints results in:
`java.lang.IllegalArgumentException: Negative position`

With the PR included I was able to open and read all of the planes within the image without issue.

This is good for inclusion in Bio-Formats 6.0.0

PR #3306, opened 2019-01-16
TIFF: ignore FillOrder unless uncompressed or fax compressed

melissalinkert:
See:

https://trello.com/c/NFuiPd4s/45-large-tiff-0-pixel-value
https://trello.com/c/7pSOeMJF/46-big-tiff-breakage-in-571
https://trello.com/c/fKOcj8sd/47-tiff-incorrect-fillorder

and in particular rationale for this change in https://trello.com/c/fKOcj8sd/47-tiff-incorrect-fillorder#comment-5c1aca32a935af64b5348fb2

Test files are ```curated/tiff/darwinjob/SESSION4_Image2.tif```, ```curated/tiff/qa-20514/819_2.tif```, and ```/ome/team/mlinkert/tiff-fill-order/stephane-dallongeville/716_2.tif``` (not copied due to permissions issues).  All have large XY dimensions, and should be completely black without this PR.  With this PR, all should look like regular whole slide images with white backgrounds.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -779,7 +779,9 @@ public class TiffParser {
 
     // reverse bits in each byte if FillOrder == 2
 
-    if (ifd.getIFDIntValue(IFD.FILL_ORDER) == 2) {
+    if (ifd.getIFDIntValue(IFD.FILL_ORDER) == 2 &&
+      compression.getCode() <= TiffCompression.GROUP_4_FAX.getCode())
+    {
       for (int i=0; i<tile.length; i++) {
         tile[i] = (byte) (Integer.reverse(tile[i]) >> 24);
       }


dgault:
From  https://www.adobe.io/content/dam/udp/en/open/standards/tiff/TIFF6.pdf as mentioned on the Trello card the logic for FillOrder 2 should be:

```
FillOrder = 2 should be used only when BitsPerSample = 1 and the data is
either uncompressed or compressed using CCITT 1D or 2D compression, to
avoid potentially ambigous situations.
```
Though as also mentioned the `BitsPerSample = 1 ` does contradict the sample files.

With the PR included I was able to read the affected sample files and they displayed without error. The original logic is maintained for uncompressed and the 3 CCITT compression types, which appears to be correct from the PDF. The PR has not caused any regressions in the repo tests or builds.

sbesson:
See https://forum.image.sc/t/tiff-file-fails-to-open/50416 and notably https://forum.image.sc/t/tiff-file-fails-to-open/50416/3 for the case of a TIFF file with Deflate compression and `FillOrder` 2.

It might be worth trying to update the conditional test here to include the Deflate compression code and test that nothing else is broken in the non-regression tests?

PR #3308, opened 2019-01-22
Support for packed UNINT12 datatype

LambertJohan:
I have made an adaptation on LiFLIMReader.java that now checks for UNINT12 Data and unpacks this to 16 bitt data.
It also checks if there is a dark image, If true, it will be appended to the Dataset.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
@@ -58,6 +58,7 @@ import ome.units.quantity.Time;
 import ome.units.unit.Unit;
 import ome.units.UNITS;
 
+
 /**
  * LiFlimReader is the file format reader for LI-FLIM files.
  */
@@ -84,7 +85,7 @@ public class LiFlimReader extends FormatReader {
   public static final String F_KEY = "frequencies";
   public static final String T_KEY = "timestamps";
   
-  // DarkImages is different w.r.t. background table,
+  // New Metadatakey for cameras that have a dark image.
   public static final String DarkImage_KEY = "hasDarkImage";
 
   // relevant keys in timestamp table
@@ -106,7 +107,6 @@ public class LiFlimReader extends FormatReader {
   public static final String DATATYPE_INT32 = "INT32";
   public static final String DATATYPE_REAL32 = "REAL32";
   public static final String DATATYPE_REAL64 = "REAL64";
-  // added new datatype 22-01-2019
   public static final String DATATYPE_UINT12 = "UINT12";
   
   // -- Fields --
@@ -158,8 +158,8 @@ public class LiFlimReader extends FormatReader {
   private int gzSeries;
   
   /** if DataType is UINT12, data is compressed in 12 bits.
-   * Similar to Gzipstream make use of a DataInputStream 
-   * to blow up the compressed UINT12 pixel data. */
+  *   Similar to Gzipstream make use of a DataInputStream 
+  *   to unpack the compressed UINT12 pixel data into 16 bits. */
   private boolean TypeUINT12;
   private DataInputStream UINT12stream;
   private int UINT12streamPos;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/LiFlimReader.java
```diff
@@ -376,6 +421,11 @@ private void initCoreMetadata() throws FormatException {
     ms.sizeZ = Integer.parseInt(zLen) * sizeF;
     ms.sizeC = Integer.parseInt(channels);
     ms.sizeT = Integer.parseInt(timestamps) * sizeP;
+    
+    /** Check for DarkImage, if "1" increment Timestamps with 1 such that DarkImage is appended to the stream  */
+    if (DarkImage.equals(new String("1"))) {
```
sbesson: See https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/51263/console this causes a NPE for  sample images where `DarkImage` is not defined so the nullity needs to be checked

LambertJohan: Hello Sébastian, fixed it and added some more fixes.

sbesson:
Hi @LambertJohan, some initial functional testing feedback using the sample files uploaded in the original [ome-users thread](http://lists.openmicroscopy.org.uk/pipermail/ome-users/2018-December/007308.html) as well as in https://github.com/openmicroscopy/bioformats/pull/3303#issuecomment-454145605.

Testing was done by reading the 2 FLIM filesets using the Bio-Formats command-line tools and importing them into a test OMERO server with a Bio-Formats library including this PR. As shown below, all three filesets imported without issues:

screen shot 2019-01-28 at 15 09 09
screen shot 2019-01-28 at 15 09 34

Two initial comments up for discussion:

- the maximum intensity is detected to be 25952 while I would have expected maximum values to be 4095 for such a data type
- the dark image is correctly detected and read by the reader. At the moment, this plane is simply appended along the time dimension leading to a black plane when scrolling to the end of the time dimension. 

screen shot 2019-01-28 at 22 05 26
 
More than a UI issue, this representation means it is no longer possible to interpret `sizeT == true number of timepoints x  moduloT`  and clients would need  extra code to detect and handle the final plane(s). Would it make sense to detect this dark image as another separate series as the reader does for background images?

As the assumptions discussed in the last comment also affects the automated tests for the some of the [LI-FLIM representative files](https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/51899/console), temporarily excluding this PR from the daily build to review #3312 against the whole set of formats but let's keep the back-n-forth on the best way to get both of these features (new data type, dark image) integrated.

LambertJohan:
@sbesson 
Hello Sebastien, thank you very much for giving it a test.

W.r.t. your comments:
- The 12-bit data is upscaled to 16-bit because I use the 16bit datatype otherwise the image would always look dark. 
- I had problems with getting the reader to put the dark image in a new series.  I agree with you that it would be much more convenient if it is handled as another series. Maybe you have suggestions about how I can improve the code?

sbesson:
Hi @LambertJohan,

- re 12-bit data, the way to communicate a smaller dynamic pixel range than the one given by the pixel type while returning the data value on disk is to set `CoreMetadata.bitsPerPixels` attribute - to 12 in that case. Clients can then use the `IFormatReader.getBitsPerPixels` API and use it to se the rendering window accordingly - see https://github.com/openmicroscopy/bioformats/blob/v6.0.0-m4/components/bio-formats-plugins/src/loci/plugins/in/Colorizer.java#L278 for the case of the ImageJ plugin.

- adding another series in a reader can be achieved by creating and another `CoreMetadata` object and adding it to the `core` list. For the LIFlimReader, the relevant code handling the optional background image is in https://github.com/openmicroscopy/bioformats/blob/6.0.0-m4/components/formats-gpl/src/loci/formats/in/LiFlimReader.java#L397. I assume a similar conditional block can be added to handle dark images.

LambertJohan:
Hello @sbesson,

Done as you suggested, please have a look at the code and let me know if it can be included in the next release :) Many thanks in advance!

dgault:
Thanks @LambertJohan, we have been very busy prepping the release of Bio-Formats 6.0.0 the past couple of days. We will have a final review of this PR next week and hope to include it in Bio-Formats 6.1.0

dgault:
Retested again and the dynamic range is updated as expected and the dark images are now being handled correctly. This has also impacted some of the other sample files in our repo which I have tested and are correct. With the new data repo config for the updated dark image included all the tests have green.

Merging for inclusion in 6.1.0

PR #3309, opened 2019-01-23
TIFF: restrict direct tile copying to cases where only one tile is read

melissalinkert:
Backported from a private PR.

If multiple tiles need to be read, then direct copying would have resulted in invalid images.  This bug was only exposed in cases where multiple tiles were read along X, but the requested image width matched the tile width.  Any tiled TIFF can be used to test, but SVS files are the most obvious choice.

Without this change:

- ```showinf svs/alexandra/UMD001_ORO.svs -crop 0,0,240,240``` should look correct
- ```showinf svs/alexandra/UMD001_ORO.svs -crop 10,0,240,240``` should have 10 rows of black pixels at the bottom of the image
- ```showinf svs/alexandra/UMD001_ORO.svs -crop 10,0,239,240``` should look correct

With this change, all 3 tests above should result in correct images.  

edc8d5c alone combined with develop should result in a unit test failure.  Adding d9818b4 should allow the tests to pass again.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -1081,10 +1081,13 @@ public class TiffParser {
             outputRowLen * (tileY - y);
           if (planarConfig == 2) dest += (planeSize * (row / nrows));
 
-          // copying the tile directly will only work if there is no overlap;
+          // copying the tile directly will only work if there is no overlap
+          // and only one tile needs to be read
           // otherwise, we may be overwriting a previous tile
           // (or the current tile may be overwritten by a subsequent tile)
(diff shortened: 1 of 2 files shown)


dgault:
Code changes look fine and the new unit is a good addition which covers this scenario. 

Tested using `svs/alexandra/UMD001_ORO.svs`. Running the suggested commands without the PR results in the second command producing an incorrect image:
```
showinf svs/alexandra/UMD001_ORO.svs -crop 0,0,240,240 
showinf svs/alexandra/UMD001_ORO.svs -crop 10,0,240,240 
showinf svs/alexandra/UMD001_ORO.svs -crop 10,0,239,240 
```

Running all 3 again with the PR results in correct images for each.

Running the new unit test without the PR results in failure. With the PR included it passes as expected.

PR #3312, opened 2019-01-25
Update SubResolutionFormatReader to extend FormatReader

melissalinkert:
```FormatReader``` now implements and uses ```currentCore()``` whenever the ```CoreMetadata``` for the current series/resolution is used.  This allows ```SubResolutionFormatReader``` to extend ```FormatReader```, which means that hundreds of lines of duplicate code are removed.

```SubResolutionFormatReader``` continues to override all of the methods specific to series and resolution handling, and uses a ```CoreMetadataList``` instead of ```List<CoreMetadata>``` to store metadata internally.

See https://trello.com/c/5JvwTpLa/6-formattools-refactorings

Builds should continue to pass, but memo files may be affected.  As noted in https://trello.com/c/5JvwTpLa/6-formattools-refactorings, ```AFIReader```, ```CellSensReader```, and ```ZeissCZIReader``` can be updated to use ```SubResolutionFormatReader```, but I will do this in a follow-up PR once this one is merged.

### CODE REVIEW

File: components/formats-api/src/loci/formats/FormatReader.java [hunk 1]
```diff
@@ -1466,6 +1466,15 @@ public void close() throws IOException {
     close(false);
   }
 
+  /**
+   * Get the CoreMetadata corresponding to the current series and resolution
+   *
+   * @return the CoreMetadata
+   */
+  protected CoreMetadata currentCore() {
```
sbesson: Since this helper is technically a getter, should it be called `getCurrentCore` for consistency?

melissalinkert: Done in d5e6abc.

PR #3315, opened 2019-01-31
Olympus OIR: compare absolute file paths when reading pixels files

melissalinkert:
Backported from private PR; opening to check test impact, but not expecting this to be included in 6.0.0 (can exclude after one test run if needed).

This ensures that the file path is normalized, so that the file
separators will be the same across all files.  This prevents weird
problems on Windows where passing ```/path/to/file.oir``` throws an
exception but ```\path\to\file.oir``` is fine.

Without this PR and any single-file .oir on Windows:

```
PS E:\tmp>./showinf ./file.oir
```

should throw an exception but ```./showinf .\file.oir``` should successfully read all planes.

With this change, both tests should successfully read all planes.

The incorrect behavior is more immediately obvious in Git Bash, since tab-completing the file name uses ```/```, while tab-completing in PowerShell replaces ```/``` with ```\```.  The exception above can be reproduced in PowerShell, though, just take care to type/paste the file name with ```/``` separators and don't touch the tab key.

Memo files and non-Windows behavior should be unaffected; this should be fine for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -614,7 +614,7 @@ public class OIRReader extends FormatReader {
     // seek past reference image blocks
     while (skipPixelBlock(file, s, false));
 
-    if (s.getFilePointer() == pixelStart && !file.endsWith(currentId)) {
+    if (s.getFilePointer() == pixelStart && !isCurrentFile(file)) {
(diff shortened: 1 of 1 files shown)


dgault:
This change definitely makes sense. Searching the code base I couldn't find any other readers with similar logic that might also be affected. With the PR included OIR files are successfully read on Windows when using either `/` or `\`

PR #3319, opened 2019-02-11
MSR: Add null pointer check for "xyz-Table Z Resolution"

dgault:
This was reported in http://lists.openmicroscopy.org.uk/pipermail/ome-devel/2019-February/004320.html and can be tested using QA-27309

With the QA file the value of "xyz-Table Z Resolution" is "false", which throws a NullPointerException when parsing the value. This PR adds a check for the null pointer and ignores the value.

To test, use QA-27309, with this PR included the file should open and display without any exception being thrown.


dgault:
Config added: https://github.com/openmicroscopy/data_repo_config/pull/370

PR #3328, opened 2019-03-13
LEOReader: length parsing fixes and global metadata support

sbesson:
#3178 introduced more systematic parsing of the INI-style metadata stored in the private TIFF tag 34118.

As reported by our community, some of these files are now failing to open in Bio-Formats 5.9.1 and later due to the strictness of the unit detection in:

https://github.com/openmicroscopy/bioformats/blob/388e21ea44d12e53226407a2ee7a5966bb0e0dc3/components/formats-gpl/src/loci/formats/in/LEOReader.java#L102-L103

To fix this problem, this PR:

- migrate  the `parseLength()` logic from `FakeReader` as a public `FormatTools` API including some unit test coverage
- updates `LEOReader` and `FakeReader` to use this API in order to create lengths. For LEO files, this means the reader should handle any length unit supported by the model .

Additionally:
- the key/value pairs following all lines starting with `AP_`, `DP_` and `SV_` are parsed and added to the Global metadata table
- one more acquisition date format is added informed by the data of our available sample files

Files from QA 27408 and 27409 should be tested with this change. Before this change, the `setId` call should fail with a parsing exception. With this change, files should read, the global metadata should contain `Image Pixel Size` keys and both the physical size and the acquisition date should be populated in the OME-XML metadata.

In addition to fixing the issue above, the physical sizes should now use the native units rather than being systematically converted into microns. Many more key/value pairs should now be available in the Global Metadata table for al the files under `curated/leo`. This PR should be suitable for a minor release.


dgault:
Adding the `parseLength` to `FormatTools` makes sense and works as expected. The new tests for `FormatTools` and `FakeReader` run correctly and provide good test coverage.

For the new sample files `QA-27408` and `QA-27409`, without the PR included opening the files results in an exception:
`java.lang.NumberFormatException: For input string: "785.0 pm"`

With the PR included the files are read without issue.

Going through the existing files in the repo which are impacted, it appears that the only changes are updates to the physical sizes which are now using nanometers for some for the files. These are all correct.

Also reading LEO files now results in a much greater amount of global metadata, all of which looks sensible and correct.

PR #3331, opened 2019-03-22
Use file-leak-detector to check for open file handles

melissalinkert:
See https://trello.com/c/3rSiAj6J/286-check-for-leaking-file-descriptors

This adds https://github.com/kohsuke/file-leak-detector as a dependency of ```test-suite```, and introduces a file leak test that will run during ```test-automated```.  The test compares the open file handles before and after the full test suite is run, so the number of threads used shouldn't affect the results.

Local testing showed a couple of obvious potential leaks in the ```isThisType``` methods in ```FormatReader```.  There was also a bug in how the thread count was passed to the tests, which should be fixed by 3207c2a (and may also help to reduce occasional ```OutOfMemoryError```s in the repo tests).

As noted in the above card, this does check in the ```file-leak-detector``` jar since the tests are run with Ant.  If anyone has a better idea for how to do this, I'm happy to change it.


melissalinkert:
Build failures with Java 9+ due to https://openjdk.java.net/jeps/220 and https://github.com/kohsuke/file-leak-detector/blob/file-leak-detector-1.13/pom.xml#L138.  Excluding the ```com.sun:tools``` transitive dependency seems to pass the build (locally, anyway), and should be OK since we aren't running the tests via Maven.

It looks like a relevant PR was merged (https://github.com/kohsuke/file-leak-detector/pull/32) some time ago, but it's unclear to me why that isn't included in 1.13.

components/test-suite/build.xml
```diff
@@ -267,10 +267,12 @@ Type "ant -p" for a list of targets.
       <sysproperty key="testng.multiplier" value="${testng.multiplier}"/>
       <sysproperty key="lurawave.license" value="${lurawave.license}"/>
       <sysproperty key="testng.in-memory" value="${testng.in-memory}"/>
+      <sysproperty key="testng.threadCount" value="${testng.threadCount}"/>
       <jvmarg value="-mx${testng.memory}"/>
       <jvmarg value="-Duser.language=${user.language}"/>
       <jvmarg value="-Duser.country=${user.country}"/>
       <jvmarg value="-Dlogback.configurationFile=${logback.configurationFile}"/>
+      <jvmarg value="-javaagent:${root.dir}/jar/file-leak-detector-1.13.jar=strong"/>
```
sbesson: From http://maven.apache.org/ant-tasks/examples/dependencies.html, 

```diff
--- a/components/test-suite/build.xml
+++ b/components/test-suite/build.xml
@@ -272,7 +272,7 @@ Type "ant -p" for a list of targets.
       <jvmarg value="-Duser.language=${user.language}"/>
       <jvmarg value="-Duser.country=${user.country}"/>
       <jvmarg value="-Dlogback.configurationFile=${logback.configurationFile}"/>
-      <jvmarg value="-javaagent:${root.dir}/jar/file-leak-detector-1.13.jar=strong"/>
+      <jvmarg value="-javaagent:${org.kohsuke:file-leak-detector:jar}=strong"/>
     </testng>
     <fail if="failedTest"/>
   </target>
```

should allow to consume the JAR without committing it to the source code. Testing it locally I get an error when lauching the JVM:

```
test-automated:
   [testng] Error occurred during initialization of VM
   [testng] agent library failed to init: instrument
   [testng] objc[95933]: Class JavaLaunchHelper is implemented in both /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/bin/java (0x10378c4c0) and /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/lib/libinstrument.dylib (0x1057ba4e0). One of the two will be used. Which one is undefined.
   [testng] Failed to find Premain-Class manifest attribute in /Users/sbesson/.m2/repository/org/kohsuke/file-leak-detector/1.13/file-leak-detector-1.13.jar
   [testng] The tests failed.
```

Indeed, the manifest of the JAR is different between the one retrieved via Maven and the one checked out locally

```shell
sbesson@ls30630:bioformats ((bdccebd6da...)) $ unzip -qc jar/file-leak-detector-1.13.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Premain-Class: org.kohsuke.file_leak_detector.AgentMain
Archiver-Version: Plexus Archiver
Built-By: kohsuke
Agent-Class: org.kohsuke.file_leak_detector.AgentMain
Can-Redefine-Classes: true
Can-Retransform-Classes: true
Created-By: Apache Maven 3.5.0
Build-Jdk: 1.8.0_172
Main-Class: org.kohsuke.file_leak_detector.Main

sbesson@ls30630:bioformats ((bdccebd6da...)) $ unzip -qc /Users/sbesson/.m2/repository/org/kohsuke/file-leak-detector/1.13/file-leak-detector-1.13.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Built-By: kohsuke
Created-By: Apache Maven 3.5.0
Build-Jdk: 1.8.0_172
```

melissalinkert: Thanks for the pointers.  The ```jar-with-dependencies``` is what was checked in, so that explains the difference.

4560e1b updates the POM to depend upon the ```jar-with-dependencies``` specifically, switches ```build.xml``` to use the relevant property, and removes the committed ```file-leak-detector-1.13.jar```.  That all seems to work locally, but we'll see what happens with the builds tonight.

components/formats-gpl/src/loci/formats/in/NDPIReader.java
```diff
@@ -131,14 +131,15 @@ public int fileGroupOption(String id) throws FormatException, IOException {
...
-      tiffParser.getStream().close();
-      return b;
+      try (RandomAccessInputStream s = new RandomAccessInputStream(currentId)) {
+        tiffParser = new TiffParser(s);
+        tiffParser.setUse64BitOffsets(true);
+        tiffParser.setYCbCrCorrection(false);
+        return tiffParser.getSamples(ifds.get(ifdIndex), buf, x, y, w, h);
+      }
+      finally {
+        tiffParser.getStream().close();
```
joshmoore: Not for this PR (and I haven't looked into the details) but this idiom makes me think that TiffParser should be made Closeable to enable:

```
  try (TiffParser tiffParser = new TiffParser(currentId)) {
      tiffParser.setUse64BitOffsets(true);
      tiffParser.setYCbCrCorrection(false);
      return tiffParser.getSamples(ifds.get(ifdIndex), buf, x, y, w, h);
  }
```

sbesson: At a quick glance, there are a few places that could benefit from such a logic

```
sbesson@ls30630:bioformats (develop) $ git grep getStream.*close
components/formats-gpl/src/loci/formats/in/APLReader.java:            p.getStream().close();
components/formats-gpl/src/loci/formats/in/CellSensReader.java:        parser.getStream().close();
components/formats-gpl/src/loci/formats/in/NDPIReader.java:      tiffParser.getStream().close();
components/formats-gpl/src/loci/formats/in/NDPIReader.java:        tiffParser.getStream().close();
components/formats-gpl/utils/WritePrecompressedPlanes.java:    saver.getStream().close();
```

Extending `TiffParser` (and `TiffSaver`?) to be `Closeable` is certainly something worth discussing in the context of the next minor release especially given the emphasis on the file leak. I do not know of obvious caveats to doing so.

melissalinkert:
Two follow-up cards added: https://trello.com/c/ZLAScq9y/367-deprecate-remove-filehandletest, https://trello.com/c/JHikBmdv/368-make-tiffparser-and-tiffsaver-closeable

PR #3332, opened 2019-03-22
Fake: remove "header" key from original metadata

melissalinkert:
See https://trello.com/c/4tlJlMuC/339-fake-ini-handling-leads-to-spurious-header-key

To test, use a simple dataset like this:

```
$ cat test.fake.ini
[GlobalMetadata]
foo = bar
$ touch test.fake
```

Without this PR, ```showinf -nopix test.fake``` should show two key/value pairs in the global metadata table: ```foo: bar``` and ```header: GlobalMetadata```.  With this PR, only ```foo: bar`` should be present.

Should be safe for a patch release if needed, and very low priority.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -620,7 +620,9 @@ public class FakeReader extends FormatReader {
       table = list.getTable("GlobalMetadata");
       if (table != null) {
         for (Map.Entry<String, String> entry : table.entrySet()) {
-          addGlobalMeta(entry.getKey(), entry.getValue());
+          if (!IniTable.HEADER_KEY.equals(entry.getKey())) {
+            addGlobalMeta(entry.getKey(), entry.getValue());
+          }
         }
       }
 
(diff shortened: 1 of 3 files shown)


melissalinkert:
Readers that use INI parsing at all (per ```git grep IniParser```):

```
FakeReader
BDReader
FEITiffReader
FV1000Reader
HamamatsuVMSReader
HitachiReader
LiFlimReader
SISReader
SimplePCITiffReader
TillVisionReader
```
Most seem to use ```IniList.flattenIntoHashMap()``` which filters the header keys automatically (https://github.com/ome/ome-common-java/blob/master/src/main/java/loci/common/IniList.java#L86), or already specifically exclude the header key (e.g. https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/LiFlimReader.java#L305)

It looks like only ```HamamatsuVMSReader``` and ```HitachiReader``` have the same problem with header keys, so 49e19bf has a similar fix for those readers.

If we want something more advanced instead, some options are:

- add ```FormatTools``` method to copy an ```IniTable``` into original metadata, omitting the header key
- rework the INI parsing classes so that ```IniTable``` stores the header name separately and not in the ```HashMap```
- override ```keySet()``` in ```IniTable``` so that the header key is excluded (but can still be accessed directly via ```get("header")```)

joshmoore:
Thanks, Melissa. :+1: on catching the other readers. I don't have an overly strong opinion on any of your suggestions, other than the standard "breaking changes" mantra.

sbesson:
Looking at the suggestions in https://github.com/openmicroscopy/bioformats/pull/3332#issuecomment-475772395, the only other option I could think of was an API of type `IniList.get(String, boolean)` which would selectively return the `IniTable` with/without the header key depending on the boolean value.

PR #3337, opened 2019-04-01
Gatan DM3/DM4: parse "Montage" tag to determine if tiles are present

melissalinkert:
Backported from a private PR.

To test, use ```curated/gatan/samples/test-tiles.dm3```, which was created using Gatan's GMS 3 (DigitalMicrograph Offline) software (http://www.gatan.com/installation-instructions).  The file has 4 planes, and a ```Montage``` tag structure that was copied from a private file that was expected to contain 4 planes each representing one tile.

Without this PR, ```showinf test-tiles.dm3``` should result in a single series with 4 Z sections.  With this PR, the same test should result in 4 series each with a single plane.  Position data for each plane should be present in the OME-XML.

Only one existing file in the repo (```gatan/qa-11392/Example_Gatan_File.dm4```) has a ```Montage``` tag, but it only has one plane so there should be no change in dimensions for any existing files.  Should be safe for a patch release (no memo file changes), but not urgent.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -342,7 +342,7 @@ public class GatanReader extends FormatReader {
 
       int digits = String.valueOf(getSeriesCount() + 1).length();
       for (int i=0; i<getSeriesCount(); i++) {
-        if (foundMontage) {
+        if (foundMontage && getSeriesCount() > 1) {
           store.setImageName(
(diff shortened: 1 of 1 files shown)


dgault:
Tested using `curated/gatan/samples/test-tiles.dm3`:
Without the PR results in a single series with 4 Z planes.
With the PR results in 4 series with a single plane. Each plane displays as expected (gradient from left to right)

The PR doesn't have any impact on any of the existing sample files we have and builds and tests have all been green.

PR #3338, opened 2019-04-07
Upgrade jhdf5 dependency to version 14.12.6

sbesson:
See https://github.com/scijava/pom-scijava/issues/87

Initially opening to assess the impact of the `jhdf5` dependency bump to the latest patch release.

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -170,9 +170,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>ch.systems.cisd</groupId>
+      <groupId>cisd</groupId>
       <artifactId>jhdf5</artifactId>
-      <version>14.12.0</version>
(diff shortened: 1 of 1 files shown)


sbesson:
https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-test-folder/62756/console looks very positive i.e. no obvious regression flagged by automated tests. I will be carrying some minimal functional testing before listing this for review.

joshmoore:
Discussed today getting this into a quick release. [Current versioning policy](https://docs.openmicroscopy.org/bio-formats/6.0.0/about/index.html#versioning-policy) specifies that that can't be a patch release, so this will be targeted for 6.1.0. (It may make most sense to roll the Guava bump in at the same time.)

To facilitate the release, `develop` should be kept extra releasable. Outstanding is testing in ImageJ.

PR #3339, opened 2019-04-08
Use Guava from OME Common Java.

mtbc:
@sbesson points out that there is no need to depend on Guava again in Bio-Formats and have to keep versions matching.


sbesson:
The few failures from the daily automated tests are unrelated to this change. Merging as this dependency should be pulled transitively from the upstream `ome-common` dependency.

Issue #3343, opened 2019-04-09
showinf crashing on TIFs created from VSI images by bfconvert 6.0.1


ebremer:
Using:
bfconvert -series 6 -pyramid-scale 2 -pyramid-resolutions 4 image.vsi image.tif
runs without error.

Using:
showinf image.tif results in
Exception in thread "main" java.lang.NullPointerException
        at loci.formats.in.APLReader.isThisType(APLReader.java:98)
        at loci.formats.ImageReader.getReader(ImageReader.java:188)
        at loci.formats.ImageReader.getFormat(ImageReader.java:156)
        at loci.formats.tools.ImageInfo.configureReaderPreInit(ImageInfo.java:433)
        at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1030)
        at loci.formats.tools.ImageInfo.main(ImageInfo.java:1121)

GIMP can read the TIF and show the pyramid.

joshmoore:
Hi @ebremer. Sorry for the troubles. This is a known problem that should be fixed in the next release. If you'd like to workaround, passing `-format CellSens` to `showinf` will likely help. If that doesn't, calling `showinf` with an absolute path may: `showinf /my/data/image.vsi`

You can take a look at https://github.com/openmicroscopy/bioformats/pull/3010 and the linked PRs if you'd like more information.

Cheers,
~Josh

ebremer:
I'm not running the showinf on the image.vsi but the bfconverted image.tif.  The problem did not occur when I used 6.0.0-m4.

joshmoore:
> I'm not running the showinf on the image.vsi but the bfconverted image.tif.

Understood. Then replace vsi with tif in my comments. The underlying problem here is that in trying to detect the format of the file given (vsi or tif or something else) one of the readers is misbehaving.

>  The problem did not occur when I used 6.0.0-m4.

Cheers.

ebremer:
Unfortunately, neither helped.  The error is different though.

D:\vsi>showinf -format CellSens d:\vsi\him4.tif
Checking CellSens VSI format [no]
Initializing reader
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: -2
        at java.lang.String.substring(Unknown Source)
        at loci.formats.in.CellSensReader.initFile(CellSensReader.java:630)
        at loci.formats.FormatReader.setId(FormatReader.java:1395)
[...]

PR #3344, opened 2019-04-10
Make ImageReader defensive against reader errors

joshmoore:
Small programming errors in potential readers (e.g.
a NullPointerException during isThisType) can potentially
cause a file of a different type to become unreadable.
ImageReader should catch all such exceptions and considers
readers that throw such exceptions invalid.

fix gh-3343

### DIFF
--- a/components/formats-api/src/loci/formats/ImageReader.java
+++ b/components/formats-api/src/loci/formats/ImageReader.java
@@ -40,6 +40,7 @@ import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Definitely :+1: to making NPE-type issues less burdensome, but does this mean that repo tests will no longer fail on any ```isThisType```-related regressions?

joshmoore:
Hmmm... good point. Shall I add a system property  check? And if so, is there already one that would cover “test running”?

melissalinkert:
I guess a quick fix would be to check the ```testng.directory``` system property, but I don't know that that's a good idea long-term.  Maybe better to add a format-independent ```MetadataOptions``` key and check that (and then update the tests to set the option)?

joshmoore:
Longer-term I could see setting an error handler on the metadata options which could be a step on the path to leniency. I don't see any format-independent options listed on https://docs.openmicroscopy.org/bio-formats/6.0.1/formats/options.html?highlight=options -- prefix with `bioformats.`, `internal.`, ... ?

melissalinkert:
I'd probably lean towards ```bioformats.```, but no really strong opinion.

joshmoore:
Or `core`.

melissalinkert:
The scenario I had in mind was where a reader's ```isThisType``` is changed such that it accidentally throws an exception on a file where it should return ```false```.  For example, a TIFF-based reader checking the value of a particular IFD tag but not checking if the tag is missing.

On second look, you're right that ```FormatReaderTest.testIsThisType``` should be covering that case, though.  No objection to getting this merged.

PR #3347, opened 2019-04-11
Add FormatTools methods for creating well names

melissalinkert:
9631911 is backported from a private PR.

This adds ```String getWellName(int row, int col)``` and ```String getWellRowName(int row)``` to ```FormatTools```, to unify the logic for creating a well name from the row and column indexes.  This was prompted by a plate with more than 26 rows, which caused weird ```Image``` names when using the standard ```'A' + rowIndex``` logic.  I'm open to other ideas on how to name rows beyond the 26th; what is here just seemed like the most obvious strategy.

I wouldn't expect any test failures, as existing ```Image``` names should not change and I don't think we have any plates larger than 384 wells at the moment.  This is just low priority cleanup and future-proofing, so feel free to exclude as needed.


melissalinkert:
Excluding for now, since this conflicts with https://github.com/openmicroscopy/bioformats/pull/3350 (which has new datasets and config changes).

sbesson:
The new API additions make sense to me and are directly used in various HCS readers with the non-regression tests staying successful. ONly consideration is whether unit tests should also be added to `formats-api` directly. Otherwise LGTM from my side.

melissalinkert:
A few tests added in 969f1c1. I think I convinced myself that it's better to explicitly forbid a negative/row column, but if anyone disagrees can revert that change.

sbesson (approved):
I agree with the proposed behavior of throwing an exception for negative indexes.

dgault:
Builds and tests have remained green with this PR included.

The new API additions look good and the behaviour matches the expected results. Testing manually with sample files from each of the impacted readers and the well names remained consistent with no regressions detected. The size limit seems fine for now and can be extended in the future if ever required. PR looks good to merge for 6.9.0

PR #3348, opened 2019-04-11
TIFF: use fast reading case when image is stored in contiguous strips

melissalinkert:
Backported from a private PR.

This adds a special case to ```TiffParser``` to treat uncompressed strips that are stored in order with no gaps as a single strip.  This should somewhat improve read times for some TIFF files, but likely won't have a noticeable effect upon repo test times.  Any TIFF files that use compression or store tiles instead of strips should be unaffected.

The main thing to verify is that all builds remain green; can definitely exclude in the morning if anything is failing.  There are no API or memo changes, so this should be safe to defer to a patch release.


melissalinkert:
Removing exclude label as discussed with @sbesson.  I will check the build tomorrow and fix or re-exclude as needed.

dgault:
Ive been dogfooding this change for a few weeks now and noticed no noticeable regression while testing various TIFF based formats. For the actual code changes, the logic itself makes sense and looks good and the tests have all been running green with the PR included.

Issue #3242, opened 2018-10-04
[WARN] No pixel block error


naomicxia:
Hello,
I am trying to read .oir files acquired from Olympus 2-P microscope. The files are quite large (2 channels, 15000 timepoints each). When I try to open them with Bio-formats I receive the following error:

Reading pixel blocks from /Volumes//5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1.oir
[WARN] Unknown Immersion value 'SCALEVIEW-A' will be stored as "Other"
Reading pixel blocks from /Volumes//5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1_00001
Reading pixel blocks from /Volumes/5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1_00002
Reading pixel blocks from /Volumes/5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1_00003
Reading pixel blocks from /Volumes/5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1_00004
Reading pixel blocks from /Volumes//5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1_00005
Reading pixel blocks from /Volumes/5-18-18/5-18-18_3001873_m219_wmdelay0to0'5s-1_00006
[WARN] No pixel blocks for plane #2202
[WARN] No pixel blocks for plane #2202
[WARN] No pixel blocks for plane #2330
[WARN] No pixel blocks for plane #2330
[WARN] No pixel blocks for plane #3008

And so on and so forth. It essentially processes the entire 30,000 images but after about 1000 frames there are no more pixels. I tried using the Bioformats command line tool (bfconvert) as well as the Matlab toolbox and replicated the same issue, so I am abit stumped here. Any help would be appreciated!

dgault:
Hi @naomicxia, thanks for opening the issue. This appears to be a new issue which we have not seen reported before, do you see the issue consistently with all your OIR files or is it only this particular fileset which is problematic? Also is it a single OIR file or is it a split multifile dataset?

If you can reproduce the issue with a smaller image then you can upload to https://www.openmicroscopy.org/qa2/qa/upload/ for us to test.

sbesson:
[...]

PR #3349, opened 2019-04-15
OIR: Allow more than 1000 time-points

dgault:
This has been raised in thread https://forum.image.sc/t/problems-opening-olympus-oir-files-using-bio-formats/24747/2 (which contains a sample file for download) and https://github.com/openmicroscopy/bioformats/issues/3242 

The issue shows up when loading a timepoint from 1000 onwards which would return an empty frame and `[WARN] No pixel blocks for plane #1000`


cudmore:
@dgault, thanks for doing this. What is the timeframe on these changes making it into a download? Is there a way I can download a branch with this change in it and drop the plugin into Fiji?

PR #3350, opened 2019-04-16
Cellomics: indexing fixes for plates with a single well or missing fields

melissalinkert:
First four commits backported from a private PR; last commit added to fix test failures.  Also fixes https://trello.com/c/Liyb44EG/279-cellomics-dib-plate-failure

For single well plates, this should fix the well row and column reported by ```showinf -nopix -omexml``` to match what is in the file name instead of always being 0.  For plates with missing fields, in particular cases when the first field index is greater than 0, this should fix indexing so that ```showinf -nopix -omexml``` does not throw an exception or result in orphaned ```Image```s.

To test, use ```inbox/cellomics/samples/large-field-index``` and ```curated/cellomics/qa-21720```.  The former mimics the test dataset for the original private PR.  The first field index is 29; ```showinf -nopix -omexml``` on one of the ```*d*.C01``` or ```*o*.C01``` files should show 2 channels x 290 series (10 fields x 29 wells), which matches the file count.  ```showinf -nopix -omexml``` on the QA 21720 dataset should result in a single series that is linked to the ```Plate``` as well row 0 and column 13 (```A14```, as in the file name).

I would expect this to impact memo files, so not safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
@@ -80,6 +80,8 @@ public class CellomicsReader extends FormatReader {
   private Pattern cellomicsPattern;
   private String[] files;
 
+  private ArrayList<String> metadataFiles = new ArrayList<String>();
+
   // -- Constructor --
 
   /** Constructs a new Cellomics reader. */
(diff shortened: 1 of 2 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-DEV-merge-push#1731](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/1731/). See the [console output](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-merge-push/1731/consoleText) for more details.
Possible conflicts:
  - PR #3347 melissalinkert 'Add FormatTools methods for creating well names'
    - components/formats-gpl/src/loci/formats/in/CellomicsReader.java

--conflicts

dgault:
Builds and tests have remained green with this and the config PR included. Tested using the files mentioned.

Without the PR:
`inbox/cellomics/samples/large-field-index` results in 290 series with 2 channels. There are 29 wells, 2 columns of 12 and 1 of 5, each with 10 fields
`curated/cellomics/qa-21720` results in a single 3 channel image with a plate but no well info

With the PR:
`inbox/cellomics/samples/large-field-index` appears to be the same as before but with the columns and rows swapped
`curated/cellomics/qa-21720` results in a single 3 channel image with a plate and a single well, column 13, row 0. This is correct and matches the naming convention

The other files affected in the config PR update rows and columns which were previously 0,0. These too look to be correct

PR #3351, opened 2019-04-16
CellWorX: ignore files that contain "_thumb"

melissalinkert:
Backported from a private PR.

Thumbnail files with a UUID in the file name may be of the form
```<plate name>_<dimensions>_thumb<UUID>.tif```, not
```<plate name>_<dimensions>_thumb_<UUID>tif```.  Checking the index of
```_thumb``` in the file name catches both cases.

To test use ```curated/cellworx/samples/thumb-test``` (copied from ```curated/cellworx/public/idr0005```, with TIFF files renamed).  With this PR, ```showinf -nopix 2011-04-19-plate-1.HTD``` should throw an exception:

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
	at loci.formats.in.CellWorxReader.getTiffFiles(CellWorxReader.java:788)
	at loci.formats.in.CellWorxReader.initFile(CellWorxReader.java:346)
	at loci.formats.FormatReader.setId(FormatReader.java:1389)
	at loci.formats.ImageReader.setId(ImageReader.java:842)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1035)
```

With this PR, the same command should have identical results compared to ```curated/cellworx/public/idr0005```.

I wouldn't expect this to impact tests or memo files, so should be fine for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
@@ -784,7 +784,7 @@ public class CellWorxReader extends FormatReader {
       for (String f : directoryList) {
         if (checkSuffix(f, new String [] {"tif", "tiff", "pnl"})) {
           String path = new Location(parent, f).getAbsolutePath();
-          if (path.startsWith(base) && path.indexOf("_thumb_") < 0) {
+          if (path.startsWith(base) && path.indexOf("_thumb") < 0) {
             files[nextFile++] = path;
           }
         }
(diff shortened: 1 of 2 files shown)


sbesson:
Briefly discussed with @melissalinkert. I  added `curated/cellworx/public/idr0005` to the curated QA repository for configuration but postponed the generation of configuration files because thumbnails are not detected and need to be manually excluded.

While looking at this change, one question is whether thumbnail files should not be detected as part of the fileset instead but not associated to any series i.e. detecting as auxiliary files. The main impact of this might be OMERO import workflows as the transfer time of the complete filesets could be multiplied by two.

melissalinkert:
With a840f17, the thumbnail files should now be included in the used files list (as non-pixels files, since nothing is read from them).  This should mean that thumbnail files do not need to be excluded when generating configuration files.

PR #3356, opened 2019-04-22
Flex: use plate barcode for grouping, so that truncated files are kept

melissalinkert:
Without this change, a truncated file would be rejected from the used file list,
which could completely break plate detection.  With an added barcode
check, plates can include truncated files, and files from different
plates will not be inadvertently grouped together.

To test, use ``` /uod/idr/filesets/idr0056-stojic-lncrnas/20190123-ftp/Stojic_screen B_platesIDR/1982/Meas_01(2014-07-04_14-48-06)/```.  Without this change, ```showinf -nopix``` on any of the files should result in a single series and a single used file.  With this change, all .flex files in the directory should be grouped into a plate.

Tests should continue to pass, and memo files should be unaffected.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FlexReader.java
+++ b/components/formats-gpl/src/loci/formats/in/FlexReader.java
@@ -251,8 +251,14 @@ public class FlexReader extends FormatReader {
       new RandomAccessInputStream(getFileHandle(file.file))) {
         IFD ifd;
         if (file.offsets == null) {
-          ifd = file.ifds.get(imageNumber);
-          factor = 1d;
+          if (imageNumber < file.ifds.size()) {
+            ifd = file.ifds.get(imageNumber);
+            factor = 1d;
+          }
+          else {
+            Arrays.fill(buf, (byte) 0);
+            return buf;
+          }
         }
         else {
           // Only the first IFD was read. Hack the IFD to adjust the offset.
@@ -1211,6 +1217,7 @@ public class FlexReader extends FormatReader {
     HashMap<String, ArrayList<String>> v = new HashMap<String, ArrayList<String>>();
     Boolean firstCompressed = null;
     int firstIFDCount = 0;
+    String firstBarcode = null;
     for (String file : fileList) {
       LOGGER.warn("parsing {}", file);
       IFD firstIFD = null;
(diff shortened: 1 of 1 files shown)


dgault:
With this PR included tests have been green and there have been no regressions. This also does not impact on memo files.

Tested using `/idr0056-stojic-lncrnas/20190123-ftp/Stojic_screen B_platesIDR/1982/Meas_01(2014-07-04_14-48-06)/`. 

Without the PR opening any of the flex files results in a single series with 4 channels. 
With the PR included all of the files are included and results in 11520 series in total and each of these with 4 channels. The images display correctly.

sbesson:
The barcode check introduced in this PR caused some detection issues with new variant of the Opera Flex format received in the context of an IDR submission. As an example,  a single plate folder had half of the `flex` files associated with barcode `TS3ab-RT-SnfGVphT-2011-12-30` and the other half associated with `TS3cd-RT-SnfGVphT-2011-12-30`, both barcode with different start times.

Quoting some feedback from the submitters:

```
Most of the plates were imaged half at a time: odd rows separately (ab) and even rows separately (cd). Plates that were re-imaged or done on separate days have _# (mostly _2, a few have _3) in the folder name.
...
The plate name is manually set by the operator; this means you will notice some inconsistencies between the plates (ab, AB, _, - ,...). 
```

Looks like we have two options to be able to support this scenario:

- either treat each subset imaged at a given time with a unique barcode as a separate plate acquisition
- or maintain the expectation to treat everything within `Meas_xx` folder as a single plate acquisition and only group plate acquisitions at  the `Meas_xx` level (see also #3537)

In the second case, relaxing/suppressing the barcode check could be minimally tested against our representative Flex samples. Before opening such a PR, @melissalinkert are you aware of other situations where relaxing this check would break the behavior?

melissalinkert:
I wouldn't be surprised if completely removing the barcode check caused some of our existing tests to fail, as a few of the ```flex``` subdirectories have files from very different plates/acquisitions in the same directory.  It might be OK to limit relaxing the check to cases where there is a parent ```Meas*``` directory, or use the plate name instead of the barcode to group files.  Adding a reader-specific option to turn off the barcode check would also be safe, but presumably doesn't help much with the IDR case.

It might be a little more work, but would it be conceptually easier in this case to model the data as one ```Plate``` with two ```PlateAcquisition```s?

PR #3362, opened 2019-04-30
TIFF: don't assume all IFDs have same XY size during non-sequential write

melissalinkert:
If ```setWritingSequentially(true)``` was not called on the ```TiffSaver```, then writing each
resolution of a pyramid in order caused the previous resolution's
dimensions to be used.  This updates the IFD reuse logic so that only
tags which are not present in the ```ifd``` argument are reused.  Use cases
where ```setWritingSequentially(true)``` is called will not be affected.

This should fix https://github.com/ome/bio-formats-examples/issues/48


dgault:
Tested using https://github.com/ome/bio-formats-examples/issues/48, this now generates the correct results with each resolution being half the size of the previous.

Testing using other scenarios with `setWritingSequentially(true)` shows files are unaffected by this change and are correct.

The logic here seems fairly sensible and avoids any incorrect metadata values being duplicated.

PR #3364, opened 2019-05-02
Make ImageConverter public

### LINKED ISSUE #3359: Why is ImageConverter private?
Is there any reason why ImageConverter is "private ImageConverter()" instead of "public ImageConverter()"?

https://github.com/openmicroscopy/bioformats/blob/4d0f6757bed4e9b219ee9bfac575f1e924d9a2b9/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java#L118

sbesson:
Fixes #3359 

There is no obvious rationale for keeping this constructor private and there are examples of downstream code working around the private scope - see https://github.com/spacetx/spacetx-writer/blob/0.1.1/src/main/java/spacetx/FOVTool.java#L430.

In addition to publicizing the constructor, this PR adds a minimal unit test showing the creation of the `ImageConverter` class

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -128,7 +128,7 @@ public final class ImageConverter {
 
   // -- Constructor --
 
-  private ImageConverter() { }
+  public ImageConverter() { }
 
   /**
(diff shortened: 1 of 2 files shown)


dgault:
No objections to making this public. The new unit test runs successfully and looks good.

PR #3373, opened 2019-05-16
LIF: Correctly read scale from polygon regions of interest

seanwarren:
Bioformats incorrectly loads elliptical/polygon ROIs from LIF files created in newer versions of LAS X because bioformats does not account for the scale factor (which was unused in previous versions of LAS X). 

This PR uses the ROI scale correctly and so can load ROIs created in both older and newer versions.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -2361,7 +2361,7 @@ public class LIFReader extends FormatReader {
           for (int i=0; i<x.size(); i++) {
             points.append(x.get(i).doubleValue() * scaleX + roiX);
             points.append(",");
-            points.append(y.get(i).doubleValue() * scaleX + roiY);
+            points.append(y.get(i).doubleValue() * scaleY + roiY);
             if (i < x.size() - 1) points.append(" ");
           }
           store.setPolygonID(shapeID, roi, 1);
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -2352,16 +2352,16 @@ public class LIFReader extends FormatReader {
         roiY = transY - 2 * cornerY;
       }
 
-      // TODO : rotation/scaling not populated
+      // TODO : rotation not populated
 
       String shapeID = MetadataTools.createLSID("Shape", roi, 1);
       switch (type) {
         case POLYGON:
           final StringBuilder points = new StringBuilder();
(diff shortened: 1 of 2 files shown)


components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -2352,16 +2352,16 @@ public void storeROI(MetadataStore store, int series, int roi, int roiIndex)
...
       String shapeID = MetadataTools.createLSID("Shape", roi, 1);
       switch (type) {
         case POLYGON:
           final StringBuilder points = new StringBuilder();
           for (int i=0; i<x.size(); i++) {
-            points.append(x.get(i).doubleValue() + roiX);
+            points.append(x.get(i).doubleValue() * scaleX + roiX);
             points.append(",");
-            points.append(y.get(i).doubleValue() + roiY);
+            points.append(y.get(i).doubleValue() * scaleX + roiY);
```
mtbc: Not `scaleY`? (Just checking.)

sbesson:
Thanks @seanwarren. Do you also have a sample file created with the new version of LAS X allowing us to test this change?

components/formats-gpl/src/loci/formats/in/LIFReader.java
seanwarren: Good catch @mtbc! Sorry this was a bit of a rushed fix

seanwarren:
@sbesson - I've uploaded a LIF file with four FRAP series with ROI sizes 1, 2.5, 5 and 10um respectively here:

https://www.dropbox.com/s/usr02vbd0umdz6g/150519_FRAP_test_ROIs_chromagreen.zip?dl=0

There are screenshots from LAS X for each series.

seanwarren:
Hi @sbesson - just wanted to check that test data was sufficient?

sbesson:
@seanwarren yes, thank you for uploading representative data. With [Bio-Formats 6.1.0](http://www.openmicroscopy.org/2019/05/20/bio-formats-6-1-0.html) being us we will now come back to the open backlog and review candidates for quick patch releases of the library.

I think your sample files should be sufficient but we will let you know here if more is required. Can the data uploaded above be made available [publicly](https://downloads.openmicroscopy.org/images/) under a CC-BY license or would you rather have it kept private?

seanwarren:
Apologies, I thought I had replied to this. Yes, the data can be made freely available.

sbesson:
Merging for inclusion in the next patch release of Bio-Formats. We will also make the samples public as per https://github.com/openmicroscopy/bioformats/pull/3373#issuecomment-496503408. Thanks for your contribution Sean!

PR #3376, opened 2019-05-17
Bump ome-common to 6.0.3

dgault:
Bumping ome-common version to https://github.com/ome/ome-common-java/releases/tag/v6.0.3


sbesson:
@dgault Should I close #3375 ?

dgault:
Sorry missed that somehow

PR #3378, opened 2019-05-21
DeltavisionReader: add lens IDs associated with rcpnl files

melissalinkert:
Backported from a private PR.

We received a list of lens definitions for IDs 18107 - 18111, which are primarily associated with the ```rcpnl``` variant of Deltavision files.  Most of these are new IDs, but 18107 was already defined with a different NA.  The original NA (0.15) will still be used if the file extension is not ```rcpnl```, otherwise the new NA (0.30) will be used.

As far as I can see, the only file that we have with a lens ID between 18107 and 18111 is ```curated/deltavision/qa-21574/RC_SevenChannelScan.rcpnl```, which uses 18107.  I would expect ```showinf -nopix -omexml RC_SevenChannelScan.rcpnl``` without this PR to have ```LensNA``` on the ```Objective``` set to 0.15, and with this PR to have it changed to 0.30 (as above).  Renaming the file to ```RC_SevenChannelScan.dv``` should show a ```LensNA``` of 0.15 even with this PR.

This should not affect tests or memo files, and should be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -2659,6 +2659,11 @@ public class DeltavisionReader extends FormatReader {
         break;
       case 18107: // API 10X, BH
         lensNA = 0.15;
+
+        if (checkSuffix(currentId, "rcpnl")) {
+          lensNA = 0.30;
+        }
+
         magnification = 10.0;
         workingDistance = 15.00;
         immersion = MetadataTools.getImmersion("Air");
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -2662,9 +2662,27 @@ public class DeltavisionReader extends FormatReader {
         magnification = 10.0;
         workingDistance = 15.00;
         immersion = MetadataTools.getImmersion("Air");
+        correction = MetadataTools.getCorrection("PlanFluor");
         break;
       case 18108:
         magnification = 20.0;
+        lensNA = 0.75;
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
```diff
@@ -2659,9 +2659,40 @@ else if (lensID < 20000) {
         break;
       case 18107: // API 10X, BH
         lensNA = 0.15;
+
+        if (checkSuffix(currentId, "rcpnl")) {
```
sbesson: Do all the objective IDs defined below apply to files with both extensions or should they also only be mapped in the case of `rcpnl` files?

melissalinkert: I'm honestly not sure.  We don't have any examples of non-```rcpnl``` files with any of these IDs, but that's not necessarily conclusive.  It would be easy enough to make 18108-18111 only apply to ```rcpnl``` files, if that's preferable.

dgault:
Tested using `curated/deltavision/qa-21574/RC_SevenChannelScan.rcpnl`:
Without this PR the file has `Lens ID Number = 18107` and `LensNA = 0.15`
With this PR the file has `Lens ID Number = 18107` and `LensNA = 0.3`
With the PR and the file renamed to `.dv` has `Lens ID Number = 18107` and `LensNA = 0.15`

In that sense this PR behaves as expected and does not break any existing tests.

Having the ID's differ between the 2 extensions is obviously not ideal and doesn't provide a great deal of confidence for the other new ID's. That being said we can only work with the known sample images we have right now, and until we have a `.dv` sample that contradicts then this seems like an appropriate approach. If we need to update the logic for the others at a later date then we can do so in a subsequent patch release.

PR #3380, opened 2019-05-23
Columbus: warn and return a blank plane if a file has too few planes

melissalinkert:
See https://github.com/IDR/idr0056-stojic-lncrnas/issues/3

This should prevent an exception from being thrown if a TIFF file is truncated, and instead return a blank plane and log the details at ```WARN```.  This is consistent with what we do for a few other HCS formats.

I wouldn't expect this to have any impact on tests or memo files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ColumbusReader.java
@@ -193,12 +193,16 @@ public class ColumbusReader extends FormatReader {
       }
     }
 
+    Arrays.fill(buf, (byte) 0);
     if (p != null && new Location(p.file).exists()) {
       reader.setId(p.file);
-      reader.openBytes(p.fileIndex, buf, x, y, w, h);
-    }
-    else {
-      Arrays.fill(buf, (byte) 0);
+      if (p.fileIndex < reader.getImageCount()) {
+        reader.openBytes(p.fileIndex, buf, x, y, w, h);
+      }
+      else {
+        LOGGER.warn("Blank plane for series {} plane {}; {} may be truncated",
+          getSeries(), no, p.file);
+      }
     }
     return buf;
   }


dgault:
Testing was carried out on` idr0056-stojic-lncrnas/20190513-ftp/RNAi screenA/1921[652]/ ` using bftools, FIJI and the short program Melissa used to reproduce the issue originally:

```

public class Test {
    try (ImageReader reader = new ImageReader()) {
      reader.setId(args[0]);
      for (int s=0; s<reader.getSeriesCount(); s++) {
        reader.setSeries(s);
        for (int p=0; p<reader.getImageCount(); p++) {
          /* debug */ System.out.println("opening plane " + p + " from series " + s);
          reader.openBytes(p);
        }
      }
    }
  }
}
```

With Bio-Formats 5.9.2 running the short program reproduces the issue:
```
Exception in thread "main" loci.formats.FormatException: Invalid image number: 4 (series=0, imageCount=4)
	at loci.formats.FormatTools.checkPlaneNumber(FormatTools.java:1007)
	at loci.formats.FormatTools.checkPlaneParameters(FormatTools.java:996)
	at loci.formats.in.MinimalTiffReader.openBytes(MinimalTiffReader.java:283)
```
This can also be seen using `showinf -series 1`.

With Bio-Formats 5.9.2 and this PR included the exception is not present and all planes open for every series. All 3 planes for series 1 however are blank.

With Bio-Formats 6.0.0 and 6.1.0 the exception is not present and the series opens with 4 channels as expected and the images display correctly.

This PR does resolve the exception and allow the remainder of the fileset to be read. It seems as suspected that a more comprehensive fix was included with 6.0.0. I have tried to backport a number of the TIFF PR's from 6.0.0 onto 5.9.2 to isolate which one resolved the issue but have not yet been able to identify the exact fix.

With the latest Bio-Formats and this PR included there is no regression in the tests and no exception is thrown with the planes all opening and displaying correctly (not blank).

Merging this PR, even if its fix has been superseded it provides a good fallback if ever required.

PR #3383, opened 2019-05-30
InCell 1000/2000: don't trust the plane count in the "Images" element

melissalinkert:
Instead, count the number of child ```Image``` elements.  See https://github.com/IDR/idr0061-wolf-spindlepositioning/issues/2.

For almost all datasets, the ```number``` attribute on ```Images``` matches the number of ```Image```s, so I wouldn't expect this to cause any test failures.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -84,6 +84,7 @@ public class InCellReader extends FormatReader {
   private List<Double> emWaves, exWaves;
   private List<String> channelNames;
   private int totalImages;
+  private transient int imagesNumber;
   private int imageWidth, imageHeight;
   private String creationDate;
   private String rowName = "A", colName = "1";
@@ -252,6 +253,7 @@ public class InCellReader extends FormatReader {
       imageFiles = null;
       tiffReader = null;
       totalImages = 0;
+      imagesNumber = 0;
       emWaves = exWaves = null;
       channelNames = null;
       wellCoordinates = null;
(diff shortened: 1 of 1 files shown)


sbesson:
@melissalinkert thanks. This feels like a typical edge case where we have to make a call between teaching Bio-Formats to internally handle datasets inconsistencies or simply failing on such datasets. I do not know of another case where the InCell acquisition software should produce an element count that does not match the number of elements.

I will take a closer look at the submitted data and confirm the image count is the only issue. If this is the case, my inclination goes towards for the second behavior. Independently of the fail-or-not behavior, the logic introduced is still useful minimally compare the expectation in terms of `Image` count with the true number of elements and give an informative error message on mismatch.

sbesson:
Following a discussion with @melissalinkert, proposing https://github.com/sbesson/bioformats/commit/13e6cee2e3299edef96e961729a92e86c6ca7d72 to compare the expected values in terms of image counts and throw a warning message in case of mismatch.

sbesson (approved):
Looking at the history of this PR - see https://github.com/IDR/idr0061-wolf-spindlepositioning/issues/2 for the full context, there were two source issues:

- the first issue was related to a mismatch between the content of the xcde metadata files the acquisition modality as described by the experimentalist
- to correct for this, amended versions of the xcde metadata files were resubmitted. This is the file where the metadata for the images count was not matching the effective number of images.

This PR proposes a defensive approach to handle this scenario in a lenient way internally calculating the effective number of images elements, comparing it to the declared `number` value and throwing a warning in case of a mismatch (but not failing the initialization).

A representative dataset has been derived from another synthetic sample of the file format, configured and is now being executed by the nightly CI builds - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/207764/console.

My feeling is that this PR deals a very edge-case which only happened as a result of a manual intervention on the metadata files. The internal computation and sanity check between the `Images.number` value and the actual number of `Images` elements remains valuable as we are iterating through these elements anyways. Happy to see this merged and released either in the upcoming release of a patch release as nothing critical depends on it.

dgault (approved):

PR #3391, opened 2019-06-12
bfconvert: add -cache, -cache-dir, and -no-sas options

melissalinkert:
Backported from a private PR.

This adds ```-cache```, ```-cache-dir```, and ```-no-sas``` to ```bfconvert```, just like in ```showinf```.  ```-cache``` and ```-cache-dir``` will generate and use a memo file for the input file, which is useful when running something like ```bfconvert -series n``` multiple times in parallel.  ```-no-sas``` turns off ```OriginalMetadata``` annotation population, which can reduce memory consumption and is particularly helpful for extracting just pixel data from plates.


dgault:
Tested each of the new options and each appears to be working as expected with the cache options generating memo files in the correct locations. Merging for inclusion in 6.2.0

PR #3392, opened 2019-06-12
Zeiss CZI: expand case where tiles are automatically stitched

melissalinkert:
Backported from a private PR.

This fixes automatic stitching for 4 existing files in the repository.  Ground truth screenshots are in https://docs.google.com/document/d/139zYc9i5TPQbS_ai3cKnAWNbdix-4G_zUOwpmZcn2Oo/edit.  With the corresponding configuration PR, tests should continue to pass.


dgault:
Tested using the 4 images from the config PR.

Without this PR all 4 open as a series of tiles. With the PR included each opens as a single series with the stitched image. The image looks sensible and matches the ground truth from the google doc. The metadata also appears to be correct in each instance.

No other files in the repo appear to be impacted and the builds and tests have been green with this PR and the config PR included.

PR #3393, opened 2019-06-13
Operetta: do not attempt to open images that are too small

melissalinkert:
Backported from a private PR.

Sometimes a TIFF file will have smaller XY dimensions than all other TIFF files, and will not match the dimensions recorded in ```Index.idx.xml```.  Instead of throwing an exception, this commit returns a blank plane and logs the file name at ```WARN```.

So far we have only seen this in one specific plate, with no clear indication of how/why one file has smaller dimensions.  It wouldn't surprise me even a little if IDR encounters this at some point, though.

To test, use ```curated/perkinelmer-operetta/samples/smaller-xy```, which is copied from QA 17826 as noted in the readme.  ```showinf -series 7 Index.idx.xml``` without this PR should throw a ```FormatException```.  With this PR, the same command should display 2 channels without an exception; the first channel contains data and the second is blank.  The name of the smaller file should be logged.

Happy to have this do something other than return a blank plane, either now or in the future.  Pasting the smaller image into the upper-left corner would be possible, as would scaling it to match the expected XY dimensions.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -228,7 +228,12 @@ public class OperettaReader extends FormatReader {
           reader = new MinimalTiffReader();
         }
         reader.setId(p.filename);
-        reader.openBytes(0, buf, x, y, w, h);
+        if (reader.getSizeX() >= getSizeX() && reader.getSizeY() >= getSizeY()) {
+          reader.openBytes(0, buf, x, y, w, h);
+        }
+        else {
+          LOGGER.warn("Image dimension mismatch in {}", p.filename);
+        }
         reader.close();
       }
     }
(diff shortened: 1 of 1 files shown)


joshmoore:
I'd think something like:
 * keep the warn
 * pad down and right
   - with zeroes
   - or if possible, NaNs
   - or if possible, with red blinking lights...

melissalinkert:
2b8d503 switches to reading the small image into the upper left corner, so there will be a border of zeroes on the bottom and right side.  ```showinf -series 7 Index.idx.xml``` with 2b8d503 should show that the second channel has data in the upper left quarter, with everything else zero.  Cropping works as usual, so reading a tile outside of the upper left quarter should result in all zeroes.

dgault:
Tested using `curated/perkinelmer-operetta/samples/smaller-xy`

Without this PR all of the series opened and displayed as expected with the exception of series 7 which threw a `FormatException` as expected.

With the PR included all of the series opened and displayed including series 7. Series 7 now displays as having the same dimensions, with the first channel containing a full plane image and the second channel having a sub image in the top left corner (approximately 680x500 pixels) with the rest of the frame being 0.

PR #3400, opened 2019-07-01
TIFF: fixes for tag overwriting

melissalinkert:
Backported from a private PR.

1062297 updates ```TiffSaver.overwriteIFDValue``` to overwrite the original value with 0s before the new value is written.  This means that the original tag value is completely gone, and no longer orphaned but still visible in a hex editor.  This is helpful for anyone using ```overwriteIFDValue``` (or higher-level things like ```tiffcomment```) to de-identify or anonymize TIFFs.

In working on this, it became clear that there was a bug in ```overwriteIFDValue``` when the new value has the same length as the original value (but neither value is inlined).  The bug is exhibited by the unit test in fe4d563, which fails without 1062297 and passes with it.

I would expect all other tests to continue to pass; as this is specific to writing, there should be no memo changes.  It probably makes sense to spot check a few common workflows of ```tiffcomment``` and ```bfconvert``` to be on the safe side, though, particularly for things like (pyramid) OME-TIFF.  I'm inclined to think this is safe for a patch release, but happy to be told otherwise.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -807,6 +807,18 @@ public class TiffSaver {
         LOGGER.debug("\tnew: (tag={}; type={}; count={}; offset={})",
           new Object[] {newTag, newType, newCount, newOffset});
 
+        // first overwrite the original value with 0s
+        // this makes sure that the original value is really gone
+        // and not just orphaned
+
+        int bytesPerElement = entry.getType().getBytesPerElement();
+        if (entry.getValueCount() > (offset / bytesPerElement)) {
+          out.seek(entry.getValueOffset());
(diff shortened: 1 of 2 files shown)


dgault:
Builds and tests have been green with this included.

Without the PR I was able to reproduce the issue using `tiffcomment` to extract the xml from an OME-TIFF. I then modified it to remove a number of fields. I the used `tiffcomment` to replace the XML. When expecting the file afterwards I could find the orphaned XML which had not been overwritten.

With only the unit test included the failure can be reproduced:
```
FAILED: testOverwriteCommentEqualLength
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
```
With the full PR included there are no exceptions and the test completes without issue.

Ive also run the `bfconvert` workflow with a selection of different formats including OME-TIFF pyramid. These completed as expected without any noticeable issues or exceptions.

Additionally Ive tested the `tiffcomment` workflow to replace XML and this too completed without issue and the original XML was not present.

PR #3402, opened 2019-07-13
BDV: Verify timepoints

dgault:
This is a follow up to https://github.com/openmicroscopy/data_repo_config/issues/409 and https://github.com/openmicroscopy/data_repo_config/issues/410

Both of these failures relate to the `astroP65` dataset.
This dataset has an XML which lists the timepoints as:
```
    <Timepoints type="range">
      <first>0</first>
      <last>0</last>
    </Timepoints>
```

However the actual h5 file no timepoint 0 and only contains timepoint 18.

This PR changes the logic to:
- parse timepoints from XML
- verify that each timepoint exists in the h5 file
- if timepoints are invalid log a warning
- if invalid then search the h5 file for valid timepoint list
- from this list recreate the timepoint pattern
- use the new valid timepoint settings

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/BDVReader.java
+++ b/components/formats-gpl/src/loci/formats/in/BDVReader.java
@@ -467,17 +467,7 @@ public class BDVReader extends FormatReader {
     core.clear();
     // read experiment structure and collect coordinates
 
-    int numTimepoints = 1;
-    if (timepointUsePattern && lastTimepoint > 0) {
-      numTimepoints = lastTimepoint - firstTimepoint + 1;
-      if (timepointIncrement > 0) {
-        numTimepoints /= timepointIncrement;
-      }
-    }
-    else {
-      if (lastTimepoint == 0) lastTimepoint = firstTimepoint;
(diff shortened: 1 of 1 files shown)


sbesson:
Retested using various workflows:

- reading `astroP65.h5` into an OMERO.server results in no regression with this PR included
- the test-suite is still broken with this PR included. The configuration of all BDV datasets fail while configuring `idr0048` alone succeeded but the execution still results in failing `testSaneUsedFiles` and `testValidXML`

sbesson:
As this was used for testing the `idr0048` submission in a CI build resulted in an import error similar to the one reported in https://github.com/ome/bioformats/pull/3402#issuecomment-515453349. Labelling this as `exclude` to confirm BDV datasets can be successfully imported without it /cc @pwalczysko

sbesson:
As per https://github.com/ome/bioformats/pull/3575, all BDV samples have now been configured and are tested daily by the non-regression builds. I will leave @dgault to confirm there is nothing critical left in this PR and I assume it can be closed.

PR #3405, opened 2019-07-16
Use javaObject to construct Boolean object in GNU Octave

sbesson:
See https://forum.image.sc/t/bio-formats-octave-error-on-bfsave/27504/4. This
should fix the execution of bfsave in the GNU Octave environment

The impact of this PR should be tested both in GNU Octave (4.2.0 e.g. available from Ubuntu 18.04) and MATLAB R2017b or later. In both environments, a call to the writer API e.g. using `bfsave(zeros(512,512,10,3,1), 'out.ome.tiff')` should successfully create an output file. Without this PR the command should fail on GNU Octave

### DIFF
--- a/components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
+++ b/components/formats-gpl/matlab/createMinimalOMEXMLMetadata.m
@@ -53,12 +53,7 @@ metadata = OMEXMLService.createOMEXMLMetadata();
 metadata.createRoot();
 metadata.setImageID('Image:0', 0);
 metadata.setPixelsID('Pixels:0', 0);
-if is_octave()
-    java_true = javaObject('java.lang.Boolean', 'TRUE');
-else
-    java_true = java.lang.Boolean.TRUE;
-end
-metadata.setPixelsBigEndian(java_true, 0);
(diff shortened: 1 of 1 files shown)


carandraug:
The proposed fix works in Octave 4.0, 4.2, 5.1, and in the current development sources so I believe it is fine to push (but this is a workaround to a bug in Octave). Two comments:

* the changed line on this commit also works in Matlab, so maybe you could remove the `is_octave` branch altogether (and leave a comment saying that we're constructing a Boolean object instead of boolean primitive to avoid Octave auto unboxing and a link to this issue?)

* on `bfsave`, the default `metadata` value (empty array `[]`) does not pass its own validation (`@(x) isa(x, 'loci.formats.ome.OMEXMLMetadata')`. Matlab does not check the validity of the default value and Octave copied that behaviour in 4.2. What that means is that for `bfsave` to work in Octave 4.0 that would also need fixing.

dgault:
Tested this with Octave 4.2 and it is working as expected and there is no regression with MATLAB . 

I have opened a separate Trello card to capture the additional comment on the default metadata value for Octave 4.0: https://trello.com/c/Lnorq1jh/377-octave-40-default-metadata-in-bfsave

PR #3410, opened 2019-07-22
TiffSaver: add overwriteIFDValue signature that takes an IFD offset

melissalinkert:
This is helpful for modifying thumbnail and sub-IFDs, which aren't
included in the list of offsets returned by ```TiffParser.getIFDOffsets()```.

I wouldn't expect this to have any impact on memo files or existing tests, but it arguably still requires a minor release since it's an API addition.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -724,6 +724,11 @@ public class TiffSaver {
    * new data to the end of the file and updates the offset field; if not, or
    * if the old data is already at the end of the file, it overwrites the old
    * data in place.
+   *
+   * @param raf the input stream representing the file to be edited
+   * @param ifd the index into the list of IFDs {@see TiffParser#getIFDOffsets()}
+   * @param tag the tag code
+   * @param value the new value for the tag
    */
   public void overwriteIFDValue(RandomAccessInputStream raf,
     int ifd, int tag, Object value) throws FormatException, IOException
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
```diff
@@ -759,14 +764,52 @@ public void overwriteIFDValue(RandomAccessInputStream raf,
...
+   *
+   * @param raf the input stream representing the file to be edited
+   * @param ifdOffset the offset to the IFD
+   * @param tag the tag code
+   * @param value the new value for the tag
+   */
+  public void overwriteIFDValue(RandomAccessInputStream raf,
+    long ifdOffset, int tag, Object value) throws FormatException, IOException
+  {
+    raf.seek(0);
```
sbesson: With this refactoring, does that mean any call to the current `overwriteIFDValue(RandomAccessInputStream, int, int, Object)` API will now perform a redundant check on the first IFD?

melissalinkert: The header will be checked twice, yes.  Could refactor further to add a ```overwriteIFDValue(RandomAccessInputStream, long, int, Object, boolean skipHeaderCheck)```?  Otherwise not sure of a good and safe way to skip that check.

sbesson: Since we're introducing new API anyways, I think introducing 

` public void overwriteIFDValue(RandomAccessInputStream raf,
    long ifdOffset, int tag, Object value, boolean skipskipHeaderCheck)`

is acceptable. From there both `public void overwriteIFDValue(RandomAccessInputStream raf,
    long ifdOffset, int tag, Object value)` and `public void overwriteIFDValue(RandomAccessInputStream raf,
    int ifd, int tag, Object value)` can delegate to this API by passing the appropriate boolean value.

melissalinkert: Added in b43e748

dgault:
With the latest update to add `skipHeaderCheck` the API updates here look to be good. 

The existing unit tests, repo tests and workflows are unaffected by the changes. I reran some local tests using the `PyramidOMETiffWriter` and the generated images and metadata all looked correct and no regressions found from these additions.

PR #3411, opened 2019-07-24
Deltavision: Add new objective info for new lensID

dgault:
The new objective metadata was provided in thread https://forum.image.sc/t/warning-when-opening-dv-files-in-fiji/27923/2 for a new lensID 20007

This is suitable for a patch release. Testing is simply to ensure that builds and tests are green with no regressions.


sbesson:
The changes look straightforward and do not cause any regression against the curated Deltavision samples. It would be useful to validate it against a sample file from the original issue. If that does not become available in time for Bio-Formats 6.2.1, it might be doable to construct a mock example using an existing sample dataset by modifying the lens ID.

dgault:
We have received a sample file with the new lens Id, I will create a new repo config with QA-27760

dgault:
Config for the uploaded QA now open: https://github.com/openmicroscopy/data_repo_config/pull/419

PR #3413, opened 2019-07-30
Leica LIF: fix units and indexing for tile-based plane positions

melissalinkert:
Backported from a private PR.

This updates the units for tile positions to meters, and fixes how tile positions are indexed when there is a mixed tile/non-tile dataset.  Specifically, if a non-tile series is present before a set of tiles, then the first tile's position is applied to the non-tile series (and all remaining positions are off by one).  I don't think we have an example of this that can be added to the repo, but I'll try to make one shortly.

I would expect 3 existing files to be affected, as described in the forthcoming config PR.  The only change is the units of ```PositionX``` and ```PositionY```, which change from ```reference frame``` to ```m```.  This seems to yield positions that are consistent with the physical pixel sizes, and should help anyone trying to stitch tiles.

With the config PR, all tests should continue to pass.  I wouldn't expect memo files to be affected, so this should be OK for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -1990,10 +1990,17 @@ public class LIFReader extends FormatReader {
           String posX = tileNode.getAttribute("PosX");
           String posY = tileNode.getAttribute("PosY");
 
+          while (fieldPosX.size() < image) {
+            fieldPosX.add(null);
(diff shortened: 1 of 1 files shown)


dgault:
The only files impacted by this are those in the updated config PR which have updated from reference frame to meters. Based on the other sizes and positions in the files changing to meters looks to be the most sensible unit for the values provided.

PR #3417, opened 2019-08-01
MinimalTiffReader: be defensive and use DataTools.safeMultiply64

sbesson:
See https://forum.image.sc/t/standard-workflow-and-format-to-handle-huge-slidescanner-images/28159/12

If the the TIFF is stored using very large tile/strips dimensions, `DataTools.safeMultiply32` might throw an IllegalArgumentException before the size comparison can happen.

To test, use the sample file available under `/uod/idr/repos/curated/tiff/tobias`. Without this PR, `showinf` or `bfconvert` should throw an `IllegalArgumentException` while retrieving the optimal tile size and importing into OMERO server should result in an exception during the pyramid generation. With the PR included, it should be possible to read and convert the TIFF file.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java
@@ -422,7 +422,7 @@ public class MinimalTiffReader extends SubResolutionFormatReader {
       // Some TIFF files only store a single tile, even if the image is
       // very, very large.  In those cases, we don't want to open the whole
       // tile if we can avoid it.
-      if (DataTools.safeMultiply32(height, getOptimalTileWidth()) >
+      if (DataTools.safeMultiply64(height, getOptimalTileWidth()) >
         10 * 1024 * 1024)
       {
         return super.getOptimalTileHeight();
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -1332,7 +1332,7 @@ public class OMETiffReader extends SubResolutionFormatReader {
           c.tileHeight = c.sizeY;
         }
         // Duplicate MinimalTiffReader getOptimalTileHeight logic
-        if (DataTools.safeMultiply32(c.tileHeight, c.tileWidth) >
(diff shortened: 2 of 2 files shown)


mtbc:
Could be worth deprecating those methods in favor of `Math.multiplyExact`. Doesn't block this PR though.

mtbc:
Good point, could be simply that we can use some Java stream folding or joining or somesuch to very simply implement `DataTools.safeMultiply` using `Math.multiplyExact`. Happy to create an issue though could be there's enough else in the backlog that it'd be more clutter than value-adding.

dgault:
This fixes one specific uses case for bfconvert. Width should also be updated. Also this only fixes 
minimaltiff, other formats will still be affected. There is no harm in putting this in but the converter should also be updated at some stage. Adding additional API to common, while I can see the benefit is probably overkill for now.

sbesson:
Interestingly, `MinimalTiffReader.getOptimalTileWidth` seems to return the width  independently of the plane size logic, only the optimal tile height is adjusted for very large sizes.  Looking at the rest of the code:

```
sbesson@ls30630:bioformats (tiff_illegalargumentexception) $ git grep safeMultiply32
components/formats-bsd/src/loci/formats/in/OMETiffReader.java:        if (DataTools.safeMultiply32(c.tileHeight, c.tileWidth) >
components/formats-gpl/src/loci/formats/in/SlidebookReader.java:          DataTools.safeMultiply32(getSizeX(), getSizeY());
components/test-suite/src/loci/tests/testng/Configuration.java:          planeSize = DataTools.safeMultiply32(reader.getSizeX(),
components/test-suite/src/loci/tests/testng/FormatReaderTest.java:          expected = DataTools.safeMultiply32(x, y, c, bytes);
components/test-suite/src/loci/tests/testng/FormatReaderTest.java:            planeSize = DataTools.safeMultiply32(resolutionReader.getSizeX(),
components/test-suite/src/loci/tests/testng/FormatReaderTest.java:          planeSize = DataTools.safeMultiply32(reader.getSizeX(),
```

I can certainly update `OMETiffReader` as part of this PR and review the `test-suite` usage althought the latter is less critical.

dgault:
Updating OMETiffReader makes sense as part of this PR, the rest do not look critical and can be left as is for now.

PR #3422, opened 2019-08-15
fake: allow injecting sleep on setId and openBytes

joshmoore:
Opening to debug an ongoing timeout issue. May expand into a (developer) feature.


dgault:
Merging this as is, i'll make a note to update the documentatn on Monday

PR #3425, opened 2019-08-25
Metamorph plate fixes

melissalinkert:
Backported from a private PR.

This fixes a few issues in Metamorph file name construction, especially when the wavelength name is not included in the file name.  It also updates some of the plate detection logic, so that potential well labels in the .nd file are considered, and labels similar to ```A1, 1 of 2```, ```A2, 1 of 2```, etc. will result in a plate.

I'll put together a test dataset on Monday, just opening this for now to make sure that tests aren't impacted.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -2000,7 +2000,7 @@ public class MetamorphReader extends BaseTiffReader {
    */
   private String locateFirstValidFile() {
     for (int q = 0; q < stks.length; q++) {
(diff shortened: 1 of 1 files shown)


melissalinkert:
Apologies for delay - test dataset is uploaded as described in https://github.com/openmicroscopy/data_repo_config/pull/425

dgault:
Tests have been green and no impacts from this PR have been seen in the Metamorph datasets. Tested with the dataset from `metamorph-test-plate`, without this PR results in an exception:
`java.lang.ArrayIndexOutOfBoundsException: 57`

With this PR included the dataset can be read and results in 96 series, `Stage 1 A1 - Stage 96 H12`

PR #3429, opened 2019-09-03
NativeND2 - Use floating point for 32 bit values

dgault:
Opening as a follow up to https://trello.com/c/4PvmeIIP/378-nd2-32-bit-float-being-treated-as-int to test impact on data repo

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -835,7 +835,7 @@ public class NativeND2Reader extends SubResolutionFormatReader {
               }
               else if (attributeName.equals("uiBpcInMemory")) {
                 core.get(0, 0).pixelType = FormatTools.pixelTypeFromBytes(
-                  valueOrLength / 8, false, false);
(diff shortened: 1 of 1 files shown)


sbesson:
The implementation change will cause the reader to read any Nikon ND2 file with `uiBpcInMemory` set to `32` as a 32-bit floating point rather than 32-bit unsigned integer.

As confirmed by the automated tests, this causes no apparent regressions for any of the numerous ND2 files we have been accumulating over the years in our curated QA repository which are either `uint16` or `uint8`.

Tested the QA file uploaded with the original image.sc thread without this PR,

Screen Shot 2019-09-10 at 14 40 57

With this PR included, the image is correctly detected as floating point and the pixel values are in agreement with NIS Elements Viewer

Screen Shot 2019-09-10 at 14 40 33
Screen Shot 2019-09-10 at 16 27 45

Will get a final sign-off from the @ome/formats team at the upcoming weekly meeting. The QA file should separately be configured and added to our curated repository for non-regression testing. Assuming there is no `uint32` ND2 sample to the best of our knowledge, this change looks like a straightforward improvement.

PR #3430, opened 2019-09-05
CZI: Use CreationDate as alternate AcquisitionDate 

dgault:
This is a follow up to the thread: https://forum.image.sc/t/omero-does-not-import-acquisition-date-from-czi-file-metadata/28910/3
A sample file is available in QA-27807

When the usual `Information > Image > AcquisitionDateAndTime` metadata is not set then also check `Information > Document > CreationDate` and use it to populate Image Acquisition Date.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -2086,6 +2086,9 @@ public class ZeissCZIReader extends FormatReader {
       }
 
       acquiredDate = getFirstNodeValue(image, "AcquisitionDateAndTime");
+      if (acquiredDate == null) {
+        acquiredDate = getFirstNodeValue(document, "CreationDate");
+      }
 
       Element objectiveSettings = getFirstNode(image, "ObjectiveSettings");
       correctionCollar =


sbesson:
A very quick auditing of a few CZI files in the curated QA repository seems to reveal that `Information > Document > CreationDate` is always present when`Information > Image > AcquisitionDateAndTime` exists but the reverse statement is not true. 

For most examples, the values of both dates are very close if not identical:

```
$ find /data/public/ -iname *.czi -exec sh -c "echo {} && /bio-formats-build/bioformats/tools/showinf -nopix {} | grep Date" \;
...
/data/public/idr0011/Plate1-Blue-A_TS-Stinger/Plate1-Blue-A-36-Scene-3-P1-D7-03.czi
Information|Document|CreationDate #1: 2013-12-10T17:46:40.7586388+00:00
Information|Image|AcquisitionDateAndTime #1: 2013-12-10T17:46:40.7596389Z
Scaling|AutoScaling|CreationDateTime #1: 01/01/0001 00:00:00
```

Unless there is a reason to consider `Information > Document > CreationDate` as the ground truth for all filesets, falling back to `Information|Document|CreationDate` if `Information|Image|AcquisitionDateAndTime` if present looks like a safe addition.

Configuration files should probably review across the curated QA repository accordingly.

dgault:
Closing temporarily to prevent merge conflicts

PR #3431, opened 2019-09-05
ICS: Update y start value when invertY

dgault:
This is a follow up to the thread https://forum.image.sc/t/trouble-with-converter-testconvert-from-bio-formats-tools/29189/5

When using an ICS file from SVI-Huygens then using tiled reading produces inverted bands as in the thread. This fix aims to resolve that issue similar to other readers such as https://github.com/ome/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/BMPReader.java#L122

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/ICSReader.java
+++ b/components/formats-bsd/src/loci/formats/in/ICSReader.java
@@ -608,6 +608,11 @@ public class ICSReader extends FormatReader {
     int len = FormatTools.getPlaneSize(this);
     int rowLen = FormatTools.getPlaneSize(this, w, 1);
 
+    // Update y value when invertY and reading tiles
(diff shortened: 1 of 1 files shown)


melissalinkert:
Change makes sense to me.  This is something that https://github.com/ome/bioformats/blob/develop/components/test-suite/src/loci/tests/testng/OpenBytesPerformanceTest.java should have caught if it was running against an appropriate file, so maybe worth resurrecting that test as a follow-up?

dgault:
Note: Latest CI is red as the config changes for this PR have been merged as part of the large config work: https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/

PR #3433, opened 2019-09-06
MetaMorph: Fix for finding first valid file

dgault:
This fix came about from a series of QA files (QA-27845 - QA-27863)

The nd file provided does have valid Tiff associated with it, it is expecting 3 channels (561, 488, 405), but only 2 are present for that nd file. The first channel is missing and resulted in being unable to locate any valid files due to a bug in the nested loop params.

With this fix the nd file should open and 2 of the channels will be correctly populated, the first channel will remain blank.


melissalinkert:
Should I revert https://github.com/ome/bioformats/pull/3425/commits/f514df4fff6eba1b2cbeedfc5a06853d48e941c5 so this doesn't conflict?

dgault:
Ah, didnt realise you already had it included. Will close this PR

PR #3435, opened 2019-09-10
AmiraParameters: Use warning for empty key

dgault:
This is an initial test PR for an issue reported on https://forum.image.sc/t/error-importing-am-files-bio-formats/29358

A sample file is expected via ftp which will be used for further testing.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tools/AmiraParameters.java
+++ b/components/formats-bsd/src/loci/formats/tools/AmiraParameters.java
@@ -40,6 +40,9 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import loci.common.RandomAccessInputStream;
 import loci.formats.FormatException;
 
@@ -63,6 +66,8 @@ public class AmiraParameters {
 
   protected int column, row;
   protected char c;
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(AmiraParameters.class);
 
   public AmiraParameters(String path) throws FormatException {
     readFile(path);
@@ -380,7 +385,11 @@ public class AmiraParameters {
       }
 
       String key = readKey();
-      if (key.equals("")) syntaxError("Invalid key");
+      if (key.equals("")) {
+        LOGGER.warn("Syntax warning:" + row + ":" + column + ": " + "Invalid key");
+        skipWhiteSpace();
+        continue;
+      }
 
       skipWhiteSpace();
       Object value;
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/tools/AmiraParameters.java
```diff
@@ -380,7 +385,11 @@ protected Map readMap() throws FormatException, IOException {
       }
 
       String key = readKey();
-      if (key.equals("")) syntaxError("Invalid key");
+      if (key.equals("")) {
+        LOGGER.warn("Syntax warning:" + row + ":" + column + ": " + "Invalid key");
```
joshmoore: Not that it needs doing in this PR, but I can see this being wrapped into a helper method managed by a metadata option to choose between strict and lenient behavior.

sbesson (commented):
The CI tests have been passing with this PR included  which demonstrates no apparent regression  on existing Amira filesets although our coverage of this particular file format might not be as thorough as we wish. From the original description and the image.sc thread, it looks like we have a sample file allowing to reproduce the issue. Is it available somewhere?

In the absence of a sample file, I would have a similar question as @joshmoore in https://github.com/ome/bioformats/pull/3435#discussion_r322742494 i.e. is there a scenario where we would like the reader to error rather than being lenient and throwing a warning?

sbesson (changes requested):
The CI repository tests demonstrate this change is not causing regressions in existing samples.

In order to reproduce the initial bug and validate the fix, I created a derived version from the public sample file at https://downloads.openmicroscopy.org/images/AmiraMesh/ignacio/ and injected an empty key under `Parameters`

```diff
        CoordType "uniform"
        Content "121x154x114 byte, uniform coordinates"
        BoundingBox 0.0 120.0 0.0 153.0 0.0 113.0
+        "" "test"
}
```

This is sufficient to trigger the code path proposed in this PR but leads to an infinite loop

```
build@32cca0aab56c:/uod/idr/scratch/inbox/amira/samples$ showinf -nopix test.am 
Checking file format [Amira]
Initializing reader
AmiraReader initializing test.am
Syntax warning:9:10: Invalid key
Syntax warning:9:10: Invalid key
Syntax warning:9:10: Invalid key
Syntax warning:9:10: Invalid key
Syntax warning:9:10: Invalid key
Syntax warning:9:10: Invalid key
Syntax warning:9:10: Invalid key
...
```

In the absence of a sample file reproducing the original issue, I could not convince myself that my synthetic example above would be a valid representation that we should use for testing this PR. If so, I would propose to update this PR to fix the initialisation in order to get the fix released.

sbesson (approved):
Retested using the synthetic sample file mentioned above. With the current release of Bio-Formats, initialisation fails as described in the original issue. With the latest state of this PR, the reader initialises with a warning indicating the presence of an invalid key.

The sample file has been configured and should be tested as part of the upcoming nightly CI builds. In the absence of an external example supplied by the original reporter, this is a best attempt to address the stack trace reported in the original image.sc thread and does not destabilizes any of our representative samples.

PR #3437, opened 2019-09-11
Make AcquisitionDate check more strict

melissalinkert:
The DeltaT test is now much more likely to fail, since the configuration
must exactly match what is reported by MetadataStore.

The test should now fail on any of the following:
  - configuration date is null, but stored date is not null
  - configuration date is not null, but stored date is null
  - both configuration and stored dates are not null, but are not equal

The test should pass if:
  - both configuration and stored dates are null
  - both configuration and stored dates are not null and equal

As discussed in the formats meeting, this will probably cause a whole bunch of test failures.  I'm happy to have this run once, gather the list of failing folders, and then exclude until I can get a configuration PR opened.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OpenlabRawReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OpenlabRawReader.java
@@ -146,13 +146,12 @@ public class OpenlabRawReader extends FormatReader {
     in.skipBytes(1);
 
     long stampMs = in.readLong();
+    String stamp = null;
     if (stampMs > 0) {
       stampMs /= 1000000;
       stampMs -= (67 * 365.25 * 24 * 60 * 60);
+      stamp = DateTools.convertDate(stampMs, DateTools.UNIX);
     }
-    else stampMs = System.currentTimeMillis();
-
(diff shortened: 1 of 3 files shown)


melissalinkert:
This does expose some problems with a few formats that store the month name or abbreviation instead of the month number.  In these cases, the test will fail when the ```user.language``` is set to something other than ```en```.

```git grep MMM *Reader.java``` identifies 16 readers to check, and I can confirm that ```AIMReader``` for sure shows the problem.

sbesson:
Short summary from the discussion yesterday with the @ome/formats team re https://github.com/ome/bioformats/pull/3437#issuecomment-531899787.

- the locale-dependent acquisition date detection is an issue that exists in the current stable version Bio-Formats
- the current assumption is that the major acquisition dates from our curated QA repository using this format are stored using an English locale
- next step is to configure all acquisition dates assuming an English locale and get the test-repo job passing with the default locale configuration
- once this repository is passing, it might be possible to run in a different locale e.g. `fr_FR` to identify whether we have representative sample filesets where the date is stored a different locale. In all cases, we will need evidence via. representative samples
- the `DateTools` parsing utilities will need to be reviewed and extended to allow the date parsing in a given locale

PR #1553, opened 2015-01-28
Fix NDPI and Leica SCN metadata issues exposed by Virtual Microscope

melissalinkert:
See files in team/atarkowska/virtual-microscope/VLE_for_edu/.  With this change, magnifications should be parsed and stored for NDPI files, and magnifications/physical pixel sizes should be set correctly for Leica SCN data (regardless of whether or not pyramid resolutions are flattened).  Note that there is also some cleanup of the Leica SCN reader to make it more maintainable going forward.

/cc @aleksandra-tarkowska

PR #3441, opened 2019-09-19
NDPIReader: fix nominal magnification population

### LINKED ISSUE #3439: Incorrect objective magnification reported for NDPI images
The NDPIReader does not seem to return the correct magnification level of the source image.
Currently, it reads from the field "Objective.Lens.Magnificant" - I believe it should instead read this magnification from the SOURCE_LENS field (TIFF tag 65421), which is 
already used inside the NDPIReader.

More information on the magnification field can be found here:
https://openslide.org/formats/hamamatsu/

"openslide.objective-power": normalized hamamatsu.SourceLens

sbesson:
Fixes #3439 

Support for the magnification in the Hamamatsu NDPI reader was introduced in Bio-Formats 5.1.0 (#1553). The original implementation makes use of the `Objective.Lens.Magnificant` key in the metadata TIFF tag (65449). Across our samples, the value is found to be always constant and equal to 35.16 for the largest resolution of the pyramid.

This PR adjusts the reader to populate the nominal magnification from the SourceLens TIFF tag (65449) for the pyramidal images. This is in agreement with what NDP.view displays when opening NDPI samples files

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPIReader.java
@@ -27,6 +27,7 @@ package loci.formats.in;
 
 import java.io.IOException;
 
+import loci.common.DataTools;
 import loci.common.DateTools;
 import loci.common.RandomAccessInputStream;
 import loci.common.services.ServiceException;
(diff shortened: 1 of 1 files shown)


dgault:
Tested with a number of files including the recommended https://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/manuel/test3-DAPI%202%20(387)%20.ndpi

With this PR the magnification is updated to `40.0` from `35.16`:
`<Objective ID="Objective:0:0" NominalMagnification="40.0"/>`

Opening the file with NDP View shows the `Source Lens` as `40x` which appears to match.

PR #3443, opened 2019-09-25
BDV: Parse physical sizes from XML

dgault:
This is a follow up to the forum thread https://forum.image.sc/t/bio-formats-xml-hdf5-reader-voxel-dimensions/29645/2

The data is present in the XML under ViewSetup as below:
```
        <voxelSize>
          <unit>um</unit>
          <size>0.3518820106983185 0.3518820106983185 0.3518820106983185</size>
        </voxelSize>
```

An example dataset with this data present is 100-U-9-MV-f0

This will be memo breaking


dgault:
Updated to fix: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/25733/consoleText

Without delving too deep into the parser code it seems that some of the lines are being split and read in 2 calls. As such the last commit is appending the parsed data and resetting the value at the end of the tag.

PR #3447, opened 2019-10-04
Artifactory deployment

sbesson:
This proposes to update the deployment mechanism of the Bio-Formats JARs (SNAPSHOT and release) to use Travis CI rather than relying on a secondary Jenkins job.

This logic is identical to what is in place for the downstream examples and documentation repositories and has been tested in the context of IDR - see https://github.com/IDR/bioformats/pull/16.

If accepted, the [Jenkins job](https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-latest-maven/) achieving this task will become redundant and should be disabled and eventually archived when this PR is merged.


dgault:
The updated settings look to be correct. Happy to use this updated deployment process for 6.3.0

PR #3452, opened 2019-10-11
Operetta: prevent series from having X or Y set to 0

melissalinkert:
Backported from a private PR.

An artificial test plate is in ```curated/perkinelmer-operetta/samples/no-first-series```.  Without this change, ```showinf -omexml Index.idx.xml``` will throw an ```IllegalArgumentException```.  This is because the first field in the first well is not recorded in the ```Index.idx.xml```, so X and Y sizes are not assigned to the first series.

With this change, the same command should result in blank planes being displayed. The first series is not removed because that would make it difficult to compare fields across wells.

Other plates and memo files should be unaffected, so this could wait for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -458,6 +458,31 @@ public class OperettaReader extends FormatReader {
       }
     }
 
+    // some series may not have valid XY dimensions,
+    // particularly if the first series was not recorded in the Index.idx.xml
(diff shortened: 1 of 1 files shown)


dgault:
Tests and builds have been green with this PR included.

Tested using the sample file `perkinelmer-operetta/samples/no-first-series`:
Without the PR this fileset produces an exception - `java.lang.IllegalArgumentException: 0 must be non-null and strictly positive.`

With the PR included the file is read and displays without error showing 8 series, each with 2 planes

PR #3453, opened 2019-10-21
Bump ome-common version to 6.0.4

sbesson:
See https://github.com/ome/ome-common-java/blob/v6.0.4/NEWS.md  for the list of changes especially https://github.com/ome/ome-common-java/pull/48

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
     <slf4j.version>1.7.6</slf4j.version>
     <kryo.version>2.24.0</kryo.version>
     <testng.version>6.8</testng.version>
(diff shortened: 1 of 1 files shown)


dgault:
Tested the merged Bio-Formats build with the `ome-common 6.0.4` version bump and some of the files which had failed in latest (https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/152/console). 

With the latest common PR's included the behaviour is as expected and described in https://github.com/ome/ome-common-java/pull/48.

No other regressions or test failures have occurred as result of this version bump. Good for inclusion in 6.3.0

PR #3458, opened 2019-10-30
Add a null check to LIFReader.checkFlip

sbesson:
This should prevent `NullPointerException` using some variants of the LIF
file format as reported in https://forum.image.sc/t/lif-cannot-open-with-bioformats-6-3-0/30856/

/cc @NicoKiaru


dgault:
Retested with the sample file on the forum thread and the file opens and displays without exception. As mentioned in the config PR some follow up to properly populate the missing position is required but getting the NPE fixed with this PR is a good starting point.

PR #3463, opened 2019-11-11
bfconvert: fix the writer's XY coordinate when the input image is cropped

### LINKED ISSUE #3461: bfconvert and showinf 6.3.0 issues
1) bvconvert issue
D:\vsi>bfconvert -series 7 -tiley 512 -tiley 512 -compression JPEG -crop 32256,8704,5120,5120 image.vsi converted.tif
image.vsi
CellSensReader initializing image.vsi
[CellSens VSI] -> converted.tif [Tagged Image File Format]
Tile size = 512 x 512
Exception in thread "main" loci.formats.FormatException: X:32256 must be < 5120
        at loci.formats.FormatWriter.checkParams(FormatWriter.java:453)
        at loci.formats.out.TiffWriter.saveBytes(TiffWriter.java:220)
        at loci.formats.tools.ImageConverter.convertTilePlane(ImageConverter.java:878)
        at loci.formats.tools.ImageConverter.convertPlane(ImageConverter.java:763)
        at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:691)
        at loci.formats.tools.ImageConverter.main(ImageConverter.java:1051)

The image in that series is 49512x38578.
The command line option "-crop" says:
 -crop: crop images before converting; argument is 'x,y,w,h'

2) showinf issue
The command line option "-series" says:
-series: specify which image series to read

If I leave off -series on say my above vsi image, all 16 series info shows.  If I put a -series 7, 16 series will be displayed.  Shouldn't it only show the info for series 7?

melissalinkert:
Fixes #3461.  The previous behavior was to use the same XY coordinate when reading and writing a tile.  This is fine when there is no cropping, but otherwise results in a coordinate that fails the writer's bounds check.

Without this PR, ``` bfconvert -crop 32256,8704,5120,5120 "test&sizeX=49512&sizeY=38578.fake" test.tif``` should throw:

```
Exception in thread "main" loci.formats.FormatException: X:32256 must be < 5120
	at loci.formats.FormatWriter.checkParams(FormatWriter.java:453)
	at loci.formats.out.TiffWriter.saveBytes(TiffWriter.java:220)
	at loci.formats.tools.ImageConverter.convertTilePlane(ImageConverter.java:878)
	at loci.formats.tools.ImageConverter.convertPlane(ImageConverter.java:763)
	at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:691)
	at loci.formats.tools.ImageConverter.main(ImageConverter.java:1051)
```

With this PR, the same test should successfully write a 5120x5120 file without throwing an exception.  The resulting file should be read correctly as well.

### DIFF
--- a/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
+++ b/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
@@ -65,6 +65,7 @@ public class ImageConverterTest {
 
   private Path tempDir;
   private File outFile;
+  private int width = 512;
   private final SecurityManager oldSecurityManager = System.getSecurityManager();
   private final PrintStream oldOut = System.out;
   private final PrintStream oldErr = System.err;
@@ -96,6 +97,7 @@ public class ImageConverterTest {
 
     tempDir = Files.createTempDirectory(this.getClass().getName());
     tempDir.toFile().deleteOnExit();
+    width = 512;
   }
 
   @AfterMethod
@@ -123,7 +125,7 @@ public class ImageConverterTest {
   public void checkImage() throws FormatException, IOException {
     IFormatReader r = new ImageReader();
     r.setId(outFile.getAbsolutePath());
-    assertEquals(r.getSizeX(), 512);
+    assertEquals(r.getSizeX(), width);
     r.close();
   }
 
@@ -215,5 +217,21 @@ public class ImageConverterTest {
             "plate&plates=1&fields=2.fake", outFile.getAbsolutePath()
     };
     assertConversion(args);
+  }
+
+  @Test
+  public void testCrop() throws FormatException, IOException {
+    outFile = tempDir.resolve("test.ome.tiff").toFile();
+    String[] args = {
+      "-tilex", "128", "-tiley", "128",
+      "-crop", "256,256,256,256", "test.fake", outFile.getAbsolutePath()};
+    width = 256;
+    try {
+      ImageConverter.main(args);
+    } catch (ExitException e) {
+      outFile.deleteOnExit();
+      assertEquals(e.status, 0);
+      checkImage();
+    }
   }
 }
--- a/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
+++ b/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
@@ -234,4 +234,22 @@ public class ImageConverterTest {
       checkImage();
     }
   }
+
+  @Test
+  public void testCropOddTileSize() throws FormatException, IOException {
+    outFile = tempDir.resolve("odd-test.ome.tiff").toFile();
+    String[] args = {
+      "-tilex", "128", "-tiley", "128",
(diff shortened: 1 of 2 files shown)


sbesson:
:+1: sbesson@1b3199e129cc91fb4b0023560d4ebedb42f49ff7 contains a minimal unit test reproducing the original failure and which should pass with this fix included.

melissalinkert:
Thanks for the test; commit cherry-picked onto this PR's branch as 0913d76.

mtbc:
Just beforehand there's the conditional,
```java
          if (nTileRows > 1) {
            tileY = 0;
          }
          if (nTileCols > 1) {
            tileX = 0;
          }
        }
```
It's okay that we're unzeroing them again for the tile writing?

components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
```diff
@@ -216,4 +218,20 @@ public void testSPWSeries() throws FormatException, IOException {
     };
     assertConversion(args);
   }
+
+  @Test
+  public void testCrop() throws FormatException, IOException {
+    outFile = tempDir.resolve("test.ome.tiff").toFile();
+    String[] args = {
+      "-tilex", "128", "-tiley", "128",
+      "-crop", "256,256,256,256", "test.fake", outFile.getAbsolutePath()};
```
mtbc: Maybe could also include a test that isn't exactly on tile boundaries.

mtbc:
Despite comments (which arise mostly from my ignorance), actually looks great and works fine as far as I can tell, no objections here. :+1:

melissalinkert:
> Just beforehand there's the conditional,
> 
> ```java
>           if (nTileRows > 1) {
>             tileY = 0;
>           }
>           if (nTileCols > 1) {
>             tileX = 0;
>           }
>         }
> ```
> 
> It's okay that we're unzeroing them again for the tile writing?

That block (https://github.com/ome/bioformats/pull/3463/files#diff-3ab3923a0dab584a4eadb99362e60687R857) only occurs if this condition is met:

https://github.com/ome/bioformats/pull/3463/files#diff-3ab3923a0dab584a4eadb99362e60687R831

which should only happen if we're writing tiles to separate files, which isn't the case here.  Admittedly, I didn't test the separate files case and I don't think we have tests for it either - maybe warrants a follow-up issue?

components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
melissalinkert: Added in 7a11de2

mtbc:
Certainly happy for the question to be relegated to a follow-up issue, it'll be good to get this present fix in.

components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
mtbc: :+1:

dgault:
Merging this initial fix as it fixes the initial issue and the unit test looks good. Further tests and fixes will be added in follow up PR's

PR #3465, opened 2019-11-14
TiffSaver: make tile writing faster

melissalinkert:
Mentioned in formats meeting 2019-10-30.  See also discussion in https://github.com/glencoesoftware/raw2ometiff/pull/1.

This should prevent unnecessary array copying when writing a tile buffer whose size matches the expected tile dimensions.  The overall conversion time for large tiled images should be noticeably faster with this PR.  ```bfconvert``` from an NDPI/SVS/etc. to pyramid OME-TIFF would be a good test.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
@@ -335,7 +335,10 @@ public class TiffSaver {
 
       // write pixel strips to output buffers
       // Check for the sane cases
-      if (ifd.getImageWidth() == w && ifd.getTileWidth() == w && channelsAllSameSize) {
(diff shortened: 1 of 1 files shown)


dgault:
I've run some initial tests on this and it looks promising. The converted images (I've been using ndpi) open and display as expected without issue and there have been no regressions. The conversion times (using bfconvert) are reduced as hoped. I will continue running some more variations of these tests and hopefully get this merged tomorrow.

dgault:
I've carried out further testing today across a range of NDPI files with various compressions and tile sizes. Each saved without error and the images are able to be read and displayed as before without any noticeable difference. The speed up varied across the results but fell in the 10-25% range depending on the tile sizes. Overall this looks good to merge.

PR #3466, opened 2019-11-18
ICSWriter: PhysicalSize NPE fix

dgault:
This is a fix for the NPE reported in https://forum.image.sc/t/bio-formats-ics-writer-bug/31423
The issue was due to the Units of the physical sizes being Pixels, which cant be converted to micrometer.

To test:

- From FIJI open the sample file Bat Cochlea Volume
- Using the Bio-Formats Exporter try to save the file as .ics
- Without the PR this should throw a NullPointerException
- With the PR the file should be saved without error
- Verify that the ICS file can be imported and displays as expected

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/ICSWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/ICSWriter.java
@@ -259,19 +259,19 @@ public class ICSWriter extends FormatWriter {
         char dim = outputOrder.charAt(i);
         Number value = 1.0;
         if (dim == 'X') {
-          if (meta.getPixelsPhysicalSizeX(0) != null) {
+          if (meta.getPixelsPhysicalSizeX(0) != null && meta.getPixelsPhysicalSizeX(0).value(UNITS.MICROMETER) != null) {
             value = meta.getPixelsPhysicalSizeX(0).value(UNITS.MICROMETER).doubleValue();
           }
           units.append("micrometers\t");
         }
         else if (dim == 'Y') {
-          if (meta.getPixelsPhysicalSizeY(0) != null) {
+          if (meta.getPixelsPhysicalSizeY(0) != null && meta.getPixelsPhysicalSizeY(0).value(UNITS.MICROMETER) != null) {
             value = meta.getPixelsPhysicalSizeY(0).value(UNITS.MICROMETER).doubleValue();
           }
           units.append("micrometers\t");
         }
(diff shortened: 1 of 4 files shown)


melissalinkert:
It sounds like this fixes https://trello.com/c/OG2iQY7q/329-ics-writer-npe-when-no-physical-size too?

sbesson:
Using the command-line utilites, created a minimal example reproducing the issue:

```
sbesson@ls30630:Downloads $ bfconvert "test&physicalSizeX=1pixel.fake" test.ics
test&physicalSizeX=1pixel.fake
FakeReader initializing test&physicalSizeX=1pixel.fake
[Simulated data] -> test.ics [Image Cytometry Standard]
	at loci.formats.out.ICSWriter.setId(ICSWriter.java:263)
	at loci.formats.ImageWriter.setId(ImageWriter.java:463)
	at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:671)
	at loci.formats.tools.ImageConverter.main(ImageConverter.java:1051)
```

and confirmed that the current pr fixes the conversion of images with physical sizes non-convertible into metric units as ICS:


```
sbesson@ls30630:Downloads $ rm test.ics 
sbesson@ls30630:Downloads $ ./bftools/bfconvert  "test&physicalSizeX=1pixel.fake" test.ics
test&physicalSizeX=1pixel.fake
FakeReader initializing test&physicalSizeX=1pixel.fake
[Simulated data] -> test.ics [Image Cytometry Standard]
	Converted 1/1 planes (100%)
[done]
0.598s elapsed (12.0+1.0ms per plane, 574ms overhead)
````

Code changes look minimal and should cover all physical sizes. To prevent any regression, would it be worth looking into creating a `ICSWriterTest` class with a unit test covering this particular scenario?

dgault:
Yeah I will put together a unit test to cover it for regressions

dgault:
New unit tests added, mostly basic tests based on the TiffWriter tests. If doing further work on the ICSWriter tests then more can be added and WriterUtilities should be updated to make it more flexible for non TIFF Writers.

There is also a test which covers the scenario of this PR: https://github.com/ome/bioformats/pull/3466/commits/09577401590470815b098e3cfd2cb24ec8403c79#diff-07f549d5e4cc20b3884341bc0dfd537eR100

This should fail and throw an NPE when run on its own without the original fix.

PR #3470, opened 2019-11-21
Zeiss CZI: fix some tile stitching and position size issues

melissalinkert:
Backported from a private PR.  See also https://trello.com/c/7MRDBwSn/178-czi-file-not-automatically-stitching

3add3da fixes a problem with the XY sizes when there are multiple positions ("Scenes" in ZEN).  Without this PR, ```qa-17179/WT_calc_after24h_stitch_2.czi``` had the same XY size for all 3 series.  With this PR, the sizes vary slightly between the 3 series.

604a1a0 fixes a problem where some datasets with multiple positions ("Scenes") and multiple mosaics (tiles) were not automatically stitched.  Without this PR, each position and mosaic in ```michael/multipos-z-time-*-tile.czi``` is split into a separate series.  With this PR, the mosaics are correctly stitched.

50da2d0 adds basic parsing for the ```Events``` annotation type, which was required in the original private PR but could be split into a separate PR if that helps with review.

The three files mentioned above need configuration changes; I'll put screenshots from ZEN on the corresponding PR.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -2821,6 +2821,37 @@ public class ZeissCZIReader extends FormatReader {
           shape = populateRectangles(text, roiCount, shape);
         }
 
+        NodeList events = getGrandchildren(layer, "Elements", "Events");
+        if (events != null) {
+          for (int s=0; s<events.getLength(); s++) {
+            Element event = (Element) events.item(s);
+
+            Element geometry = getFirstNode(event, "Geometry");
+            Element textElements = getFirstNode(event, "TextElements");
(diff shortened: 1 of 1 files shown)


dgault:
Builds and tests have been green with this PR included. Tested with the sample files recommended.

For file `qa-17179/WT_calc_after24h_stitch_2.czi`, without the PR the series all had the same dimensions of `1470 x 1475`. With this PR the dimensions are now `1470 x 1475`, `1470 x 1470` and `1469 x 1468`. 

For file `michael/multipos-z-time-*-tile.czi`, without this PR each scene is split into a separate series. With this PR the stitching is correctly applied.

Adding support for the `Events` annotation type does not appear to have any negative impact on existing datasets so adding support for it here is fine.

PR #3471, opened 2019-11-22
QPTIFF: metadata fixes

melissalinkert:
Backported from a private PR.

9e15251 fixes how the ```X_POSITION``` and ```Y_POSITION``` TIFF tags are recorded in the original metadata table for all TIFFs.  Without this change, the ```XPosition``` and ```YPosition``` values were treated as integers, and now they are treated as doubles.  ```showinf -nopix``` on any of the *.qptiff files in ```data_repo/curated/vectra-qptiff/``` should show the difference, and ```tiffinfo /path/to/file | grep Position``` can confirm that this PR is correct.

e9fc01a uses the ```X_POSITION``` and ```Y_POSITION``` TIFF tags to set ```Plane.PositionX``` and ```Plane.PositionY``` in the OME-XML for QPTIFF files.  ```showinf -nopix -omexml``` on any of the *.qptiff files in ```data_repo/curated/vectra-qptiff/``` should show that both values are populated with this PR.

a65cfe7 adds support for known companion files to QPTIFF.  As far as I know, this is a new addition since the instruments/software that produce QPTIFFs were acquired by Akoya Biosciences.  This commit will affect memo files, so the PR will require a minor release as-is.  I can split it into a separate PR if necessary.

All tests should continue to pass without configuration changes.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/BaseTiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/BaseTiffReader.java
@@ -230,15 +230,8 @@ public abstract class BaseTiffReader extends MinimalTiffReader {
     putInt("MinSampleValue", firstIFD, IFD.MIN_SAMPLE_VALUE);
     putInt("MaxSampleValue", firstIFD, IFD.MAX_SAMPLE_VALUE);
 
-    TiffRational xResolution = firstIFD.getIFDRationalValue(IFD.X_RESOLUTION);
-    TiffRational yResolution = firstIFD.getIFDRationalValue(IFD.Y_RESOLUTION);
-
-    if (xResolution != null) {
-      put("XResolution", xResolution.doubleValue());
-    }
-    if (yResolution != null) {
-      put("YResolution", yResolution.doubleValue());
-    }
+    putDouble("XResolution", firstIFD, IFD.X_RESOLUTION);
+    putDouble("YResolution", firstIFD, IFD.Y_RESOLUTION);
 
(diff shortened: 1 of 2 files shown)


sbesson:
Thanks @melissalinkert. Unless there is also an open request for these fixes, I think it's completely acceptable to keep it standalone for now and target it for the next minor version increment, currently `6.4.0`

chris-allan:
There is a bug with this PR that we are currently investigating. Please don't merge until you've heard from us.

melissalinkert:
c0decf7 fixes the issue referenced in https://github.com/ome/bioformats/pull/3471#issuecomment-564006628

dgault:
Without this PR the position data in the file from `tiffinfo /path/to/file | grep Position` is:
`Position: 0.595628, 3.25669`

The data is not present on the OME-XML plane element but is in global metadata as:
```
XPosition: 0
YPosition: 3
```

With this PR the position data is now seen in the OME-XML (in this case the ifd resolution unit was 3 for cm) as:
`<Plane ExposureTime="10000.0" ExposureTimeUnit="µs" PositionX="5956.2802305343275" PositionXUnit="µm" PositionY="32566.888350392826" PositionYUnit="µm" TheC="0" TheT="0" TheZ="0"/>`

and shows correctly in the global metadata as:
```
XPosition	0.5956280230534328
YPosition	3.2566888350392826
```
Overall, switching to using double rather than int makes sense and overall this has had no impact on the repo tests or builds.

PR #3477, opened 2019-12-04
ImageConverter: Fixes for separate tiles with other combined options

dgault:
This is a follow up to https://github.com/ome/bioformats/pull/3475 and fixes https://github.com/ome/bioformats/issues/3472

Along with the previous fix this was originally looking at the below scenario for bfconvert:
- series which is not the first
- tilex and tiley
- pattern output with %x and %y or %m

In adding the new unit tests it also raised an issue with:
- choosing a single channel, timepoint or Z
- pattern output with %x and %y or %m

The new unit tests should cover both of these test cases but manual testing to confirm the validity of the converted planes in both scenarios should be carried out.

The original sample file recommended is a good starting point for tests: https://downloads.openmicroscopy.org/images/Ventana/openslide/OS-1.bif

I also found the artifical sample files useful for ensuring the right planes have been converted: https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/


components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
```diff
@@ -252,4 +261,51 @@ public void testCropOddTileSize() throws FormatException, IOException {
       checkImage();
     }
   }
+  
+  @Test
+  public void testCropLargertThanTileSize() throws FormatException, IOException {
```
mtbc: Is the extra "t" intended?

mtbc:
Just a small thought -- seeing as we're working with both *x* and *y* tile dimensions I don't know if it's useful to have test coverage for them not being the same value to ensure that the code never wrongly mixes them up. Doesn't block this PR though.

sbesson:
Tested using the public [4D-series.ome.tiff](https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/4D-series.ome.tiff) and [HandEcompressed_Scan1.qptiff](https://downloads.openmicroscopy.org/images/Vectra-QPTIFF/perkinelmer/PKI_scans/HandEcompressed_Scan1.qptiff) samples

The following conversion commands were executed against both  

```
bfconvert -series 4 HandEcompressed_Scan1.qptiff HandEcompressed_Scan1.ome.tiff
bfconvert -tilex 512 -tiley 512 -series 4 HandEcompressed_Scan1.qptiff HandEcompressed_Scan1_tiled.ome.tiff
bfconvert -tilex 512 -tiley 512 -series 4 -crop 0,0,1024,1024 HandEcompressed_Scan1.qptiff HandEcompressed_Scan1_tiled_cropped.ome.tiff
bfconvert -timepoint 1 -channel 0 -z 4 4D-series.ome.tiff 4D-series_converted.ome.tiff

bfconvert -tilex 512 -tiley 512 -series 4 HandEcompressed_Scan1.qptiff HandEcompressed_Scan1_X%x_Y%y.ome.tiff
bfconvert -tilex 512 -tiley 512 -series 4 HandEcompressed_Scan1.qptiff HandEcompressed_Scan1_%m.ome.tiff
bfconvert -tilex 512 -tiley 512 -series 4 -crop 0,0,1024,1024 HandEcompressed_Scan1.qptiff HandEcompressed_Scan1_tiled_cropped_%m.ome.tiff
bfconvert -tilex 256 -tiley 256 -timepoint 1 -channel 0 -z 4 4D-series.ome.tiff 4D-series_%m.ome.tiff
bfconvert -tilex 256 -tiley 256 -timepoint 1 -channel 0 -z 4 4D-series.ome.tiff 4D-series_X%x_Y%y.ome.tiff
```

With Bio-Formats 6.3.0, the 5 last commands fail as expected from the opened issue and this PR.
With this PR, all conversion commands completed as executed and generated the expected number of files depending on the tiling/cropping conditions. A few comments:

- in the  default OS preview, all generated TIFF files render as expected  except for the ones with a tile Y smaller than tileY. I could not find anything obviously wrong with the TIFF metadata so this might be an artifact of the viewer itself

Screen Shot 2019-12-09 at 22 15 57

- some of the metadata mismatches between TIFF and OME-TIFF. For instance `4D-series_0.ome.tiff` is ` 256x167 tiled TIFF but the OME metadata indicates `SizeX="183"`
- `../bftools/bfconvert  -timepoint 1 -channel 0 -z 4 4D-series.ome.tiff 4D-series_%m.ome.tiff`  fails with a NPE which fails to explain the underlying issue (assuming it is the incompatibility between the tiled-based naming scheme and the non-tiled reading). As a RFE, an earlier detection and error might be usefu

Otherwise, the newly added tests should now improve the test coverage of the various use cases and prevent future regressions. Once merged, it might be useful to add some description of the `%m/%x/%y` expansion patterns in the utility help as well as the online documentation

dgault:
The idea of having different dimensions for the 2 tiles is something that should be added along with a similar test at the TiffWriter level. Additionally the same should probably be done for the crop sizes. I will create a separate issue for these new tests.

The spelling mistake in the test name has been corrected.

The patterns for `%m/%x/%y` are now present in the utility help

The NPE when using the tile name output but without the tile sizes set, this was actually a bug in the image converter, essentially it was assuming it was a tiled conversion and never setting the writer ID. This should now be resolved. (It will write a single non tiled image but the tile indexes in the name will be 0) 

The incorrect dimensions is a good spot and was definitely a bug. In the original scenario the converter would write the first image correctly (156 x 167) but before closing the writer it would update the metadata for the next ouput file and overwrite the sizeX and sizeY. This should now be corrected so the image sizes are correct for each tile.

PR #3478, opened 2019-12-05
OBF: Incorrect dimensions fix

dgault:
This is a follow up to address the issues which have been reported in https://github.com/ome/bioformats/pull/3141

In this scenario the dimensions are only updated from the label if they have not already been set.
The provided sample file can be used for testing. Without this PR the second series should throw an exception. With this PR it should read and display without error. 

I will open a config PR with the new sample file.


dgault:
The sample file for testing can be found at `inbox/from_bjoernthiel`

dgault:
Merging for release as part of 6.3.1

bjoernthiel:
@sbesson keep the sample image private please.

PR #3479, opened 2019-12-06
DeltavisionReader.java: synchronise objectives metadata with softWoRx 7.2.0

carandraug:
This is pretty much the same as #3204 . I used a [script](https://github.com/MicronOxford/scripts/blob/master/auto/dvlenses2readercases.pl) to parse the `dvlenses.tab` from the last release of softWoRx and convert to the java code required for the deltavision reader.

This does adds more objectives but also removes a few which I found odd since I don't know from where those were coming (mainly 33a19733b32b and a644a10df03e). At least the change made by ace7daf99e43 are odd because they mark them as Nikon objectives but they are not in the 12000-13999 range.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -1836,7 +1836,7 @@ public class DeltavisionReader extends FormatReader {
       case 10211: // Olympus U-Plan S-Apo 20X/0.75
         lensNA = 0.75;
         magnification = 20.0;
-        workingDistance = 0.65;
+        workingDistance = 0.60;
         immersion = MetadataTools.getImmersion("Air");
         model = "1-U2B825";
         break;
@@ -1989,6 +1989,14 @@ public class DeltavisionReader extends FormatReader {
         workingDistance = 0.2;
         immersion = MetadataTools.getImmersion("Oil");
         model = "1-U2B530";
+        break;
+      case 10417: // Olympus 40X/1.25, UPLSAPO 40XS, Super Apochromat Silicone Oil
+        lensNA = 1.25;
+        magnification = 40.0;
+        workingDistance = 0.3;
+        immersion = MetadataTools.getImmersion("Other");
+        model = "";
+        correction = MetadataTools.getCorrection("Apo");
         break;
       case 10000: // Olympus 100X/1.30, DApo/340, IMT2
         lensNA = 1.30;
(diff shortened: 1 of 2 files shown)


sbesson:
@carandraug thanks again for opening this periodic update of the DV lenses metadata. I think the mismatching lens IDs were  introduced in https://github.com/ome/bioformats/pull/3378 and primarily associated with the `rcpnl` variant of the Deltavision format.

In Bio-Formats 6.3.0, the logic for this format variant was split into a separate reader - see #3421. From your description, it feels like more rcpnl-specific metadata will need to migrate to `RCPNLReader.populateObjective()`. @melissalinkert might be able to confirm and/or provide more guidance here

melissalinkert:
Thanks @carandraug.  As @sbesson mentions, IDs 18107-18111 are used in the .rcpnl variant.  Removing them here is fine with me, but they will then need to be added to ```RCPNLReader```.  I can do that in a separate PR if that's easiest.

carandraug:
> IDs 18107-18111 are used in the .rcpnl variant

I'm not familiar with the rcpnl format but are its objective ids documented somewhere? What is weird about them is that it contradicts the deltavision documentation (Objectives manufactured by Nikon get an ID between 12000 and 13999.  Objectives with ID in the 18107-18111 range identify them as manufactured by API/GE -- which I get are not objective manufacturers).

> I can do that in a separate PR if that's easiest.

I guess you can just push commits to this PR.

carandraug:
On a related note, would it be of interest to add the data file with the lenses information and have an ant target to update the source file (or have the DeltaVisionReader source file being generated as part of the build?)

melissalinkert:
PR opened against this branch (I can't push directly to your branch): https://github.com/carandraug/bioformats/pull/1

The .rcpnl-specific IDs are not publicly documented, unfortunately.

I defer to @sbesson and @dgault on the question of adding build targets to update the objective information.  I'd assume we don't want to have ```dvlenses.tab``` committed, but maybe that's OK.

carandraug:
> PR opened against this branch (I can't push directly to your branch): carandraug#1

Now merged.

I think it would make sense to invert the order of the commits though. First move those IDs into `RCPNLReader` and then update the list on `DeltavisionReader`. I don't know how you fell about such history manipulation so I'll leave it up to you.

melissalinkert:
I don't have a problem with the commits being reordered, especially since this PR isn't yet included in the integration build.

carandraug:
I have reordered the commits as mentioned above. Here's some notes about this PR:

1) for the objective with lens id 10211, the working distance value has changed. I confirm that the value did change on the metadata provided by softworx between versions 6.5.2 and 7.2.0.

2) this also adds lens with ids 18107, 18108, and 18109 which are different from the ones in RCPLNReader.

3) this removes the case for the lens id 20007. This was added on issue #3411 after a [report on the image sc forum](https://forum.image.sc/t/warning-when-opening-dv-files-in-fiji/27923). The id is not included on any of the softworx versions we have around (I have checcked 6.0, 6.5.2, and 7.2.0). Is it possible that this lens id was added by some third party to cover an objective used there, and in which case it might mean another objective in another place?

dgault:
Thanks @carandraug, the new additions and updates look good. Moving the RCPNL specific data also makes sense. For id 20007 I think its best we err on the side of caution and leave it in for the time being, I will push a commit to re-add it and a few removed lines from 18107. 

Otherwise this PR has no negative impact on repo tests or builds and is good to merge and include in the 6.4.0 release.

Longer term I will have a think to see if there is a nice easy way we can automate these updates for future minor releases.

carandraug:
> For id 20007 I think its best we err on the side of caution and leave it in for the time being, I will push a commit to re-add it

Shouldn't err on the side of the caution be to not add it back? My interpretation of the situation suggests that this id might be site specific, so I would argue that it would be better to return no lens information (with a warning that there's no information of it) that one that is wrong.

> and a few removed lines from 18107.

I'm not sure those lines are correct. 18107 falls on the range 18000-19999 and so, per specs, manufacturer should be API/GE. I also don't see where the correction information comes from.

Either way, if those lines are to be kept, would be nice to have a comment explaining their origin since they're not documented anywhere. It will make it clear next time there's a softworx release with more objectives.

PR #3486, opened 2019-12-20
InCell: invert Y coordinate in plane/field positions

melissalinkert:
Backported from a private PR.

This corrects for the difference in how Y coordinates are treated between the acquisition software and our model (see the code comment).  To test, open ```incell/michael/firstStack``` (1 well, 6 fields in a 3x2 grid) in ImageJ with ```Open all series``` and ```Stitch tiles``` checked.  Without this PR, there should be a very obvious horizontal line in the middle of the stitched image even when zoomed out.  With this PR, the stitched image should look much better.

Shouldn't affect memo files, but does require a substantial configuration update to change every Y position for every InCell plate.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -1129,8 +1129,15 @@ public class InCellReader extends FormatReader {
           index = offsetPointCounter++;
         }
 
+        // Y value multiplied by -1 to correct for difference in assumptions
+        // - position is relative to center of well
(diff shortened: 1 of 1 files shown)


dgault:
Tested using `incell/michael/firstStack` as recommended.

With the PR included the PositionY values are altered and match those in the updated config files. The stitched tiles now appear inverted (as in the bottom row of tiles are now the top row). When combined this tiled combination looks to be correct and joins up better than the previous view.

With the updated configs the tests are green again and this looks good to merge.

PR #3488, opened 2019-12-30
CellWorx: add support for multiple Z sections

melissalinkert:
Backported from a private PR.

If multiple Z sections are present, each one is expected to be in a separate directory as described on line 898.  A test dataset is being uploaded to ```curated/cellworx/samples/multi-z/``` (generated from idr0005).  Without this PR, ```showinf -nopix 2011-04-19-plate-1.HTD``` should throw an exception.  With this PR, ```SizeZ``` should be 2 for every series.  ```showinf -trace 2011-04-19-plate-1.HTD -series n``` for any series should show that the 2 planes are coming from different files.  The planes won't be visually different, so it's worth checking the file names. 

Existing tests should continue to pass.  This will impact memo files, so requires a minor version.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
@@ -937,9 +937,8 @@ public class CellWorxReader extends FormatReader {
           }
         }
         if (nextFile != files.length) {
-          throw new FormatException(
-            "Well " + well + " expected " + files.length +
-            " files; found " + nextFile);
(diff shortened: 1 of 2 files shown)


dgault:
Tested using the sample dataset `curated/cellworx/samples/multi-z/`
Without this PR the file is not able to be read and throws an exception.

With this PR the file is read and planes displayed without exception. Each series has 2 Z planes and using `showinf -trace 2011-04-19-plate-1.HTD -series n` shows the planes coming from different files.

No other repo tests failed and with the new dataset configured the tests are all green. PR is good to merge.

PR #3489, opened 2020-01-06
Fix a handful of deprecation warnings

melissalinkert:
Cleans up a few places where readers were using a deprecated method or variable.  There should be no behavior changes or failing tests; the only difference is that the build log should be a bit cleaner.  Should be safe for a patch release, but definitely not urgent.


dgault:
Switching to use MetadataTools and UNITS.MICROMETER look good. All tests have continued to pass and builds have been unaffected. Changes are good to merge.

Issue #3416, opened 2019-08-01
Ventana: Physical Size not being read


dgault:
This issue was first raised in https://forum.image.sc/t/missing-metadata-for-ventana-images-in-qupath-with-bio-formats/28123

A sample file is available in QA-27763 and reproduces the issue.

From debugging it seems that most of our Ventana samples have the XML on ifd 2 which is the first resolution (0 and 1 being a label and thumbnail) - https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/VentanaReader.java#L460. For the provided QA there is no XML tag present on this ifd, but instead on ifd 0.

dgault:
Resolved by https://github.com/ome/bioformats/pull/3494

PR #3494, opened 2020-01-08
Ventana: try harder to find physical sizes for prestitched TIFFs

melissalinkert:
See #3416.

Without this PR, ```showinf -nopix -noflat -omexml``` on the file from QA 27763 should show that ```PhysicalSizeX``` and ```PhysicalSizeY``` are not included in the OME-XML.  With this PR, the same command should show both ```PhysicalSizeX``` and ```PhysicalSizeY``` set to ```0.25```.

Ventana TIFFs (which is what QA 27763 is) are typically converted from .bif files and don't have as much metadata stored in the IFDs for each pyramid level.  Most of the XML on the pyramid IFDs in a .bif file is related to stitching tiles; when the .bif file is converted to a Ventana TIFF, all of that XML is presumably stripped out since it's no longer relevant.

I wouldn't expect any test failures, and memo files should not be impacted.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/VentanaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/VentanaReader.java
@@ -456,7 +456,7 @@ public class VentanaReader extends BaseTiffReader {
       }
       String xml = ifds.get(index).getIFDTextValue(XML_TAG);
       LOGGER.debug("XMP tag for series #{} = {}", i, xml);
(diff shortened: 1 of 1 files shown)


dgault:
Tested with QA-27763, without this PR the physical sizes are not present. With this PR included the sizes are present in the OME-XML and are correct. No regressions have been seen with existing repo files and with the new config included the tests are now green.

PR #3495, opened 2020-01-09
FormatWriter: create output file's parent directory if needed

melissalinkert:
See https://trello.com/c/zxnORoSU/380-bfconvert-create-subfolders

Without this change, something like ```bfconvert test.fake plate/well/field/test.tiff``` (where ```./plate/well/field/``` doesn't exist) should throw an exception.  With this change, the conversion should be successful.

It's also worth checking something like ```bfconvert "test&sizeZ=3.fake" plate/well/%z/test.tiff``` where there is a pattern in the directory, just to make sure.


dgault:
Tested using the given example and also the pattern file example. Without this PR a `FileNotFoundException` exception is thrown. With this PR included there is no exception and the conversion completes successfully with the directories created as expected.

PR #3496, opened 2020-01-09
Make TiffParser and TiffSaver implement Closeable

melissalinkert:
See https://trello.com/c/JHikBmdv/368-make-tiffparser-and-tiffsaver-closeable and https://github.com/openmicroscopy/bioformats/pull/3331#pullrequestreview-223532670

```TiffSaver``` already implemented ```close()```, so the only change there is to explicitly implement ```Closeable``` which should be pretty safe.

```TiffParser``` needed a new ```close()``` method though, and I'm not 100% sure of the best approach.  Always closing the underlying ```RandomAccessInputStream``` would make sure that the file handle is always closed, but would be problematic for existing code of this form:

```
RandomAccessInputStream s = new RandomAccessInputStream("file.tiff");
TiffParser p = new TiffParser(s);
...
p.close();
s.seek(0);
```

The current approach is to only close the ```RandomAccessInputStream``` if it was opened by ```TiffParser```, i.e. if the ```TiffParser(String)``` constructor was used, which seems safer.  Happy to hear other thoughts.

Shouldn't impact tests, but I'd assume requires a minor release.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -32,6 +32,7 @@
 
 package loci.formats.tiff;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -59,7 +60,7 @@ import org.slf4j.LoggerFactory;
  * @author Melissa Linkert melissa at glencoesoftware.com
  * @author Chris Allan callan at blackcat.ca
  */
-public class TiffParser {
+public class TiffParser implements Closeable {
 
   // -- Constants --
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by joshmoore** [comment]
Use of the ctor to decide close/don't-close makes sense to me. I imagine if need be we could add other ctors (with a boolean?) or a static initializer, though likely subclassing would get most people what they need.

**Comment by dgault** [comment]
The logic used here makes sense with the existing code that is out there. For example we create new TiffParsers in isThisType to check if a stream is TIFF, in these cases you certainly wouldn't want the stream closed after. Also a lot of our other uses already assume that we close the stream after the fact such as:

```
      try (RandomAccessInputStream in = new RandomAccessInputStream(path)) {
        TiffParser parser = new TiffParser(in);
```

Happy to have this included as is, and as mentioned if it ever becomes necessary then worst case scenario we can add an additional constructor with boolean at a later date.

PR #3497, opened 2020-01-09
xmlvalid: fix lowercasing of file names

melissalinkert:
See https://trello.com/c/fa7RLnwL/366-character-case-in-validating-xml

Without this PR, this test should throw a ```FileNotFoundException```:

```
$ showinf -nopix -omexml-only test.fake > TEST.ome.xml
$ xmlvalid TEST.ome.xml
```

With this PR, the same test should successfully validate ```TEST.ome.xml```.


dgault:
The change make sense and when tested using the recommended example and some other suitable sample files the validation runs without exception and tests pass successfully and validate with the PR included.

Issue #3346, opened 2019-04-11
mkfake fails on ini file


joshmoore:
```
/tmp/fake $ mkfake .
/tmp/fake $ echo "sizeC=3" >>  screen.fake/Plate000/Run000/WellA000/Field000.fake.ini
/tmp/fake $ showinf screen.fake
Checking file format [Simulated data]
Initializing reader
FakeReader initializing screen.fake
Exception in thread "main" java.lang.NumberFormatException: For input string: "000.fake"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at loci.formats.in.FakeReader.extractTokensFromFakeSeries(FakeReader.java:1209)
	at loci.formats.in.FakeReader.initFile(FakeReader.java:555)
	at loci.formats.FormatReader.setId(FormatReader.java:1397)
	at loci.formats.ImageReader.setId(ImageReader.java:842)
	at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:650)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1034)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1120)
```

melissalinkert:
Closing in favor of https://github.com/ome/bioformats/pull/3499

PR #3499, opened 2020-01-09
Fake: allow .ini files in plates created by mkfake

melissalinkert:
See https://github.com/ome/bioformats/issues/3346

Without this PR, the following should throw a ```NumberFormatException``` as described in #3346 :

```
$ mkfake .
$ echo "sizeC=3" >>  screen.fake/Plate000/Run000/WellA000/Field000.fake.ini
$ showinf screen.fake
```

With this PR, the same test should successfully display one plane.  Note that ```SizeC``` will be 1 with the test case as written, since only the only .ini file parsed is the one matching what was passed to ```setId``` (```screen.fake.ini``` in this case).  To have ```SizeC``` set to 3, either run:

```
$ showinf screen.fake/Plate000/Run000/WellA000/Field000.fake
```

or:

```
$ echo "sizeC=3" >> screen.fake.ini
$ showinf screen.fake
```

I wouldn't expect this to have any test or memo impact, so should be fine for a patch release if necessary.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -1294,7 +1294,7 @@ public class FakeReader extends FormatReader {
         rows = ResourceNamer.alphabeticIndexCount(elements[0]);
         cols = Integer.parseInt(elements[1]) + 1;
       } else if (pathToken.startsWith(ResourceNamer.FIELD)) {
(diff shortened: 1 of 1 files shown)


dgault:
Tested using the recommended scenario. Without the PR this results in:
```
FakeReader initializing screen.fake
Exception in thread "main" java.lang.NumberFormatException: For input string: "000.fake"
```

With this PR the reader does not throw any exception and opens a single plane as expected.

PR #3500, opened 2020-01-16
Add a note about supplying IFD instances for TIFF writing.

mtbc:
Documents the advice given in https://github.com/ome/omero-downloader/issues/17#issuecomment-574896977.


dgault:
The documented advice makes sense and is good to have there.

PR #3505, opened 2020-01-31
NDPI: support JPEG-XR and files larger than 4 GB

melissalinkert:
Backported from private PRs.

Based on new (private) documentation, this updates NDPI support to allow JPEG-XR compressed tiles, full support for files larger than 4 GB, and some additional metadata tags.

JPEG-XR support required changing the component of the JPEG-XR codec and service from ```formats-gpl``` to ```formats-bsd```.  The ```jxrlib``` version also needed to be updated to include https://github.com/glencoesoftware/jxrlib/pull/20, since overview images in files that use JPEG-XR compression store channels in BGR order.

Definitely requires a minor release; I would expect memo files to be affected.  Requires some configuration changes to NDPI files larger than 4 GB, but otherwise should not fail existing tests.

I suspect, but have not yet thoroughly tested, that this will fix at least a few of the outstanding NDPI issues.  To be tested:

* https://trello.com/c/1jxzzrWk/332-ndpi-null-pointer-exception
* https://trello.com/c/YMsm1dzp/64-ndpi-to-tiff-conversion-is-failling
* https://github.com/ome/bioformats/issues/3059
* https://github.com/ome/bioformats/issues/2544
* https://github.com/ome/bioformats/issues/3432


melissalinkert:
311e1c8 is an attempt at fixing the ongoing repo test failures.  With it, I can no longer reproduce failures locally, but builds should probably be green for several days in a row before thinking about merging.

I don't like this as a long-term solution, since it will slow down JPEG-XR tile reading especially in use cases like viewing in OMERO.  Making jxrlib more thread safe (https://github.com/glencoesoftware/jxrlib/issues/22) is better long-term, but not a quick project.

Another option is to re-add the original ```JPEGXRServiceImpl``` decoding logic as a new service method, revert the changes to ```ZeissCZIReader```, and have ```ZeissCZIReader``` call the new service method.  That's not great either and still not a long-term fix, but would hopefully reduce the scope of the performance impact.

Happy to discuss.

melissalinkert:
One more try at fixing the ongoing test failures in 2381e5a.

dgault:
Builds and repo tests have been green for a number of days now with the latest commit.

Manually tested using the new and updated files from the config PR. Each opened and displayed without issue and the pixel and metadata looked sensible. The additional tags all look like positive additions and I cannot see them causing any issues with existing filesets.

I also sanity checked some of the CZI files which had been failing in the builds previously. These too opened and displayed without any issues.

Merging for release with 6.4.0

PR #3506, opened 2020-02-04
bfconvert: add -nobigtiff option

melissalinkert:
See https://github.com/ome/bioformats/issues/3404

Adds a ```setCanDetectBigTiff(boolean)``` method to ```TiffWriter``` and corresponding ```-nobigtiff``` option to ```bfconvert```.  If ```-nobigtiff``` is used (== ```setCanDetectBigTiff(false)```), then automatic switching to BigTIFF based upon the number of pixel bytes is turned off.  Extension-based switching is unchanged.

Using ```-bigtiff``` and ```-nobigtiff``` together should exit quickly with an error message, since the correct behavior is ambiguous.

Should not impact memo files or anything on the reading side of the stack, but does require a minor release due to new API.  I don't feel strongly that this needs to be in 6.4.0.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -106,6 +106,7 @@ public final class ImageConverter {
   private String compression = null;
   private boolean stitch = false, separate = false, merge = false, fill = false;
   private boolean bigtiff = false, group = true;
+  private boolean nobigtiff = false;
   private boolean printVersion = false;
   private boolean lookup = true;
   private boolean autoscale = false;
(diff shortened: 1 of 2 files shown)


dgault:
Tested conversions with a number of different formats. With the new option set the converted images are able to be read and displayed without issue. The message is displayed to show that BigTiff is disabled.

Converting a large file without compression and forcing bigtiff to be disabled will result in the following exception:
`FormatException: File is too large; call setBigTiff(true)`

Converting the same file with compression and forcing bigtiff to be disabled did result in the file being written correctly without exception.

Comparing the converted images with those converted as bigtiff and the pixel and metadata looks correct in both cases. 

The new API addition looks fine for a minor release and repo tests have been unaffected. I will update the docs to document the new option.

PR #3507, opened 2020-02-04
Zeiss CZI: check for uncompressed pixels incorrectly flagged as JPEG-XR

melissalinkert:
See https://github.com/ome/bioformats/issues/3401.

The JPEG-XR decompression warnings occurred because the pixel data blocks were actually uncompressed.  ```showinf -range 2 2 -series 8``` on the file from QA 27508 without this PR will show an image with all 0 pixel values.  With this PR, there should clearly be image data that aligns with the other channels.  Note that there are still blank areas in each corner, but that is correct.

I don't think this will affect tests (feel free to exclude if it does), and should be safe for a patch release.


dgault:
Tested with QA-27508

Without this PR the file opens with all 0 pixel values for the 3rd channel
With this PR included the file opens and the image displayed looks correct and matches the other channels in the series.

With this PR and the new config file the builds and repo tests are all green.

PR #3519, opened 2020-03-08
TIFF: prevent integer overflow when reading from a large tile

melissalinkert:
When reading from a single tile whose byte count is > 2GB,
use ```seek``` instead of ```skipBytes``` to prevent a possible integer overflow.

/cc @chris-allan, @douglasrennick

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -1026,7 +1026,9 @@ public class TiffParser implements Closeable {
           // we only want a piece of the tile, so read each row separately
           // this is especially necessary for large single-tile images
           int bpp = bytes * effectiveChannels;
-          in.skipBytes((int) (y * bpp * tileWidth));
+          // don't use skipBytes here, as the number of bytes to skip may
+          // be greater than Integer.MAX_VALUE if the tile is large
+          in.seek(in.getFilePointer() + (y * bpp * tileWidth));
           for (int row=0; row<height; row++) {
             in.skipBytes(x * bpp);
             int len = (int) Math.min(buf.length - offset, width * bpp);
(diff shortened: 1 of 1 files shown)


chris-allan:
As a matter of background, we were seeing very weird behaviour trying to do large pyramid conversions from multi-GB TIFFs with no internal pyramids.

/cc @seb, @dgault

components/formats-bsd/src/loci/formats/tiff/TiffParser.java
```diff
@@ -1026,7 +1026,9 @@ else if (stripByteCounts[countIndex] < 0 && countIndex > 0) {
           // we only want a piece of the tile, so read each row separately
           // this is especially necessary for large single-tile images
           int bpp = bytes * effectiveChannels;
-          in.skipBytes((int) (y * bpp * tileWidth));
+          // don't use skipBytes here, as the number of bytes to skip may
+          // be greater than Integer.MAX_VALUE if the tile is large
```
mtbc: Could just use `skipBytes(long)`.

melissalinkert: Forgot about that API method; fixed in b47dafa.

melissalinkert:
An artificial test file is now in ```inbox/gh-3519/```.  The test file contains a gradient from black at the top to white at the bottom.

To see the problem without this PR:

```
$ bfconvert -tilex 1024 -tiley 1024 -bigtiff -pyramid-scale 2 -pyramid-resolutions 4 large-gradient.tiff bfconvert.ome.tiff
$ showinf -noflat -resolution 3 bfconvert.ome.tiff
```

The top portion of the image should show a gradient as expected, but the bottom portion will be oddly striped.  With this PR, the same test should result in an image with a completely smooth gradient.

dgault:
Using the new sample file the behaviour can be easily reproduced and is visible across the different resolution levels. I couldn't find any obvious pattern in the starting point of the corruption though.

Screen Shot 2020-03-10 at 22 18 09


With the PR included the image is read and displayed without andy issues. There have been no regressions in any tests, merging this to included for 6.4.0

PR #3524, opened 2020-03-12
Added bfopen5D.m and modified bfsave.m.

RainerHeintzmann:
bfopen5D.m now opens up to 5D datasets, since bfopen seemed to not work correctly.

In bfsave.m I also made some changes. Also a warning is issued, if bfsave is called with a filename NOT ending with ome.tif changing the filename to be an ome tiff.

It would be great, if we could change the copyright for bfopen5D.m and bfsave.m to NOT be GPL-2 such that it then complies also with closed source distributions. The reason is that one may want to distribute a compiled Matlab version, which naturally does not come with source code. This would allow the user to install bioformats.jar separately from the distribution, but the bfopen5D and bfsave functions would need to be distributed closed source.

### DIFF
--- a/components/formats-gpl/matlab/bfopen5D.m
+++ b/components/formats-gpl/matlab/bfopen5D.m
@@ -1,0 +1,345 @@
+function [ResData,AllColorMaps,LabelList,FileInfo,OMEMetaData,seriesNumber] = bfopen5D(id, varargin)
+% Open microscopy images using Bio-Formats.
+% based on the bfopen function, but optimized for importing up to 5D data.
+%
+% SYNOPSIS [ResData,AllColorMaps,LabelList,FileInfo,OMEMetaData] = bfopen5D(id)
+%          [ResData,AllColorMaps,LabelList,FileInfo,OMEMetaData] = bfopen5D(id, x, y, w, h)
+%
+% Input
+%    r - the reader object (e.g. the output bfGetReader)
+%
+%    x - (Optional) A scalar giving the x-origin of the tile.
+%    Default: 1
+%
+%    y - (Optional) A scalar giving the y-origin of the tile.
+%    Default: 1
+%
+%    w - (Optional) A scalar giving the width of the tile.
+%    Set to the width of the plane by default.
+%
+%    h - (Optional) A scalar giving the height of the tile.
+%    Set to the height of the plane by default.
+%
+% Output
+%
+%    result - a cell array of cell arrays of (matrix, label) pairs,
+%    with each matrix representing a single image plane, and each inner
+%    list of matrices representing an image series.
+%
+% Portions of this code were adapted from:
+% http://www.mathworks.com/support/solutions/en/data/1-2WPAYR/
+%
+% This method is ~1.5x-2.5x slower than Bio-Formats's command line
+% showinf tool (MATLAB 7.0.4.365 R14 SP2 vs. java 1.6.0_20),
+% due to overhead from copying arrays.
+%
+% Thanks to all who offered suggestions and improvements:
+%     * Ville Rantanen
+%     * Brett Shoelson
+%     * Martin Offterdinger
+%     * Tony Collins
+%     * Cris Luengo
+%     * Arnon Lieber
+%     * Jimmy Fong
+%     * Rainer Heintzmann
+%
+% NB: Internet Explorer sometimes erroneously renames the Bio-Formats library
+%     to bioformats_package.zip. If this happens, rename it back to
+%     bioformats_package.jar.
+%
+% For many examples of how to use the bfopen function, please see:
+%     https://docs.openmicroscopy.org/latest/bio-formats/developers/matlab-dev.html
+
+% OME Bio-Formats package for reading and converting biological file formats.
+%
+% Copyright (C) 2007 - 2017 Open Microscopy Environment:
+%   - Board of Regents of the University of Wisconsin-Madison
+%   - Glencoe Software, Inc.
+%   - University of Dundee
+%
+% This program is free software: you can redistribute it and/or modify
+% it under the terms of the GNU General Public License as
+% published by the Free Software Foundation, either version 2 of the
+% License, or (at your option) any later version.
+%
+% This program is distributed in the hope that it will be useful,
+% but WITHOUT ANY WARRANTY; without even the implied warranty of
+% MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+% GNU General Public License for more details.
+%
+% You should have received a copy of the GNU General Public License along
+% with this program; if not, write to the Free Software Foundation, Inc.,
+% 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+% -- Configuration - customize this section to your liking --
+
+% Toggle the autoloadBioFormats flag to control automatic loading
+% of the Bio-Formats library using the javaaddpath command.
+%
+% For static loading, you can add the library to MATLAB's class path:
+%     1. Type "edit classpath.txt" at the MATLAB prompt.
+%     2. Go to the end of the file, and add the path to your JAR file
+%        (e.g., C:/Program Files/MATLAB/work/bioformats_package.jar).
+%     3. Save the file and restart MATLAB.
+%
+% There are advantages to using the static approach over javaaddpath:
+%     1. If you use bfopen within a loop, it saves on overhead
+%        to avoid calling the javaaddpath command repeatedly.
+%     2. Calling 'javaaddpath' may erase certain global parameters.
+% autoloadBioFormats = 1;
+
+ResData=[];AllColorMaps=[];FileInfo=[];OMEMetaData=[];LabelList=[];
+
+% Toggle the stitchFiles flag to control grouping of similarly
+% named files into a single dataset based on file numbering.
+stitchFiles = 0;
+seriesNumber=1;
+
+% To work with compressed Evotec Flex, fill in your LuraWave license code.
+%lurawaveLicense = 'xxxxxx-xxxxxxx';
+
+% -- Main function - no need to edit anything past this point --
+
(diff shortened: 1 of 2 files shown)


sbesson:
@RainerHeintzmann thanks for opening this contribution. Before we can review the code changes  in details, I encountered a few execution errors (see below) and had some high-level questions on the overall goal and the design of the new functionalities. For review and integration purposes, it will be probably be worth splitting this PR into a few separate PRs more focused on few changes but this can happen as a follow-up of this discussion.

- `bfOpen5D`
  1. can you expand  on ` bfopen seemed to not work correctly`, maybe with a more concrete example and/or a description of the issue you have been facing?
  2. can you describe the expected behavior of the `bfopen5d` utility and more precisely the differences with `bfopen`. As the code is largely a modification of the existing utility, it makes it hard to review specifically the divergences.
  3. running `bfopen5D('test.tiff')` on MATLAB R2018b leads to the following error

  ```
  Undefined function or variable 'equalsizes'.

  Error in bfopen5D (line 231)
        if ~equalsizes(size(ResData),TotalSize)
  ```

- `bfsave`:
  1. can you summarize the proposed changes in terms of functionalities? it reads like a lot of the changes are tailored towards a given use case writing OME-TIFF with some assumptions on the input data.
  2. running `bfsave(ones(100,100),'out.tiff')` on MATLAB R2018b leads to the following error

  ````
  Dot indexing is not supported for variables of this type.

Error in bfsave2 (line 107)
newid=[newid 'spacing='
num2str(double(metadata.getPixelsPhysicalSizeZ(0).value))];
  ````

- re copyright, that's an interesting question and I do not remember it being raised before. Thinking briefly about it, I agree conceptually that the MATLAB `bfopen`, `bfsave`... utility function should primarily expose BSD-2 licensed code like the Bio-Formats API.  A closer review of the codebase will probably be needed to confirm this. The `bfmatlab.zip` bundle obviously need to remain GPL-2 but would a lighweight bundle including only the M files be useful for your purpose?

RainerHeintzmann:
Dear Sébastian Besson,
Thanks for checking the code. I forgot to include the equalsizes funtuon which is in a different self written Toolbox. Will fix that by including the equal size function. Sorry. 
I tried some simple tests in the old bfopen and could not reproduce a problem. Maybe I somehow started with some old code. 

However what I needed is a tool that returns a multidimensional image rather than a one-d list of 2d images. I also wanted to avoid asking the user for specifying the order of dimensions. 

As for bfsave the changes were triggered by the observation that a tif file that I wrote via the Bioformats library, specifying all the dimensions in the metadata correctly ended up always with a 1um Z-pixelsize in Imaris. After some effort we discovered that this was because we called it “.tif” instead of “.ome.tif”. 
Admittedly the current idea of issuing a warning and just changing the name to “.ome.tif” is also not great and even a bit dangerous with respect to unwanted LG overwriting files. 
I do realise that the TIFF standard may have no default way of storing plane spacing, however the ImageJ default reader gets this from the ImageDescription Tag.
Probably the best way of dealing with this would be if Bioformats does write at least this Z-spacing information into the appropriate TIF tag. 
I do also realise that there may be political reasons not to do this, but in this case it would be good to force the user to leave this metadatatag empty by returning an appropriate error. 
My attempt was to write both tags even into ome.tiff files, since this would allow even the drag and drop functionality to work correctly in ImageJ. 
I did fix the indicated error by now but the problem remains since that part of my code seems to not create that tiff tag but just an InageDescription in the ome System at least when it is ome.tag. Any idea how to achieve the ImageDescription tif Tag via Bioformats?

Will send a now commit later today. 

Regarding the lincensing: Yes, a lightweight bundle under a less viral licence would be exactly what I need. Would this be achieved by moving these functions to the “-bsd” part of the directory ?
But “lightweight sounds even better.

RainerHeintzmann:
... just uploaded the changed version.

sbesson:
Hi Rainer,

> However what I needed is a tool that returns a multidimensional image rather than a one-d list of 2d images. I also wanted to avoid asking the user for specifying the order of dimensions.

That makes complete sense especially in the case where the output of `bfopen` is used directly for processing. Rather than copying most of the logic of `bfopen` into another function, would modifying `bfpen` to return a different output depending on a key/value pair be an option here? For instance, a signature like `bfopen(..., 'output', '{2D/5D}')` could return either return a nPlanes x 2 cell array of planes and labels or a 5D matrix in the first element of the output depending on the value of the parameter.

> I do realise that the TIFF standard may have no default way of storing plane spacing, however the ImageJ default reader gets this from the ImageDescription Tag.

You are right that the core TIFF specification has no concept of multiple Z-planes. This is something that both the ImageJ TIFF implementation and the OME-TIFF specification are attempting to solve by storing relevant metadata in the `ImageDescription` tag. 

When writing a regular TIFF, `TIFFWriter` should write the ImageJ metadata by default. For instance, the two following calls should create respectively a Z-stack and a 3 channel TIFF image that opens as expected in ImageJ/Fiji. Can you give us a minimal example of the scenario that you are trying to write?

```
>> bfsave(zeros(100,100,10,1,1,'uint8'), 'zstack.tif')
>> bfsave(zeros(100,100,1,3,1,'uint8'), '3channels.tif')
```

> Regarding the lincensing: Yes, a lightweight bundle under a less viral licence would be exactly what I need. Would this be achieved by moving these functions to the “-bsd” part of the directory ?

A starting point for the lightweight bundle might be the Bio-Formats Octave package already available from the [downloads page](https://www.openmicroscopy.org/bio-formats/downloads/) which only contains the `.m` files. With regard to licensing, using the current layout the source files would need to be migrated to the a BSD-2 component. Note however the relicensing needs to be discussed and agreed by all copyright holders. Would you like to open a dedicated [issue](https://github.com/ome/bioformats/issues/new) to initiate this discussion?

RainerHeintzmann:
Hi Sébastien,

Would modifying bfpen to return a different output depending on a key/value
> pair be an option here? For instance, a signature like bfopen(...,
> 'output', '{2D/5D}') could return either return a nPlanes x 2 cell array
> of planes and labels or a 5D matrix in the first element of the output
> depending on the value of the parameter.
>
Yes. This would also work and not break backwards compatibility. The only
disadvantage is that it would be more difficult for the user to discover
such hidden functionality, if it is not documented well.
So this should be prominently mentioned in the help file.

> I do realise that the TIFF standard may have no default way of storing
> plane spacing, however the ImageJ default reader gets this from the
> ImageDescription Tag.
>
> You are right that the core TIFF specification has no concept of multiple
> Z-planes. This is something that both the ImageJ TIFF implementation and
> the OME-TIFF specification are attempting to solve by storing relevant
> metadata in the ImageDescription tag.
>
> When writing a regular TIFF, TIFFWriter should write the ImageJ metadata
> by default. For instance, the two following calls should create
> respectively a Z-stack and a 3 channel TIFF image that opens as expected in
> ImageJ/Fiji. Can you give us a minimal example of the scenario that you are
> trying to write?
>
> >> bfsave(zeros(100,100,10,1,1,'uint8'), 'zstack.tif')
>
> >> bfsave(zeros(100,100,1,3,1,'uint8'), '3channels.tif')
>
> Of course this would work but does NOT contain the spacing information,
which what I was trying to explain. Even if you set the metadata correctly
it seems to not carry over the spacing between planes.


>
>  Regarding the lincensing: Yes, a lightweight bundle under a less viral
> licence would be exactly what I need. Would this be achieved by moving
> these functions to the “-bsd” part of the directory ?
>
> A starting point for the lightweight bundle might be the Bio-Formats
> Octave package already available from the downloads page
> <https://www.openmicroscopy.org/bio-formats/downloads/> which only
> contains the .m files. With regard to licensing, using the current layout
> the source files would need to be migrated to the a BSD-2 component. Note
> however the relicensing needs to be discussed and agreed by all copyright
> holders. Would you like to open a dedicated issue
> <https://github.com/ome/bioformats/issues/new> to initiate this
> discussion?
>
This may be useful. Are you referring to the following licence under BSD-2 ?
https://opensource.org/licenses/BSD-2-Clause

I uploaded a new version of bfOpen5D with some features about the time
series. The issue was that a series of stk files was always fully loaded,
even if the user may want to open only a single stk file.

Best wishes,
  Rainer

sbesson:
> Yes. This would also work and not break backwards compatibility. The only disadvantage is that it would be more difficult for the user to discover such hidden functionality, if it is not documented well. So this should be prominently mentioned in the help file.

Promoting this functionality both in the function help as well as https://docs.openmicroscopy.org/bio-formats/6.4.0/developers/matlab-dev.html#opening-an-image-file would make complete sense to me. Eventually this might become the default but that is a breaking change.
If a separate function remains valuable, adding the feature in `bfopen` and having a shallow `bfopen5D` function which would delegate to `bfopen` with the appropriate options set might be an option to solve the maintenance problem.

> which what I was trying to explain. Even if you set the metadata correctly
it seems to not carry over the spacing between planes.

The storage of the metadata is certainly dependent of whether the target format supports it. By spacing between planes, are you referring to the [physical pixel size in Z](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Pixels_PhysicalSizeZ) or the [plane position](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Plane_PositionZ) metadata?
Do you have an example of plain TIFF file which would store this information currently?

>This may be useful. Are you referring to the following licence under BSD-2 ?
https://opensource.org/licenses/BSD-2-Clause

Yes, almost all software maintained by OME is either licensed under BSD-2-Clause or GPL2+.

sbesson:
Coming back to this issue. Since the MATLAB functions have  be relicensed as BSD-2 in Bio-Formats 6.7.0 (thanks again for driving this!), the files modified/added in this PR are in the wrong component. Closing this PR but feel free to reopen a new PR with the changes against the files under `components/formats-bsd`.

If both proposed features are still in scope, I suggest to open two separate Pull Requests, one for each function, to ease the review and allow features to be released more rapidly.

PR #3526, opened 2020-03-13
JPEG service: fix off-by-one when using restart markers to build a tile's stream

### LINKED ISSUE #3059: NDPIReader & makeRGBImage last scanlines wrong
When using NDPIReader and converting the data to an image using a BufferdImageReader - or makeRGBImage directly - on some images (probably depending on the height) the last few pixel rows (always 8?) have value 128 instead of the correct pixel value.

The problem can be reproduced on the example image
http://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/manuel/test3-FITC%202%20(485).ndpi

Code:

```
import loci.formats.FormatReader;
import loci.formats.gui.AWTImageTools;
import loci.formats.in.NDPIReader;
import java.awt.image.BufferedImage;

public class TestNDPIReader {
    public static void main(String[] args) {
        // image from http://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/manuel/test3-FITC%202%20(485).ndpi
        String filename = "test3-FITC 2 (485).ndpi";
        try (FormatReader r = new NDPIReader()) {
            r.setId(filename);
            byte[] bytes = r.openBytes(0);
            int last = bytes[bytes.length-1] & 0xf;   // here the last byte (blue value) is 0, which is correct
            System.out.println(last);    // 0
            // openImage results in value 128 - which is wrong. Trace down to makeRGBImage...
           // BufferedImage bi = AWTImageTools.openImage(bytes,r,r.getSizeX(),r.getSizeY());
           // BufferedImage bi = AWTImageTools.makeImage(bytes, r.getSizeX(), r.getSizeY(), 3, true, 1, false, true, false);
            BufferedImage bi = AWTImageTools.makeRGBImage(bytes, 3, r.getSizeX(), r.getSizeY(), true);
            int lastBi = bi.getRaster().getSample(bi.getWidth()-1,bi.getHeight()-1,0);
            System.out.println(lastBi);  // 128 but should be 0
            // this problem holds for the last 8 scanline pixels (samples width*(height-1..8)*3), before it's correct
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```
Maybe this is related to [#875](https://github.com/openmicroscopy/bioformats/issues/875) and/or [#836](https://github.com/openmicroscopy/bioformats/pull/836)

Result of the image using a BufferedImageReader (you can see the last part (8 lines) is gray)
testndpi
Zoomin:
testndpi-zoom

melissalinkert:
Fixes #3059

Without this fix, the bytes between the last restart marker and the end of the original JPEG
stream were ignored.

Without this PR, ```showinf hamamatsu/public/manuel/test3-FITC 2 (485).ndpi``` will show a horizontal gray stripe at the bottom of the image approximately 8 pixels high.  With this PR, the same test should show no stripe.

I'd expect this to be safe for a patch release, but some configuration updates may be necessary depending upon the results of tonight's repo test.

### DIFF
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
@@ -291,13 +291,17 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
     for (int row=0; row<restartInterval; row++) {
       int end = start + 1;
 
+      long startOffset = restartMarkers.get(start);
+      long endOffset = in.length();
       if (end < restartMarkers.size()) {
-        long startOffset = restartMarkers.get(start);
-        long endOffset = restartMarkers.get(end);
-
-        dataLength += (endOffset - startOffset);
+        endOffset = restartMarkers.get(end);
       }
+
+      dataLength += (endOffset - startOffset);
       start += xTiles;
+      if (start >= restartMarkers.size()) {
+        break;
+      }
     }
 
     byte[] data = new byte[(int) dataLength];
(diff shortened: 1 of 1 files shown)


sbesson:
As expected from the description, this change breaks the assumptions of the non-regression tests against the `hamamatsu` sample files - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/52408/.

Discussing with @melissalinkert, proposed next step is to import all the NDPI sample files into test servers with and without this PR. Once the impact of this PR is validated, we can schedule the regeneration of the configuration files.

sbesson:
Visually comparing the impact of this changes on a few files. Without this PR, the last lines display at green when zooming to the maximum resolution

Screen Shot 2020-03-18 at 17 26 29

With this PR included, the last lines renders as expected

Screen Shot 2020-03-18 at 17 26 18

mstritt:
I just want to thank you for your efforts, it's really great work you're doing!
Manuel

PR #3529, opened 2020-03-16
Use skipBytes(long) in places where integer overflow is possible

melissalinkert:
Inspired by https://github.com/ome/bioformats/pull/3519, https://github.com/ome/bioformats/pull/3303, and https://github.com/ome/bioformats/pull/3157.  This uses ```skipBytes(long)``` in more places where integer overflow seems plausible.

I would expect this to be safe for a patch release and have no impact on tests.  I don't anticipate this will fix any outstanding issues, it's just an attempt at being more defensive.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LiFlimReader.java
@@ -231,7 +231,7 @@ public class LiFlimReader extends FormatReader {
       int thisSeries = getSeries();
       for (int i=0; i<thisSeries; i++) {
         setSeries(i);
-        in.skipBytes(getImageCount() * FormatTools.getPlaneSize(this));
+        in.skipBytes((long) getImageCount() * FormatTools.getPlaneSize(this));
       }
       setSeries(thisSeries);
       readPlane(in, x, y, w, h, buf);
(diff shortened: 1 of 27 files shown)


components/formats-gpl/src/loci/formats/in/LiFlimReader.java
```diff
@@ -231,7 +231,7 @@ else if (TypeUINT12) {
       int thisSeries = getSeries();
       for (int i=0; i<thisSeries; i++) {
         setSeries(i);
-        in.skipBytes(getImageCount() * FormatTools.getPlaneSize(this));
+        in.skipBytes((long) getImageCount() * FormatTools.getPlaneSize(this));
```
sbesson: This seems to have impacted the non-regression tests at https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/52532/console. We might need to look at the data with or without this PR to establish whether the configuration needs to be regenerated or there is an issue with this change.

melissalinkert: Reasonably certain that the configuration needs to be updated, as proposed here: https://github.com/ome/bioformats/pull/3529#pullrequestreview-375846632

Note that without that change, the MD5s for series 0 and series 1 were the same, which points to this PR fixing a problem.

PR #3531, opened 2020-03-18
Add standard GPL-2 license to MikroscanTiffReader

sbesson:
The reader was added in #3333 but the correct license was not included back then. This updates the file header using the standard assumptions in terms of license and copyright owner holders.


melissalinkert:
Makes sense to me.  No objections to merging.

PR #3532, opened 2020-03-20
Fix test-config on datasets with planes larger than 2 GB

melissalinkert:
The plane read time configuration didn't correctly check that the plane size is smaller than 2 GB in addition to fitting into available memory.  This resulted in ```IllegalArgumentException```s when setting ```-Dtestng.memory``` higher than the default value.

The problem was observed when running ```ant -Dtestng.memory=2g ... test-config``` to regenerate configurations for .ndpi files.  The change in this PR was used to create https://github.com/openmicroscopy/data_repo_config/pull/447.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -563,12 +563,8 @@ public class Configuration {
     globalTable.put(TEST, "true");
     globalTable.put(MEMORY, String.valueOf(TestTools.getUsedMemory()));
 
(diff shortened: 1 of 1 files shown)


sbesson:
@melissalinkert tested on a few random Hamamatsu sample files with `testng-memory=4g` but I could not reproduce the initial issue without this PR. Do you remember which sample file would be a good candidate for reproducing? Otherwise the code changes look fine.

melissalinkert:
```ant -Dtestng.memory=4g -Dtestng.directory=/path/to/curated/hamamatsu/david -Dtestng.configDirectory=/home/melissa/ test-config``` shows the problem locally.

PR #3535, opened 2020-03-30
Correct pixel size for "Olympus - TIFF Images (Stream 1.9.2)"

swg08:
We came across a similar issue with Olympus TIFF images as in here: https://forum.image.sc/t/spatial-calibration-in-tiff-not-recognized-when-picture-captured-using-olympus-stream-software/22418.

The image reports a pixel size of 352,27microns which is wrong for the image, but perfectly fine with respect to the 72 DPI / inch found in the TIFF tags. 
So, I tried to understand the topic a bit better and debugged the SisReader.java.

When I checked with https://gist.github.com/tbenst/7db5a9aa9b221e67ba85c5642b23f094 I found that it can read the values correctly (I think I read that in image.sc issue above).

The PR is more a discussion/draft PR - as I have not deleted all non-necessary Code and there is at least one place where a Resolution/discussion is needed.


dgault:
Thanks for opening the PR and investigating the issue. For the first step I will include the PR in the nightly tests to see how it impacts the existing sample files we have for the format. Unfortunately we have a limited set of files to test but it will at least provide some idea if there are potentially files without the units present. 

If you have any additional files we can test with please do upload them to https://www.openmicroscopy.org/qa2/

swg08:
Dear David,

I had Trouble with Adobe on the upload page below and am not allowed to
install it on my Laptop.
So, I tried uploading it to the Glencoe Zeiss FTP upload Directory.
I'm not using FileZilla yet and am not sure if that upload worked. Could
you give me sign if the file is there now. I'm not in the Office today and
will check back tomorrow.

Best regards,
Stephan

dgault:
Thanks Stephen, we received the file upload OK. As expected the overnight tests resulted in some differing values of physical sizes with some of the existing datasets. I will try and review each of these this week to determined if the newly updated values are indeed correct.

swg08:
Hi David, good that the file made it to you. As mentioned in the PR I would
expect that the lines with diving by 1000 Need to adapted to a concrete
Pixel size estimation - meaning that we might need to identify from which
TIFF Tag a certain pixel size came from. The easy cases might be where the
Pixel size is now 354 microns which seems to be related to the dpi value in
the General TIFF tags.
Best regards,
Stephan

swg08:
Hi David, can I help finishing this Topic off?

dgault:
Sorry for the delay in getting this merged. The reason its taking so long is because it has updated the pixel sizes for a large number of files in our data repo and I need to review each to ensure the new values are correct and then update the configuration files to reflect the changes. I had hoped to try and have all of the impacted repo files reviewed for the 6.5.1 release but did not get through all of them in time. I will try and finish the remaining review this week and then get a configuration PR open so we can get this scheduled for a release.

dgault:
Repo tests are now green with the updated configuration. Each of the updated values has been verified and this looks good to merge.

swg08:
Hi David,
That's great!
That sounded like a lot of Checks. I guess there are many Olympus TIFF
files collected over the course of years.
Thank you for Your efforts!
Best regards,
Stephan

David Gault <notifications@github.com> schrieb am Do., 23. Juli 2020, 10:18:

PR #3540, opened 2020-04-05
CellWorX: add support for single site HCS variant

sbesson:
The number of well samples/fields of view per well is computed from the total number of acquisition sites and the number of selected sites, determined by XSites, YSites and the SiteSelection<index> entries in the HTD file.

For single site acquisitions (XSites=1, YSites=1), the SiteSelection1 key might be absent from the HTD files. This commit updates the reader to assume the single site is selected.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
@@ -217,7 +217,7 @@ public class CellWorxReader extends FormatReader {
     }
 
     int planeIndex = no;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Propose to add one more commit here that assumes a single site any time ```Sites``` is set to ```FALSE```: https://github.com/sbesson/bioformats/pull/6

dgault:
Tested using the newly configured data from `idr0081`. With this change the file was able to be read without any exceptions and the data looked to be correct. No repo test failures or regressions have been seen with this PR included either. Looks good to merge.

PR #3542, opened 2020-04-08
Metamorph: don't allow underlying TIFF reader to treat each IFD as a series

melissalinkert:
Instead, trust dimension from tag metadata and/or .nd file.  Backported from a private PR.

I wouldn't expect this to have any impact on tests.  I will put together a test file in the next day or two once it's clear that this does not break anything.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -171,6 +171,7 @@ public class MetamorphReader extends BaseTiffReader {
     suffixSufficient = false;
     datasetDescription = "One or more .stk or .tif/.tiff files plus an " +
       "optional .nd or .scan file";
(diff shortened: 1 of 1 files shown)


melissalinkert:
See test dataset configured in https://github.com/openmicroscopy/data_repo_config/pull/455.  Without this change, ```showinf test.nd``` should throw an exception; with this change, the same test should display images with the image count matching the contents of ```test.nd```.

dgault:
Tested using the new sample file. Without this PR it results in an exception:

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at loci.formats.in.MetamorphReader.getComment(MetamorphReader.java:1549)
```

With the PR the file is read without exception and the displayed image and metadata looks correct. 

Builds and tests have been green with no regressions

PR #3543, opened 2020-04-08
Deltavision: sanity check the panel count and fall back to old style header

melissalinkert:
Backported from a private PR.

Some files have the file type value set to indicate a "new" header format,
but actually have an "old" header format.  This causes the panel/series count
to be wildly inaccurate, which in extreme cases can trigger an ```OutOfMemoryError```.
A basic check on the number of detected panels allows the reader to fall
back to parsing an "old" header instead.

An artificial test file created from one of the public samples is in ```inbox/dv-bad-header``` (see the readme in that directory).  Without this PR, ```showinf``` on the test file should eventually throw an ```OutOfMemoryError```.  With this PR, images should be displayed without error and the output should match the original file from which the test was created.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -351,12 +351,23 @@ public class DeltavisionReader extends FormatReader {
     if (fileType == NEW_TYPE) {
       in.seek(852);
       int secondaryT = in.readInt();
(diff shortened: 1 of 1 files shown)


dgault:
Tested using the new sample file. Without this PR it results in an OutOfMemoryError, with this PR included the file is read without exception.

Builds and tests have green and are not showing any regressions in existing datasets, looks good to merge.

Issue #3541, opened 2020-04-07
Conversion of OME Creator attribute


sbesson:
The  [6.1.0 release of ome-model](https://github.com/ome/ome-model/blob/v6.1.0/NEWS.md) includes proper support for the `OME/Creator` attribute in the `ome.xml.meta` interfaces and implementations.

Inline with the [OME schema](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#OME_Creator), the OME-XML and OME-TIFF writers unconditionally populate this attribute with the version of Bio-Formats.

https://github.com/ome/bioformats/blob/b1be2a9baa7b780ca258315d80132479099c57de/components/formats-bsd/src/loci/formats/out/OMETiffWriter.java#L297-L299

https://github.com/ome/bioformats/blob/b1be2a9baa7b780ca258315d80132479099c57de/components/formats-bsd/src/loci/formats/out/OMETiffWriter.java#L320-L322

https://github.com/ome/bioformats/blob/b1be2a9baa7b780ca258315d80132479099c57de/components/formats-bsd/src/loci/formats/out/OMEXMLWriter.java#L106-L108

With the new API, a conversion workflow can read and convert the value of the `Creator` attribute. ON the other hand, capturing the version of the software which created the converted OME metadata is still valuable.

Listing a few options up for discussion:

1. give the user the option to either preserve or override the Creator attribute
2. always override the Creator attribute but store the previous Creator value as an annotation

dgault:
Fixed in https://github.com/ome/bioformats/pull/3959

PR #3545, opened 2020-04-13
Bump ome-model version to 6.1.0

sbesson:
See https://github.com/ome/ome-model/blob/v6.1.0/NEWS.md for the list of changes

The latest version of ome-model has been pulled in the daily merge CI builds so I expect tests will keep passing.

Bumping the  component version in the pom.xml is necessary to consume the new `OME Creator` API - see #3541.


dgault:
Builds and tests have all been passing and Ive verified the OME@Creator changes. Good for inclusion with 6.5.0 release

PR #3546, opened 2020-04-13
Zeiss CZI: fix position count when the starting index is > 0

melissalinkert:
Backported from a private PR.

This is meant to fix a case where only one position is present, but the starting index is 10.  The file contains a pyramid, so an incorrect position count completely throws off pyramid building logic.

If this doesn't fail tests, additional commits could be added (here or in a separate PR) to calculate the illumination, phase, etc. dimensions similarly.  I'll find or put together a test file once it's clear that this doesn't break any existing files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1694,6 +1694,9 @@ public class ZeissCZIReader extends FormatReader {
 
     ArrayList<Integer> uniqueT = new ArrayList<Integer>();
 
+    int minPositions = Integer.MAX_VALUE;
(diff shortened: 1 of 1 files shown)


sbesson:
This seems to have caused regressions on other CZI filesets - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/57376/console. Exclusing for now. Feel free to unexclude when the issues are resolved.

melissalinkert:
2b5922d fixed the failing tests locally, so removing exclude label for one more try.

dgault:
The updated calculations look fine and repo tests are green with this included. Manually tested with the new sample file from https://github.com/openmicroscopy/data_repo_config/pull/457 and the file was read without issue and the metadata looked to be correct.

PR #3222, opened 2018-09-12
DICOM file grouping fixes

melissalinkert:
Backports 3 private PRs.  This improves file grouping and file-to-series mapping for multi-file datasets with and without a ```DICOMDIR``` file.

To test, use the following new datasets:

1. ```data_repo/curated/dicom/melissa/two-files-single-series``` (2 files)
2. ```data_repo/curated/dicom/melissa/two-files-two-series``` (2 files)
3. ```data_repo/curated/dicom/public/nema/Liver/DICOMDIR``` (```DICOMDIR``` + 183 files)

Datasets 1 and 2 are artificial datasets constructed as noted in the respective readmes.  Dataset 3 is directly copied from ftp://medical.nema.org/medical/dicom/DataSets/WG16/Philips/ClassicSingleFrame/Liver/

Without this PR, ```showinf -nopix 0.dcm``` for both datasets 1 and 2 should show 2 series and 3 used files (both datasets are grouped together, with one file omitted).  The first series has 1 Z section, and the second series has 2 Z sections.

Without this PR, ```showinf -nopix DICOMDIR``` for dataset 3 will show 80 series, each with a small but variable number of Z sections.

With this PR, ```showinf -nopix 0.dcm``` for dataset 1 should show 1 series with 2 Z sections and two used files (```two-files-single-series/0.dcm``` and ```two-files-single-series/1.dcm```).

```showinf -nopix 0.dcm``` for dataset 2 should show 2 series each with 1 Z section and varying SizeY.  There should be two used files (```two-files-two-series/0.dcm``` and ```two-files-two-series/1.dcm```).

```showinf -nopix DICOMDIR``` for dataset 3 should show 5 series, each with 18-50 Z sections.  The total number of planes across all series should be 173, which matches the number of (single plane) ```IM_*``` files.

Viewing datasets 1 and 2 in ImageJ should result in all images matching ```data_repo/curated/dicom/public/samples/CT-MONO2-16-chest.dcm``` (this is the base file from which the datasets were constructed).  Viewing dataset 3 in ImageJ should result in 5 clearly distinct Z stacks with no planes obviously out of order.

All builds should pass; there should be no failures or configuration changes for existing DICOM datasets.  I would expect memo files to be affected.

PR #3548, opened 2020-04-20
DICOM: do not check other directories for matching files

melissalinkert:
Files spread across multiple directories are now handled by a ```DICOMDIR``` file that groups everything together.  See previous work in #2785 and #3222.

Main thing to check is that this has no impact on existing tests.  I would expect this to speed up initialization of multi-file datasets, especially in a directory structure similar to:

```
/path/to/dicom/set/1/
/path/to/dicom/set/2/
```

with each fileset containing hundreds or thousands of files.

Not urgent for 6.5.0; assuming tests pass, should be safe for a later patch release.


dgault:
PR hasn't caused any regressions which is the main thing, all builds and tests have remained green.
Initialization should be sped up for some filesets, testing locally with some data I have didn't show any issues or unintended side effects. Looks good to merge.

PR #3518, opened 2020-03-06
ImageConverter: MultiSeries fix

dgault:
This is a fix for https://github.com/ome/bioformats/issues/3512

To test:
- Use bfconvert on a multi series file
- Select a single timepoint, z slice or channel
- Without this PR opening the converted file will throw an ArithmeticException: / by zero
- With this PR the converted file should open without exception

Closes #3512

PR #3553, opened 2020-05-04
Image converter: fix multi-series regression

sbesson:
Reported in https://forum.image.sc/t/bfconvert-throwing-an-error-for-an-vsi-file/37185, this fixes a regression introduced in #3518 for multi-series files with varying image sizes

`bfconvert "test&resolutions=2.fake" out.ome.tiff` should suffice to reproduce the issue reported in the forum thread. 205dae6  fixes the issue by only setting the pixel X and Y sizes to the first image. 5561dbb adds a simple test to `ImageConverterTest` which should fail without the fix and pass with it.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -578,10 +578,9 @@ public final class ImageConverter {
           writer.setMetadataRetrieve((MetadataRetrieve) meta);
         }
         else {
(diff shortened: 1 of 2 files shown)


dgault:
Retested using bfconvert and multiple resolutions. Without this PR an exception is thrown, as reported on the forum thread:
`FormatException: Buffer is too small; expected 262144 bytes, got 65536 bytes.`

With this PR included no exception is thrown and the image is converted successfully. 

The updated unit tests will help to catch any future regressions and all other tests are currently passing with this PR included.

PR #3554, opened 2020-05-05
Add Zarr Reader and Writer

dgault:
Initial Zarr reader and writer for testing, this is an ongoing work in progress but can be used for testing purposes.

In its current form the writer expects a .zarr folder and will place the ome.xml within that location.
The reader will accept any file so long as its path has .zarr in it somewhere.

Notes:
- Trying to overwrite a file will not work due to a folder being used as the ID
- Dimension order of the Zarr is currently fixed to XYZTC
- Chunking is currently in X and Y dimensions only, when converting a file the readers tile size will be used for chunk size
- Javadoc needs added/finalised

Things to agreed:
- Location of XML
- File to select for SetID

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZarrReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZarrReader.java
@@ -1,0 +1,242 @@
+package loci.formats.in;
+
+import java.io.IOException;
+
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+import loci.common.DataTools;
+import loci.common.Location;
+import loci.common.RandomAccessInputStream;
+import loci.common.services.DependencyException;
+import loci.common.services.ServiceException;
+import loci.common.services.ServiceFactory;
+import loci.common.xml.XMLTools;
+import loci.formats.CoreMetadata;
+import loci.formats.FormatException;
+import loci.formats.FormatReader;
+import loci.formats.FormatTools;
+import loci.formats.MetadataTools;
+import loci.formats.MissingLibraryException;
+import loci.formats.meta.MetadataStore;
+import loci.formats.ome.OMEXMLMetadata;
+import loci.formats.services.ZarrServiceImpl;
+import ome.xml.meta.MetadataConverter;
+import loci.formats.services.OMEXMLService;
+import loci.formats.services.ZarrService;
+
+
+public class ZarrReader extends FormatReader  {
+
+  private transient ZarrService zarrService;
+  
+  public ZarrReader() {
+    super("Zarr", "zarr");
+    suffixSufficient = true;
+    domains = new String[] {FormatTools.UNKNOWN_DOMAIN};
+  }
+  
+  /* @see loci.formats.IFormatReader#isThisType(String, boolean) */
+  @Override
+  public boolean isThisType(String name, boolean open) {
+    Location zarrFolder = new Location(name).getParentFile();
+    if (zarrFolder != null && zarrFolder.exists() && zarrFolder.getAbsolutePath().indexOf(".zarr") > 0) {
+      return true;
+    }
+    return super.isThisType(name, open);
+  }
+
+  /* @see loci.formats.IFormatReader#getOptimalTileHeight() */
+  @Override
+  public int getOptimalTileHeight() {
+    FormatTools.assertId(currentId, true, 1);
+    return zarrService.getChunkSize()[1];
+  }
+  
+  /* @see loci.formats.IFormatReader#getOptimalTileHeight() */
+  @Override
+  public int getOptimalTileWidth() {
+    FormatTools.assertId(currentId, true, 1);
+    return zarrService.getChunkSize()[0];
+  }
+  
+  /* @see loci.formats.FormatReader#initFile(String) */
+  @Override
+  protected void initFile(String id) throws FormatException, IOException {
+    super.initFile(id);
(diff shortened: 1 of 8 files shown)


components/formats-gpl/src/loci/formats/out/ZarrWriter.java
```diff
@@ -0,0 +1,208 @@
...
+      xml = service.getOMEXML(omeMeta);
+    }
+    catch (ServiceException se) {
+      throw new FormatException(se);
+    }
+    return xml;
+  }
+    
+  private String getUUID(String filename) {
+    String uuid = UUID.randomUUID().toString();
```
jburel, line 192: could be simplified ``return UUID...``

components/formats-gpl/src/loci/formats/out/ZarrWriter.java
```diff
@@ -0,0 +1,208 @@
+package loci.formats.out;
```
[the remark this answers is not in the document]
sbesson: Before headers get added, proposing to put these new readers/writers under BSD-2 license.

components/formats-gpl/src/loci/formats/out/ZarrWriter.java
```diff
@@ -0,0 +1,209 @@
...
+      else if (bpp==2) {
+        ByteBuffer bb = ByteBuffer.wrap(buf);
+        ShortBuffer sb = bb.asShortBuffer();
+        short[] image = new short[w * h];
+        for (int x_i=0; x_i < w; x_i++) {
+            for (int y_i=0; y_i < h; y_i++) {
+                short value = sb.get(y_i*w + x_i);
+                image[y_i*w + x_i] = value;
+            }
+        }
```
melissalinkert, line 84: Since the source and destination indexes are the same, is it necessary to iterate over each pixel or could this be replaced by ```sb.get(image)```?

components/formats-gpl/src/loci/formats/out/ZarrWriter.java
```diff
@@ -0,0 +1,209 @@
...
+        for (int x_i=0; x_i < w; x_i++) {
+            for (int y_i=0; y_i < h; y_i++) {
+                int value = sb.get(y_i*w + x_i);
+                image[y_i*w + x_i] = value;
+            }
+        }
+        zarrService.saveBytes(image, shape, offsets);
+      }
+      else {
+        throw new FormatException("CellH5Writer: Pixel type not supported");
```
melissalinkert, line 100: s/CellH5/Zarr

components/formats-gpl/src/loci/formats/in/ZarrReader.java
```diff
@@ -0,0 +1,243 @@
...
+      Integer w = omexmlMeta.getPixelsSizeX(i).getValue();
+      Integer h = omexmlMeta.getPixelsSizeY(i).getValue();
+      Integer t = omexmlMeta.getPixelsSizeT(i).getValue();
+      Integer z = omexmlMeta.getPixelsSizeZ(i).getValue();
+      Integer c = omexmlMeta.getPixelsSizeC(i).getValue();
+      if (w == null || h == null || t == null || z == null | c == null) {
+        throw new FormatException("Image dimensions not found");
+      }
+
+      Boolean endian = null;
```
melissalinkert, line 137: Is the value of ```BigEndian``` in the OME-XML ignored intentionally?

jburel:
It was discussed during a "io/zarr" call that we wanted to reduce the number of dependencies used to connect to S3.
``omero-ms-zarr`` already uses ``software.amazon.awssdk`` to connect to S3
This PR depends on ``jzarr`` will use ``com.amazonaws``. (via https://github.com/lasersonlab/Amazon-S3-FileSystem-NIO2)
This is a non-official repo. 
This library should probably be excluded and replaced by the one used in ``omero-ms-zarr``
if the connection to S3 is required from B-F.

joshmoore:
A few thoughts and testing observations:
 - I'd vote for having the output be at least one zgroup containing resolutions as in [spec.md](https://github.com/ome/omero-ms-zarr/blob/master/spec.md). `bioformats2raw` goes further and has one more zgroup on top for multiple series. That may be a complication we don't want to deal with at the moment.
- Which probably brings the specter of a version number. This should at least go in the zattrs of the highest level (whatever that is)
- Whichever zgroup represents the set of resolutions could also re-use [this code](https://github.com/glencoesoftware/bioformats2raw/blob/master/src/main/java/com/glencoesoftware/bioformats2raw/Converter.java#L955) for adding metadata about the resolutions.
- `showinf` on the .zarr directory itself shows `java.io.FileNotFoundException: foo.zarr (Is a directory)`
- and `bfconvert a.fake foo.zarr && showinf foo.zarr` fails with:
```
	at loci.formats.in.ZarrReader.getOptimalTileWidth(ZarrReader.java:63)
```
All that being said, I assume the highest priority issue is having the OMERO build passing (i.e. jars & repositories). At that point, we could take any of the resolutions out of an existing dataset and import it.

ngladitz:
@dgault I was somewhat excited to see work on this happening. Any reason this got so quietly closed?

joshmoore:
Hi @ngladitz. No worries. We'll get back to it. And it will certainly be a topic of discussion at https://www.openmicroscopy.org/events/ome-community-meeting-2020/ -- one of the primary issues was library incompatibilities between Bio-Formats and OMERO causing daily builds to break. With the upcoming meeting, there was just enough going on. But we'd definitely love feedback.

ngladitz:
@joshmoore alright looking forward to it, thanks!

PR #3555, opened 2020-05-08
Bump jxrlib version to 0.2.4

chris-allan:
Should be functionally identical but resolves duplicate class leakages
outlined here:

  https://forum.image.sc/t/how-to-resolve-duplicate-classes-in-unshaded-ome-jxrlib-all-artifact/37290


sbesson:
As mentioned in https://forum.image.sc/t/how-to-resolve-duplicate-classes-in-unshaded-ome-jxrlib-all-artifact/37290/6, some additional shaded content has been reported. @callan just created on a `0.2.4` release to address it. Assuming this will get pushed to this branch so that we can review everything on one go.

mtbc:
Maybe reword commit message?

PR #3556, opened 2020-05-08
Add OBF file format version and stack version to the global metadata

sbesson:
Follow-up of https://github.com/ome/bioformats/pull/3528, this change adds the file version and stack version (see https://imspectordocs.readthedocs.io/en/latest/fileformat.html for more details) to the global metadata populated by the reader.

/cc @ngladitz


dgault:
New additions look good. This wont have any impact on existing filesets and is safe to include in the next patch release.

PR #3557, opened 2020-05-11
Update kryo to latest version

ctrueden:
From 2.24.0 to 3.0.0, the kryo project changed its groupId from `com.esotericsoftware.kryo` to `com.esotericsoftware`. Therefore, from Maven's perspective, both of these artifacts can be present on the classpath at the same time, creating duplicate class conflicts. Unfortunately, this scenario is happening when attempting to combine Bio-Formats (which wants kryo 2.24.0) with `graphics.scenery:scenery:0.7.0-beta-7` (which wants kryo 4.0.2).

One solution to this dilemma is to update the version of kryo here to 4.0.2, which is what this PR does. The code still compiles, and tests pass, although I am not certain how thorough (if at all) the kryo-related functionality is exercised.

See also ome/ome-common-java#50.

### DIFF
--- a/components/formats-bsd/pom.xml
+++ b/components/formats-bsd/pom.xml
@@ -69,7 +69,7 @@
       <version>${jgoodies-forms.version}</version>
     </dependency>
     <dependency>
-     <groupId>com.esotericsoftware.kryo</groupId>
+     <groupId>com.esotericsoftware</groupId>
      <artifactId>kryo</artifactId>
      <version>${kryo.version}</version>
     </dependency>
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -126,7 +126,7 @@
       <version>2.2</version>
     </dependency>
     <dependency>
-      <groupId>com.esotericsoftware.kryo</groupId>
+      <groupId>com.esotericsoftware</groupId>
       <artifactId>kryo</artifactId>
       <version>${kryo.version}</version>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
     <ome-metakit.version>5.3.2</ome-metakit.version>
     <metakit.version>${ome-metakit.version}</metakit.version>
     <slf4j.version>1.7.6</slf4j.version>
-    <kryo.version>2.24.0</kryo.version>
+    <kryo.version>4.0.2</kryo.version>
     <testng.version>6.8</testng.version>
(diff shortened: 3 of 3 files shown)


sbesson:
Thanks for opening this companion PR @ctrueden. Summarizing some initial feedback from its inclusion in the daily CI builds. Overall, things are looking quite promising:

- the multi-component build of Bio-Formats compiled and all unit tests were passing
- no regression was detected against the curated QA repository especially the memoization tests which are part of the test-suite
- an OMERO server was successfully built and deployed including this PR and the OMERO integration tests were still passing passing
- a preliminary round of manual testing has been performed using images imported into OMERO across a combination of imaging modalities and file formats. From the server logs, the existing memo files were invalidated as expected, successfully regenerated and subsequently reloaded when accessing the data.

As the OME team is switching its focus to the preparation of the upcoming community meeting, these PRs will remain opened and included in the daily CI builds so that it can indirectly benefit from the testing of other downstream components.

Post community meeting, we will discuss a more thorough round of testing which should hopefully lead to the inclusion of this dependency bump in an upcoming minor release of Bio-Formats.

sbesson:
In addition to https://github.com/ome/bioformats/pull/3557#issuecomment-627521978, I tested specifically the cache generation, reloading and invalidation workflow using the last stable release of Bio-Formats 5.6.1 as well as a build with this PR included (`showinf -nopix -debug -cache`). Using a sample OME-TIFF file, I confirmed:

- a memo file generated by Bio-Formats 6.5.1 is successfully reloaded by Bio-Formats 6.5.1 but invalidated by Bio-Formats with this PR included and a new cache file is regenerated
- a memo file generated by Bio-Formats with this PR included is successfully reloaded by the same Bio-Formats vuild but invalidated by Bio-Formats 6.5.1 and a new cache file regenerated

Overall, the generation/reloading logic has been successfully tested over the last 2 months. Given the absence of issues, as discussed on Monday, my proposal is to get this merged early in the 6.6.0 roadmap so that we can keep testing it. If needed, a milestone build can be created in order to test the regeneration workflow.

The only minor addition I would propose after merging this PR would be to the bump the internal memo version value https://github.com/ome/bioformats/blob/v6.5.1/components/formats-bsd/src/loci/formats/Memoizer.java#L336 as this should invalidate the memo file with a more graceful error message.

dgault:
Agreed on bumping the internal Memoizer version, otherwise it looks good for 6.6.0

PR #3558, opened 2020-05-12
Gatan DM4: fix support for montages with single Z dimensions

sbesson:
See https://forum.image.sc/t/problem-reading-some-dm4-files-in-bioformats-6-4-0/36590

This commit tries to address issues with variants of the DM4 file formats which likely were not available when support for montage was added in #3337.

This should be tested using the new sample files uploaded in the image.sc thread as well as checking for regressions using the other DM4 montage sample files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -85,6 +85,9 @@ public class GatanReader extends FormatReader {
   private static final int ELLIPSE = 6;
   private static final int TEXT = 13;
 
+  public static final String SPLIT_MONTAGE = "gatan.split_montage";
+  public static final boolean SPLIT_MONTAGE_DEFAULT = true;
+
   // -- Fields --
 
   /** Offset to pixel data. */
@@ -180,6 +183,14 @@ public class GatanReader extends FormatReader {
 
   /* @see loci.formats.FormatReader#initFile(String) */
   @Override
+  protected ArrayList<String> getAvailableOptions() {
+    ArrayList<String> optionsList = super.getAvailableOptions();
+    optionsList.add(SPLIT_MONTAGE);
+    return optionsList;
+  }
+
+  /* @see loci.formats.FormatReader#initFile(String) */
+  @Override
   protected void initFile(String id) throws FormatException, IOException {
     super.initFile(id);
     in = new RandomAccessInputStream(id);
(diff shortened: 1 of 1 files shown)


sbesson:
Unfortunately, the simple fix proposed above does not resolve the issue. Although the original `setId` exception is  fixed, some uploaded samples throw an exception during the `openBytes` call for some series. This can be confirmed visually by inspecting the sample files loaded in a test OMERO server:

Screen Shot 2020-05-18 at 09 24 37

The primary question is whether the assumption introduced #3337 i.e.  dm4 files containing a montage tag as well as a stage position tag should be detected as multi-series is still valid and if not how to refine the logic.

melissalinkert:
Best guess looking at the tag structure in ```showinf -nopix -debug SmallMontage0001.dm4``` is that the ```Images Stored to Disk``` value under ```Montage > Acquisition``` needs to take precedence over the number of recorded stage positions.

sbesson:
Thanks @melissalinkert, this matches a quick investigation of the dm4 sample files containing a Montage tag. For reference, half of the uploaded sample files as part of the image.sc thread contain `Images Stored To Disk: 0.0` while the other half contain `Images Stored To Disk: 1.0`. I will look into adding an additional check for the value of this tag.

dgault:
Pushed a commit that will fix the 2 new sample files from the Image.sc thread. However its certainly not the proper solution. 

Looking at the tags on the various samples along with the samples from the previous montage work I haven't yet found an appropriate value that would indicate what we need, for example Im not sure how to handle `Images Stored To Disk: 0.0`. The fact that these new samples also each have the multiple positions listed in each file is also confusing me.

sbesson:
As discussed a few weeks ago, the current implementation should still fix the issue with the [sample datasets](https://downloads.openmicroscopy.org/images/Gatan/imagesc-36590/) uploaded as part of https://forum.image.sc/t/problem-reading-some-dm4-files-in-bioformats-6-4-0/36590. However it is still unclear from the tag structure how to decide whether montages should be split into multi-series or combined as multidimensional images.

d6a2f70 defensively adds a reader option allowing users to disable the montage split functionality.

dgault:
The new reader option makes sense to have with our current knowledge of the files. Retested with the same samples to ensure that the new option behaviour was working as expected. Everything appears to be behaving correctly so go ahead and merging this PR and the config PR.

PR #3562, opened 2020-05-14
SCN: No color correction for SCN400

dgault:
This is a test PR for a potential fix to https://github.com/ome/bioformats/issues/3510

Currently color correction is applied for any non Versa systems, in this case an SCN400 system has been used that also does not require correction. Unfortunately I cant find anything else in the metadata that would indicate if correction is or isn't required. Opening this PR as a test to see its impact on other files in the data repo.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
@@ -230,7 +230,7 @@ public class LeicaSCNReader extends BaseTiffReader {
     // newer files from Versa systems do not
     if (handler != null) {
       Image i = handler.imageMap.get(0);
-      tiffParser.setYCbCrCorrection(!"versa".equalsIgnoreCase(i.devModel));
+      tiffParser.setYCbCrCorrection(!"versa".equalsIgnoreCase(i.devModel) && !"leica scn400".equalsIgnoreCase(i.devModel);
     }
   }
 
--- a/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
@@ -230,7 +230,7 @@ public class LeicaSCNReader extends BaseTiffReader {
     // newer files from Versa systems do not
     if (handler != null) {
       Image i = handler.imageMap.get(0);
-      tiffParser.setYCbCrCorrection(!"versa".equalsIgnoreCase(i.devModel) && !"leica scn400".equalsIgnoreCase(i.devModel);
+      tiffParser.setYCbCrCorrection(!("versa".equalsIgnoreCase(i.devModel) || "leica scn400".equalsIgnoreCase(i.devModel)));
     }
   }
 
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
```diff
@@ -230,7 +230,7 @@ protected void initTiffParser() {
     // newer files from Versa systems do not
     if (handler != null) {
       Image i = handler.imageMap.get(0);
-      tiffParser.setYCbCrCorrection(!"versa".equalsIgnoreCase(i.devModel));
+      tiffParser.setYCbCrCorrection(!"versa".equalsIgnoreCase(i.devModel) && !"leica scn400".equalsIgnoreCase(i.devModel);
```
mtbc, line 233:
```suggestion
      tiffParser.setYCbCrCorrection(!("versa".equalsIgnoreCase(i.devModel) || "leica scn400".equalsIgnoreCase(i.devModel)));
```

dgault:
A number of repo files have been impacted by this and will need further investigation to confirm if the new pixel values are correct: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

dgault:
Builds and repo tests are now green with the latest config PR included, alternate MD5 values have been added as per the other SCN files as the values differ depending on the Java version.

components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
```diff
@@ -230,7 +230,7 @@ protected void initTiffParser() {
     // newer files from Versa systems do not
     if (handler != null) {
       Image i = handler.imageMap.get(0);
-      tiffParser.setYCbCrCorrection(!"versa".equalsIgnoreCase(i.devModel));
+      tiffParser.setYCbCrCorrection(!("versa".equalsIgnoreCase(i.devModel) || "leica scn400".equalsIgnoreCase(i.devModel)));
```
sbesson, line 233: If the list of scanners keeps growing, should we start maintaining a static list at the reader level?

sbesson (approved):
Functionally tested with the sample Leica SCN files from the curated QA repository acquired with the SCN400 slide scanner.

Without this PR, both the WSI and the macro image appear pink as described in the original issue. With this PR included, the color is restored. The proposed change is effectively the same as the one that was used for the datasets acquired with the Versa scanner in #3240.

One inline question about the value of extracting the list of scanners into a reader static attribute. Otherwise, looks good to me. I'll double check with @melissalinkert whether we have additional private sample files that should be tested with this change included.

melissalinkert:
I don't think we have additional private samples that need testing - the case @sbesson and I discussed earlier is a different format not handled by this reader.

I'm fine with this being merged as-is. As a follow-up, it might be worth an option to make the color correction configurable, so we don't have to worry so much about changes like this in the future. That shouldn't hold up the 6.11.1 release though.

sbesson (approved):

PR #3563, opened 2020-05-15
CZI: Parse TimeSpan Default Units

dgault:
Fixes https://github.com/ome/bioformats/issues/3560

This issue was raised in https://forum.image.sc/t/czi-reader-illuminationtype-mapping-and-exposuretime-units/37557/4 and a sample file showing the issue is available in QA-29273

This now parses the metadata value `Experiment|AcquisitionBlock|TimeSeriesSetup|Interval|TimeSpan|DefaultUnitFormat` and makes use of the units which had previously been hardcoded to seconds (note, seconds remains the default if the tag is not available).

This may have a knock on effect on sample files in the data repo, in which they will need to be reviewed to ensure the new values are correct.

To test:
- Using the file QA-29273, without the PR the exposure time will be in seconds
- With this PR included the exposure time should correctly be ms

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -69,6 +69,7 @@ import ome.units.quantity.Power;
 import ome.units.quantity.Pressure;
 import ome.units.quantity.Temperature;
 import ome.units.quantity.Time;
+import ome.units.unit.Unit;
 import ome.units.UNITS;
 
 import org.xml.sax.SAXException;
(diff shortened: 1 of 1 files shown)


dgault:
This has caused changes to many existing files (https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/62854/consoleFull), many of those using `ms` are likely to be correct and worth updating. Some have updated to `min` which needs further investigation as this looks to be incorrect.

dgault:
@melissalinkert, as discussed, I am unable to verify the ground truth for this change. There are a significant number of failures, the ones that I would be most concerned about are the ones which have updated the units to `mins` which looks incorrect: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/62854/consoleFull

Below are a sample of some of the failures with min as units and may be worth verifying first:
```
/data/public/idr0011/Plate1-Blue-A_TS-Stinger/Plate1-Blue-A-26-Scene-2-P1-E5-02.czi
/data/public/idr0011/Plate1-Blue-A_TS-Stinger/Plate1-Blue-A-40-Scene-2-P2-D8-02.czi
/data/public/idr0011/Plate1-Blue-A_TS-Stinger/Plate1-Blue-A-24-Scene-1-P1-F4-01.czi
```

melissalinkert:
@dgault: screenshots for several of the failing files are now in https://github.com/openmicroscopy/data_repo_config/issues/552 (can add to that if needed).

dgault:
Thanks Melissa, closing this PR as it is resulting in a large number of values which do not match the values in Zen.

PR #3564, opened 2020-05-15
Cellomics: use maximum row/column index to calculate plate size

melissalinkert:
Backported from a private PR.

Test plate is in ```inbox/cellomics-plate-size```.  Compare ```showinf -nopix -omexml``` with and without this PR.  Without this PR, only the first well is linked to the plate.  With this PR, both wells are linked to the plate, and the plate has 384 instead of 96 wells.

This should be safe for a minor release; memo files and tests should be unaffected.


melissalinkert:
Hold off on merging this for the moment please, one more commit may be required.

melissalinkert:
No more commits planned after 213ff7e, assuming tests pass.

PR #3568, opened 2020-06-10
SVS/AFI: fix NPE when exposure time is not defined

melissalinkert:
Backported from a private PR.

Defends against a potential NPE if either the exposure time or units are missing from an SVS file.  Tests should remain green, and this should be safe for a patch release.


sbesson:
https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/419/ was green today

PR #3569, opened 2020-06-10
Cellomics: set physical sizes for all series

melissalinkert:
Backported from a private PR.

Fixes typo that caused physical sizes to only be set for the first series.  Requires a configuration change, but tests should otherwise pass and this should be safe for a patch release.


sbesson:
Despite Travis being green, it looks like the status did not make it back to GitHub causing this PR to be excluded from the daily CI builds and https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/418/ to fail due to mismatches with the configuration changes. /cc @mtbc 

Closing/re-opening for inclusion in tomorrow's builds.

Issue #3571, opened 2020-06-11
Bug: Invalid z position in getIndex


Nicholas-Schaub:
**Summary**
I'm receiving an error when trying to directly read an image plane by specifying the z-position, but when I directly set the index I receive no error and can successfully read the plane.

**Image Information**
I'm working with an `.ome.tif` that only has XYZ data (no channels, timepoints, pyramids, etc). The image is stored using the tile settings (tile width and height are both 1024). 

**Additional Information**
All of my work is currently done using `python-bioformats` using the ImageReader class. The issue traces back to the FormatWriter in `loci_tools` jar (as discussed below). I believe this is an issue with newer versions of `loci_tools.jar`, because I've swapped out different versions of the jar and the issue no longer appears. It only seems to happen with newer versions of the jar.

When I get the number of z-positions from the metadata, it correctly returns the number of z-positions. However, when I try to read the image and specify a z-position other than 0, I get an error that traces back to `FormatWriter.getIndex` that says the z value is out of range. If I manually calculate the index and use `setIndex` then read the plane, it gives me the expected plane.

mtbc:
Could you give us a clear, simple example that shows the problem? Go ahead and attach an example image here if you like, link to a script in a gist, etc. We'd be happy to try to figure out what's going on.

Did you manage to pin down which BF version introduces the regression? That may help us to pinpoint the culprit.

Nicholas-Schaub:
I could give you error traces if that would be helpful. While I have used Bioformats in Java in the past, I mainly interact with it through Python, so I'll be sharing what I do in Python.

The version of `loci_tools.jar` used in `python-bioformats` is from Bioformats 5.9.0:
https://github.com/CellProfiler/python-bioformats/tree/master/bioformats/jars

For various reasons, I started using `loci_tools.jar` from Bioformats 6.5.0, and I have a link to the jar I use in my repo. I built a tool we use in our organization for reading/writing all images in a cloud computing platform we are building:
https://github.com/Nicholas-Schaub/polus-plugins/tree/bfio/document/utils/polus-bfio-util

To use Bioformats 6.5.0, you have to manually copy the `loci_tools.jar` into the `bioformats` Python package after installing with `pip install python-bioformats`.

Then, the following code should reproduce the error:
```python
import bioformats
import javabridge as jutil

jutil.start_vm(class_path=bioformats.JARS)

file_path = '/path/to/3d/file`

try:
    with bioformats.ImageReader(self._file_path) as reader:
[...]

PR #3576, opened 2020-06-19
MikroscanTiffReader: set suffixSufficient to false

sbesson:
See #3571 for more context

This change should allow `MikroscanTiffReader.isThisType(name, open)` to fail in case `open` is set to false i.e. file access is disallowed

I expect this change should have no impact on the standar Bio-Formats detection logic in `ImageReader`. Non-regression CI builds should keep passing against all sample datasets including the Mikroscan TIFF ones.

When the `MikroscanTiffReader.isThisType(name, open)` API is used with `open` set to false like in the case `python-bioformats`, this should prevent all TIFF files to be erroneously flagged as Mikroscan TIFF. The code snippet from https://github.com/ome/bioformats/issues/3571#issuecomment-643281905 could be used for testing this PR after replacing the JAR with a version of Bio-Formats including this PR.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MikroscanTiffReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MikroscanTiffReader.java
@@ -58,6 +58,7 @@ public class MikroscanTiffReader extends SVSReader {
     super("Mikroscan TIFF", new String[] {"tif", "tiff"});
     domains = new String[] {FormatTools.HISTOLOGY_DOMAIN, FormatTools.LM_DOMAIN};
     suffixNecessary = false;
+    suffixSufficient = false;
   }
 
   // -- IFormatReader API methods --


melissalinkert:
Might be worth setting ```suffixSufficient = false``` in the constructor if you haven't tried that already. That should automatically cause ```super.isThisType(name, false)``` to return false.

dgault:
Tested with a short Java program calling the `isThisType` API with standard TIFF files.

Without this PR:
`isThisType(fileID, false)` returned true
`isThisType(fileID, true)` returned false

With this PR:
`isThisType(fileID, false)` returned false
`isThisType(fileID, true)` returned false

I carried out some additional tests using ImageReader to confirm that this has caused no unexpected regressions. Builds and repo tests have also been unaffected. Looks good to merge.

PR #3577, opened 2020-06-23
do not log if omero

jburel:
Do not log id if omero
cc @dgault @sbesson 

see: https://github.com/ome/omero-insight/issues/136

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1381,7 +1381,10 @@ public abstract class FormatReader extends FormatHandler
    */
   @Override
   public void setId(String id) throws FormatException, IOException {
-    LOGGER.debug("{} initializing {}", this.getClass().getSimpleName(), id);
+    if (!isOmero(id)) {
+      LOGGER.debug("{} initializing {}", this.getClass().getSimpleName(), id);
+    }
+    
 
     if (currentId == null || !new Location(id).getAbsolutePath().equals(
       new Location(currentId).getAbsolutePath()))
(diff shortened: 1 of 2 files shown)


joshmoore:
Added a link to the original issue so we have some context.

joshmoore:
@jburel : would another option be to only print the first line everywhere? If this had been formatted as an `URI` then parsing with that class would allow us to strip out parameters, but that generalization doesn't work. If ImageJ's file/url-protocol is newline separated, though, then that may be a more general solution.

jburel:
In the case of OMERO, it does not really offer much, since it also printed out without the password immediately after, admittedly in the debug mode i.e. there was a lot of redundancy in the printing out.
We also already have the initial connecting step recored when the user logs in to the server when using the plugin.

For the formatting, I don't think it is wise to change it. It will require a review of the component using it, including scifio. This is for another review in my view

joshmoore:
Sorry, I don't follow. I agree that all the other locations need work, too. What I'm wondering about is whether or not we can't do something more general than adding an `isOmero` method to a component which should have no knowledge of OMERO.

jburel:
The ``isOmero`` is not a new check. I just made a method since it is now used in 2 places. I went for the less disruptive change.

components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -2076,4 +2076,9 @@ public static int getRequiredDirectories(String[] files)
 
     return (int) Math.max(maxDirCount - dirCount, 0);
   }
+
+  static boolean isOmero(String id) {
```
sbesson, line 2080: Note this is now turning an API addition and would redefine the scope of the release to a minor release increment. Should we discuss it during the weekly formats meeting.

dgault: Yeah, lets discuss at the formats meeting and make the call then.

jburel: alternative is to have the same method as private in 2 classes.

dgault:
Following the discussion at the formats meeting Im happy to have this included in 6.5.1 using the 2 private methods

PR #3579, opened 2020-06-29
Metamorph: expand wavelength parsing

melissalinkert:
Backported from a private PR.

This will attempt to use wavelengths from Metamorph XML or original metadata to populate ```Channel.EmissionWavelength``` and ```Channel.LightSourceSettings.Wavelength``` if no other wavelengths are defined.

I think this would be safe for a patch release.  The only potential test impact I would expect is that some files might need to have an emission wavelength added to the configuration.  If that turns out to be the case, I will open a config PR tomorrow.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -39,6 +39,7 @@ import java.util.Map;
 import java.util.HashMap;
 import java.util.AbstractMap;
 import java.util.Collections;
+import java.util.Vector;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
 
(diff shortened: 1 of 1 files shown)


sbesson:
See test failures https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/69432/console and https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/69485/console. A companion configuration PR will need to be opened for all Metamorph samples where the wavelength is populated as per these changes.

dgault:
Overall these changes look to be fine. Due to the delayed release of 6.5.1 this should be able to be included as part of that release.

There are a number of unrelated changes in the config PR which seem to be older changes that are just getting picked up now when the config are regenerated. Once I've confirmed these are all safe I will get this merged.

dgault:
The other config changes mentioned in the above comment are unrelated to this PR and safe to add. Merging for inclusion in 6.5.1

PR #3583, opened 2020-07-06
Operetta: fix metadata files logic to skip plate folders

sbesson:
This commit deals with the scenario of an IDR submission where plate folders containing the TIFF data + the metadata XML files directly (without the usual Images subdirectory) were found under the same current level. The assumptions of the reader currently causes all other plates to be added to the fileset as ancillary metadata files, increasing the size of the original 7K fileset to 55K and slowing down import down the line.

The new logic adds early directory checks when adding metadata files to skip either the directory containing the master XML file or other directories containing XML files.

From the mainline Bio-Formats, the main impact is that there should be no regression at the level of the automated tests. There should be a slight performance improvement to `setId` as the current directory should be skipped early during the discovery of associated metadata files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -288,16 +288,25 @@ public class OperettaReader extends FormatReader {
     for (String f : list) {
       Location path = new Location(parent, f);
       if (path.isDirectory()) {
+        // the current file's parent directory will usually be "Images",
+        // but may have been renamed especially if there are no
+        // analysis results
+        if (f.equals(currentFile.getParentFile().getName())) {
+          LOGGER.trace("Skipping current directory {}", f);
+          continue;
+        }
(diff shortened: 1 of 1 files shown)


melissalinkert:
This looks fine to me; no obvious signs of breakage.

dgault:
Changes look sensible and have not caused any breakage in repo tests. Some manual testing showed the updated behaviour to be as expected with no regressions. Good to merge.

sbesson:
Unfortunately it looks like the logic introduced here only solves half the problem. As shown by the fileset of https://idr.openmicroscopy.org/webclient/?show=well-1533562, all TIFF files of mismatching plates are indeed excluded from the used files list however the top-level `Index.idx.xml` files are still added as ancillary files.

The major implication is probably that `omero import --exclude clientPath` should not be used on such datasets. I will occasionally look into an additional commit to further remove the unrelated `Index.idx.xml`.

PR #3586, opened 2020-07-14
MetamorphReader: allow loading of single stk files

imagejan:
In case the id is pointing to an *stk* file (and not to the main *nd* file), returning `CAN_GROUP` instead of `MUST_GROUP` here should provide the flexibility of choosing between loading a single stk file or the entire dataset.

This leads to people using more or less ugly workarounds, e.g.:

https://github.com/TissueMAPS/TissueMAPS/blob/59f44ce1c76fe3c3abb0a932a8badc834b122e98/tmlibrary/tmlib/workflow/metaconfig/visiview.py#L168-L176

or entire macros for custom metadata handling, using ImageJ's `open` instead of bio-formats to open the single stk files:

https://github.com/bvernay/ImageJ-Macros/blob/558415ef8ea619e45f0d5010fdcf791bc3e46aab/Metamorph%20Utilities/MetamorphFilesFolderToHyperstacks.ijm

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -249,7 +249,7 @@ public class MetamorphReader extends BaseTiffReader {
       if (checkSuffix(file, ND_SUFFIX) &&
        l.getName().startsWith(file.substring(0, file.lastIndexOf("."))))
       {
-        return FormatTools.MUST_GROUP;
+        return FormatTools.CAN_GROUP;
       }
     }
 


dgault:
Thanks @imagejan for opening the PR. I will include it in our overnight builds to test against the samples in our data repo to ensure there are no unwanted knock on effects.

imagejan:
Hi @dgault, any news on this?

dgault:
Hi @imagejan, sorry for the delay in getting back to this. Within the ImageJ world, there is an option in the importer window to `Open file individually`. Would this achieve the same behaviour you are looking for with this PR?

imagejan:
@dgault indeed, that's what I was using in the UI. For opening via API, I *thought* I had tried `ImporterOptions.setGroupFiles(false)` before without any success, but another trial just now proves me wrong: the following script works just fine with the latest version of Bio-Formats:

```groovy
#@ File inputFile

import loci.plugins.in.ImporterOptions
import loci.plugins.BF

options = new ImporterOptions()
options.setGroupFiles(false)
options.setId(inputFile.getAbsolutePath())

imps = BF.openImagePlus(options)
imps[0]
```

PR #3588, opened 2020-07-16
Cellomics: parse channel data from companion .mdb file

melissalinkert:
Backported from a couple of private PRs.  See also https://trac.openmicroscopy.org/ome/ticket/7092

This updates the ```MDBService``` API to allow retrieval of one table at a time.  ```CellomicsReader``` uses the new API to retrieve just the channel table from the optional companion .mdb file, so that channel names, colors, and plane exposure times are now populated.

This will require a minor release, and may require a configuration PR to add channel/plane metadata.  If there are any other test failures, feel free to exclude for now.

### DIFF
--- a/components/formats-gpl/src/loci/formats/services/MDBService.java
+++ b/components/formats-gpl/src/loci/formats/services/MDBService.java
@@ -58,6 +58,21 @@ public interface MDBService extends Service {
    */
   public Vector<Vector<String[]>> parseDatabase() throws IOException;
 
+  /**
+   * Read named table from a pre-initialized .mdb files.
+   * Each String[] in a table represents a row, and each element of the String[]
+   * represents the value of a particular column within the row.
+   *
+   * The first row in the table contains the names for each column.
+   *
+   * {@link #initialize(String)} must be called before calling parseTable.
+   *
+   * @param name table name
+   * @return table data or null if the named table does not exist
+   * @throws IOException if there is a problem reading the table data
+   */
+  public Vector<String[]> parseTable(String name) throws IOException;
+
   /** Close the currently initialized file. */
   public void close();
 
--- a/components/formats-gpl/src/loci/formats/services/MDBService.java
+++ b/components/formats-gpl/src/loci/formats/services/MDBService.java
@@ -71,7 +71,9 @@ public interface MDBService extends Service {
    * @return table data or null if the named table does not exist
    * @throws IOException if there is a problem reading the table data
    */
-  public Vector<String[]> parseTable(String name) throws IOException;
+  public default Vector<String[]> parseTable(String name) throws IOException {
+    return null;
+  }
 
   /** Close the currently initialized file. */
   public void close();
--- a/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellomicsReader.java
@@ -149,6 +149,7 @@ public class CellomicsReader extends FormatReader {
     if (!fileOnly) {
       files.clear();
       cellomicsPattern = null;
+      metadataFiles.clear();
     }
   }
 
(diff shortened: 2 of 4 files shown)


sbesson:
This seems to have caused some regressions on our Cellomics datasets - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/72403/console

components/formats-gpl/src/loci/formats/services/MDBService.java
```diff
@@ -58,6 +58,21 @@
...
+   *
+   * The first row in the table contains the names for each column.
+   *
+   * {@link #initialize(String)} must be called before calling parseTable.
+   *
+   * @param name table name
+   * @return table data or null if the named table does not exist
+   * @throws IOException if there is a problem reading the table data
+   */
+  public Vector<String[]> parseTable(String name) throws IOException;
```
sbesson, line 74: Since this is an interface we might want to add a default method returning `null` to avoid a breaking change

sbesson, line 74: Since this is an interface, a default method should be declared to prevent a breaking change e.g. returning `null` unconditionally.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#513](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/513/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/513/consoleText) for more details.
Possible conflicts:
  - PR #3561 melissalinkert 'Split CellWorX and MetaXpress support into separate readers'
    - components/test-suite/src/loci/tests/testng/FormatReaderTest.java

--conflicts

components/formats-gpl/src/loci/formats/services/MDBService.java
melissalinkert: Fixed in 51e6ee7

sbesson:
Tested an OMERO based workflow using the sample Cellomics plate in our curated QA repository. The `setId` call detects and groups 121 files including the mdb file vs 120 without this PR. The plate dimensions are unchanged by this PR.

Looking at the detail of individual images without this PR
Screen Shot 2020-07-24 at 14 13 15

vs with this PR

Screen Shot 2020-07-24 at 14 13 10

- the channel names and colors are no longer the default (0,1,2) (red, green,blue) suggesting they have been populated from the mdb metadata file as expected
- from the acquisition panel, one can confirm the plane timestamps are also populated

Two issues which might be outside the scope of this PR but are revealed:

- I was expecting the repository tests to fail on the sample `cellomics` dataset due to the extra metadata population but they have been passing. From the following code snippet, it feels like there is some special handling of the `ChannelName_xx = null` scenario. In addition to regenerating the configuration file to include the new metadata (channel names and plane timestamps), should we review this assumption and always fail when channel names are mismatching?

https://github.com/ome/bioformats/blob/6f50e4d52c9d96112635fd8b2dde737f31041cf0/components/test-suite/src/loci/tests/testng/FormatReaderTest.java#L1094-L1098

- as shown in the screenshot above, the default thumbnail generation process at import time created dark thumbnails. The same behavior happened when skipping thumbnails at import time and regenerating them on the fly with the clients. When (re)saving the rendering settings from import, the expected thumbnail is generated. I am unsure whether this code is at fault or some other PR in the stack.

melissalinkert:
Also have a report of the dark thumbnail issue on the GS side, so assuming it's due to this PR somehow.  Investigating now, but feel free to exclude until there is a solution.

melissalinkert:
I think 0fa032f will fix the dark thumbnail issue.  As @sbesson suggested, having the channel color's alpha component set to 0 seems to be the problem.  I was able to reproduce that separately by importing a fake OME-TIFF with 2 channels set to 16711680 (green, alpha = 0) and 65280 (blue, alpha = 0).  Single channel datasets are unaffected.

Other readers take the same approach of setting alpha to 255, so I don't have a problem with doing the same thing here.  At some point, though, this should maybe be addressed either in the channel handler during import or when viewing/generating thumbnails.

PR #3593, opened 2020-07-24
Increment the Memoizer version

sbesson:
Follow up of #3557, this PR bumps the internal version number of the `Memoizer` wrapper and the `ome-common` dependency to use the same `kryo` version across the board.

Tests should be unaffected and keep passing. A test for the version bump would be to test the memo file invalidation workflow:
- create a memo file using the last stable Bio-Formats release e.g. by running `showinf -nopix -cache -debug` 
- reopen the file with a version of Bio-Formats including  this PR  using `showinf -nopix -cache -debug`
- check the memo file has been invalidated and recreated as well as the error message in the log

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -333,7 +333,7 @@ public class Memoizer extends ReaderWrapper {
    * cached items. This should happen when the order and type of objects stored
    * in the memo file changes.
    */
-  public static final Integer VERSION = 3;
+  public static final Integer VERSION = 4;
 
   /**
    * Default value for {@link #minimumElapsed} if none is provided in the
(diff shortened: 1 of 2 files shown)


dgault:
Builds and tests have continued to be green with this PR included.

Tested using the memo generation on `showinf` as recommended. First running the command (`showinf -nopix -cache -debug`) without the PR included and the old memoizer version.

Then with this PR and running the same command a new memo file is generated and the debug output shows that the version mismatch was detected as expected:
```
Old version of memo file: 3 not 4
start[1595858460408] time[1] tag[loci.formats.Memoizer.loadMemo]
....
saved to temp file: /Users/dgault/Documents/Sample Images/ome_tiff/.time seriesTest.ome.tif.bfmemo1685298991441587511
start[1595858476102] time[89] tag[loci.formats.Memoizer.saveMemo]
saved memo file: /Users/dgault/Documents/Sample Images/ome_tiff/.time seriesTest.ome.tif.bfmemo (49493 bytes)
start[1595858460369] time[15823] tag[loci.formats.Memoizer.setId]
```

PR #3596, opened 2020-07-28
Bump ome-common to version 6.0.6

sbesson:
See https://forum.image.sc/t/fiji-wont-open-czi-files-after-update-there-was-a-problem-with-the-class-java/40932
for the rationale and the issue with ome-common 6.0.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,6 @@ matrix:
   exclude:
   - jdk: openjdk11
     env: BUILD=ant
-  - jdk: openjdk8
-    env: BUILD=maven
 
 script:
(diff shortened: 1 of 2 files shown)


sbesson:
dd6be1e re-enables the Maven + JDK8 combination in Travis where tests fail as expected due to the compilation issue with `ome-common 6.0.5`.  6c9dc93 bumps to `ome-common 6.0.6` and passes the tests on the minimal supported JDK version.

dgault:
Reenabling the Travis tests should hopefully catch this in the future. The new om-common version looks correct and builds as expected with all tests passing.

PR #3602, opened 2020-07-31
SIS: Trim imageName and channelName

### LINKED ISSUE #3598: SISReader: Image name and channel name include trailing null byte
Hi,

I recently was using QuPath 0.2.1 which uses Bioformats 6.5.1 with some (SIS-)TIFF files created by Olympus cell^F and found that the image name and channel name as provided by the Bioformats reader include the trailing null byte at the end of the string.

I looked up the code in RandomAccessInputStream.java and found that readCString essentially calls ReadString which in my understanding reads through the null byte and returns it along with the string.

I added a call to trim() to the readCString function call in lines 174 and 239 in SISreader.java and this fixes the problem for me.

The image file I used to test this can be found at
https://box.med.uni-heidelberg.de/s/qjr7vvn7P9akLMx
Password for access: QuPathrocks1! (please include the exclamation mark)

The associated QuPath issue is: https://github.com/qupath/qupath/issues/573

Thank you for your work & best regards
Simon

dgault:
This PR is implementing the recommended changes from https://github.com/ome/bioformats/issues/3598
Opening to test the impact on the data repo

Fixes #3598

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SISReader.java
@@ -171,7 +171,7 @@ public class SISReader extends BaseTiffReader {
     acquisitionDate = DateTools.formatDate(acquisitionDate, "yyyy-M-d H:m");
 
     in.skipBytes(6);
-    imageName = in.readCString();
+    imageName = in.readCString().trim();
 
     in.seek(metadataPointer);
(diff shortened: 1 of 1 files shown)


dgault:
PR has had no negative impact on any existing files and repo tests continue to pass.

PR #3611, opened 2020-08-24
Remove FileStitcher from FormatReaderTest and FormatReaderTestFactory

sbesson:
The `FileStitcher` has been included in the automated tests since the inception of the `test-suite` component. This means that file stitching based on name pattern is automatically executed fo every fileset for every format in the curated QA repository - see [here](https://downloads.openmicroscopy.org/images/DV/U2OS/), [here](https://downloads.openmicroscopy.org/images/DICOM/nema/WG04/IMAGES/J2KR/) or. [here](https://downloads.openmicroscopy.org/images/PNG/user-4/dataset-user-4/) for public filesets where several files are grouped by stitching. For some filesets, the non-stitched filesets have mismatching dimensions and the stitching logic fails. In these cases, the only way to configuration the files for testing is to artificially split them into separate folders - see [here](https://downloads.openmicroscopy.org/images/Leica-SCN/openslide/).

This PR proposes to remove the `FileStitcher` from the `FormatReaderTest` and other relevant locations. A separate private PR updates all configuration files for all filesets that were previously grouped by the stitcher, testing each non-stitched fileset instead. The performance of the daily CI job should be largely unaffected.

The stitching functionality will remain tested by the non-regression tests specifically in the case of the `.pattern` filesets since the associated `FilePatternReader` largely delegates its logic to `FileStitcher`.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -38,7 +38,6 @@ import loci.common.IniParser;
 import loci.common.IniTable;
 import loci.common.IniWriter;
 import loci.common.Location;
-import loci.formats.FileStitcher;
 import loci.formats.FormatException;
 import loci.formats.FormatTools;
 import loci.formats.IFormatReader;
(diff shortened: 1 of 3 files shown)


dgault:
As has been discussed, merging this PR now to get this included in 6.6.0. Builds and tests have been green for quite a while with both this and the config PR included.

I will follow up with investigation into adding configs for `samples/dub` dataset and reviewing the Bio-Rad datasets for potential updates..

PR #3613, opened 2020-09-08
Review and update the instances of image index in the API documentation

sbesson:
Reported in https://forum.image.sc/t/loci-formats-in-ometiffreader-loads-different-slices-than-loci-formats-in-tiffreader/42536/9, the Javadoc of various APIs still refers to the `the image index within the file.` which erroneously lead to thinking the reader is iterating over the IFDs in order within a TIFF file for instance.

This PR updates all relevant locations to use `the plane index within the current series` as was done in `ome-files-cpp`.


dgault:
The updated Javadoc is a much better representation of the actual parameters and should help avoid confusion in future

PR #3617, opened 2020-10-03
Options: test-suite improvements

sbesson:
Folllowing the new API introduced in https://github.com/ome/bioformats/pull/3605, this PR makes a few modifications primarily in the `test-suite` component in order to run the non-regression tests against filesets containing `.bfoptions` files.

- the readers from `Configuration` and `FormatReaderTest` are systematically initialized with a `DynamicMetadataOptions` instance as the latter is required to load options file
- skip `.bfoptions` file in `FormatReaderTest.isThisType()`
- the options file is appended to the `FormatReader.getUsedFiles` API rather than being prepended. The rationale for doing so is that the `ImageReader.getRequiredDirectories()` determines the reader on the first file in this list. Without this change, `FormatReaderTest.testRequiredDirectories()` was failing with an `UnknownFormatException` against a fileset with an options file.
-  `DynamicMetadataOptions.getMetadataOptionsFile()` and `DynamicMetadataOptions.loadOptions()` are modified to return a `String` and take a `String` as an input (like `FormatReader.getCurrentId`). Internally, the function logic is updated to use `loci.common.Location` rather than `java.io.File`.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatReader.java
+++ b/components/formats-api/src/loci/formats/FormatReader.java
@@ -1020,11 +1020,6 @@ public abstract class FormatReader extends FormatHandler
   public String[] getUsedFiles(boolean noPixels) {
     String[] seriesUsedFiles;    
     Set<String> files = new LinkedHashSet<String>();
-    File optionsFile = DynamicMetadataOptions.getMetadataOptionsFile(currentId);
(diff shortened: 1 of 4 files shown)


dgault:
Changes all look good and look testing hasn't identified any regressions. I will merge once I get the repo tests back green again.

PR #3618, opened 2020-10-07
Memoizer: Persist caching to sub readers

dgault:
Initial PR for testing and discussion of a possible option for passing Memoization to sub readers. In this case each reader must wrap its sub readers, which leaves a significant amount of testing required to cover the large number of readers involved.

- MetadataOptions used to store Memoizer otpions
- A wrap method in Memoizer to handle the wrapping
- Each reader then wraps its sub readers

Likely to be further commits to bring this more inline with the IDR branch
New MetaxpressTiffReader also needs to be updated

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -46,6 +46,8 @@ import loci.common.RandomAccessOutputStream;
 import loci.common.services.DependencyException;
 import loci.common.services.ServiceException;
 import loci.common.services.ServiceFactory;
+import loci.formats.in.DynamicMetadataOptions;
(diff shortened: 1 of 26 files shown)


joshmoore:
First off, wow. I find this very exciting. I only ever approached propagating to the most necessary readers in the IDR context.

I do wonder if we're doing this across the board now if there are any other wrapping concerns we should consider. I imagine that might work by letting the `Options` object decide which wrappings to apply.

joshmoore:
With the change of IDR's strategy (in this case, to not support the .screen format) this PR is currently unneeded. I've backed up the branch to my repo just in case.

_Discussed in today's formats meeting with @melissalinkert, @sbesson, @jo-mueller, @dominikl and @kkoz_

PR #3624, opened 2020-10-19
Optionsfiles fixes

sbesson:
Discussed during the weekly Formats meeting,

- 6f1716c and be2b2bd address the concern initially raised in https://github.com/ome/bioformats/pull/3605#discussion_r501092971 by making option filenames unique
- b51a2f6 is a minor fix for a warning discovered during testing and should avoid unnecessary warnings related to the `IniList` way of storing headers


dgault:
Changes work as expected with options files now unique and match the expectation from the formats meeting feedback. The new tests also look good and all pass. No other regressions have been found so this looks good to merge.

PR #3626, opened 2020-10-27
Actions on PR and tag

jburel:
Add support for actions.
Build on windows and ubuntu on pull_request
When a tag is pushed, the artifacts will be uploaded to the GitHub tag (similar to downloads.o.org)
I have not removed the hard-coded version for this round
Token will need to set on the repo in order to push the tag

I have not removed the travis or appeyvor files
I have tested the workflow using on my account see https://github.com/jburel/bioformats/releases/tag/v6.6.0-m3 for example

cc @dgault


components/formats-bsd/src/loci/formats/UpgradeChecker.java
```diff
@@ -85,7 +85,7 @@
    * Location of the JAR artifacts for the stable releases.
    */
   public static final String STABLE_BUILD =
-    "http://downloads.openmicroscopy.org/bio-formats/" +  STABLE_VERSION + "/artifacts/";
+    "http://downloads.openmicroscopy.org/bio-formats/";
```
sbesson, line 88: For anyone consuming this field, this would be a breaking change. Could we use the unversioned https://downloads.openmicroscopy.org/bio-formats/latest/artifacts/ URL as a replacement and deprecate the static `STABLE_VERSION` field?

If we publish using GitHub artifacts, I assume this will be replaced by the appropriate URL?

jburel: I think that's what I had something got lost when cleaning up stuff

manics:
Did you investigate any existing actions e.g. https://github.com/actions/create-release https://github.com/actions/upload-release-asset ?

jburel:
@manics I have played with both of them, yes but opted for that solution due to the steps required like creation of checksum etc.

sbesson:
A few comments:

- :+1: for the Python 3 upgrade and the UpgradeChecker deprecation
- re https://github.com/ome/bioformats/pull/3626#issuecomment-718543084, I am definitely in favor of using upstream organizational roles whenever possible. A limitation is that the GitHub actions syntax does not support for loops so I assume it would mean duplicating the action for each asset including checksums. Mid-term I would hope https://github.com/actions/upload-release-asset/issues/60 will allow us to use `actions/upload-release-asset` with a wildcard.
- most of the Travis CI workflow is now converted with the exception of the artifact deployment to Artifactory - https://github.com/ome/bioformats/blob/c14984f78990f7d37f81e444f609125f4ec1e5b1/.travis.yml#L38-L46. Is this in scope of this work
- the current workflow combines Ant and Maven as matrix components. An alternate representation could be to split them into two separate workflows: `ant.xml` and `maven.xml`. The main benefit of this approach is to be able to re-use the latter as the Maven commands are very generic. This can also be reviewed as a follow-up with other Maven-based Java libraries in mind.

jburel:
I think:
* I will remove the "release" file since we may not want to push all artifacts to GH for next release
* split into ant and maven as suggested above

jburel:
closing this PR see #3635 for the build part of this work

PR #3627, opened 2020-10-29
ND2Handler: Catch NPE when parsing key value pairs

dgault:
Issue was raised in forum thread https://forum.image.sc/t/command-line-bfconvert-to-convert-nd2-to-tif-java-error/43728/3

A sample file is available in QA-29634 which reproduces the reported bug in BF 6.5.1

To test:
Using the file in QA-29634 attempt to open the image using BF 6.5.1, this should throw a NullPointerException
With the PR included repeat the above and the file should open and display without exception. A warning message will be logged: `[WARN] Could not parse Power value: -nan(ind)`

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Handler.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Handler.java
@@ -919,7 +919,7 @@ public class ND2Handler extends BaseHandler {
         }
       }
     }
-    catch (NumberFormatException exc) {
+    catch (NumberFormatException | NullPointerException exc) {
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Code changes are straightforward and make complete sense.

Tested using the sample QA file with and without this PR. Without it, initializing the reader led into a `NullPointerException` as described. With this PR included, the reader could be initialized with all metadata and binary data accessible.

Only additional thing would be to add the sample file to the curated QA repository to prevent regressions. This should not prevent the inclusion of this PR in the upcoming minor release of Bio-Formats.

dgault:
Merging as tests are green with new sample file

PR #3633, opened 2020-11-09
APNGReader: fix hanging openBytes call

sbesson:
Following an investigation on a hanging import of a PNG, this bug addresses an issue when opening the pixel data of a PNG in tiles and the requested tile height is not a multiple of 8. Under some conditions, this could lead to a height larger than the image height passed to the `APNGReader.decode` and the reader hanging.

This PR should fix the behavior by capping the height passed to `APNGReader.decode` to the image height. It also adds some  preliminary checks to the `APNGReader.decode(PNGInputStream, int, int)` to throw an exception if the dimensions are too large.

Without this PR, some PNG files would fail an OMERO import, hanging during the tile-based min/max calculation step and freezing a server thread indefinitely. With this PR included, the same file should import successfully. 

/cc @pwalczysko @joshmoore

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/APNGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/APNGReader.java
@@ -358,6 +358,14 @@ public class APNGReader extends FormatReader {
     int bpp = FormatTools.getBytesPerPixel(getPixelType());
     int rowLen = width * getRGBChannelCount() * bpp;
 
+    if (width > getSizeX()) {
+      throw new FormatException(
+        "Width (" + width + ") exceeds image width (" +  getSizeX() + ")");
+    }
+    if (height > getSizeY()) {
+      throw new FormatException(
+        "Height (" + height + ") exceeds image height (" +  getSizeY() + ")");
+    }
     if (getBitsPerPixel() < bpp * 8) {
       int div = (bpp * 8) / getBitsPerPixel();
       if (div < rowLen) {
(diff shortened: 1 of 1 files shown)


joshmoore:
The fact that https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/582/ is green is certainly reassuring, though I wonder if we need to add such an evil PNG to the mix. The changes make general sense, though I wonder if we don't need to report upstream.

Copying URLs from https://trac.openmicroscopy.org/ome/ticket/13246#comment:11 (since it sometimes doesn't load):
* ​https://bugs.openjdk.java.net/browse/JDK-8060193
* https://issues.jenkins-ci.org/browse/JENKINS-14362
* http://stackoverflow.com/questions/6584765/thread-locking-when-flushing-jsp-file

(though I don't see any clear place on any of them where someone wants our feedback)

sbesson:
> I wonder if we need to add such an evil PNG to the mix. 

Happy to add a configuration for a representative sample PNG. I held off on it because I failed to reproduce the issue with our classical non-regression test suite which does not test this scenario. The issue is that [FormatReaderTest](https://github.com/openmicroscopy/bioformats/blob/develop/components/test-suite/src/loci/tests/testng/FormatReaderTest.java) includes a tile-based `openBytes` test but it only tests the case of a tile starting at `(0,0)` while this bug will show up for tiles close to `sizeY`.

pwalczysko:
Screen Shot 2020-11-10 at 14 01 48

The upload of all four pngs (including the problematic one) into merge-ci was successful.
Also, on merge-ci webclient, the thumbnails were generated successfully, and the problematic image could be opened successfully in

-  iviewer
- figure
- old image viewer

All looks good fmpov

dgault:
The change looks to make sense and hasn't caused any regressions in tests or existing datasets. Given that the new problematic files are also now working without issue lets get this merged and included with the 6.6.0 release.

PR #3638, opened 2020-11-19
Populate plate row and column dimensions

melissalinkert:
Noticed while working on https://github.com/glencoesoftware/bioformats2raw/pull/69.

Older HCS readers didn't set ```Plate.Rows``` and ```Plate.Columns```, since those values are optional and were introduced in the 2010-04 schema.  The only remaining reader that sets ```Plate.ID``` but not ```Plate.Rows``` and ```Plate.Columns``` is ```CellH5Reader```, but I suspect all of the plate population needs to be rewritten there so will save it for a later date.

Tests should continue to pass, and this should be safe for a patch release.


dgault:
Changes here look good and all the tests continue to pass. Manually tested a few samples for each format and the updates look to be behaving as expected. Good to merge.

PR #3646, opened 2020-12-10
Bump ome-model version to 6.2.2

sbesson:
See https://github.com/ome/ome-model/blob/v6.2.2/NEWS.md for the list of changes.


dgault:
Update has been successful without any regressions as far as I can tell. Merging for 6.6 release.

PR #3647, opened 2020-12-10
Deploy

jburel:
This PR deploys the snapshots to artifactory when a PR is merged into ``develop``.

I have also renamed the ``ant`` workflow so it is easier to include it as a badge and so it matches the approach taken in other files.

secrets will need to be set for the deployment to work
cc @sbesson @dgault 

Screenshot 2020-12-10 at 17 06 36

Edit:
The logic to push is as follow:
 * if a commit with version **with** snapshot is pushed to develop. Artifacts are pushed to ``ome.snapshots``
 * if a commit with version **without** snapshot is pushed to develop. Artifacts are pushed to ``ome.staging``
 * No deployment when a tag is push, we can deploy on tag but we need to add check to make sure we do not push twice i.e. develop without snapshot and tag


.github/workflows/maven.yml
```diff
@@ -32,3 +32,22 @@ jobs:
...
+  deploy:
+    if: contains('refs/heads/develop', github.ref)
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Repository
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
```
sbesson, line 44:
```suggestion
          java-version: 8
```

Due to fundamental compatibility issue in the Java ecosystem, artifacts should always be produced using the minimal Java version.

jburel: I am okay with that but previously in travis we had ``openjdk11``

jburel: I have pushed a commit since the suggestion is incorrect. it should ``1.8``

sbesson: Agreed. That was probably done before we realized compiling with higher versions of Java could cause issues. See https://forum.image.sc/t/fiji-wont-open-czi-files-after-update-there-was-a-problem-with-the-class-java/40932/8 for a recent occurence i can think of.

.github/workflows/maven.yml
```diff
@@ -32,3 +32,22 @@ jobs:
...
+    if: contains('refs/heads/develop', github.ref)
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Repository
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+          server-id: ome.snapshots
```
sbesson, line 45: Is this going to work for non-SNAPSHOTs? I expect they should be deployed to `ome.staging`?

.github/workflows/maven.yml
```diff
@@ -32,3 +32,22 @@ jobs:
           restore-keys: ${{ runner.os }}-m2
       - name: Build
         run: mvn ${{ env.maven_commands }}
+  deploy:
```
sbesson, line 35: Main downside of the separate step is probably that the whole lifecycle (checkout, compile, test, install deploy) gets duplicated. But I would say it's mostly an optimization concern and happy to go with something that works for now.

sbesson (commented):
Only concern is that the current setup is likely to fail for non-SNAPSHOT versions. On the other side, builds are currently hanging on Travis and this converts the remaining bit of Travis into a workflow and ideally should preserve the automatic build and deployment of the 6.6 Maven artifacts.

No objection to seeing this merged tomorrow in from my side. But we might need to be ready for some adjustments in the upcoming builds. Additionally, it might be worth testing with a `6.6.0-rc1` workflow to confirm everything works as expected with artifacts being built and uploaded to `ome.staging`.

.github/workflows/maven.yml
jburel: I think I need to revisit that i.e. change the server target if it is a tag.

.github/workflows/maven.yml
jburel: that's the problem. I was debating that point, that it will mean having many ``if:...`` on some several steps

.github/workflows/maven.yml
sbesson: Main downside of this approach is that in our current workflow there will be two events associated with a commit with a non-SNAPSHOT version: the push and the tag. If we handle the tag, the push event will still fail I think when deploying to `ome.snapshots`.

https://github.com/actions/setup-java/pull/86 might be useful in our case but it's not integrated yet.

jburel: When a commit is pushed to develop,  I extract the version out of the pom.xml.
If the version contains ``SNAPSHOT`` -> ``ome.snapshots`` is used,  otherwise``ome.staging`` will be used.

jburel:
With the last commit:
branch name ``develop`` on my fork, https://github.com/jburel/bioformats/actions/runs/414839434
Screenshot 2020-12-11 at 07 26 57

jburel:
https://github.com/jburel/bioformats/actions/runs/414967940 This is a run testing rc1 push to develop
Screenshot 2020-12-11 at 09 05 53

sbesson (approved):
The changes faithfully converts the current Travis deployment set-up i.e. allowing to deploy the artifacts built on every commit to the `develop` branch either to `ome.snapshots` or to `ome.staging`. 

Organization-wide `CI_DEPLOY_USER/CI_DEPLOY_PASS` secrets have been configured with a dedicated `github-actions` agent allowing the deployment to the OME artifactory. Merging to start testing the integration.

A few comments:
- while the SNAPHSOT deployment should be thoroughly tested, we might want to pre-test the non-SNAPSHOT deployment ahead of the next release of Bio-Formats /cc @dgault 
-  `CI_DEPLOY_USER/CI_DEPLOY_PASS` might need to be reviewed when we start deploying to Sonatype via GitHub actions
- as mentioned by @jburel, the current deployment workflow expects a commit bumping the version to trigger a deployment. An alternative workflow used in bio-formats-example is to deploy on tags. In the case of this repository where every commit triggers a deployment, this would require either extra conditional logic to prevent two actions (one on push and one on tag) to try and deploy the same version to `ome.staging`.

PR #3652, opened 2020-12-15
MetaXpress/CellWorX: ignore case when checking for thumbnail paths

melissalinkert:
Backported from a private PR.  Handles the case when thumbnail files are named ```*_Thumb.TIF```.

Tests should continue to pass, and should be safe for a patch release.


dgault:
All builds and tests have been passing with this PR included. The change looks good, merging for inclusion with 6.6.1

PR #3663, opened 2021-02-14
CV7000: fix channel-wise ordering of planes

melissalinkert:
See ```inbox/idr0093/020/```, discussed with @sbesson.  This seems to fix the order of planes returned by ```openBytes``` without affecting channel ordering in the OME-XML.  No idea what if any impact this will have on other CV7000 data though, so may need further adjustment if tests fail.

Note also that the affected dataset actually has 5 channels of TIFF files, but only 4 defined in metadata.  The fifth channel will be ignored completely; the corresponding files will not be preserved on import.  I don't think the reader can realistically do anything about that.

Discussed with @sbesson that the memo file impact here is a little unpleasant.  Memo files won't be invalidated, but trying to use this change with an old memo file means that the fix just won't appear.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -379,7 +379,7 @@ public class CV7000Reader extends FormatReader {
       planeLengths[2] = core.get(p.series).sizeT;
 
       p.no = FormatTools.positionToRaster(planeLengths,
-        new int[] {p.channelIndex - m.minC, p.z - m.minZ, p.timepoint - m.minT});
+        new int[] {p.channel - m.minC, p.z - m.minZ, p.timepoint - m.minT});
       if (reversePlaneLookup[p.series][p.no] < 0) {
         reversePlaneLookup[p.series][p.no] = i;
       }
@@ -612,6 +612,7 @@ public class CV7000Reader extends FormatReader {
 
   private Plane lookupPlane(int series, int no) {
     int index = reversePlaneLookup[series][no];
+    LOGGER.trace("lookupPlane(series={}, no={}), index = {}", series, no, index);
     if (index < 0 || index >= planeData.size()) {
       return null;
     }
@@ -720,6 +721,8 @@ public class CV7000Reader extends FormatReader {
           }
 
           p.channelIndex = channelMap.get(p.channel);
+          LOGGER.trace("p.channel = {}, p.channelIndex = {}", p.channel, p.channelIndex);
+          LOGGER.trace("    channelMap = {}", channelMap);
 
           p.xpos = DataTools.parseDouble(attributes.getValue("bts:X"));
           p.ypos = DataTools.parseDouble(attributes.getValue("bts:Y"));
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -28,6 +28,7 @@ package loci.formats.in;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -91,6 +92,7 @@ public class CV7000Reader extends FormatReader {
   private int fields;
   private String startTime, endTime;
   private ArrayList<String> extraFiles;
+  private Map<Field, Integer[]> fieldChannels = new HashMap<Field, Integer[]>();
 
   // -- Constructor --
 
@@ -206,6 +208,7 @@ public class CV7000Reader extends FormatReader {
       endTime = null;
       reversePlaneLookup = null;
       extraFiles = null;
+      fieldChannels.clear();
     }
   }
 
@@ -271,6 +274,13 @@ public class CV7000Reader extends FormatReader {
       if (settingsPath != null) {
         settingsPath = new Location(parent, settingsPath).getAbsolutePath();
       }
+
+      channels.sort(new Comparator<Channel>() {
+        @Override
+        public int compare(Channel c1, Channel c2) {
+          return c1.index - c2.index;
+        }
+      });
     }
 
     if (settingsPath != null && new Location(settingsPath).exists()) {
@@ -296,7 +306,7 @@ public class CV7000Reader extends FormatReader {
 
     for (Plane p : planeData) {
       if (p != null) {
-        int wellIndex = p.row * plate.getPlateColumns() + p.column;
+        int wellIndex = p.field.row * plate.getPlateColumns() + p.field.column;
         if (!minMax.containsKey(wellIndex)) {
           minMax.put(wellIndex, new MinMax());
         }
@@ -325,8 +335,8 @@ public class CV7000Reader extends FormatReader {
           m.minC = p.channelIndex;
         }
 
-        if (p.field >= fields) {
-          fields = p.field + 1;
+        if (p.field.field >= fields) {
+          fields = p.field.field + 1;
         }
 
         uniqueWells.add(wellIndex);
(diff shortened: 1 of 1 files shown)


sbesson:
Thanks @melissalinkert. With a5c2865, `setId` failed for one of the representative datasets - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/109243/console. The `MeasurementRecord` for this plate points to channels 1, 2, 3, 4 and 9. So I suspect some extra logic will be needed to map these indices to `0,1,2,3,4`, which was the primary goal of `channelIndex` as far as I understand from the reader.

The review of the reader change on against representative plates from `idr0093` where the original issue was discovered looks promising so far.

melissalinkert:
Re-reading the ```nih``` plate story (especially #3278), my impression is that we didn't have a clear idea on what the correct plane ordering was supposed to be.  It ended up being the natural ordering of planes within the ```MeasurementData.mlf```, with the channel indexing logic meant to address the fact that there are gaps between channel indexes in the same field, and different channels are acquired across different fields.

eb705ce sorts so that ```bts:Ch``` is the authority on how planes are ordered, which is what I thought this reader was doing in the first place.  That shouldn't change the behavior for ```idr0093/020``` compared to a5c2865, but will change the plane ordering for the ```nih``` plate (config PR forthcoming).  If anyone feels that the original ordering for the ```nih``` plate was indeed correct (i.e. this is breaking), then definitely happy to discuss - I'm not 100% confident in how that plate should be handled.  It's also worth checking that the channel metadata reported by ```showinf -nopix -omexml``` matches both the submitter expectations and current plane ordering.

eb705ce is a more substantial change, and will now invalidate memo files.  I would expect that this is not safe for a patch release.

sbesson:
Completely agreed that the `nih` plate is pretty challenging. On the bright side, its complexity might foruce us to increase the robustness of the reader logic and adapt to the complex acquisitions scenarios.

Writing down a few comments from my evening investigation of the various metadata files associated with our sample CV7000 filesets. There are a few files of interest:

- `MeasurementData.mlf` contains a list of `bts:MeasurementRecord` elements which are largely equivalent to `OME:Plane` as well as the Plane/TIFF mapping. These records are ordered by acquisition time and contain a reference to the associated dimension (row, column, z-index...).

   ```xml
  <bts:MeasurementData bts:Version="1.0" xmlns:bts="http://www.yokogawa.co.jp/BTS/BTSSchema/1.0">
  <bts:MeasurementRecord bts:Type="IMG" bts:Time="2018-05-21T13:49:39.748-04:00" bts:Column="6" bts:Row="2" bts:TimePoint="1" bts:FieldIndex="1" bts:ZIndex="1" bts:TimelineIndex="1" bts:ActionIndex="1" bts:Action="BF3D" bts:X="-500.0" bts:Y="500.0" bts:Z="0.0" bts:Ch="9">Test-plate-1_B06_T0001F001L01A01Z01C09.tif</bts:MeasurementRecord>
   <bts:MeasurementRecord bts:Type="IMG" bts:Time="2018-05-21T13:49:39.809-04:00" bts:Column="6" bts:Row="2" bts:TimePoint="1" bts:FieldIndex="1" bts:ZIndex="2" bts:TimelineIndex="1" bts:ActionIndex="1" bts:Action="BF3D" bts:X="-500.0" bts:Y="500.0" bts:Z="1.0" bts:Ch="9">Test-plate-1_B06_T0001F001L01A01Z02C09.tif</bts:MeasurementRecord>
   ```

- `MeasurementDetail.mrf` contains high-level characteristics for the plate, a pointer to a secondary file containing the measurement settings, a pointer to the plate metadata as well as various `MeasurementChannel` elements. These `MeasurementChannel` elements are currently parsed by the reader to populate some channel metadata including the physical pixel sizes. Note however this looks like metadata associated with  correction images for each of the channels.
- the measurement settings file referenced by `MeasurementDetail.mrf` is the primary source of instrument metadata
  - `bts:LightSourceList` is a list of `bts:LightSource` objects which are analogous to `OME:Laser`
  - `bts:ChannelList` is a list of `bts:Channel` objects which naturally map to `OME:Channel` and refers to the `bts:LightSource` element by name

  ```xml
    <bts:ChannelList>
    <bts:Channel bts:Ch="1" bts:Target="Microscope Image" bts:ObjectiveID="40000" bts:Objective="40x" bts:Magnification="40" bts:MethodID="1" bts:Method="Confocal Fluorescence 405/488/561/640 nm" bts:FilterID="1" bts:Acquisition="BP445/45" bts:ExposureTime="50" bts:Binning="2" bts:Color="#FF002FFF" bts:MinLevel="0" bts:MaxLevel="0.949691661150897" bts:Kind="ConfocalFluorescence" bts:AndorParameterID="32" bts:AndorParameter="Gain 1,4" bts:CameraType="Andor" bts:InputLevel="2000" bts:Fluorophore="">
      <bts:LightSourceName>405nm</bts:LightSourceName>
      <bts:LightSourceName>561nm</bts:LightSourceName>
    </bts:Channel>
    <bts:Channel bts:Ch="2" bts:Target="Microscope Image" bts:ObjectiveID="40000" bts:Objective="40x" bts:Magnification="40" bts:MethodID="1" bts:Method="Confocal Fluorescence 405/488/561/640 nm" bts:FilterID="2" bts:Acquisition="BP525/50" bts:ExposureTime="20" bts:Binning="2" bts:Color="#FF00FFA1" bts:MinLevel="0" bts:MaxLevel="0.98993833223017935" bts:Kind="ConfocalFluorescence" bts:AndorParameterID="32" bts:AndorParameter="Gain 1,4" bts:CameraType="Andor" bts:InputLevel="10000" bts:Fluorophore="">
  ...
  ```

  - `bts:Timelapse` is a list of `bts:Timeline` elements which contain a `bts:WellSequence` object i.e. a list of wells and a `bts:ActionList` element which is roughly a subset of channels from `bts:ChannelList` in a structure reflecting the acquisition process e.g.
    
  ```xml
      <bts:ActionList bts:RunMode="sequential">
        <bts:ActionAcquireBF3D bts:XOffset="0" bts:YOffset="0" bts:AFShiftBase="4" bts:TopDistance="4" bts:BottomDistance="-4" bts:SliceLength="1">
          <bts:Ch>9</bts:Ch>
        </bts:ActionAcquireBF3D>
        <bts:ActionAcquire3D bts:XOffset="0" bts:YOffset="0" bts:AFShiftBase="-4" bts:TopDistance="4" bts:BottomDistance="-4" bts:SliceLength="1" bts:UseSoftFocus="false">
          <bts:Ch>1</bts:Ch>
          <bts:Ch>3</bts:Ch>
        </bts:ActionAcquire3D>
        <bts:ActionAcquire3D bts:XOffset="0" bts:YOffset="0" bts:AFShiftBase="-4" bts:TopDistance="4" bts:BottomDistance="-4" bts:SliceLength="1" bts:UseSoftFocus="false">
          <bts:Ch>2</bts:Ch>
          <bts:Ch>4</bts:Ch>
        </bts:ActionAcquire3D>
  ```

My primary consideration at this stage is that this `bts:Timelapse` element seems to contains a lot of metadata that is currently recalculated by the internal reader logic especially eb705ce. It is possible that parsing it and extracting the subset of acquisition channels associated with a given well might simplify the plane/channel assignment?

The XML extract above also makes me reconsider whether the channel order for a given OME:Image should be sorted by the channel index e.g. `1, 2, 3, 4, 9` in the sample above or the order of apparition in the `ActionList` object e.g. `9, 1, 3, 2, 4` in the example above.

While reviewing the configuration file, another consideration is that we might want to maintain the original channel index as the channel name  and drop `CameraName` as this is only a concept associated with `MeasurementChannel` i.e. the channels might be `Channel 1`, `Channel 2`, `Channel 3`, `Channel 4` and`Channel 9` in the example above. At least, this might be useful for testing the associated metadata/underlying data.

Leaving these here for initial feedback. Probably more will come as part of the testing. Happy to schedule a discussion to discuss these and make decisions.

melissalinkert:
f13f627 takes the action indexes into account, so channel orders for the ```nih``` plate and ```idr0093/020``` should now match what @sbesson describes above (and what we discussed separately).  https://github.com/openmicroscopy/data_repo_config/pull/483 has been updated accordingly.

A few local tests indicate that memo files are not invalidated across the board, but obviously will still need to be regenerated for CV7000 data that has multiple actions.

sbesson (approved):
This PR has been extensively tested in the context of and IDR submission (`idr0093`). After validation by the original depositor, I can confirm the channel mapping and the plane reading works as discussed in https://github.com/ome/bioformats/pull/3663#issuecomment-782921211 i.e. following the ordering defined in the acquisition file. This should also support the `TimelineIndex` and the `ActionIndex` metadata from these acquisition files.

In terms of deployment, this PR by itself does not invalidate memo files apart from the ones for CV7000 filesets created by previous releases of Bio-Formats due to the implementation changes to the internal structure of the reader. As such it is amenable to a patch release, although we could also use a minor release increment to communicate the potential impact on channel order.

dgault:
Including this with the 6.6.1 patch release as discussed in the formats meeting. A note will be added to the release announcements to communicate the impact to users.

Issue #3282, opened 2018-11-29
NDPIS images with missing wavelengths fail to open


petebankhead:
Some NDPIS images are failing to open, seemingly due to the wavelength value required by [NDPISReader](https://github.com/openmicroscopy/bioformats/blob/f9f8f17d5037dab717cb22507ef9b0d591ed9a8d/components/formats-gpl/src/loci/formats/in/NDPISReader.java#L215) being missing.  There error shown when attempting to open the image with Fiji is
```
java.lang.NullPointerException: Length: Length cannot be constructed with a null value.
 at ome.units.quantity.Length.<init>(Length.java:59)
 at loci.formats.in.NDPISReader.initFile(NDPISReader.java:215)
```

There is more information at https://github.com/qupath/qupath-bioformats-extension/issues/9 and https://groups.google.com/d/msg/qupath-users/AOWezK8-MKE/Go_h9MthBAAJ (with a sample file).

sbesson:
Thanks @petebankhead, looking at the code, the issue is related to this code block 

https://github.com/openmicroscopy/bioformats/blob/9fc607f85b8900be786813296f1eee75cc1ed883/components/formats-gpl/src/loci/formats/in/NDPISReader.java#L212-L215

It looks like what is required is either a check for `null` values of `wavelength` or using the `FormatTools.getEmissionWavelength` API.

The sample file provided with the original user report is definitely valuable for reproducing the issue and adding to our non-regression tests. Can we add the dataset to our [public sample files](https://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/) under CC-BY licenses?

petebankhead:
Excellent, thanks Sebastien.  Regarding the file, it's not mine and so I'm afraid I can't answer.  I'll link to this discussion in Google Groups in case the owner wants to comment.

JohnKPan:
Hi @sbesson I provided the file to @petebankhead. 

Great tools, from both of you. I'm glad that the files were valuable. 

Unfortunately, I asked and was told by my bosses that it's a hard no to adding the file to a public dataset. 

They've actually asked me to request that you delete the files once the bug is fixed. 

sorry,

John

sbesson:
Hi @JohnKPan, no worries at all. We perfectly understand all datasets cannot be shared.

Would you have other sample files exhibiting the isssue that we could keep in our private QA repository i.e. only available to the Bio-Formats team for non-regression testing and the maintenance of Bio-Formats?

[...]

PR #3664, opened 2021-02-16
NDPIS: use transmittance values to pick valid channels if no wavelength

melissalinkert:
Backported from a private PR.  This adds another way to pick the correct band from NDPI files in an NDPIS dataset when the emission wavelengths are missing.

The behavior without this PR was to always pick the first band if the wavelength was not present.  To test, compare ```showinf``` on the ```hamamatsu/svidro``` dataset with and without this PR.  The first two channels in that dataset do not have a defined wavelength.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPISReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPISReader.java
@@ -230,19 +230,21 @@ public class NDPISReader extends FormatReader {
           else if (490 < wavelength && wavelength <= 580) bandUsed[c] = 1;
           else if (580 < wavelength && wavelength <= 780) bandUsed[c] = 0;
         }
-        else {
-          String extraMetadata = ifd.getIFDStringValue(METADATA_TAG);
-          String[] metadataLines = extraMetadata.split("\r\n");
-          for (String line : metadataLines) {
-            if (line.trim().startsWith(";NDP Shading Data")) {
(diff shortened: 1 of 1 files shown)


sbesson:
I assume this might be fixing the issue described in https://github.com/ome/bioformats/issues/3282?

melissalinkert:
Certainly looks like it could; I missed that issue when looking for things that might be related to the original problem.

sbesson (approved):
Using the sample file provided as part of #3282. Without this PR I can confirm the wrong brand is being read for the first two channels leading to a signal near 0:

Screenshot 2021-02-19 at 08 28 25

With this PR included, the data from the correct band is read. The color assignment (FITC/green and DAPI/blue) also makes sense

Screenshot 2021-02-19 at 08 28 32

Code changes look good and make sense. Only other consideration is whether the content from tag 65449 might be worth parsing and adding unconditionally to the global metadata. This should not block the merging of this PR

melissalinkert:
4bba33f adds the key/value pairs from tag 65449 to original metadata.  There is a little bit of restructuring there so that the tag is parsed even if a wavelength is present, but the transmittance values will only be used if the wavelength is missing.

sbesson (approved):
Screenshot 2021-02-20 at 22 49 30

dgault:
Changes look good and all builds and tests have been green. Merging for next patch release

PR #3665, opened 2021-02-17
Leica SCN: change position units from reference frame to nm

melissalinkert:
Existing comment on line 390 indicates that the image position and size values stored in XML are in nm.  This can be verified by comparing the ```view.sizeX``` and ```view.sizeY``` values with the image width, height, and physical pixel size stored in the corresponding IFD. 

A configuration PR is forthcoming, but otherwise tests should continue to pass and this should be safe for a patch release.


dgault:
Builds and tests have been green with the PR included. Change looks good and makes sense, testing with sample files from the repo also seems to suggest that nm is correct. For example:

```
sizeX: 36832 
sizeY: 38432
view.sizeX: 18416000
view.sizeY: 19216000
PhysicalSizeY: 0.5µm
PhysicalSizeX: 0.5µm

18416000 / 36832 = 500
19216000 / 38432 = 500

```

PR #3671, opened 2021-03-22
TiffParser: handle Deflate compressed tiles/strips with lsb2msb order

sbesson:
See https://forum.image.sc/t/tiff-file-fails-to-open/50416 for a TIFF file failing to open with Bio-Formats

As suggested by @cgohlke in https://forum.image.sc/t/tiff-file-fails-to-open/50416/3, the issue is that `FillOrder` is lsb2msb and the bytes need to be reordered before decompression.

e0c2e8e proposes to extend the scope of https://github.com/ome/bioformats/pull/3306 to handle TIFF files with Deflate compression (codes 8 and 32946).

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -814,7 +814,9 @@ public class TiffParser implements Closeable {
     // reverse bits in each byte if FillOrder == 2
 
     if (ifd.getIFDIntValue(IFD.FILL_ORDER) == 2 &&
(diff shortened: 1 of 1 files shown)


dgault:
The change here looks sensible and all builds and tests have remained green.

With BF 6.6.1 reading the sample file results in an exception:
```
loci.formats.FormatException: java.util.zip.ZipException: incorrect header check
	at ome.codecs.BaseCodec.decompress(BaseCodec.java:199)
	at loci.formats.codec.WrappedCodec.decompress(WrappedCodec.java:86)
	at loci.formats.codec.ZlibCodec.decompress(ZlibCodec.java:1)
```

With this PR included the file is able to be read without exception and it displays correctly.

PR #3673, opened 2021-03-23
Update POM version numbers to 6.6.1

dgault:
Bumping manually as job is failing (https://ci.openmicroscopy.org/job/BIOFORMATS-DEV-latest-version-bump/)


sbesson:
@dgault looks like the purpose of this job is fairly minimal anyways and we are updating the version numbers by hand in most other components. Should we decommission the job and make this a manual process?

PR #3686, opened 2021-04-21
Operetta: populate channel color using emission wavelength

melissalinkert:
Backported from a private PR.

Similar strategy as OMERO uses by default, but supports 6 different colors using narrower wavelength ranges.  See table on page 1 of https://www.perkinelmer.com/CMSResources/Images/44-6551APP_PhotocellApplicationNotes.pdf; this should be a better match for the colors shown by Columbus.

To test, use any of the ```curated/perkinelmer-operetta``` plates and verify that ```showinf -nopix -omexml``` shows a valid ```Color``` on each ```Channel```.


dgault:
Builds and tests have been unaffected with this PR included. The change looks sensible and the colors match those shown in the pdf, the rgb values are also correct.

Tested using a range of sample files from curated/perkinelmer-operetta and with this PR included the Colors are now being correctly populated onto Channel.

PR #3687, opened 2021-04-21
Zeiss CZI: detect plates

melissalinkert:
Backported from a private PR.

To test, use any of the files with updated configurations from the forthcoming configuration PR.  Most have more than one resolution per field; it's also not uncommon to have an attachment image similar to a slide overview, although it doesn't look like any of the plates in the repo have this.

The test change allows for series to be marked as unlinked from a plate by design, by setting ```Plate = -1``` explicitly.  Omitting ```Plate``` from the configuration or setting ```Plate = 0``` should behave as before.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -855,8 +855,9 @@ public class ZeissCZIReader extends FormatReader {
         angles = 1;
       }
       else {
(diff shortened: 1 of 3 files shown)


dgault:
Builds and tests have remained green with this PR and the associated config PR included.
Tested using each of the datasets from the config PR and compared using the Zen, the updated plate values from the new config look to be correct.

Getting this merged for inclusion in 6.7.0

PR #3688, opened 2021-04-23
Pattern reader: populate channel names from pattern tokens

melissalinkert:
To test, use something like:

```
$ cat test.pattern
c<DAPI,FITC>.fake
$ showinf -nopix -omexml test.pattern
```

and verify that channel names are set to ```DAPI``` and ```FITC```.  This will only work when using a pattern file, not with ```showinf -stitch c<DAPI,FITC>.fake```.  I'm not completely opposed to adding this feature to ```FileStitcher``` instead, it just seems simpler and less breaking in ```FilePatternReader```.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FilePatternReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FilePatternReader.java
@@ -39,6 +39,7 @@ import java.util.List;
 
 import loci.common.DataTools;
 import loci.common.Location;
+import loci.formats.AxisGuesser;
 import loci.formats.ClassList;
 import loci.formats.CoreMetadata;
 import loci.formats.FileStitcher;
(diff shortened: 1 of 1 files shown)


melissalinkert:
@joshmoore mentioned there may be something similar in IDR.  The closest I can find is https://github.com/ome/bioformats/pull/2895, which I think is a bit different in that it allows the channel names to be specified as key/value pairs within the pattern file.

dgault:
No objections to adding this to FilePattern, the changes wont cause any breaking issues. Tested using the recommended fake pattern. Without this PR included it produces 2 channels but without the channel names populated:
```
            <LightPath/>
         </Channel>
         <Channel ID="Channel:0:1" SamplesPerPixel="1">
            <LightPath/>
         </Channel>
```

With the PR included the channel names are now also correctly populated:
```
         <Channel ID="Channel:0:0" Name="DAPI" SamplesPerPixel="1">
            <LightPath/>
         </Channel>
         <Channel ID="Channel:0:1" Name="FITC" SamplesPerPixel="1">
            <LightPath/>
         </Channel>
```

PR #3691, opened 2021-05-05
MATLAB/Octave: relicense m-files under BSD-2 clause

sbesson:
Fixes https://github.com/ome/bioformats/issues/3530

See the issue and related discussion for more historical context. Summarizing briefly, it was agreed there is no requirement for the whole set of M functions used by the Bio-Formats/Octave package to be distributed under the GPL license and this licensing choice is rather historical. In the associated issue, all current and former contributors have given their consent to relicense their contributions under the permissive BSD-2 clause license.

This PR implements this change with the following:

- migrates the functions from `components/formats-gpl` under `components/formats-bsd`
- updates the headers to reflect the new licenses
- updates various places in the build system to use the new location

In terms of distribution,

- this PR does not modify the licensing of the `bfmatlab.zip` package itself since this bundle also distributes `bioformats_package.jar` which includes all the GPL licensed readers
- this PR does modify the licensing of the `bio-formats-octave` bundle to BSD-2 since the installation of the dependencies is separate
- with this change included it should be possible to combine the m-files as well as `format-bsd.jar` and all its dependencies and be able to read/write BSD-licensed readers/writers under a permissive license both under MATLAB and under GNU Octave

One question still remains for the GNU Octave package. Since the package only bundles the M functions, the `.tar.gz` bundle could be also relicensed as BSD-2. However as per the `SystemRequirements` in https://github.com/ome/bioformats/blob/304290bcf9d8edf906793d9e7074d8db47b71b3a/components/formats-gpl/octave/DESCRIPTION.in#L11, it is unclear to me whether this should remain GPL. @carandraug do you have more specific guidance here?
Update (2021-05-05): solved by also migrating the Octave packaging infrastructure under `components/formats-bsd`


sbesson:
Thanks @carandraug for the quick response. I'll work on a a few extra commits to move the Octave packaging  under `formats-bsd` and use the proper licensing.

> Why is it possible for the Matlab package to fallback to formats-bsd.jar and not in Octave?

Which fallback are you referring to?

carandraug:
> [...]
> > Why is it possible for the Matlab package to fallback to formats-bsd.jar and not in Octave?
> 
> Which fallback are you referring to?

You said "with this change included it should be possible to use the MATLAB functions as well as `format-bsd.jar` and all its dependencies and be able to read/write BSD-licensed readers/writers under a permissive license". I believe the same should be possible with the Octave package.

sbesson:
> You said "with this change included it should be possible to use the MATLAB functions as well as format-bsd.jar and all its dependencies and be able to read/write BSD-licensed readers/writers under a permissive license". I believe the same should be possible with the Octave package.

Yes, I  meant to refer to the m-files rather than a specific  environment. I updated the PR description and title to reflect the fact the impact of these changes should be applicable to both MATLAB and GNU Octave.

dgault:
The changes to the licensing appear to be complete and the updated builds look good. Tested using the MATLAB functions with formats-bsd and you are able to read and write as expected. Also tested the Octave distribution with formats-bsd and again it was able to read and write just fine with bsd licensed readers/writers. As all contributors have signed off on the relicensing this is good to merge for inclusion in 6.7.0

mkitti:
Thank you all.

PR #3694, opened 2021-05-19
Update directory for OME-TIFF filesets with BinaryOnly across multiple folders

sbesson:
See https://forum.image.sc/t/opening-images-from-a-yokogawa-cq1-with-bio-formats/52615/

This PR deals with the case of a multi-file OME-TIFF fileset distributed over multiple folders, where the master file containing the metadata is up one directory from the individual files referring to it using the `BinaryOnly.MetadataFile` element.

53621b2 adjust the current directory to use the directory of the metadata file. This allows to compute all the  paths of the TiffData elements relative to this directory.

bd13a7c  also uses `getCanonicalPath()` to resolve the relative paths stored in `BinaryOnly.MetadataFile`

7c28bcc sets the `currentId` to the master metadata file. This is not a rquirement for the fix and can be reviewed/discussed separately if problematic. 

A sample fileset can be downloaded from the [Zenodo record](https://zenodo.org/record/4751185) and is also available under `/uod/idr/scratch/inbox/imagesc-52615`. Without this PR, `showinf` should work when pointing to the master OME-TIFF file (`MeasurementResultMIP.ome.tif`) but fail when pointing at one of the TIFF under the `Projection` folder with a `Mismatch UUID` error. With this PR, the fileset should be detected independently of the TIFF chosen. Additionally the list the used files should always start with the master metadata file and be canonical paths (no `../`).

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -507,6 +507,8 @@ public class OMETiffReader extends SubResolutionFormatReader {
           meta = service.createOMEXMLMetadata(xml);
           // Compute all paths relative to the directory of the metadata file
           dir = path.getParentFile().getCanonicalPath();
+          // Set the current ID to the metadata file
+          currentId = metadataFile;
         }
         catch (ServiceException se) {
           throw new FormatException(se);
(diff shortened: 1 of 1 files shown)


dgault:
Overall looks fine to me, no issue with using the metadata file for currentId in this scenario. Will be good to merge once the repo tests are green (failures are unrelated to this PR).

dgault:
Builds and tests have remained green with this PR included. 

Tested with the original sample fileset `imagesc-52615`. Without this PR opening the `ome.tiff` or `ome.xml` file sees the full dataset read correctly with 309 series available, the images open and display as expected. Opening one of the single tiff files however results in the exception as reported:
`loci.formats.FormatException: Unmatched UUID: urn:uuid:1b973f47-3036-416c-9f3c-6e64ddd0dd58`

With this PR included opening the `ome.tiff` or `ome.xml` results in unchanged behaviour, but opening one of the tiff files now sees the full dataset being read and displayed without exception.

PR #3695, opened 2021-05-20
200809-Transforms: Add tests for custom attributes

dgault:
This is a follow up to https://github.com/ome/ome-model/issues/142

From the 2003-FC spec the CustomAttributes can be positioned in 4 different places. The CustomAttributes which are not at the top level must be copied there before transforming to XMLAnnotations. 

These tests should fail until a transform fix has been included with OME-Model.


sbesson (approved):
Apart from the mismatching indentation in the XML file, the additions to the test file make sense and cover the main objects annotated with `CustomAttributes` (`OME`, `Datasets`, `Image`).

The CI build  confirms the transformation are lossy without the ome-model fix (https://github.com/ome/bioformats/runs/2637611507) and that the attributes are properly converted with the ome-model fix (https://github.com/ome/bioformats/pull/3695/checks).

Issue #3699, opened 2021-05-27
Use the cached XML schema for improved validation


sbesson:
_Originally posted by @sbesson in https://github.com/ome/bioformats/issues/3697#issuecomment-847306437_

The OME-XML validation calls used by certain readers and the utilities try to retrieve the upstream XML schema from the W3C website which is throttling such calls - see https://www.w3.org/Help/Webmaster#slowdtd. For libraries, the accepted recommendation is to cache these schemas and use catalog files.

The ome-model repository already includes a [copy of the XML schema](https://github.com/ome/ome-model/blob/v6.2.2/specification/src/main/resources/released-schema/external/xml.xsd) as well as [a catalog file](https://github.com/ome/ome-model/blob/master/specification/src/main/resources/released-schema/external/catalog.xml) and these resources are bundled as part of the `org.openmicroscopy:specification` JAR but it looks like these are not being used. 

Being able to use the internal cache file should greatly improve the experience and performance when it comes to validating OME-XML fragments.

ngladitz:
Don't quite know what I am doing here or if it is generic enough or how it is intended exactly but this does seem to work:
```patch
diff --git a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
index 2ba3a7069a..a20d7047a5 100644
--- a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
+++ b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
@@ -174,6 +174,8 @@ public class OMEXMLServiceImpl extends AbstractService implements OMEXMLService
             /* from specification.jar */
             return getClass().getResourceAsStream("/released-schema/" +
                  matcher.group(1) + "/" + matcher.group(2) + ".xsd");
+          } else if(url.equals("http://www.w3.org/2001/xml.xsd")) {
+            return getClass().getResourceAsStream("/released-schema/external/xml.xsd");
           } else {
             return null;
           }
```

Got me the sameish speed improvement as removing validation in the OBF reader for my small test file.
This may indicate that removing validation (especially if it turns out to be problematic after all) can perhaps be omitted but I've still got to test this with something more bulky.

dgault:
Closing this issue as the PR https://github.com/ome/bioformats/pull/3701 provides a solution

PR #3701, opened 2021-05-28
OMEXMLService: Use XML catalog in specifications.jar

ngladitz:
Before this change OME schema references for
"http://www.w3.org/2001/xml.xsd" lead to throttled downloads.

After this change an embedded copy from specifications.jar
should be used instead.

This seems to be a more generic solution than what I proposed for #3699 earlier and it does seem to work as well (maybe a tiny bit slower due to the additional catalog processing but still much faster than the initial throttled download).

Again I only half know what I am doing here so I'd appreciate a thorough review of the changes.

### DIFF
--- a/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
+++ b/components/formats-api/src/loci/formats/services/OMEXMLServiceImpl.java
@@ -174,6 +174,9 @@ public class OMEXMLServiceImpl extends AbstractService implements OMEXMLService
             /* from specification.jar */
             return getClass().getResourceAsStream("/released-schema/" +
                  matcher.group(1) + "/" + matcher.group(2) + ".xsd");
+          } else if(url.equals("http://www.w3.org/2001/xml.xsd")) {
+            return getClass().getResourceAsStream(
+              "/released-schema/external/xml.xsd");
           } else {
             return null;
           }


dgault:
Thanks @ngladitz, I have included this PR for testing with the nightly data repository

dgault:
@ngladitz, like many other projects, we require contributors to sign our Contributor License Agreement (CLA). 

For legal reasons we need all the people listed to sign the CLA before we can review and merge your contribution. Please download the [CLA form](https://drive.google.com/file/d/1ld6s678JWRF15qUT4_3LyTE4IGYh-wRZ/view), fill and sign it and return it by email to contributors@openmicroscopy.org.
Thanks again for the contribution, we look forward to reviewing it!

ngladitz:
@dgault in case this isn't covered by internal communication ....
@sbesson confirmed retrieval of my signed CLA on June 1st.

sbesson (approved):
I will leave @dgault to review and confirm on return but from my experience testing other ome-model/bio-formats improvements with this PR included, running `showinf -nopix -omexml` show a massive performance improvement effectively validating the OME-XML instantaneously

dgault:
Tested across OME-TIFF, OME-XML and OBF, across the board the performance improvement was pretty consistent with over 15 seconds saved from initialisation each time with the PR included. This is very much in line with the times expected without the PR and validation turned off, though I tested with some invalid elements to verify that validation is still taking place. 

The reason this was likely more pronounced in OBF is due to that particular reader not using the metadata option to turn validation on or off so it was validating every time regardless of the option. Modifying the OBF reader to use `getMetadataOptions().isValidate()` is something which can happen as a follow up to this, though it provides little performance improvement compared to this PR.

Overall this looks like a great improvement, thanks again @ngladitz for your work contributing this PR, we will get this included in the next 6.7.0 release.

dgault (approved):

PR #3707, opened 2021-06-16
Deltavision: fix IndexOutOfBoundsException

sbesson:
This PR tries to address a regression in the Deltavision reader introduced in Bio-Formats 6.2.0.
Initial support for multi-stage positions was introduced in https://github.com/ome/bioformats/commit/1e6686286760bbac249a64c05e42b5af6ec5a8a2. https://github.com/ome/bioformats/pull/1324 introduced some special handling of cases where some coordinates are equal to 0 and https://github.com/ome/bioformats/pull/3398 added support for detecting backward stages in X.

A representative sample files was uploaded as part of an IDR submission where `uniqueTileX.size()` is 1, `uniqueTileY.size()` is 3 and `hasZeroX` is true. As per the following block

https://github.com/ome/bioformats/blob/b68a64959b9f17ceb6b9cb57c15d7b46b56f23ed/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java#L506-L516

`xTiles` is incremented by one and causes the following block to fail when accessing the second element:

https://github.com/ome/bioformats/blob/b68a64959b9f17ceb6b9cb57c15d7b46b56f23ed/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java#L518-L520

The PR updates the coordinates detection to use `Constants.EPSILON` and round small values to zero.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import loci.common.Constants;
 import loci.common.DataTools;
 import loci.common.DateTools;
 import loci.common.Location;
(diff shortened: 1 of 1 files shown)


melissalinkert:
I think the root cause of this problem is that all of the positions should be detected as zero (i.e. ```uniqueTileX``` should be empty), but are not because of the incorrect floating point math in lines 489-501.  Had you already tried changing the checks to something like:

```
      if (!uniqueTileX.contains(hdr.stageXCoord) &&
              Math.abs(hdr.stageXCoord.value().floatValue()) > Constants.EPSILON) {
        uniqueTileX.add(hdr.stageXCoord);
      }
      else if (Math.abs(hdr.stageXCoord.value().floatValue()) <= Constants.EPSILON) {
```

sbesson:
:+1: locally the proposed change suffices to open the representative file. Pushed 03e91d1 and I will let the CI jobs validate it against the curated repository.

dgault:
With the latest commit the updated logic looks sensible to me and a better approach overall. Looking at the position values it does appear that they should all be zero with some rounding errors in there. PR hasn't caused any failures or regressions with existing datasets in the repo tests and the new IDR sample file is read correctly without the Exception.

sbesson:
Thanks @dgault.

PR #3700, opened 2021-05-27
Relicense JHDF service and CellH5 readers under BSD-2

sbesson:
Fixes #3696

- migrates  the `JHDFService` interface and its implementation as well as the `CellH5` reader/writer classes under `formats-bsd` and updates the headers of these classes 
- moves the `jhdf5` dependency under `formats-bsd`
- migrates the `JHDFServiceTest` and `MissingJHDFServiceTest` tests, adds `JHDFServiceTest` to the default testng suite and migrate the `test-no-jhdf` target under `formats-bsd`
- adds a `test-no-jhdf` execution to the Maven test phase of `formats-bsd` to cover the `MissingJHDFServiceTest` as part of the Maven build

The CI tests (Ant & Maven) as well as the repository tests should be unaffected by this relicensing and the impact on downstream users should be minimal as these packages are moved to an upstream dependency of `formats-gpl`.

PR #3712, opened 2021-07-13
Relicense BDV and KLB readers under BSD-2 clause

sbesson:
Support was added for both open file formats to the Bio-Formats code base within the context of the academic [BB/P027032/1](https://gtr.ukri.org/projects?ref=BB%2FP027032%2F1) grant. The original introduction under `formats-gpl` was done for historical reasons rather than a real desire to license those under the GPL.

This PR proposes to relicense these readers under the more permissive BSD-2 license in the same spirit as #3700.


dgault:
Both CI and repo tests have been unaffected by these changes. This will have minimal impact on upstream users and the relicense is suitable for a minor release. Given that all contributors have agreed to the relicense this is good to merge for inclusion in 6.7.0

PR #3718, opened 2021-07-26
Add Olympus omp2info reader

melissalinkert:
Backported from private PRs.  Support for this format was added in partnership with OLYMPUS EUROPA SE & Co. KG.

The .omp2info file is XML that allows .oir files to be stitched together.  Opening/importing an .omp2info should automatically detect and stitch the .oir files, but opening/importing an .oir file will not attempt to detect an .omp2info.  The documentation page may need to warn about importing a whole directory of data in these formats in OMERO.

Datasets are in ```inbox/olympus-new``` and configuration is forthcoming.  Some metadata parsing needed to be added to ```OIRReader```; the dataset that uses the new timestamp logic is in ```2021-07-19```.  This ImageJ plugin can be used to verify correct behavior:

https://imagej.net/formats/olympus

### DIFF
--- a/components/formats-api/src/loci/formats/readers.txt
+++ b/components/formats-api/src/loci/formats/readers.txt
@@ -102,6 +102,7 @@ loci.formats.in.SPEReader             # spe
 loci.formats.in.OIRReader             # oir
 loci.formats.in.KLBReader             # klb
 loci.formats.in.MicroCTReader         # vff
+loci.formats.in.OlympusTileReader     # omp2info
 
 # multi-extension messes
 loci.formats.in.JEOLReader            # dat, img, par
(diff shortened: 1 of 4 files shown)


melissalinkert:

dgault:
Builds have been green with this PR included and no regressions in existing datasets.

Tested using each of the new datasets. Without this PR the individual `OIR` files can be opened and displayed but the `omp2info` files are not recognised and result in an `UnknownFormatException`. With this PR the `omp2info` files are now opened and display the stitched image, this matched the stitched OIR file and looks correct compared to the individual tiles and Olympus Viewer. 

For the 2021-07-19 dataset, without this PR the file can be opened and displayed but there is no timestamps present. With this PR the file opens and displays but this time the timestamps are read. I tried to verify these using the Olympus Viewer but it didn't appear to contain the timestamp data either. The values did look to be sensible and consistent though.

PR #3719, opened 2021-07-30
Gatan: Strip null chars from value string

dgault:
This issue was raised on forum thread: https://forum.image.sc/t/gatan-dm4-files-wont-open/55627/3

The issue was related to the parsing of metadata values. In this case some of the position where as below:

- -199.355 (This String contains extra null 0 bytes in between each char)
- -199.355 (The same String with the null bytes removed parses without exception)

A sample file was provided which I will include in the data repo and a config PR is to follow before review


sbesson (approved):
Tested with the sample files provided by the original submitter. Without this change, `setId` errors as described in the thread. With this change included, the metadata is correctly detected and the `xPos (µm)` is populated in the global metadata table.

PR #3721, opened 2021-08-09
Fix case where JPEGTurboService is used to read single-tile images

melissalinkert:
Backported from a private PR.  Had hoped this might fix https://trello.com/c/pVIlQ3Xe/327-jpeg-ioexception, but it only changes the error message.

Very low priority and should be fine for a patch release.


dgault:
A couple of failures in the repo tests: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

melissalinkert:
Closing since this does more harm than good, and I don't have a good plan for fixing the test failures.

PR #3726, opened 2021-09-07
ND2 metadata parsing fixes

melissalinkert:
Backported from a private PR.

For a private test file, the original metadata table contained some bad data e.g. ```Data: ��D�������;����;?��D����```.  The changes here fixed it so that the original metadata table no longer contains invalid characters, and expected key/value pairs are present.

I wouldn't expect any test failures or configuration changes, but there is a possibility that both commits here are specific to newer versions of ND2 and/or NIS Elements and will need to be adjusted to handle older files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -747,18 +747,48 @@ public class NativeND2Reader extends SubResolutionFormatReader {
           long startFP = in.getFilePointer();
           in.seek(startFP - 1);
 
+          // text block can contain XML (which may be cut off)
+          // or sequence of string objects
+
           int typeByte = in.read();
-          int charCount = in.read();
-          textStrings.add(DataTools.stripString(in.readString(charCount * 2)));
-          int numTextInfos = in.readInt();
-          long remainingBytes = in.readLong();
 
-          iterateIn(in, startFP + dataLength - 1);
-          validDimensions.add(blockCount > 2);
+          // 11 is the max object type code (see iterateIn case statement)
+          // don't look for '<' because old ND2s might have
+          // cut off or incorrectly aligned XML
 
-          // make sure the file pointer is at the end of the block
-          in.seek(startFP + dataLength - 1);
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
```diff
@@ -747,13 +747,14 @@ else if (blockType.startsWith("ImageText")) {
...
+          int typeByte = in.read();
+          int charCount = in.read();
+          textStrings.add(DataTools.stripString(in.readString(charCount * 2)));
+          int numTextInfos = in.readInt();
+          long remainingBytes = in.readLong();
 
-          if (!textString.startsWith("<")) {
-            skip = 0;
-          }
+          iterateIn(in, startFP + dataLength - 1);
```
dgault, line 756: PR has caused some failures in the repo tests: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

I think dataLength here might need updated based on the values already read (so  - 14 + (charCount * 2))

It also seems that the value of skip might need reset to 0 as before

melissalinkert: c4a1a39 seems to fix this when testing locally, but I'll check the repo test tomorrow.  Certainly feel free to exclude if too disruptive right now.

dgault:
Builds and tests have been green with this PR included for a while now.


I was able to find a sample file with similar issues. Without the PR the global metadata has scrambled data with invalid chars.
```
baOpticalPathsCorrections	CLxOpticalPathsCorrectionTable�wstrSystemUUIDdb3cff94e6db465699f4289fc8d08fddvecIdsSizematRows��L
baScanArea	dZoom�*7O�?dAngle	dXOffset���L�x�?	dYOffset��G�A��?dScanAreaAspect?�����?bIsCrop
dCropZoom�?dCropXOffsetdCropYOffsetdCropAspect�?uiPointCoint
```

With the PR included all of the original valid values are still present and the mangled values are appearing as correct key value pairs.

PR #3727, opened 2021-09-10
Zeiss CZI: add support for zstd compression

melissalinkert:
Ported from a private PR.  Adds support for both zstd variants in Zeiss CZI, using https://github.com/airlift/aircompressor to do the actual decompression.

Test data for ZSTD0, ZSTD1, and corresponding uncompressed data is being uploaded to ```inbox/zeiss-czi-zstd```.  Compressed data should mostly match the corresponding uncompressed data (comparing MD5s/pixel values).  The ```/LLS*``` datasets are an exception due to how they were exported, but should match visually despite the pixel value differences.

I'd expect this to require a minor release since there is a new dependency.

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -130,6 +130,11 @@
       <artifactId>kryo</artifactId>
       <version>${kryo.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.airlift</groupId>
+      <artifactId>aircompressor</artifactId>
+      <version>0.18</version>
+    </dependency>
 
     <!-- NB: dependency:analyze has false warning about xml-apis:xml-apis. -->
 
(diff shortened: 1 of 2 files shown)


melissalinkert:

dgault:
This PR has been ggreen in builds and hasn't caused any regressions in existing datasets. 

Manually tested with each of the new sample files. Without this PR each of the uncompressed files was able to be opened and displayed as normal. For each of the compressed files the image would fail to open and an exception occurred:
```
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:373)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at loci.formats.in.ZeissCZIReader.initFile(ZeissCZIReader.java:923)
```

With the PR included each of the compressed files now opens and displays without exception. The metadata and displayed images match those of the uncompressed data. Overall this appears to be behaving as expected and looks good for inclusion in 6.8.0

PR #3728, opened 2021-09-20
AVIWriter: Set fps from Pixels Time Increment

dgault:
This PR is in response to an Issue raised (https://github.com/ome/bioformats/issues/3702)
With the sample file provided you can see the time increment is 0.039 which matches the plane delta T times, however when the file is converted to an AVI the resulting file plays with an fps value of 10 which is the default for all format writers.

With this PR the converted sample file should now playback for ~5secs with the correct fps set.

fixes https://github.com/ome/bioformats/issues/3702

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/AVIWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/AVIWriter.java
@@ -44,7 +44,6 @@ import loci.formats.FormatWriter;
 import loci.formats.meta.MetadataRetrieve;
 import ome.units.UNITS;
 import ome.units.quantity.Time;
-import ome.units.unit.Unit;
 
 /**
  * AVIWriter is the file format writer for AVI files.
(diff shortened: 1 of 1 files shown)


components/formats-bsd/src/loci/formats/out/AVIWriter.java
```diff
@@ -42,6 +42,9 @@
 import loci.formats.FormatTools;
 import loci.formats.FormatWriter;
 import loci.formats.meta.MetadataRetrieve;
+import ome.units.UNITS;
+import ome.units.quantity.Time;
+import ome.units.unit.Unit;
```
melissalinkert, line 47: This import shouldn't be necessary.

melissalinkert (changes requested):
Looks fine to me apart from the minor issue noted in-line.

melissalinkert (approved):

PR #3731, opened 2021-09-23
New Open Source SlideBook Format 7 Reader

nicolapapp:
SlideBook 2021 from Intelligent Imaging Innovations introduces a new open source format for their image storage. The format keeps all the metadata in yaml , while the binary images are in numpy format. SlideBook7Reader.java basically implements the
Bio-Formats API without the need of dll's as our previous Format 6 reader. The suffix of the new format is "sldy" , where y is for yaml.

The reader depends on the org.yaml:snakeyaml so this dependency will have to be added. We have used version 1.6 .


dgault:
Hi @nicolapapp,

Like many other projects, we require contributors to sign our Contributor License Agreement (CLA). 

For legal reasons we need all the people listed to sign the CLA before we can review and merge your contribution. Please download the [CLA form](https://drive.google.com/file/d/1ld6s678JWRF15qUT4_3LyTE4IGYh-wRZ/view), fill and sign it and return it by email to contributors@openmicroscopy.org.

Thanks again for the contribution, I have tagged it for inclusion with our nightly builds and image repository tests, we look forward to reviewing it.

dgault:
Hi @nicolapapp, thank you again for opening the PR and contribing to the project.

For the dependency you will need to add the below to the pom.xml file. 
```
<dependency>
    <groupId>org.yaml</groupId>
    <artifactId>snakeyaml</artifactId>
    <version>1.6</version>
</dependency>
```

The 1.6 version of `snakeyaml` seems to be quite old, would it also be worth testing with a later version to see if that also works?

Also would there be any objection to moving this from formats-gpl to formats bsd to have the reader licensed under a more permissive [BSD-2](https://github.com/ome/bioformats/blob/master/components/formats-bsd/LICENSE.txt) license?

nicolapapp:
The failure was in the testing. Do you attribute it to the new code SlideBook7Reader.java? It downloaded the snakeyaml o, and for what I can see in the tests, the SlideBook7Reader: isThisType returned false as none of the format presented was a sldy

components/formats-bsd/src/loci/formats/in/SlideBook7Reader.java
```diff
@@ -0,0 +1,2703 @@
...
+				}
+			}
+		}
+		return buf;
+	}
+
+	// -- Internal FormatReader API methods --
+	public void close(boolean fileOnly) throws IOException {
+		super.close(fileOnly);
+        mDataLoader.CloseFile();
```
dgault, line 2508: Potential null pointer exception if mDataLoader has not been initialized

dgault:
I think it might be caused by a NullPointerException in the close method. You should be able to run the full build and tests by using the `mvn` command from the top level folder to check if everything passes.

dgault:
Hi @nicolapapp, our apologies for the long delay in properly reviewing this PR. When attempting to include the PR in our builds we see a large number of failures in our test suites. These are tests that we run on a nightly basis against a large repository of files from every different format. The failures we see come from a wide range of different formats : https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/1080/consoleFull

The other formats affected are:
https://docs.openmicroscopy.org/bio-formats/6.9.0/formats/i2i.html
https://docs.openmicroscopy.org/bio-formats/6.9.0/formats/evotecperkinelmer-opera-flex.html
https://docs.openmicroscopy.org/bio-formats/6.9.0/formats/ome-tiff.html
https://docs.openmicroscopy.org/bio-formats/6.9.0/formats/gatan-digital-micrograph-2.html
https://docs.openmicroscopy.org/bio-formats/5.8.2/formats/micro-manager.html
https://docs.openmicroscopy.org/bio-formats/5.8.2/formats/zeiss-lsm.html
https://docs.openmicroscopy.org/bio-formats/5.8.2/formats/becker-hickl-spcimage.html
https://docs.openmicroscopy.org/bio-formats/5.8.2/formats/bd-pathway.html

This is most likely caused by the `isThisType` methods in the Slidebook7Reader also identifying files from a number of different formats. Unfortunately many of the sample files causing the failures are private, but there are some public samples for the below formats:

https://downloads.openmicroscopy.org/images/OME-TIFF/
https://downloads.openmicroscopy.org/images/Micro-Manager/

In order to get this PR in a position to fully review it would be good if you could correct the isThisType functionality so that it only identifies SlideBook 7 files. A good start would be to ensure that the public samples files for OME-TIFF and Micro-Manager above are still able to be opened and displayed without issue.

dgault:
I pushed one last commit to tidy up some of the warnings and adjust the logging levels. With that this PR is in a good state so we will get this merged for release with the next 6.10.0 release which will be out by the end of the month.

Issue #3682, opened 2021-04-07
SVS: handle optional label and macro images


sbesson:
See https://forum.image.sc/t/correct-way-to-convert-svs-slides-for-omero-import/50942/.

As described in the `Digital Slides and Third-Party Data Interchange`, a SVS file may contain a label and/or a macro image. In the majority of the samples, these images are typically tagged with the SubFileType 0x1 and 0x9 although this is not part of the document above. 

While other translational libraries like `openslide` and `tifffile` handle this optional presence, the current version of the SVSReader contains some assumptions that seem to unconditionally expect these ancillary images e.g.

https://github.com/ome/bioformats/blob/25645389e076a7bd0011e04c4dd8982c0f0614ed/components/formats-gpl/src/loci/formats/in/SVSReader.java#L331

https://forum.image.sc/t/correct-way-to-convert-svs-slides-for-omero-import/50942/13 and https://forum.image.sc/t/correct-way-to-convert-svs-slides-for-omero-import/50942/19 describe how to generate a secondary SVS file from public samples illustrating this use case.

From a quick search across our curated QA repository and the OpenSlide public samples, I  found no original SVS file without a label or a macro image but it might be useful to  look further for additional representative samples.

MITEL-UNIUD:
Yesterday I was browsing throught SVSreader.java to find where macro and label were managed, and there are a couple of areas:
- https://github.com/ome/bioformats/blob/25645389e076a7bd0011e04c4dd8982c0f0614ed/components/formats-gpl/src/loci/formats/in/SVSReader.java#L334 (look at the comment in the `else` branch)
- https://github.com/ome/bioformats/blob/25645389e076a7bd0011e04c4dd8982c0f0614ed/components/formats-gpl/src/loci/formats/in/SVSReader.java#L475  (again, in the `else` branch there is the case of pyramid levels without macro/label). 

Regarding the availability of examples, by default both macro and label are saved. However, looking at the manual, there is a Scanscope system property `DoLabelCapture` that sets whether the label is captured or not. As soon as I can go to the scanner place, I will try to acquire an example at least without label.

sbesson:
[...]

PR #3733, opened 2021-09-28
SVS: check for label/macro instead of assuming they exist

melissalinkert:
Backported from a private PR.  Instead of assuming that exactly two extra images (label and macro) exist, this adds a check on the IFD descriptions for any non-pyramid IFDs to determine whether a label and/or macro exist.

I would not expect this to cause any test failures ~or memo file changes~, but it may help with #3682.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SVSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SVSReader.java
@@ -283,6 +283,7 @@ public class SVSReader extends BaseTiffReader {
       if (comment == null) {
         continue;
       }
+      comments[i] = comment;
       String[] lines = comment.split("\n");
       String[] tokens;
       String key, value;
@@ -299,10 +300,10 @@ public class SVSReader extends BaseTiffReader {
               zPosition[index] = DataTools.parseDouble(value);
             }
           }
-          else if (t.toLowerCase().startsWith("label")) {
+          else if (t.toLowerCase().indexOf("label") >= 0) {
             labelIndex = i;
           }
-          else if (t.toLowerCase().startsWith("macro")) {
+          else if (t.toLowerCase().indexOf("macro") >= 0) {
             macroIndex = i;
           }
         }
--- a/components/formats-gpl/src/loci/formats/in/SVSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SVSReader.java
@@ -90,6 +90,7 @@ public class SVSReader extends BaseTiffReader {
   private ArrayList<String> dyeNames = new ArrayList<String>();
 
   private transient Color displayColor = null;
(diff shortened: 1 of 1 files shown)


sbesson:
See https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/146689/. At least one sample file seems to be affected by this change. Looking at the content of the file, it looks like the proposed changes here are legit i.e. with current Bio-Formats, three images are detected (in non-flattened mode) and the image identified as a `label image` is really a pyramidal level of the WSI. With this PR included, the reader only detects a WSI and a macro image. I assume the configuration file should be opened with the corresponding updates.

melissalinkert:

sbesson (commented):
I'll carry out functional testing tomorrow when our testing systems are available again. Overall, the proposed functionality makes sense. 

I found two locations in the reader which seem to still have the hardcoded expectations of  2 extra images:

https://github.com/ome/bioformats/blob/758a5f996748995523c9d469bd4468bea1eeb1f3/components/formats-gpl/src/loci/formats/in/SVSReader.java#L183-L186

https://github.com/ome/bioformats/blob/758a5f996748995523c9d469bd4468bea1eeb1f3/components/formats-gpl/src/loci/formats/in/SVSReader.java#L564

Also should https://github.com/ome/bioformats/blob/758a5f996748995523c9d469bd4468bea1eeb1f3/components/formats-gpl/src/loci/formats/in/SVSReader.java#L506-L517 be updated to use the new detection logic?

melissalinkert:
I think cc1112f should fix everything listed in https://github.com/ome/bioformats/pull/3733#pullrequestreview-807503142. That does now mean that memo files will change, but since this is already scheduled for a minor release that hopefully isn't a problem.

sbesson (approved):
With a handful of configuration changes the repository tests have been passing with this PR included.

Looking specifically at the images for which configuration has been regenerated. Without this PR, the pyramidal resolutions are split into 2 series, with the second one marked as `label image`

Screenshot 2021-11-24 at 12 51 36

With this PR included, all the pyramidal levels are correctly detected as part of the WSI and there is no label image.

Screenshot 2021-11-24 at 12 51 49

We do not have any representative example of SVS file with a label image only and/or without any extra images.  This should not block the merging of this PR which can be safely included in the upcoming Bio-Formats 6.8.0

dgault:
Builds and tests have been green with this PR and associated config included. 

Breaking memo files is fine as this is going to be a minor release. The logic and code looks good and makes sense. 

I retested manually against the 2 datasets that have been impacted, for both of the datasets the label image without the PR should actually be part of the image pyramid. With this PR the image is correctly part of the pyramid and only a macro image is present. Pr looks good to merge for 6.8.0

PR #3736, opened 2021-10-06
OIR: Fix NullPointerException for Laser Data ID

dgault:
This issue was raised on forum thread https://forum.image.sc/t/nullpointerexception-when-opening-oir-files-with-bio-formats/58487 and a sample file was provided which reproduces the exception with Bio-Formats 6.7.0

To test:
- Builds and repo tests should remain green
- Without this PR opening the sample file from the thread should result in `java.lang.NullPointerException`
- With this PR included the file should be opened without exception and the pixel and metadata look sensible

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -1224,7 +1224,7 @@ public class OIRReader extends FormatReader {
           if (laserId != null) {
             for (int l=0; l<lasers.size(); l++) {
               Laser laser = lasers.get(l);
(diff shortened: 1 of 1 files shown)


melissalinkert (approved):

shonkoren:
Wanted to comment that I'm still having the same issue opening OIR files. Not sure if the commit is live or not. Happy to supply information if necessary.

melissalinkert:
@shonkoren : this change was included in 6.8.0, so if you still see problems after updating to 6.8.0 please do let us know the exact error message you see and any other details.

PR #3738, opened 2021-10-22
Improve tile read speeds for some pyramid OME-TIFFs

melissalinkert:
Backported from a private PR.  This addresses the pyramid performance problem discussed with @sbesson.

Any time a SubIFD contains large amounts of non-pixel data (e.g. a full copy of the OME-XML), ```openBytes``` is slowed down substantially.  This changes the SubIFD reading in ```openBytes``` to only request a single SubIFD, and not even try to read the corresponding ```ImageDescription``` tag since it will be ignored anyway.

7cee004 is a minor improvement to reduce the number of tiles that are checked when reading pixel data.  It could be reverted if any objections.

I wouldn't expect this to address other issues with OME-TIFF read performance, but maybe points to IFD size instead of modality as a point for further investigation.

I would not expect tests to fail or memo files to be impacted.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -345,8 +345,11 @@ public class OMETiffReader extends SubResolutionFormatReader {
     try (RandomAccessInputStream s = new RandomAccessInputStream(info[series][no].id, 16)) {
       TiffParser p = new TiffParser(s);
       if (resolution > 0) {
-        IFDList subifds = p.getSubIFDs(ifd);
-        ifd = subifds.get(((OMETiffCoreMetadata)core.get(series, resolution)).subresolutionOffset);
+        p.setDoCaching(false);
+        long offset = ifd.getIFDLongArray(IFD.SUB_IFD)[((OMETiffCoreMetadata)core.get(series, resolution)).subresolutionOffset];
+        ifd = p.getIFD(offset);
+        ifd.remove(IFD.IMAGE_DESCRIPTION);
+        p.fillInIFD(ifd);
       }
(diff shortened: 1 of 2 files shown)


dgault:
The logic here makes sense to me and I don't see it having any negative consequences. Builds and tests have also remained passing for some time now and the performance of the repo tests has remained consistent.

I tried testing some tiled reads with the public sample files and generally speaking the times were either slightly quicker or more or less the same.

https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/sub-resolutions/Fluorescence/

Without the PR:
Resolution 1: 16.656s
Resolution 2: 4.95s

With the PR
Resolution 1: 14.464s
Resolution 2: 4.201s

https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/sub-resolutions/Brightfield/Leica-1/

Without the PR
Resolution 0: 19.827s
Resolution 1: 2.659s
Resolution 2: 0.772s

With the PR
Resolution 0: 15.466s
Resolution 1: 2.741s
Resolution 2: 0.829s

https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/sub-resolutions/EM/

Without the PR
Resolution 1: 36.95s
Resolution 2: 7.041s

With the PR
Resolution 1: 29.681s
Resolution 2: 7.148s

https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/sub-resolutions/Z-stack/

Without the PR
Resolution 0: 26.783s
Resolution 1: 22.652s
Resolution 2: 20.32s

With the PR
Resolution 0: 25.933s
Resolution 1: 21.013s
Resolution 2: 19.685s

PR looks good from my perspective. Merging this for inclusion with 6.9.0

PR #3739, opened 2021-10-27
Reduce memory required to read tiles from large JPEGs

melissalinkert:
Backported from a private PR.

This should reduce the amount of memory needed to read large JPEGs, in particular the contents of .ndpi files.  Without this PR, picking an .ndpi file and running something like:

```
$ export BF_MAX_MEM=100m
$ showinf -crop 24000,12000,1024,1024 file.ndpi
```

would work, but setting a lower value for ```BF_MAX_MEM``` would result in an ```OutOfMemoryError``` (the precise limit depends a bit on the file, but ~100MB is about right from what I have seen).  With this PR, setting ```BF_MAX_MEM``` to a much lower value (on the order of 40-50% of the original limit) should work without throwing an ```OutOfMemoryError``` or displaying an incorrect image.

I would not expect tests to fail or memo files to be impacted.  I would suggest testing with QuPath or similar to verify that the actual tile contents are as expected.

### DIFF
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
@@ -88,7 +88,8 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
   private long sos;
   private long imageDimensions;
 
-  private int tileDim;
+  private int tileWidth;
+  private int tileHeight;
   private int xTiles;
   private int yTiles;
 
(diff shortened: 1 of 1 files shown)


petebankhead:
FWIW I tried this quickly in QuPath and it looks great to me.  The images looked fine and on my computer it reduced the time to read tiles dramatically, from >400 ms to <200 ms for 1024x1024 pixels.

This could make it feasible to convert .ndpi to .ome.tif in QuPath (which was prohibitively slow before, but I could finally do it with the help of this change).

<details>
  <summary>QuPath script to time tile reading</summary>

 ```groovy
def rng = new Random()
int n = 50
int w = 1024
int h = 1024
def server = getCurrentServer()
long startTime = System.currentTimeMillis()
for (int i = 0; i < n; i++) {
    def region = RegionRequest.createInstance(server.getPath(), 1.0,
        rng.nextInt(server.getWidth()-w),
        rng.nextInt(server.getHeight()-h),
        w, h, ImagePlane.getDefaultPlane())
    def img = server.readBufferedImage(region)
}
long endTime = System.currentTimeMillis()
long duration = endTime - startTime
println "Total time: ${duration/1000} s (${duration/n} ms per tile)"
```

</details>

**Comment by dgault** [comment]
The changes in this PR make sense to me and look good for inclusion. With the PR included the builds and tests have remained green and there have been no regressions detected in any existing datasets.

Testing without the PR:

I tried testing the recommended showinf command with ndpi files. Without the PR I had to set the BF_MAX_MEM value to 170m in order to successfully read the selection, otherwise I would receive the out of memory:
```
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
	at org.libjpegturbo.turbojpeg.TJDecompressor.decompress(TJDecompressor.java:338)
```

With the PR included I was able to open the image with BF_MAX_MEM set to 140m.

Similar tests with other samples from https://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/ produced various results with the PR performing better or as good as in each scenario.

PR #3740, opened 2021-11-04
FEI TIFF: fix physical pixel sizes for Phenom data

melissalinkert:
Backported from a private PR.

I don't yet have a file that can be used for testing, but as the title suggests this should fix a problem with pixel sizes not being detected in TIFFs acquired by FEI/Thermo Fisher/? Phenom instruments.  The XML schema for Phenom metadata is slightly different from other FEI TIFFs.

I would not expect this to cause test failures or memo file issues.


dgault:
Builds and tests have remained green with this PR included. I don't have a suitable sample file to test but the changes here look safe and sensible, no regressions have been detected in any of the existing sample files. PR looks good to merge,

PR #3745, opened 2021-11-12
Bump cisd:jhdf5 to version 19.04.0

sbesson:
See https://github.com/scijava/pom-scijava/issues/181

Also explicitly declare common-langs as dependency. It was previously coming as a transitive dependency of cisd:jhdf5:14.12.6 and is used across a few readers.

The `JHDF5Service` consuming this library is in-use for two formats: `CellH5` and `BDV`

### DIFF
--- a/components/formats-bsd/pom.xml
+++ b/components/formats-bsd/pom.xml
@@ -74,6 +74,11 @@
      <version>${kryo.version}</version>
     </dependency>
     <dependency>
+      <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+      <version>2.6</version>
+    </dependency>
+    <dependency>
       <groupId>org.perf4j</groupId>
       <artifactId>perf4j</artifactId>
       <version>0.9.16</version>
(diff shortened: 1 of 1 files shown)


sbesson:
Initial inclusion of this dependency bump in the nightly builds indicaes there is no apparent regression in the repository tests - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/959/console - failures are unrelated and caused by the ongoing DICOM work but the `cellh5` and `bdv` tests were passing. So this is positive on the Bio-Formats front.

With this PR included, the downstream `omero-build` job  is failing with errors which seem to be related to `org.apache:commons-io` - see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-build-build/986/. `cisd:jhdf5:14.12.6` was depending on `commons-io:commons-io:1.4` so I'll exclude this PR for tomorrow's build to confirm the issues are related and we can discuss the next steps.

dgault:
Manually tested a selection of CellH5 and BDV files and all opened and displayed without issue showing no regressions. I also tested the new jar in a standard FIJI environment, the Java-8 site has commons-lang 2.6 and jhdf 14.12.6 but the older jhdf behaves as expected in that environment. I will mention to dependency bump when updating the downstream apps.

PR #3753, opened 2021-12-10
CellSens: Set physicalSizeZ

dgault:
This is an attempt to fix the issues reported in https://forum.image.sc/t/bio-formats-vsi-olympus-cellsens-bug/29879/5 and previously on Trello card https://trello.com/c/v47N2hGC/382-vsi-incorrect-metadata-posz-and-exp-time

Initially this is simply setting the physicalSizeZ based on the zIncrement value which is already being parsed.


sbesson (approved):
Straightforward metadata population change. A handful of VSI datasets in the curated QA repositories were affected and needed to be reconfigured to include the physical size in Z. The newly populated value was validated against screenshots of the filesets in Olympus' OlyVIA by @melissalinkert.

PR #3754, opened 2021-12-12
Tiff performance

petebankhead:
This PR aims to address https://github.com/ome/bioformats/issues/3752 while slightly improving TIFF read/write performance.

The performance impact when using JPEG compression can be much bigger when combined with https://github.com/ome/ome-codecs/pull/19

To test, I converted *CMU-1-Small-Region.svs* and *CMU-1.svs* to JPEG with 2 pyramid levels (to avoid memory errors) with bfconvert, with and without the `-no-sequential` flag on a MacBook Pro (M1). I also converted the larger image with Bio-Formats via QuPath.

The changes here were made after using VisualVM to identify which aspects of conversion were slow. Sampling with VisualVM suggested the modifications helped reduce time spent in the relevant methods, and pushed compression/decompression to be the main bottlenecks (along with writing IFDs and creating BufferedImages) - although I find the real-world impact a bit harder to decipher.

Summary:
* Conversion time with this PR is similar to that with 6.7.0 and 6.8.0, but the compression problem in 6.8.0 goes away, at least in my test
* There appears to be some minor performance improvement in this PR (but it depends on image size, number of tiles - so may be inconsistent across images)
* The codec PR makes a much bigger difference
* Parallel reading/writing in QuPath + these PRs can reduce the time converting the image to <20% of using bfconvert currently - although it's mostly achieved by the parallelisation + reduced tile reading

### CMU-1-Small-Region.svs

```
./bfconvert -overwrite -no-upgrade -series 0 -compression JPEG -tilex 256 -tiley 256 -pyramid-scale 4 -pyramid-resolutions 2 "/path/to/CMU-1-Small-Region.svs" "/path/to/converted.ome.tif"
```

Method                  | Sequential | Time (s) | Valid
------------------------|------------|----------|------
v6.7.0                  | :white_check_mark:          | 5.36     | :white_check_mark:
v6.8.0                  | :white_check_mark:          | 5.31     | :x:
v6.8.0                  | :x:          | 5.59     | :white_check_mark:
Proposed                | :white_check_mark:          | 5.38     | :white_check_mark:
Proposed                | :x:          | 5.58     | :white_check_mark:
**Proposed + Codec PR** | :white_check_mark:          | 3.48     | :white_check_mark:
Proposed + Codec PR     | :x:          | 3.57     | :white_check_mark:

### CMU-1.svs

```
./bfconvert -overwrite -no-upgrade -series 0 -compression JPEG -tilex 512 -tiley 512 -pyramid-scale 4 -pyramid-resolutions 2 "/path/to/CMU-1.svs" "/path/to/converted.ome.tif"
```

Method                  | Sequential | Time (s) |
------------------------|------------|----------|
v6.7.0                  | :white_check_mark:          | 224.6    |
v6.8.0                  | :x:          | 231.6    |
Proposed                | :white_check_mark:          | 216.5    |
Proposed                | :x:          | 227.8    |
**Proposed + Codec PR** | :white_check_mark:          | 120.4    |
Proposed + Codec PR     | :x:         | 133.6    

> Opening the images isn't easy because only 2 pyramid levels results in high memory use.

## Convert with QuPath + Bio-Formats

The image is opened in QuPath's viewer using Bio-Formats, and exported via *File &rarr; Export images... &rarr; OME TIFF*, with the tile size and JPEG compression set in the dialog.

This calls Bio-Formats, but uses multiple threads and readers, along with tile caching and use of pyramid levels, to try to improve performance. The export image has 4 pyramid levels.

#### Using 256 x 256 tiles

Method                                  |  Time (s) |
----------------------------------------|-----------|
QuPath v0.3.0 + Bio-Formats v6.7.0      | 62.1      |
QuPath v0.3.0 + Proposed                | 51.5      |
**QuPath v0.3.0 + Proposed + Codec PR** | 51.3      |

#### Using 512 x 512 tiles

Method                                  |  Time (s) |
----------------------------------------|-----------|
QuPath v0.3.0 + Bio-Formats v6.7.0      | 45.7      |
QuPath v0.3.0 + Proposed                | 44.5      |
**QuPath v0.3.0 + Proposed + Codec PR** | 42.3      |

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/IFD.java
+++ b/components/formats-bsd/src/loci/formats/tiff/IFD.java
@@ -221,7 +221,7 @@ public class IFD extends HashMap<Integer, Object> {
 
   /** Gets the given directory entry value from this IFD. */
   public Object getIFDValue(int tag) {
-    return get(new Integer(tag));
+    return get(Integer.valueOf(tag));
   }
 
   /**
@@ -229,7 +229,7 @@ public class IFD extends HashMap<Integer, Object> {
    * performing some error checking.
    */
   public Object getIFDValue(int tag, Class checkClass) throws FormatException {
-    Object value = get(new Integer(tag));
+    Object value = get(Integer.valueOf(tag));
     if (checkClass != null && value != null && !checkClass.isInstance(value)) {
       // wrap object in array of length 1, if appropriate
       Class cType = checkClass.getComponentType();
@@ -523,8 +523,8 @@ public class IFD extends HashMap<Integer, Object> {
   }
 
   public boolean isTiled() throws FormatException {
-    Object offsets = get(new Integer(STRIP_OFFSETS));
-    Object tileWidth = get(new Integer(TILE_WIDTH));
+    Object offsets = get(Integer.valueOf(STRIP_OFFSETS));
+    Object tileWidth = get(Integer.valueOf(TILE_WIDTH));
     return offsets == null && tileWidth != null;
   }
 
@@ -934,22 +934,22 @@ public class IFD extends HashMap<Integer, Object> {
 
   /** Adds a directory entry to this IFD. */
   public void putIFDValue(int tag, Object value) {
-    put(new Integer(tag), value);
+    put(Integer.valueOf(tag), value);
   }
 
   /** Adds a directory entry of type BYTE to this IFD. */
   public void putIFDValue(int tag, short value) {
-    putIFDValue(tag, new Short(value));
+    putIFDValue(tag, Short.valueOf(value));
   }
 
   /** Adds a directory entry of type SHORT to this IFD. */
   public void putIFDValue(int tag, int value) {
-    putIFDValue(tag, new Integer(value));
+    putIFDValue(tag, Integer.valueOf(value));
   }
 
   /** Adds a directory entry of type LONG to this IFD. */
   public void putIFDValue(int tag, long value) {
-    putIFDValue(tag, new Long(value));
+    putIFDValue(tag, Long.valueOf(value));
   }
 
   // -- Debugging --
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -219,7 +219,7 @@ public class TiffParser implements Closeable {
       return null;
     }
 
-    return new Boolean(littleEndian);
+    return Boolean.valueOf(littleEndian);
   }
 
   /** Returns whether or not the current TIFF file contains BigTIFF data. */
@@ -442,8 +442,8 @@ public class TiffParser implements Closeable {
     IFD ifd = new IFD();
 
     // save little-endian flag to internal LITTLE_ENDIAN tag
-    ifd.put(new Integer(IFD.LITTLE_ENDIAN), new Boolean(in.isLittleEndian()));
-    ifd.put(new Integer(IFD.BIG_TIFF), new Boolean(bigTiff));
+    ifd.put(Integer.valueOf(IFD.LITTLE_ENDIAN), Boolean.valueOf(in.isLittleEndian()));
+    ifd.put(Integer.valueOf(IFD.BIG_TIFF), Boolean.valueOf(bigTiff));
 
     // read in directory entries for this IFD
     LOGGER.trace("getIFD: seeking IFD at {}", offset);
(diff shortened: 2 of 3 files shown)


petebankhead:
I added an extra check using a smaller tile size for *CMU-1.svs*, because my understanding is that performance can degrade a lot if there are a large number of tiles.

```
./bfconvert -overwrite -no-upgrade -series 0 -compression JPEG -tilex 256 -tiley 256 -pyramid-scale 4 -pyramid-resolutions 2 "/path/to/CMU-1.svs" "/path/to/converted.ome.tif"
```

Method                  | Sequential | Time (s) |
------------------------|------------|----------|
v6.7.0                  | Y          | 495.9    |
v6.8.0                  | N          | 635.7    |
Proposed                | Y          | 549.0    |
Proposed                | N          | 587.6    |
**Proposed + Codec PR** | Y          | 225.5    |
Proposed + Codec PR     | N          | 340.9    |


Based on this, I wondered whether [no longer passing an IFD in `ImageConverter`](https://github.com/ome/bioformats/issues/3752#issuecomment-991179536) was causing the conversion in v6.8.0 to be slower with large tile counts - and undoing any benefits from this PR.

To test this, I ran it again using the `ImageConverter` from https://github.com/ome/bioformats/commit/48e14845e42c8350afee616a31ea23a40aa2d6f5#diff-cb592158a6af58e0453459ed7811838aa3ec86efd6809c066020bd14b094c160 and saw things become substantially faster.

Method                  | Sequential | Time (s) |
------------------------|------------|----------|
Proposed + old ImageConverter                | Y          | 414.9    |
**Proposed + old ImageConverter + Codec PR** | Y          | 177.8    |

My initial conclusion was that removing the IFD had slowed things down, and the changes in this PR had speeded them up again (albeit in a different way).

Although, full disclosure: when I ran the conversion (twice) with a tile size of 512 x 512 pixels, I found this PR code to be *slower* than the 512 x 512 px results in the previous post (vs. 216.5s) if the JPEGCodec changers weren't included... and very slightly faster if they were (vs. 120.4s). So I remain a bit confused.

Method                  | Sequential | Time (s) |
------------------------|------------|----------|
Proposed + old ImageConverter (512x512)                | Y          | 266.8, 253.3  |
**Proposed + old ImageConverter + Codec PR** (512x512) | Y          | 119.7, 118.3  |

As with https://github.com/ome/bioformats/pull/3680 I'm finding TIFF performance hard to figure out.
There's a chance I got tangled up in my test scripts and a more systematic check is needed, but hopefully something in here is useful.

dgault:
The changes in the PR all look sensible. I am initially looking at this PR as a fix for the bug when converting with tiled compression. As this was generating invalid files with the 6.8.0 release then having this as a fix would be suitable for an immediate 6.8.1 patch release to address the bug. The performance elements and the codec PR will be tested further for a 6.9.0 release.

Using this PR and testing the scenario from https://github.com/ome/bioformats/issues/3752 with tiles compression. With the PR a valid file is generated and the tile offsets and byte counts match identically to those reported in 6.7.0 or when using the non sequential flag in 6.8.0. Enabling the non-sequential flag with this PR also generates a valid file as expected. 

Testing the bfconvert pathway with a combination of parameters (compression, tiles, crop, pyramids) and using range of sample data from https://downloads.openmicroscopy.org/images/:
https://downloads.openmicroscopy.org/images/ND2/karl/
https://downloads.openmicroscopy.org/images/Olympus-OIR/etienne/
https://downloads.openmicroscopy.org/images/Leica-SCN/openslide/Leica-1/
https://downloads.openmicroscopy.org/images/CellH5/samples/reduced/

Each converted file was valid and matched those converted using 6.7.0

Some quick testing of the performance and it the difference in 6.7.0 and 6.8.0 times is surprising to me but with this PR there does seem to be an improvement. It would be worth spending more time analysing the TiffSaver performance for 6.9.0 to fully understand what is going on, especially between the 6.7.0 and 6.8.0 changes:

| File  | 6.7.0 | 6.8.0 | With PR |
| ------------- | ------------- |------------- |------------- |
| CMU-1-Small-Region | 3.175s  | 3.456s | 3.158s |
| CMU-1 | 523.976s | 479.796s | 377.769s |
| CMU-2 | 1223.51s | 881.015s | 473.068s |
| CMU-3 | 1531.622s | 1370.692s | 614.238s |

petebankhead:
Thanks @dgault that sounds great. Was the JPEG codec PR https://github.com/ome/ome-codecs/pull/19 included here, or not?

I'm curious as to whether in-memory JPEG compression is contributing to the timing results, or if there is room for it to give further improvements at some point.

dgault:
I didn't include the codec PR as part of my initial testing, that was purely with this PR. We will get this merged an included in a 6.8.1 patch release and then follow up with the codec PR for a larger 6.9.0 minor release

PR #3758, opened 2022-01-06
Remove loci_tools bundle

sbesson:
Fixes https://github.com/ome/bioformats/issues/3755

See https://www.openmicroscopy.org/2021/12/13/log4j-issue.html and https://www.openmicroscopy.org/security/advisories/2021-SV4/ for more information


components/bundles/bioformats_package/assembly.xml
```diff
@@ -16,7 +16,6 @@
         <exclude>aopalliance:aopalliance</exclude>
         <exclude>org.aspectj:aspectj*</exclude>
         <exclude>org.slf4j:slf4j-log4j12</exclude>
-        <exclude>log4j:log4j</exclude>
```
melissalinkert, line 19: Does it make sense to keep this exclude with a comment that it explicitly prevents log4j from ever being included through other dependencies?

sbesson: You're correct. That's an exclusion statement not an inclusion and is not even in the right commit. I cannot remember the rationale so I will force push this change away

melissalinkert (approved):

components/bio-formats-plugins/src/loci/plugins/Updater.java
```diff
@@ -55,7 +55,7 @@
   /** Flag indicating whether last operation was canceled. */
   public boolean canceled;
 
-  /** Path to ueber-jar whether bioformats_package.jar or loci_tools.jar. */
+  /** Path to ueber-jar whether bioformats_package.jar. */
```
dgault, line 58:
```suggestion
  /** Path to bioformats_package.jar uber-jar. */
```

dgault (commented):
PR looks good and all instances of loci_tools are covered. This should have no impact on end users as bioformats_package has superseded it.

The only remaining instance of log4j I could find was in a MATLAB logging test: https://github.com/ome/bioformats/blob/25645389e076a7bd0011e04c4dd8982c0f0614ed/components/formats-gpl/test/matlab/TestBfInitLogging.m#L38

sbesson:
> The only remaining instance of log4j I could find was in a MATLAB logging test:

Yes, I also came across this during my review of this repository as well as the documentation in https://github.com/ome/bio-formats-documentation/pull/221.

As I didn't capture it elsewhere, log4j is the Java logging framework shipped within MATLAB itself - see https://uk.mathworks.com/matlabcentral/answers/1610640-apache-log4j-vulnerability-cve-2021-44228-how-does-it-affect-matlab for their statement re log4shell. 

So I think the set of statements in https://docs.openmicroscopy.org/bio-formats/6.8.0/developers/logging.html remains valid i.e.
- Bio-Formats  uses slf4j for logging
- logback is the recommended framework implementation (and effectively the one that in use across all OME products)
- log4j primary usage is in the MATLAB environment

dgault (approved):

Issue #3757, opened 2022-01-05
SVS: Incorrect pyramid levels with 6.8.0


dgault:
Issue was raised on the forum thread https://forum.image.sc/t/problem-about-opening-some-svs-slides-in-qupath-v0-3-1-bio-formats-6-8-0/61404/15

A sample file was provided on the thread which reproduces the problem. The problem seems to be as a result of changes to the handling of macro and label images in https://github.com/ome/bioformats/pull/3733

Using the provided sample file with 6.8.0 you get:
```
Series count = 1
Series #0 -- :
	Resolutions = 6
		sizeX[0] = 62094
		sizeX[1] = 15523
		sizeX[2] = 3880
		sizeX[3] = 1024
		sizeX[4] = 1549
		sizeX[5] = 685
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (true color)
	Width = 62094
	Height = 36729
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Tile size = 256 x 256
	Thumbnail size = 685 x 530
	Endianness = intel (little)
	Dimension order = XYCZT (uncertain)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0
```

With 6.8.0 and flattened resolutions you get:
```
Series #0 -- Series 1:
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (true color)
	Width = 62094
	Height = 36729
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Tile size = 256 x 256
	Thumbnail size = 685 x 530
	Endianness = intel (little)
	Dimension order = XYCZT (uncertain)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0

Series #1 -- Series 2:
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (true color)
	Width = 15523
	Height = 9182
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Tile size = 256 x 256
	Thumbnail size = 685 x 530
	Endianness = intel (little)
	Dimension order = XYCZT (uncertain)
	Pixel type = uint8
[...]

PR #3759, opened 2022-01-06
SVS: Detect label and macro if no image description

dgault:
This is a rather simplistic attempt to deal with the issue detected in https://github.com/ome/bioformats/issues/3757
If no image description is present then it will try to treat the series as a label or macro.

Fixes https://github.com/ome/bioformats/issues/3757

To test:
- All existing builds and repo tests should remain green
- Test manually with the new sample file from the thread (also included in the new config PR)
- Test with the sample files impacted by the original PR in https://github.com/openmicroscopy/data_repo_config/pull/503

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SVSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SVSReader.java
@@ -283,6 +283,12 @@ public class SVSReader extends BaseTiffReader {
 
       String comment = ifds.get(index).getComment();
       if (comment == null) {
+        if (labelIndex == -1) {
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested using both the newly provided SVS sample as well as former samples using for https://github.com/ome/bioformats/pull/3733. Everything works as expected with macro and label images detected appropriately.

In the absence of additional representative samples, this feels like the best compromise allowing to fix the regression for some SVS variants reported in #3757 while preserving the improvements introduced in 6.8.0

PR #3764, opened 2022-01-19
SVS: Macro and Label image improvements

dgault:
This is a follow up to the previous SVS fix: https://github.com/ome/bioformats/pull/3759

It was reported on thread https://forum.image.sc/t/problem-about-opening-some-svs-slides-in-qupath-v0-3-1-bio-formats-6-8-0/61404/27 and Issue https://github.com/ome/bioformats/issues/3757

The one sample file from the thread that was still causing issues looks as though the extra images can be identified based on tiled vs strips in the IFD. This also raised an issue in that the extra image in this case was not at the end of the indexes and so reordering is needed after identifying the label and/or macro images.

Fixes https://github.com/ome/bioformats/issues/3757

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SVSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SVSReader.java
@@ -280,14 +280,15 @@ public class SVSReader extends BaseTiffReader {
       setSeries(i);
       int index = getIFDIndex(i, 0);
       tiffParser.fillInIFD(ifds.get(index));
-
+      long rowsPerStrip = ifds.get(index).getRowsPerStrip()[0];
+      long tileLength = ifds.get(index).getIFDLongValue(IFD.TILE_LENGTH, 0);
       String comment = ifds.get(index).getComment();
-      if (comment == null) {
+      if (comment == null || (tileLength == 0 && rowsPerStrip >=1)) {
         if (labelIndex == -1) {
-          labelIndex = i;
+          labelIndex = index;
         }
         else if (macroIndex == -1) {
-          macroIndex = i;
+          macroIndex = index;
         }
         continue;
       }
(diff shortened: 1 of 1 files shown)


sbesson:
As expected, a few failures associated with this PR related to the pixel hashes tests.

Quickly tested it functionally using the TCGA file from https://github.com/ome/bioformats/issues/3757#issuecomment-1150075421.

Importing the image into OMERO, certainly improves  the experience as the thumbnail is fixed and the navigation through the different resolution level is working as expected.

Screenshot 2022-06-09 at 12 15 25

With the current state of this PR and due to the strip/tile criteria, the thumbnail plane (second IFD) is identifed as a separate series and imported as a separate image (suffixed with `label`):

```
Reading core metadata
filename = /uod/idr/scratch/inbox/TCGA-OR-A5K9-01Z-00-DX2.7C0DDEF6-8CCC-4640-820F-7703616E2FEF.svs
Series count = 2
Series #0 :
	Resolutions = 4
		sizeX[0] = 95616
		sizeX[1] = 23904
		sizeX[2] = 5976
		sizeX[3] = 2988
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (true color)
	Width = 95616
	Height = 57015
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Tile size = 240 x 240
	Thumbnail size = 128 x 76
	Endianness = intel (little)
	Dimension order = XYCZT (uncertain)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0

Series #1 :
	Image count = 1
	RGB = true (3) 
	Interleaved = false
	Indexed = false (true color)
	Width = 1024
	Height = 610
	SizeZ = 1
	SizeT = 1
	SizeC = 3 (effectively 1)
	Tile size = 1024 x 16
	Thumbnail size = 128 x 76
	Endianness = intel (little)
	Dimension order = XYCZT (uncertain)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = true
	-----
	Plane #0 <=> Z 0, C 0, T 0
```

Should the thumbnail IFD  be treated as an extra associated image or as the smallest resolutions level for the primary WSI? The first option is inline with how OpenSlide handles it - see https://openslide.org/formats/aperio/  while I think the second option is how  Bio-Formats has been handling the SVS format (and generally any WSI format).

PR #3766, opened 2022-01-26
OME-TIFF reader: reduce open file handle count during setId

melissalinkert:
Without this PR, creating a fake plate:

```
bfconvert "test&plateRows=16&plateCols=24&fields=40.fake" "test_s%s.ome.tiff" -option ometiff.companion true
```
and then attempting to initialize:

```
 showinf -nopix -omexml test_s0.ome.tiff
```

throws an exception due to too many open files. With this PR, the same ```showinf``` command should eventually display OME-XML without an error.

@sbesson reported seeing a similar problem with real data, so hopefully this fixes that case too.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -281,6 +281,27 @@ public class OMETiffReader extends SubResolutionFormatReader {
       FormatTools.NON_SPECIAL_DOMAINS;
   }
 
+  @Override
+  public void setSeries(int series) {
+    super.setSeries(series);
+
+    // don't keep inactive series' files open
+    for (int i=0; i<info.length; i++) {
+      if (i != series && info[i] != null) {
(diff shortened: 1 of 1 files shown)


dgault:
Looks like this may haved caused some failures in https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/166283/consoleFull and https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/166234/consoleFull

melissalinkert:
Excluding for now, since I probably won't have time to investigate the failures today.

melissalinkert:
With f0a3e81, the `ome-tiff` and `olympus-oir` tests seem to be passing locally. Removing exclude label to check tomorrow's build.

sbesson (approved):
See https://github.com/IDR/bioformats/pull/27

At least from a functional perspective, this PR brings us to a good place where the handling of OME-TIFF  HCS filesets is comparable to HCS PFFs of similar characteristics (number of files, numbers of images).

Issue #3513, opened 2020-02-25
Leica LIF: Missing laser settings


dgault:
Issue was raised on forum thread https://forum.image.sc/t/leica-lif-metadata-fiji-extracting-laser-settings-for-multiple-channels/34250/4

Device information is correctly populated on OME-XML
However additional metadata such as laser intensity shown in 'Show Info' only contains values for a single laser. This may be because the same key values are being used for each laser and overwriting the previous entries. Further investigation is required to confirm.

A sample file has been made available in QA-29127 and issue reproduced with Bio-Formats 6.3.1

dgault:
It has also been reported on the same image sc thread that this also impacts other per-channel metadata such as exposure.

dgault:
I haven't gone through this in detail, and although its a slightly different sceanrio, we have encountered issues with series metadata having the same key names across each series. In these cases we append a prefix for each series (such as "Series0 |" ). This is done using a helper method to merge the full list of original metadata and the series data, appending a different prefix each time (https://github.com/ome/bioformats/blob/develop/components/formats-api/src/loci/formats/FormatReader.java#L1462). 

Perhaps it is possible to use the same arrays from the OMEXML parsing and use the MetadataTools.merge helper function to merge them into the original metadata with a different prefix such as Laser0 etc.

For the DetectionRange values, I opened up the sample file you provided and the raw metadata seems to have the values set to 0:
```
< D e t e c t o r   N a m e = " P M T   2 "   T y p e = " P M T "   S c a n T y p e = " I n t e r n a l "   C h a n n e l = " 2 "   C h a n n e l N a m e = " C h a n n e l   2 "   I s A c t i v e = " 1 "   S T E D D e t e c t o r M o d e = " 1 "   S T E D D e t e c t o r M o d e N a m e = " C o n f o c a l "   S T E D _ L a t e r a l S a t u r a t i o n F a c t o r = " 0 "   G a i n = " 3 9 9 . 9 9 6 1 8 5 2 4 4 5 2 6 "   O f f s e t = " 0 "   I s S T E D D e t e c t o r = " 1 "   I s S R S D e t e c t o r = " 0 "   D e t e c t i o n R a n g e B e g i n = " 0 "   D e t e c t i o n R a n g e E n d = " 0 "   P h a s e = " 0 "   I s F l i m D e t e c t o r = " 0 "   I s H P D D e t e c t o r = " 0 "   C a n D o T i m e G a t e = " 0 " / >
```

[...]

PR #3769, opened 2022-01-28
Display all channel info in .lif metadata

zacsimile:
Proposed fix for #3513. Also a change in how laser indices are calculated in the absence of a qualifier. The main reason for the latter is I think this helps line 754 evaluate properly.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -1227,7 +1227,7 @@ public class LIFReader extends FormatReader {
       translateDetectors(image, index);
 
       final Deque<String> nameStack = new ArrayDeque<String>();
-      populateOriginalMetadata(image, nameStack);
+      populateOriginalMetadata(image, nameStack, null, -1);
       addUserCommentMeta(image, i);
     }
     setSeries(0);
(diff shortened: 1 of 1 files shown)


dgault:
Hi @zacsimile, thank you for opening the PR and contributing to the project. I have marked the PR for inclusion in our automated testing. We will aim to have this PR reviewed for the next 6.9.0 milestone.

We have recently introduced a CLA for contributors, would you be able to sign and return the form following the instructions on https://ome-contributing.readthedocs.io/en/latest/cla.html

dgault:
The initial test runs with this PR included look positive. Additional metadata was detected in existing sample files which will require me to make additions to the repo test configuration files but the changes all looked to be additional metadata being parsed with no obvious regressions.

dgault:
I have reverted the change for dealing with the empty qualifier as it was resulting in some failures with a set of new sample files which we have received.

dgault:
In its current state this PR is passing all of the builds and repo tests. Testing against the original sample file shows that the core metadata remains unchanged, as it does for all samples in the repo. However the global metadata now contains the full list of metadata, including the previously missing values. All of the previously read metadata is present along with the new values. I will merge this for inclusion with the 6.9.0 release.

PR #3770, opened 2022-02-04
Operetta: Use stream for parsing XML

dgault:
This PR is in response to an issue raised on forum thread https://forum.image.sc/t/operetta-bioformats-reading-error-with-big-dataset-and-xml-parsing-crazy-slow/62768/3

In this case the initial XML file is > 800Mb, reading the entire file into a byte array and then String was very memory intensive and time consuming. Reading it in from a Stream provide much improved performance. Without this PR I was struggling to profile the issue as I would run into out of memory exceptions even with what should have been plenty of available memory (8Gb).

With the PR the initial parsing completes in around 26 seconds without the same memory issues.

The later indexing of each of the planes in the dataset still however takes a long time, I will continue profiling that code section to see if further improvements can be made.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -327,9 +327,10 @@ public class OperettaReader extends FormatReader {
 
     // parse plate layout and image dimensions from the XML file
 
-    String xmlData = DataTools.readFile(id);
     OperettaHandler handler = new OperettaHandler();
-    XMLTools.parseXML(xmlData, handler);
+    try (RandomAccessInputStream xml = new RandomAccessInputStream(id)) {
+      XMLTools.parseXML(xml, handler);
+    }
 
     // sort the list of images by well and field indices
 


sbesson (approved):
The Operetta samples in our curated QA repository are unaffected from this change with all the tests passing in today's build. Switching from reading the entire XML content as a string to using a stream should certainly improve the handling of very large files both in terms of time and JVM requirements. This seems to be confirmed by @NicoKiaru's testing on his 800M sample files - see https://forum.image.sc/t/operetta-bioformats-reading-error-with-big-dataset-and-xml-parsing-crazy-slow/62768/5.

As mentioned, it is possible that further performance improvements could be made but given the 10-fold improvement of these large datasets, getting this PR included as is in the upcoming Bio-Formats 6.9.0 release would already be a great step forward.

dgault:
Merging this now for 6.9.0, any further improvements can be handled as a separate follow up PR

NicoKiaru:
I am a bit stupid, I thought I tested on a 800Mb xml file, but in fact it was 200Mb. 

So the bug is resolved and I can work with the dataset (Thanks!), but the reading time gain for planes is probably not there (>20 minutes and the opening is still not done for the 800Mb file).

PR #3772, opened 2022-02-08
OBF: Suppress deflate errors when all expected output bytes are available

ngladitz:
@dgault this is a workaround for the issue at https://forum.image.sc/t/warning-error-loading-thumbnail-for-series-1-for-series-2-and-3/62793

Basically it ignores zlib errors that happen right after all bytes have been decompressed.
It might be preferable to still output some warning in this case.
Is there a standard / preferred way to generate a user visible warning from a Bio-Formats reader?

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -740,13 +740,26 @@ public class OBFReader extends FormatReader {
           throw new FormatException("Unsupported zlib compression");
         }
 
+        long bytesWrittenBefore = inflater.getBytesWritten();
+
         try {
           int decompressedBytes = inflater.inflate(buffer, bufferOffset, remainingBytes);
           bufferOffset += decompressedBytes;
           remainingBytes -= decompressedBytes;
         }
         catch (DataFormatException exception) {
-          throw new FormatException(exception.getMessage());
+          long bytesWrittenAfter = inflater.getBytesWritten();
+
+          long decompressedBytes = bytesWrittenAfter - bytesWrittenBefore;
+
+          bufferOffset += decompressedBytes;
+          remainingBytes -= decompressedBytes;
+
+          if(remainingBytes != 0) {
+            throw new FormatException(exception.getMessage());
+          } else {
+            LOGGER.warn("Error past the end of deflated stream", exception);
+          }
         }
       }
     }
--- a/components/formats-bsd/src/loci/formats/in/OBFReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OBFReader.java
@@ -748,6 +748,11 @@ public class OBFReader extends FormatReader {
           remainingBytes -= decompressedBytes;
         }
         catch (DataFormatException exception) {
+          /* For older OBF files a zlib error may occur after all bytes have been decompressed.
+           In this scenario we will attempt to ignore the error if no bytes remain and output a warning.
+           This will only work for Java versions 11 and upwards, prior versions will still throw a FormatException.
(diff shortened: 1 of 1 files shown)


dgault:
Thanks @ngladitz, adding a warning sounds like a sensible idea. Normally we do so using LOGGER.warn similar to below:

`LOGGER.warn("Could not parse OME-XML metadata", e);`

dgault:
@ngladitz, I retested this PR against the 2 sample files from the original thread (https://zenodo.org/record/5971265#.YgaDKR2nw6g) and I am still getting an exception at https://github.com/ome/bioformats/pull/3772/files#diff-107190afde145589cd213c4948a27ed0fecd3415392571849480f290149b22e3R759 as it seems there is still 512 bytes remaining. Is that what you would expect for those files?

ngladitz:
@dgault no I've tested the same two sample files with my branch and `./tools/ijview`.
I get the Bio-Formats Import-Dialog and the Series-Selection-Dialog with 3 series and visible (non gray) thumbnails.
I can open all three series of both files.
On the console I get:
```
Error past the end of deflated stream
java.util.zip.DataFormatException: invalid stored block lengths
	at java.base/java.util.zip.Inflater.inflateBytesBytes(Native Method)
	at java.base/java.util.zip.Inflater.inflate(Inflater.java:385)
	at loci.formats.in.OBFReader.readFromStack(OBFReader.java:746)
	at loci.formats.in.OBFReader.readStackFrame(OBFReader.java:916)
...
```
Which indicates the new condition and warning diagnostic seem to be working.
I am unsure what might explain the difference in behavior that you are seeing.
Could it be platform or Java version differences? 
ImageJ (started via `./tools/ijview`) reports Java 11.0.13 (64-bit).
I am using Ubuntu 21.10.

dgault:
Thanks for checking, using ijview and the other command line tools works as expected for me also. For some reason testing the jars in a recent FIJI install is throwing the exception and no images are loaded:
```
loci.formats.FormatException: invalid block type
	at loci.formats.in.OBFReader.readFromStack(OBFReader.java:759)
	at loci.formats.in.OBFReader.readStackFrame(OBFReader.java:916)
	at loci.formats.in.OBFReader.openBytes(OBFReader.java:1009)
	at loci.formats.ChannelFiller.openBytes(ChannelFiller.java:167)
	at loci.formats.ChannelSeparator.openBytes(ChannelSeparator.java:229)
```

 I will investigate further next week to see what differs between the 2 setups.

dgault:
As a follow up, the issue I was seeing does seem to be linked to the Java versions, with anything from Java 11.0 onwards the files open and display (with the expected warning) with the PR included. With prior Java versions the `loci.formats.FormatException: invalid block type` is thrown and the planes are blank.

ngladitz:
Thanks @dgault. 

I am not sure I am looking in the right place but this looks like it might be the corresponding change:
https://github.com/openjdk/jdk/commit/883d41fefc2b5da40b159b24e7387f3bdbd22a5a

Seems to be there since "jdk-11+11" and adds this comment:

```
These byte totals, as well as
the {@linkplain #getBytesRead() total bytes read}
and the {@linkplain #getBytesWritten() total bytes written}
values, will be updated even in the event that a {@link DataFormatException}
is thrown to reflect the amount of data consumed and produced before the
exception occurred.
```

I suppose that means I am out of luck. The commit supposedly won't hurt and it might start working if ImageJ ever switched to something more recent but I guess that doesn't happen too often. Feel free to merge or leave it unmerged as you see fit.

dgault:
Thanks @ngladitz, as this will still improve the experience for users with later versions and wont have any negative impact on other users then I think its worth including. I have added an additional commit just to document the differing behaviour and I will make a similar comment on the version specific behaviour in the release notes

PR #3783, opened 2022-02-25
ND2 metadata updates

melissalinkert:
Attempts to improve some ND2 metadata parsing, particularly around objective data, timestamps, and exposure times.

This is backported from a private PR; corresponding files aren't available. I would expect at least a few of the existing .nd2 files to be affected though (hopefully for the better).

I wouldn't expect this to break memo files, but there are a couple of new API methods in the reader.


melissalinkert:
I'm working on reverting the exposure time and excitation wavelength changes here, but won't have something pushed in time to get the next build to pass (but hopefully will tomorrow).

dgault:
Changes here look sensible and builds and tests have remained green with this PR and the data-repo PR included. 

I've been through all of the config changes on the data-repo PR and the updates there all look to be correct and match the values being displayed in Elements. A manual testing of a subset of those files confirmed that the updated values from the PR are correct.

PR #3775, opened 2022-02-10
Improves performance of OperettaReader

NicoKiaru:
* Avoids a N^2 algorithm by using a hashmap to find the correct plane
* Very significant speedup for big dataset (k.100 000 planes), from 30 minutes to 30s with a 800 Mb xml file

PR #3784, opened 2022-02-25
OperettaReader improvements

melissalinkert:
Backported from a few private PRs.

This makes well names more readable now that #3347 is merged, and adds better checks for invalid TIFFs and supplemental metadata files.

I don't think this will conflict with #3775, and it should be OK for a patch release too.


dgault:
Builds and tests have continued to stay green with this PR and the associated config PR included.

The changes here all look to be sensible, the smarter checks for companion files and TIFF validation all look good. The only immediate difference in any of our repo test files is the image name updates to provide better naming for Wells. The updated names look to be correct and are an improvement. Overall the PR looks good to merge.

PR #3794, opened 2022-03-14
Chunk API: Add new API and functionality for reading and writing chunks

dgault:
An initial prototype of potential new API methods for reading and writing N dimensional chunks of data rather than 2d planes/tiles. These new methods currently have default implementations in FormatReader and FormatWriter than pass through to the current 2D API, so they should be useable. This will be accompanied by a PR on ZarrReader which would be the first instance of a Reader which uses the chunk API by default.

The naming and parameters are all open for debate/suggestions, one of the trickier areas with this is how to handle the DimensionOrder going forward. Currently it is reusing the existing 5D dimensionOrder string.


components/formats-api/src/loci/formats/FormatTools.java
```diff
@@ -697,6 +697,21 @@ public static int getIndex(String order,
     };
   }
 
+  public static int[] getXYZCTIndexes(String order, int[] values) {
```
melissalinkert, line 700: Is this meant to return in `XYZCT` order (as the name suggests) or `XYZTC` order (as implemented)?

dgault: Good catch, updated now to match the name

components/formats-api/src/loci/formats/FormatWriter.java
```diff
@@ -330,6 +362,20 @@ public int setTileSizeY(int tileSize) throws FormatException {
...
+  @Override
+  public int[] getChunkSize() throws FormatException {
+    int[] defaultChunkSize = {getTileSizeX(), getTileSizeY(), 1, 1, 1};
+    return defaultChunkSize;
+  }
+
+  /* @see IFormatWriter#setChunkSize(int[]) */
+  @Override
+  public int[] setChunkSize(int[] chunkSize) throws FormatException {
+    int[] returnChunkSize = {setTileSizeX(chunkSize[0]), setTileSizeY(chunkSize[1]), 1, 1, 1};
```
joshmoore, line 375: Back to @melissalinkert's question about what would test this code, this could warn or error if a >2D chunk size were requested.

melissalinkert:
Temporarily excluding since this is scheduled for 8.0.0 and conflicts with things scheduled for 7.1.0.

joshmoore:
With the change of IDR's strategy (in this case, to the use of the dedicated omero-zarr-pixel-buffer rather than ZarrReader) this PR is currently unneeded. I've backed up the branch to my repo just in case.

_Discussed in today's formats meeting with @melissalinkert, @sbesson, @jo-mueller, @dominikl and @kkoz_

PR #3795, opened 2022-03-15
CZI: use fill color of 255 when brightfield data is suspected

melissalinkert:
Backported from a private PR. This should result in a white background for brightfield data (instead of black), which better matches Zen and other software.

There are a few different ways we can check for a brightfield pyramid (including by channel name),
but checking for RGB + multiple resolutions seems to be reliable so far and does not require a
memo file update.  @swg08 if you have any thoughts on a better way to determine the correct fill color in general, I'd be happy to hear them.

I would expect some MD5 test failures, and will open a matching configuration PR after the tests have run once.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -356,7 +356,7 @@ public class ZeissCZIReader extends FormatReader {
     }
 
     byte fillColor = (byte) 0;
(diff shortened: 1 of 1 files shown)


dgault:
Change looks sensible and is suitable for a patch release. Manually testing each of the files impacted in the repo tests and each is indeed a Brightfield image. Each of the files opens and displays correctly with this PR included and with the associated config the builds and tests are also green. PR looks good to merge.

PR #3803, opened 2022-03-31
Zeiss LSM: assume 32 bit data is floating point

### LINKED ISSUE #3604: Zeiss LSM: Incorrect data type
Issue was raised in QA-29464 and a sample file has been provided showing the issue.

Opening the file in FIJI using the native reader shows floating point pixel type (data type 5) with pixel values between 0 and 1.
When using the Bio-Formats reader (tested with 6.5.1) the data type is read as 0 (varying data types) and pixels are displayed as uint32 values.

melissalinkert:
Fixes #3604.

Compare the pixel type and values in QA 29464 with and without this PR. It doesn't look like we have any other 32 bit .lsm files that would be affected.

I'd expect this to be safe (but low priority) for a patch release.


dgault:
Tested using the sample QA file and without this PR the pixel type is UINT32, with the PR included it is now detected as float and the image opens and displays as expected. No other sample files in the test repo are impacted by this change.

PR #3811, opened 2022-04-15
Deltavision: another fix for incorrect XY tile count detection

### LINKED ISSUE #3601: ArrayIndexOutOfBoundsException from DeltavisionReader.initExtraMetadata with v6.5.1
See [QA 29449](https://www.openmicroscopy.org/qa2/qa/feedback/29449/) and [showinf.log](https://github.com/ome/bioformats/files/5000269/showinf.log).

melissalinkert:
This adds better handling for the case when the calculated XY tile count is smaller than the number of panels (series) and at least one of the axes is acquired in reverse order.

Fixes #3601. Without this PR, `showinf` on the .dv file from QA 29449 should throw an exception as indicated in #3601. With this PR, the same command should successfully initialize and display an image. The dimensions should match what is in the .dv.log file:

```
XY Dimensions:              1024 x 1024
ZWT Dimensions (expected):  1 x 1 x 1
Panels:                     86
```

Assuming tests pass, this should be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DeltavisionReader.java
@@ -899,6 +899,19 @@ public class DeltavisionReader extends FormatReader {
         yTiles = 1;
       }
     }
+    else if (xTiles * yTiles < getSeriesCount() && (backwardsStageX || backwardsStageY)) {
+      if (backwardsStageX && yTiles == 1) {
+        xTiles = getSeriesCount();
+      }
+      else if (backwardsStageY && xTiles == 1) {
+        yTiles = getSeriesCount();
+      }
+      else {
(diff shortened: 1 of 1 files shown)


dgault:
With the PR included builds and tests have remained green.

Tested using the sample file from the linked Issue. 
Without the PR the QA file results in an exception being thrown:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 167
	at loci.formats.in.DeltavisionReader.initExtraMetadata(DeltavisionReader.java:936)
	at loci.formats.in.DeltavisionReader.initFile(DeltavisionReader.java:337)
```
With the PR included the file is able to be opened and displayed without any exception. The pixel data looks sensible and the dimensions match those from the log file.

PR #3826, opened 2022-05-18
logback version bump from 1.2.0 to 1.2.9

dgault:
To match the version bumps in other repos.


sbesson (approved):
Same as https://github.com/ome/ome-model/pull/160#pullrequestreview-977170927

sbesson:
Also opened https://github.com/ome/bioformats/pull/3827 to bump all components with logback as a dependency to their latest release.

PR #3830, opened 2022-05-25
Update xercesImpl to version 2.12.2

sbesson:
Also declare xml-apis 1.4.01 in the dependency tree of formats-bsd to avoid downstream components to resolve xml-apis 1.3.02, which is a transitive dependency of ome-common but incompatible with the recent version of xercesImpl


dgault:
Update looks good to me, builds and tests are all green again with the PR included. The xercesImpl version will be ahead of that used in FIJI but that should have no impact for plugin users and I can make note of the update when bumping the downstream jars.

PR #3834, opened 2022-05-31
Update metadata extractor to version 2.18.0

sbesson:
See https://github.com/drewnoakes/metadata-extractor/releases for a list of changes in this version.

This pulls a more recent version of `xmpcore` where packages have been renamed internally but apart from the `EXIFServiceImpl` class, all changes should be internal to the library.

With Bio-Formats 6.10.0 behing us, opening to see the impact on the curated QA repository builds.

### DIFF
--- a/components/formats-bsd/pom.xml
+++ b/components/formats-bsd/pom.xml
@@ -110,7 +110,7 @@
     <dependency>
 	    <groupId>com.drewnoakes</groupId>
 	    <artifactId>metadata-extractor</artifactId>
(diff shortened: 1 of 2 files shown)


dgault:
Version bump looks good to me and all builds and tests have remained green with this PR included. Tested manually with JPEG sample files as the EXIFServiceImpl seems to be the only impacted area for us. With the PR included all files opened and displayed as expected and the metadata remained consistent compared to the latest 6.10.0 release.

PR #3835, opened 2022-06-02
Zeiss CZI: use scene names as image names

melissalinkert:
Backported from a private PR.

This is meant to preserve scene names in the `Image` name, in cases where the scene name has been set during acquisition. I expect to see some test failures which will require a configuration PR (and will indicate which files can be used to test).

These changes should not affect memo files, and should be safe for a patch release.


dgault (approved):
Builds and tests have continued passing with this change included. There have been no repo test failures which means I unfortunately don't have any sample files to test or verify. 

The code changes themselves look good though and don't impact memo files so are suitable for the upcoming patch release.

PR #3841, opened 2022-06-27
TIFF: prevent ClassCastException if the SubfileType tag has the wrong type

melissalinkert:
Backported from a private PR.

This adds a check for the `Number` class when retrieving the `SubfileType` TIFF tag. Some files incorrectly store an array in this tag, which causes a `ClassCastException` without this PR.


dgault (approved):
Builds and tests have remained green with this PR included. This solution seems like the most sensible approach for handling the incorrect data types. Looks good for the upcoming patch release

PR #3831, opened 2022-05-27
OME-TIFF: faster initialization when there are lots of planes

melissalinkert:
Backported from a private PR. Opening to see if repo tests fail, will exclude tomorrow if they do. 

This is meant to speed up `setId` time for cases such as the output of `bfconvert "test&sizeZ=4000.fake" test.ome.tiff`.

Unit tests fail without 2ebbc58, which I think actually exposes an issue with bfconvert. `bfconvert "test&sizeC=2&series=3.fake" -tilex 256 -tiley 256 -channel 1 test-split_%x_%y_%m.ome.tiff` results in in this OME-XML:

```
<?xml version="1.0" encoding="UTF-8"?>
<!-- Warning: this comment is an OME-XML metadata block, which contains crucial dimensional parameters and other important metadata. Please edit cautiously (if at all), and back up the original data before doing so. For more information, see the OME-TIFF web site: https://docs.openmicroscopy.org/latest/ome-model/ome-tiff/. -->
<OME xmlns="http://www.openmicroscopy.org/Schemas/OME/2016-06" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Creator="OME Bio-Formats 6.10.0-SNAPSHOT" UUID="urn:uuid:aa1ec1c8-4f03-4f22-9634-345a61d7275a" xsi:schemaLocation="http://www.openmicroscopy.org/Schemas/OME/2016-06 http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd">
  <Image ID="Image:0" Name="test">
    <Pixels BigEndian="false" DimensionOrder="XYZCT" ID="Pixels:0" Interleaved="false" SignificantBits="8" SizeC="1" SizeT="1" SizeX="256" SizeY="256" SizeZ="1" Type="uint8">
      <Channel ID="Channel:0:0" SamplesPerPixel="1">
        <LightPath/>
      </Channel>
      <Channel ID="Channel:0:1" SamplesPerPixel="1">
        <LightPath/>
      </Channel>
      <TiffData FirstC="0" FirstT="0" FirstZ="0" IFD="0" PlaneCount="1">
        <UUID FileName="test-split_0_0_0.ome.tiff">urn:uuid:aa1ec1c8-4f03-4f22-9634-345a61d7275a</UUID>
      </TiffData>
    </Pixels>
  </Image>
  <Image ID="Image:1" Name="test 2">
[...]

PR #3842, opened 2022-06-29
OME-TIFF: improve performance with open file handle reduction

melissalinkert:
Continuation of #3766.

#3766 introduced degraded `setId` performance for single file OME-TIFF datasets with a larger number of series. `bfconvert "test&sizeZ=20&series=300.fake" lots-of-series.ome.tiff` produces a file that demonstrates the issue.

With this PR, `showinf -nopix -omexml lots-of-series.ome.tiff` should be noticeably faster than the same command using 6.9.1 or 6.10.0. There should also be fewer `Reading IFDs` lines in the `showinf` output with this PR.

The file handle test case in #3766 should still work with these changes, and tests should continue to pass.

Technically should be safe for a patch release, but given how much we rely on OME-TIFF will definitely need some careful testing to make sure this doesn't make anything worse.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -289,7 +289,9 @@ public class OMETiffReader extends SubResolutionFormatReader {
     for (int i=0; i<info.length; i++) {
       if (i != series && info[i] != null) {
         for (OMETiffPlane p : info[i]) {
-          if (p != null && p.reader != null) {
+          if (p != null && p.reader != null &&
+            !getCurrentFile().equals(p.reader.getCurrentFile()))
+          {
             try {
               p.reader.close();
             }
@@ -1432,7 +1434,13 @@ public class OMETiffReader extends SubResolutionFormatReader {
         }
         core.add(s, c);
       }
-      r.close();
+      // if we have multiple OME-TIFF files, close to reduce the
+      // number of open file handles
+      // keep the "main" file's handle open though, to improve
+      // performance when there is only one file
+      if (!getCurrentFile().equals(r.getCurrentFile())) {
+        r.close();
+      }
     }
     core.reorder();
 


sbesson:
Thanks @melissalinkert. The challenge is to optimize the reader behavior in order to support two very different but valid file layouts allowing to store multi-image data (such as HCS) as OME-TIFF filesets:

- a single large OME-TIFF like the sample plate at https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/BBBC/ (2300 images/well samples with 3 planes each)
- one OME-TIFF file per series e.g. using a similar layout such as https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/plate-companion/

From a specification perspective, both layouts are equally valid and have their respective advantages and drawbacks.  

Taking the first sample as the representative example for testing, with Bio-Formats 6.9.0 (i.e. before #3766), the initialization is almost instant

```
[root@8311f729a5ac /]# ./bftools/showinf -nopix  -omexml /uod/idr/repos/curated/ome-tiff/public/2016-06/BBBC/NIRHTa-001.ome.tiff | grep Init
Initializing reader
Initialization took 7.941s
```

With Bio-Formats 6.9.1, the same command quickly runs the first pass mapping the `IFDs` to the `Image/Plane` but the IFDs are parsed several times as suggested in the description leading to much longer initialization times.

With the daily merge build, the `TiffData` processing is extremely slow (1-2s / image in debug mode) making it hard to assess the impact of this PR. Looking at the open PRs, https://github.com/ome/bioformats/pull/3831 appears as the most plausible candidate. Would it be acceptable to exclude this so that we can assess the impact of this PR across our entire workflow?

melissalinkert:
#3831 now excluded. Makes complete sense to focus on one set of changes at a time.

sbesson:
Tested the initialization times using a combination of files derived from the first plate in BBBC017 as well as the fake file in the description.
Both source files were converted into 2 OME-TIFF layouts:

- single OME-TIFF
- multi-file OME-TIFF with 1 file / series and a companion file

Below  are the initialization times measured by multiple runs of `showinf` with Bio-Formats 6.7.0, Bio-Formats 6.9.1 and the nighthly CI build including this PR

```
./showinf -nopix -omexml  <path_to_file> | grep Ini
```

The first initialization time for each fileset was subsequently larger than any other initialization time and was discarded as it includes some cold access.

|                              	| Series 	| Planes/series 	| Files 	| Nightly build (incl. this PR) 	| 6.9.1             	| 6.9.0             	| 6.7.0             	|
|------------------------------	|--------	|---------------	|-------	|-------------------------------	|-------------------	|-------------------	|-------------------	|
| NIRHTa-001.ome.tiff          	| 2304   	| 3             	| 1     	| 9.833,8.641,8.568             	| 762.572           	| 7.871,8.092,7.381 	| 7.587,7.004,7.524 	|
| NIRHTa-001.companion.ome     	| 2304   	| 3             	| 2305  	| 8.617,8.334,8.095             	| 8.378,7.333,7.419 	| 6.741,5.628,5.781 	| 5.796,5.837,6.343 	|
| lots-of-series.ome.tiff      	| 300    	| 20            	| 1     	| 7.25,7.305,7.204              	| 712.715s          	| 4.992,5.084,5.09  	| 5.081,5.184,5.142 	|
| lots-of-series.companion.ome 	| 300    	| 20            	| 301   	| 3.493,3.584,3.436             	| 3.842,3.68,3.694  	| 3.271,3.107,3.034 	| 3.193,3.166,3.166 	|


With Bio-Formats 6.9.1, the initialization times for the single OME-TIFF were only measured once but clearly demonstrate the performance issue.

Overall, there is a still a slight performance reduction compared to the initialization times in Bio-Formats 6.7.0 but the  issue for single file OME-TIFF with large number of series is addressed. I will load these files into a testing OMERO.server to compare access times.

sbesson:
Imported all filesets mentioned above into an OMERO.server with this PR included.

NIRHTa-001 

```
(.venv3) bash-4.2$ omero fs importtime Fileset:126609
Using session for user-3@localhost:4064. Idle timeout: 10 min. Current group: read-only-1
   upload time of  50.11s for 2305 files (0.022s/file)
    setId time of  13.19s
 metadata time of 347.28s
   pixels time of 394.20s for 6912 planes (0.057s/plane)
    rdefs time of 429.05s for 2304 rendering settings (0.186s/rdef)
thumbnail time of 253.98s for 2304 thumbnails (0.110s/thumbnail)
(.venv3) bash-4.2$ omero fs importtime Fileset:126608
Using session for user-3@localhost:4064. Idle timeout: 10 min. Current group: read-only-1
   upload time of   1.19s for 1 file (1.185s/file)
    setId time of  14.16s
 metadata time of 317.88s
   pixels time of 396.00s for 6912 planes (0.057s/plane)
    rdefs time of 476.52s for 2304 rendering settings (0.207s/rdef)
thumbnail time of 296.56s for 2304 thumbnails (0.129s/thumbnail)
```

lots-of-series

```
(.venv3) bash-4.2$ omero fs importtime Fileset:126610
Using session for user-3@localhost:4064. Idle timeout: 10 min. Current group: read-only-1
   upload time of   0.46s for 1 file (0.455s/file)
    setId time of   7.28s
 metadata time of  12.79s
   pixels time of  45.88s for 6000 planes (0.008s/plane)
    rdefs time of 112.02s for 300 rendering settings (0.373s/rdef)
thumbnail time of  62.20s for 300 thumbnails (0.207s/thumbnail)
(.venv3) bash-4.2$ omero fs importtime Fileset:126611
Using session for user-3@localhost:4064. Idle timeout: 10 min. Current group: read-only-1
   upload time of   6.40s for 301 files (0.021s/file)
    setId time of   2.97s
 metadata time of  15.72s
   pixels time of  45.10s for 6000 planes (0.008s/plane)
    rdefs time of  61.33s for 300 rendering settings (0.204s/rdef)
thumbnail time of  29.89s for 300 thumbnails (0.100s/thumbnail)
```

So the import times are fully comparable. Unless we want to investigate more deeply the performance hit since 5.9.0 (related to the looping over several 100-1000 planes?), this looks good for inclusion in a patch release.

Additionally, this reveals that it would be useful to have a few reference real-wolrd OME-TIFF dataets for these use cases in our curated repository. https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/BBBC/ would be a relevant example although this testing reminded me the converted data is shifted by one column to the right due to a bug in the reader which got corrected and released after the conversion was made.
A possible suggestion would be to regenerate the converted representation of this plate both as single file and multi-file OME-TIFF fileset, add these to the curated repository as public data, register them in https://docs.openmicroscopy.org/ome-model/latest/ome-tiff/data.html#plate and hide the incorrectly converted OME-TIFF from the index?

sbesson (approved):

dgault:
This has been in the merge builds and tests for a while now without issue. The repo test times themselves are too inconsistent to draw any conclusions from, including BBBC and other real world sample sets may be worth considering going forward. The strategy of leaving the main file handle open seems like a sensible compromise to prevent the regression in the single file use case. 

Seb's testing makes it clear that the immediate issue of the performance regression for single file, multi series in 6.9.1 has been addressed. Additionally I retested the use case from the previous PR https://github.com/ome/bioformats/pull/3766 and can confirm that it appears to be unaffected by this change and the open file handle exception is not seen. I have been using the merge build with this PR locally for day to day use and have not encountered any noticeable performance hits or regressions.

PR #3848, opened 2022-07-12
Bump ome-common and ome-poi versions

sbesson:
These include dependency version bumps that should address https://github.com/ome/openmicroscopy/pull/6328

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,10 +49,10 @@
     <slf4j.version>1.7.30</slf4j.version>
     <kryo.version>4.0.2</kryo.version>
     <testng.version>6.8</testng.version>
-    <ome-common.version>6.0.9</ome-common.version>
+    <ome-common.version>6.0.11</ome-common.version>
     <ome-model.group>org.openmicroscopy</ome-model.group>
     <ome-model.version>6.3.1</ome-model.version>
-    <ome-poi.version>5.3.6</ome-poi.version>
+    <ome-poi.version>5.3.7</ome-poi.version>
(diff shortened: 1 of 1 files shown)


jburel:
it leads to the following error (See https://github.com/ome/openmicroscopy/runs/7302418261?check_suite_focus=true) when building BF and openmicroscopy
```
/home/runner/.m2/repository/org/openmicroscopy/ome-common/6.0.9/ome-common-6.0.9.jar
		::::::::::::::::::::::::::::::::::::::::::::::
		::              FAILED DOWNLOADS            ::
		:: ^ see resolution messages for details  ^ ::
		::::::::::::::::::::::::::::::::::::::::::::::
		:: org.openmicroscopy#ome-common;6.0.9!ome-common.jar
		::::::::::::::::::::::::::::::::::::::::::::::
```

sbesson:
Does this imply that for the coupled build to work, all the versions have to be unified across components? That's doable and we have effectively done so in the past but it's also fairly contraining in terms of release and raises the question of whether these components should be isolated at all.

Is there some short (1-2 days max) investigation that we could schedule to see if we could lift this limitation at the OMERO level?

jburel:
It is a constraint. I have not looked into, I just ran the build to see what was happening

jburel:
the commit https://github.com/ome/openmicroscopy/pull/6328/commits/e1d9101174440c3f765e68f663ed7fcd9ea3c848 seems to be enough to build everything together

jburel (approved):

PR #3851, opened 2022-07-26
OME-TIFF: allow individual file reading for datasets with one file per series

melissalinkert:
The intended use case is something like:

```
$ bfconvert "test%series=10.fake" output_s%s.ome.tiff
$ showinf -nogroup output_s9.ome.tiff
```

https://github.com/glencoesoftware/raw2ometiff/pull/80 is related.

Without this change, `showinf` would have thrown an exception. With this change, it should detect a single series which corresponds to the chosen file.

All of the changes are limited to the case where the dataset has multiple files and file grouping has been turned off explicitly. I would expect this to be safe for a patch release, but don't have strong feelings about including in 6.10.1.

/cc @DavidStirling

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMETiffReader.java
@@ -611,11 +611,14 @@ public class OMETiffReader extends SubResolutionFormatReader {
     if (!isGroupFiles() && !isSingleFile(currentId)) {
       IFormatReader reader = new MinimalTiffReader();
       initializeReader(reader, currentId);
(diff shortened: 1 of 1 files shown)


dgault:
All builds and tests have continued to pass with this PR included.

Testing without this PR and running showinf with the nogroup option results in an exception as below:
```
Reading IFDs
Populating metadata
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 2, Size: 2
	at loci.formats.in.OMETiffReader.initFile(OMETiffReader.java:755)
```

With this PR included there is no exception thrown and a single series is detected as expected.

PR #3852, opened 2022-07-26
NDPI: fix non-number tag handling for files > 4GB

### LINKED ISSUE #3725: Large NDPI files fail to open with java.lang.ClassCastException
Reported via QA 30384, some large NDPI files (>4GB) fail to open with the following stack trace

```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
	at loci.formats.in.NDPIReader.initStandardMetadata(NDPIReader.java:381)
	at loci.formats.in.BaseTiffReader.initMetadata(BaseTiffReader.java:97)
	at loci.formats.in.BaseTiffReader.initFile(BaseTiffReader.java:586)
	at loci.formats.in.NDPIReader.initFile(NDPIReader.java:311)
```

Two representative NDPI samples have been uploaded (as multi-part) as https://doi.org/10.5281/zenodo.5337013. 

Support for large NDPI files (>4GB) had been improved in https://github.com/ome/bioformats/pull/3505 but the latest released version of Bio-Formats (6.7.0) still fails against these files.  `tiffinfo` sanity checks also fail against these files. However the files themselves load without problem into Hamamatsu NDP.View2.

melissalinkert:
Tag 65459 in particular has ASCII type. The tag type may be stored incorrectly in the file, since the value is more consistent with
a LONG type. We have no documentation on this tag, though, so for now just preventing a ClassCastException is enough.

Fixes #3725. Should be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NDPIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NDPIReader.java
@@ -377,7 +377,7 @@ public class NDPIReader extends BaseTiffReader {
                   entry.getValueCount(), newOffset);
                 ifd.put(tag[t], newEntry);
               }
-              else {
+              else if (value instanceof Number) {
                 ifd.put(tag[t], ((Number) value).longValue() + highOrder[t]);
               }
             }


dgault (approved):
With this PR the sample files from the original report are able to be opened without the exception occurring. Without any relevant docs avoiding the cast exception is as much as we can do for now. The sample files are now copied to the repo and are included and passing in nightly tests. PR looks good to merge.

jburel:
Reading the original issue, should we make an attempt at converting the  "String" into a "Number"?

melissalinkert:
I intentionally didn't make any attempt to force this tag to be interpreted as a `Number`, since we just don't know what it actually represents. It's not clear if the bytes should be read as an `int`, a `float`, or a meaningless placeholder value. If/when we have more information about tag 65459, it's definitely worth going back and refining this.

PR #3857, opened 2022-08-08
Replace one-jai with jai-imageio-jpeg2000

dgault:
A follow up to https://github.com/ome/ome-codecs/pull/21 and the forum thread https://forum.image.sc/t/status-of-jai-imageio-forks/69438
Replaces the deprecated ome-jai with https://github.com/jai-imageio/jai-imageio-jpeg2000

### DIFF
--- a/components/formats-bsd/src/loci/formats/services/JAIIIOServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JAIIIOServiceImpl.java
@@ -47,17 +47,17 @@ import javax.imageio.spi.IIORegistry;
 import javax.imageio.stream.ImageOutputStream;
 import javax.imageio.stream.MemoryCacheImageInputStream;
 
+import com.github.jaiimageio.jpeg2000.J2KImageReadParam;
+import com.github.jaiimageio.jpeg2000.J2KImageWriteParam;
+import com.github.jaiimageio.jpeg2000.impl.J2KImageReader;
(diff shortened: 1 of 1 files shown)


sbesson:
Interestingly, the JAI service interface and implementation have been completely copied over to the `ome-codecs` component and they do not seem to be used anywhere in the codebase. From https://github.com/search?q=org%3Aome+JAIIIOService&type=code, they are imported but apparently unused in https://github.com/ome/bioformats/blob/25645389e076a7bd0011e04c4dd8982c0f0614ed/components/formats-bsd/src/loci/formats/codec/JPEG2000Codec.java#L54-L55 and the primary usage I could find in our code base was https://github.com/ome/omero-romio/blob/8e53730d44244952d6e93ac9ab8222e85db6c189/src/main/java/ome/io/nio/SimpleBackOff.java#L15.

Do you remember whether the rationale for keeping them in both repositories was primarily to avoid breaking changes and cause a major version increment? Or are both services still meeting different needs?
I wonder if we could consider removing the content of theses services classes, forwarding the `ome.codecs` classes and deprecating them in preparation for a removal in a future major version increment. Additionally, this might be sufficient to pass the CI builds.

dgault:
It's likely that we could remove those services all together but that would require a more in depth review at some point. Closing this PR with the intention of following up post release on potentially removing JAIIIOService

PR #3860, opened 2022-08-10
bfconvert: fix overwrite checking when output path is a pattern

### LINKED ISSUE #3746: bfconvert `-overwrite` option is ignored while converting to several output images
For instance such command works fine:

```
> bfconvert fixtures/images/lif1.lif output/lif1.ome.tiff
Output file output/lif1.ome.tiff exists.
Do you want to overwrite it? ([y]/n)
```

`bfconvert` correctly warns warns me about existing file.

But on the other hand, when I'm trying, for example extract separate per-serie images, like:

`bfconvert fixtures/images/lif1.lif output/lif1_%s.ome.tiff`

the first run is  perfectly fine, but the second run does not give any confirmation. Moreover, it seems that `-overwrite` and `-nooverwrite` options have no effect. And, here is the weirdest part, it seems that `bfconvert` appends data to the already existing files, e.g.:

1. Directory listing from the first run:
```
➜ ll -h output
total 41640
drwxr-xr-x   4 lucassus  staff   128B 13 Nov 11:21 .
drwxr-xr-x  83 lucassus  staff   2.6K 13 Nov 11:21 ..
-rw-r--r--   1 lucassus  staff    15M 13 Nov 11:22 lif1_0.ome.tiff
-rw-r--r--   1 lucassus  staff   4.3M 13 Nov 11:22 lif1_1.ome.tiff
```

2. ...and from the second run
```
total 79056
drwxr-xr-x   4 lucassus  staff   128B 13 Nov 11:21 .
drwxr-xr-x  83 lucassus  staff   2.6K 13 Nov 11:21 ..
-rw-r--r--   1 lucassus  staff    29M 13 Nov 11:22 lif1_0.ome.tiff
-rw-r--r--   1 lucassus  staff   8.3M 13 Nov 11:22 lif1_1.ome.tiff
```

The output files are doubled in size.

I believe that in case of multi-file conversion bftools should override destination files or exit with an error.

melissalinkert:
Fixes #3746.

This changes how the `-overwrite`/`-no-overwrite` options and corresponding user prompt work with a use case such as `bfconvert  "test&sizeZ=10.fake" "test_z%z.ome.tiff"`. Without this change, running that command twice with any combination of overwriting options will always result in `test_z0.ome.tiff`, `test_z1.ome.tiff`, etc. being appended to instead of deleted/overwritten or left as-is.

With this change, `bfconvert  "test&sizeZ=10.fake" "test_z%z.ome.tiff"` when run twice with any combination of overwriting options should behave the same as `bfconvert "test&sizeZ=10.fake" test.ome.tiff` when run twice with the same options. There should be no case where the second `bfconvert` command appends to files that already existed.

Right now, if neither `-overwrite` nor `-no-overwrite` is specified this will only prompt to overwrite for the first existing file. If overwriting is enabled, that is assumed to apply to any other existing files. I think that's probably what we want, but open to other thoughts (or suggestions on how to log that more clearly).

This should be safe for a patch release.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -137,6 +137,11 @@ public final class ImageConverter {
   private boolean firstTile = true;
   private DynamicMetadataOptions options = new DynamicMetadataOptions();
 
+  // record paths that have been checked for overwriting
+  // this is mostly useful when "out" is a file name pattern
+  // that may be expanded into multiple actual files
+  private HashMap<String, Boolean> checkedPaths = new HashMap<String, Boolean>();
+
   // -- Constructor --
 
   public ImageConverter() { }
@@ -447,21 +452,9 @@ public final class ImageConverter {
     CommandLineTools.runUpgradeCheck(args);
 
     if (new Location(out).exists()) {
-      if (overwrite == null) {
-        LOGGER.warn("Output file {} exists.", out);
-        LOGGER.warn("Do you want to overwrite it? ([y]/n)");
-        BufferedReader r = new BufferedReader(
-          new InputStreamReader(System.in, Constants.ENCODING));
-        String choice = r.readLine().trim().toLowerCase();
-        overwrite = !choice.startsWith("n");
-      }
-      if (!overwrite) {
-        LOGGER.warn("Exiting; next time, please specify an output file that " +
-          "does not exist.");
+      boolean ok = overwriteCheck(out, false);
+      if (!ok) {
         return false;
-      }
-      else {
-        new Location(out).delete();
       }
     }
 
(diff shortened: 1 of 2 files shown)


dgault:
Builds and tests have continued to be green with these changes and the updated unit tests. Overall the code looks good to me and the strategy of using the same behaviour for all files makes sense to me, having a mixture of overwriting options for different files seems like it would be overkill.

Tested using the suggested command with pattern output: `bfconvert "test&sizeZ=10.fake" "test_z%z.ome.tiff"`

With 6.10.1, when running for the second time with no options there is no prompt for overwriting and the files are appended to the existing files.

With 6.10.1, when running for the second time with the `-nooverwrite` argument, there are no new files created and the original files have been appended to.

With 6.10.1, when running for the second time with the `-overwrite` argument, the same result is seen with no new files created and the original files have been appended to.

With this PR included, when running for the second time with no options, there is a prompt for overwriting and selecting yes will overwrite each of the existing files in the dataset. Selecting no will mean the original files are kept and no new files created.

With this PR included, when running for the second time with the `-nooverwrite` argument, there are no new files created and the original files have not been appended.

With this PR included, when running for the second time with the `-overwrite` argument, the original files have been correctly overwritten with the new outputs.

dgault (approved):

PR #3863, opened 2022-08-12
ND2: remove chunkmap skipping logic

### LINKED ISSUE #3838: ND2: Image stack lost for image with small dimensions
Issue was raised on forum thread https://forum.image.sc/t/can-not-read-nd2-image-stack-when-image-dimension-is-small/68155 but no sample file reproducing the issue is available as of yet.

The issue reports that when opening an image stack (in this case 5 planes) with a small image (50 x 50) then the image stack is lost and only a single plane is displayed. The same issue does not occur if the image is slightly larger (100 x 106). This was reported with Bio-Formats 6.10.0.

melissalinkert:
Fixes #3838.

This logic did not appear to be used by any files in the current data repo, and causes problems for files with very small images.

Opening the file in #3838 (copied to `curated/nd2/gh-3838`) without this PR should show a single plane. Without this PR and with `nativend2.chunkmap` set to `false`, there should be 5 timepoints as expected. With this PR and default options, there should be 5 timepoints as well.

Probably safe for a patch release assuming tests pass.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -545,7 +545,6 @@ public class NativeND2Reader extends SubResolutionFormatReader {
         }
       }
 
-      int chunkmapSkips = 0;
       Boolean currentCountSetted = false;
       int XYCount = 1;
       int timeCount = 1;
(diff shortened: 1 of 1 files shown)


dgault:
Builds and tests have continued to pass with this PR and the associated config PR included. Existing sample files in the repo tests all appear to be unaffected by these changes

Tested using the new sample file `curated/nd2/gh-3838`

Without the PR included and using the default option a single plane is detected which is incorrect. 
Without the PR and setting the `nativend2.chunkmap` option to false results in 5 timepoints which is correct.

With this PR included and using the default options now results in 5 timepoints being detected which is correct. 
With this PR included and setting the `nativend2.chunkmap` option to false results in 5 timepoints which is correct.

dgault (approved):

PR #3867, opened 2022-08-19
- Implement zstd decompression for tiffs.

wimpomp:
Implement zstandard decompression for tif files.

Test tif files compressed with zstd in https://surfdrive.surf.nl/files/index.php/s/BYO0qLnHE9H7fhn, together with original czi images for comparison.

### DIFF
--- a/components/formats-bsd/pom.xml
+++ b/components/formats-bsd/pom.xml
@@ -167,7 +167,11 @@
       <artifactId>snakeyaml</artifactId>
       <version>1.29</version>
     </dependency>
-
+    <dependency>
+      <groupId>io.airlift</groupId>
+      <artifactId>aircompressor</artifactId>
+      <version>0.21</version>
+    </dependency>
   </dependencies>
 
   <properties>
--- a/components/formats-bsd/pom.xml
+++ b/components/formats-bsd/pom.xml
@@ -170,7 +170,7 @@
     <dependency>
       <groupId>io.airlift</groupId>
       <artifactId>aircompressor</artifactId>
-      <version>0.21</version>
+      <version>0.18</version>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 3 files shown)


dgault:
Hi @wimpomp, thank you for contributing this PR to the project. We have introduced a Contributor License Agreement, would you be able to sign and return the form following the instructions on https://ome-contributing.readthedocs.io/en/latest/cla.html

Also do you have any suitable sample files which could be used to test this PR with? If you need a suitable upload location for sample files then we recommend using [Zenodo](https://www.google.com/search?q=zenodo&oq=zeno&aqs=chrome.1.69i57j69i59j46i433i512j46i175i199i512j0i433i512j46i131i175i199i433i512j69i65l2.2345j0j7&sourceid=chrome&ie=UTF-8)

wimpomp:
Hi @dgault, thanks for the reply. I have updated the PR with a link to some test files. Also, I see that some checks have failed, notably because a dependency could not be found. I'll see how I can fix this and try again, any ideas?

dgault:
Thanks @wimpomp, all the checks are passing now, I can also confirm that we recieved your CLA. I have marked the PR for inclusion in our nightly CI builds and repository tests.

dgault:
This PR has been included with the builds and tests for a while now without any issues.
The implementation of the codec looks good from side and the addition of the `aircompressor` dependency causes no issues as we already use the same dependency for the Zeiss CZI reader decompression.

Testing the codec against the new codec shows that the Tiffs are decompressed as expected without any errors and the resulting output matches that of the corresponding CZI files as expected.

We also have a large number of existing CZI sample files with zstd compression, testing these files using the new Codec also showed no issues. Each file opened and displayed without any exception and the resulting output matched that of the current implementation.

Thank you again @wimpomp for contributing the PR to the Bio-Formats project. I will open 2 follow up pr's in the coming days to further make use of the new Codec:
- Replace the existing hardcoded CZI implementation to use the codec
- Migrate the codec to https://github.com/ome/ome-codecs to match the majority of the existing codecs

dgault (approved):

sbesson:
@dgault thinking of the roadmap and given the addition of new classes and enums, would that be rather amenable for a minor release rather than a patch release?

wimpomp:
Thanks for including it!

PR #3868, opened 2022-08-25
Plugins: Close reader after format exception in Importer

dgault:
This is in response to an issue encountered in forum thread https://forum.image.sc/t/exception-issue-when-opening-nd2-files-with-bio-formats/40462/9

When using the BFImport command in a macro, if a FormatException occurred the underlying reader was not properly closed and the file was unable to be renamed.

This will not impact any repo tests and is safe for a patch release.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/Importer.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/Importer.java
@@ -72,32 +72,37 @@ public class Importer {
 
       ImportProcess process = new ImportProcess(options);
 
-      BF.debug("display option dialogs");
-      showDialogs(process);
-      if (plugin.canceled) return;
-
-      BF.debug("display metadata");
-      DisplayHandler displayHandler = new DisplayHandler(process);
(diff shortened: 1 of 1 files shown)


melissalinkert:
Should the reader only be closed when `FormatException` is thrown, or if any exception is thrown?

jburel:
A ``try/catch/finally`` might be more suitable.

dgault:
It really is a unique case when the underlying reader has been created but throws a `FormatException` during the `setID` call. Exceptions before that point won't need to close the reader and exceptions after will be able to close the reader as normal. So this is only dealing with that 1 particular scenario of a FormatException being thrown by the setID call.

melissalinkert (changes requested):
I understand that this is specific to the case where an exception is thrown by `setId`, and that's definitely an improvement. If an exception other than `FormatException` is thrown during `setId` though, shouldn't the reader be closed as well?

melissalinkert (approved):

PR #3869, opened 2022-08-29
CZI: fix pyramid dimension calculation

melissalinkert:
Ported from a private PR.

This addresses a problem where the XY dimensions of a slide were incorrectly set to the XY dimensions of a single tile. Aside from just being wrong, the sizes were small enough that smaller resolutions were calculated to have 0 pixels, which caused other errors.

I expect this to have limited impact on existing datasets; a configuration PR will be opened shortly.


sbesson (approved):
Tested against the handful of sample CZI files affected by this change. The image dimensions now match the expectations from ZEN, the pixel data itself is unaffected except for the position of the right and bottom borders which are slightly modified

dgault (approved):
As with Seb, reviewing the impacted datasets and the adjusted values are matching those shown in Zen. With the config PR included all the builds and tests have remained green for quite some time now without any regressions in other datasets. Looks good to merge.

PR #3874, opened 2022-09-15
FV1000: Use DataTools to parse Double values

dgault:
This PR is in response to https://forum.image.sc/t/cannot-open-oib-files-in-imagej/71616/5
Sample files have been provided which use a different Locale format for the double values.

With the current release (6.10.1) this results in an exception as below:

```
java.lang.NumberFormatException: For input string: “211,761.0”
at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2043)
at sun.misc.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
at java.lang.Double.parseDouble(Double.java:538)
at loci.formats.in.FV1000Reader.initFile(FV1000Reader.java:498)
at loci.plugins.in.ImportProcess.initializeFile(ImportProcess.java:499)
at loci.plugins.in.ImportProcess.execute(ImportProcess.java:142)
at loci.plugins.in.Importer.showDialogs(Importer.java:140)
```

With this PR there should be no exception and the files open and display without error.
I will add a new config for the sample files from the imagesc thread.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/FV1000Reader.java
@@ -495,8 +495,8 @@ public class FV1000Reader extends FormatReader {
       code[i] = commonParams.get("AxisCode");
       size[i] = commonParams.get("MaxSize");
 
-      double end = Double.parseDouble(commonParams.get("EndPosition"));
-      double start = Double.parseDouble(commonParams.get("StartPosition"));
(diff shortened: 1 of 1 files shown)


sbesson (approved):
The usage of `DataTools` parsing API should be more defensive in handling different formats. Tested using OMERO with this PR included and the metadata is correctly parsed.

Screenshot 2022-09-20 at 11 20 57

PR #3876, opened 2022-09-20
Add ReportEnabledStatus class to list files that aren't tested

melissalinkert:
Usage right now is something like:

```
$ java -Dtestng.directory=/path/to/data -Dtestng.configDirectory=/path/to/config loci.tests.testng.ReportEnabledStatus --not-enabled
```

This should help with https://github.com/openmicroscopy/data_repo_config/issues/534.

I haven't turned this into an actual TestNG test or ant task (do we want that?), but in theory at least this could be run by Jenkins just like repo tests. Not expecting this to be merged as-is, it's just a place to start.


melissalinkert:
Removed draft status as discussed in today's formats meeting. The next steps are:

- verify that tests continue to pass with this included
- write a wrapper script or something to make this easier to run

melissalinkert:
This does not appear to have broken anything so far.

329045d is one way to turn this feature into a more usable tool. The commit message has details, but I would expect something like:

```
$ unzip bftools.zip
$ cp bio-formats-testing-framework.jar bftools/
$ echo $BF_DEVEL

$ bftools/unconfigured $DATA/curated/mng/ $CONFIG/curated/mng/
```

to print a list of the 7 unconfigured .mng files (other directories can be used, that's just one example). We can debate whether the script belongs in bftools.zip, if it makes sense to include the test jar in bftools.zip, etc. - so I wouldn't expect this to be merged in the current state.

In doing this I also noticed that there was a previous attempt at this feature in `components/test-suite/utils`. As it didn't even read the current version of the config files, ecae467 removes it to avoid confusion.

joshmoore:
Big :+1: in general for the addition of a wrapper in https://github.com/ome/bioformats/commit/329045d7464edaea3a86baeeea62c1feda9e771f but a slight heads-up that at least on my system these commands end up on the default `$PATH` and therefore we might want to consider, e.g., starting to prefix everything with `bf-`. Then as with git tools, we could consider having a single entrypoint like `bf` which could take `bf unconfigured` or provide general help, etc.

melissalinkert:
> Big 👍 in general for the addition of a wrapper in [329045d](https://github.com/ome/bioformats/commit/329045d7464edaea3a86baeeea62c1feda9e771f) but a slight heads-up that at least on my system these commands end up on the default `$PATH` and therefore we might want to consider, e.g., starting to prefix everything with `bf-`. Then as with git tools, we could consider having a single entrypoint like `bf` which could take `bf unconfigured` or provide general help, etc.

See https://github.com/ome/bioformats/issues/3998 and 8ffb35d.

joshmoore:
Thanks, @melissalinkert, and :+1: for the issue.

dgault (approved):
I don't see any issue with having this included as part of bftools, and bf-unconfigured seems like a suitable name to me.

Testing, both directly through ReportEnabledStatus and through the new bf-unconfigured correctly listed the  correct files as expected.

PR #3886, opened 2022-10-06
Make logback dependencies optional

ctrueden:
At runtime, the `ch.qos.logback` classes are only accessed via reflection.

Specifically, the IJStatusEchoer class is used only by ome-common (!) [in a `ReflectedUniverse` block](https://github.com/ome/ome-common-java/blob/v6.0.13/src/main/java/loci/common/DebugTools.java#L160-L182).

It is best practice for libraries not to impose any specific SLF4J binding on downstream consumers.

> Embedded components such as libraries or frameworks should not declare a dependency on any SLF4J binding but only depend on slf4j-api.

[Source](https://www.slf4j.org/manual.html#libraries)

### DIFF
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -72,11 +72,13 @@
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-core</artifactId>
       <version>${logback.version}</version>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <version>${logback.version}</version>
+      <optional>true</optional>
     </dependency>
 
     <!-- NB: dependency:analyze has false warning about xml-apis:xml-apis. -->
--- a/components/bio-formats-tools/pom.xml
+++ b/components/bio-formats-tools/pom.xml
@@ -67,6 +67,7 @@
       <artifactId>logback-core</artifactId>
       <version>${logback.version}</version>
       <scope>runtime</scope>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <!-- NB: We want this, despite warning from dependency:analyze. -->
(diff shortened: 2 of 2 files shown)


dgault:
Thanks Curtis for opening the PR. You are correct, marking the dependency is the correct solution as the library shouldn't impose the binding on downstream consumers. The only remaining instances of the dependency are limited to the testing scope and should be fine as is. 

I will try to review any impact this may have on other downstream clients so that the change can be properly communicated to them and we will look at trying to get this PR and https://github.com/ome/ome-common-java/pull/68 included for an upcoming release.

ctrueden:
Thanks @dgault. With this change in place, we will be able to remove the corresponding exclusions from the pom-scijava dependency management of these components.

dgault:
Moving this PR to the 6.12.0 milestone to give more time to test downstream applications

sbesson (commented):
Trying to build this in combination with https://github.com/ome/ome-common-java/pull/68 locally and running `mvn dependency:tree`, logback is definitely declared as optional in high-level components such as `bio-formats-plugins`

```
[INFO] +- ch.qos.logback:logback-core:jar:1.2.9:compile (optional) 
[INFO] +- ch.qos.logback:logback-classic:jar:1.2.9:compile (optional) 
```

However, the `bioformats_package` bundle still declares it as

```
[INFO] |  +- ome:formats-gpl:jar:6.12.0-SNAPSHOT:compile
[INFO] |  |  +- org.openmicroscopy:ome-mdbtools:jar:5.3.2:compile
[INFO] |  |  +- org.openmicroscopy:metakit:jar:5.3.4:compile
[INFO] |  |  |  \- ch.qos.logback:logback-classic:jar:1.2.9:compile
[INFO] |  |  |     \- ch.qos.logback:logback-core:jar:1.2.9:compile
```
Could you also update https://github.com/ome/ome-metakit/blob/623cb7435b2bc098d193ab0eb54cf2ae0773a893/pom.xml#L139-L143 to declare the correct scope?

dgault (approved):

PR #3887, opened 2022-10-10
PCI: Parse units for physical size

dgault:
This is a follow up to the original PR https://github.com/ome/bioformats/pull/3880

Comparing the data repo updates to the equivalent HCImage  data suggests that the magnification value should not be used for physical size when the units are Pixels. https://github.com/openmicroscopy/data_repo_config/pull/539

This may mean further repo updates if there are other datasets with pixel units.


melissalinkert:
Code changes make sense to me, I think this is fine to merge once configuration issues are sorted out and tests are confirmed to pass.

dgault:
Merge CI is now green without the unit changes. Getting this merged for 6.11.0 and we can follow up on the units addition for the next release

Issue #3607, opened 2020-07-15
nii vs nii.gz


dominikl:
Importing  unzipped `nii` files works fine, whereas the same files in their original `nii.gz` form take so long that it's practically not possible to import. See https://forum.image.sc/t/importing-large-neuroimaging-data-into-omero/40114 
I'm capturing the problem here, because I'm not sure yet which component is actually causing this.

sbesson:
A few additional comments here:

- support for gzipped compressed NIfTI data is relatively recent (it was added via an external contribution in https://github.com/ome/bioformats/pull/2508) and has primarily been tested with the sample NIH datasets (https://nifti.nimh.nih.gov/nifti-1/data). Thus I am not completely surprised by the performance degradation for large volumes as we hold no representative sample of real-world large `.nii.gz`
- the reader logic uses the filename to determine which handle should be used to stream the file which probably explain why the trick suggested in https://github.com/ome/bioformats/issues/3607#issuecomment-671648632 did not work
- if conversion is an option and the NifTi format is not a requirement else where in the pipeline, it might be worth testing the conversion into e.g. zlib-compressed tiled OME-TIFF using either the command-line tools and/or the [bioformats2raw/raw2ometiff pipeline](https://forum.image.sc/t/converting-whole-slide-images-to-ome-tiff-a-new-workflow/32110) as an alternative
- to fix the issue at the reader level, some profiling might help identifying which calls are causing the bottleneck although I am pretty  likely this happens during compression/decompression. Does anyone have experience with how this problem is solved in reference NifTi libraries or other software? Is there some expected benchmark metric for the random access to a file under `.nii` or `.nii.gz` format?

dgault:
Was running some quick profiling on this. Opening a small .nii.gz resulted in a very long read time, caused by a seek call for each openBytes:

<img width="836" alt="Screen Shot 2020-08-19 at 17 45 52" src="https://user-images.githubusercontent.com/1628865/90665890-8f6a8e00-e244-11ea-8797-ab724ebfa8d3.png">
Looking at the code I retested after the removing the below line which looks unnecessary:
https://github.com/ome/bioformats/pull/2508/files#diff-1696608d5491b8873c77d696b4f4cc4aR150

Without that line the same file then produced the below times and still loaded as expected:

[...]

PR #3891, opened 2022-10-18
NIFTI: reduce number of backwards seeks

melissalinkert:
May improve performance for gzip-compressed files. See #3607.

This seems to improve `openBytes` performance during `showinf`, and tests continue to pass locally. This should be OK for a patch release, if deemed useful (otherwise feel free to close).

`showinf` or really anything else is still going to be pretty slow to initialize `.nii.gz` files though. A lot of that comes down to `GZipHandle` needing to calculate the length of the uncompressed stream each time a new handle is opened, which happens several times during type checking in `ImageReader`. The only way to do the length calculation is to skip through the stream and count the number of bytes, then reset. I looked into ways to make better use of the gzip format's internal metadata, but unfortunately the only uncompressed size recorded is the original file length mod 2^32 (`ISIZE`, see https://datatracker.ietf.org/doc/html/rfc1952#page-8). That's not particularly helpful when we know we can have uncompressed sizes larger than 4 GB.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NiftiReader.java
+++ b/components/formats-gpl/src/loci/formats/in/NiftiReader.java
@@ -28,6 +28,7 @@ package loci.formats.in;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 
+import loci.common.DataTools;
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.formats.CoreMetadata;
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested functionally using `showinf` and the sample file provided in the original forum post.

Without these changes, both the reader initialisation and the loading of every single plane are extremely slow. With these changes, the reader initialised in 91.068s and the planes loaded in 3.842s (0.118580244ms per plane). The viewing of the image is then smooth.

Also tested in the context of OMERO and confirmed that passed the initialisation step, pixel data access is vastly improved. Without this PR, `omero fs importtime` reports

```
   upload time of   6.55s for 1 file (6.555s/file)
    setId time of  99.28s
 metadata time of   4.08s
   pixels time of 67037.61s for 32400 planes (2.069s/plane)
    rdefs time of  96.98s for 1 rendering setting (96.979s/rdef)
thumbnail time of   9.47s for 1 thumbnail (9.474s/thumbnail)
```

With this PR

```
   upload time of   2.73s for 1 file (2.730s/file)
    setId time of  96.21s
 metadata time of   1.54s
   pixels time of   9.09s for 32400 planes (0.000s/plane)
    rdefs time of  96.16s for 1 rendering setting (96.162s/rdef)
thumbnail time of   9.53s for 1 thumbnail (9.528s/thumbnail)
```

When accessing the data via the Web client, every request triggers a `setId` call under the hood which takes ~10s even with the memo file. 

Overall, this feels like a stepwise improvement of the experience with gzipped compressed NIfTI files which primarily applies in the scenarios where the pixel data is being accessed from an initialised reader e.g. in a conversion workflow.

PR #3892, opened 2022-10-20
QPTIFF: preserve biomarker data in channel name

melissalinkert:
Backported from a private PR.

If a biomarker is set in the original metadata, it will now be saved as the `Name` on `Channel`. The previous name will be saved as `Fluor` on `Channel`, as this is a better match for what the name attribute actually is.

If no biomarker is present, there should be no change in behavior. This seems to be the more common case, as I don't believe we have any example files in the repo that contain a biomarker. The files used to test the private PR unfortunately can't be shared.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/VectraReader.java
+++ b/components/formats-gpl/src/loci/formats/in/VectraReader.java
@@ -476,6 +476,11 @@ public class VectraReader extends BaseTiffReader {
 
           if (name.equals("Name")) {
             for (int series=0; series<omeImageCount; series++) {
+              store.setChannelFluor(value, series, c);
+            }
+          }
+          else if (name.equals("Biomarker")) {
+            for (int series=0; series<omeImageCount; series++) {
               store.setChannelName(value, series, c);
             }
           }
--- a/components/formats-gpl/src/loci/formats/in/VectraReader.java
+++ b/components/formats-gpl/src/loci/formats/in/VectraReader.java
@@ -366,6 +366,9 @@ public class VectraReader extends BaseTiffReader {
     int omeImageCount = hasFlattenedResolutions() ? pyramidDepth : 1;
 
     for (int c=0; c<getEffectiveSizeC(); c++) {
+      String storedChannelName = null;
+      String biomarker = null;
+
       String xml = getIFDComment(c);
       try {
         Element root = XMLTools.parseDOM(xml).getDocumentElement();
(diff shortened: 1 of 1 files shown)


sbesson (approved):
As indicated by the daily CI builds, we have currently no representative sample of this QPTIFF variant in the curated QA repository and this PR has not been modifying the metadata of our collection on QPTIFF files.

Relevant context can be found in the Akoya QPTIFF Specification which is available in Appendix B of the [PhenoChart User Manual](https://www.akoyabio.com/wp-content/uploads/2021/01/PhenochartUserManual_1_1_0_rev0.pdf). More specifically, the following rows in page 55 are most relevant

| Tag       | Optional ? | Contents                                                                                                |
|-----------|------------|---------------------------------------------------------------------------------------------------------|
| Name      | Y          | Band name for FL whole slide scans, component name for extracted components. Not present for RGB images |
| ...       |            |                                                                                                         |
| Biomarker | Y          | String indicating target such as CD8, Ki67, etc.                                                        |

When available, the `Biomarker` attribute will typically contain some metadata that users might want to see as the primary name to identify the channel. In that case, the `Name` attribute if it exists will still be stored.

As mentioned on Monday, in the absence of sample, it should be possible to craft synthetic examples derived from our [public samples](https://downloads.openmicroscopy.org/images/Vectra-QPTIFF/) if that gave us more confidence in this change and was deemed valuable to prevent future regressions.

components/formats-gpl/src/loci/formats/in/VectraReader.java
```diff
@@ -538,6 +542,25 @@ else if (name.equals("ExposureTime")) {
...
+      boolean validName = storedChannelName != null && !storedChannelName.isEmpty();
+      for (int series=0; series<omeImageCount; series++) {
+        if (validBiomarker) {
+          store.setChannelName(biomarker, series, c);
+          if (validName) {
+            store.setChannelFluor(storedChannelName, series, c);
+          }
+        }
+        else {
+          store.setChannelName(storedChannelName, series, c);
```
jburel, line 561: an empty string could be stored. Make check if it is a ``validName`` too.

melissalinkert: I intentionally didn't do that, so that the behavior would be unchanged in the common case where biomarker data is missing. If you're OK with the behavior changing in that case, I'll switch line 560 to `else if (validName)` (which will mean that `setChannelName` isn't guaranteed to be called for every channel).

dgault (approved):
Based on the QPTIFF spec this change looks like it makes sense, it has no impact on any of the existing files in our repo, if we receive any samples in the future we will want to add them to the repo tests also.

PR #3893, opened 2022-10-24
switch to new output command

jburel:
Check that the build is green
Background https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


jburel:
Green

PR #3897, opened 2022-10-24
Olympus vsi: fix exposure time handling

### LINKED ISSUE #3642: VSI: incorrect exposure times
This issue was raised in forum thread https://forum.image.sc/t/bio-formats-exposure-time-incorrect-when-importing-vsi-files/46080

In the OME-XML the Exposure Time is the same for each plane. However the correct values can be found in the original metadata tags as shown in the images attached in the imagesc thread.

melissalinkert:
Exposure times can be recorded with no tag prefix or with the `Microscope ` tag prefix. Some datasets have just one or the other, some have both. The times with no prefix are preferred if present, but the `Microscope ` ones are now used if that is all that is available.

Fixes #3642.

Compare `showinf -nopix -omexml -noflat` on QA 29751 with and without this PR.

This should be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -899,6 +899,9 @@ public class CellSensReader extends FormatReader {
               if (c < pyramid.exposureTimes.size()) {
                 exp = pyramid.exposureTimes.get(c);
               }
+              else if (c < pyramid.otherExposureTimes.size()) {
+                exp = pyramid.otherExposureTimes.get(c);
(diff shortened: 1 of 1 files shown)


dgault:
Logic here looks sensible and looks good from my side.

Testing with the sample file reported in QA-29751, with the PR included the exposure time is now correctly reporting the expected values.

Also tested the `olivier/small-tiles` sample file from the associated config PR and again with the PR included the exposure time now matches that displayed in the OlyVIA software. 

All other remaining metadata values are unchanged and the builds and tests have been green with this included.

PR #3898, opened 2022-10-28
InCell 1000/2000 field count and plane metadata fixes

### LINKED ISSUE #3895: InCellReader: ArrayIndexOutOfBoundsException
See [QA 31270](https://www.openmicroscopy.org/qa2/qa/feedback/31270/)

```
java.lang.ArrayIndexOutOfBoundsException: 0
	at loci.formats.in.InCellReader$MinimalInCellHandler.startElement(InCellReader.java:885)
	at org.apache.xerces.parsers.AbstractSAXParser.startElement(Unknown Source)
	at org.apache.xerces.parsers.AbstractXMLDocumentParser.emptyElement(Unknown Source)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanStartElement(Unknown Source)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(Unknown Source)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanDocument(Unknown Source)
	at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
	at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
	at org.apache.xerces.parsers.XMLParser.parse(Unknown Source)
	at org.apache.xerces.parsers.AbstractSAXParser.parse(Unknown Source)
	at org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser.parse(Unknown Source)
	at org.apache.xerces.jaxp.SAXParserImpl.parse(Unknown Source)
	at javax.xml.parsers.SAXParser.parse(SAXParser.java:195)
	at loci.common.xml.XMLTools.parseXML(XMLTools.java:551)
	at loci.common.xml.XMLTools.parseXML(XMLTools.java:533)
	at loci.formats.in.InCellReader.initFile(InCellReader.java:378)
...
```

melissalinkert:
Fixes #3895 and https://trello.com/c/gFmUKSp6/362-incell-exposure-times

`showinf -nopix -omexml` on the .xdce file in QA 31270 should show an exception similar to what is noted in #3895. With this PR, the same command should show 8 series, each of which is a field in well A1.

That's almost but not quite correct with respect to what's in the .xdce. The .xdce defines 8 fields, but each field is only present in one of 3 wells (A1, A2, A3). So 8 total series makes sense, but the `Well`/`WellSample` mapping does not. We would need to decide how to best represent sparse fields here - does 24 series (16 of which are blank) make sense instead? Or is it better to have 8 series and ignore the fact that they are 8 distinct fields?

QA 31270 also exposes the exposure time issue noted in https://trello.com/c/gFmUKSp6/362-incell-exposure-times, so the last 2 commits here address that problem as well. This requires some configuration updates, which are in a forthcoming PR.

Opening as draft for now since this requires some discussion around sparse field handling.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -1026,8 +1026,12 @@ public class InCellReader extends FormatReader {
         int z = Integer.parseInt(attributes.getValue("z_index"));
         int c = Integer.parseInt(attributes.getValue("wave_index"));
         int t = Integer.parseInt(attributes.getValue("time_index"));
+        int channels = channelsPerTimepoint.get(t).intValue();
+
         currentImage = t;
-        currentPlane = z * getSizeC() + c;
+        currentPlane = FormatTools.getIndex("XYZCT", getSizeZ(),
+          channels, 1, getSizeZ() * channels, z, c, 0);
+
         int well = currentRow * wellCols + currentCol;
         Image img = imageFiles[well][currentField][currentImage][currentPlane];
         if (img != null) {
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -898,7 +898,7 @@ public class InCellReader extends FormatReader {
           lastImage.max = DataTools.parseDouble(attributes.getValue("max"));
         }
       }
-      else if (qName.equals("offset_point")) {
+      else if (qName.equals("offset_point") || qName.equals("Point")) {
         fieldCount++;
       }
       else if (qName.equals("TimePoint") && doT) {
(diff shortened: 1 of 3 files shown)


sbesson (commented):
Unless there is a data rationale, my preference would be to avoid injecting artificial blank images especially with HCS data where we should be able to handle sparse data. In addition to the two solutions proposed in the description, a third option could be for the reader to generate a representation as close as possible to the content of the xdce:
```
Image 0 <-> WellRow 0, WellCol 0, WellSample 0
Image 1 <-> WellRow 0, WellCol 0, WellSample 1
Image 2 <-> WellRow 0, WellCol 1, WellSample 2
Image 3 <-> WellRow 0, WellCol 1, WellSample 3
Image 4 <-> WellRow 0, WellCol 1, WellSample 4
Image 5 <-> WellRow 0, WellCol 2, WellSample 5
Image 6 <-> WellRow 0, WellCol 2, WellSample 6
Image 7 <-> WellRow 0, WellCol 2, WellSample 7
```

We do have a few examples of sparse plates in the sense that the number populated wells varies as a function of the field index - see for instance https://idr.openmicroscopy.org/webclient/?show=plate-5563. It would be interesting to have a feeling of  whether our clients can handle the extreme scenario of this InCell dataset. 

If we wanted to make a comparative review of different layouts, would it be useful to try and export the source dataset into a set of various OME-XML/OME-TIFF samples that could be imported?

melissalinkert:
Closing as discussed today with @ome/formats.

PR #3902, opened 2022-11-04
LSID validation

melissalinkert:
Backported from a private PR.

This is intended to make it easier to set valid IDs when the file format's metadata provides an ID that might be usable. This won't apply to most formats, but Zeiss CZI (as demonstrated here) and a few other readers that store OME-XML-like metadata could use this feature.

The new `MetadataTools.createLSID(String, String, int...)` signature checks the ID stored in the file's metadata and returns it if it matches the expectations of https://github.com/ome/ome-model/blob/46a69d3b3478997c80fb4c4928624494bf54f371/specification/src/main/resources/released-schema/2016-06/ome.xsd#L3248-L3252. Otherwise, it delegates to the original `MetadataTools.createLSID(String, int...)` signature to create a new ID.

The .czi file that prompted this change is private, but the Dichroic ID stored in the file was `BeamsplitterLsm.2412-296`. Without this change, `showinf -nopix -omexml` showed OME-XML validation errors. With this change, there were no more validation errors as the Dichroic ID was set to `Dichroic:0:0:BeamsplitterLsm.2412-296` in the OME-XML.

### DIFF
--- a/components/formats-api/src/loci/formats/MetadataTools.java
+++ b/components/formats-api/src/loci/formats/MetadataTools.java
@@ -927,4 +927,27 @@ public final class MetadataTools {
     }
   }
 
+  /**
+   * Get a valid LSID based on a stored ID.
+   * The returned ID is expected to match the "(\S+):(\S+)" pattern.
+   * If the srcID matches this pattern already, it is returned intact.
+   * If the srcID is null or empty, the altID (expected to be the result of
+   * calling createLSID(...)) is returned instead.
+   * If the srcID is not valid but also not empty, then the srcID is appended
+   * to the altID and separated by ":".
+   *
+   * This is mainly intended for use with formats that store metadata in a
+   * structure similar to OME-XML. The ID recorded in the metadata may not be
+   * quite valid, but is worth saving if possible.
+   */
+  public static String sanitizeID(String srcID, String altID) {
+    if (srcID == null || srcID.isEmpty()) {
+      return altID;
+    }
+    if (srcID.indexOf(":") >= 0) {
+      return srcID;
+    }
+    return altID + ":" + srcID;
+  }
+
 }
--- a/components/formats-api/src/loci/formats/MetadataTools.java
+++ b/components/formats-api/src/loci/formats/MetadataTools.java
@@ -336,6 +336,28 @@ public final class MetadataTools {
   }
 
   /**
+   * Construct a valid LSID based on a stored ID.
+   * The returned ID is expected to match the "(\S+):(\S+)" pattern.
+   * If the srcID matches this pattern already, it is returned intact.
+   * If the srcID is null or empty, this falls back to createLSID(String, int...)
+   * If the srcID is not valid but also not empty, then the srcID is appended
+   * to the return value of createLSID(String, int...) and separated by ":".
+   *
+   * This is mainly intended for use with formats that store metadata in a
+   * structure similar to OME-XML. The ID recorded in the metadata may not be
+   * quite valid, but is worth saving if possible.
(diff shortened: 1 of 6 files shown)


sbesson (commented):
Caputring various thoughts on this API extension and the underlying question of preserving original identifiers. Some of these comments are also related to a background investigation of ongoing indexing issues with reading OME-XML based formats. There are clear trade-offs with all options and I have not convinced myself of a clear path forward  but hopefully, this can be used as a starting point for the discussion.

Bio-Formats internally makes extensive use of `MetadataTools.createLSID` to create unique valid object identifiers. These identifiers typically following a simple and explicit naming scheme `<Object>:<index>[:<index>[:<index>]]]` where the indices match the values you would pass to the `MetadataStore/MetadataRetrieve` API. `MetadataTools.populatePixels` will typically use this scheme for all `Image/Pixels/Channel` object. One thing I have not reviewed is whether additional APIs e.g. in `MetadataTools` rely on this naming scheme. If that was the case, departing from this convention could lead to metadata incompatibilities and we might want to clarify this in the documentation.

On the other hand, preserving proprietary metadata at least in the OriginalMetadata and ideally in the OME metadata adheres with the Bio-Formats principles of metadata handling.

When the proprietary metadata includes identifiers fully compliant with the schema e.g. `Dichroic:<value>`, I think the trade-offs are as follow:
 -  pros: using valid identifiers as `ID` values a priori look like the best way to preserve original metadata. Additionally, there are scenarios where these identifiers might be cross-referenced elsewhere in the metadata (e.g. `WellSampleRef/ImageID`). At the moment our readers have no logic for updating these references with its own generated identifiers which can lead to metadata corruption under some conditions
 - cons: the primary benefit of the handling described in 2. is that it ensures some level of consistency in the OME metadata generation. Blindly inheriting original identifiers could also lead to broken scenarios under some conditions e.g. the creation of objects with duplicate identifiers

When the proprietary metadata includes identifiers which do not match our schema expectation, the current PR proposes to support the appending of the original metadata to the generated ID. This looks reasonable as the XSD restriction is typically `<Object>:\S+` or its LSID equivalent. If we agree with this API, it might be worth adding an additional check ensuring `srcID` does not contain whitespaces

Thinking of other approaches to preserve the metadata, the most immediate one coming to ming would be to annotate the relevant object. The typical disadvantage is that most consumers have at best minimal support for such annotation and under many conditions this would be equivalent to hiding this information.

joshmoore:
I'm certainly no expert on what the possible assumptions on the LSID strings here are in existing consuming code (and it doesn't help that the LSID spec has been replaced by 2 or 3 _generations_ of other identifiers in the meantime...) but looking at this naively I'd say introducing a type that gets instantiated even if just internally would help to bundle the logic and perhaps let us write tests. https://github.com/ome/omero-model/blob/master/src/main/java/ome/util/LSID.java ([tests](https://github.com/ome/omero-model/blob/master/src/test/java/ome/util/utests/LSIDEquivalenceTest.java)) is somewhat in this vein.

melissalinkert:
This needs more discussion and potentially changes to the readers listed in https://github.com/ome/bioformats/pull/3902#issuecomment-1359694831 before I'd be OK with this being merged. Moving out of 6.12.0, and we can reconsider for the following non-patch release.

melissalinkert:
Discussed a bit at today's formats meeting. Next step is to update other readers as noted in https://github.com/ome/bioformats/pull/3902#issuecomment-1359694831 and then take out of draft for minor release after 6.13.0.

melissalinkert:
Closing as discussed in today's formats meeting.

PR #3908, opened 2022-11-11
Operetta/Harmony: ignore any 32-bit TIFF files

melissalinkert:
Backported from a private PR.

Some Operetta/Harmony/etc. plates will have one or more 32-bit TIFF files essentially at random. These are not recorded as being any different in the `Index.idx.xml` file. We have been told that these files contain invalid data (even if it "looks OK"), so a blank plane (all 0s) should be returned in place of the 32-bit data.

An artificial example is being uploaded to `curated/perkinelmer-operetta/samples/32-bit-file` (see the readme for how it was created). Without this PR, `showinf -series 2 Index.idx.xml` should throw an exception during `openBytes`. With this PR, the same command should show a warning to indicate that one of the files is 32-bit, but should not throw an exception. The first plane in the series should be all 0s.

Additional reports of this issue are in https://forum.image.sc/t/post-processing-of-plates-in-omero/53700.

I would expect this to be safe for a patch release (assuming tests pass), but am assigning to 6.12.0 so as not to overload 6.11.1.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -241,7 +241,11 @@ public class OperettaReader extends FormatReader {
           return buf;
         }
 
-        if (reader.getSizeX() >= getSizeX() && reader.getSizeY() >= getSizeY()) {
+        if (reader.getPixelType() != getPixelType()) {
+          throw new FormatException("Pixel type mismatch in " + p.filename +
+            " (got " + FormatTools.getPixelTypeString(reader.getPixelType()) + ")");
+        }
+        else if (reader.getSizeX() >= getSizeX() && reader.getSizeY() >= getSizeY()) {
           reader.openBytes(0, buf, x, y, w, h);
         }
         else {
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Using the synthetic sample file mentioned above, I can confirm `openBytes` fails on the third series without this PR included. An OMERO import leads to an import failure during the `ManagedImportRequestI.pixelData` step (and unfortunately leaves the plate in a broken unrecoverable state) as the planes are being parsed to compute the min/max values.

With this PR included, `openBytes` no longer throws an exception and the plate imports successfully. The plane corresponding to the HOECHST 33342 channel of the third series (`Well E04, Field 1`) renders black as expected by the change.

The new sample fileset is successfully being tested by the daily repository tests

NicoKiaru:
Just to comment on that: the 32 bits arise when the values in the tiffs overshoot 65 536. This typically happens with digital phase contrast channel, where images are the result of a computation between two acquired bright images.

In a previous script we made, we specified an optional rescale:

https://github.com/BIOP/biop-operetta-tools/blob/ae675e108c8aecf77f0d08450a512941969f9fc9/deprecated/BIOP_Operetta_Import.groovy#L289-L301

I guess it's fine to drop the 32 bits images, but I don't think it's invalid data

PR #3924, opened 2022-12-15
Fixes Null Pointer Exception in case no plane is present in OperettaReader

NicoKiaru:
Using this xml file: https://drive.google.com/file/d/1JTfk48qBCQQdb5_jIrZzLKLfPjFt_C7r/view?usp=share_link
the Operetta Reader creates a null pointer exception on the line:

filename = planes[i][planeIndex].filename;

This PR fixes the issue by creating an empty plane. 

Maybe this bug was introduced with my previous PR on Operetta performance improvement (can't find the previous PR link, sorry)

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -438,6 +438,9 @@ public class OperettaReader extends FormatReader {
                 String key = rows[row]+":"+cols[col]+":"+fields[field]+":"+cs[c]+":"+zs[z]+":"+ts[t];
                 if (hashToPlane.containsKey(key)) {
                   planes[nextSeries][nextPlane] = hashToPlane.get(key);
+                } else {
+                  LOGGER.warn("No plane found for key {}",key);
+                  planes[nextSeries][nextPlane] = new Plane();
                 }
                 nextPlane++;
               }


melissalinkert:
For internal testing, the file is now in `inbox/gh-3924`.

@NicoKiaru, is it correct to say this fix is only for the case when no TIFF file exists at all? If so, would it make more sense to throw `FormatException` in `setId` rather than produce a whole plate of empty planes with int8 pixel type?

dgault (changes requested):
Hi @NicoKiaru, thank you again for contributing this fix. Unfortunately this PR has caused some failures in our repository tests for some existing datasets. 

The particular files impacted also have a large number of missing planes, however with this PR some of the channel metadata for those files that was previously populated has now been lost. Looking at the reader this appears to due to logic put in previously (likely for those very files) to handle channel metadata specifically when missing planes are present.

In that case the code below is using a check of `plane == null` to identify the missing planes:
https://github.com/ome/bioformats/blob/14fc1dd6c706798c31582ffad2b2c610d3c6fda0/components/formats-gpl/src/loci/formats/in/OperettaReader.java#L635-L648 

So there are 2 main fixes I can see to address both the above scenario and the bug you are seeing:

- Keep this fix to create empty planes, but additionally modify the channel logic to detect an empty plane rather than null plane
- Keep the existing logic of using null planes and add extra null pointer checks for https://github.com/ome/bioformats/blob/14fc1dd6c706798c31582ffad2b2c610d3c6fda0/components/formats-gpl/src/loci/formats/in/OperettaReader.java#L489

NicoKiaru:
Ok, thanks for the report. I will try to collect the tiff data associated to the failing xml file, will share it with you and try to dig further.

NicoKiaru:
I'm lagging too far behind. I will close this PR and open a new clean one

PR #3925, opened 2022-12-21
Olympus .vsi: only read pixels from frame_*.ets files

### LINKED ISSUE #3859: VSI: Exception loading files from VS200 scanner with software version 3.4.1
Issue has been reported by multiple users on forum thread https://forum.image.sc/t/vsi-format-update-unable-to-open-new-files/70366/5

This appears to be recently occurring after performing a software update for VS200 scanner to software version 3.4.1

I can reproduce the issue with the sample dataset provided and standalone Bio-Formats 6.10.0. I have not got to root of the issue but it looks to be due to Bio-Formats finding more available files than it was expecting to. My initial though is that it may be related to the blob_21_f_Frame#0 files found in the stack10000 folder, moving those out of the fileset seems to allow for the image to be opened. 

I have not confirmed what these additional files relate to, though there is a `blob_21_f.meta file` included. This looks likely to a be an update to the proprietary VSI spec as a result of the software update.

melissalinkert:
Backported from a private PR. This is expected to fix #3859.

~`inbox/gh-3859`~`curated/cellsens/gh-3859` can be used to test. I may have additional test datasets, but am waiting for confirmation that they can be shared.

Without this change, `showinf -nopix -noflat` on the .vsi file should throw an exception. With this change, usual `showinf` commands, opening in ImageJ, etc. should display images and metadata without error. The `blob*.ets` and `blob*.meta` files should be included at the end of the used files list.

I would expect this to impact memo files, so requires a minor release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -376,6 +376,8 @@ public class CellSensReader extends FormatReader {
   // -- Fields --
 
   private String[] usedFiles;
+  // non-pixels files that are included in the directory (e.g. "blobs")
+  private ArrayList<String> extraFiles = new ArrayList<String>();
   private HashMap<Integer, String> fileMap = new HashMap<Integer, String>();
 
   private TiffParser parser;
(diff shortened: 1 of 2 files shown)


dgault (approved):
All builds and tests have been green with this PR and the config PR's included. PR does not show any regressions or behaviour changes in existing datasets.

Tested using the newly provided sample files. Without this PR and testing `curated/cellsens/gh-3859` results in an exception as below:

```
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 3, Size: 3
```

With the PR included there is no exception and the file can be opened and displayed without issue. The list of included files also looks to be correct.

Testing the additional new samples from `curated/cellsens/public/evident` and each opens and displays without issue, matching the screenshots provided from OlyVIA.

PR #3926, opened 2023-01-03
VSI: respect tile origin metadata

### LINKED ISSUE #3822: VSI: Cropped image due to incorrect offsets
Issue was raised on forum thread https://forum.image.sc/t/bio-formats-accidental-cropping-of-olympus-images/55689 and can be reproduced with the sample file provided.

The dimensions read by Bio-Formats are correct however the image appears to be shifted due to some zero padding on the left hand side resulting in the right hand side appearing to be cropped. This is likely due to the offsets being used by Bio-Formats being incorrect, when the user tried converting the image to tiff the result produced `invalid page offset` errors:

```
>>>image = skimage.io.imread("image.tif")
<tifffile.TiffPages @8> invalid page offset 90397154
```

melissalinkert:
Fixes #3822.

Without this PR, opening the .vsi file in `curated/cellsens/gh-3822` in QuPath or ImageJ vs. Olympus' OlyVIA shows a subtle difference between the left and right edges, as noted in the initial issue report. The associated .jp2 file visually approximates what OlyVIA shows, but cannot be used for quantitative comparison as it has a different pixel type and channel count.

With this PR, opening the largest resolution from the .vsi file in ImageJ or similar should show that the image is now shifted 34 pixels in X and 1 pixel in Y, matching what OlyVIA shows. This is easiest to see when zooming in on the corners of the image.
The pixel shift is scaled with the resolution index, so opening smaller resolutions should also show a slight difference with this PR. 

This requires a minor release, as additional fields are added to the `Pyramid` inner class (which I expect will break memo files). Assigning to 6.12.0 for now, but can but can push to the next non-patch release as needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -530,11 +530,22 @@ public class CellSensReader extends FormatReader {
         getRGBChannelCount() * FormatTools.getBytesPerPixel(getPixelType());
       int outputRowLen = w * pixel;
 
+      Pyramid pyramid = getCurrentPyramid();
+
       for (int row=0; row<tileRows; row++) {
         for (int col=0; col<tileCols; col++) {
           int width = tileX.get(getCoreIndex());
           int height = tileY.get(getCoreIndex());
           Region tile = new Region(col * width, row * height, width, height);
+
+          // the pixel data in the stored tiles may be larger than the defined image size
+          // the "tile origin" information indicates how to crop the pixel data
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested using the sample file provided on the original image.sc thread. Without this PR, the reading of the data clearly indicates a shift towards the right of the image which is most visible on highest resolution levels and towards the bottom of the image.

The proposed change reads the `TILE_ORIGIN` metadata when present and uses it to correct the origin of the tiles in `openBytes`. With it included, the reading of the pixel data for the sample file matches the ground truth from the manufacturer's OlyVIA software.

This should be ready for inclusion in the upcoming minor release.

PR #3928, opened 2023-01-04
CZI: tile stitching fixes when pyramid does not exist

### LINKED ISSUE #3710: CZI pyramid not detected / tile width not correct
See  https://forum.image.sc/t/czi-slide-scanner-image-bio-formats/52892/ and data in ```inbox/romainGuiet```.  I can confirm that the two files reported as not working open with skewed images and multiple series which are individual tiles from a pyramid.
Currently open .czi PRs (#3681, #3687) do not fix this.  In Zen, the files open as Z stacks with multiple channels (no scenes).

I suspect #3323 is related, but not identical.  Fixing the pyramid detection first may fix the image skew as well.

melissalinkert:
Fixes #3710.

See test files in `curated/zeiss-czi/gh-3710`.

Without this PR, `showinf` should report several small series corresponding to individual tiles in a larger image. In some cases, viewing the tiles will show a skewed image.

 With this PR, the tiles should be correctly stitched together into a single series which no longer looks skewed. Screenshots from ZEN for comparison are on the corresponding configuration PR.

I would expect one existing file to be affected by this, as noted in the configuration PR. I believe the previous handling of that file was incorrect, and this PR fixes it. Otherwise, I would not expect this to introduce test failures.

This should be safe for a patch release, so assigning to 6.12.1 for now since it isn't especially urgent and there is already a lot in 6.12.0.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -877,10 +877,12 @@ public class ZeissCZIReader extends FormatReader {
         int newY = lastPlane.y;
         if (allowAutostitching() && (ms0.sizeX < newX || ms0.sizeY < newY)) {
           prestitched = true;
-          mosaics = 1;
+          if (maxResolution > 0) {
+            mosaics = 1;
+          }
         }
         else {
-          prestitched = maxResolution > 0;
+          prestitched = true;
         }
 
         // don't shrink the dimensions if prestitching is allowed


dgault (approved):
Builds and tests have continued to be green with this PR and the config PR included.

Tested using the sample file in `curated/zeiss-czi/gh-3710`

Without the PR and the autostitch option disabled, the file reads as having 8 series inlcuding a label and macro image. In Scene 1 each series is 1388x799 (15 planes 5Cx3Z), for Scene 2 it is 738x255(9 planes 3Cx3Z). When displaying the individual series, half of the images show the 'skewed' effect that we have seen reported in CZI files.

Without the PR and the autostitch option enabled the file reads as before, having 8 series inlcuding a label and macro image.

With the PR included and the autostitch option disabled the file reads as before, having 8 series inlcuding a label and macro image.

With the PR included and the autostitch option enabled, the image now shows correctly 3 series including a label and macro image. Scene 1 shows as 2575x3378 15 planes 5Cx3Z), for Scene 2 it is 738x255(9 planes 3Cx3Z). The images display correctly without any skewing effect.

For the remaining Partikelnormal file altered in the config PR, opening the image with this PR and the autostitch option displays the stitched image matching that from the Zen screenshot

PR #3932, opened 2023-01-17
ND2: fix reading of blocks larger than 2 GB

melissalinkert:
See https://github.com/glencoesoftware/bioformats2raw/issues/184

To test, use the provided test file, which is copied to `curated/nd2/bioformats2raw-184`. Without this PR, `showinf -nopix` should throw `NegativeArraySizeException` eventually. Adding `-option nativend2.chunkmap false` results in the same exception.

With this PR, `showinf -nopix` should initialize without exception, though it may be a little slow. Reading tiles from the image should result in reasonable looking images. I was not able to open the full image in Nikon's NIS Elements viewer for comparison, but can view a thumbnail and the basic metadata (a screenshot is included in the configuration PR).

The source of the problem was the `CustomDataSeq|edfbox_d` block, which has a data length greater than 2GB. I would expect files which do not have a data block larger than 2GB (including files smaller than 2GB) to be unaffected.

Assuming tests pass, this should be safe for a patch release. Assigning to 6.12.1 for initial review.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -611,7 +611,7 @@ public class NativeND2Reader extends SubResolutionFormatReader {
 
         in.seek(helper + 12);  // Return to starting position
 
-        int len = (int) (nameLength + dataLength);
+        long len = nameLength + dataLength;
 
         long fp = in.getFilePointer();
         String blockType = in.readString(12);
@@ -620,7 +620,7 @@ public class NativeND2Reader extends SubResolutionFormatReader {
         LOGGER.info("Parsing block '{}' {}%", blockType, percent);
         blockCount++;
 
-        int skip = len - 12 - nameLength * 2;
+        long skip = len - 12 - nameLength * 2;
         if (skip <= 0) skip += nameLength * 2;
 
         // Image calibration for newer nd2 files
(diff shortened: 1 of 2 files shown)


dgault:
A few failures in the overnight repo tests: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

melissalinkert:
Hopefully 3e1587b6f4 fixes this, otherwise feel free to exclude for now.

The two test failures are only on the new file, and both indicate that the test is attempting to read whole planes, each of which is 1394238204 bytes. `TestTools.canFitInMemory` incorrectly returns true, allowing the whole plane to be read when it's too big to reasonably do so.

Iterating over every plane using 256x256 tiles does not throw an exception (https://gist.github.com/melissalinkert/8732d57e667fc5664d2232e262f7d0b6).

dgault (approved):
Builds and tests, including the updated FormatReaderTest have been green with this PR included.

Tested using the recommended sample file - `curated/nd2/bioformats2raw-184`. Without the PR included, the file fails to open and an exception in seen as below:

```
Exception in thread "main" java.lang.NegativeArraySizeException
	at loci.common.RandomAccessInputStream.readString(RandomAccessInputStream.java:632)
	at loci.formats.in.NativeND2Reader.initFile(NativeND2Reader.java:588)
	at loci.formats.DelegateReader.setId(DelegateReader.java:291
```

With the PR included the file is able to be read without exception and the metadata matches that from the screenshots in the associated config PR.

PR #3933, opened 2023-01-18
upgrade setup-java action. Current version uses set-output

jburel:
see also https://github.com/ome/action-gradle/pull/1

Check that the build via GHA is green

PR #3912, opened 2022-11-21
Upload one set of artifacts from the GitHub Actions ant build

melissalinkert:
As discussed with @ome/formats earlier today.

For the Java 1.8/ubuntu-latest ant build only, artifacts should be now be archived, for 30 days only (instead of default 90). Hopefully that's enough to prevent any problems with usage limits.

I did not do anything extra to add commit hashes as suggested in discussions. All of the Bio-Formats builds should already include the commit hash in the `Implementation-Build` of the manifest and the `FormatTools.VCS_REVISION` variable. Using the `bioformats_package.jar` archived here with ImageJ (`Help > About Plugins > Bio-Formats Plugins...`), `showinf -version`, or `java -jar bioformats_package.jar` and then `Help > About...` should all show the 938eca1... commit hash.

PR #3815, opened 2022-04-21
Java 17 support

sbesson:
Fixes #3659 

Java 17 LTS was released in September 2021 and in getting increasing used in the community by consumers of Bio-Formats e.g. `bioformats2raw` or `QuPath`. While there has been few reports so far indicating, this PR aims to bring full Java 17 support to Bio-Formats in the upcoming minor release, adding the required testing infrastructure and making the minimal set of code changes.

Summary of changes:

- [7ef4d15](https://github.com/ome/bioformats/pull/3815/commits/7ef4d1522ba719c4e24f851dc84cab32b34d1c52) and [f6d9c4b](https://github.com/ome/bioformats/pull/3815/commits/f6d9c4bcda975178bcece9fe19c7f73ba3c0dea7) add Java 17 to the GitHub Actions testing matrix 
- [6bf30d7](https://github.com/ome/bioformats/pull/3815/commits/6bf30d7be1e0de8edfe4a8a620f1a9bd6daace80) and [bf047fb](https://github.com/ome/bioformats/pull/3815/commits/bf047fbb3204db221a089fd62e148b9d8f163ff1) make the minimal changes to the `test-suite` infrastructure to have the automated tests running against JDK 17
- [d27a825](https://github.com/ome/bioformats/pull/3815/commits/d27a82539749c0edc006ecb4876778057e8c4a78), [767733f](https://github.com/ome/bioformats/pull/3815/commits/767733f35c8c8769a9466b732db6ede1681870d6), [43b572f](https://github.com/ome/bioformats/pull/3815/commits/43b572f22fda1e923fbb9fff3220d7c98f53a49a), [a315e5b](https://github.com/ome/bioformats/pull/3815/commits/a315e5bc07bb8e1e98058b8fc122c623606a912b) and [7769fa8](https://github.com/ome/bioformats/pull/3815/commits/7769fa8dee3bad5b81f10102ac987c32ea27a5e6) contain the minimal reader changes required for the memoization on Java 17
-  [07b8507](https://github.com/ome/bioformats/pull/3815/commits/07b85072d482d459fc9ad28eae20e44e2d0e8bc4) upgrades `ome-codecs` to `0.4.2` which includes the required upgrade of `io.airlit.aircompressor` to version 0.21. As demonstrated in the discussion of this PR, this change required https://github.com/ome/bioformats/pull/3923 to fix the Ant build

Related work:

- https://github.com/ome/bioformats/pull/3796 also fixes an issue exposed when using Java 17
[...]

PR #3934, opened 2023-01-21
Release artifacts

jburel:
This PR creates and uploads the release artifacts when a tag is pushed
4b1f6859f76cae9feb4865b66c19cb487699b903 is needed since the reference ``compile/runtime.classpath`` is no longer found when ant 1.10 is used (that's the version installed via GHA). Open to alternative is someone knows one

Tested via https://github.com/jburel/bioformats/releases/tag/v6.11.1-m1

### DIFF
--- a/components/bundles/bioformats_package/build.xml
+++ b/components/bundles/bioformats_package/build.xml
@@ -6,15 +6,26 @@ Download Apache Ant from http://ant.apache.org/.
 Type "ant -p" for a list of targets.
 -->
 
-<project name="bioformats_package" default="bundle" basedir=".">
+<project name="bioformats_package" default="bundle" basedir="."  xmlns:resolver="antlib:org.apache.maven.resolver.ant">
   <description>Build file for Bio-Formats bundle</description>
   <property name="root.dir" location="../../.."/>
-  <import file="${root.dir}/ant/java.xml"/>
+  <import file="${root.dir}/ant/java.xml" />
   <property file="build.properties"/>
+
 
   <target name="docs" depends="javadoc-properties"
     description="generate the Javadocs for most components">
     <echo>----------=========== Javadocs ===========----------</echo>
+    <path id="compile.classpath"/>
+    <path id="runtime.classpath"/>
+
+    <resolver:resolve>
+      <dependencies>
+        <pom refid="pom"/>
+      </dependencies>
+      <path refid="compile.classpath" classpath="compile"/>
+      <path refid="runtime.classpath" classpath="runtime"/>
+    </resolver:resolve>
     <tstamp>
       <format property="YEAR" pattern="yyyy"/>
     </tstamp>
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -17,4 +17,6 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
       - name: Build with Ant
-        run: ./tools/test-build ant
+        run: |
+            ./tools/test-build ant
+            ant docs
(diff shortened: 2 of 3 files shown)


sbesson (commented):
A few additional questions in terms of the proposed changes:

1- I am going back and forth about injecting the `resolver:resolve` step in `components/bundle/bioformats_package.xml` as proposed here vs centralizing all the targets involving some dependency resolution to `ant/java.xml` 
2a- should the `docs` target be built as part of every run to catch future regressions?
2b- https://github.com/ome/bioformats/pull/3912 also modifies the logic of the Ant action to upload a set of artifacts for consumption. This addition could also be constructed a tag-specific step of the `ant` workflow assuming the relevant artifacts are all built

No blockers but possibly a few decisions that might be worth discussing as part of the @ome/formats weekly meeting.

jburel:
> 1- I am going back and forth about injecting the `resolver:resolve` step in `components/bundle/bioformats_package.xml` as proposed here vs centralizing all the targets involving some dependency resolution to `ant/java.xml` 
I have tried various things to no success i.e. keeping it in `ant/java.xml` 


 >2a- should the `docs` target be built as part of every run to catch future regressions? 
This is an important part of the release process and it does not take long so I think it should be included by default.


> 2b- #3912 also modifies the logic of the Ant action to upload a set of artifacts for consumption. This addition could also be constructed a tag-specific step of the `ant` workflow assuming the relevant artifacts are all built

It could be. I follow the same pattern that we introduced in openmicroscopy mainly to  keep things clearer.
Convoluted workflows can be difficult sometimes to understand and adjust.
> 
> No blockers but possibly a few decisions that might be worth discussing as part of the @ome/formats weekly meeting.

components/bundles/bioformats_package/build.xml
```diff
@@ -6,15 +6,26 @@ Download Apache Ant from http://ant.apache.org/.
...
   <property name="root.dir" location="../../.."/>
-  <import file="${root.dir}/ant/java.xml"/>
+  <import file="${root.dir}/ant/java.xml" />
   <property file="build.properties"/>
 
+
   <target name="docs" depends="javadoc-properties"
     description="generate the Javadocs for most components">
     <echo>----------=========== Javadocs ===========----------</echo>
+    <path id="compile.classpath"/>
```
sbesson, line 19: As mentioned in a previous version of this commit, this is equivalent to the approach I took in https://github.com/ome/bioformats/pull/3923/commits/bf8076ec08fffc2a5a2deedd6e6240a33ecc11a8#diff-fdeb882f159787e553f8d2b7f340d498880a8a54e60ae5dee163671dc9009018R203-R211 to fix the bundle generation target.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#359](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/359/). See the [console output](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/359/consoleText) for more details.
Possible conflicts:
  - PR #3815 sbesson 'Java 17 support'
    - .github/workflows/ant.yml
  - PR #3912 melissalinkert 'Upload one set of artifacts from the GitHub Actions ant build'
    - .github/workflows/ant.yml
  - PR #3933 jburel 'upgrade setup-java action. Current version uses set-output'
    - .github/workflows/ant.yml

--conflicts

dgault (approved):
Overall the changes look good, agreed that `ant docs` should indeed be part of the process. The artifacts generated by the test in https://github.com/jburel/bioformats/releases/tag/v6.11.1-m1 all look to be correct.

PR #3939, opened 2023-02-02
ImagePlusReader: handle IndexColorModel

### LINKED ISSUE #3730: Custom reader supplied LUT gets lost when using ImageJ
I am trying to evaluate the possibility of having the OBF reader supply custom LUTs via e.g. `get8BitLookupTable()`.
By default ("Color mode: Default") loading an OBF in ImageJ results in grayscale display.
Returning any LUT via `get8BitLookupTable()` seems to result (irregardless of actual content) in a gradient LUT display from black to yellow.

I found the following snippet:
https://github.com/ome/bioformats/blob/b67dfcac19b898e9a8e2481e5288bdc4b7aa782d/components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java#L193-L209

I don't understand it or know the history behind it or possible side effects but commenting it out seems to result in my custom LUT being used properly.

dgault:
This is a follow up to forum thread https://forum.image.sc/t/bio-formats-color-wrong-for-imagej-images/76021/7 and existing issue https://github.com/ome/bioformats/issues/3730

The GitHub Issue has a comment with further details on the need to handle IndexColorModel alonside LUT: https://github.com/ome/bioformats/issues/3730#issuecomment-1387258241

This was a quick initial test fix and will require further testing and work before it is ready for review.

Fixes #3730

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java
@@ -34,6 +34,7 @@ import ij.process.ImageProcessor;
 import ij.process.LUT;
 
 import java.awt.image.ColorModel;
+import java.awt.image.IndexColorModel;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


dgault:
Change produces a number of failures in the Importer unit tests, will require some investigation to confirm if the new values are indeed correct.

dgault:
Closing PR as I wont have time to investigate the failing tests in time for upcoming release.

PR #3942, opened 2023-02-08
Bump version to 6.12.0-m1

dgault:
Bump pom versions to 6.12.0-m1

### DIFF
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>ome</groupId>
     <artifactId>pom-bio-formats</artifactId>
-    <version>6.12.0-SNAPSHOT</version>
+    <version>6.12.0-m1</version>
     <relativePath>../..</relativePath>
   </parent>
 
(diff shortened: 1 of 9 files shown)


jburel:
I assume you will merge when green and push a tag so we can test the release workflow
You will also need to push a tag in the doc repo

jburel:
```package ch.qos.logback.core does not exist``` for javadoc

**Comment by dgault** [comment]
As for possible resolutions, the 2 pathways I see at the moment are:
- add logback as an optional dependency to bioformats_package pom
- try to move the docs generation to the toplevel ant task

At the minute I am leaning towards the first option but open to alternative ideas @sbesson, @jburel, @melissalinkert

**Comment by jburel** [comment]
Option 1 will be the less disruptive one.

**Comment by sbesson** [comment]
> add logback as an optional dependency to bioformats_package pom

No objection from my side. Also the distributed command-line tools and MATLAB package use this uber-JAR and we certainly want them to have some logging implementation shipped by default.
Does this mean that logging for these utilities is broken with the current state of the development branch?

**Comment by dgault** [comment]
Yeah, the artifacts build without error but when trying to run the tools I get:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

PR #3947, opened 2023-02-21
Fix NPEs when a no-op file pattern is encountered

melissalinkert:
Backported from a private PR.

If the helper reader performs file grouping, then `externals` will be null and `noStitch` will be set to `true`. This cleans up a few overlooked places where `externals` was assumed to be non-null. It also logs an `INFO` message if the file pattern is discarded in favor of grouping by the helper reader. We can debate whether this should just be logged, or if throwing an exception is more appropriate.

To test, use `curated/ome-tiff/samples/no-op-file-pattern/multifile.pattern`. Without this PR, run `showinf -nopix multifile.pattern -cache` twice. The first time will succeed, the second time will fail with a `NullPointerException`. With this PR, remove the memo file and perform the same test. Both `showinf` calls should succeed without an exception, and there should be a `File pattern ignored...` message in the output.

This can also be tested without memo files by using the following code:

```
try (ImageReader reader = new ImageReader()) {
  reader.setId("multifile.pattern");
  reader.close(true);
  reader.reopenFile();
}
```

Without this PR, the above should throw an exception; with this PR it should not.

### DIFF
--- a/components/formats-bsd/src/loci/formats/FileStitcher.java
+++ b/components/formats-bsd/src/loci/formats/FileStitcher.java
@@ -205,7 +205,7 @@ public class FileStitcher extends ReaderWrapper {
    */
   public int[] getAxisTypes() {
     FormatTools.assertId(getCurrentFile(), true, 2);
-    return externals[getExternalSeries()].getAxisGuesser().getAxisTypes();
+    return getAxisGuesser().getAxisTypes();
   }
 
   /**
@@ -220,7 +220,7 @@ public class FileStitcher extends ReaderWrapper {
    */
   public void setAxisTypes(int[] axes) throws FormatException {
     FormatTools.assertId(getCurrentFile(), true, 2);
-    externals[getExternalSeries()].getAxisGuesser().setAxisTypes(axes);
+    getAxisGuesser().setAxisTypes(axes);
     computeAxisLengths();
   }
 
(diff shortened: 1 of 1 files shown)


dgault:
Some failed tests today for a pattern file https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

It passed yesterday with the PR included so perhaps related to the Java version, todays was JDK11.

sbesson:
From the logs, it looks like all failures are associated to the new sample dataset added to the repository tests for testing this feature rather than regressions on existing datasets.

sbesson (approved):
Using the generated sample, confirmed that:

- without this PR, `showinf -nopix -cache` succeeds but fails on a second invocation with a NPE in reopenFile
- with this PR, the same commands are both successful.

CI repository tests have been consistently green with this PR included as well as the new sample dataset.

Additionally, with #3950 included, the sample OME-TIFF fileset + pattern was successfully imported into an OMERO.server where the binary repository is a symlink. Without #3950, the import failed with the `SecurityViolation` described in https://forum.image.sc/t/no-access-to-some-images-in-omero-securityviolation-metadata-companion-ome-accesses-data-outside-managed-repository/69453. With both PRs included, the import was successful and the data can be accessed by the clients demonstrating that the reopenFile logic is working.

PR #3948, opened 2023-02-22
vsi: slightly rework tile decoding to fail faster on invalid tiles

melissalinkert:
Backported from a private PR.

This reads the expected number of compressed bytes into an array, then attempts to decompress the array. This will fail faster on invalid tiles than passing the entire .ets file to a codec, which may attempt to read the entire file in search of the start or end of the compressed block.

This is meant to address a very specific case where a memo file is generated and then the underlying files are changed. In certain circumstances, particularly if the new files are smaller than the old files, this can cause `openBytes` to not return (or at least, be very very slow). This case is not expected to really work, but should at least fail faster.

This is a little tricky to test, but something like the following (without this PR) should show the issue:

```
# make a local copy of the data in curated/cellsens/brian
$ showinf -noflat -cache BZ\ cfos1-2_05.vsi -resolution 5
# image should be displayed without error
$ cp .BZ\ cfos1-2_05.vsi.bfmemo .BZ\ cfos1-2_05.vsi.bfmemo.bak
$ cp SW10082402-02_21.vsi BZ\ cfos1-2_05.vsi
$ cp _SW10082402-02_21_/stack1/frame_t_0.ets _BZ\ cfos1-2_05_/stack1/frame_t_0.ets  
$ showinf -noflat -cache BZ\ cfos1-2_05.vsi -resolution 5
# this may work, in which case follow the next two steps
$ cp .BZ\ cfos1-2_05.vsi.bfmemo.bak .BZ\ cfos1-2_05.vsi.bfmemo
$ showinf -noflat -cache BZ\ cfos1-2_05.vsi -resolution 5
# image not displayed after several minutes; process must be killed as it is blocked on call to openBytes
```

With this PR and the same setup/memo file, all calls to `showinf -noflat -cache BZ\ cfos1-2_05.vsi -resolution 5` should throw an exception quickly.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -1121,29 +1121,33 @@ public class CellSensReader extends FormatReader {
       if (tileSize == 0) {
         tileSize = tileX.get(getCoreIndex()) * tileY.get(getCoreIndex()) * 10;
       }
-      options.maxBytes = (int) (offset + tileSize);
 
       long end = index < tileOffsets.get(getCoreIndex()).length - 1 ?
         tileOffsets.get(getCoreIndex())[index + 1] : ets.length();
 
+      int compression = compressionType.get(getCoreIndex());
+      int compressedBufSize = compression == PNG || compression == BMP ?
+        (int) (end - offset) : tileSize;
+      byte[] compressedBuf = new byte[compressedBufSize];
+      ets.read(compressedBuf);
+
       String file = null;
 
-      switch (compressionType.get(getCoreIndex())) {
+      switch (compression) {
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested using the workflow from the description. Confirmed that with Bio-Formats 6.12.0, the sequence leads to a hanging utility while it results in a `NullPointerException` in the second case.

Arguably, the issue is really an edge case and the combination of data corruption used with an outdated memo file should definitely be avoided as much as possible. Overall, I find the new code more readable, separating the buffer initialization from the decompression logic and as shown by the tests above, it handles some tricky scenarios more defensively.

The repository tests have been consistently passing with this PR included providing additional confidence the refactoring does not cause any regression for any JDK version.

PR #3952, opened 2023-02-28
Metamorph: don't try to read metadata keys that don't exist

### LINKED ISSUE #3578: MetaMorph: NegativeArraySizeException
This issue was raised on forum thread https://forum.image.sc/t/negativearraysizeexception-when-opening-metamorph-nd-or-stk-files/39526

The issue seems to be due to an incorrect value read at https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/MetamorphReader.java#L2062

It was reported using Bio-Formats 6.5.0

melissalinkert:
Prevents a NegativeArraySizeException (or IOException) if the key length is too small or too large. This is expected when the key/value pair is not actually defined in the file. Fixes #3578.

To test, use the datasets in `curated/metamorph/imagesc-39526`. Without this PR, `showinf -nopix -omexml` on the .nd files should throw a `NegativeArraySizeException` as noted in #3578. With this PR, both files should initialize without error.

Should be safe for any patch release; adding to 6.12.1 for initial review.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -2114,6 +2114,9 @@ public class MetamorphReader extends BaseTiffReader {
   private void readPlaneData() throws IOException {
     in.skipBytes(4);
     int keyLength = in.read();
+    if (keyLength < 0 || keyLength >= in.length() - in.getFilePointer()) {
+      return;
+    }
     String key = in.readString(keyLength);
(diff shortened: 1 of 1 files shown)


dgault (approved):
Tested using the sample file set in `curated/metamorph/imagesc-39526`

Without this PR the file fails to open and a `NegativeArraySizeException` is thrown. With the PR included the file can be opened and displayed as expected.

With the associated config PR all of the builds and repo tests have remained green with the PR included.

Issue #3951, opened 2023-02-28
OperettaReader does not set the pixel type for single plane images


aherbert:
We cannot import Operetta images into OMERO if they have only one plane per image series. The import breaks on the computation of the MinMax for the plane as it cannot open the pixel bytes. The pixel type defaults to 8-bit as set in the CoreMetadata but the TIFF is actually 16-bit.

I know where the bug is. I cannot reproduce it using the `showinf` tool as this correctly opens the requested 16-bit TIFF image.

Here is the issue in the OperettaReader:

```Java
        boolean validFile = false;
        while ((filename == null || !new Location(filename).exists() || !validFile) &&
          planeIndex < planes[i].length - 1)
        {
          LOGGER.debug("Missing TIFF file: {}", filename);
          planeIndex++;
          filename = planes[i][planeIndex].filename;

```
It sets the file as invalid. It does not bother to read the current (and first and only) filename and get the pixel type. It immediately notices the `validFile` is false and attempts to move to the next plane if available. Since my data is only single plane images there is no next file and the reader logs a warning in the Blitz.log that:
```
"Could not find valid TIFF file for series 0; pixel type may be wrong"
```
I have fixed this locally by changing the loop to actually open the first filename.

The error can be triggered with the following snippet but the Operetta file must be a single plane per image (no multiple channels, multiple z, etc):
```Java
    try (ImageReader reader = new ImageReader()) {
      final IMetadata omeMeta = MetadataTools.createOMEXMLMetadata();
      reader.setMetadataStore(omeMeta);
      reader.setId("/path/to/operetta_data/Images/Index.idx.xml");
      int c = omeMeta.getImageCount();
      for (int image = 0; image < c; image++) {
        reader.setSeries(image);
        reader.openBytes(0);
      }
    }
```

I am happy to provide a minimal Operetta image which triggers this condition. I can open a PR but do not know where to put a unit test with an example image.

aherbert:
PR submitted.

I filed the OME CLA for a fix to OMERO blitz on 12-Jul-2021. Let me know if you require another one.

I look into creating a minimal image that triggers this issue and post a link to the data.

PR #3953, opened 2023-03-01
OperettaReader: Do not skip the first filename when validating series

aherbert:
See GH Issue #3951

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -498,13 +498,7 @@ public class OperettaReader extends FormatReader {
         ms.sizeY = planes[i][planeIndex].y;
         String filename = planes[i][planeIndex].filename;
         boolean validFile = false;
-        while ((filename == null || !new Location(filename).exists() || !validFile) &&
-          planeIndex < planes[i].length - 1)
-        {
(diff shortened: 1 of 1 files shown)


dgault:
Thanks again @aherbert for contributing the PR, overall the changes make sense and look to be correct. My suspicion  is that the original `planeIndex < planes[i].length - 1` was a mistake that wasn't caught as we don't seem to have any single plane sample files. All builds and repository tests have remained green with the PR included and haven't shown any signs of regressions in existing sample files. Overall the PR looks good and Im happy to get this merged. 

If you do have a link to a minimal sample that we could add to our nightly tests that would great.

aherbert:
Hi @dgault 

Apologies for the delay in obtaining an example image. The following is a dataset using a single plane image:

[DOI: 10.5281/zenodo.7841359](https://doi.org/10.5281/zenodo.7841359)

Note that the image can be viewed using the BioFormats `showinf` command line tool. But it previously could not be opened using the code snippet provided in the linked GH Issue due to a pixel type mismatch. When I pulled the latest `develop` branch with the fix then the same test code is successful.

You can use the image as required in your test suite (there are no license restrictions). If you require some other minimal datasets (e.g. multiple images per well, all single plane) then please let me know.

dgault:
Thanks, thats a great help! That should help us catch any future regressions of this bug.

PR #3956, opened 2023-03-03
Bump sl4j-api to 2.x

sbesson:
See discussions at https://github.com/ome/bioformats/pull/3943#issuecomment-1422854526 and https://github.com/ome/bio-formats-examples/pull/102#issuecomment-1422904376

Bio-Formats 6.12.0 included the declaration of logback-core and logback-classic as optional dependencies in agreement with the recommendations of the slf4j framework. As. part of the testing and while trying to upgrade logback to modern versions across the board, it was found that the current version of slf4-api shipped with ome-common & Bio-Formats imposes a constraint on logback. This can be highly dependent of the build system and the way dependencies are resolved but the logback 1.2.9 released Dec 17, 2021 is the last version that uses slf4-api 1.x.

This commit explores bumping the slf4-api version to 2.x. For library usages, this means logback should be upgraded to a compatible version, either  1.3.x (Java 8+) or 1.4.x (Java 11+)

Tentatively assigning to 6.13.0


melissalinkert (approved):

dgault (approved):
Happy to get all of these included as part of 7.1.0 as long as it is OK for OMERO downstream (https://github.com/ome/omero-blitz/pull/137)

PR #3958, opened 2023-03-06
Memoizer: minor API updates

### LINKED ISSUE #3820: Memoizer API updates
See https://github.com/glencoesoftware/bioformats2raw/pull/145#issuecomment-1117116897

melissalinkert:
Fixes #3820.

This allows a null reader to be specified in the constructors, in which case a new `ImageReader` will be wrapped. The memo file can also now be deleted and retrieved from public methods.

Unit tests should cover the new functionality, but can certainly add more tests if I've missed anything.

Requires a minor release due to API additions.


joshmoore (approved):
Other than possibly listing the (positive) `null` argument behavior in the javadocs, :+1: since the ctors here were never my favorite (and I can't remember any code that would depend on an NPE)

sbesson (approved):

dgault (approved):

PR #3959, opened 2023-03-07
Add writer options to preserve the OME Creator attribute

melissalinkert:
By default, the OME-XML and OME-TIFF writers will still set the Creator to FormatTools.CREATOR, even if that means overwriting an existing value.

This is option 1 in #3541.

Not super high priority, but opening for 6.13.0 discussion. If we decide that option 2 in #3541 is better, I can update this PR. There is some duplicate code between the two writers now, but not a great common place to put it (though happy to hear other ideas).

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/OMETiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/OMETiffWriter.java
@@ -77,6 +77,7 @@ public class OMETiffWriter extends TiffWriter {
     FormatTools.URL_OME_TIFF + ". -->";
 
   public static final String COMPANION_KEY = "ometiff.companion";
+  public static final String CREATOR_KEY = "ometiff.preserve_creator";
 
   // -- Fields --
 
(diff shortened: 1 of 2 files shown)


dgault (approved):
Going with the configurable approach works for me and gives the end user the option to decide the behaviour.

Tested using a modified OME-TIFF sample file with the Creator tag modified. 
- Converting the file without the PR resulted in the tag being set to `OME Bio-Formats 6.12.0`
- Converting the file with the PR  and the default behaviour resulted in the tag being set to `OME Bio-Formats 6.12.1-SNAPSHOT ` and a warning that the tag was being overwritten was displayed
- Converting the file with the PR and the new option set to true resulted in the original tag being maintained and no warning was shown
- Converting the file with the PR and the new option set to false resulted in the tag being set to `OME Bio-Formats 6.12.1-SNAPSHOT ` and a warning that the tag was being overwritten was displayed

So the new behaviour appears to be working as expected

@melissalinkert, can we also get a docs PR to add the new writer options to https://github.com/ome/bio-formats-documentation/blob/master/sphinx/formats/options.rst

melissalinkert:
> can we also get a docs PR to add the new writer options to https://github.com/ome/bio-formats-documentation/blob/master/sphinx/formats/options.rst

https://github.com/ome/bio-formats-documentation/pull/299

PR #3960, opened 2023-03-08
push to releases

jburel:
This PR proposes to change the target on artifactory
This matches the changes introduced in the omero-* repositories e.g. https://github.com/ome/omero-common/blob/master/.github/workflows/gradle.yml#L27
Permissions have been granted to the GitHub user pushing to artifactory
This will remove the manual step


sbesson (approved):
This proposal aligns with what has been used for the `org.openmicroscopy:omero-*` components

dgault (approved):

PR #3962, opened 2023-03-10
SVS: Remove broken special case path when there is only one (pyramidal) image.

alex-virodov:
Hello. We deidentify some .svs slides using a fork of https://github.com/pearcetm/svs-deidentifier/blob/master/wsideidentifier.py. That code removes the macro and label images from an .svs file. After that removal, the slide does not show correctly in QuPath anymore. 

I traced the problem to this line:
https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/SVSReader.java#L172
```
  public byte[] openBytes(...)
  {
    if (core.size() == 1) {
      return super.openBytes(no, buf, x, y, w, h);
    }
```

It seems that there is a special case for a slide that contains only one (pyramidal) image. However, that special case does not use the resolution field that is set on the object earlier, and ends up reading the wrong pyramid level - always the 0th one. 

The code below this special case seems to be the general case, and seems to work properly. In my testing, I removed the special case `if (core.size() == 1)` and the deidentified slide works properly with QuPath and seems to pass the unit tests.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SVSReader.java
+++ b/components/formats-gpl/src/loci/formats/in/SVSReader.java
@@ -169,9 +169,6 @@ public class SVSReader extends BaseTiffReader {
   public byte[] openBytes(int no, byte[] buf, int x, int y, int w, int h)
     throws FormatException, IOException
   {
-    if (core.size() == 1) {
-      return super.openBytes(no, buf, x, y, w, h);
-    }
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
     if (tiffParser == null) {
       initTiffParser();


dgault:
Thank you @alex-virodov for contributing to the project. I have marked your PR for inclusion in our nightly builds and tests, this should flag any known issues that resulted in the special case being added in the first place.

We have introduced a Contributor License Agreement for the project, would you be able to sign and return the form following the instructions on https://ome-contributing.readthedocs.io/en/latest/cla.html

alex-virodov:
@dgault Hello. I've sent the signed Contributor License Agreement, is there anything else I can do to proceed with this PR?
Please let me know if you did not receive it.

Thank you.
Alex.

dgault:
Thanks @alex-virodov, thats great, i will try to get this PR reviewed in the next few days so it can be included in the next Bio-Formats release.

dgault (approved):
Going through the commit history for the reader and this section of the code was introduced many years ago when the reader was first created. As such there does not appear to be any specific reasoning given for why the special case was introduced.

There have been a number of reported issues with SVS files which do not contain macro or label images, in 6.8.0 there was an attempt to solve these with https://github.com/ome/bioformats/pull/3733. This particular use case did not seem to be caught.

We have one current open issue https://github.com/ome/bioformats/issues/3877 which looks to be related. The sample file does not contain a label or macro image and with this PR included each of the resolution levels is able to be opened and displayed as expected.

The PR has no impact on any existing datasets with our data repo and looks to be safe to be merged. I will open a follow up PR to configure and add the sample file from https://github.com/ome/bioformats/issues/3877 to the data repo.

PR #3963, opened 2023-03-13
Add API to configure the value of undefined pixels

### LINKED ISSUE #3800: Fill color configuration
See #3795 and https://github.com/glencoesoftware/bioformats2raw/pull/65 for prior work on this topic.

Some formats allow sparse pixel data, which means that a particular image tile may contain pixels that have no value stored in the file(s). This is especially true for whole slide formats. Most readers fill these undefined pixels with 0s, which doesn't always match the vendor software or users' expectations. Brightfield slides in particular are usually expected to have undefined pixels set to 255 (i.e. white).

To better address this across all readers, one approach would be:

* Define a new `IFormatReader` method `Color getFillColor()` which returns the `ome.xml.model.primitives.Color` that will be used for undefined pixels (the return type is up for debate). By default this returns black/0.
* Define a new `IFormatReader` method `setFillColor(...)` which allows the color to be changed. This alters the value returned by `getFillColor()`.
* Add an option to bfconvert/bioformats2raw/etc. that calls `setFillColor`.
* Update to always use `getFillColor()` as the default pixel value. This might be worth adding to `FormatTools.checkPlaneParameters` (or some other method that all `openBytes` implementations call), and would need to be changed in the ~20 places where `Arrays.fill(buf, (byte) 0)` is called by `openBytes`.
* Update individual readers to override `getFillColor()` if color metadata is defined or can be inferred for a particular format. This could be limited to the readers for whole slide formats, at least to start.

Specific API changes are up for discussion, as is the priority since this would require at least a minor version update.

melissalinkert:
Fixes #3800.

This generalizes the work in https://github.com/glencoesoftware/bioformats2raw/pull/65 to allow a fill value to be set on all readers. Unless otherwise specified, the fill value will be 0, so the default behavior should be unchanged.

`ZeissCZIReader` is currently the only reader that overrides the default fill value. This is an extension of #3795.

`showinf` and `bfconvert` also have a new `-fill byte` option that uses the new API.

Any of the files in https://github.com/openmicroscopy/data_repo_config/pull/518 could be used to test the new API, but something like this should clearly demonstrate this feature:

```
$ showinf -crop 0,0,512,512 zeiss-czi/qa-10972/01042015-36.czi
# white tile is displayed (default for brightfield)
$ showinf -crop 0,0,512,512 zeiss-czi/qa-10972/01042015-36.czi -fill 0
# black tile is displayed (matching `-fill` value)
$ bfconvert -series 0 -crop 0,0,512,512 zeiss-czi/qa-10972/01042015-36.czi default.png
# resulting PNG is all white
$ bfconvert -series 0 -crop 0,0,512,512 zeiss-czi/qa-10972/01042015-36.czi fill-0.png -fill 0
# resulting PNG is all black
```

The original proposal in #3800 was to allow a `Color` to be specified. A `Byte` is used instead here as in https://github.com/glencoesoftware/bioformats2raw/pull/65. This is simpler to implement, as the specified byte value is just stored directly in pixel byte arrays without checking pixel type, channel count/RGB, etc. If anyone feels strongly that a `Color` needs to be specified, we can discuss, but that opens up other questions around how to handle anything that isn't 3 RGB channels.

Assigning to 6.13.0 for initial discussion; this will obviously require a non-patch release. Since the default behavior is unchanged, I would not expect tests to fail.

### DIFF
--- a/components/formats-api/src/loci/formats/IFormatReader.java
+++ b/components/formats-api/src/loci/formats/IFormatReader.java
@@ -354,6 +354,22 @@ public interface IFormatReader extends IFormatHandler, IPyramidHandler {
    */
   boolean isGroupFiles();
 
+  /**
+   * Set the fill value for undefined pixels.
+   * This can be used to override the default fill value
+   * defined in a reader.
+   *
+   * @param color value that will be used to fill pixel byte arrays
+   */
+  void setFillColor(Byte color);
+
+  /**
+   * Return the fill value for undefined pixels.
+   *
+   * @see #setFillColor(Byte)
+   */
+  Byte getFillColor();
+
   /** Returns true if this format's metadata is completely parsed. */
   boolean isMetadataComplete();
 
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -128,6 +128,7 @@ public final class ImageConverter {
   private boolean originalMetadata = true;
   private boolean noSequential = false;
   private String swapOrder = null;
+  private Byte fillColor = null;
 
   private IFormatReader reader;
   private MinMaxCalculator minMax;
@@ -259,6 +260,10 @@ public final class ImageConverter {
         else if (args[i].equals("-swap")) {
           swapOrder = args[++i].toUpperCase();
         }
+        else if (args[i].equals("-fill")) {
+          // allow specifying 0-255
+          fillColor = (byte) Integer.parseInt(args[++i]);
+        }
         else if (!args[i].equals(CommandLineTools.NO_UPGRADE_CHECK)) {
           LOGGER.error("Found unknown command flag: {}; exiting.", args[i]);
           return false;
@@ -334,7 +339,7 @@ public final class ImageConverter {
       "    [-nolookup] [-autoscale] [-version] [-no-upgrade] [-padded]",
       "    [-option key value] [-novalid] [-validate] [-tilex tileSizeX]", 
       "    [-tiley tileSizeY] [-pyramid-scale scale]", 
-      "    [-swap dimensionsOrderString]",
+      "    [-swap dimensionsOrderString] [-fill color]",
       "    [-pyramid-resolutions numResolutionLevels] in_file out_file",
       "",
       "            -version: print the library version and exit",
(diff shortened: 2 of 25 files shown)


dgault (approved):
Overall the changes here look good to me. Using Byte as opposed to Color definitely simplifies things and still gives the end user flexibility to control as they wish. This seems like it would be suitable for anyone using either via the conversion tools or a developer setting the value for a viewer.

All builds and tests have continued to pass with this included and the default behaviour remains unchanged. The CZI changes have also maintained the default behaviour added in https://github.com/ome/bioformats/pull/3795

Tested using showinf and bfconvert tool with the recommended CZI files 
- curated/zeiss-czi/qa-10972
- curated/zeiss-czi/qa-10448
- curated/zeiss-czi/qa-17711
- curated/zeiss-czi/zeiss/zen-2012/Young_mouse

For each, with showinf the default fill value is seen (white for the brightfield samples). When using the new fill option the selected fill value is now being used as expected. 

Similarly with bfconvert, using the tool without the new option results in the default fill value being used in the converted image. Then when retesting with the fill value set it results in the converted file using the specified fill value as expected.

PR looks good from my perspective, happy to get this merged and included for 6.13.0.

@melissalinkert, can you also open a PR against bioformats-documentation to add the new option to the command line tools docs for [conversion](https://github.com/ome/bio-formats-documentation/blob/master/sphinx/users/comlinetools/conversion.rst) and [showinf](https://github.com/ome/bio-formats-documentation/blob/master/sphinx/users/comlinetools/display.rst)

components/formats-api/src/loci/formats/IFormatReader.java
```diff
@@ -354,6 +354,22 @@ Object openPlane(int no, int x, int y, int w, int h)
...
   boolean isGroupFiles();
 
+  /**
+   * Set the fill value for undefined pixels.
+   * This can be used to override the default fill value
+   * defined in a reader.
+   *
+   * @param color value that will be used to fill pixel byte arrays
+   */
+  void setFillColor(Byte color);
```
sbesson, line 364: Do these new methods to the `IFormatReader` interface constitute backward incompatible changes since third-party reader which does not extend `FormatReader` will need to be updated to implement them?

melissalinkert: Yes. Anything that extends `FormatReader` (or any other existing reader implementation) shouldn't require changes, but just implementing `IFormatReader` means that `setFillColor(Byte)` will need to be implemented as part of updating to 6.13.0.

Issue #3961, opened 2023-03-09
Import failure of Ionpath MIBITIFF with ImageJ and Bio-Formats 6.10 / 6.12 (current)


christianrickert:
There might be a issue with Bio-Formats when importing of Ionpath MIBITIFF-compatible files in ImageJ 1.54c (Java 1.8.0_322, 64-bit):

Example: [[FOV1-1] MoQC-100.zip](https://github.com/ome/bioformats/files/10935583/FOV1-1.MoQC-100.zip)

![image](https://user-images.githubusercontent.com/19319377/224143915-c31105a7-dca8-4c3e-a9d2-c09178500d2c.png)

Bio-Formats 6.10.1  (plugins/bio-formats_plugins-6.10.1-20220731.012559-15.jar), from [Bio-Formats](https://sites.imagej.net/Bio-Formats/) in ImageJ Updater
```
loci.formats.FormatException: Unexpected format in SIMS description JSON.
	at loci.formats.in.IonpathMIBITiffReader.initStandardMetadata(IonpathMIBITiffReader.java:142)
	at loci.formats.in.BaseTiffReader.initMetadata(BaseTiffReader.java:98)
	at loci.formats.in.BaseTiffReader.initFile(BaseTiffReader.java:610)
	at loci.formats.FormatReader.setId(FormatReader.java:1443)
	at loci.plugins.in.ImportProcess.initializeFile(ImportProcess.java:499)
	at loci.plugins.in.ImportProcess.execute(ImportProcess.java:142)
	at loci.plugins.in.Importer.showDialogs(Importer.java:140)
	at loci.plugins.in.Importer.run(Importer.java:76)
	at loci.plugins.LociImporter.run(LociImporter.java:78)
	at ij.IJ.runUserPlugIn(IJ.java:244)
	at ij.IJ.runPlugIn(IJ.java:210)
	at ij.IJ.runPlugIn(IJ.java:199)
	at loci.plugins.shortcut.ShortcutPanel$1.run(ShortcutPanel.java:189)
```

Bio-Formats 6.11.1 (plugins/bio-formats_plugins-6.11.1.jar), currently ships with Fiji
```
(no issue)
```

Bio-Formats 6.12.0 (plugins/bio-formats_plugins-6.12.0.jar), updated by ImageJ Updater's default sites
```
java.lang.IllegalStateException: Too early in import process: current step is FILE, but must be after STACK
[...]

PR #3966, opened 2023-03-21
Bump json dependency to 20220924

dgault:
This is a follow up to issue #3961. In the ImageJ environment, json-20220924 is available as a dependency through the updater. When used alongside Bio-Formats 6.12.0 this caused a failure in the Ionpath MIBITIFF reader due to an unexpected return type.

This PR will cause breakages in the repo tests and is aimed to test the full impact of readers affected and issues caused by the upgrade.

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -163,7 +163,7 @@
     <dependency>
     	<groupId>org.json</groupId>
     	<artifactId>json</artifactId>
-    	<version>20090211</version>
+    	<version>20220924</version>
     </dependency>
     <dependency>
       <groupId>org.xerial</groupId>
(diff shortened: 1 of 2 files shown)


dgault:
In nightly tests the ionpath-mibi repo was the only format with failures, see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

dgault:
Added some small changes to the IonpathMIBITiffReader which hould make it compatible with both versions of the json dependency

melissalinkert:
This came up in a discussion with @sbesson and @muhanadz - I was concerned that `jzarr` might depend on some other version of `org.json:json`, in which case we'd need to be careful with ZarrReader, bioformats2raw, etc.

I was wrong, though, `jzarr` in fact uses `jackson-databind` for JSON handling, and does not have a dependency on `org.json:json` at all.

sbesson:
@dgault should we get this merged so that we can consider https://github.com/ome/bioformats/pull/3978 as a follow-up?

sbesson (approved):

dgault:
Yeah, lets get this one in and consider [3978](https://github.com/ome/bioformats/pull/3978) as a follow up

PR #3967, opened 2023-03-22
Update kryo to 5.4.0

melissalinkert:
This keeps the default configuration from 4.0.2, as described in https://github.com/EsotericSoftware/kryo/wiki/Migration-to-v5#configuration-changes. We can debate whether that's appropriate going forward, but this was the least disruptive way to do the upgrade for now.

See https://github.com/ome/ome-common-java/pull/74 and https://forum.image.sc/t/plugin-maintainers-can-you-test-fiji-2-11-0/78852/14.

### DIFF
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -29,24 +29,9 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.openmicroscopy</groupId>
-      <artifactId>ome-common</artifactId>
-      <version>${ome-common.version}</version>
-    </dependency>
-    <dependency>
       <groupId>${ome-model.group}</groupId>
       <artifactId>ome-xml</artifactId>
       <version>${ome-model.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>${project.groupId}</groupId>
-      <artifactId>formats-api</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>${project.groupId}</groupId>
-      <artifactId>formats-bsd</artifactId>
-      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -29,9 +29,24 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.openmicroscopy</groupId>
+      <artifactId>ome-common</artifactId>
+      <version>${ome-common.version}</version>
+    </dependency>
+    <dependency>
       <groupId>${ome-model.group}</groupId>
       <artifactId>ome-xml</artifactId>
       <version>${ome-model.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>formats-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>formats-bsd</artifactId>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/components/bio-formats-tools/pom.xml
+++ b/components/bio-formats-tools/pom.xml
@@ -29,19 +29,9 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.openmicroscopy</groupId>
-      <artifactId>ome-common</artifactId>
-      <version>${ome-common.version}</version>
-    </dependency>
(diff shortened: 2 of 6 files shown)


melissalinkert:
I'm not entirely sure why `ant docs` fails here, possibly it's picking up kryo 4.0.2 (from ome-common-java) instead of 5.4.0?

sbesson:
Yes definitely feels like a Maven dependency resolution issue. Running `mvn dependency:tree` on this project confirms that `com.esotericsoftware:kryo:jar:5.4.0` is selected for the `formats-gpl` component but `com.esotericsoftware:kryo:jar:4.0.2` is selected for `bio-formats_plugins`, `bio-formats-tools`, `bioformats_package`. This is because both `ome-common` and `formats-gpl` are declared in the dependencies and since `kryo` is found at the same depth, the first declaration wins - https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html.

Once `ome-common` gets released with the kryo bump, versions should be consistent again but for now, we need a strategy to get both PR included in the nightly CI builds. One option might be to only declare one dependency in each pom i.e.

```
formats-bsd -> formats-api
formats-gpl -> formats-bsd
bio-formats-tools, bioformats_package, bio-formats_plugins -> formats-gpl
```

I expect this should be sufficient to have the kryo dependency declared at the `formats-gpl` level win the conflict resolution in Maven (Gradle always selects the highest version). Note this might have other implications including different resolution for other dependencies that we might need to review. Happy to discuss later today.

NicoKiaru:
What will happen with the existing 'old' memo files ? Is there already a mechanism that detects when a memo file fails and erases it ? Will it work for this transition ?

sbesson:
@NicoKiaru definitely something that should be part of the testing. Ideally, I would expect old memo files will fail to be deserialized using the new library, deleted and then recreated after the reader is reinitialised.

melissalinkert:
Thanks, @sbesson. e2c0a8c simplifies the dependencies as described, which does indeed fix the build.

> Ideally, I would expect old memo files will fail to be deserialized using the new library, deleted and then recreated after the reader is reinitialised.

That's my expectation as well.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#418](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/418/). See the [console output](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/418/consoleText) for more details.
Possible conflicts:
  - PR #3956 sbesson 'Bump sl4j-api to 2.x'
    - pom.xml

--conflicts

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#1571](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/1571/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/1571/consoleText) for more details.
Possible conflicts:
  - PR #3956 sbesson 'Bump sl4j-api to 2.x'
    - pom.xml

~~--conflicts~~ Conflict resolved in build [BIOFORMATS-push#1572](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/1572/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/1572/consoleText) for more details.

sbesson:
Given the conflicts, converted the set of slf4j api PRs into draft mode again and relaunched the nightly CI builds.

sbesson (commented):
The nightly CI builds have been successfully passing for the last three occurences i.e. using different JDKs with this dependency upgrade included. The automated tests include the memo file generation/loading for every supported file format.
Additionally, tested that memo files created by versions 6.12.0 and earlier are successfully deleted and regenerated e.g. in an OMERO context. Confirmed this is also the case for a sample of file formats/modalities. 

From my perspective, this should be suitable for inclusion in the upcoming minor release of Bio-Formats. Let's discuss at the weekly Formats meeting whether we want to carry out additional testing and the roadmap for this dependency review.

jburel:
I have also tested this PR with the "double-build" strategy (broken with logback changes)
This is green
See https://github.com/jburel/openmicroscopy/actions/runs/4530341369

sbesson (approved):
Following today's discussion, happy for this change to get in so that we can start evaluating other dependency upgrades.

On the pom.xml changes specifically, the declaration of all transitive dependencies seem to be an historical legacy of the project (specifically the addition of Maven as a build system alongside Ant). Since Ant is now using Maven to resolve dependencies, this is largely unnecessary and it should be possible to go one step further than what is proposed here. I can try and open an additional PR to further clean-up the dependency graph that we can evaluate.

dgault (approved):

sbesson (approved):

PR #3973, opened 2023-03-30
Remove unneeded joda-time dependency

dgault:
Following on from the dependency updates in https://github.com/ome/bioformats/issues/3970

The joda-time dependency is currently used in a number of different projects, all of which also include ome-common. Rather than managing the dependency in each individual project it could be easier to simply inherit from the upstream ome-common project.

See the conversion on https://github.com/ome/ome-common-java/pull/77

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -121,11 +121,6 @@
       <version>1.4.0</version>
     </dependency>
     <dependency>
-      <groupId>joda-time</groupId>
-      <artifactId>joda-time</artifactId>
-      <version>2.2</version>
-    </dependency>
-    <dependency>
       <groupId>com.esotericsoftware</groupId>
       <artifactId>kryo</artifactId>
       <version>${kryo.version}</version>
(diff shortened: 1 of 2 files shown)


sbesson (approved):
As mentioned in the linked discussion above, this feels like the most straightforward way to streamline the maintenance of this dependency. A single version is defined, managed and inherited by all components across the entire stack.

melissalinkert (approved):

jburel:
FYI this has broken the ``openmicroscopy`` build.
The dependency will need to be added to ``openmicroscopy``. Other projects might have to do the same adjustment
see https://github.com/ome/openmicroscopy/actions/runs/4613815262/jobs/8156187436

sbesson:
Note this build has been performed from a revision Bio-Formats which does not include `ome-common` with the `joda-time` upgrade. I would suggest we first get this component properly released and bumped in Bio-Formats and retest the impact on the downstream builds before considering options.

jburel:
Sounds good. But it will have been preferable not to merge the BF PR before we are ready since now 
the GAs will always fail on ``openmicroscopy`` repo

What is the timeline?

jburel:
I have turned off the build of BF for now see https://github.com/ome/openmicroscopy/commit/e23c38e6fc92e8681a8bae619019b49f51613d07

PR #3975, opened 2023-04-04
MRC: read IMOD flags to determine int8 vs uint8 pixel type

melissalinkert:
See https://github.com/glencoesoftware/bioformats2raw/issues/198#issuecomment-1494790737

The test file has been uploaded to `curated/mrc/bioformats2raw-198`. Without this PR, `showinf test_uint8.mrc` should read the file, but will incorrectly identify it as `int8`. With this PR, the same test should show a pixel type of `uint8`.

I would not expect test failures, and this should be safe for a patch release.

cc @blowekamp

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MRCReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MRCReader.java
@@ -198,11 +198,21 @@ public class MRCReader extends FormatReader {
     m.rgb = false;
 
     int mode = in.readInt();
+
+    // see the 'imodStamp' and 'imodFlags' fields in
+    // https://bio3d.colorado.edu/imod/doc/mrc_format.txt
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Changes looks good and inline with the specification linked from the reader. Nightly CI tests are passing.

I defer on the merging as we are prioritizing the dependencies upgrade work but from my side, this feels like a valuable fix to include in Bio-Formats 6.13.0.

dgault:
Changes here look good and inline with the spec. Testing with the new sample file and with the PR included the file is able to read without issue and `PixelType = uint8` which is the correct expected value. No other sample files have been impacted by the PR and all builds and repo tests continue to pass. Merging this for inclusion with 6.13.0.

PR #3976, opened 2023-04-07
Bump low level component versions

dgault:
Bumping the version of low level components as part of the dependency update work in https://github.com/ome/bioformats/issues/3970

- ome-common to 6.0.16
- ome-model to 6.3.3
- ome-codecs to 0.4.5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,13 +49,13 @@
     <slf4j.version>1.7.30</slf4j.version>
     <kryo.version>4.0.2</kryo.version>
     <testng.version>6.8</testng.version>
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Since the nightly CI builds bump the components to the local built versions, I suspect there is not much to be tested beyond the GitHub Actions validation.

Should we get this merged and update https://github.com/ome/bioformats/pull/3967 to remove the temporary commit?

melissalinkert (approved):
I'll push the update to #3967 when this is merged.

PR #3979, opened 2023-04-14
CellWorX/MetaXpress: fix an off-by-one when retrieving channel metadata

melissalinkert:
This only happened when the files for the entire first field were missing (in a multi-field dataset). The problem was introduced in https://github.com/ome/bioformats/pull/3806. Backported from a private PR.

To test, use the new artificial sample in `curated/metaxpress/samples/multi-channel-missing-corner`. This was generated from existing sample data to mimic the plates for which the problem was initially noticed.

Without this PR, `showinf -nopix -omexml multi-channel.HTD` should throw an exception indicating that the plane index exceeds the plane count. With this PR, `showinf -omexml -series 13 -minmax multi-channel.HTD` should successfully show two channels, neither of which is empty. The first 12 series should be readable, but with all 0 pixel values as the .HTD file shows these fields/wells as acquired but the files are missing.

I'd expect this to be safe for a patch release.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellWorxReader.java
@@ -395,7 +395,7 @@ public class CellWorxReader extends FormatReader {
       planeIndex = c;
       file = getFile(seriesIndex, planeIndex);
       while (!new Location(file).exists()) {
-        if (planeIndex < zSteps * nTimepoints * wavelengths.length) {
+        if (planeIndex < zSteps * nTimepoints * (wavelengths.length - 1)) {
           planeIndex += (zSteps * nTimepoints);
         }
         else if (seriesIndex < seriesCount - 1) {


sbesson (approved):
Same considerations as https://github.com/ome/bioformats/pull/3977#pullrequestreview-1391906162, we might want to discuss the timeline for shipping this as it effectively addresses a regression introduced in a previous version of Bio-Formats

dgault (approved):
Builds and tests have remained green with this PR and the new config PR included.

Tested using the new sample dataset `curated/metaxpress/samples/multi-channel-missing-corner`. Without the PR the file fails to open and an exception is thrown as below:

```
Exception in thread "main" java.lang.IllegalArgumentException: Invalid plane index: 2/2
	at loci.formats.FormatTools.getZCTCoords(FormatTools.java:646)
	at loci.formats.FormatTools.getZCTCoords(FormatTools.java:565)
	at loci.formats.FormatReader.getZCTCoords(FormatReader.java:1140)
	at loci.formats.in.CellWorxReader.getFile(CellWorxReader.java:722)
	at loci.formats.in.CellWorxReader.populateMetadata(CellWorxReader.java:408)
	at loci.formats.in.CellWorxReader.initFile(CellWorxReader.java:310)
```
With the PR included and testing with the same dataset, the image now opens without any exception.

None of the other existing datasets appear to have been impacted.

PR #3982, opened 2023-04-17
CV7000: ignore any extra folders mixed in with the plate data

melissalinkert:
Backported from a private PR.

This fixes an issue where if a CV7000 plate directory contains a subdirectory, the subdirectory path gets added to the used files list. The contents of the subdirectory (if any) were not included in the used files list.

To test, use `curated/cv7000/samples/extra-dirs`, which is created from an existing dataset. Without this PR, `showinf -nopix` on the .wpi file should include the subdirectory path (`.../DW 0003...`) in the used files list. With this PR, the same test should not show the subdirectory path at all.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -130,7 +130,7 @@ public class CV7000Reader extends FormatReader {
     ArrayList<String> files = new ArrayList<String>();
     files.add(new Location(currentId).getAbsolutePath());
     for (String file : allFiles) {
-      if (!files.contains(file) && (!noPixels || !checkSuffix(file, "tif"))) {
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested using the synthetic sample. Without this PR, the additional directory gets listed in `getUsedFiles`. This departs from the API contract and breaks the semantics of downstream consumers, in particular the import of these plates into OMERO is broken.

With the proposed fix, the directory is no longer listed in the output of `getUsedFiles` and the plate imports successfully into OMERO.

Possibly the primary decision is whether the content of these extra folders should be listed or not. We have precedents for both behaviors and this PR implements the simplest approach where all files which are not as the same level as the `.wpi` file are ignored.

As this effectively maintains the existing statu quo, I would vote for getting this in as it is as we have real examples of these datasets where our reader needs to be more defensive. We can always revisit in the future if some (all) of the additional folders need some registration .

PR #3984, opened 2023-04-19
IFormatReader: add default implementations for new setFillValue/getFillValue

melissalinkert:
Provides backwards compatibility, for any external readers that implement `IFormatReader` but do not extend an existing implementation.

This is a follow-up to #3963 and was suggested by @sbesson. There should be no change in behavior compared to #3963, since all of our readers extend `FormatReader`.


sbesson (approved):
This change ensures that the new methods added to the interface have a reasonable default implementation and enforce the backwards compatibility of the new API addition making it fully suitable for a minor release increment. As noted in the description, all the readers shipped by Bio-Formats extend `FormatReader` which overrides the default implementation and are unaffected.

PR #3985, opened 2023-04-24
Rewrite of ZeissCZIReader to support Tb files (fast blocks retrieval …

NicoKiaru:
…and low memory footprint).

Draft

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -23,7 +23,7 @@ jobs:
             ant docs
       - name: Upload artifacts
         # upload just one set of artifacts for easier PR review
-        if: matrix.os == 'ubuntu-latest' && matrix.java == '1.8'
+        if: matrix.os == 'ubuntu-latest' && matrix.java == '8'
         uses: actions/upload-artifact@v3
         with:
           path: artifacts/
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -28,8 +28,41 @@ jobs:
           cache: 'maven'
       - name: Build
         run: mvn ${{ env.maven_commands }}
-  deploy:
+  deploy_snapshots:
     if: ${{ github.ref == 'refs/heads/develop' && github.repository_owner == 'ome' }}
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Retrieve version
+        id: get_version
+        run: |
+          VERSION=$( mvn help:evaluate -Dexpression=project.version -q -DforceStdout )
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+      - name: Set server
+        id: set_server
+        run: |
+          if [[ ${{ steps.get_version.outputs.version }} =~ 'SNAPSHOT' ]]; then
+              echo server='ome.snapshots' >> $GITHUB_OUTPUT
+            else
+              echo server='ome.releases' >> $GITHUB_OUTPUT
+          fi
+      - name: Set up Repository
+        uses: actions/setup-java@v3
+        with:
+          java-version: 8
+          distribution: 'zulu'
+          server-id: ${{ steps.set_server.outputs.server }}
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+      - name: Deploy SNAPSHOT
+        if: ${{ steps.set_server.outputs.server == 'ome.snapshots' }}
+        run: mvn deploy
+        env:
+          MAVEN_USERNAME: ${{ secrets.CI_DEPLOY_USER }}
+          MAVEN_PASSWORD: ${{ secrets.CI_DEPLOY_PASS }}
+  deploy_tags:
+    if: startsWith(github.ref, 'refs/tags') && github.repository_owner == 'ome'
     needs: build
     runs-on: ubuntu-latest
     steps:
@@ -55,7 +88,8 @@ jobs:
           server-id: ${{ steps.set_server.outputs.server }}
           server-username: MAVEN_USERNAME
           server-password: MAVEN_PASSWORD
-      - name: Deploy SNAPSHOT
+      - name: Deploy Tags
+        if: ${{ steps.set_server.outputs.server == 'ome.releases' }}
         run: mvn deploy
         env:
           MAVEN_USERNAME: ${{ secrets.CI_DEPLOY_USER }}
--- a/components/bio-formats-plugins/pom.xml
+++ b/components/bio-formats-plugins/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>ome</groupId>
     <artifactId>pom-bio-formats</artifactId>
-    <version>6.12.1-SNAPSHOT</version>
+    <version>6.12.1-BIOP</version>
     <relativePath>../..</relativePath>
   </parent>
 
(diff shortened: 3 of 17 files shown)


NicoKiaru:
## This PR = refactoring the logic of CZI Reader

This is a very early draft related to https://github.com/ome/bioformats/issues/3839
I'll copy paste what's in the header of the reader that details the logic of the rewrite:

## Logic overview
------

ZeissCZIReader is the file format reader for Zeiss .czi files.
 See  [https://zeiss.github.io](CZI reference documentation)
 
 Essentially, all data is stored into subblocks where each subblock location is specified by its dimension indices.
 There are standard spatial and time dimensions, as well as extra ones necessary to describe channel, scenes,
 acquisition modalities, etc:
 *  X,Y,Z, // 3 spaces dimension
 * T, Time
 * M, Mosaic but why is there no trace of it in libczi ???
 * C, Channel
 * R, Rotation
 * I, Illumination
 * H, Phase
 * V, View
 * B, Block = deprecated
 * S  Scene
 
 A subblock may represent a lower resolution level. How to know this ? Because its stored size (x or y) is lower
 than its size (x or y). Its downscaling factor can thus be computed the ratio between stored size and size.
 For convenience, this reader adds the downscaling factor as an extra dimension named `PY`
 
A CZI file consists of several segments. The majority of segments are data subblocks, as described before. But other segments are present. Essentially this reader reads the {@link loci.formats.in.libczi.LibCZI.FileHeaderSegment} that contains some metadata as well as the location of the {@link loci.formats.in.libczi.LibCZI.SubBlockDirectorySegment}


The SubBlockDirectorySegment is a critical segment because it contains the dimension indices and file location of all data subblocks. Thus, by reading this segment only, there is no need to go through all file segments while initializing the reader.

Using this initial reading of the directory segment, all dimensions and all dimension ranges are known in advance. This is used to compute the number of core series of the reader, as well as the resolution levels. This is done by creating a core series signature {@link CoreSignature} where the dimension are sorted according to a priority {@link ZeissCZIReader#dimensionPriority(String)}. If autostitching is true, all mosaics belong to the same core series. If autostitching is false, each mosaic is split into different core series.
(core series = series + resolution level)

Notes:
1. It is assumed that all subblocks from a single core index have the same compression type {@link ZeissCZIReader#coreIndexToCompression}

2. This reader is not thread safe, you can use memoization or {@link ZeissCZIReader#copy()} to get a new reader and perform parallel reading.

3. This reader is optimized for fast initialisation and low memory footprint. It has been tested to work on Tb czi size files. To save memory, the data structures used for reading are trimmed to the minimal amount of data necessary for the reading at runtime. To illustrate this point, for a 6Tb dataset, each 'int' saved per block saves 7Mb (in RAM and in memo file). Trimmimg down libczi dimension entries to {@link MinimalDimensionEntry} leads to a memo file of around 100Mb for a 4Tb czi file. Its initialisation takes below a minute, with memo building. Then a few seconds to generate a new reader from a memo file is sufficient.

4. Even with memoization, at runtime, a reader for a multi Tb file will take around 300Mb on the heap. While this is reasonable for a single reader, it becomes an issue to create multiple readers for parallel reading: 10 readers will take 3 Gb. Thus the method {@link ZeissCZIReader#copy()} exist in order to create a new reader from an existing one, which saves memory because it reuses all fields from the previous reader. Using this method, 10 readers can be created to read in parallel en single czi file, but it will use only the memory of one reader.
WARNING: calling {@link ZeissCZIReader#close()} on one of these readers will prevent the use of all the other readers created with the copy method!

The annotation {@link CopyByRef} is used to annotate the fields that should be initialized in the duplicated reader using the reference of the model one, see the constructor with the reader in argument.

5. This reader uses the class {@link LibCZI} which contains the czi data structure translated to Java and which contains very little logic related to the reader itself, which should be in this class.

## A few TODOs...
- test auto-stitching
- exposure time read from subblock do not seem to work
- use seriesToOx and seriesToOy for correct series positioning in XY (center or topleft?)
- check camera orientation with regards to origin ?
- add multi-part file support
- test RGB file
- test compressed files
- position from subblock originx and originy works, but it does not seem to work with some other images
- get optimal tile size should vary depending on compression: on raw data it's easy to partially read planes, but for compressed data that's much harder so it would be better to read the whole block rather that decompressing it multiple times the same block to extract a partial region

Features:
- add two methods that map forth and back czi dimension indices to bio-formats series
- add a method that returns a 3D matrix per series (for lattice skewed dataset?)

----

## What next ?

It's a long term project to potentially get this merged. Right now it's the only reader that can open reasonably fast Lattice Light Sheet datasets. But many of the features from the previous reader are probably missing.

In the future:
- either you think it's a good idea to try to get this merged and I take the time to edit this reader to support the various use-cases from the previous reader
- either there's a way to swap the reading to this reader when a lattice light sheet dataset is detected. And thus there is a ZeissCZIReader and a ZeissLatticeCZIReader

Currently, I'll put a custom version of this jar in a 'Lattice CZI' update site in Fiji, so that it does mess with anything, and I can use it for my data.

Clearly it's a very early draft version, but happy to discuss if you think it's interesting

dgault:
Thanks @NicoKiaru for the excellent effort to work on this. It will likely be next week before I get a chance to take a serious deep dive into this and start testing. The key factor in deciding how to handle this will be in determining how many different use cases and variations of the format have been impacted. As it is one of the more complex formats with many different variations, I suspect it may be likely that we look at having a separate ZeissLatticeCZIReader as an initial step in getting this integrated and then have the option of aligning with the original reader as a follow up over time.

NicoKiaru:
@dgault thanks for looking at it. But this can wait a bit: I'll do a few more modfications to include (what I understood of) the previous metadata logic.

Quick question: do you think it is a good idea to use JAXB to map Zeiss ImageDocument xml to a Java structure instead of the succession of getroot, getchildren, that's current used to read xml ? Maybe it's a bad idea because of xml variants ? And I do not know if the xml structure of the ImageDocument is public.

dgault:
Temporarily removing this from draft in order to test it with our data repository and get an indication of the impact on existing CZI variants.

NicoKiaru:
Ah, let me do a small push then, I've added quite a few metadata part back

NicoKiaru:
Hmm I've messed up the version, sorry, that was because I made a partial release for an experimental update site.

NicoKiaru:
I've just discovered https://docs.openmicroscopy.org/ome-files-cpp/0.5.0/ome-model/manual/html/developers/6d-7d-and-8d-storage.html.

Is it possible to precise modulo in series ?

dgault:
Currently you can only use ModuloAlongZ, ModuloAlongT and ModuloAlongC. The CZI reader previously had used these for rotations, illuminations and phases - https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java#L815-L843

When included with the tests last night it threw a number of exceptions just when trying to create the test instance (https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console). Most of the failures occurred when calling [populatePixels](https://github.com/ome/bioformats/pull/3985/files#diff-29b01fdc8a1f59d6bcd347df58cf572f332d56e81aad01ae82f69a6a821e0aa6R1134) and it seems that one of the dimensions Z, C or T at that point in time was 0 for some filesets resulting in `ArithmeticException: / by zero`.

It also seemed that a number of open files had not been closed properly during the close method.

NicoKiaru:
I close this PR because I was a bit lost in the branches and remote repos.

The new PR is https://github.com/ome/bioformats/pull/4009

PR #3990, opened 2023-04-27
Point to Release

jburel:
Switching from staging to releases implies that a commit without snapshot could be pushed to artifacts.o.org under releases.
This PR introduces a new job.
* the artifacts will be pushed to ``snapshots`` only when ``SNAPSHOT`` is in the version
* the artifacts will be pushed to ``releases`` only when ``SNAPSHOT`` is **not** in the version and a tag is pushed. (just in case a tag is made by accident with snapshot)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -28,8 +28,41 @@ jobs:
           cache: 'maven'
       - name: Build
         run: mvn ${{ env.maven_commands }}
-  deploy:
+  deploy_snapshots:
     if: ${{ github.ref == 'refs/heads/develop' && github.repository_owner == 'ome' }}
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Retrieve version
+        id: get_version
+        run: |
+          VERSION=$( mvn help:evaluate -Dexpression=project.version -q -DforceStdout )
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+      - name: Set server
+        id: set_server
+        run: |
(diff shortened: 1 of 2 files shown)


sbesson (commented):
If I understand correctly the logic: 
- SNAPSHOT version / merge commit -> deploy to `ome.snapshots`
- release version / merge commit -> no deployment
- SNAPSHOT version / tag -> no deployment
- release version / tag -> deploy to `ome.releases`

If my reading is correct, this seems reasonable and should match our current release workflow.  Two additional comments:
- as the current deployment failure has shown, it is extremely hard to validate these deployment steps as they are skipped on Pull Request. Is there a test Maven project that could be used for testing? Or is the idea to have code review only, get this in and test a SNAPSHOT build and a tag?
- there is a lot of redundancy between the two `deploy` jobs especially since the only element that truly varies is the condition for executing `mvn deploy`. As long as we are all comfortable on maintaining this workflow, happy to go for this variant though.

jburel:
Your understanding is correct.
There is redundancy but unless we go for something similar to https://github.com/ome/action-workflows/blob/main/.github/workflows/gradle_publish.yml. This is difficult to avoid
We could look into that strategy post release since it will add complexity to the setup/review at this stage

dgault:
The overall logic here makes sense to me. I will push the 6.13.0 release until next week so there is time to review this and not rush it through

jburel:
Merging this PR will test the "snapshots" deployment 
@dgault doing a release candidate will validate the "releases" deployment

melissalinkert (approved):
I don't see anything obviously problematic. :+1: to merging and using a RC to validate that `ome.releases` is still used correctly.

dgault (approved):

PR #4008, opened 2023-05-17
KLB: Use z index to increment block input offset

dgault:
This is a follow up to an issue discovered on idr0044 in which scrolling through z planes only incremented every 8 frames.

The issue can be reproduced with Bio-Formats 6.13.0 using all of the sample files from https://downloads.openmicroscopy.org/images/KLB/

In each of the samples above the z dimension has a block size of 8. The reader was choosing the correct block but was not correctly setting the Z offset within the block, which is why the planes updated every 8 steps.

To test use any of the public samples, img.klb from https://downloads.openmicroscopy.org/images/KLB/samples/ makes for a nice quick test file, but this should also be tested against the idr0044 data. With the PR, scrolling through Z should show the planes update with each step.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/KLBReader.java
+++ b/components/formats-bsd/src/loci/formats/in/KLBReader.java
@@ -261,6 +261,13 @@ public class KLBReader extends FormatReader {
           int imageRowSize = w * bytesPerPixel;
           int blockRowSize = blockSizeAux[0] * bytesPerPixel;
           int fullBlockRowSize = dims_blockSize[0] * bytesPerPixel;
+          
+          // Actual block values used as offsets for  Z planes
+          // This covers the use case when you have a block that overlaps a tile and also 
+          // is a partial block at the end of a row or column
+          int actualBlockWidth = Math.min(dims_blockSize[0], (getSizeX() - coordBlock[0]));
+          int actualBlockHeight = Math.min(dims_blockSize[1], (getSizeY() - coordBlock[1]));
+          int actualBlockPlaneSize = bytesPerPixel * actualBlockHeight * actualBlockWidth;
 
           // Location in output buffer to copy block
           outputOffset = (imageRowSize * (coordBlock[1] - y)) + ((coordBlock[0] - x) * bytesPerPixel);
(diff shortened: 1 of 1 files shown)


dominikl (approved):
Tested with `showinf` tool and https://downloads.openmicroscopy.org/images/KLB/samples/img.klb . Looks good. Could see the jump every 8 planes with the release build, and the smooth scrolling through Z with this PR. Couldn't test against idr0044 image, because they are too big.

dgault:
Most recent commit should hopefully resolve the tiling issues with https://idr-testing.openmicroscopy.org/webclient/img_detail/4007801/?dataset=3351

As before you should still be able to scroll through the Z stack with smooth transitions for each frame. 

- To test the tiled scenario I used the partial idr dataset from https://downloads.openmicroscopy.org/images/KLB/idr0044/
- I then picked a plane that had very obvious tiling issues and converted that single plane using:
 ```
 ./bfconvert -tilex 32 -tiley 32 -series 0 -range 21239 21239 /Mmu_E1_CAGTAG1.TM000000_timeFused_blending/SPM00_TM000000_CM00_CM01_CHN00.fusedStack_xyProjection.corrected.shifted.klb KLB-tileTest.ome.tiff
 ```
- Without the PR the resulting image should have very obvious issues with the tiled reading
- With the PR included (replacing the bioformats_package.jar in bftools) the resulting converted image should open and display as expected.

sbesson (dismissed):
Tested within the context of an IDR deployment. Without this fix, testing any of the multi-Z images of https://idr.openmicroscopy.org/webclient/?show=project-502 shows that the plane data is updated every 8 Z-sections.

With this fix, the planes are updated for each increment of the Z-slider and the tile-based reading does not create any artifact. Also confirmed that `idr0051` which also uses the same reader but is unaffected by the issue still reads accurately with this change included.

dgault:
Unfortunately a few repo test failures for the smaller img.klb sample, checking the failures now
https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

dgault:
@sbesson, repo tests with this PR are now green again so should be ready for a retest on IDR whenever suits

sbesson (approved):
Looks good to me at least after deploying the latest fix in a pilot IDR server.
I would certainly wait until this has been tested against all combinations of JDK using the new OME CI infrastructure. Afterwards, this can be safely included in the upcoming Bio-Formats release from my perspective.

melissalinkert:
:+1: merging as discussed during today's PR review.

PR #4009, opened 2023-05-22
ZeissCZIReader rewrite - fast opening by reading directory segments.

NicoKiaru:
Not all functionalities supported yet.

This is a new PR that replaces https://github.com/ome/bioformats/pull/3985

The reason is just to keep things cleaner on the git side.

-------------------

## This PR = refactoring the logic of CZI Reader

This is a very early draft related to https://github.com/ome/bioformats/issues/3839
I'll copy paste what's in the header of the reader that details the logic of the rewrite:

## Logic overview
------

ZeissCZIReader is the file format reader for Zeiss .czi files.
 See  [https://zeiss.github.io](CZI reference documentation)
 
 Essentially, all data is stored into subblocks where each subblock location is specified by its dimension indices.
 There are standard spatial and time dimensions, as well as extra ones necessary to describe channel, scenes,
 acquisition modalities, etc:
 *  X,Y,Z, // 3 spaces dimension
 * T, Time
 * M, Mosaic but why is there no trace of it in libczi ???
 * C, Channel
 * R, Rotation
 * I, Illumination
 * H, Phase
 * V, View
 * B, Block = deprecated
 * S  Scene
 
 A subblock may represent a lower resolution level. How to know this ? Because its stored size (x or y) is lower
 than its size (x or y). Its downscaling factor can thus be computed the ratio between stored size and size.
 For convenience, this reader adds the downscaling factor as an extra dimension named `PY`
 
A CZI file consists of several segments. The majority of segments are data subblocks, as described before. But other segments are present. Essentially this reader reads the {@link loci.formats.in.libczi.LibCZI.FileHeaderSegment} that contains some metadata as well as the location of the {@link loci.formats.in.libczi.LibCZI.SubBlockDirectorySegment}

The SubBlockDirectorySegment is a critical segment because it contains the dimension indices and file location of all data subblocks. Thus, by reading this segment only, there is no need to go through all file segments while initializing the reader.

Using this initial reading of the directory segment, all dimensions and all dimension ranges are known in advance. This is used to compute the number of core series of the reader, as well as the resolution levels. This is done by creating a core series signature {@link CoreSignature} where the dimension are sorted according to a priority {@link ZeissCZIReader#dimensionPriority(String)}. If autostitching is true, all mosaics belong to the same core series. If autostitching is false, each mosaic is split into different core series.
(core series = series + resolution level)

Notes:
1. It is assumed that all subblocks from a single core index have the same compression type {@link ZeissCZIReader#coreIndexToCompression}

2. This reader is not thread safe, you can use memoization or {@link ZeissCZIReader#copy()} to get a new reader and perform parallel reading.

3. This reader is optimized for fast initialisation and low memory footprint. It has been tested to work on Tb czi size files. To save memory, the data structures used for reading are trimmed to the minimal amount of data necessary for the reading at runtime. To illustrate this point, for a 6Tb dataset, each 'int' saved per block saves 7Mb (in RAM and in memo file). Trimmimg down libczi dimension entries to {@link MinimalDimensionEntry} leads to a memo file of around 100Mb for a 4Tb czi file. Its initialisation takes below a minute, with memo building. Then a few seconds to generate a new reader from a memo file is sufficient.

4. Even with memoization, at runtime, a reader for a multi Tb file will take around 300Mb on the heap. While this is reasonable for a single reader, it becomes an issue to create multiple readers for parallel reading: 10 readers will take 3 Gb. Thus the method {@link ZeissCZIReader#copy()} exist in order to create a new reader from an existing one, which saves memory because it reuses all fields from the previous reader. Using this method, 10 readers can be created to read in parallel en single czi file, but it will use only the memory of one reader.
WARNING: calling {@link ZeissCZIReader#close()} on one of these readers will prevent the use of all the other readers created with the copy method!

The annotation {@link CopyByRef} is used to annotate the fields that should be initialized in the duplicated reader using the reference of the model one, see the constructor with the reader in argument.

5. This reader uses the class {@link LibCZI} which contains the czi data structure translated to Java and which contains very little logic related to the reader itself, which should be in this class.

## A few TODOs...
- test auto-stitching
- exposure time read from subblock do not seem to work
- use seriesToOx and seriesToOy for correct series positioning in XY (center or topleft?)
- check camera orientation with regards to origin ?
- add multi-part file support
- test RGB file
- test compressed files
- position from subblock originx and originy works, but it does not seem to work with some other images
- get optimal tile size should vary depending on compression: on raw data it's easy to partially read planes, but for compressed data that's much harder so it would be better to read the whole block rather that decompressing it multiple times the same block to extract a partial region

Features:
- add two methods that map forth and back czi dimension indices to bio-formats series
- add a method that returns a 3D matrix per series (for lattice skewed dataset?)

----

## What next ?

It's a long term project to potentially get this merged. Right now it's the only reader that can open reasonably fast Lattice Light Sheet datasets. But many of the features from the previous reader are probably missing.

In the future:
- either you think it's a good idea to try to get this merged and I take the time to edit this reader to support the various use-cases from the previous reader
- either there's a way to swap the reading to this reader when a lattice light sheet dataset is detected. And thus there is a ZeissCZIReader and a ZeissLatticeCZIReader

Currently, I'll put a custom version of this jar in a 'Lattice CZI' update site in Fiji, so that it does mess with anything, and I can use it for my data.

Clearly it's a very early draft version, but happy to discuss if you think it's interesting

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -9,33 +9,21 @@
  * %%
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as
- * published by the Free Software Foundation, either version 2 of the 
+ * published by the Free Software Foundation, either version 2 of the
  * License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
- * You should have received a copy of the GNU General Public 
+ *
+ * You should have received a copy of the GNU General Public
  * License along with this program.  If not, see
  * <http://www.gnu.org/licenses/gpl-2.0.html>.
  * #L%
  */
 
 package loci.formats.in;
-
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.IOException;
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import javax.xml.parsers.DocumentBuilder;
 
 import loci.common.ByteArrayHandle;
 import loci.common.Constants;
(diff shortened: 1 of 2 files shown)


NicoKiaru:
List of publicly available czi files:

- https://zenodo.org/record/7129425 (plate)
- https://zenodo.org/record/7254229
- https://zenodo.org/record/4662053 (Airyscan and unprocessed Airyscan)
- https://zenodo.org/record/6848342 (Airyscan)
- https://zenodo.org/record/7015307 (demo camera images by Sebastien Rhode)
- https://zenodo.org/record/7254229
- https://zenodo.org/record/7117784 (lattice lightsheet dataset)
- https://zenodo.org/record/7147844 (plate)
- https://zenodo.org/record/7260610
- https://zenodo.org/record/5101351
- https://zenodo.org/record/3991919 (multi scene)
- https://zenodo.org/record/7430767

- https://rodare.hzdr.de/record/1849 (zipped parts)

I'll edit this comment to add more dataset

What's missing for sure:
- Multi part file (different angles)
- PALM

dgault:
Thanks thats a great start. The only public CZI files we currently have are at https://downloads.openmicroscopy.org/images/Zeiss-CZI/
We will use the same location to add any additional files we are able to make public.

For distributing this as an external jar initially a good example of a small external reader would be https://github.com/ome/ZarrReader. The package names can be unique however, it isn't required to keep the `loci.formats.in`. Once the naming is agreed I can open a PR to add it to the readers list (it is a simple online change similar to https://github.com/ome/bioformats/pull/3639/files).

NicoKiaru:
For the multi part files, we have a system at the BIOP that can generate such a dataset. But for PALM data I'm stuck.

NicoKiaru:
Ok, here's finally the czi external reader:

https://github.dev/BIOP/quick-start-czi-reader/blob/main/src/main/java/loci/formats/in/ZeissQuickStartCZIReader.java

I kept `loci.formats.in` because I like the consistency of having all the readers under the same package name. Hope that's fine.

So, could you add `loci.formats.in.ZeissQuickStartCZIReader` into the readers file or do you want me to file a PR ?

joshmoore:
Couple of things:
 * :+1: though maybe some discussion around the package name :wink:
 * non github.dev link for the interested: https://github.com/BIOP/quick-start-czi-reader/blob/main/src/main/java/loci/formats/in/ZeissQuickStartCZIReader.java
 * so this PR can be closed?
 * feel free to open the PR adding an `[external]` entry

NicoKiaru:
Yes, you can close the PR. 

> though maybe some discussion around the package name

You prefer the package name to be different ? Like `ch.epfl.biop` ? I'm not aware of any other reader that's outside `loci.format.in` I feared that I could not access protected fields if I was not in the same package, but I can try to change it.

> so this PR can be closed? 
> feel free to open the PR adding an [external] entry

Yes, maybe after this discussion about the package name ?

NicoKiaru:
> I feared that I could not access protected fields if I was not in the same package, but I can try to change it.

Nope, seems fine. Better like that ? 

`ch.epfl.biop.formats.in.ZeissQuickStartCZIReader`

I hope it won't break any logic elsewhere

NicoKiaru:
PR opened in https://github.com/ome/bioformats/pull/4019

I'm closing this PR. I'll move the discussion to issues of https://github.com/BIOP/quick-start-czi-reader

PR #4006, opened 2023-05-15
CV7000: handle case where wells are recorded, but all files were removed

melissalinkert:
See #4005.

I think this actually does fix the issue in #4005, but needs further testing. Mostly opening this PR now to see if/how many tests fail. At the moment, I'd expect memo file regeneration to be required, but not re-import.

PR #4011, opened 2023-05-22
CV8000: fix channel indexing

melissalinkert:
Backported from private PR. I don't think this will conflict with #4006.

This is meant to handle a case where multiple channels are defined in `MeasurementDetail.mrf`, but a non-consecutive subset of those channels were acquired. In the original (private) data, 6 channels are defined but only the first and 5th were acquired. Without this PR, that meant that `showinf -omexml` on the .wpi file reported 5 channels per series - but the middle 3 channels were completely blank images.

With this PR, the same test reports 2 channels per series, matching what was actually acquired.

We'll likely need to generate test data that demonstrates the problem; opening this now to verify that it does not break existing data.

The original dataset for this issue was CV8000 (not CV7000), so 4d9c6ea adds some extra logging to help distinguish between CV7000 and CV8000 (or later systems) as the former has better support than the latter.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/CV7000Reader.java
@@ -306,6 +306,7 @@ public class CV7000Reader extends FormatReader {
 
     fields = 0;
     HashSet<Integer> uniqueWells = new HashSet<Integer>();
+    HashSet<Integer> uniqueChannels = new HashSet<Integer>();
 
     for (Plane p : planeData) {
       if (p != null) {
@@ -333,12 +334,17 @@ public class CV7000Reader extends FormatReader {
         if (p.z < m.minZ) {
           m.minZ = p.z;
         }
+
+        // min and max channel indexes not currently used,
+        // but continue to calculate for completeness
+        // they may be needed in future CV7000/8000 work
         if (p.channelIndex > m.maxC) {
           m.maxC = p.channelIndex;
         }
         if (p.channelIndex < m.minC) {
           m.minC = p.channelIndex;
         }
+        uniqueChannels.add(p.channelIndex);
 
         if (p.field.field >= fields) {
           fields = p.field.field + 1;
@@ -360,6 +366,9 @@ public class CV7000Reader extends FormatReader {
     Arrays.sort(wells);
     reversePlaneLookup = new int[realWells * fields][];
 
+    Integer[] channelIndexes = uniqueChannels.toArray(new Integer[uniqueChannels.size()]);
+    Arrays.sort(channelIndexes);
+
     for (int i=0; i<realWells * fields; i++) {
       if (i > 0) {
         core.add(new CoreMetadata(core.get(0)));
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/CV7000Reader.java
```diff
@@ -306,6 +306,7 @@ public int compare(Channel c1, Channel c2) {
 
     fields = 0;
     HashSet<Integer> uniqueWells = new HashSet<Integer>();
+    HashSet<Integer> uniqueChannels = new HashSet<Integer>();
```
sbesson, line 309: Mostly a semantics question but where do you see the delineation between `unique{Wells,Channels}` and `acquiredWells` e.g. as introduced in https://github.com/ome/bioformats/pull/4006?

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
```diff
@@ -382,6 +391,13 @@ public int compare(Channel c1, Channel c2) {
     extraFiles = new ArrayList<String>();
     for (int i=0; i<planeData.size(); i++) {
       Plane p = planeData.get(i);
+
+      // reindex so that the plane's channel index is into
+      // the list of unique acquired channels, not the list of all channels
+      // that might have been acquired
+      // this eliminates the need to correct for the minimum C index later on
+      p.channelIndex = Arrays.binarySearch(channelIndexes, p.channelIndex);
```
sbesson, line 399: Understanding this would cause a serialization breaking change as the `PlaneData` element would need an extra field, do we anticipate scenarios where preserving the original channel index (using the channels declared in the original metadata file as the reference) would be useful?

components/formats-gpl/src/loci/formats/in/CV7000Reader.java
melissalinkert: `uniqueWells` and `uniqueChannels` are only indexes for valid wells/channels, and I wouldn't think of them as useful on their own - they're intermediate data used to set up the Plate/Well/WellSample/Image links and plane/file mappings.

`acquiredWells` is really just a cache - we could remove it and expect the same functionality, just maybe slower. 

I'm not particularly attached to any of these names, so happy to take other suggestions if renaming would help.

sbesson: Thanks, this was mostly for my own curiosity. From a quick grep, `unique<Element>s` is already used widely across our readers so happy to continue with this pattern.

dgault:
Do we know if this specifically related to CV8000? Is this likely to be common with those datasets or is it a more unique scenario?

In such a case, when it does occur I can see it making sense to update the indexing as has been here. The code change itself looks fine to me and hasn't had any negative regressions with existing datasets.

melissalinkert:
I don't believe this is specific to CV8000, it's just that a CV8000 plate was the first time we encountered this specific scenario. My understanding is that similar workflows are possible with CV7000.

sbesson (approved):
This PR has been tested with a sample plate which unfortunately cannot be shared. Trying to summarize the main features of the `.mes` file:

- 6 channels (1 to 6) are defined in the `bts:ChannelList` element
- the `bts:ActionList` element contains two `bts:ActionAcquire3D` elements i.e. 3D acquisition, the first one associated with channel 3 and the second one associated with channel 1

Without this PR, the plate included 5 channels with all the middle channel containing black planes. With this PR included, the images from the same plate was detected as 2-channel matching the metadata from the measurement file.

The fact that more channels can be defined than the real number used in the action list is consistent with has been seen from the acquisition UI for a CV7000 system. This supports the statement made in https://github.com/ome/bioformats/pull/4011#issuecomment-1611816595 that this is likely not CV8000 specific.

The only addition I could think of would be to construct and configure a synthetic plate representative of the acquisition settings that could be added to the QA repository and tested in the nightly CI builds

dgault (approved):
Thanks, it sounds like the functionality is definitely needed and I certainly can't see it having any negative effect on other datasets. Looks good from side too.

PR #4013, opened 2023-05-24
Update CV7000 isThisType test to skip extended type checking

melissalinkert:
There is no need to iterate over each file and check the type if the results are just ignored anyway.

See https://github.com/openmicroscopy/data_repo_config/pull/572#issuecomment-1560636849. I'd expect this to noticeably reduce test run time for CV7000/8000 plates with lots of files.

If we're happy with this approach, it looks like there are a few other similar checks that could also be moved outside the file loop.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -2479,6 +2479,32 @@ public class FormatReaderTest {
       "open = " + isThisTypeOpen + ", !open = " + isThisTypeNotOpen);
   }
 
+  /**
+   * In most cases, this test will:
+   *  - get the lowest-level reader for the file being tested
+   *  - iterate over each used file in the fileset
+   *      * iterate over each available reader (i.e. all of readers.txt)
+   *          - check if the reader picks up the used file
+   *          - fail the test if either is true:
+   *              * the reader picks up the file, but is not an instance of the original lowest-level reader
+   *              * the reader does not pick up the file, and is an instance of the original lowest-level reader
+   *
+   * There are a number of special cases, as many formats cannot be reliably
+   * detected from all files in the fileset. TIFF-based formats are especially
+   * prone to this issue. When adding a special case, take care to choose
+   * the appropriate location (type checking can be expensive for filesets with many files):
+   *  - inside the used file loop, but outside the reader loop
+   *     * if some used files are not expected to be passed to setId,
+   *       and could be picked up by many different readers, then a special case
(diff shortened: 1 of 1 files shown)


sbesson (approved):
With this PR included, the automated tests against the curated CV7000 samples completed in ~2h ([log](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/244940/) instead of ~3h30 ([log](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/244776/)) previously.

The time increase was caused by the addition of a real-world CV7000 plate (originally mentioned in https://github.com/ome/bioformats/pull/3663) containing 15K files. Prior to this change, `IFormatIReader.isThisType` was invoked `nFiles x nReaders` times with the result effectively skipped for all files except for the master `.wpi` file. With this change, `IFormatIReader.isThisType` should only be invoked once per reader for the wpi file resulting in a drastic performance improvement.

Very supportive of applying similar changes to other formats especially HCS ones where representative samples with 1-10K files are fairly common. Were you thinking of opening per-format PRs or combine all `isThisType` improvements in a single one?

melissalinkert:
> Were you thinking of opening per-format PRs or combine all `isThisType` improvements in a single one?

Probably combine all changes into one PR - I can do that here if no objections.

melissalinkert:
All done in 023c257.

dgault (approved):
Changes here make sense and have not had any negative impact on the repo tests with the PR included. With the new dataset configs this PR does provide a substantial performance improvement. Overall this looks good to me.

sbesson (approved):
Sorry for the delay here, I had started measuring execution times across affected formats but this was right before OME CI became unavailable so I have lost all my metrics/links.

Adding to what's said above:
- the overall execution times of the repository jobs have not been adversely affected by this change
- for all the large (HCS) formats that I have checked individually, the proposed changes rather improved the execution time cutting down some unnecessary loops
- the block of text added to the top of `testIsThisType` is a very useful addition to document the intent of the test and the rationale for the different categories of manual exclusions

dgault:
With https://github.com/openmicroscopy/data_repo_config/pull/572#issuecomment-1560636849 now merged the latest CI completed the full repo tests in 212 minutes 28 seconds ([log](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/231108/console)). Merge-CI with this PR  completed in 135 minutes 26 seconds ([log](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/11387/console)). These times and improvements seem to be consistent with the previous testing.

PR #4014, opened 2023-06-05
Remove Woolz support

melissalinkert:
As discussed in today's formats meeting, this is backported from a private PR.

This will remove all support for reading and writing Woolz. It does not attempt to deprecate first, though we could do that separately as a first step. Removing this functionality is proposed because we have limited ability to test or otherwise maintain it, and the maintenance status of the underlying library (https://github.com/ma-tech/Woolz) is unclear.

This will be a breaking change which requires at least a minor version, and probably a major version. Excluding from builds for now.


dgault:
Apologies @melissalinkert, accidentally pulled in other commits while trying to resolve the conflicts, going to close this and reopen a clean PR

PR #4019, opened 2023-06-15
Update readers.txt to include an external CZI Reader (https://github.com/BIOP/quick-start-czi-reader)

NicoKiaru:
Adds ZeissQuickStartCZIReader as an alternative external czi reader. The alternative reader is located in https://github.com/BIOP/quick-start-czi-reader.


NicoKiaru:
Hi everybody, 

Do you think this PR could be merged before the 11th of July ? What would you need from me ? I have a workshop on the 11th of July, and it would simplify a lot the installation process if the SNAPSHOT version of bio-formats contains this PR.

If it's complicated, then let me know, I'll make an update site with some jars of mine overriding Fiji's bioformats version. 

Note: Am I correct to think that the bio-formats Fiji update site ships the latest snapshot ?

Best, 

Nicoals

dgault (approved):
Hi @NicoKiaru, we will get this merged and included with the Bio-Formats 6.14.0 release that will be out next week, so it should be live on the FIJI update site before the 11th July.

NicoKiaru:
Thanks!!

PR #4023, opened 2023-06-20
Add deprecation tags for Woolz format

dgault:
This is a precursor to the work in https://github.com/ome/bioformats/pull/4014

This will also be accompanied with a warning in the release announcement, alongside steps for users to convert their existing files to supported open formats.


melissalinkert (approved):
Looks good to me. Is the plan to add a deprecation note on the corresponding documentation page as well (in addition to release notes)?

dgault:
Yeah, that seems like a sensible idea, I will get a note added to the format page also

sbesson (approved):

PR #4025, opened 2023-06-22
MicroManager: Prevent infinite loop and don't swallow IOException

### LINKED ISSUE #4018: Infinite loop in MicromanagerReader
The MicromanagerReader can swallow an IOException and infinite loop. The offending method is:
```java
private void parsePosition(int posIndex) throws IOException, FormatException
```
I have a TIFF file that has an incorrect IFD (due to file truncation). When the code executes `int nIFDs = parser.getMainIFDs().size()` the method `getMainIFDs` throws an EOFException. This is caught within the loop and logged. But the loop iteration counter `plane` is not incremented. Thus the loop just repeats.
Reproduce with:
```
> showinf -version
Version: 6.13.0
Build date: 2 May 2023
VCS revision: afaff5ef177c51d99e673cad6031d31637b431eb

> showinf -nopix -debug mrc5\ 60_17_MMStack_Pos0.ome.tif
```
Debug output is:
```
...
Initializing reader
Location.mapFile: embedded-stream.raw -> null
Location.mapFile: embedded-stream.raw -> null
MicromanagerReader initializing mrc5 60_17_MMStack_Pos0.ome.tif
MicromanagerReader initializing mrc5 60_17_MMStack_Pos0.ome.tif
loci.formats.in.MicromanagerReader.initFile(mrc5 60_17_MMStack_Pos0.ome.tif)
Reading metadata file
Populating metadata
Finding image file names
Building list of TIFFs
Failed to read metadata from /data/images/laura/test/mrc5 60_17_MMStack_Pos0.ome.tif
java.io.EOFException: Attempting to read beyond end of file.
        at loci.common.NIOFileHandle.readShort(NIOFileHandle.java:452)
        at loci.common.NIOFileHandle.readUnsignedShort(NIOFileHandle.java:467)
        at loci.common.RandomAccessInputStream.readUnsignedShort(RandomAccessInputStream.java:658)
        at loci.formats.tiff.TiffParser.readTiffIFDEntry(TiffParser.java:1422)
        at loci.formats.tiff.TiffParser.getIFD(TiffParser.java:464)
        at loci.formats.tiff.TiffParser.getMainIFDs(TiffParser.java:243)
        at loci.formats.in.MicromanagerReader.parsePosition(MicromanagerReader.java:496)
        at loci.formats.in.MicromanagerReader.initFile(MicromanagerReader.java:308)
        at loci.formats.ImageReader.setId(ImageReader.java:863)
        at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:660)
        at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1043)
Caused by: java.nio.BufferUnderflowException: null
        at java.base/java.nio.Buffer.nextGetIndex(Buffer.java:651)
        at java.base/java.nio.HeapByteBuffer.getShort(HeapByteBuffer.java:350)
        at loci.common.NIOFileHandle.readShort(NIOFileHandle.java:450)
        ... 12 common frames omitted
...
(Output then loops starting at 'Failed to read metadata from')
```

If I remove the sidecar metadata.txt file written by MicroManager then Bioformats selects the `TiffReader`. This trickles up the EOFException and the user is notified that there is an error with the image.

This bug was identified during an import into OMERO which just hangs due to the infinite loop.

Image to reproduce the error is attached:

[mrc5 60_17_MMStack_Pos0.zip](https://github.com/ome/bioformats/files/11744606/mrc5.60_17_MMStack_Pos0.zip)

A simple fix is to increment `plane` in the catch block. But it is not clear why it is acceptable to swallow an IOException for this format inside the loop. The call to `private void parsePosition(String jsonData, int posIndex)` outside the loop is allowed to throw an IOException.

dgault:
Issue was raised in https://github.com/ome/bioformats/issues/4018

The cause of the problem is a truncated file with an invalid IFD, when attempting to read the file an infite loop is encountered. This is due to the IOException being swallowed by the reader and the current loop continuing without incrementing. As the exception is due to an invalid truncated file, I believe it is best to throw an exception in this instance rather than continuing to increment the loop.

A sample file is available in the issue which can be used to test the PR.
Without the PR it should result in an infinite loop
With the PR the exception should be thrown and the user alerted to the error.

Fixes #4018

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
+++ b/components/formats-bsd/src/loci/formats/in/MicromanagerReader.java
@@ -599,7 +599,8 @@ public class MicromanagerReader extends FormatReader {
         plane += ifds.size();
       }
       catch (IOException e) {
-        LOGGER.debug("Failed to read metadata from " + path, e);
+        LOGGER.error("Failed to read metadata from " + path, e);
+        throw new FormatException(e);
       }
     }
   }


melissalinkert (approved):
Looks fine to me. Tested with data in `inbox/gh-4018`, and can easily reproduce the loop without this PR. With this PR, the IOException is actually thrown. `tiffinfo` confirms that the .tif file in the dataset is truncated, so agreed that this is the best we can do.

sbesson (approved):

PR #4030, opened 2023-06-29
Deprecate LegacyND2Reader

sbesson:
This reader depends on an outdated DLL which has not been built in years, is untested and fully superseded by the new NativeND2Reader

In addition to the Java reader, the following components are scheduled for removal in Bio-Formats 7:
- `lib/LegacyND2Reader.dll`
- `components/formats-gpl/src/loci/formats/in/loci_formats_in_LegacyND2Reader.cpp`
- `components/formats-gpl/src/loci/formats/in/loci_formats_in_LegacyND2Reader.h`

Possibly biggest question is whether we want to keep the delegation logic implemented in `ND2Reader` which allows to select between the `LegacyND2Reader` and `NativeND2Reader`. Listing a series of options briefly discussed with @melissalinkert are:

1. redirect https://github.com/ome/bioformats/blob/971fb341745e5c7e4b0cbf8c2af59deacc73db75/components/formats-gpl/src/loci/formats/in/ND2Reader.java#L47 to `NativeND2Reader`.
2. keep `ND2Reader` but have it simply extend `NativeND2Reader`
3. remove `ND2Reader` and update `readers.txt` to use `NativeND2Reader`
4. remove the delegate reader and rename `NativeND2Reader` as `ND2Reader`

Given this is a backwards-incompatible change, my personal preference would go towards options 3 or 4 but definitely open to feedback. In case we go for the latter options, we might want to add other deprecation flags to some of these classes

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LegacyND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/LegacyND2Reader.java
@@ -38,7 +38,10 @@ import loci.formats.meta.MetadataStore;
 /**
  * LegacyND2Reader is a file format reader for Nikon ND2 files that uses
  * the Nikon ND2 SDK - it is only usable on Windows machines.
+ *
+ * @Deprecated LegacyND2Reader will be removed in Bio-Formats 7.0.0
  */
+@Deprecated
 public class LegacyND2Reader extends FormatReader {
 
   // -- Constants --
--- a/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/NativeND2Reader.java
@@ -70,7 +70,11 @@ import ome.units.UNITS;
  * (2) the JAI jar file precedes JJ2000 in the classpath.
  *
  * Thanks to Tom Caswell for additions to the ND2 metadata parsing logic.
+ *
+ * @Deprecated NativeND2Reader will be unified with ND2Reader in Bio-Formats 7.0.0
+ * and should no longer be used
  */
+@Deprecated
 public class NativeND2Reader extends SubResolutionFormatReader {
 
   // -- Constants --


melissalinkert (approved):
Option 4, 3, 2, 1 would be my order of preference for handling `ND2Reader`.

dgault:
We will also want to add warnings to the format pages for each of these similar to Woolz https://github.com/ome/bio-formats-documentation/pull/308

sbesson:
Adding to Monday's agenda, we should decide on our removal strategy - see https://github.com/ome/bioformats/pull/4030#pullrequestreview-1505615759. In particular if we decide to go with option 3 or and 4, I could add a commit to annotate the classes that will be renamed with `@deprecated` to notify developers.

joshmoore:
Brief discussion of an additional option:

 5. move all logic to ND2Reader and have NativeND2Reader subclass it

It's not a strong opinion, but as came out in the chat, I would tend to opt for the least breaking change (but perhaps only since if I started letting myself adjust namings, I probably wouldn't stop... 😉) So a slight preference for 1 (with a warning if anyone chooses `legacyReader`) but a slightly stronger partial-preference for 5>4 with a deprecation of the updated NativeND2Reader subclass.

sbesson:
Summarizing the weekly discussion, the preferred option seems to be migrate the `NativeND2Reader` logic into `ND2Reader`. For backwards compatibility,  `NativeND2Reader` might be retained and simplified as a simple extension for as mentioned in https://github.com/ome/bioformats/pull/4030#issuecomment-1618338650.

In all cases, this means `NativeND2Reader` is deprecated and consumers should be encouraged to depend directly on `ND2Reader`. Pushed [74f337c](https://github.com/ome/bioformats/pull/4030/commits/74f337c4a60e983b40c0f4fa68d1b25b04340603) to try and capture this message although open to better ways to express this.

dgault (approved):

joshmoore (approved):

PR #4031, opened 2023-06-29
Add deprecation flags for LegacyQTTools/LegacyQTReader/LegacyQTWriter

sbesson:
These classes depended on the external QuickTime for Java library which is long obsolete. This commit marks them as deprecated in preparation for removal in the upcoming major release of Bio-Formats

The removal of these classes will affect the following components:

- `components/bio-formats-plugins/src/loci/plugins/config/ConfigWindow.java`
- `components/formats-bsd/src/loci/formats/gui/DataConverter.java`
- `components/formats-bsd/src/loci/formats/in/PictReader.java`
- `components/formats-bsd/src/loci/formats/in/QTReader.java`
- `components/formats-bsd/src/loci/formats/out/QTWriter.java`

For `LegacyQTReader`, the same range of options as the one discussed in https://github.com/ome/bioformats/pull/4030 are available ranging from minimal changes to removing the reader delegation logic.

### DIFF
--- a/components/formats-bsd/src/loci/formats/gui/LegacyQTTools.java
+++ b/components/formats-bsd/src/loci/formats/gui/LegacyQTTools.java
@@ -53,7 +53,10 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Utility class for working with QuickTime for Java.
+ *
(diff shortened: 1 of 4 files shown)


melissalinkert (approved):
Same preference for handling the delegate reader in #4030.

dgault (approved):

PR #4032, opened 2023-06-29
Add deprecation flags for LuraWave/LuraWaveCode/LuraWaveService/LuraWaveServiceImpl

sbesson:
This codec depends on a third-party proprietary library with a license code. This commit marks these classes as deprecated in preparation for their removal in the upcoming major release of Bio-Formats

The removal of these classes will minimally affect the following locations:

- `build.xml`
- `components/bio-formats-plugins/src/loci/plugins/config/FlexWidgets.java`
- `components/bio-formats-plugins/src/loci/plugins/config/libraries.txt`
- `components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java`
- `components/bio-formats-plugins/src/loci/plugins/in/ImportProcess.java`
- `components/formats-api/src/loci/formats/readers.txt`
- `components/formats-bsd/matlab/bfGetReader.m`
- `components/formats-bsd/matlab/bfopen.m`
- `components/formats-bsd/src/loci/formats/tiff/TiffCompression.java`
- `components/formats-bsd/test/loci/formats/utests/CompressDecompressTest.java`
- `components/formats-gpl/src/loci/formats/in/FlexReader.java`
- `lwf-stubs` dependency (see https://github.com/ome/bioformats/pull/4032#pullrequestreview-1505634706)

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/LuraWave.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/LuraWave.java
@@ -34,7 +34,10 @@ import loci.formats.services.LuraWaveServiceImpl;
 
 /**
  * Utility methods for dealing with proprietary LuraWave licensing.
+ *
+ * @deprecated LuraWave will be removed in Bio-Formats 7.0.0
  */
+@Deprecated
 public final class LuraWave {
 
   // -- Constants --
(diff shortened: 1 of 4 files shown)


melissalinkert (approved):
As discussed separately, it's probably worth deprecating the underlying codec/service implementations in ome-codecs, but otherwise :+1:. For removal, keep in mind there is an `lwf-stubs` dependency (https://github.com/ome/ome-stubs/tree/master/lwf-stubs) that will need to be removed; that's easy to miss if doing `git grep -i lurawave` or similar.

dgault (approved):

PR #4040, opened 2023-07-06
Remove support for proprietary LuraWave license

sbesson:
See https://github.com/ome/bioformats/pull/4032 for the original deprecation PR as well as https://github.com/ome/ome-common-java/pull/81 and https://github.com/ome/ome-codecs/pull/30 for companion PRs dropping additional components

* Remove LurawaveCode, LuraWaveService and LuraWaveServiceImpl
* Remove lwf-stubs dependencies
* Remove LURAWAVE as supported enum in TiffCompression
* Remove all references to lurawave.license system property
* Remove all code dealing with Lurawave in bio-formats-plugins


dgault:
Looks like we have a few Flex files in the repo that are listed as having the LuraWave compression (https://github.com/ome/bioformats/pull/4040/files#diff-4484ee34f26bcf07288b7899e5a097d3829b8bbeb54fd4de33a274343f73f042L198)

 https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/lastFailedBuild/console

melissalinkert:
I think what's causing the test failures is the few cases where there are Lurawave-compressed files (that are set to `test=false` in our repo) mixed in with uncompressed files that are being tested. The scan for other files in the same plate checks for compression type (among other things), which causes a problem now that the `TiffCompression` enum has been changed.

Just ignoring anything with a compression type not in the `TiffCompression` enum appears to cause some other differences in the file grouping though; it's possible those differences are actually correct, but I'm not sure yet.

dgault (approved):
With the Flex PR merged all builds and tests have remained green. All the changes here look good and I cant find any additional LuraWave references. Looks good to merge.

PR #4044, opened 2023-07-07
ARM64 libturbojpeg linux

CGDogan:
Deterministically generated using https://github.com/CGDogan/jpegjni/blob/master/Dockerfile , Sha256 `8d5ed1780ab3552ada8981434400be2d8f6e8c4d4030d43229a2d9af8f6afee8`


melissalinkert:
Thanks, @CGDogan. While we understand this is useful, we're not comfortable accepting precompiled binaries at this point.

We intend to update the list of architectures supported by turbojpeg over the coming months; see https://github.com/glencoesoftware/libjpeg-turbo-java/issues/2 for details. We will consider adding Linux and Mac ARM64 to the list of supported architectures, which means building binaries via our existing CI.

PR #4047, opened 2023-07-12
Remove Woolz support

dgault:
This will remove all support for reading and writing Woolz. It does not attempt to deprecate first, though we could do that separately as a first step. Removing this functionality is proposed because we have limited ability to test or otherwise maintain it, and the maintenance status of the underlying library (https://github.com/ma-tech/Woolz) is unclear.


melissalinkert (approved):
Matches the changes in https://github.com/ome/bioformats/pull/4014/commits/be55f1feb726542933b0bb57ddd2f582008f93b0, see https://github.com/ome/bioformats/pull/4014

sbesson (approved):

PR #4049, opened 2023-07-14
Remove LegacyND2Reader and options

dgault:
This the follow up to https://github.com/ome/bioformats/pull/4030

- LegacyND2Reader has been removed
- ND2Reader has been removed
- NativeND2Reader has been renamed to ND2Reader and deprecation flags removed
- Options to use the LegacyReader in Bio-Formats Plugins have been removed


melissalinkert:
Diff looks fine to me; only thought is it's going to be a little difficult to look at historical changes to what used to be `NativeND2Reader`, as `git log -p ND2Reader.java` or similar won't show commits on `NativeND2Reader`. Would it be worth doing something like:

- `git rm ND2Reader.java`
- `git mv NativeND2Reader.java ND2Reader.java`

so that `git log --follow -p ND2Reader.java` can show the history of `NativeND2Reader.java`?

sbesson (changes requested):
Seconding @melissalinkert's comment, I think it would be valuable to evaluate if `ND2Reader` could inherit the history of `NativeND2Reader` from a maintenance perspective.

Two other suggestions/comments:

- should `lib/LegacyND2Reader.dll` also be removed as part of this PR?
- while reviewing https://github.com/ome/bio-formats-documentation/pull/319, I was reminded of `nativend2.chunkmap` option. Would we consider renaming the option as `nd2.chunkmap`? And if so, would it be feasible to keep `nativend2.chunkmap` as a property alias (and deprecate it for 8.0.0)?

dgault:
Closing this to open a fresh PR that retains the NativeND2Reader history for ND2Reader

PR #4051, opened 2023-07-15
fix aplreader throwing nullptrexception

CGDogan:
If you run this with `/images/a.tif`, you get a nullptr exception. However not for `/a.tif` or for `/images/img/a.tif`. So our case wasn't tested.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/APLReader.java
+++ b/components/formats-gpl/src/loci/formats/in/APLReader.java
@@ -91,12 +91,12 @@ public class APLReader extends FormatReader {
         try {
           parent = parent.getParentFile();
           parent = parent.getParentFile();
+          Location aplFile = new Location(parent, parent.getName() + ".apl");
+          return aplFile.exists();
         }
         catch (NullPointerException e) {
           return false;
         }
-        Location aplFile = new Location(parent, parent.getName() + ".apl");
-        return aplFile.exists();
       }
     }
     return false;


melissalinkert:
With 6.14.0 and the results of `bfconvert test.fake test.tif` placed in `/tmp/`:

```
$ showinf /tmp/test.tif -format APL
	at loci.formats.in.APLReader.isThisType(APLReader.java:98)
	at loci.formats.FormatReader.isThisType(FormatReader.java:1443)
	at loci.formats.tools.ImageInfo.configureReaderPreInit(ImageInfo.java:446)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1038)
```

I could not duplicate this exception with just `showinf /tmp/test.tif`. With this PR:

```
$ showinf /tmp/test.tif -format APL -debug
No update needed
Checking Olympus APL format [no]
Initializing reader
APLReader initializing /tmp/test.tif
loci.formats.in.APLReader.initFile(/tmp/test.tif)
Initializing /tmp/test.tif
Failure during the reader initialization

loci.formats.FormatException: .mtb file not found
	at loci.formats.in.APLReader.initFile(APLReader.java:240)
	at loci.formats.ReaderWrapper.setId(ReaderWrapper.java:660)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1043)
```

I'd be fine with including this in builds, but @CGDogan could you please clarify the exact command or code you used to produce the `NullPointerException` above?

CGDogan:
```
ca-iip    | 04:28:10.632 [main] DEBUG loci.formats.ImageReader - loci.formats.in.APLReader@37900532 raised on isThisType
ca-iip    | java.lang.NullPointerException: null
ca-iip    | 	at loci.formats.in.APLReader.isThisType(APLReader.java:98)
ca-iip    | 	at loci.formats.ImageReader.isThisType(ImageReader.java:878)
ca-iip    | 	at loci.formats.ImageReader.getReader(ImageReader.java:193)
ca-iip    | 	at org.camicroscope.BFBridge.BFIsCompatible(BFBridge.java:113)
ca-iip    | 04:28:10.632 [main] ERROR loci.formats.ImageReader -  *** One or more readers is misbehaving. See the debug output for more information. e.g.:
ca-iip    |      loci.formats.in.APLReader@37900532 -> java.lang.NullPointerException('null') ***
```

from our method:

```
reader.getReader(filePath) != null
```

CGDogan:
```
ca-iip    | 04:28:10.632 [main] DEBUG loci.formats.ImageReader - loci.formats.in.APLReader@37900532 raised on isThisType
ca-iip    | java.lang.NullPointerException: null
ca-iip    | 	at loci.formats.in.APLReader.isThisType(APLReader.java:98)
ca-iip    | 	at loci.formats.ImageReader.isThisType(ImageReader.java:878)
ca-iip    | 	at loci.formats.ImageReader.getReader(ImageReader.java:193)
ca-iip    | 	at org.camicroscope.BFBridge.BFIsCompatible(BFBridge.java:113)
ca-iip    | 04:28:10.632 [main] ERROR loci.formats.ImageReader -  *** One or more readers is misbehaving. See the debug output for more information. e.g.:
ca-iip    |      loci.formats.in.APLReader@37900532 -> java.lang.NullPointerException('null') ***
```

from our method:

```
reader.getReader(filePath) != null
```

sbesson:
@CGDogan could you share a link to publicly available source code which would demonstrate the sequence of API calls ?

CGDogan:
There's nothing but: `(new ImageReader()).getReader(filePath)`

melissalinkert:
Thanks, @CGDogan. I can reproduce something like the log in https://github.com/ome/bioformats/pull/4051#issuecomment-1641394791 using a simple `new ImageReader().getReader("/tmp/test.tif")`.

To be clear though, that log (and what I can reproduce with `new ImageReader().getReader("/tmp/test.tif")`) is an exception being caught and logged, not an exception being thrown that prevents `getReader(String)` from being used. The catch/log logic is in https://github.com/ome/bioformats/blob/develop/components/formats-api/src/loci/formats/ImageReader.java#L880, called from https://github.com/ome/bioformats/blob/develop/components/formats-api/src/loci/formats/ImageReader.java#L193.

CGDogan:
That's correct, I'd like to keep logs clean

CGDogan:
Would you rather like:

```
        try {
          parent = parent.getParentFile();
          parent = parent.getParentFile();
        }
        catch (NullPointerException e) {
          return false;
        }
+       if (parent == null) return false;
        Location aplFile = new Location(parent, parent.getName() + ".apl");
        return aplFile.exists();
      }
```

instead?

I think you're concerned with preserving its current behaviors as much as possible.

melissalinkert:
> I think you're concerned with preserving its current behaviors as much as possible.

No, we're just trying to understand the full context of what was going wrong and how to test that this pull request fixes it, since the original pull request description has very few specifics. I don't have a problem with including this change in builds as-is, but defer to @dgault.

dgault:
In this instance i think the existing behaviour is correct and having the exception caught is preferable. The exception is also only be logged at DEBUG level which seems sensible. If the main concern is with keeping the logs clean then setting the logging level to a higher level using DebugTools (https://bio-formats.readthedocs.io/en/v7.0.0/developers/logging.html#initialization) would be an easier solution. Closing this for now as the existing code is behaving as expected.

PR #4052, opened 2023-07-16
Allow finding potential readers when the file extension is not known

CGDogan:
I think this is useful, please let me know if you have any suggestions?


melissalinkert:
Could you please explain a bit the intended use case for this method? Any actual file reading would require a name with an appropriate extension for the underlying format; initializing from an unnamed stream is not possible.

CGDogan:
I have now discovered that IFormatreader.isthistype is public so for rare usecases such as file recovery this can be implemented by the user

PR #4053, opened 2023-07-17
Fixes a NPE in OperettaReader.

NicoKiaru:
The next plane checked in the initializer may be null, in which case the plane needs to be iterated until the end of the series is found or a non-null plane is found.

See https://forum.image.sc/t/null-pointer-exception-in-perkin-elmer-operetta-dataset-with-bio-formats-6-14/83784.

That's hopefully a better fix than https://github.com/ome/bioformats/pull/3924

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -523,6 +523,13 @@ public class OperettaReader extends FormatReader {
           }
           // Check the next plane if possible
           planeIndex++;
+
+          // The next plane may be null, in which case we need to move to the next non-null one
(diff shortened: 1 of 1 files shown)


dgault (approved):
With the PR included there have been no regressions in the repo tests for existing files. Testing with the sample file provided in the original PR (available at scratch/inbox/gh-4053). 

Without this PR, attempting to read the file resulted in an exception as below:
```
	at loci.formats.in.OperettaReader.initFile(OperettaReader.java:489)
```
With this PR included the file opened without exception, displaying blank planes as expected.Overall the PR looks good to be included in the BF 7.0.0 release. I will open a separate PR to add the associated sample file to the data repo tests.

NicoKiaru:
Thanks @dgault !

PR #4059, opened 2023-07-26
TiffSaver: produce valid 4-channel RGBA images

CGDogan:
As I asked on https://forum.image.sc/t/bfconvert-breaks-images-with-alpha-layer-wrong-interleave/83482, bfconvert, when given 4-channel images, outputs with the monochrome tag. This means that many image decoders will try to read it as monochrome and show an invalid image. In general, we should save 4-channel images like 3-channel ones, because RGBA tagged as RGB is a supported configuration in the TIFF specification (https://www.itu.int/itudoc/itu-t/com16/tiff-fx/docs/tiff6.pdf pg 77), where even the order is specified: R,G,B, then A. This will result in BioFormats producing much more compatible tiff files.

EDIT: **Rebased**, considering that the TIFF specification does not limit the additional alpha channel to RGB, but it limits its premultipliedness property only to RGB. As such, YCbCr can also have an alpha channel


melissalinkert:
For reference, the original logic was introduced in #3057. It would help to know which viewer(s) can't handle TIFFs converted without this change, so that we can better evaluate the behavior of this change across a range of use cases.

CGDogan:
My mac can’t, but the reason I’m concerned with this is that I bfconvert images to tiff so that I can view with OpenSlide, but OpenSlide then shows it as monochrome. But for comparison, I searched for “view tiff online” and opened 5 different sites, none of them could view it.

CGDogan:
My mac can’t, but the reason I’m concerned with this is that I bfconvert images to tiff so that I can view with OpenSlide, but OpenSlide then shows it as monochrome. But for comparison, I searched for “view tiff online” and opened 5 different sites, none of them could view it. Here's the file: https://github.com/ome/bioformats/files/11995909/out5.tiff.zip

melissalinkert (changes requested):
Running `bfconvert` with an RGBA input file and `test.tiff` as the output file name does result in a 4-channel RGB TIFF which can be displayed with `showinf`, ImageMagick's `display`, and Windows Photos. However, `display` shows this warning:

```display-im6.q16: Sum of Photometric type-related color channels and ExtraSamples doesn't match SamplesPerPixel. Defining non-color channels as ExtraSamples.. `TIFFReadDirectory' @ warning/tiff.c/TIFFWarnings/912.```

The `ExtraSamples` tag should be set in this case, with a single value (likely 0). See pages 31-32 of the TIFF specification.

**Comment by CGDogan** [comment]
It should work now?

File: components/formats-bsd/src/loci/formats/tiff/TiffSaver.java [hunk 1]
```diff
@@ -1021,13 +1021,22 @@ private void makeValidIFD(IFD ifd, int pixelType, int nChannels) {
...
         pi = PhotoInterp.Y_CB_CR;
+        if (nChannels == 4) {
+          int[] extraSamplesArray = {2};
+          ifd.putIFDValue(IFD.EXTRA_SAMPLES, extraSamplesArray);
+        }
       }
       else {
         pi = PhotoInterp.RGB;
+        if (nChannels == 4) {
+          int[] extraSamplesArray = {1};
```
melissalinkert: I don't think it's safe to assume here that the 4th channel is always what the TIFF specification refers to as "associated alpha data", which is what the `1` here indicates. That may be true in many cases, but I'm uncomfortable with this assumption for *every* 4-channel RGB input. As noted in the previous review, I think setting the `ExtraSamples` entry to 0 ("unspecified data" per the TIFF specification) is the safest option.

components/formats-bsd/src/loci/formats/tiff/TiffSaver.java
```diff
@@ -1021,13 +1021,22 @@ private void makeValidIFD(IFD ifd, int pixelType, int nChannels) {
...
       pi = PhotoInterp.RGB_PALETTE;
     }
-    else if (nChannels == 3) {
+    else if (nChannels == 3 || nChannels == 4) {
+      // TODO: handle CMYK; nChannels == 4, extraSamples 0 (can be omitted)
       if (ifd.getIFDValue(IFD.COMPRESSION).equals(TiffCompression.JPEG.getCode())) {
         // see https://github.com/ome/bioformats/issues/3856
         pi = PhotoInterp.Y_CB_CR;
+        if (nChannels == 4) {
+          int[] extraSamplesArray = {2};
```
melissalinkert: As in line 1037, I am unconvinced that we can safely assume that the 4th channel is "unassociated alpha data". I would prefer to see this value changed to 0 unless there is a strong argument to the contrary.

melissalinkert (changes requested):

melissalinkert (approved):
This looks fine to me now. If I convert a 4-channel RGBA PNG to TIFF with bfconvert, the resulting TIFF has the `PhotometricInterpretation`, `BitsPerSample`, and `ExtraSamples` values that are expected. Opening the file with ImageMagick's `display` command shows a correct image with no warnings.

Converting the same file with `bfconvert -separate` continues to produce 4 separate IFDs, none of which have an `ExtraSamples` tag. The `PhotometricInterpretation` in this case is 1 (`BLACK_IS_ZERO`) and `BitsPerSample` has one entry, all as expected.

PR #4070, opened 2023-08-07
Gatan DM3: be more defensive in parsing acquisition mode

### LINKED ISSUE #3938: Gatan: StringIndexOutOfBoundsException when parsing Mode
Issue was raised via InsightBug-32569, a StringIndexOutOfBoundsException occurs during the parsing of the Mode metadata value. Adding a check for `token.indexOf(' ') != -1 `or simply using substring to length would be potential fixes. Having a suitable sample file would be ideal to decide how to handle it best but currently there is not one available that reproduces the issue.

The stack trace is as below:
```
Parameters: {stacktrace=java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1967)
	at loci.formats.in.GatanReader.initFile(GatanReader.java:351)
	at loci.formats.ImageReader.setId(ImageReader.java:849)
	at ome.formats.importer.OMEROWrapper$4.setId(OMEROWrapper.java:167)
	at loci.formats.ChannelFiller.setId(ChannelFiller.java:234)
	at loci.formats.ChannelSeparator.setId(ChannelSeparator.java:293)
	at loci.formats.Memoizer.setId(Memoizer.java:690)
```

melissalinkert:
Fixes #3938.

To test, use the artificial file in `curated/gatan/samples/mode-metadata`. Without this PR, `showinf` should result in an exception matching what is in #3938. With this PR, the file should initialize without error.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GatanReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GatanReader.java
@@ -347,8 +347,15 @@ public class GatanReader extends FormatReader {
       for (String token : scopeInfo) {
         token = token.trim();
         if (token.startsWith("Mode")) {
-          token = token.substring(token.indexOf(' ')).trim();
-          mode = token.substring(0, token.indexOf(' ')).trim();
+          if (token.indexOf(' ') > 0) {
+            token = token.substring(token.indexOf(' ')).trim();
+          }
+          if (token.indexOf(' ') > 0) {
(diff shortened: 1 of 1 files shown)


sbesson (approved):

dgault (approved):
Tested using the new sample file, without the PR the file fails to open with an `StringIndexOutOfBoundsException`, with the PR included the file now opens with exception.

PR #4071, opened 2023-08-07
IVision: improve version and data type checks in isThisType

### LINKED ISSUE #3582: IVision - Use stricter format recognition
Reported first via forum thread https://forum.image.sc/t/imported-txt-file/39874/4
Sample file reproducing the issue is available at `inbox/imagesc-39874/`

The IVision reader does not require a suffix (and we have one dataset in the repo which does not have a suffix). As such the reader is relying on reading a version from the start of the file. As the text file provided started with a version like String (`4.000`) the reader has treated it as an IVision file. The format recognition of this reader will need to be tightened up.

melissalinkert:
Fixes #3582.

Without this PR, `showinf` on the file in `inbox/imagesc-39874/` will incorrectly detect the file type as `Ivision`, and successfully initialize with dimensions that are obviously very wrong. With this PR, the same test should throw `UnknownFormatException`, since the file is not actually something Bio-Formats should read.

Tests of existing Ivision files should continue to pass with this change.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/IvisionReader.java
+++ b/components/formats-gpl/src/loci/formats/in/IvisionReader.java
@@ -94,7 +94,12 @@ public class IvisionReader extends FormatReader {
     String version = stream.readString(3);
     try {
       Double.parseDouble(version);
-      return version.indexOf('.') != -1 && version.indexOf('-') == -1;
+      boolean validVersion = version.indexOf('.') != -1 && version.indexOf('-') == -1;
+      boolean validPatch = Character.isAlphabetic(stream.read());
+      stream.skipBytes(1);
+      int dataType = stream.read();
+      boolean validType = dataType >= 0 && dataType <= 8;
+      return validVersion && validPatch && validType;
     }
     catch (NumberFormatException e) { }
     return false;


dgault (approved):
Testing with the invalid sample file. Without this PR the file is wrongly detected as ivision:

```
Checking file format [IVision]
Initializing reader
IvisionReader initializing /imagesc-39874/Results.txt
Populating metadata
Looking for XML metadata
Populating core metadata
Populating MetadataStore
Unknown Binning value 'nullxnull' will be stored as "Other"
Initialization took 0.073s

Reading core metadata
filename =  /imagesc-39874/Results.txt
Series count = 1
Series #0 :
	Image count = 12336
	RGB = false (1) 
	Interleaved = true
	Indexed = false (false color)
	Width = 808464432
	Height = 808464432
	SizeZ = 12336
	SizeT = 1
	SizeC = 1
	Tile size = 808464432 x 0
	Thumbnail size = 1 x 1
	Endianness = motorola (big)
	Pixel type = int8
	Valid bits per pixel = 8
	Metadata complete = false
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0
	Plane #6166 <=> Z 6166, C 0, T 0
	Plane #6167 <=> Z 6167, C 0, T 0
	Plane #6168 <=> Z 6168, C 0, T 0
	Plane #6169 <=> Z 6169, C 0, T 0
	Plane #6170 <=> Z 6170, C 0, T 0
	Plane #12335 <=> Z 12335, C 0, T 0

```

With this PR included an `UnknownFormatException` is thrown as would be expected. The PR has not had any negative impact on any existing datasets in our repo tests. Looks good to merge.

PR #4072, opened 2023-08-07
Ventana: fix tile arithmetic for smallest resolutions

### LINKED ISSUE #3927: Ventana BIF: grid of zero pixel values at lowest resolution levels
This issue was raised on 2 separate forum threads:
https://forum.image.sc/t/ventana-bif-images-imported-to-omero-web-do-not-stitch-together/75002/4
https://forum.image.sc/t/bif-opened-with-bioformats-contains-black-lines-on-the-highest-overview-levels/75221

Sample files from the first thread have been tested with BF 6.11.1 and will be copied to inbox shortly. With those datasets the lowest resolution level shows a grid of zero pixel values across the image.

Screenshot 2023-01-04 at 15 26 35

The second thread reports that it the issue can also be seen in other resolution levels (usually 0-4) but is file dependant.

melissalinkert:
Fixes #3927.

Any of the .bif files in `curated/ventana` should show the issue, but the 3 files for which the problem was reported are in `curated/ventana/gh-3927`. Without this PR, `showinf -noflat` on the last 2 resolutions in the first series should show a grid of 0-value pixels. With this PR, the same command should show a normal-looking image that matches the larger resolutions in the pyramid.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/VentanaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/VentanaReader.java
@@ -230,8 +230,8 @@ public class VentanaReader extends BaseTiffReader {
     int outputRowLen = w * tilePixel;
 
     int scale = getScale(getCoreIndex());
-    int thisTileWidth = tileWidth / scale;
-    int thisTileHeight = tileHeight / scale;
+    int thisTileWidth = scaleCoordinate(tileWidth, getCoreIndex());
+    int thisTileHeight = scaleCoordinate(tileHeight, getCoreIndex());
 
     byte[] subResTile = null;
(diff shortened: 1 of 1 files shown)


dgault (approved):
Tested with `curated/ventana/gh-3927` and saw the same results for other files in the curated Ventana datasets.

Without the PR attempting to open the lowest resolution resulted in the grid like pattern appearing on the images.

Screenshot 2023-10-04 at 13 18 14

With the PR included the image now opens and displays as expected without the same pattern. The image dimensions and metadata remain unaffected.

Screenshot 2023-10-04 at 13 22 13

PR #4074, opened 2023-08-07
Gel: check IFD count in isThisType

### LINKED ISSUE #4034: Gel Tiff: IndexOutOfBoundsException
This issue was raised on forum thread https://forum.image.sc/t/error-converting-to-ome-ngff/82872 when attempting to convert a dataset for the Bio Image Archive using bioformats2raw.

A sample dataset was provided which is available on scratch/inbox/imagesc-82872

Using the provided sample I was able to reproduce the issue using Bio-Formats 6.13.0 through showinf.
The stack trace is below:

```
java.lang.IndexOutOfBoundsException: Index: 2, Size: 2
	at java.util.ArrayList.rangeCheck(ArrayList.java:659)
	at java.util.ArrayList.get(ArrayList.java:435)
	at [loci.formats.in](http://loci.formats.in/).MinimalTiffReader.openBytes(MinimalTiffReader.java:295)
	at [loci.formats.in](http://loci.formats.in/).GelReader.openBytes(GelReader.java:132)
```

melissalinkert:
Fixes #4034.

Actual gel images have a single plane, and a maximum of 2 IFDs (one image, one extra metadata).

The `imagesc-82872/SG...` file from #4034 is not actually a gel, but was incorrectly identified as it has some of the same tags that `GelReader` expects. With this PR, the file should be picked up by `TiffDelegateReader`, and have 4 single-plane series all of which are readable. 

There is some additional metadata in the file which doesn't appear to correspond to an existing reader. I would guess that this data should be represented as a single series with 2 resolutions and 2 channels, but that would require a new reader.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/GelReader.java
+++ b/components/formats-gpl/src/loci/formats/in/GelReader.java
@@ -90,7 +90,11 @@ public class GelReader extends BaseTiffReader {
   public boolean isThisType(RandomAccessInputStream stream) throws IOException {
     TiffParser parser = new TiffParser(stream);
     parser.setDoCaching(false);
-    IFD ifd = parser.getFirstIFD();
+    long[] offsets = parser.getIFDOffsets();
+    if (offsets.length == 0 || offsets.length > 2) {
+      return false;
+    }
+    IFD ifd = parser.getIFD(offsets[0]);
     if (ifd == null) return false;
     return ifd.containsKey(MD_FILETAG);
   }


dgault (approved):
Tested using the sample file from `curated/tiff/imagesc-82872`. Without this PR attempting to open the file results in the GelReader being used and an exception thrown:
```
java.lang.IndexOutOfBoundsException: Index: 3, Size: 2
	at java.util.ArrayList.rangeCheck(ArrayList.java:659)
	at java.util.ArrayList.get(ArrayList.java:435)
	at loci.formats.in.MinimalTiffReader.openBytes(MinimalTiffReader.java:295)
	at loci.formats.in.GelReader.openBytes(GelReader.java:132)

```

With this PR included the TiffDelegateReader is now used to read the image and it can be opened and each of the four planes does display.

PR #4076, opened 2023-08-08
VSI: fix uncompressed BGR data

### LINKED ISSUE #3819: VSI: Inverted colours (RGB instead of BGR)
Issue was raised on forum thread https://forum.image.sc/t/qupath-0-3-2-inverts-colors-of-vsi-file/66609/3
A sample file was provided that reproduces the issue with Bio-Formats 6.9.1

The image appears to be detected by the reader as an RGB image, in the reality the channels should be BGR. A quick inspection of the metadata does not show any obvious values that could be used to detect this. Further inspection of the file and the format spec would be needed to understand better the channel order.

melissalinkert:
Fixes #3819.

To test, use `curated/cellsens/gh-3819/` and see ground truth screenshot in https://github.com/openmicroscopy/data_repo_config/pull/583. Without this PR, `showinf -noflat` on any resolution should show an image that is more red than the screenshot. With this PR, the same test should show an image that matches the screenshot.

This will require a minor release, as it affects memo files.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -1241,7 +1241,7 @@ public class CellSensReader extends FormatReader {
 
     etsFile.skipBytes(4 * 10 - color.length); // background color
     int componentOrder = etsFile.readInt();
-    bgr[s] = componentOrder == 1;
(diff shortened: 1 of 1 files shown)


dgault (approved):
Tested with `curated/cellsens/gh-3819/`. Without the PR displaying any of the resolutions produces an image which displays with more red than the ground truth image

With the PR included the same image displays matching the color of the ground truth image in the config PR.

PR #4080, opened 2023-08-09
Exporter plugin: remove any extra Channel elements

### LINKED ISSUE #3539: ICS Writer: export fails when cropping dimensions
This bug was reported on forum thread https://forum.image.sc/t/bio-formats-export-as-ics1-2-fails-when-cropping-more-than-one-dimension/35685 and a sample file provided in QA-29206 (the sample file is also available in from the FIJI samples `File > Open Samples > Mitosis (26MB, 5D stack)`).

To reproduce the issue the file can be imported using specify range. In the reported example the range used was 1C, 5Z, 10T. When exporting the file to TIFF there is no exception and the file exports correctly. When exporting to ICS the below exception is seen:
```
java.lang.ArrayIndexOutOfBoundsException: 50
	at loci.formats.out.ICSWriter.saveBytes(ICSWriter.java:137)
	at loci.formats.FormatWriter.saveBytes(FormatWriter.java:132)
	at loci.plugins.out.Exporter.run(Exporter.java:847)
```

melissalinkert:
Fixes #3539.

Without this PR, the problem is reproduceable by following the steps in the issue; in particular, the image needs to be opened with `Specify range for each series` checked and at least one channel omitted from the range.

With this PR, the same test should successfully write an .ics file. The converted file should also be readable, and have the correct (smaller) dimensions.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -446,6 +446,15 @@ public class Exporter {
             store.setPixelsSizeC(new PositiveInteger(channels*imp.getNChannels()), 0);
             store.setPixelsSizeT(new PositiveInteger(imp.getNFrames()), 0);
 
+            try {
+              // if a subset of the data was opened, the number of Channels
+              // in the OME-XML may not match the ImagePlus channel count
+              // channel count mismatches can cause problems when actually writing pixels
+              service.removeChannels(service.getOMEMetadata(store), 0, imp.getNChannels());
+            }
+            catch (ServiceException e) {
+            }
+
             if (store.getImageID(0) == null) {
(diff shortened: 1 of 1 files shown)


dgault (approved):
Tested with the original file from the issue. Without this PR and reducing the range of the C dimension. When attempting to save the file it results in the same error as the issue:

```
java.lang.ArrayIndexOutOfBoundsException: 50
	at loci.formats.out.ICSWriter.saveBytes(ICSWriter.java:137)
	at loci.formats.FormatWriter.saveBytes(FormatWriter.java:132)
	at loci.plugins.out.Exporter.run(Exporter.java:847)
```

With the PR included and retesting with the same file, the image is now exported without exception. Opening the newly created image also opens and displays without exception, the dimensions are correct and the pixels all appear correct.

PR #4088, opened 2023-08-24
CZI: unify channel color parsing

### LINKED ISSUE #4087: ZeissCZIReader get lut error
I have a `czi` file , read color of one channel , the color is `#FF0000FF`, in the source code `get16Lut` method is used int to parse this number.

melissalinkert:
Fixes #4087.

`get*LookupTable` now parses colors using the same logic that is used when setting `Channel.Color`. This allows for both RGB and ARGB colors; if specified, the alpha is ignored. If anything other than '#' is removed from the original color string, the change will be logged at `DEBUG`.


dgault (approved):
The new logic here looks good and makes sense. All builds and tests have remained green and been unaffected by the PR. Looks good to merge.

PR #4090, opened 2023-08-28
Throw error when cannot load library

CGDogan:
openBytes returned 0 bytes so there was a silent failure

### DIFF
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
@@ -105,8 +105,10 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
     logger = Logger.getLogger(NATIVE_LIB_CLASS);
     logger.setLevel(Level.SEVERE);
     if (!libraryLoaded) {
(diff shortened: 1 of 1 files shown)


dgault:
Thanks @CGDogan for getting the PR opened. Have you encountered issues with the library loading? If so what system and architecture were you using when seeing the problem?

CGDogan:
On ARM64 where TurboJPEG isn't currently added (https://github.com/ome/bioformats/pull/4044 https://github.com/glencoesoftware/libjpeg-turbo-java/issues/2) BioFormats was reading OS-1.ndpi downloaded from OpenSlide test files server. The most zoomed out resolutions it could read, the most zoomed in resolutions it could not read so openBytes returned "0 bytes" but not exception. (Another difference is that the zoomed out resolutions are uninterleaved whereas zoomed in ones are)

To reproduce this, you may recompile BioFormats on any system after deleting this folder https://github.com/ome/bioformats/tree/develop/components/forks/turbojpeg/lib

dgault (approved):
Throwing an exception in this scenario is certainly preferable to setting libraryLoaded to true and returning anyway. Ideally a checked exception would be perferable but that would require a minor release, so is something that can be looked at longer term. For the upcoming patch release this is an improvement on the existing behaviour.

PR #4094, opened 2023-09-10
File header update

sbesson:
This PR reviews the file headers using the output `mvn license:update-file-header`.

Most of the changes are simple whitespace differences or description updates. The only change of signficance is 39dedca38b01055a4e57689c6f5a711da4162b35 which I believe corrects an oversight of 8b12b69faa1b01ba84b2c872975da20bb37c2d5a where the reader was moved under the BSD-2 licensed components for inclusion without updating the header.


dgault (approved):
All the changes look good, the SlideBook correction does look to be a genuine oversight when moving the reader to BSD.

Issue #3643, opened 2020-12-08
bfconvert use wrong tile size during pyramid creation


iwbh15:
Under specific condition bfconvert uses a wrong tile size for one of the planes during pyramid creation.
('Specific' condition means that the problem does not occur always.
The exact conditions could not be found.
It seems as if the image size has an influence.)

The problem can be reproduced e.g. by converting an 8bit single channel tif file with the command:
	bfconvert -tilex 256 -tiley 256 -pyramid-resolutions 8 -pyramid-scale 2 ..\test_1chn_8bit.tif ..\bf_out_1chn_8bit_jpeg_pyr.ome.tif
or
	bfconvert -compression JPEG -tilex 256 -tiley 256 -pyramid-resolutions 8 -pyramid-scale 2 ..\test_1chn_8bit.tif ..\bf_out_1chn_8bit_jpeg_pyr.ome.tif

The test image was created with the following ImageJ macro:
	newImage("Untitled", "8-bit black", 6400, 6400, 1);
	path= "C:/tmp/test_1chn_8bit.tif";
	saveAs("Tiff", path);

This problem seems to occur when pyramid tiles are created smaller than the defined tilex/tiley size.
It occurs with other image sizes (<> 6400x6400) but not all.
	
	
The windows console output of bfconvert is:
 
C:\tmp\bftools>bfconvert -no-upgrade -compression JPEG -tilex 256 -tiley 256 -pyramid-resolutions 8 -pyramid-scale 2 ..\test_1chn_8bit.tif ..\bf_out_1chn_8bit_jpeg_pyr.ome.tif
..\test_1chn_8bit.tif
TiffDelegateReader initializing ..\test_1chn_8bit.tif
Reading IFDs
Populating metadata
Checking comment style
Populating OME metadata
[Tagged Image File Format] -> ..\bf_out_1chn_8bit_jpeg_pyr.ome.tif [OME-TIFF]
Tile size = 256 x 256
        Converted 1/1 planes (100%)
Tile size = 256 x 256
        Converted 1/1 planes (100%)
Tile size = 256 x 256
        Converted 1/1 planes (100%)
Tile size = 256 x 256
        Converted 1/1 planes (100%)
Tile size = 256 x 256
        Converted 1/1 planes (100%)
Tile size = 200 x 163
        Converted 1/1 planes (100%)
Tile size = 100 x 100
        Converted 1/1 planes (100%)
Tile size = 50 x 50
        Converted 1/1 planes (100%)
[done]
[...]

Issue #3630, opened 2020-11-05
bfconvert: crop, downsampling and multi-resolution issues


sbesson:
See https://forum.image.sc/t/cropping-large-pyramidal-tiffs-with-bfconvert-crashes/44857

The issue can be reproduced using a pyramidal fake image and trying to convert and crop it

```
printf "sixeX=40000\nsizeY=100000\nresolutions=10\n" test.fake.ini
bfconvert -crop 30000,0,10000,10000 -tilex 512 -tiley 512 -noflat test.fake out.tif
```

With Bio-Formats 6.5.1, this will result in a `FormatException` when checking for the tile size as soon as a resolution of size smaller than the crop size is reached. One of the underlying issue is that the cropped region is never adjusted at each resolution scale. This is not that a simple scenario which involves computing  the downsampling scale as well as handling  rounding errors.
The current workaround is to first crop the largest resolution "only", then convert it into a pyramidal image by recomputing the resolution levels.

A variation of this issue had been reported in https://forum.image.sc/t/bfconvert-throwing-loci-formats-formatexception-for-a-czi-file/37478 where the number of requested resolutions is larger than the number of resolutions in the image but the downsampling is inconsistent.

As a general rule, it feels like like for multi-resolution images, we should only support one of the two cases:
- convert the original image exactly as it is including all pyramidal level
- take the largest resolution  of each multi-resolution image and perform computations (cropping, downscaling...) from this resolution level
The  latter is also the approach taken by the Glencoe [bioformats2raw](https://github.com/glencoesoftware/bioformats2raw) pipeline.

Two additional issues specifically related to cropping is the modification of any positional metadata or region of interests as well as the absence of recording of the crop region in the metadata.

For upcoming release, we should probably define a set of recommended or disallowed scenarios, and minimally fail fast with an informative message for some combination of options and image metadata.

rleigh-codelibre:
This is a situation where I think the best solution is at the model level, as detailed below.

You don't want to change any of the raw tile data.  Particularly when lossy compression is used.  In order to correctly preserve the data, you could treat the crop region as a bounding box, and select all tiles which intersect the region for retention.  All non-intersecting tiles can be discarded.  This will handle overlapping tiles at all resolution levels.  In order to correctly set the crop region for rendering and the image size, you will need a separate bit of metadata in Pixels (or attached to Pixels via a reference) which describes the bounding box, for example an x+y offset and an x+y size.  It would also generalise to z and t, and potentially c as well.

[...]

PR #4097, opened 2023-09-12
DICOM writer: throw an exception if the provided tiles don't match the expected tile size

melissalinkert:
Without this PR, viewing the smallest resolution from the output of:

```
bfconvert HandEcompressed_Scan1.qptiff test.dcm -tilex 2048 -tiley 2048 -noflat
```

shows an almost all black image; only the top 16 rows of pixels are present. Viewing in OMERO, QuPath, or `showinf -noflat -resolution 4 test_0_0.dcm` will show the problem. `HandEcompressed_Scan1.qptiff` is from https://downloads.openmicroscopy.org/images/Vectra-QPTIFF/perkinelmer/PKI_scans/.

The source of the problem is a combination of things:
- the chosen 2048x2048 tile size is smaller than all resolutions except resolution 4 (the last one)
- for tile sizes larger than a particular resolution, `bfconvert` reverts to the reader's optimal tile size
- the reader's optimal size is 1920x16 for resolution 4 (all others have an optimal size of 512x512)
- `DicomWriter` expects full tiles matching the chosen tile size to be supplied to `saveBytes`, and will pad any smaller tiles to the correct dimensions. So in this case, the first 1920x16 tile supplied gets padded to 2048x2048, and the image is considered completely written.

The proposed solution here is to throw an exception in `saveBytes` if a tile is provided that is not on an expected tile boundary (`x` and `y` coordinates not a multiple of tile size) or is the wrong width or height without being bottom or right edge tile. I think that's most compatible with #3992, minimizes impact by not changing `bfconvert`, and catches cases where the provided tile might not be the correct size when `bfconvert` is not being used.

With this PR, the conversion command above should throw `FormatException`. Changing the tile size parameters to `-tilex 1024 -tiley 1024` or similar should work, and each resolution should be readable.

Updating `bfconvert` to use the full image instead of the reader's optimal size in cases such as resolution 4 would also potentially work, but I'm not convinced that's appropriate in general.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/DicomWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/DicomWriter.java
@@ -188,6 +188,10 @@ public class DicomWriter extends FormatWriter {
   {
     checkParams(no, buf, x, y, w, h);
 
+    int resolutionIndex = getIndex(series, resolution);
+    int thisTileWidth = tileWidth[resolutionIndex];
+    int thisTileHeight = tileHeight[resolutionIndex];
+
     MetadataRetrieve r = getMetadataRetrieve();
     if ((!(r instanceof IPyramidStore) ||
       ((IPyramidStore) r).getResolutionCount(series) == 1) &&
@@ -195,11 +199,17 @@ public class DicomWriter extends FormatWriter {
     {
       throw new FormatException("DicomWriter does not allow tiles for non-pyramid images");
     }
+    else if (x % thisTileWidth != 0 || y % thisTileHeight != 0 ||
+      (w != thisTileWidth && x + w != getSizeX()) ||
+      (h != thisTileHeight && y + h != getSizeY()))
+    {
+      throw new FormatException("Tile too small, expected " + thisTileWidth + "x" + thisTileHeight +
+        ". Setting the tile size to " + getSizeX() + "x" + getSizeY() + " or smaller may work.");
+    }
     checkPixelCount(false);
 
     boolean first = x == 0 && y == 0;
     boolean last = x + w == getSizeX() && y + h == getSizeY();
-    int resolutionIndex = getIndex(series, resolution);
 
     // the compression type isn't supplied to the writer until
     // after setId is called, so metadata that indicates or
@@ -264,8 +274,6 @@ public class DicomWriter extends FormatWriter {
 
     byte[] paddedBuf = null;
 
-    int thisTileWidth = tileWidth[resolutionIndex];
-    int thisTileHeight = tileHeight[resolutionIndex];
     int thisTilePixels = thisTileWidth * thisTileHeight;
 
     // pad the last row and column of tiles to match specified tile size


sbesson (commented):
Confirmed that with the proposed change, the command that was previously writing incorrect planes for the smallest resolution now fails with an exception indicating the reduce the tile size. 


Within the context of this writer, the new error message is definitely a improvements as it allows the user to select a different tile size. Only downside is that this call happens in `saveBytes` so  all the previous resolutions have been successfully generated and the suggested tile size might still be not suitable for the next resolution levels if any.


For comparison, I tried to perform the same tiled writing using OME-TIFF as the format output and the conversion fails with a different error message on the last resolution.

```
(java11) sbesson@Sebastiens-MacBook-Pro-3 bioformats % ./tools/bfconvert ~/Downloads/HandEuncompressed_Scan1.qptiff test.ome.tiff -bigtiff -tilex 2048 -tiley 2048 -noflat -compression zlib
Output file test.ome.tiff exists.
Do you want to overwrite it? ([y]/n)
y
/Users/sbesson/Downloads/HandEuncompressed_Scan1.qptiff
VectraReader initializing /Users/sbesson/Downloads/HandEuncompressed_Scan1.qptiff
Reading IFDs
Populating metadata
Populating OME metadata
[PerkinElmer Vectra/QPTIFF] -> test.ome.tiff [OME-TIFF]
Tile size = 2048 x 2048
	Series 0: converted 1/1 planes (100%)
Tile size = 2048 x 2048
	Series 0: converted 1/1 planes (100%)
Tile size = 2048 x 2048
	Series 0: converted 1/1 planes (100%)
Tile size = 2048 x 2048
	Series 0: converted 1/1 planes (100%)
Tile size = 1920 x 16
Exception in thread "main" java.lang.IndexOutOfBoundsException: Range [0, 0 + 30720) out of bounds for length 5760
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckFromIndexSize(Preconditions.java:82)
	at java.base/jdk.internal.util.Preconditions.checkFromIndexSize(Preconditions.java:343)
	at java.base/java.util.Objects.checkFromIndexSize(Objects.java:424)
	at java.base/java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:155)
	at java.base/java.io.DataOutputStream.write(DataOutputStream.java:107)
	at loci.formats.tiff.TiffSaver.writeImage(TiffSaver.java:354)
	at loci.formats.tiff.TiffSaver.writeImage(TiffSaver.java:278)
	at loci.formats.out.TiffWriter.saveBytes(TiffWriter.java:282)
	at loci.formats.out.OMETiffWriter.saveBytes(OMETiffWriter.java:219)
	at loci.formats.out.PyramidOMETiffWriter.saveBytes(PyramidOMETiffWriter.java:89)
	at loci.formats.out.OMETiffWriter.saveBytes(OMETiffWriter.java:209)
	at loci.formats.tools.ImageConverter.convertTilePlane(ImageConverter.java:966)
	at loci.formats.tools.ImageConverter.convertPlane(ImageConverter.java:833)
	at loci.formats.tools.ImageConverter.testConvert(ImageConverter.java:766)
	at loci.formats.tools.ImageConverter.main(ImageConverter.java:1199)
```

The error above makes me wonder whether we have a more fundamental issue with the behavior of `ImageConverter` when converting resolutions smaller than the supplied tile size.

melissalinkert:
> The error above makes me wonder whether we have a more fundamental issue with the behavior of `ImageConverter` when converting resolutions smaller than the supplied tile size.

Is that something you would like to see addressed in this PR, or would a follow-up issue/PR be reasonable? #3643 and #3630 may be related as well.

sbesson:
> Is that something you would like to see addressed in this PR, or would a follow-up issue/PR be reasonable?

I avoided requesting changes in my review as I am not sure of the best approach and the amount of work required to deal with this tile size adjustment issue.

As mentioned in my previous review, throwing an error message definitely feels like a step-wise improvement compared to silently writing corrupted data. In the context of a patch release, I think it's fair to limit the scope of the changes to this write and schedule a separate discussion regarding changes to the converter logic.
@dgault any thoughts?

dgault (approved):
Agreed that this has highlighted a more fundamental issue within the ImageConverter (or conversions in general). This PR seems like a sensible first step for DICOM writing though and from the perspective of including it in a patch release then this seems a definite improvement on the existing behaviour. I would be happy including this in the upcoming release and handling the wider use case as a follow up for a future release.

sbesson (approved):

PR #4098, opened 2023-09-12
TIFF: throw exception on 64-bit int data

### LINKED ISSUE #4089: Loading 64-bit int TIFFs
Dear all,
As a Fiji user I'm a long time user and appreciator of BioFormats.
However... there appears to be quite dangerous behaviour when loading 64-bit int or uint tiffs, as I posted here:

https://forum.image.sc/t/saving-and-loading-64-bit-int-tiffs/85523

@cgohlke proposes a very reasonable solution, and I'd be interested in your opinion

melissalinkert:
Fixes #4089. Test file in `inbox/gh-4089`, generated as described in https://forum.image.sc/t/saving-and-loading-64-bit-int-tiffs/85523/2.

Without this PR, `showinf im-i8.tif` should display an image with `double` type. `tiffinfo` or `tiffdump` should confirm that there 64 bits per sample, but with a `SampleFormat` of 2 indicating signed int data.

With this PR, `showinf -debug im-i8.tif` should throw `FormatException` indicating that the pixel type is unsupported.

We have no real way of supporting this data in Bio-Formats right now, since the OME model does not allow for 64-bit ints. https://trac.openmicroscopy.org/ome/ticket/11860 and https://trello.com/c/FxSUHfGp/273-add-more-model-enum-values are the closest relevant discussions to updating supported pixel types.

### DIFF
--- a/components/formats-bsd/src/loci/formats/tiff/IFD.java
+++ b/components/formats-bsd/src/loci/formats/tiff/IFD.java
@@ -615,6 +615,9 @@ public class IFD extends HashMap<Integer, Object> {
       case 24:
         return FormatTools.FLOAT;
       case 64:
+        if (bitFormat != 3) {
+          throw new FormatException("64-bit int data not supported");
+        }
         return FormatTools.DOUBLE;
       case 32:
         if (bitFormat == 3) return FormatTools.FLOAT;


sbesson (approved):
Given the set of pixel types supported by the OME model, this is the most reasonable behavior for the moment.

dgault (approved):
Agreed that this is currently the best approach for dealing with this. Testing with the sample file in, inbox/gh-4089 and with this PR a FormatException is now seen as expected.

PR #4111, opened 2023-10-26
OIR: channel order and missing pixel block fixes

melissalinkert:
Backported from a private PR.

Changes through line 742 were meant to address a single missing pixel block issue. Unfortunately, the data for that case is private and I have not successfully made an artificial example that demonstrates the problem.

The remainder of the changes are meant to address an issue with channel ordering. The problem occurs when the primary channel elements (whose `order` attributes were respected) are deemed invalid, and the secondary channel elements are used instead. The changes here update the secondary channel element handling in `parseChannel` to respect the `order` attribute, so that both sets of channel handling logic match. An artificial example that demonstrates is in `curated/olympus-oir/samples/channel-reorder`.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -687,7 +687,9 @@ public class OIRReader extends FormatReader {
     }
     // total block length could include multiple strings of XML
     int totalBlockLength = s.readInt();
+    LOGGER.debug("total block length = {}", totalBlockLength);
     long end = s.getFilePointer() + totalBlockLength - 4;
+    LOGGER.debug("end = {}", end);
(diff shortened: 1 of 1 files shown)


jburel:
@melissalinkert Will the testing data we currently have be enough for review or do we need a new set of data?

sbesson (approved):
Retested with the artificial sample generated to simulate the channel order scenario as well as the private samples.

With this PR included, the channel returned match the screenshots from the Olympus Viewer plugin. No other regression was raised as part of the nightly CI builds and the new sample file has been configured appropriately.

dgault (approved):
Tested the channel ordering using the new artificial sample file. With the PR included the channel ordering for the new sample file matches that from the description of the config PR. All other builds and tests have remained green with this PR included.

PR #4112, opened 2023-10-30
Imaris HDF: populate channel names

melissalinkert:
See https://forum.image.sc/t/imaris-files-and-qupath/88061 and file copied to `inbox/imagesc-88061`.

As noted in the commit message, this appears to have been an oversight dating to the early days of this reader, as the channel names were already parsed correctly.

This does require configuration changes; a corresponding PR is forthcoming.


sbesson (approved):
Successfully tested across the set of Imaris `ims` files available in the curated QA repository.
Channel names are now properly set when populating the OME metadata using `showinf -omexml` and the values match what is displayed in ImarisViewer 10.1.0 as the source of truth

Issue #4103, opened 2023-10-10
CZI: Label and macro image incorrectly labelled


dgault:
Issue was raised on forum thread https://forum.image.sc/t/label-macro-image-switch-after-cropping/87104/7

The user originally saved the file and was able to see the label and macro correctly labelled. However after cropping the image and resaving it using Zen Blue 3.5 (Lite, version 3.5.093.00000) , the label and macro have now inverted.

A sample file was provided at https://zenodo.org/record/8424082 that can reproduce the issue using Bio-Formats 7.0.0

PR #4114, opened 2023-11-10
Add test for image names when resolutions are not flattened

melissalinkert:
In most cases, the `Image` names stored in a `MetadataStore` will not depend on `hasFlattenedResolutions()`. For some pyramid formats (e.g. SVS, CZI), the `Image` names are noticeably different.

This updates the config file generation to add an `Alternate_Name` representing the unflattened `Image` name, if the unflattened `Image` name is different from the flattened `Image` name for a particular series/resolution. The new `testUnflattenedImageNames` checks this value, and errors if the name provided by an unflattened reader does not match either configured name.

I would expect this to cause test failures for SVS, CZI, and potentially others. I am happy for this to be excluded once at least test run has completed (so we know the scope of config changes required).

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -42,8 +42,9 @@ import loci.formats.FormatException;
 import loci.formats.FormatTools;
 import loci.formats.IFormatReader;
 import loci.formats.ImageReader;
-import loci.formats.in.DynamicMetadataOptions;
+import loci.formats.MetadataTools;
 import loci.formats.ReaderWrapper;
+import loci.formats.in.DynamicMetadataOptions;
 import loci.formats.meta.IMetadata;
 
 import ome.xml.model.primitives.PositiveInteger;
@@ -108,6 +109,7 @@ public class Configuration {
   private static final String EXCITATION_WAVELENGTH_UNIT = "ExcitationWavelengthUnit_";
   private static final String DETECTOR = "Detector_";
   private static final String NAME = "Name";
+  private static final String UNFLATTENED_NAME = "Alternate_Name";
   private static final String DESCRIPTION = "Description";
   private static final String SERIES_COUNT = "series_count";
   private static final String RESOLUTION_COUNT = "resolution_count";
@@ -418,6 +420,10 @@ public class Configuration {
     return currentTable.get(NAME);
   }
 
+  public String getUnflattenedImageName() {
+    return currentTable.get(UNFLATTENED_NAME);
+  }
+
   public boolean hasImageDescription() {
     return currentTable.containsKey(DESCRIPTION);
   }
@@ -551,6 +557,7 @@ public class Configuration {
     IFormatReader unflattenedReader = reader;
     if (seriesCount > 1) {
       unflattenedReader = new ImageReader();
+      unflattenedReader.setMetadataStore(MetadataTools.createOMEXMLMetadata());
       unflattenedReader.setFlattenedResolutions(false);
       unflattenedReader.setMetadataOptions(new DynamicMetadataOptions());
       try {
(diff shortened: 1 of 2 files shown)


melissalinkert:
Unsurprisingly, this caused a number of build failures:

- `samples`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/34972/console
- `hamamatsu`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35019/console
- `afi`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35023/console
- `svs`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35025/console
- `zeiss-czi`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/34996/console
- `ventana`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35043/console
- `ome-tiff`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35003/console
- `imaris`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35059/console
- `leica-scn`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35067/console
- `cellsens`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35063/console
- `pyramid`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35089/console
- `jpeg2000`: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/35098/console

Excluding for the moment, can re-include once a configuration PR is ready.

sbesson:
On the positive side, it looks like tests have been failing if and only if the format is multi-resolution which was the expectation.

components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -108,6 +109,7 @@ public class Configuration {
   private static final String EXCITATION_WAVELENGTH_UNIT = "ExcitationWavelengthUnit_";
   private static final String DETECTOR = "Detector_";
   private static final String NAME = "Name";
+  private static final String UNFLATTENED_NAME = "Alternate_Name";
```
sbesson: Re-reading this PR, I wonder whether we should use a more explicit key for this configuration. `Unflattened_Name` is probably the closest to the current logic but it is a bit of a mouthful. Arguably we could also reverse the logic and store `Flattened_Name` for multi-resolution samples.

melissalinkert: Attribute name updated in 7ed65ecdbc. I'm not entirely sure whether reversing the logic is more or less confusing overall; doing that would have a bigger impact on configuration changes, as `Name` would change and `Flattened_Name` would be added (vs just adding `Unflattened_Name`).

sbesson:
As discussed today with @melissalinkert and @joshmoore, I agree the latest state of this PR puts us in a nice compromise between:
- allowing the non-regression test to check image names when using the sub-resolution API - which is a prerequisite to add coverage for issues of the like of #4103 
- and making minimal changes to the existing test-suite infrastructure and the set of existing configuration samples.

Next step here is to update the configuration files for all pyramidal file formats and re-include this PR in the nightly CI builds

sbesson (approved):
Together with the associated configuration changes for all multi-resolution formats, the non-regression tests has been passing over several execution of the nightly CI builds

dgault (approved):
The approach taken here seems sensible to me and the new test looks good. With all of the config updates the tests have been passing without any issues.

PR #4115, opened 2023-11-15
Revert RunTimeException when JPEGTurbo cannot be loaded

### LINKED ISSUE #4109: #4090 breaks support for Apple Silicon for formats that don't need TurboJPEG
Hi,

https://github.com/ome/bioformats/pull/4090 seems to break the use of
```java
```
on Apple Silicon - even for formats that otherwise worked without problem (e.g. CMU-1.svs, LuCa tiff images).

The issue seems to be that the `try/catch` block doesn't handle the `RuntimeException` thrown whenever it is not possible to load TurboJPEG:
https://github.com/ome/bioformats/blob/e484e1faefd6c7590fa5dbb56e56f3cc360d9139/components/formats-api/src/loci/formats/ImageReader.java#L134-L139

The workaround is to pass a `ClassList` that has been already been checked to exclude classes that can't be instantiated, e.g. using
```java
        private static ClassList<IFormatReader> getDefaultClassList() {
		ClassList<IFormatReader> classes = new ClassList<>(IFormatReader.class);
		for (var cls : ImageReader.getDefaultReaderClasses().getClasses()) {
			try {
				var instance = cls.getConstructor().newInstance();
				if (instance != null)
					classes.addClass(cls);
			} catch (Throwable t) {
				logger.debug("Cannot instantiate reader class " + cls.getName() + ": " + t.getMessage());
			}
		}
		return classes;
	}

...

ImageReader reader = new ImageReader(getDefaultClassList());
```

Adding code like this should be enough for us to be able to include 7.0.1 in QuPath - but since Bio-Formats is the now only dependency we have without Apple Silicon support, it would be even better if https://github.com/ome/bioformats/issues/3756 and https://github.com/ome/bioformats/issues/3858 could be addressed :) The QuPath Apple Silicon builds are now faster and more reliable on M1/M2 Macs (due in part to some OpenCV changes), so will be recommended with the caveat that most .czi images won't work.

dgault:
Reverts the changes from  https://github.com/ome/bioformats/pull/4090
Fixes https://github.com/ome/bioformats/issues/4109

To test you should attempt to instance ImageReader reader = new ImageReader(); while using Apple Silicon. This should throw a RunTimeException even if the desired reader does not require JPEGTurbo. 

I opted to revert the change rather than catch the exception in ImageReader as this would negate the original reason for adding the exception.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/TileJPEGReader.java
@@ -147,6 +147,7 @@ public class TileJPEGReader extends FormatReader {
       service = null;
       throw new IOException("Could not initialize JPEG service", se);
     }
+    if (!service.isLibraryLoaded()) throw new IOException("JPEG service failed to load Turbo JPEG library");
   }
 
 }
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboService.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboService.java
@@ -58,5 +58,7 @@ public interface JPEGTurboService extends Service {
   byte[] getTile(int xTile, int yTile) throws IOException;
 
   void close() throws IOException;
+  
+  boolean isLibraryLoaded();
 
 }
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
@@ -105,8 +105,7 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
     logger = Logger.getLogger(NATIVE_LIB_CLASS);
     logger.setLevel(Level.SEVERE);
     if (!libraryLoaded) {
-      NativeLibraryUtil.loadNativeLibrary(TJ.class, "turbojpeg");
-      libraryLoaded = true;
+      libraryLoaded = NativeLibraryUtil.loadNativeLibrary(TJ.class, "turbojpeg");
     }
   }
 
@@ -379,6 +378,11 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
     xTiles = 0;
     yTiles = 0;
     header = null;
+  }
+
+  @Override
+  public boolean isLibraryLoaded() {
+    return libraryLoaded;
   }
 
   // -- Helper methods --
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
@@ -105,10 +105,8 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
     logger = Logger.getLogger(NATIVE_LIB_CLASS);
     logger.setLevel(Level.SEVERE);
     if (!libraryLoaded) {
-      libraryLoaded = NativeLibraryUtil.loadNativeLibrary(TJ.class, "turbojpeg");
-      if (!libraryLoaded) {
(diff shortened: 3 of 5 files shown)


melissalinkert:
I'm completely fine with this approach for 7.1.0, but for a future release would it be worth something like:

- set `libraryLoaded` as in #4090, to accurately represent the library loading state
- add a getter for `libraryLoaded`, so that individual readers can query it
- update `initFile` in `NDPIReader` (and other readers that use `JPEGTurboService`) to call the new getter and throw an exception if the library was not loaded

That would be in the spirit of #4090 (fail before `openBytes` if data can't be read), but limited to the case where unreadable files are actually being initialized.

dgault:
Yeah, that seems like a good approach that would meet the demands of the initial PR but without failing unnecessarily. I will look at getting it implemented as part of this PR.

joshmoore:
Testing on my arm64 machine:

* Without this PR I see the TurboJPEG runtime exception. :+1:
* With this PR, I do not. :+1:

Testing required running:

```
bioformats/artifacts $zip -d bioformats_package.jar META-INF/lib/osx_64/libturbojpeg.dylib
```
after each build, even though I had deleted turbojpeg:

```
(mvn) /opt/own-bf/bio-formats-build/bioformats/tools $git status
On branch JPegTurbo-RevertRTE
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   ../ant/toplevel.xml
	deleted:    ../components/forks/turbojpeg/LICENSE.txt
	deleted:    ../components/forks/turbojpeg/build.properties
	deleted:    ../components/forks/turbojpeg/build.xml
	deleted:    ../components/forks/turbojpeg/lib/linux_32/libturbojpeg.so
	deleted:    ../components/forks/turbojpeg/lib/linux_64/libturbojpeg.so
	deleted:    ../components/forks/turbojpeg/lib/osx_32/libturbojpeg.dylib
	deleted:    ../components/forks/turbojpeg/lib/osx_64/libturbojpeg.dylib
	deleted:    ../components/forks/turbojpeg/lib/osx_ppc/libturbojpeg.dylib
	deleted:    ../components/forks/turbojpeg/lib/windows_32/turbojpeg.dll
	deleted:    ../components/forks/turbojpeg/lib/windows_64/turbojpeg.dll
	deleted:    ../components/forks/turbojpeg/pom.xml
	deleted:    ../components/forks/turbojpeg/src/TJExample.java
	deleted:    ../components/forks/turbojpeg/src/TJUnitTest.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJ.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJCompressor.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJCustomFilter.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJDecompressor.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJLoader.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJScalingFactor.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJTransform.java
	deleted:    ../components/forks/turbojpeg/src/org/libjpegturbo/turbojpeg/TJTransformer.java
	modified:   ../pom.xml
```

🤷🏽

melissalinkert (approved):
Code changes look reasonable to me, with the caveat that I am not set up to actually test functionality.

joshmoore (approved):

PR #4117, opened 2023-11-21
Work-around for breaking change in CellSens 4.1

### LINKED ISSUE #4116: Change in CellSens 4.1 causing exception in VSI format reader.
I am getting an exception reading a set of Olympus VSI files produced by CellSens software 4.1 with BioFormats 7.0.1.

This problem looks similar to, but is **not** the same as, the issue discussed in Issue #3859 (VSI: Exception loading files from VS200 scanner with software version 3.4.1), which was fixed in PR #3925 (Olympus .vsi: only read pixels from frame_\*.ets files).  This is a new problem with a different cause; the version of BioFormats I am using includes the aforementioned fix.

I have two sample VSI files, each with their associated frame_\*.ets files, produced from CellSens version 4.1.  Each of these VSI files exhibit the same symptoms. We are presently looking to get permission to share these two images.  When executing "ImageInfo -nopix -no-upgrade", this exception is thrown:

```
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
        at java.util.ArrayList.rangeCheck(ArrayList.java:659)
        at java.util.ArrayList.get(ArrayList.java:435)
        at loci.formats.in.CellSensReader.parseETSFile(CellSensReader.java:1272)
```

The offending code line is:
    `HashMap<String, Integer> dimOrder = pyramids.get(s).dimensionOrdering;`

In this exception, the 'pyramids' array has one element, and 's' is 1.

I have done a deep dive into the logic to find the cause of this problem and here's what I've discovered.  I'll start by recapping the relevant logic in the CellSensReader.java.  The '`pyramids`' array is built while parsing through the metadata in the .vsi file; a new element is added to the array each time an EXTERNAL_FILE_PROPERTIES tag is encountered.  Apparently each instance of this tag in the metadata introduces a new metadata parameter block that is associated with one individual external frame_\*.ets file, and as well, the block provides all the information needed for describing the image stored in the corresponding frame_\*.ets file.  The logic appears to work on the assumption that there will exist exactly one frame_\*.ets file for each EXTERNAL_FILE_PROPERTIES tag, and vice versa.

It is worth noting here that the method used by the code to correlate each EXTERNAL_FILE_PROPERTIES tag with the appropriate frame_*.ets file is solely by sequence.  The sequence of the frame_*.ets files is established via sorting, using the folder names as the primary sort key, and the frame_*.ets file name as the secondary key.  The frame_*.ets files are then matched one by one with the EXTERNAL_FILE_PROPERTIES tags in the sequence in which they are encountered in the .vsi metadata.

In the cases of my two failure images, the EXTERNAL_FILE_PROPERTIES tag appears in the metadata exactly **one** time, yet each of the failed vsi images has **four** associated frame_*.ets files!  I have done extensive examination of the metadata in the vsi file to verify this is true.

The proximate cause of the exception is that there are more frame_\*.ets files than there are EXTERNAL_FILE_PROPERTIES tags.  Apparently some change in the latest version (V4.1) of scanner software causes, or at least allows, this situation to occur.  In addition to the exception being thrown, this mismatch between the numbers of EXTERNAL_FILE_PROPERTIES tags and frame_\*.ets files brings up another problem, which is, there is now ambiguity as to how to properly associate each EXTERNAL_FILE_PROPERTIES metadata block with the correct frame_\*.ets file.  In the cases of my two sample images, I have examined the image parameters (specifically pixel width and height) within the one EXTERNAL_FILE_PROPERTIES tag's present metadata block, and it's clear that this block of metadata is NOT associated with the **first** frame_\*.ets file (based on the present sort order) but with the **second**.  Thus, for any work-around to be successful it would not suffice to simply ignore any frame_\*.ets files which have no associated EXTERNAL_FILE_PROPERTIES tag; it would first be necessary to establish a reliable way to correctly associate each EXTERNAL_FILE_PROPERTIES metadata block with the appropriate frame_\*.ets file.

With all the above investigation completed, I am creating a PR for a proposed code change for a work-around that will:
  1. Attempt to correctly extract the most amount of pixel data from the .vsi and frame_\*.ets using the information available.
  2. Will not alter the present behavior of the code for any vsi whose count of frame_\*.ets files matches the number of EXTERNAL_FILE_PROPERTIES metadata blocks.  

The logic of the work-around does this:
  1. When there are more frame_\*.ets files than EXTERNAL_FILE_PROPERTIES metadata blocks, it attempts to associate each frame_\*.ets file with the appropriate metadata block by using the image's pixel width and height from the metadata block.
  2. For performing this association, the pixel width and height of the highest resolution image in each frame_\*.ets file can be calculated to within one tile boundary.  This is done by examining the tile coordinates of all tiles in the file to find the maxX and maxY coordinate at the highest resolution (in a similar manner to logic already in the program).
  3. If the pixel width and height from the metadata block matches to within one tile boundary of the calculated width and height of the frame_\*.ets, they will be associated.
  4. If a Frame_\*.ets file does not have an associated metadata block, it will be handled as usual but without any associated metadata.  At minimum the pixel width and height of the image series within the frame_\*.ets files are required.  These are calculated and filled in using the pixel width and height calculated from the tile coordinates during the matching process.

ed-scanlon:
Missing metadata from .vsi file is causing exception when parsing .ets files.  Added logic to dynamically associate .efs files with .vsi file metadata.

I have submitted a detailed write-up in Issue #4116.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -771,8 +771,15 @@ public class CellSensReader extends FormatReader {
       if (s < files.size() - 1) {
         setCoreIndex(index);
         String ff = files.get(s);
+        /**
+         * If there are fewer metadata 'pyramids' defined in the vsi than there
+         * are frame_*.ets files, we are missing the metadata associated with
+         * one or more frame_*.ets files.  In this case we need to dynamically
+         * match the metadata we do have with the appropriate frame_*.ets file.
+        */
+        boolean missingMetadata = pyramids.size() < (files.size() - 1);
         try (RandomAccessInputStream stream = new RandomAccessInputStream(ff)) {
-            parseETSFile(stream, ff, s);
+            parseETSFile(stream, ff, s, missingMetadata);
         }
 
         ms.littleEndian = compressionType.get(index) == RAW;
@@ -796,10 +803,6 @@ public class CellSensReader extends FormatReader {
           }
         }
         index += ms.resolutionCount;
-
-        if (s < pyramids.size()) {
-          ms.seriesMetadata = pyramids.get(s).originalMetadata;
-        }
 
         setCoreIndex(0);
         ms.dimensionOrder = "XYCZT";
@@ -1188,7 +1191,8 @@ public class CellSensReader extends FormatReader {
     return buf;
   }
 
-  private void parseETSFile(RandomAccessInputStream etsFile, String file, int s)
+  private void parseETSFile(RandomAccessInputStream etsFile, String file, int s,
+                            boolean missingMetadata)
     throws FormatException, IOException
   {
     fileMap.put(core.size() - 1, file);
@@ -1286,7 +1290,40 @@ public class CellSensReader extends FormatReader {
     int[] maxC = new int[maxResolution];
     int[] maxT = new int[maxResolution];
 
-    HashMap<String, Integer> dimOrder = pyramids.get(s).dimensionOrdering;
+    HashMap<String, Integer> dimOrder = new HashMap<String, Integer>();
+    Pyramid pyramid = null;
+    /**
+    * If there is metadata missing from the vsi file, we need to see if we can match up
+    * this frame_*.ets file with its correct metadata block if it exists.  To do that we
+    * search the metadata blocks (ie 'pyramids') for the one whose width and height are
+    * within the correct range for this frame_*.ets file.
+    */
+    if (missingMetadata) {
+      int maxXAtRes0 = 0;
+      int maxYAtRes0 = 0;
+      for (TileCoordinate t : tmpTiles) {
+        if (!usePyramid || t.coordinate[t.coordinate.length - 1] == 0) {
+          maxXAtRes0 = Math.max(maxXAtRes0, t.coordinate[0]);
+          maxYAtRes0 = Math.max(maxYAtRes0, t.coordinate[1]);
+        }
+      }
+      int maxPixelWidth  = (maxXAtRes0 + 1) * tileX.get(tileX.size()-1);
+      int maxPixelHeight = (maxYAtRes0 + 1) * tileY.get(tileY.size()-1);
+      for (Pyramid p : pyramids) {
+        if (  (p.width  <= maxPixelWidth ) && (p.width  >= maxPixelWidth  - tileX.get(tileX.size()-1))
+           && (p.height <= maxPixelHeight) && (p.height >= maxPixelHeight - tileY.get(tileY.size()-1)) ) {
+            pyramid = p;
+            break;
+        }
+      }
+    }
+    else {
+      pyramid = pyramids.get(s);
+    }
+
+    if (pyramid != null) {
+      dimOrder = pyramid.dimensionOrdering;
+    }
 
     for (TileCoordinate t : tmpTiles) {
       int resolution = usePyramid ? t.coordinate[t.coordinate.length - 1] : 0;
@@ -1381,12 +1418,23 @@ public class CellSensReader extends FormatReader {
         maxC[resolution] = t.coordinate[cIndex];
       }
     }
-
-    if (pyramids.get(s).width != null) {
-      ms.sizeX = pyramids.get(s).width;
+    /**
+     * If this ets file has an associated metadata block, grab the correct width and height,
+     * and link it to the full metadata block.
+     */
+    if (pyramid != null) {
+      ms.sizeX = pyramid.width;
+      ms.sizeY = pyramid.height;
+      ms.seriesMetadata = pyramid.originalMetadata;
     }
-    if (pyramids.get(s).height != null) {
-      ms.sizeY = pyramids.get(s).height;
+    else {
+    /**
+     * Otherwise, compute the closest approximation, since the real info couldn't be found.
+     */
+      ms.sizeX = (maxX[0] + 1) * tileX.get(tileX.size()-1);
+      ms.sizeY = (maxY[0] + 1) * tileY.get(tileY.size()-1);
+      ms.seriesMetadata = new Hashtable<String, Object>();
+      ms.seriesMetadata.put("Metadata", "Not Found"); // leave a trail that the metadata was missing for this ets.
     }
     ms.sizeZ = maxZ[0] + 1;
     if (maxC[0] > 0) {
--- a/components/formats-gpl/src/loci/formats/in/CellSensReader.java
+++ b/components/formats-gpl/src/loci/formats/in/CellSensReader.java
@@ -772,14 +772,28 @@ public class CellSensReader extends FormatReader {
         setCoreIndex(index);
         String ff = files.get(s);
         /**
-         * If there are fewer metadata 'pyramids' defined in the vsi than there
-         * are frame_*.ets files, we are missing the metadata associated with
-         * one or more frame_*.ets files.  In this case we need to dynamically
-         * match the metadata we do have with the appropriate frame_*.ets file.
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thank you for the detailed write-up and proposed fix, @ed-scanlon. I don't see anything immediately concerning in code review, so am including in nightly builds to assess impact on our existing data.

I do have a private sample dataset (not included in automated tests) that exhibited the problem in #4116. This PR does allow that dataset to be read, but note that while this PR attempts to read the extra frame*.ets files, Olympus/Evident's OlyVIA software appears to ignore these files.

It would be good to understand what the extra frame*.ets files represent - are these legitimate acquisitions that were incorrectly omitted from the .vsi file? Or are they invalid/duplicate/deleted acquisitions that were correctly removed from the .vsi and the .ets file was just not deleted?

ed-scanlon:
Hi Melissa, thank you for reviewing my write-up, and for including the proposed work-around in the nightly builds.  I have just a few more comments to add.

Yes, from what I can tell too, Olyvia does seem to ignore the extra files - if I remove all but one of the .ets files, Olyvia will still display everything the same way as when they're all present (the trick is knowing which ones to remove and which one to keep).

My proposed work-around is based on the very limited information I have, with the main unknown being the very question you wondered too, why are these extra frame_*.ets files present?  Since I didn't know, it was hard for me to make the decision to just totally ignore them.

As for determining which .ets file is the correct one to use, picking the one with the best matching width and height may work well, but I have no way to know how well that method would hold up over many samples.  A more ideal situation would be if we could find something within each EXTERNAL_FILE_PROPERTIES metadata block that would explicitly indicate exactly which .ets file it is referencing.  (Unfortunately I have no more information about the format and contents of the .vsi file than I have ascertained by reading the bioformats code.)  If anyone could find such information then it may be possible to simply construct the actual path to each .ets file from the associated metadata block.  In that case it would no longer be necessary to perform directory scan for finding .ets files, and we'd have an actual "fix" instead of just a "work-around" for this problem.

ed-scanlon:
Please see comment regarding the second commit here. https://github.com/ome/bioformats/issues/4116

melissalinkert (changes requested):
I definitely agree that ignoring orphaned .ets files (but keeping them on the used file list) is the better approach; that's more consistent with how the vendor's software treats this data.

One of the two test datasets I have works as expected with this PR, but the other does not. The non-working data should have one .ets file that is removed by the block in line 788, but that doesn't appear to happen. The initialization is successful, but viewing the last several series (e.g. `showinf -noflat -series 24`) shows an exception similar to the original issue:

```
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index 24 out of bounds for length 24
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at loci.formats.in.CellSensReader.getCurrentPyramid(CellSensReader.java:1070)
	at loci.formats.in.CellSensReader.openBytes(CellSensReader.java:545)
	at loci.formats.FormatReader.openBytes(FormatReader.java:922)
	at loci.formats.ImageReader.openBytes(ImageReader.java:449)
	at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:334)
	at loci.formats.tools.ImageInfo.readPixels(ImageInfo.java:829)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1063)
```

The working dataset has series with distinct XY sizes; no two have the same X or Y dimension. The non-working dataset has several small series, some with identical XY sizes, and some that vary by less than 10 pixels in either dimension. I suspect the orphaned .ets file has the same XY size as one of the valid files, and so is incorrectly determined to have a matching metadata block.

I unfortunately can't share either dataset, but can certainly re-test with additional changes or capture the output of `showinf -debug` if additional debug logging is added.

melissalinkert:
@ed-scanlon : thanks again for all of your work here. In order to get this pull request merged, the comments in https://github.com/ome/bioformats/pull/4117#pullrequestreview-1806459537 will need to be addressed, and we'll need a signed [Contributor License Agreement](https://ome-contributing.readthedocs.io/en/latest/cla.html).

Could you please let us know if you're still working on this? If so, we'll leave this pull request open, but if not we'll plan to close it and leave #4116 open to be fixed at a later date.

ed-scanlon:
@melissalinkert : Thank you very much for message.  Yes, this change is important to us and I would like to see it through.  I have put it into my schedule to address the comments and post a new commit within the next couple of days.  Thank you for your detailed analysis of the problem, I understand, and I have what I believe is a suitable resolution in mind.

components/formats-gpl/src/loci/formats/in/CellSensReader.java
```diff
@@ -771,8 +771,29 @@ else if (zCount > 0) {
...
+        boolean hasOrphanEtsFiles = pyramids.size() < (files.size() - 1);
         try (RandomAccessInputStream stream = new RandomAccessInputStream(ff)) {
-            parseETSFile(stream, ff, s);
+            boolean validFrameFile = parseETSFile(stream, ff, s, hasOrphanEtsFiles);
+            /*
+             * If this frame file is an orphan, "undo" the work that was done for
+             * it and change its status to being an "extra" file.
+             */
+             if (!validFrameFile) {
+                core.remove(core.size()-1);
```
melissalinkert, line 789:
```suggestion
                fileMap.remove(core.size() - 1);
                core.remove(core.size()-1);
```

melissalinkert (changes requested):
Only final change I think is required is suggested inline - the `fileMap` variable just needs to be updated when an invalid file is found. That only causes an issue for things that call `openThumbBytes`; the easiest way to see this is by running `showinf -thumbs` which should throw something like:

```
Exception in thread "main" java.lang.IllegalArgumentException: Invalid series: 64
	at loci.formats.FormatReader.setCoreIndex(FormatReader.java:1430)
	at loci.formats.in.CellSensReader.openThumbBytes(CellSensReader.java:514)
	at loci.formats.ImageReader.openThumbBytes(ImageReader.java:479)
	at loci.formats.ReaderWrapper.openThumbBytes(ReaderWrapper.java:360)
	at loci.formats.gui.BufferedImageReader.openThumbImage(BufferedImageReader.java:100)
	at loci.formats.tools.ImageInfo.readPixels(ImageInfo.java:828)
	at loci.formats.tools.ImageInfo.testRead(ImageInfo.java:1063)
```

With the suggested change, the same command should no longer throw `IllegalArgumentException` (though it may fail with `OutOfMemoryError`, which is far outside the scope of this pull request).

Otherwise, this looks good and now behaves as I'd expect with the test data I used previously. I'm fine with this being merged once the CLA is in place, and the above change is made and included in at least one passing nightly build.

components/formats-gpl/src/loci/formats/in/CellSensReader.java
ed-scanlon: Good catch Melissa!  I tried to be thorough but that one escaped me.  I place the filemap.remove statement within the parseETSFile() function; it seemed more appropriate since the file was added to the filemap from within that function.  Please let me know if you encounter any other issues.  Thank you.

melissalinkert (approved):

ed-scanlon:
Thank you Melissa, we appreciate your help and collaboration on this issue. -Ed

dgault (approved):
The changes here look good and the logic used makes sense from my side. Tested using the new artificial sample file and with the PR included the file was able to be opened and displayed as expected without the previous exception.

Repo tests have continued to pass with this PR and the new sample file included. CLA has also been received so the PR looks good to merge.

PR #4125, opened 2023-12-06
Remove duplicate org.json:json dependency declaration

sbesson:
Implements https://github.com/ome/bioformats/pull/4123#pullrequestreview-1764908440

Since org.json:json is now a dependency of formats-bsd, this removes the duplicate declaration in formats-gpl so that the version is only managed in one place.

An alternative would be to declare `json.version` as a top-level property


dgault (approved):
Makes sense to me to simply remove it rather than having it declared it both components

PR #4126, opened 2023-12-08
Bump dependency versions

dgault:
Bumping the various dependency versions ahead of release.
I have also bumped the Xalan dependency to match that in ome-common: https://github.com/ome/ome-common-java/pull/84


dgault:
Missed a step promoting new ome-codecs release from staging to release. Should be fixed now

PR #4130, opened 2023-12-13
matlab/bfGetPlane.m: avoid use of makeDataArray2D when Octave (ome/bio-formats-octave-docker#29)

### LINKED ISSUE #29: test_bfopen fails with
While trying to create a minimal example file to reproduce https://forum.image.sc/t/bio-formats-octave-error-on-bfsave/27504, I realized that the current Travis build is failing - see https://travis-ci.org/ome/bio-formats-octave-docker/builds/545619855.

```
Populating metadata
Reading series #1
    .error: octave_base_value::reshape (): wrong type argument 'octave_java'
error: called from
    bfGetPlane at line 99 column 7
    bfopen at line 150 column 13
    test_bfopen at line 6 column 6
    travis_test at line 9 column 1
```

There are two  issues associated with this error:

- I would have expected the `docker run` command to fail with a non-zero status code on a failing script rather than a false negative
- `octave_base_value::reshape (): wrong type argument 'octave_java'` is thrown when calling `bfGetPlane` for all recent builds using GNU Octave 4.2 / Bio-Formats 6. The only related report I could find was https://savannah.gnu.org/bugs/?44881. @carandraug: do you have any immediate thoughts on this before we start investigating.

carandraug:
This fixes ome/bio-formats-octave-docker#29

Octave does not autobox multi-dimensional arrays (Matlab does) so `makeDataArray2D` can't be used in Octave. this commit makes use of `makeDataArray` to get a 1D array and then reshapes it.  This commit partially reverts 3aa468e9 (see ome/bioformats#3301) for Octave only. the data copy that that commit was meant to prevent is back but is limited to Octave.

For reference, the issue of Octave not autoboxing ND Java arrays is [Octave bug #58745](https://savannah.gnu.org/bugs/?func=detailitem&item_id=58745) (the issue description references a function `java2mat` that in theory should make this conversion but that function never worked for me and newer Octave functions have marked it somewhat internal by renaming it `__java2mat__`).

### DIFF
--- a/components/formats-bsd/matlab/bfGetPlane.m
+++ b/components/formats-bsd/matlab/bfGetPlane.m
@@ -82,9 +82,20 @@ plane = r.openBytes(...
     ip.Results.iPlane - 1, ip.Results.x - 1, ip.Results.y - 1, ...
     ip.Results.width, ip.Results.height);
 
-% Convert byte array to MATLAB image
-I = javaMethod('makeDataArray2D', 'loci.common.DataTools', plane, ...
-    bpp, fp, little, ip.Results.height);
+% Convert byte array into the appropriate primitive type array which
+% Octave and Matlab then autobox into their own array type.
+if is_octave()
+    % Octave will not autobox multi-dimensional arrays, so use
+    % makeDataArray (returns 1D vector) instead of makeDataArray2D
+    % See https://github.com/ome/bio-formats-octave-docker/issues/29
+    I = javaMethod('makeDataArray', 'loci.common.DataTools', plane, ...
+        bpp, fp, little);
+    I = reshape(I, [ip.Results.width ip.Results.height]).';
+else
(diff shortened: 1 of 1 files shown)


dgault:
Thank you @carandraug for opening the PR and the continued support for the the project. Since your previous contributions we have introduced a Contributor License Agreement for the project, would you be able to sign and return the form following the instructions on https://ome-contributing.readthedocs.io/en/latest/cla.html

carandraug:
I'm not willing to that CLA at the moment. If you prefer, this patch is too trivial and I do not claim copyright for it.

joshmoore:
I've reached out separately to David P. to discuss this and his other contributions.

joshmoore:
Chatted with David today. General support still for getting this fix in (finally). To be on the safe side, @melissalinkert will open a clean room version of the PR.

As always, thanks for your contributions and support for the project, @carandraug!

PR #4131, opened 2023-12-15
VirtualImagePlus: Retain B&C settings for each slice

dgault:
This is a follow up to a previous PR https://github.com/ome/bioformats/pull/3850/files and is an attempt to address the issues raised in https://forum.image.sc/t/b-c-for-a-whole-virtual-stack-cont/57811/10

The fix here reapplies the table settings for each slice using the same approach as the ContrastAdjuster. This may be overkill and there may be a simpler solution to ensure that the settings are applied for each slice.


dgault:
Failures relate to the 16bit tests, requires further investigation

joshmoore:
Following on from https://forum.image.sc/t/bio-formats-support/101109, it's unclear if there will be capacity to move this PR forward. We've removed it from 8.0.0 for the moment. If anyone is interested in getting involved, please let us know.

melissalinkert:
Closing, as we do not currently have capacity to work on this.

PR #4141, opened 2024-01-12
CZI: fix image naming, especially for attachments

melissalinkert:
Fixes #4103. Makes use of the new image name test from #4114.

A configuration PR is forthcoming that demonstrates affected data. For datasets with changed names, compare the name and actual image with and without this PR to verify that this fixes label and macro naming. Multi-scene and plate data should also now have unflattened names that reflect the acquisition type, instead of just `Scene #1` etc.

PR #4145, opened 2024-01-23
CZI: use alternate source for well identifiers when available

melissalinkert:
Backported from a private PR. This fixes Image/Well mapping for "meander" acquisitions.

Several of our existing test plates are affected, see the forthcoming configuration PR which includes video of the affected data in ZEN. Without this PR, `showinf -nopix -omexml` on the affected plates should show that each row of wells starts at the left-most (smallest index) column and proceeds to the right-most (largest index) column. With this PR, the same command should show that  rows alternate between left-to-right and right-to-left column indexing. The order of rows is unaffected.

This is a really subtle problem and not obvious from visually inspecting the images or even a cursory look at ZEN. Even a basic stitching workflow wouldn't necessarily have picked this up. It's really only obvious when carefully scrolling through each image in ZEN, so reviewing the videos in the configuration PR is important.

Importantly, well metadata can exist in multiple places in CZI datasets, and appears to not necessarily be consistent. That might reflect the physical plate vs the actual acquisition, and can be seen by looking at the original metadata (look for names like `A1` and IDs like `1-1`).

Added to 7.2.0 for a first look, but since this is a later addition I'm OK with it being bumped to the next milestone if needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -1395,7 +1395,7 @@ public class ZeissCZIReader extends FormatReader {
         else {
           if (i < imageNames.size()) {
             String completeName = imageNames.get(i);
-            if (i < fieldNames.size()) {
+            if (i < fieldNames.size() && !fieldNames.get(i).equals(completeName)) {
               completeName += " " + fieldNames.get(i);
             }
             store.setImageName(completeName, i);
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested against a few of the sample CZI plates modified by this change. Without this change, the well-order always start from the left-most colum of each row. With this change included, the order follows a meander-style acquisition strategy. The new order is consistent with the original metadata particular the `Information|Image|S|Scene|Name` as well as the `Information|Image|S|Scene|Shape|{ColumnIndex/RowIndex}` values as well as the inspection of theses plates in ZEN while scrolling through the scenes.

The nightly CI builds have been successful with the associated configuration changes which only modify the `UnflattenedName`, `WellColumn` keys. Given that #4141 which is already integrated and scheduled for the 7.2.0 release also fixes some image names for the CZI format, I think including this change in the minor release makes sense so that all the known image naming issues are addressed at once.

dgault (approved):
Tested the sample files which have been affected in the config PR and compared to the videos showing the inspection in Zen. Based on the video comparison the updated column indexes appear to be correct.

PR #4061, opened 2023-07-30
Don't fail on non-4:4:4 JPEG

CGDogan:
The previous pull request description was hard to read, so I'm rewriting it: MCU size is 8x8 for 4:4:4. We should read relative subsampling "numbers" from SOF0 and calculate MCU size accordingly. This way, we won't get IndexOutOfBounds when we read 4:2:0 etc.Closes #3616

PR #4148, opened 2024-02-01
JPEG: reset `useLegacy` flag to its original state upon close

melissalinkert:
This problem was exposed by #4061, but likely not directly caused by it.

I would expect tests to pass again with this included.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/JPEGReader.java
+++ b/components/formats-bsd/src/loci/formats/in/JPEGReader.java
@@ -65,6 +65,15 @@ public class JPEGReader extends DelegateReader {
 
   private static final int MAX_SIZE = 8192;
 
+  /**
+   * setId may alter the "useLegacy" variable depending upon
+   * the type of JPEG encountered. Storing the state of
+   * the "useLegacy" variable prior to any internal changes
+   * allows the reader to reset "useLegacy" to its original
+   * state upon close.
+   */
+  private boolean initialUseLegacy;
+
   // -- Constructor --
 
   /** Constructs a new JPEGReader. */
@@ -83,6 +92,10 @@ public class JPEGReader extends DelegateReader {
   /* @see FormatReader#setId(String) */
   @Override
   public void setId(String id) throws FormatException, IOException {
+    // capture the current state of "useLegacy"
+    // it may be updated later in this method and should be reset upon close
+    initialUseLegacy = isLegacy();
+
     try {
       super.setId(id);
     }
@@ -176,6 +189,10 @@ public class JPEGReader extends DelegateReader {
   public void close(boolean fileOnly) throws IOException {
     super.close(fileOnly);
     Location.mapId(currentId, null);
+
+    // reset to whatever legacy setting was originally requested
+    // this erases any changes made during setId
+    setLegacy(initialUseLegacy);
   }
 
   // -- Helper reader --


sbesson (commented):
A few initial thoughts:
- if I understand the root of the issue, it would apply to any `DelegateReader` (in practice `JPEGReader` being the only one) that would be re-used to initialize different files requiring either the legacy reader or the native reader, is that correct?
- in terms of automated tests, whether the legacy or native reader is used is currently not captured in the configuration files. Is this something we would also like to serialize and test? Or would we assume as in the current case that an incorrect selection would either lead to an exception in `setId` or failures in the downstream tests?
- should the fix be implemented more generically in `DelegateReader`? Naively, I was thinking that  `DelegateReader.close` could simply reset `useLegacy`

melissalinkert:
> A few initial thoughts:
> 
> * if I understand the root of the issue, it would apply to any `DelegateReader` (in practice `JPEGReader` being the only one) that would be re-used to initialize different files requiring either the legacy reader or the native reader, is that correct?

I think the underlying problem is that `useLegacy` is actually changed in `JPEGReader.setId`, which `DelegateReader` and `TiffDelegateReader` (our only other `DelegateReader` implementation right now) don't do.

> * in terms of automated tests, whether the legacy or native reader is used is currently not captured in the configuration files. Is this something we would also like to serialize and test? Or would we assume as in the current case that an incorrect selection would either lead to an exception in `setId` or failures in the downstream tests?

I'd expect tests to fail if the wrong delegate was used.

> * should the fix be implemented more generically in `DelegateReader`? Naively, I was thinking that  `DelegateReader.close` could simply reset `useLegacy`

That's a fair point. I was thinking this was more JPEG-specific since that's the only reader that alters `useLegacy` internally, but I can do the same fix in `DelegateReader`, and we can decide which one to keep?

melissalinkert:
On second thought, I don't think a change in `DelegateReader` will actually work here. See https://github.com/ome/bioformats/compare/develop...melissalinkert:delegate-reader-reset?expand=1 for comparison.

These two lines in `JPEGReader`:

https://github.com/melissalinkert/bioformats/blob/delegate-reader-reset/components/formats-bsd/src/loci/formats/in/JPEGReader.java#L147
https://github.com/melissalinkert/bioformats/blob/delegate-reader-reset/components/formats-bsd/src/loci/formats/in/JPEGReader.java#L156

mean that `DelegateReader` resets to the updated value of `useLegacy` - not the default when a new reader is created or whatever was explicitly set by a call to `setLegacy`.

sbesson:
Thanks for the feedback @melissalinkert. Today was busy but I'll take a closer look at this change on Monday.
In the meantime, I have another naive question: would it be correct/equivalent to set `useLegacy = false` before https://github.com/ome/bioformats/blob/194a643d02ed766bf57705a3fa3f9e8d1c880cbc/components/formats-bsd/src/loci/formats/in/JPEGReader.java#L87 i.e. enforce that the first `setId` call is always made using the `DefaultJPEGReader` independently of the state of `isLegacy()`?

sbesson (approved):
From a CI perspective, the proposed change has been addressing regression in the nightly tests caused by the addition of new samples. 

As discussed, one approach to reproduce the failure is to recycle an existing `ImageReader` where the `JPEGReader` has been initialized and set `useLegacy` to true.

To perform a more direct testing of the underlying issue, I have been using the command-line `showinf` utility with the local modification on top of `v7.1.0`

```diff
diff --git a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
index e52ccfcdc7..2cd05328db 100644
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -50,6 +50,7 @@ import loci.common.xml.XMLTools;
 import loci.formats.ChannelFiller;
 import loci.formats.ChannelMerger;
 import loci.formats.ChannelSeparator;
+import loci.formats.DelegateReader;
 import loci.formats.DimensionSwapper;
 import loci.formats.FilePattern;
 import loci.formats.FileStitcher;
@@ -134,6 +135,7 @@ public class ImageInfo {
   private int xmlSpaces = 3;
   private DynamicMetadataOptions options = new DynamicMetadataOptions();
   private Byte fillColor = null;
+  private boolean legacy = false;
 
   private IFormatReader reader;
   private IFormatReader baseReader;
@@ -188,6 +190,7 @@ public class ImageInfo {
     map = null;
     cachedir = null;
     fillColor = null;
+    legacy = false;
     if (args == null) return false;
     for (int i=0; i<args.length; i++) {
       if (args[i].startsWith("-")) {
@@ -211,6 +214,7 @@ public class ImageInfo {
         else if (args[i].equals("-no-sas")) originalMetadata = false;
         else if (args[i].equals("-normalize")) normalize = true;
         else if (args[i].equals("-fast")) fastBlit = true;
+        else if (args[i].equals("-legacy")) legacy = true;
         else if (args[i].equals("-autoscale")) {
           fastBlit = true;
           autoscale = true;
@@ -439,6 +443,12 @@ public class ImageInfo {
       if (new Location(id).exists()) {
         LOGGER.info("Checking file format [{}]", ir.getFormat(id));
       }
+      IFormatReader[] r = ir.getReaders();
+      for (int i=0; i<r.length; i++) {
+        if (r[i] instanceof DelegateReader) {
+         DelegateReader delegate = (DelegateReader) r[i];
+         delegate.setLegacy(legacy);}
+      }
     }
     else {
       // verify format
```

and executed `showinf -crop 0,0,400,400` as well as `showinf -legacy -crop 0,0,400,400` on the set of JPEG samples in the curated QA repository.

All JPEG samples currently configured for testing opened successfully without the `-legacy` flag. With the `-legacy` flag, only a subset of the samples initialized with severa; of them failing with `Restart interval and markers invalid`. The current logic of reader is slightly confusing as it provides an API allowing the user to control the legacy flag but in practice, the reader initialization might modify this value internally. Given the exposition of this API, it is quite likely few people came across this issue and most usage are being redirected to either implementation depending on the JPEG.

In conclusion, happy for this change included to fix the testing regression and effectively allow a wider set of JPEG samples to be included in the the daily test suite.

dgault (approved):
Happy to get this in as a JPEG specific fix and address the testing regressions

PR #4154, opened 2024-02-13
Update GitHub actions versions and add M1/Java 21 build

melissalinkert:
First 3 commits are just cleaning up versions, which hopefully should remove the warnings about Node.js 16 actions being deprecated (e.g. https://github.com/ome/bioformats/actions/runs/7829106830).

Last commit here adds a single configuration to the Maven build, which runs on M1 with Java 21. I intentionally added this separately instead of fully expanding the matrix to `java: [8, 11, 17, 21]` and `os: [ubuntu-latest, windows-latest, macos-latest, macos-14]`, as discussed previously with @ome/formats. If we feel that running M1 + Java 21 every time is too much, we can debate other options; this is just a place to start.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -33,7 +33,7 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Retrieve version
         id: get_version
         run: |
@@ -66,7 +66,7 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Retrieve version
         id: get_version
         run: |
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK ${{ matrix.java }}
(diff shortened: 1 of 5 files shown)


.github/workflows/maven.yml
```diff
@@ -15,13 +15,16 @@ jobs:
       matrix:
         java: [8, 11, 17]
         os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - os: 'macos-14'
```
sbesson: Super nitpicky, could we swap the order of the two items so that the display name in https://github.com/ome/bioformats/actions/runs/7909123358/job/21589620169 is consistent with the other builds ?

sbesson (commented):
All Node.js 16 warnings are now gone with the actions version updates. The dependency updates and test changes look good to me.

Adding some initial coverage for JDK 21 / Apple M1  in the CI builds makes sense to me. At a glance, the addition does not look too onerous in terms of build time. We might want to watch out for the  `macos-14` usage of the builds over the next few weeks.

I could certainly see the testing matrix reducing the number of JDK 8 builds and adding more JDK 21 builds as time progresses. Ideally, we would need supporting data allowing us to identify the combination of operating system/Java versions which are most actively used and should be tested daily.

.github/workflows/maven.yml
melissalinkert: Done in a07fcde.

sbesson (approved):

dgault (approved):
Changes here all look good from my side

PR #4147, opened 2024-02-01
Added support for ColumbusReader to import data with z stacks

TBlackmore:
Following up on issue #3791
I had the same problem importing data exported from PerkinElmer (Revvity) Columbus into Omero. No Z stacks were coming through. I've ammended the ColumbusReader and run some tests.
I have some test data I can provide, but I'm having an issue getting into Zenodo at the moment.

PR #4155, opened 2024-02-13
Prairie: look for "<PV" string instead of just "PV"

melissalinkert:
Type checking is meant to verify that there is an XML element named something like `PVScan` or `PVStateShard`, so checking for `<PV` is a little more accurate.

I'd expect `isThisType` tests to fail for some of the data in #4147 without this PR. Actual Prairie data should continue to pass with this change.


sbesson (approved):

dgault (approved):
Change looks fine to me and all existing datasets continue to pass testing.

PR #4157, opened 2024-02-22
OME-XML: populate original metadata table

melissalinkert:
Without this PR, `showinf -nopix` on any OME-XML file with `OriginalMetadata` annotations should show that the annotations are not stored in the reader's original metadata table. This issue was originally noticed with `OME/METADATA.ome.xml` files produced by bioformats2raw.

With this PR, `showinf -nopix` on the same test data should show that the original metadata table is populated as expected.

I'm not entirely sure why this was commented out in the first place, but that was ~14 years ago. I suspect we did not have a working method that created a `Hashtable` from `OriginalMetadata` annotations in an `OMEXMLMetadata`.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
+++ b/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java
@@ -35,6 +35,7 @@ package loci.formats.in;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Hashtable;
 import java.util.List;
 
 import loci.common.CBZip2InputStream;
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested by taking any sample file with original metadata e.g. https://downloads.openmicroscopy.org/images/MRC/EMDB/EMD-2225/, converting to to OME-TIFF and OME-XML using `bfconvert` and then reading it with `showinf -nopix`.

Without this PR, reading the OME-TIFF displays the same original metadata as the original format but reading the OME-XML displays no original metadata.

With this PR included, reading either the OME-TIFF or the OME-XML displays the same original metadata as the original format.

The same logic was activated in the OME-TIFF reader in a1c7b5f6c68eb29edc56176c411e0a1e737b4583 alongside with the OME-XML service implementation parsing the annotations. Update the OME-XML reader to use the same API seems completely reasonable.

dgault (approved):
Change looks good. Tested a similar workflow to Seb, converting some CZI samples to OME-XML. Without this PR the original metadata is not displayed when using `showinf` on the OME-XML. With this PR included the original metadata matches that expected.

PR #4158, opened 2024-03-01
Bump file-leak-detector to version 1.18 for JDK 21 support

sbesson:
With the current infrastructure, the file-leak-detector fails with

```
test-automated:
   [testng] Could not load field socket from SocketImpl: java.lang.NoSuchFieldException: socket
   [testng] Could not load field serverSocket from SocketImpl: java.lang.NoSuchFieldException: serverSocket
   [testng] File leak detector installed
   [testng] FATAL ERROR in native method: processing of -javaagent failed, processJavaStart failed
   [testng] Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
   [testng] V  [libjvm.so+0x974f64]  jni_FatalError+0x74
   [testng] Exception in thread "main" java.lang.reflect.InvocationTargetException
   [testng] V  [libjvm.so+0xaddf24]  JvmtiExport::post_vm_initialized()+0x394
   [testng] 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
   [testng] V  [libjvm.so+0xea50c4]  Threads::create_vm(JavaVMInitArgs*, bool*)+0x8a4
   [testng] 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
   [testng] V  [libjvm.so+0x97651e]  JNI_CreateJavaVM+0x4e
   [testng] 	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:560)
   [testng] C  [libjli.so+0x42cb]  JavaMain+0x8b
   [testng] 	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:572)
   [testng] C  [libjli.so+0x7e49]  ThreadJavaMain+0x9
   [testng] Caused by: java.lang.ClassNotFoundException: java.net.PlainSocketImpl
   [testng] 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
   [testng] 	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
   [testng] 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
   [testng] 	at java.base/java.lang.Class.forName0(Native Method)
   [testng] 	at java.base/java.lang.Class.forName(Class.java:421)
[...]

PR #4159, opened 2024-03-01
Turn off file-leak-detector if tests are run with Java 21

melissalinkert:
See discussion in #4158.

Not sure if there is a better way to do this (in particular, to exclude >= 21), but this should at least allow tests to run on Java 21 for now.

### DIFF
--- a/components/test-suite/build.xml
+++ b/components/test-suite/build.xml
@@ -6,7 +6,7 @@ Download Apache Ant from http://ant.apache.org/.
 Type "ant -p" for a list of targets.
 -->
 
-<project name="tests" default="jar" basedir="." xmlns:if="ant:if">
+<project name="tests" default="jar" basedir=".">
   <description>Build file for Bio-Formats testing framework project</description>
   <property name="root.dir" location="../.."/>
   <import file="${root.dir}/ant/java.xml"/>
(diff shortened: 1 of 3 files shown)


sbesson:
Discussed earlier today with @dgault and @melissalinkert, the current proposal offers a good compromise allowing to run the non-regression repository tests on JDK 21 without dropping support for JDK 8 yet and still testing file leaks.  
To facilitate debugging and identification of the JDK used in a test run, we might want to update `FormatReaderTestFactory.createInstances` to also print the Java version as an INFO level statement.

melissalinkert:
Java version should now be logged each time tests are run.

As suggested by @dgault, fafc56b picks a different file-leak-detector version based on the JDK version being used. The Ant build changes have been reverted so that file-leak-detector remains enabled. That seems to work locally - I can now both compile and run tests on both 8 and 21. Assuming it also works in the next 4 nightly builds in which it's included, that was easier than I expected and worth considering in the context of https://github.com/ome/ome-common-java/pull/72.

sbesson (commented):
Looks good and the nightly CI builds have been running on JDK 8, 11, 17 and 21 with this change included. Only remaining stylistic question is whether we should use something similar to https://github.com/ome/ome-codecs/pull/36 i.e. keep the `dependency` block in the dependencies and manage a new `file-leak-detector.version` property via profile.

sbesson (approved):

PR #4167, opened 2024-03-14
DICOM: treat ultrasound ECG data as timelapse

melissalinkert:
This is the very beginning of basic timelapse support, and does not extend to other modalities.

I don't have data that can be shared at the moment, but the use case that prompts this change is ultrasound data containing an ECG. Without this change, `showinf` that all planes are Z sections; with this change, all planes are timepoints. I am opening as a draft for feedback, but expect this will be substantially changed (if not closed).

Implementation details aside, this raises a few questions:

- is checking for the presence of specific modality/data type metadata the right way to determine whether a set of planes are a Z stack or a timelapse?
- if checking the modality/data type is the way to determine Z and T dimensions, which specific modalities should be treated as timelapse?

I don't recall if this has come up in discussion with @dclunie / @fedorov in the past, but any thoughts would be welcome.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -280,6 +280,7 @@ public class DicomReader extends SubResolutionFormatReader {
     Region currentRegion = new Region(x, y, w, h);
     int z = getZCTCoords(no)[0];
     int c = getZCTCoords(no)[1];
+    int timepoint = getZCTCoords(no)[2];
 
     if (!tilePositions.containsKey(getCoreIndex())) {
       LOGGER.warn("No tiles for core index = {}", getCoreIndex());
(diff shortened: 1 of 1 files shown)


fedorov:
@melissalinkert I always thought of `bioformats` as the library for microscopy imaging. Are you suggesting you are expanding the scope to cover radiology modalities? 

I do not have experience with US ECG imaging, so I am mostly responding based on my experience in other radiology modalities.

> is checking for the presence of specific modality/data type metadata the right way to determine whether a set of planes are a Z stack or a timelapse?

At least not always. In MR/CT/PET you would typically use `ImagePositionPatient` to stack the slices into a volume.

I don't have experience working with ultrasound images, but from what I know, they are notorious for the difficulties of figuring out the geometry and reconstructing volume. This page might be helpful to get a perspective: https://github.com/SlicerHeart/SlicerHeart/blob/master/Docs/ImageImportExport.md. 

> if checking the modality/data type is the way to determine Z and T dimensions, which specific modalities should be treated as timelapse?

I don't think you can make a simple rule based on modality.

melissalinkert:
Thanks for the feedback, @fedorov. Given that this a more complex issue than it first appears, I'll close this pull request and open something new at a later date.

Bio-Formats has had basic support for other modalities for a long time, in that you should be able to at least read images. Most metadata for radiology modalities does not have a place in the OME schema, though, so is not as well-supported as WSI data. https://downloads.openmicroscopy.org/images/DICOM/samples/ has public examples of files in other modalities that are included in our nightly tests; those have been readable since the early days of Bio-Formats (~2007-2008).

dclunie:
FYI, the short answer for traditional US and similar XA (non-enhanced-family) multi-frame (NumberOfFrames present and > 1, [Multi-frame Module](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.6.html)) is FrameTime or FrameTimeVector ([Cine Module](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.5.html)) that are the target of FrameIncrementPointer.

You can probably find some relevant samples in the images at ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG12/ and also ftp://dicom.offis.de/pub/dicom/images/

PR #4168, opened 2024-03-18
DICOM reader: better handling of nested sequences that include pixel data

### LINKED ISSUE #4165: BioFormats DICOM WSI reader problem with nested Pixel Data in private data element and/or offset tables
One of my tools creates DICOM WSI images that are dual-personality TIFF, but also
contain a "hidden" pyramid of TIFF sub-layers, using a private data element, so
that ordinary TIFF readers have the pyramid they are expecting, even though DICOM
spreads the pyramid over multiple files.

Normal DICOM readers ignore these hidden, and use the PixelData element in the
top level data set.

Unfortunately, adding these seems to confuse the BioFormats reader.

See the samples at:

  https://www.dropbox.com/scl/fo/za255deqf1rrmmwqsv5lv/h?rlkey=b3uow8ormdaz0tr4q7g3u5yyl

with screenshots of QuPath (5) using BioFormats and OpenSlide readers.

I am just guessing that the nested Pixel Data in the private data element is the
problem, since if I don't include it, the BioFormats reader works OK. It may be
something else.

These test images also contain both the Basic Offset Table and Extended
Offset Tables with pointers to the actual pixel data frame Items. I am testing including
these to allow faster reading when present. Note that the standard says that only
one or the other should be present, not both, so these are just for testing. Not
sure if BioFormats takes advantage of the presence of either.

melissalinkert:
Fixes #4165.

Using the data in `inbox/gh-4165`, `showinf` should show that each pyramid resolution starts with "scrambled" tiles. `showinf -noflat -resolution 2 img_0.dcm` or `showinf -nogroup img_3.dcm` in particular can easily reproduce the issue.

With this change, each resolution and extra image in the dataset should display correctly.

As described in #4165, there are PixelData elements (which Bio-Formats' reader should ignore) that are nested within a private sequence element. The reader will no longer stop at the first PixelData encountered, and instead should skip over any PixelData contained within a sequence.

### DIFF
--- a/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
+++ b/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
@@ -310,15 +310,33 @@ public class DicomTag implements Comparable<DicomTag> {
                 break;
               }
               else if (child.attribute == PIXEL_DATA) {
-                stop = fp;
-                break;
+                child.parent = this;
+                children.add(child);
+                if (child.elementLength == -1) {
+                  // look for end of sequence
+                  long seek = fp - 2;
+                  int nextTag = 0;
+                  while (seek < in.length() && (nextTag != SEQUENCE_DELIMITATION_ITEM.getTag() && nextTag != ITEM_DELIMITATION_ITEM.getTag())) {
+                    seek += 2;
+                    in.seek(seek);
+                    try {
+                      nextTag = getNextTag(in);
+                    }
+                    catch (Exception e) {
+                    }
+                    if (nextTag == 0xfeffdde0 || nextTag == 0xfeff0d0e) {
+                      in.order(!in.isLittleEndian());
+                      break;
+                    }
+                  }
+                }
               }
               else if (child.attribute != ITEM && child.attribute != ITEM_DELIMITATION_ITEM) {
                 child.parent = this;
                 children.add(child);
               }
             }
-            if (elementLength == 0) {
+            if (elementLength <= 0) {
               elementLength = (int) (stop - start);
             }
             in.seek(stop);
@@ -438,6 +456,7 @@ public class DicomTag implements Comparable<DicomTag> {
 
     switch (vr) {
       case OB:
+      case OV:
       case OW:
       case SQ:
       case UN:
--- a/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
+++ b/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
@@ -719,6 +719,17 @@ public class DicomTag implements Comparable<DicomTag> {
     }
   }
 
+  /**
+   * See https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_7.8.html
+   *
+   * @return true if this is a private content creator tag
+   */
+  public boolean isPrivateContentCreator() {
+    int highWord = (tag >> 16) & 0xffff;
+    int lowWord = tag & 0xffff;
+    return highWord % 2 == 1 && lowWord == 0x0010;
+  }
+
   @Override
   public String toString() {
     if (key == null) {
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -37,9 +37,11 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMap.Builder;
@@ -126,6 +128,8 @@ public class DicomReader extends SubResolutionFormatReader {
   private boolean edf = false;
 
   private List<DicomTag> tags;
+
+  private Set<Integer> privateContentHighWords = new HashSet<Integer>();
 
   // -- Constructor --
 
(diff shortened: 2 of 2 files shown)


sbesson (commented):
Tested using the public DICOM samples shared in #4165. Without this PR I confirm the whole slide image appears as scrambled as the result of the Pixel Data handling.
Screenshot 2024-03-29 at 12 06 41

With this PR included, all resolution levels of the whole slide image as well as the associated image read without issue using the command-line utility.

One minor concern looking the metadata output of the `showinf -noflat` and comparing it with the output of the `dcdump` utility on the same file is that the series metadata includes information for all resolutions of the hidden pyramid.

```
% ./tools/showinf -noflat -nopix ~/Downloads/gh-4165/DCM/img_0.dcm | grep Total
0048,0006 Total Pixel Matrix Columns #1: 46000
0048,0006 Total Pixel Matrix Columns #2: 23000
0048,0006 Total Pixel Matrix Columns #3: 11500
0048,0006 Total Pixel Matrix Columns #4: 5750
0048,0006 Total Pixel Matrix Columns #5: 2875
0048,0006 Total Pixel Matrix Columns #6: 1437
0048,0006 Total Pixel Matrix Columns #7: 718
0048,0006 Total Pixel Matrix Columns #8: 359
0048,0006 Total Pixel Matrix Columns #9: 179
0048,0007 Total Pixel Matrix Rows #1: 32914
0048,0007 Total Pixel Matrix Rows #2: 16457
0048,0007 Total Pixel Matrix Rows #3: 8228
0048,0007 Total Pixel Matrix Rows #4: 4114
0048,0007 Total Pixel Matrix Rows #5: 2057
0048,0007 Total Pixel Matrix Rows #6: 1028
0048,0007 Total Pixel Matrix Rows #7: 514
0048,0007 Total Pixel Matrix Rows #8: 257
0048,0007 Total Pixel Matrix Rows #9: 128
0048,0303 Total Pixel Matrix Focal Planes: 1
```

This information is slightly at odds with the DICOM resolutions detected by the reader which only include 3 levels of sizeX 46000 11500 and 2875.

melissalinkert:
daaa2f9 should fix the original metadata table issue in https://github.com/ome/bioformats/pull/4168#pullrequestreview-1968481092

sbesson (approved):
With the last commit, the difference in the original metadata for the first series of the sample shared in #4165 is

 ```diff
8,9c8
< 0008,0008 Image Type #1: DERIVED\PRIMARY\VOLUME\NONE
< 0008,0008 Image Type #2: DERIVED\PRIMARY\VOLUME\RESAMPLED
---
> 0008,0008 Image Type: DERIVED\PRIMARY\VOLUME\NONE
87,88d85
< 0008,1150 Referenced SOP Class UID: 1.2.840.10008.5.1.4.1.1.77.1.6
< 0008,1155 Referenced SOP Instance UID: 1.3.6.1.4.1.5962.99.1.3527156866.2038118173.1708629206146.9.0
104,105c101
< 0020,000e Series Instance UID #1: 1.3.6.1.4.1.5962.99.1.3527156866.2038118173.1708629206146.4.0
< 0020,000e Series Instance UID #2: 1.3.6.1.4.1.5962.99.1.3527156866.2038118173.1708629206146.4.0
---
> 0020,000e Series Instance UID: 1.3.6.1.4.1.5962.99.1.3527156866.2038118173.1708629206146.4.0
119,130c115,120
< 0028,0002 Samples per pixel #1: 3
< 0028,0002 Samples per pixel #2: 3
< 0028,0004 Photometric Interpretation #1: RGB
< 0028,0004 Photometric Interpretation #2: YBR_FULL_422
< 0028,0006 Planar Configuration #1: 0
< 0028,0006 Planar Configuration #2: 0
< 0028,0008 Number of frames #1: 23220
< 0028,0008 Number of frames #2: 5850
< 0028,0010 Rows #1: 256
< 0028,0010 Rows #2: 256
< 0028,0011 Columns #1: 256
< 0028,0011 Columns #2: 256
---
> 0028,0002 Samples per pixel: 3
> 0028,0004 Photometric Interpretation: RGB
> 0028,0006 Planar Configuration: 0
> 0028,0008 Number of frames: 23220
> 0028,0010 Rows: 256
> 0028,0011 Columns: 256
132,137c122,124
< 0028,0100 Bits Allocated #1: 8
< 0028,0100 Bits Allocated #2: 8
< 0028,0101 Bits Stored #1: 8
< 0028,0101 Bits Stored #2: 8
< 0028,0102 High Bit #1: 7
< 0028,0102 High Bit #2: 7
---
> 0028,0100 Bits Allocated: 8
> 0028,0101 Bits Stored: 8
> 0028,0102 High Bit: 7
159,162c146,147
< 0048,0006 Total Pixel Matrix Columns #1: 46000
< 0048,0006 Total Pixel Matrix Columns #2: 23000
< 0048,0007 Total Pixel Matrix Rows #1: 32914
< 0048,0007 Total Pixel Matrix Rows #2: 16457
---
> 0048,0006 Total Pixel Matrix Columns: 46000
> 0048,0007 Total Pixel Matrix Rows: 32914
```

The closest analogy I could think of would be TIFF files using private tags. Currently, a TIFF-based reader might elect to read this data and add it to the original metadata table and/or use it to interpret the data structure but there is no expectation that the reader will parse and read such private tags by default. 
From my perspective, the last commit is a similar interpretation of DICOM private data. In this case, the metadata from a private sequence excluded when reading the pixel data is also excluded from the original metadata table. It would be useful to get additional reviews and make sure we are happy with the behavior.

dgault (approved):
Tested using the sample files shared in `inbox/gh-4165` and confirmed that the issue with the 'scrambled' images has been resolved for all resolutions with this PR. All other builds and tests have been green with this PR included.

As for handling the metadata , based on the original description in https://github.com/ome/bioformats/issues/4165, I would agree with Seb's private tag analogy.  In this scenario leaving the tags related to the hidden pyramids out of the original metadata seems like the correct course of action from my side.

PR #4172, opened 2024-03-25
testSaneUsedFiles: check getCurrentFile is the first file in the getUsedFiles list

sbesson:
From an discussion with @melissalinkert and @jburel, this was opening initially to assess how many formats deviate from this assumption in the nightly CI builds

`FormatReaderTest.testSaneUsedFiles ` is currently testing that the first item of [IFormatReader.getUsedFiles()](https://downloads.openmicroscopy.org/bio-formats/8.0.0/api/loci/formats/IFormatReader.html#getUsedFiles--) is equal to [IFormatReader.getCurrentFile()](https://downloads.openmicroscopy.org/bio-formats/8.0.0/api/loci/formats/IFormatReader.html#getCurrentFile--) for single-file formats.

This change extends the test logic to assert this condition for all file formats manually excluding the 10 readers where the automated tests have identified the condition was not met.

As next steps, the individual readers can be updated individually to meet the condition and reincluded in the tests. Once all readers have been updated, the API contract could be updated to clarify the expectation and the relationship between the first file of `getUsedFiles` and the output of `getCurrentFile` although doing this will likely require a major version increment.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
+++ b/components/test-suite/src/loci/tests/testng/FormatReaderTest.java
@@ -1866,9 +1866,19 @@ public class FormatReaderTest {
         msg = "Used files list contains duplicates";
       }
 
-      if (!base[0].equals(file)) {
+      if (!(reader.getFormat().equals("Bio-Rad PIC")) &&
+          !(reader.getFormat().equals("Metamorph STK")) &&
+          !(reader.getFormat().equals("Evotec Flex")) &&
+          !(reader.getFormat().equals("CellSens VSI")) &&
+          !(reader.getFormat().equals("PerkinElmer")) &&
+          !(reader.getFormat().equals("Fuji LAS 3000")) &&
+          !(reader.getFormat().equals("Micro-Manager")) &&
+          !(reader.getFormat().equals("BDV")) &&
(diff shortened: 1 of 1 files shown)


sbesson:
As expected, quite a few failures reported in the nightly CI builds. The reader for the following curated QA folders should likely be reviewed as a next step:

- `biorad`
- `metamorph`
- `flex`
- `cellsens`
- `ultraview`
- `fuji`
- `micromanager`
- `perkinelmer`
- `bdv`
- `zeiss-tiff`
- `scanr`

Marking this PR as draft in the meantime

sbesson:
As discussed today, assigning to the `8.0.0` milestone with the aim to:
1- review and update the contract of `IFormatReader` to enforce consistency between `getCurrentFile` and `getUsedFiles`
2- review and update all the readers for the failing formats

sbesson:
Updated the PR and the description to test `getUsedFiles[0] == getCurrentFile` and exclude the formats that do not meet the criterion. These readers can be reviewed and relaxed as part of future PRs.

melissalinkert (approved):
I think this makes sense since, as discussed, the goal is to remove the reader format checks incrementally in follow-up PRs.

This would have been included in https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/244/, so merging.

PR #4173, opened 2024-03-26
bfconvert: remove extra channels if `-channel` was used

### LINKED ISSUE #4149: bfconvert: Using channel option does not remove additional channel elements from OME-XML
Issue was reported on forum thread https://forum.image.sc/t/extracting-a-single-channel-using-bfconvert/91632

The original error was reported when trying to read using AICSImage resulted in the error:
```
ValueError: conflicting sizes for dimension 'C': length 1 on the data but length 2 on coordinate 'C'
```

The issue can be reproduced with command line tools from BF 7.1.0 and using sample file from https://github.com/NEUBIAS/training-resources/blob/2145f31cd46b81fc24f31cad4de1ca30ccaf24f6/image_data/xyc_8bit__membranes_nuclei.tif

Running the below command then generates an OME-TIFF which contains 2 channel elements in the OME-XML:
```
bfconvert -channel 1 xyc_8bit__membranes_nuclei.tif xyc_8bit__membranes_nuclei.ome.tiff
```

Note that this is not apparent when using `showinf -omexml` but instead can be seen when using a tool such as tiffinfo to view the image description tag.

When viewed with `showinf -omexml` only a single channel element is seen:
```
   <Image ID="Image:0" Name="xyc_8bit__membranes_nuclei.tiff">
      <Description/>
      <Pixels BigEndian="true" DimensionOrder="XYCZT" ID="Pixels:0" Interleaved="false" SignificantBits="8" SizeC="1" SizeT="1" SizeX="475" SizeY="354" SizeZ="1" Type="uint8">
            <LightPath/>
         </Channel>
         <TiffData FirstC="0" FirstT="0" FirstZ="0" IFD="0" PlaneCount="1">
            <UUID FileName="xyc_8bit__membranes_nuclei.ome.tiff">urn:uuid:26dd8113-b0f8-465d-b860-52ef2d031195</UUID>
         </TiffData>
      </Pixels>
   </Image>
```

When using tiffinfo you see 2 channel elements:
```
   <Image ID="Image:0" Name="xyc_8bit__membranes_nuclei.tiff">
       <Description/>
       <Pixels BigEndian="true" DimensionOrder="XYCZT" ID="Pixels:0" Interleaved="false" SignificantBits="8" SizeC="1" SizeT="1" SizeX="475" SizeY="354" SizeZ="1" Type="uint8">
              <LightPath/>
           </Channel>
           <Channel ID="Channel:0:1" SamplesPerPixel="1"><LightPath/></Channel>
           <TiffData FirstC="0" FirstT="0" FirstZ="0" IFD="0" PlaneCount="1">
              <UUID FileName="xyc_8bit__membranes_nuclei.ome.tiff">urn:uuid:26dd8113-b0f8-465d-b860-52ef2d031195</UUID> 
           </TiffData>
       </Pixels>
   </Image>
```

melissalinkert:
Fixes #4149.

Compare the test case in #4149 and/or something like `bfconvert "test&sizeC=2.fake" test.ome.tiff -channel 1` with and without this PR. Without this PR, the output OME-TIFF should have 2 `Channel`s defined in the OME-XML produced by `tiffcomment` or `showinf -nopix -omexml`. With this PR, the output OME-TIFF should have a single `Channel`.

This should also be tested with both `-channel` and `-series` used, so that the case where most `Image` and `Pixels` elements are removed is also covered.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -41,6 +41,7 @@ import java.io.InputStreamReader;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.HashMap;
+import java.util.List;
 import java.util.TreeMap;
 import java.util.SortedMap;
 
@@ -84,6 +85,7 @@ import loci.formats.services.OMEXMLService;
 import loci.formats.services.OMEXMLServiceImpl;
 
 import ome.xml.meta.OMEXMLMetadataRoot;
+import ome.xml.model.Channel;
 import ome.xml.model.Image;
 import ome.xml.model.Pixels;
 import ome.xml.model.enums.PixelType;
@@ -591,11 +593,22 @@ public final class ImageConverter {
         String xml = service.getOMEXML(service.asRetrieve(store));
         OMEXMLMetadataRoot root = (OMEXMLMetadataRoot) store.getRoot();
         IMetadata meta = service.createOMEXMLMetadata(xml);
+        OMEXMLMetadataRoot newRoot = (OMEXMLMetadataRoot) meta.getRoot();
         if (series >= 0) {
-          Image exportImage = new Image(root.getImage(series));
-          Pixels exportPixels = new Pixels(root.getImage(series).getPixels());
+          Image exportImage = newRoot.getImage(series);
+          Pixels exportPixels = newRoot.getImage(series).getPixels();
+
+          if (channel >= 0) {
+            List<Channel> channels = exportPixels.copyChannelList();
+
+            for (int c=0; c<channels.size(); c++) {
+              if (c != channel) {
+                exportPixels.removeChannel(channels.get(c));
+              }
+            }
+          }
+
           exportImage.setPixels(exportPixels);
-          OMEXMLMetadataRoot newRoot = (OMEXMLMetadataRoot) meta.getRoot();
           while (newRoot.sizeOfImageList() > 0) {
             newRoot.removeImage(newRoot.getImage(0));
           }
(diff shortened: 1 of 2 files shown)


components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -591,11 +593,22 @@ public boolean testConvert(IFormatWriter writer, String[] args)
...
+            for (int c=0; c<channels.size(); c++) {
+              if (c != channel) {
+                exportPixels.removeChannel(channels.get(c));
+              }
+            }
+          }
+
           exportImage.setPixels(exportPixels);
-          OMEXMLMetadataRoot newRoot = (OMEXMLMetadataRoot) meta.getRoot();
           while (newRoot.sizeOfImageList() > 0) {
```
sbesson, line 612: Outside to the scope of this PR, this block could benefit from the new `clearImageList` API introduced in https://github.com/ome/ome-model/pull/185 and which should be available in the next minor release of `ome-xml`.

sbesson (changes requested):
Functionally tested using 
```
bfconvert "test&sizeC=2.fake" test.ome.tiff -channel 1
bfconvert "test&series=3&sizeC=2.fake" test2.ome.tiff -channel 1
bfconvert "test&series=3&sizeC=2.fake" test3.ome.tiff -series 2 -channel 1
```

Without this PR, all the channels from the original files are transferred to the output file creating a mismatch with the value of `sizeC`.

Without this PR, only the exported channel is now copied to the target file OME metadata.

Before merging this, would it be possible to add a minimal test to `ImageConverterTest` covering this scenario to prevent future regressions?

sbesson (approved):

PR #4177, opened 2024-04-04
Fix some Java 9+ warnings in formats-bsd

melissalinkert:
I was working on something else and didn't want to scroll through as many warnings to find compile errors. See for example https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#%3Cinit%3E(double).

Only touches the formats-bsd component (others could use the same treatment though), and tried not to modify things that are actively being worked on in other PRs.

Very low priority for review and can be excluded if needed.


dgault (approved):
The code changes all look fine from my side and align with the recommendations on the deprecation notices. Builds and tests have been green with this PR included and no regressions have been seen.

PR #4178, opened 2024-04-08
Fix most Java 9+ deprecation warnings in bio-formats-plugins component

melissalinkert:
In the same spirit as #4177.

The `Class.newInstance()` changes are as recommended in https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#newInstance(). There should now be only 8 warnings in the bio-formats-plugins component.

Scheduling for 8.0.0 as discussed earlier today, so that we can sort out warnings right after 7.3.0.


dgault (approved):
Changes all look good to me and the builds and tests are green with this PR included. Some manual testing of the plugin component, reproducing the release tests, also showed no issues or regressions.

PR #4179, opened 2024-04-09
Fix Java 9+ deprecation warnings in smaller components

melissalinkert:
In particular, `bio-formats-tools`, `formats-api`, and `test-suite`. This is similar to #4177 and #4178; only the `formats-gpl` component still requires work.

### DIFF
--- a/components/test-suite/src/loci/tests/testng/Configuration.java
+++ b/components/test-suite/src/loci/tests/testng/Configuration.java
@@ -202,12 +202,12 @@ public class Configuration {
     if (delim >= 0) {
       test = test.substring(0, delim);
     }
-    return new Boolean(test.trim()).booleanValue();
+    return Boolean.parseBoolean(test.trim());
   }
 
   public boolean hasValidXML() {
     if (globalTable.get(HAS_VALID_XML) == null) return true;
-    return new Boolean(globalTable.get(HAS_VALID_XML)).booleanValue();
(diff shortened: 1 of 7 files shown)


components/test-suite/src/loci/tests/testng/Configuration.java
```diff
@@ -828,7 +828,7 @@ private Length getPhysicalSize(String valueKey, String unitKey) {
     String units = currentTable.get(unitKey);
     try {
       UnitsLength unit = units == null ? UnitsLength.MICROMETER : UnitsLength.fromString(units);
-      return physicalSize == null ? null : UnitsLength.create(new Double(physicalSize), unit);
+      return physicalSize == null ? null : UnitsLength.create(DataTools.parseDouble(physicalSize), unit);
```
dgault: Excluding for now due to test failures in a number of formats. I have so far traced it down to this section of code as the cause.

melissalinkert: I believe 4662b65c55 fixes this, but leaving excluded until after 7.3.0 is out just in case.

melissalinkert:
Re-excluding due to https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/5396/, https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/5543/

dgault (approved):
All of the changes here look good. Builds and tests are all green with the PR included.

PR #4180, opened 2024-04-11
Columbus: update to use micrometers instead of `reference frame` for positions

melissalinkert:
See https://github.com/ome/bioformats/pull/4147#pullrequestreview-1989199061

This just updates the units for `Plane` and `WellSample` positions, since we are reasonably confident that the values are in micrometers.

Scheduling for 8.0.0 and excluding for now, in case I missed any config changes which would cause a test failure.


dgault (approved):
Changes look to be fine from my side and with the accompanying config PR all of the tests have remained green.

PR #4182, opened 2024-04-12
Fix Java 9+ warnings in formats-gpl component

melissalinkert:
Last planned PR that goes together with #4177, #4178, and #4179.

As with #4179, excluding for now so this doesn't risk a failing build leading up to 7.3.0. I wouldn't really expect test failures, but this does modify many readers.


dgault (approved):
All of the code changes included here look goof from my perspective. The builds and tests have continued to pass with this PR included and local testing of multiple formats shows no regressions.

PR #4183, opened 2024-04-16
Bump lower level component versions prior to 7.3.0

dgault:
Bumping versions of lower level components prior to the 7.3.0 release


dgault:
Merging as all builds are green. I re-released ome-codecs and ome-model due to incorrect Java build versions for the initial 6.3.5 and 1.0.2 bumps.

PR #4188, opened 2024-04-18
Fix a few string comparisons, fixes #4184

### LINKED ISSUE #4184: Strings are being compared with == and !=
2 occurrences where 2 Strings are being compared with == and !=.

They should be using String.equals()

https://github.com/ome/bioformats/blob/6fa35f6deb652ac34e3bb2fba5b8b694481d3585/components/formats-bsd/src/loci/formats/gui/PreviewPane.java#L218

https://github.com/ome/bioformats/blob/6fa35f6deb652ac34e3bb2fba5b8b694481d3585/components/formats-bsd/src/loci/formats/gui/PreviewPane.java#L263

should be?

Line 218 if (id.equals(lastId)) continue;
Line 263 if (!id.equals(loadId)) {

melissalinkert:
I don't think we have relevant automated tests, so it's probably a good idea to check `File > Open` in the image window shown by

* double-clicking on `bioformats_package.jar`
* running `showinf` on anything (even `test.fake`)
 
to make sure that a thumbnail is displayed in the file chooser and there isn't anything obviously broken. I wouldn't expect an actual change in behavior with this PR.

### DIFF
--- a/components/formats-bsd/src/loci/formats/gui/PreviewPane.java
+++ b/components/formats-bsd/src/loci/formats/gui/PreviewPane.java
@@ -215,7 +215,9 @@ public class PreviewPane extends JPanel
 
       try { // catch-all for unanticipated exceptions
         final String id = loadId;
-        if (id == lastId) continue;
+        if ((id == null && lastId == null) || (id != null && id.equals(lastId))) {
+          continue;
+        }
         if (id != null && lastId != null) {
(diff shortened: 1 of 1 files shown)


dgault (approved):
Changes look good and attempting to use `File > Open` from the showinf display, results in a preview thumbnail being displayed as expected.

PR #4181, opened 2024-04-12
Add support for reading compressed NDPI tiles

melissalinkert:
This expands on the "precompressed" features added in #3992.

The main goal was to just add compressed tile reading to the NDPI reader, so that conversion from NDPI to DICOM wouldn't require much if any recompression. This required a variety of changes not limited to NDPI though:

- update `JPEGTurboService` so that the repackaged compressed tiles can be retrieved, as well as the tile size and counts
- move general TIFF tile copying logic out of `SVSReader`, and into `MinimalTiffReader` (to prevent duplicate logic in `NDPIReader`)
  - this should also make it easier to support precompressed tiles in other TIFF-based formats later on
- update `NDPIReader` to report optimal sizes based on actual IFD/JPEG stream, instead of constant 1024x1024
- update `DicomWriter` to allow different tile sizes in each resolution, since NDPI reports different sizes per resolution
- update `NDPIReader` to use the `saveCompressedBytes` API

The end result is that something like this should now work:

```bfconvert -noflat -precompressed -compression JPEG CMU-1.ndpi cmu.dcm```

and should complete within a few seconds as all tiles can be copied without any recompression.

I would not expect SVS -> DICOM support to change, but it probably makes sense to double-check just in case the refactoring caused any problems.

I don't expect this to fail tests, but excluding for now so this doesn't distract from 7.3.0. Given the API changes in `JPEGTurboService`, this probably does require a major version.

cc @dclunie, @fedorov

PR #4164, opened 2024-03-13
Reduce DICOM write time

melissalinkert:
Opening as a draft since this is a work in progress, with additional changes to both reading and writing planned.

4e81795 should reduce the total time to run `bfconvert` with DICOM output. This is just buffering the header writes, and does not change the tile writing. A simple test conversion using https://downloads.openmicroscopy.org/images/Vectra-QPTIFF/perkinelmer/PKI_scans/HandEcompressed_Scan1.qptiff:

```
bfconvert -no-upgrade -noflat -tilex 256 -tiley 256 -compression JPEG HandEcompressed_Scan1.qptiff test.dcm
```

is noticeably faster with 4e81795 when testing locally.

PR #4190, opened 2024-05-23
Extend precompressed support to read DICOM and write TIFF/OME-TIFF

melissalinkert:
Opening as a draft for now, as #4181 is higher priority. This may need some updates once #4181 is merged.

I would expect basic conversions such as `bfconvert -noflat -precompressed -compression JPEG input.dcm output.ome.tiff` to work, where the input DICOM is part of a whole slide dataset. Tile sizes should be correctly picked up without being specified in the `bfconvert` options.

A few outstanding things related to this PR and #4181:

- `-precompressed` does still require the `-compression` option to be set correctly; removing that requirement could be done here
- documentation started in https://github.com/ome/bio-formats-documentation/pull/349 should be updated to include additions here and in #4181 - we may want a separate page just for this feature

/cc @fedorov, @dclunie

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -1260,11 +1260,9 @@ public final class ImageConverter {
   private boolean doTileConversion(IFormatWriter writer, String outputFile)
     throws FormatException
   {
-    if (writer instanceof DicomWriter ||
-      (writer instanceof ImageWriter && ((ImageWriter) writer).getWriter(outputFile) instanceof DicomWriter))
-    {
-      MetadataStore r = reader.getMetadataStore();
-      return !(r instanceof IPyramidStore) || ((IPyramidStore) r).getResolutionCount(reader.getSeries()) > 1;
+    MetadataStore r = reader.getMetadataStore();
+    if ((r instanceof IPyramidStore) && ((IPyramidStore) r).getResolutionCount(reader.getSeries()) > 1) {
+      return true;
     }
     return DataTools.safeMultiply64(width, height) >= DataTools.safeMultiply64(4096, 4096) ||
       saveTileWidth > 0 || saveTileHeight > 0;
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -1268,16 +1268,13 @@ public final class ImageConverter {
   private boolean doTileConversion(IFormatWriter writer, String outputFile)
     throws FormatException
   {
-    MetadataStore r = reader.getMetadataStore();
-    if ((r instanceof IPyramidStore) && ((IPyramidStore) r).getResolutionCount(reader.getSeries()) > 1) {
-      // if we asked to try a precompressed conversion,
-      // then the writer's tile sizes will have been set automatically
-      // according to the input data
-      // the conversion must then be performed tile-wise to match the tile sizes,
-      // even if precompression doesn't end up being possible
-      if (precompressed) {
-        return true;
-      }
+    // if we asked to try a precompressed conversion,
+    // then the writer's tile sizes will have been set automatically
+    // according to the input data
+    // the conversion must then be performed tile-wise to match the tile sizes,
+    // even if precompression doesn't end up being possible
+    if (precompressed) {
+      return true;
     }
     return DataTools.safeMultiply64(width, height) >= DataTools.safeMultiply64(4096, 4096) ||
       saveTileWidth > 0 || saveTileHeight > 0;
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -767,7 +767,9 @@ public final class ImageConverter {
         int writerSeries = series == -1 ? q : 0;
         writer.setSeries(writerSeries);
         writer.setResolution(res);
+
         writer.setInterleaved(reader.isInterleaved() && !autoscale);
+
         writer.setValidBitsPerPixel(reader.getBitsPerPixel());
         int numImages = writer.canDoStacks() ? reader.getImageCount() : 1;
 
@@ -837,6 +839,12 @@ public final class ImageConverter {
               setCodecOptions(writer);
               writer.setId(tileName);
               if (compression != null) writer.setCompression(compression);
+            }
+          }
+
+          if (precompressed && FormatTools.canUsePrecompressedTiles(reader, writer, writer.getSeries(), writer.getResolution())) {
+            if (getReaderCodecName().startsWith("JPEG")) {
+              writer.setInterleaved(true);
             }
           }
 
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -1284,6 +1284,14 @@ public final class ImageConverter {
     if (precompressed) {
       return true;
     }
+    // tile size has already been set in the writer,
+    // so tile-wise conversion should be performed
+    // independent of image size
+    if ((writer.getTileSizeX() > 0 && writer.getTileSizeX() < width) ||
+      (writer.getTileSizeY() > 0 && writer.getTileSizeY() < height))
+    {
+      return true;
+    }
     return DataTools.safeMultiply64(width, height) >= DataTools.safeMultiply64(4096, 4096) ||
       saveTileWidth > 0 || saveTileHeight > 0;
   }
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -62,6 +62,7 @@ import loci.formats.FilePattern;
 import loci.formats.FileStitcher;
 import loci.formats.FormatException;
 import loci.formats.FormatTools;
+import loci.formats.ICompressedTileReader;
 import loci.formats.IFormatReader;
 import loci.formats.IFormatWriter;
 import loci.formats.ImageReader;
@@ -71,7 +72,9 @@ import loci.formats.Memoizer;
 import loci.formats.MetadataTools;
 import loci.formats.MinMaxCalculator;
 import loci.formats.MissingLibraryException;
+import loci.formats.codec.Codec;
 import loci.formats.codec.CodecOptions;
+import loci.formats.codec.CompressionType;
 import loci.formats.codec.JPEG2000CodecOptions;
 import loci.formats.gui.Index16ColorModel;
 import loci.formats.in.DynamicMetadataOptions;
@@ -557,6 +560,11 @@ public final class ImageConverter {
     }
 
     reader.setId(in);
+
+    if (compression == null && precompressed) {
+      compression = getReaderCodecName();
+      LOGGER.info("Implicitly using compression = {}", compression);
+    }
 
     if (swapOrder != null) {
        dimSwapper.swapDimensions(swapOrder);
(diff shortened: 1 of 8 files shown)


components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -832,6 +842,12 @@ else if (saveTileWidth > 0 && saveTileHeight > 0) {
             }
           }
 
+          if (precompressed && FormatTools.canUsePrecompressedTiles(reader, writer, writer.getSeries(), writer.getResolution())) {
+            if (getReaderCodecName().startsWith("JPEG")) {
+              writer.setInterleaved(true);
```
sbesson, line 847: Is there a reason to special case here or should we ensure that `reader.getInterleaved` returns true appropriately so that

```
writer.setInterleaved(reader.isInterleaved() && !autoscale);
```

works correctly above?

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -558,6 +561,11 @@ public boolean testConvert(IFormatWriter writer, String[] args)
 
     reader.setId(in);
 
+    if (compression == null && precompressed) {
+      compression = getReaderCodecName();
```
sbesson, line 565: Actually looking forward to possible handling mixed compressions, is it preferable moving forward to _not_ specify the compression as an option and leave the internal logic detect the codec?

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -1260,19 +1276,13 @@ private boolean isTiledWriter(IFormatWriter writer, String outputFile)
...
-      }
-      MetadataStore r = reader.getMetadataStore();
-      return !(r instanceof IPyramidStore) || ((IPyramidStore) r).getResolutionCount(reader.getSeries()) > 1;
+    // if we asked to try a precompressed conversion,
+    // then the writer's tile sizes will have been set automatically
+    // according to the input data
+    // the conversion must then be performed tile-wise to match the tile sizes,
+    // even if precompression doesn't end up being possible
+    if (precompressed) {
+      return true;
```
sbesson, line 1285: Definitely good to see `instanceof` checks removed. Are there specific tests that should be carried out here if the target format is neither DICOM not TIFF/OME-TIFF?

sbesson (changes requested):
In the spirit of https://github.com/ome/bioformats/pull/3992 and https://github.com/ome/bioformats/pull/4181, used `CMU-1.svs` and `CMU-1.npdi` as the source files for conversion.

For both files, the `bfconvert` utility with this PR included was executed:
- using both OME-TIFF and DICOM as the target format
- using `-compression JPEG`, `-compression JPEG-2000` and `-precompressed` as the options
Additionally the two DICOM datasets generated with the JPEG and JPEG-2000 compression from CMU-1.ndpi were used as the source of a conversion from DICOM to OME-TIFF using the same three compression options as above.

| Source file               	| Target   	| Precompressed 	| Compression 	| Conversion time 	| File size 	|
|---------------------------	|----------	|---------------	|-------------	|----------------:	|-----------	|
| CMU-1.svs                 	| DICOM    	| no            	| JPEG        	| 229.057s        	| 109M      	|
| CMU-1.svs                 	| DICOM    	| no            	| JPEG-2000   	| 682.503         	| 539M      	|
| CMU-1.svs                 	| DICOM    	| yes           	|             	| 83.013s         	| 177M      	|
| CMU-1.svs                 	| OME-TIFF 	| no            	| JPEG        	| 152.064s        	| 247M      	|
| CMU-1.svs                 	| OME-TIFF 	| no            	| JPEG-2000   	| 518.381s        	| 735M      	|
| CMU-1.svs                 	| OME-TIFF 	| yes           	|             	| 27.083s         	| 176M      	|
| CMU-1.ndpi                	| DICOM    	| no            	| JPEG        	| 113.465s        	| 127M      	|
| CMU-1.ndpi                	| DICOM    	| no            	| JPEG-2000   	| 640.609s        	| 1.2G      	|
| CMU-1.ndpi                	| DICOM    	| yes           	|             	| 16.811s         	| 191M      	|
| CMU-1.ndpi                	| OME-TIFF 	| no            	| JPEG        	| 111.965s        	| 127M      	|
| CMU-1.ndpi                	| OME-TIFF 	| no            	| JPEG-2000   	| 620.038s        	| 1.3G      	|
| CMU-1.ndpi                	| OME-TIFF 	| yes           	|             	| 6.852s          	| 191M      	|
| CMU-1_0_0.dcm             	| OME-TIFF 	| no            	| JPEG        	| 137.521s        	| 127M      	|
| CMU-1_0_0.dcm             	| OME-TIFF 	| no            	| JPEG-2000   	| 1150.285s       	| 1.3G      	|
| CMU-1_0_0.dcm (JPEG)      	| OME-TIFF 	| yes           	|             	| 6.654s          	| 127M      	|
| CMU-1_0_0.dcm (JPEG-2000) 	| OME-TIFF 	| yes           	|             	| 16.244s         	| 1.2G      	|

Overall all the conversion times and sizes  are consistent with the previous measurements made during the work on reading precompressed SVS and NDPI files. The speed-up associated with the `-precompressed` flag is comparable when writing OME-TIFF and the absolute values are typically smaller.

Trying to validate the binary data in a viewer, all converted images were found to look good except for the two OME-TIFF files created from the DICOM datasets without `-precompressed` where the tiles look to be incorrectly ordered

Screenshot 2024-08-13 at 13 52 42

Once the above is addressed, I would re-run the tests above for consistency and I would propose to carry a few additional functional tests including:
- converting to TIFF rather than OME-TIFF (small images)
- converting to other supported formats
- mixing `-precompressed` / `-compression` flags
- for TIFF/OME-TIFF, testing a few other compression flags

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
melissalinkert: `SVSReader` was the particular issue here; my concern was that swapping the return value of `reader.isInterleaved` is a potentially a lot more invasive. When that value changes, what `openBytes` returns has to change, which means things like pixels hashes change, etc.

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
melissalinkert: That was definitely the intent - `-precompressed` on its own with no `-compression` should work, and doesn't require knowledge of the input data.

melissalinkert:
> In the spirit of #3992 and #4181, used `CMU-1.svs` and `CMU-1.npdi` as the source files for conversion.
> Trying to validate the binary data in a viewer, all converted images were found to look good except for the two OME-TIFF files created from the DICOM datasets without `-precompressed` where the tiles look to be incorrectly ordered
> 
> Screenshot 2024-08-13 at 13 52 42

Currently investigating, but it looks like this issue is reproducible if the same DICOM -> OME-TIFF conversion is done with develop. It's still worth fixing here, but may not have been introduced by this PR.

melissalinkert:
I believe 29e8c05f22 fixes the scrambled tiles issue. As noted above, I could reproduce the problem both with this PR and with develop alone, using something like:

```
bfconvert -noflat -compression JPEG /path/to/cmu1_0_0.dcm cmu1.ome.tiff
```

However, specifying the tile sizes explicitly:

```
bfconvert -noflat -compression JPEG -tilex 512 -tiley 512 /path/to/cmu1_0_0.dcm cmu1.ome.tiff
```

no longer showed the problem. With 29e8c05f22, I would expect either command to result in a correct OME-TIFF.

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
melissalinkert: See 29e8c05f22.

It's probably worth a couple of tests, maybe with fake data, for things like ICS. Keep in mind most writers do not allow tile-wise conversion at all (e.g. https://github.com/ome/bioformats/blob/develop/components/formats-bsd/src/loci/formats/out/APNGWriter.java#L84), and DICOM/TIFF/OME-TIFF are the only ones that would support pyramids.

sbesson (commented):
Reproducing the tests from https://github.com/ome/bioformats/pull/4190#pullrequestreview-2234915792 with the last commit, the issue raised when converting from DICOM to OME-TIFF without precompression are now addressed 

Screenshot 2024-08-14 at 10 59 52

Timings and file sizes are consistent with the table of the last review

sbesson (approved):
In addition to the last review, performed a few additional conversion using different input/output formats (fake, ics, tiff) with different source compression and tile/compression options.

The first set of conversion used a [fake file](https://bio-formats.readthedocs.io/en/stable/developers/generating-test-images.html) with XY dimensions 2000x2000, 10 focal points and 2 channels. All the conversions using TIFF as the source format used one of the TIFF files generated from the fake files

| Source format 	| Source compression 	| Target format 	| Precompressed 	| Target tile size 	| Target compression 	| Conversion time 	| File size 	|
|---------------	|--------------------	|---------------	|---------------	|------------------	|--------------------	|----------------:	|-----------	|
| Fake          	| Uncompressed       	| ICS           	| no            	|                  	|                    	| 1.143           	| 197M      	|
| Fake          	| Uncompressed       	| ICS           	| no            	| 512x512          	|                    	| 1.105s          	| 197M      	|
| Fake          	| Uncompressed       	| ICS           	| yes           	|                  	|                    	| Failed          	|           	|
| Fake          	| Uncompressed       	| TIFF          	| no            	|                  	| Uncompressed       	| 2.558s          	| 130M      	|
| Fake          	| Uncompressed       	| TIFF          	| no            	|                  	| zlib               	| 5.001s          	| 18M       	|
| Fake          	| Uncompressed       	| TIFF          	| no            	| 512x512          	| Uncompressed       	| 3.222s          	| 81M       	|
| Fake          	| Uncompressed       	| TIFF          	| no            	| 512x512          	| zlib               	| 3.21s           	| 448K      	|
| Fake          	| Uncompressed       	| TIFF          	| no            	| 512x512          	| JPEG               	| 3.889s          	| 2.5M      	|
| Fake          	| Uncompressed       	| TIFF          	| no            	| 512x512          	| JPEG-2000          	| 7.178s          	| 448K      	|
| TIFF          	| zlib               	| ICS           	| no            	|                  	|                    	| 1.952s          	| 77M       	|
| TIFF          	| zlib               	| ICS           	| no            	| 512x512          	|                    	| 1.065s          	| 105M      	|
| TIFF          	| zlib               	| ICS           	| yes           	|                  	|                    	| 1.074s          	| 105M      	|
| TIFF          	| zlib               	| TIFF          	| no            	|                  	| Uncompressed       	| 2.436s          	| 77M       	|
| TIFF          	| zlib               	| TIFF          	| yes           	|                  	|                    	| failed          	|           	|
| TIFF          	| zlib               	| TIFF          	| no            	| 512x512          	| JPEG               	| 3.724s          	| 2.5M      	|
| TIFF          	| JPEG               	| TIFF          	| no            	|                  	| Uncompressed       	| 2.494s          	| 77M       	|
| TIFF          	| JPEG               	| TIFF          	| yes           	|                  	|                    	| failed          	|           	|
| TIFF          	| JPEG               	| TIFF          	| no            	| 512x512          	| zlib               	| 3.26s           	| 546K      	|

The conversion with a `-precompressed` flag failed with

```
java.lang.UnsupportedOperationException: Reader does not support pre-compressed tile access
```

which I believe is the expectation.

All generated files were validated visually with all planes/tiles being available for all coordinates.

From my side this is ready for inclusion in the next Bio-Formats release. Given the nature of the changes, this is definitely not amenable for a patch release but could be included in a release incrementing the minor or major version to communicate the backwards-compatible API additions.

In terms of documentation, I can see a few follow-up tasks including:

- update the `bfconvert` utility help to list the formats/codecs supporting pre-compressed tiles like we do for compression
- update the user/technical documentation especially https://bio-formats.readthedocs.io/en/latest/users/comlinetools/conversion.html to remove the requirement to specify the compression format and list the additional reader/writers/codecs supporting the `-precompressed` flag
In the mid-term, I wonder whether we'll want to maintain a table of format/codecs that support precompressed reading in reading/writing mode separately.

sbesson:
Merging to fix conflicts with #4164 and unblock the performance work.

PR #4198, opened 2024-06-18
SDT: fix NegativeArraySizeException when reading files recorded by SPCM64 9.90

cqian89:
I encountered a java.lang.NegativeArraySizeException when reading files recorded by the latest version (9.90) of Becker Hickl's SPCM64 software. Files recorded with older versions don't seem to have this issue.

On further investigation, it seems to be related to a value setupLength. This should be an unsigned short, but was read as a signed short previously. In the SDT files recorded by the latest SPCM64, this seems to have exceeded the range of a signed short but is still within range of an unsigned short.

This PR fixes the error by changing readShort to readUnsignedShort and the type of setupLength to int.

Test data here: https://doi.org/10.5281/zenodo.12094862

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/SDTInfo.java
+++ b/components/formats-gpl/src/loci/formats/in/SDTInfo.java
@@ -106,7 +106,7 @@ public class SDTInfo {
   public int setupOffs;
 
   /** Length of the setup data. */
-  public short setupLength;
+  public int setupLength;
 
   /** Offset of the first data block. */
   public int dataBlockOffs;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thanks, @cqian89. This change makes sense to me, so it will be included in tonight's CI builds to see if any of our existing SDT test files are affected.

Before merging this pull request, we will need to have a signed Contributor License Agreement. Please submit the CLA according to the instructions here: https://ome-contributing.readthedocs.io/en/latest/cla.html

cqian89:
Thank you. I just submitted the signed CLA as per the instructions.

melissalinkert (approved):

sbesson (approved):

PR #4214, opened 2024-07-22
Fix incorrect logical operators in Zeiss LSM and OME-XML readers

### LINKED ISSUE #4209: && needed instead of &
https://github.com/ome/bioformats/blob/6fa35f6deb652ac34e3bb2fba5b8b694481d3585/components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java#L1137

I think **& blue** should be replaced with **&& blue**

 if (red == 0 && green == 0 && blue == 0) {

### LINKED ISSUE #4212: May need to replace single | with double ||
https://github.com/ome/bioformats/blob/7dd002d53cef8e2a30b19da572e920f82f8ce903/components/formats-bsd/src/loci/formats/in/OMEXMLReader.java#L293

| c== null

should be?

|| c == null

sbesson:
Fixes #4209 
Fixes #4212


joshmoore:
:clap:

melissalinkert (approved):
No objections to merging as long as builds remain green tomorrow (just to be safe).

dgault (approved):
Good catch, changes look good to me

PR #4215, opened 2024-07-23
Fix DICOM export in ImageJ

melissalinkert:
See https://forum.image.sc/t/exporting-a-dicom-stack/99400/6

I was able to reproduce the exception in that thread using ImageJ + Bio-Formats 7.3.1 by selecting `File > Open Samples > Organ of Corti...`, then `Plugins > Bio-Formats > Bio-Formats Exporter` and saving to a file with the `.dcm` extension.

With this change, the same test should result in a converted `.dcm` file (with no exception) that can be re-opened.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -505,7 +505,7 @@ public class Exporter {
                     String lsid = MetadataTools.createLSID("Channel", 0, c);
                     store.setChannelID(lsid, 0, c);
                 }
-                store.setChannelSamplesPerPixel(new PositiveInteger(channels), 0, 0);
+                store.setChannelSamplesPerPixel(new PositiveInteger(channels), 0, c);
 
                 if (imp instanceof CompositeImage) {
                     luts[c] = ((CompositeImage) imp).getChannelLut(c + 1);
--- a/components/formats-bsd/src/loci/formats/out/DicomWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/DicomWriter.java
@@ -2082,7 +2082,12 @@ public class DicomWriter extends FormatWriter implements IExtraMetadataWriter {
    * proper physical units (e.g. mm), so need to be handled specially.
    */
   private Length fixUnits(Length size) {
-    if (size == null || size.unit() == UNITS.PIXEL || size.unit() == UNITS.REFERENCEFRAME) {
+    if (size == null) {
+      return null;
+    }
+    if (size.unit() == UNITS.PIXEL || size.unit() == UNITS.REFERENCEFRAME) {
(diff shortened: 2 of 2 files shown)


sbesson (commented):
Using the current release JARS (7.3.1) shipped in the Java-8 distribution, I was able to reproduce the exception described in the PR while exporting the sample file as a DICOM.

Using the JARs built from this branch and replacing them in my local distribution, the same sample file exports successfully using the exporter and can be reopened with the plugin.

Testing the export of a few other sample files, I came across another NPE using ``File > Open Samples > Cardio (RGB DICOM)` and selecting the Bio-Formats exporter to create a DICOM 

```
	at loci.formats.out.DicomWriter.setId(DicomWriter.java:1057)
	at loci.formats.FormatWriter.changeOutputFile(FormatWriter.java:123)
	at loci.plugins.out.Exporter.run(Exporter.java:815)
	at ij.plugin.filter.PlugInFilterRunner.processOneImage(PlugInFilterRunner.java:266)
	at ij.plugin.filter.PlugInFilterRunner.<init>(PlugInFilterRunner.java:114)
	at ij.IJ.runUserPlugIn(IJ.java:246)
	at ij.IJ.runPlugIn(IJ.java:210)
	at ij.IJ.runPlugIn(IJ.java:199)
	at loci.plugins.shortcut.ShortcutPanel$1.run(ShortcutPanel.java:189)
	Suppressed: java.lang.NullPointerException
		at loci.formats.out.DicomWriter.writeIFDs(DicomWriter.java:2008)
		at loci.formats.out.DicomWriter.close(DicomWriter.java:1402)
		at loci.plugins.out.Exporter.run(Exporter.java:343)
		... 7 more
```

Is that something we also want to fix as part of this PR?

melissalinkert:
> Testing the export of a few other sample files, I came across another NPE using ``File > Open Samples > Cardio (RGB DICOM)` and selecting the Bio-Formats exporter to create a DICOM

ab8234f870 is one way to handle, but open to other ideas.

sbesson:
> https://github.com/ome/bioformats/commit/ab8234f870221f6e56ddec8688a16de4d00f85e0 is one way to handle, but open to other ideas.

No objections to this approach from my side. The proposal is very similar to https://github.com/glencoesoftware/omero-ms-image-region/pull/99 which was addressing a very similar use case. In the same spirit, could the private method handle separately the `null` case from the unsupported length units case and log a warning level statement in the latter case to communicate that metadata is lost on conversion?

sbesson (approved):

sbesson:
Successfully tested by exporting a few ImageJ samples (`Cardio (RGB DICOM)`, `CT (16-bit DICOM)`, `Organ of Corti (4D stack)`) as DICOM and rereading them in Fiji modified with the JARs built from this PR.

Noticed that if all DICOM files are exported in the same folder, the initialization fails as I assume the reader tries to read multiple DICOM files in the same directory. But this is the current behavior and was not introduced by this PR

PR #4219, opened 2024-08-06
ImageJ exporter: save original metadata

### LINKED ISSUE #4201: Bug(?): IJ exported OME-TIFFs lack original metadata
image

Tischi recently asked about what he considered "missing/loss metadata". After some testing, we found the following:

| File          | Created with   | OME-XML | Metadata Pane |
|---------------|----------------|---------|---------------|
| xy.oir        | N/A            | rich    | rich          |
| xy.ij.tiff    | IJ save        | poor    | rich          |
| xy-1.ome.tiff | B-F exporter   | rich    | **poor** 👈🏽           |
| xy-2.ome.tiff | bfconvert      | rich    | rich          |
| xy-3.ome.tiff | ngff-converter | rich    | rich          |

When reading and then writing with B-F importer/exporter, the original metadata annotation does not get included in the OME-XML of the xy-1.ome.tiff.

cc: @tischi

melissalinkert:
Fixes #4201.

I tested with https://downloads.openmicroscopy.org/images/ND2/karl/sample_image.nd2 since it has lots of original metadata, but pretty much any other test data would work as well.

Without this change, open in ImageJ with `Plugins > Bio-Formats > Bio-Formats Importer` and the `Display metadata` box checked. Export to OME-TIFF with `Plugins > Bio-Formats > Bio-Formats Exporter`, and default options (no boxes checked). Open the exported OME-TIFF as with the original file, and note that most of the original metadata is no longer available.

With this change, the same test should show that the original metadata window is more or less the same for the original file and the exported OME-TIFF.

cc @joshmoore, @tischi

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -52,6 +52,7 @@ import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Hashtable;
 import java.util.List;
 import java.util.Vector;
 
@@ -60,6 +61,7 @@ import javax.swing.JOptionPane;
 import javax.swing.filechooser.FileFilter;
 
 import loci.common.DataTools;
+import loci.common.DebugTools;
 import loci.common.services.DependencyException;
 import loci.common.services.ServiceException;
 import loci.common.services.ServiceFactory;
@@ -538,6 +540,7 @@ public class Exporter {
             }
 
             Object info = imp.getProperty("Info");
+            Hashtable<String, Object> originalMetadata = new Hashtable<String, Object>();
             if (info != null) {
                 String imageInfo = info.toString();
                 if (imageInfo != null) {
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Confirmed that up to Bio-Formats 8, the OME-TIFF generated by the ImageJ Bio-Formats exporter lacks any original metadata.

With this change included, the same export workflow now generates an OME-TIFF with original metadata as can be confirmed both by introspecting the TIFF using `tiffinfo` as well as opening the exported OME-TIFF using the Bio-Formats Importer and selecting `Display metadata` option.

It is worth noting the checkboxes in the `Bio-Formats Import Options` window only affect the display of the metadata. In all cases, the original metadata is populated at reader initialization and exported to the format if applicable. I believe this is consistent with the expectation highlighted in the original issue i.e. the exporter should preserver all metadata but in case there is a use case for skipping the original metadata export (e.g. due to the size of the annotation), I assume it would need to be implemented as an additional export option.

In the meantime, I think including this in the upcoming 8.0.1 patch release of Bio-Formats makes complete sense.

joshmoore:
Thanks both!

Issue #4139, opened 2024-01-10
Imspector MSR: NumberFormatException when parsing timepoints


dgault:
This issue was raised on Imagesc thread https://forum.image.sc/t/msr-files-in-bio-formats/90582 and a sample file was provided at https://zenodo.org/records/10476252?token=eyJhbGciOiJIUzUxMiJ9.eyJpZCI6IjQ1ODRiNjkwLWExNTUtNGJlZS1hYjJhLTNjOTgzZWRiNmUwZCIsImRhdGEiOnt9LCJyYW5kb20iOiJhZjI1M2JjZDc4Y2Q5ZjE0MzhiMTZhOWQwMWRkYWQ2ZiJ9.Xz8eb6cxS-jWu5GvAnYEqkKlFzjLNoYYc3bI3mjtci7tinEdRZIvj8CjTyHqTNJj1waBVpaNJZvK8IvN0y3Y0g

Using the provided sample file the issue can be reproduced with Bio-Formats 7.1.0

The stack trace for the exception is as below:
```
java.lang.NumberFormatException: For input string: "2.8E-43"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at loci.formats.in.ImspectorReader.initFile(ImspectorReader.java:429)
	at loci.formats.FormatReader.setId(FormatReader.java:1480)
```

scuniff:
I was able to re-produce the issue with the test image in Fiji.

In Eclipse, I modified the code in https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/ImspectorReader.java

The changes are below.

The variable **valueType** was 9 during the error so I moved the **case 9** statement down to the section that does **readInt**().
It was doing **readFloat**() which I think causes the Exception later at line 429 as noted in stack trace in original message.

Line 316:
 ```
       switch (valueType) {
          case 10:
          case 11:
          case 12:
              key = value;
              value = String.valueOf(in.readInt() == 1);
              break;         case 2:
          case 3:          
//       case 9:                                    // moved to section below that does readInt() 
          case 13:
          case 14:
          case 17:
              key = value;
              value = String.valueOf(in.readFloat());
              break;
          case 0:
          case 1:
          case 4:
          case 5:
          case 6:
          case 8:
          case 9:                   // added case 9
             key = value;
[...]

PR #4233, opened 2024-09-09
Handle UnknownFormatException for batch processing multiple files.

tstoeter:
When sequentially processing many files the unhandled exception terminates the program and no further files will be processed that may still be readable. This commit catches the exception, provides a warning message and continues processing.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -1144,31 +1144,31 @@ public class ImageInfo {
         newArgs[idx] = scanner.nextLine();
         System.out.println("====% " + newArgs[idx]);
         try {
-        	 new ImageInfo().testRead(newArgs);
+          new ImageInfo().testRead(newArgs);
         }
         catch (FormatException e) {
-        	 LOGGER.error("Caught FormatException. " + e.getMessage());
-        	 exitCode = 1;
-        	 continue;
+          LOGGER.error("Caught FormatException. " + e.getMessage());
+          exitCode = 1;
+          continue;
         }
         catch (IOException e) {
-        	 LOGGER.error("Caught IOException. " + e.getMessage());
-        	 exitCode = 1;
-        	 continue;
+          LOGGER.error("Caught IOException. " + e.getMessage());
+          exitCode = 1;
+          continue;
         }
         catch (ServiceException e) {
-        	 LOGGER.error("Caught ServiceException. " + e.getMessage());
-        	 exitCode = 1;
-        	 continue;
+          LOGGER.error("Caught ServiceException. " + e.getMessage());
+          exitCode = 1;
+          continue;
         }
       }
       scanner.close();
     }
     else {
       if (!new ImageInfo().testRead(args)) System.exit(1);
-   }
+    }
 
-   System.exit(exitCode);
+    System.exit(exitCode);
   }
 }
 
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -1146,18 +1146,8 @@ public class ImageInfo {
         try {
           new ImageInfo().testRead(newArgs);
         }
-        catch (FormatException e) {
-          LOGGER.error("Caught FormatException. " + e.getMessage());
-          exitCode = 1;
-          continue;
-        }
-        catch (IOException e) {
-          LOGGER.error("Caught IOException. " + e.getMessage());
-          exitCode = 1;
-          continue;
-        }
-        catch (ServiceException e) {
-          LOGGER.error("Caught ServiceException. " + e.getMessage());
+        catch (Exception e) {
+	  LOGGER.error("Caught " + e.getClass().getSimpleName(), e);
           exitCode = 1;
           continue;
         }
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -77,6 +77,7 @@ import loci.formats.meta.MetadataRetrieve;
 import loci.formats.meta.MetadataStore;
 import loci.formats.services.OMEXMLService;
 import loci.formats.services.OMEXMLServiceImpl;
+import loci.formats.UnknownFormatException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -1141,8 +1142,14 @@ public class ImageInfo {
 
       while (scanner.hasNext()) {
         newArgs[idx] = scanner.nextLine();
-	System.out.println("====% " + newArgs[idx]);
-        if (!new ImageInfo().testRead(newArgs)) System.exit(1);
+        System.out.println("====% " + newArgs[idx]);
+        try {
+        	new ImageInfo().testRead(newArgs);
+        }
+        catch (UnknownFormatException e) {
+			LOGGER.warn("Unknown file format: " + newArgs[idx]);
+        	continue;
+        }
       }
       scanner.close();
     }
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -77,7 +77,6 @@ import loci.formats.meta.MetadataRetrieve;
 import loci.formats.meta.MetadataStore;
 import loci.formats.services.OMEXMLService;
 import loci.formats.services.OMEXMLServiceImpl;
-import loci.formats.UnknownFormatException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
```diff
@@ -1141,8 +1142,14 @@ public static void main(String[] args) throws Exception {
...
       while (scanner.hasNext()) {
         newArgs[idx] = scanner.nextLine();
-	System.out.println("====% " + newArgs[idx]);
-        if (!new ImageInfo().testRead(newArgs)) System.exit(1);
+        System.out.println("====% " + newArgs[idx]);
+        try {
+        	new ImageInfo().testRead(newArgs);
+        }
+        catch (UnknownFormatException e) {
+			LOGGER.warn("Unknown file format: " + newArgs[idx]);
```
sbesson: `ImageInfo.testRead` can throw three types of exception:
- `FormatException` (of which `UnknowFormatException` is a subclass of)
- `IOException`
- `ServiceException`

What is the rationale for handling `UnknownFormatException` differently from the other exceptions?

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
```diff
@@ -1141,8 +1142,14 @@ public static void main(String[] args) throws Exception {
...
+        	continue;
```
sbesson: What should the return code be if any of the processed files fails? With the current implementation it will be zero

sbesson (commented):
Thanks @tstoeter for the follow-up. A few inline comments to try to better understand the expected behavior.

As we are still a few weeks away from the next major release of Bio-Formats, there is still time to refine the expectation of the batch functionality and document it appropriately in https://bio-formats.readthedocs.io/en/latest/users/index.html#command-line-tools. However, note that post-release, this will be treated like the rest of the API and any backwards-incompatible change will require an increment of the major version number.

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
tstoeter: Thanks a lot @sbesson for your code review and for pointing out the other possible exceptions. I intended to do the least changes to the current behaviour. I can add exception handling for the other cases, too, i.e. to handle general FormatException and IOException. Could you please tell me, in which case a ServiceException would occur?

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
tstoeter: Very good point, thank you. I would suggest a non-zero return code, then. Is there any preference?

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
sbesson: You can read more about `ServiceException` in https://bio-formats.readthedocs.io/en/latest/developers/service.html. Typically this would happen when a dependency is missing.

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
sbesson: I definitely see the value of returning a non-zero return code if any input file fails. This would give an indication to the called that the output of the utility should be reviewed for exceptions.

components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
```diff
@@ -1144,19 +1144,31 @@ public static void main(String[] args) throws Exception {
         newArgs[idx] = scanner.nextLine();
         System.out.println("====% " + newArgs[idx]);
         try {
-        	new ImageInfo().testRead(newArgs);
+        	 new ImageInfo().testRead(newArgs);
```
sbesson: Stylistic, can these be updated to use an indentation of two spaces - https://bio-formats.readthedocs.io/en/latest/developers/code-formatting.html#java

sbesson (approved):
Tested the last commit with a mixture of valid and invalid input files

```
sbesson@Sebastiens-MacBook-Pro-3 bioformats % cat files | ./tools/showinf -nopix - || echo "Completed with failures"
====% /tmp/test.ini
Initializing reader
Caught IOException. /tmp/test.ini (No such file or directory)
====% /tmp/foo
File has length 0 and may be corrupt
Caught FormatException. Unknown file format: /tmp/foo
====% /tmp/test.fake
Initializing reader
FakeReader initializing /tmp/test.fake
Initialization took 0.039s

Reading core metadata
filename = /tmp/test.fake
Series count = 1
Series #0 :
	Image count = 1
	RGB = false (1) 
	Interleaved = false
	Indexed = false (true color)
	Width = 512
	Height = 512
	SizeZ = 1
	SizeT = 1
	SizeC = 1
	Tile size = 512 x 512
	Thumbnail size = 128 x 128
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata

Reading metadata
Completed with failures
```

Happy to get the proposed changes included in Bio-Formats 8.0.0

melissalinkert (changes requested):
This is definitely an improvement over how checked exceptions were handled previously, but  unchecked exceptions (e.g. `NullPointerException`, `NumberFormatException`) are still not being caught and will stop processing. This now means two different sets of behavior depending upon whether a checked or unchecked exception was thrown.

As an example, take [test.msr](https://zenodo.org/records/10476252?token=eyJhbGciOiJIUzUxMiJ9.eyJpZCI6IjQ1ODRiNjkwLWExNTUtNGJlZS1hYjJhLTNjOTgzZWRiNmUwZCIsImRhdGEiOnt9LCJyYW5kb20iOiJhZjI1M2JjZDc4Y2Q5ZjE0MzhiMTZhOWQwMWRkYWQ2ZiJ9.Xz8eb6cxS-jWu5GvAnYEqkKlFzjLNoYYc3bI3mjtci7tinEdRZIvj8CjTyHqTNJj1waBVpaNJZvK8IvN0y3Y0g) (see #4139), which is currently expected to throw `NumberFormatException` during initialization, plus artificial files:

```
$ touch abc.fake
$ touch xyz.fake
$ ls -gG
total 268952
-rwxrwxrwx 1         0 Sep 16 11:58 abc.fake
-rwxrwxrwx 1 275404088 Sep 16 11:51 test.msr
-rwxrwxrwx 1         0 Sep 16 11:58 xyz.fake
```

Then attempt to process all 3 files:

```
$ ls | showinf -nopix - || echo "Completed with failures"
====% abc.fake
Checking file format [Simulated data]
Initializing reader
FakeReader initializing abc.fake
Initialization took 0.128s

Reading core metadata
filename = abc.fake
Used files = [/home/melissa/data/gh-4233-test/abc.fake]
Series count = 1
Series #0 :
	Image count = 1
	RGB = false (1) 
	Interleaved = false
	Indexed = false (true color)
	Width = 512
	Height = 512
	SizeZ = 1
	SizeT = 1
	SizeC = 1
	Tile size = 512 x 512
	Thumbnail size = 128 x 128
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata

Reading metadata
====% test.msr
Checking file format [Lavision Imspector]
Initializing reader
ImspectorReader initializing test.msr
Exception in thread "main" java.lang.NumberFormatException: For input string: "2.8E-43"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at loci.formats.in.ImspectorReader.initFile(ImspectorReader.java:429)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1147)
Completed with failures
```

So `xyz.fake` is never processed in this case.

Separately, since the `catch` blocks are pretty much identical, I'd prefer to see them unified. The error logging could be something like `LOGGER.error("Caught " + e.getClass().getSimpleName(), e)`. Passing the thrown exception to `LOGGER.error(...)` means that the stack trace is preserved if needed for debugging. I'd also be OK with logging just the message at `ERROR`, and the stack trace at `DEBUG`, something like:

```
LOGGER.error("Caught {}. {}", e.getClass().getSimpleName(), e.getMessage());
LOGGER.debug("Failed to process " + newArgs[idx], e);
```

tstoeter:
Thank you very much @sbesson and @melissalinkert for your thorough tests and reviews. I'll work on it!

tstoeter:
@melissalinkert I have implemented the requested changes and made sure your test case (with the same test.msr file) now passes.

melissalinkert (approved):
Thanks, @tstoeter. This looks good to me now, and the latest changes have been included in a passing build: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/192/

PR #4245, opened 2024-10-07
Upgrade cdm-core (netcdf Java) to 5.6.0

melissalinkert:
See https://github.com/ome/bioformats/pull/4217#issuecomment-2388204999

See also netcdf-java release notes: https://github.com/Unidata/netcdf-java/releases.

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -66,7 +66,7 @@
     <dependency>
       <groupId>edu.ucar</groupId>
       <artifactId>cdm-core</artifactId>
-      <version>5.3.3</version>
+      <version>5.6.0</version>
       <exclusions>
         <exclusion>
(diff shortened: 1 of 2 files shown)


sbesson (approved):
The version bump looks sensible, and from a testing perspective, the nightly repository CI builds are passing with this included.  From my side, this is ready for inclusion in Bio-Formats 8.0.0. As noted in https://github.com/ome/bioformats/pull/4217#issuecomment-2388204999, this might also assist future improvements to the ImarisIMSReader.
The `ome-common-java` component were also adjusted and released so that the correct versions of the low-level dependencies are resolved independently of the build system. This should allow this to be safely included  in the following OMERO release. I'll leave @jburel to review this on his return.

jburel (approved):
All good. Recent failures in the jobs are due the migration and url not available from new VM.

PR #4248, opened 2024-10-21
Operetta reader: don't add directories to the used files list

melissalinkert:
Backported from a private PR.

Without this change, and with an extra directory in the Operetta/Phenix plates hierarchy like this:

```
/path/to/plate_name/images/index.xml
/path/to/plate_name/eval/ID/file.csv
```

`showinf` should report the directory `/path/to/plate_name/eval/ID` on the used files list, but not any of the files it contains.
Including directories on the used file list causes problems when importing to OMERO.

This change should remove any directories from the used files list, but will not recursively look for files in any unexpected directories - so `/path/to/plate_name/eval/ID/file.csv` would still not appear on the list.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -350,10 +350,16 @@ public class OperettaReader extends FormatReader {
           LOGGER.trace("Found folder {}", folder);
           if (!checkSuffix(folder, "tiff"))
           {
-            String metadataFile = new Location(path, folder).getAbsolutePath();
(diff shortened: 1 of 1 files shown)


sbesson (commented):
At the code level, the proposed changes add some defensive checks to the reader which should handle subdirectories similarly to what was done for the CV7000 reader https://github.com/ome/bioformats/pull/3982.

Would it be possible to construct and configure a synthetic plate reproducing the issue and add it to the curated OME QA repository?

melissalinkert:
See `curated/perkinelmer-operetta/samples/extra-directories`, as noted in https://github.com/openmicroscopy/data_repo_config/pull/625.

sbesson (approved):

PR #4249, opened 2024-10-24
Imaris Reader: support for LZ4 compression and performance improvements

marcobitplane:
Hello,

this pull request adds support for LZ4 compressed ims files and modifies ImarisHDFReader to avoid multiple reads of the same 3D chunks. See also #4217.

LZ4 support is added using NetCDF-Java's ucar.nc2.filter package, which provides a mechanism for user-supplied filters as described [here](https://docs.unidata.ucar.edu/netcdf-java/5.5/userguide/reading_zarr.html). 

Regarding performance, ImarisHDFReader is modified to have a caching mechanism that reads a stack of planes (as many planes as the chunk z-size) from all channels into a buffer, which only needs to be updated after all data in it has been read. If the size of the buffer would exceed 1GB, the reader falls back to reading the requested plane only. The exact performance improvement will depend on the details of the dataset: in our testing, for 3D datasets the new reader can be over an order of magnitude faster than the existing implementation.


melissalinkert (commented):
Thanks, @marcobitplane, this definitely looks like a reasonable approach.

Two higher-level questions before we proceed with more thorough review and testing:

- `lz4-java` is added as a dependency, but the existing dependency on [aircompressor](https://github.com/airlift/aircompressor) via [ome-codecs](https://github.com/ome/ome-codecs/blob/master/pom.xml#L111) should already support LZ4. Is there a reason to use `lz4-java` specifically, or could `aircompressor` be used instead?
- The new `LZ4Filter` uses the `ucar.nc2.filter` package; if at all possible, I'd prefer to see this in the existing Bio-Formats package structure (`loci.formats.*`) rather than `ucar.nc2.filter`. A brief read of https://docs.unidata.ucar.edu/netcdf-java/5.5/userguide/reading_zarr.html#implementing-a-filter suggests the package name of the custom filter itself doesn't matter, as long as it extends/implements the correct class/interface. Could you please either change the package name, or explain a bit why putting `LZ4Filter` in `ucar.nc2.filter` is necessary?

Finally, in order to merge this we would need a signed [Contributor License Agreement](https://ome-contributing.readthedocs.io/en/latest/cla.html). That's not urgent at this point, but it would be good to have sooner rather than later.

marcobitplane:
Thank you @melissalinkert!

I modified the PR to use aircompressor instead of lz4-java and moved LZ4Filter to loci/formats/filter, let me know if that is not ideal or there's more to change there. I'll send the CLA today.

melissalinkert:
Adding to tonight's build, so we should know in the morning if there are any test failures on existing data.

@marcobitplane : maybe I missed this, but I don't see here or in #4217 a reference to lz4 test files. Do you have any Imaris HDF files with lz4 compression that we can use to test this?

marcobitplane:
Thank you @melissalinkert, [here](https://github.com/user-attachments/files/17713197/demoImagesLz4.zip) are two lz4-compressed cropped versions of Imaris demo images (retina and CellDemoMembrane3D). Feel free to let me know if you would prefer to have more images or have them shared in a different way.

sbesson:
@marcobitplane thanks for sharing some sample data. Our preferred way to collect such samples is to upload them to the [Bio-Formats Zenodo community collection](https://zenodo.org/communities/bio-formats/records). For the OME team and the community, has the benefit of unambiguously assigning a license for the distribution and re-usage of these samples. In addition, it is possible to add additional samples to the upload and create new versions of the dataset as necessary during the review process.

I had a quick go at testing these with the proposed changed and the Bio-Formats command-line utility. With the just released 8.0.1 command-line tools, Bio-Formats fails with 


```
sbesson@Sebastiens-MacBook-Pro-3 bioformats % ~/Downloads/bftools/showinf -nopix ~/Downloads/demoImagesLz4/CellDemoMembrane3Dlz4.ims
Checking file format [Bitplane Imaris 5.5 (HDF)]
Initializing reader
ImarisHDFReader initializing /Users/sbesson/Downloads/demoImagesLz4/CellDemoMembrane3Dlz4.ims
Exception in thread "main" java.lang.RuntimeException: Unknown filter type=32004
	at ucar.nc2.iosp.hdf5.H5tiledLayoutBB$DataChunk.getByteBuffer(H5tiledLayoutBB.java:227)
	at ucar.nc2.iosp.LayoutBBTiled.hasNext(LayoutBBTiled.java:101)
	at ucar.nc2.iosp.hdf5.H5tiledLayoutBB.hasNext(H5tiledLayoutBB.java:125)
	at ucar.nc2.iosp.IospHelper.readData(IospHelper.java:332)
	at ucar.nc2.iosp.IospHelper.readDataFill(IospHelper.java:292)
	at ucar.nc2.iosp.hdf5.H5iosp.readData(H5iosp.java:161)
	at ucar.nc2.iosp.hdf5.H5iosp.readData(H5iosp.java:134)
	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2122)
	at ucar.nc2.Variable.reallyRead(Variable.java:817)
	at ucar.nc2.Variable._read(Variable.java:768)
	at ucar.nc2.Variable.read(Variable.java:600)
	at ucar.nc2.Variable.read(Variable.java:546)
	at loci.formats.services.NetCDFServiceImpl.getArray(NetCDFServiceImpl.java:172)
	at loci.formats.in.ImarisHDFReader.getImageData(ImarisHDFReader.java:473)
	at loci.formats.in.ImarisHDFReader.initFile(ImarisHDFReader.java:308)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1159)
```

With a local build of Bio-Formats using the HEAD of this PR, I have

```
sbesson@Sebastiens-MacBook-Pro-3 bioformats % ./tools/showinf -nopix  -nopix ~/Downloads/demoImagesLz4/CellDemoMembrane3Dlz4.ims 
Checking file format [Bitplane Imaris 5.5 (HDF)]
Initializing reader
ImarisHDFReader initializing /Users/sbesson/Downloads/demoImagesLz4/CellDemoMembrane3Dlz4.ims
Failure during the reader initialization
```

adding the `-debug` flag

```
sbesson@Sebastiens-MacBook-Pro-3 bioformats % ./tools/showinf -nopix  -nopix ~/Downloads/demoImagesLz4/CellDemoMembrane3Dlz4.ims -debug  
...
loci.formats.FormatException: loci.common.services.ServiceException: java.io.IOException: ucar.nc2.filter.UnknownFilterException: Unknown filter: no filter found with id 32004
	at loci.formats.in.ImarisHDFReader.getImageData(ImarisHDFReader.java:518)
	at loci.formats.in.ImarisHDFReader.initFile(ImarisHDFReader.java:332)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1158)
Caused by: loci.common.services.ServiceException: java.io.IOException: ucar.nc2.filter.UnknownFilterException: Unknown filter: no filter found with id 32004
	at loci.formats.services.NetCDFServiceImpl.getArray(NetCDFServiceImpl.java:183)
	at loci.formats.in.ImarisHDFReader.getImageData(ImarisHDFReader.java:514)
	... 6 common frames omitted
Caused by: java.io.IOException: ucar.nc2.filter.UnknownFilterException: Unknown filter: no filter found with id 32004
	at ucar.nc2.internal.iosp.hdf5.H5tiledLayoutBB.<init>(H5tiledLayoutBB.java:90)
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:226)
	at ucar.nc2.internal.iosp.hdf5.H5iospNew.readData(H5iospNew.java:204)
	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2122)
	at ucar.nc2.Variable.reallyRead(Variable.java:817)
	at ucar.nc2.Variable._read(Variable.java:768)
	at ucar.nc2.Variable.read(Variable.java:600)
	at ucar.nc2.Variable.read(Variable.java:546)
	at loci.formats.services.NetCDFServiceImpl.getArray(NetCDFServiceImpl.java:175)
	... 7 common frames omitted
Caused by: ucar.nc2.filter.UnknownFilterException: Unknown filter: no filter found with id 32004
	at ucar.nc2.filter.Filters.getFilter(Filters.java:80)
	at ucar.nc2.internal.iosp.hdf5.H5tiledLayoutBB.<init>(H5tiledLayoutBB.java:88)
	... 15 common frames omitted
```

The other sample `retinalz4.ims` fails with the same stack trace  so there still seems to be some issue with the registration of the custom LZ4 filter.

marcobitplane:
Thank you @sbesson for the clarification, I'll check with our team if it is ok to share the images on Zenodo for usage by the community. I think it should be fine since we already distribute retina.ims and CellDemoMembrane3D.ims with our free ImarisViewer, but they have been modified in the version I shared with you so I'd first like to make sure everything is fine there.

Regarding the error with showinf, apologies, I have been testing the changes to the reader in Fiji (by replacing the formats-gpl jar) and have no issues there with the filter, but I indeed have the same error with the command line tool. I will investigate that and hope to understand the reason for this as soon as possible.

marcobitplane:
The issue with showinf was a conflict between META-INF/services/ucar.nc2.filter.FilterProvider files. Bioformats_package.jar bundles many things including the NetCDF-Java library, which contains its own META-INF/services/ucar.nc2.filter.FilterProvider file providing the default filters like deflate and shuffle. Now we would also have a META-INF/services/ucar.nc2.filter.FilterProvider for the lz4 filter coming from formats-gpl, which creates a conflict since there can be only one file with the same path and name in a jar. Merging the two files seems to me a good solution here, and I read [here](https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/using-container-descriptor-handlers.html) one way to do it using the maven assembly plugin and Descriptor Handlers. I already committed a fix for the command line issue, let me know if you see problems with this approach.

melissalinkert:
With 2c4dfbd, I still see the same error as in https://github.com/ome/bioformats/pull/4249#issuecomment-2476969617. The `META-INF/services/ucar.nc2.filter.FilterProvider` included in `bioformats_package.jar` looks like this:

```
$ cat META-INF/services/ucar.nc2.filter.FilterProvider 
ucar.nc2.filter.Blosc$Provider
ucar.nc2.filter.Deflate$Provider
ucar.nc2.filter.Checksum32$Fletcher32Provider
ucar.nc2.filter.Checksum32$Adler32Provider
ucar.nc2.filter.Checksum32$CRC32Provider
ucar.nc2.filter.ScaleOffset$Provider
ucar.nc2.filter.Shuffle$Provider
```

marcobitplane:
I tried a fresh build of Bio-Formats using the HEAD of this PR and I get:

```
% cat META-INF/services/ucar.nc2.filter.FilterProvider 
loci.formats.filter.LZ4Filter$LZ4FilterProvider
ucar.nc2.filter.Blosc$Provider
ucar.nc2.filter.Deflate$Provider
ucar.nc2.filter.Checksum32$Fletcher32Provider
ucar.nc2.filter.Checksum32$Adler32Provider
ucar.nc2.filter.Checksum32$CRC32Provider
ucar.nc2.filter.ScaleOffset$Provider
ucar.nc2.filter.Shuffle$Provider
```

I'm now thinking, I built Bio-Formats with maven for this test, maybe @melissalinkert your build of `bioformats_package.jar` was an Ant build? I will investigate this first thing tomorrow, if so I'll research a similar merging mechanism for Ant.

melissalinkert:
> I'm now thinking, I built Bio-Formats with maven for this test, maybe @melissalinkert your build of bioformats_package.jar was an Ant build? I will investigate this first thing tomorrow, if so I'll research a similar merging mechanism for Ant.

Yes, that's correct, I used `ant clean jars tools`.

marcobitplane:
Thank you, I committed a fix for the ant build, let me know if you encounter any other issues.

The two lz4-compressed ims demo images have also been uploaded to Zenodo.

components/formats-gpl/pom.xml
```diff
@@ -120,6 +120,11 @@
       <artifactId>kryo</artifactId>
       <version>${kryo.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.airlift</groupId>
+      <artifactId>aircompressor</artifactId>
+      <version>0.27</version>
+    </dependency>
```
melissalinkert, line 127: I was initially going to suggest removing this and updating ome-codecs instead, but realized we don't have a released version of ome-codecs that includes https://github.com/ome/ome-codecs/pull/39.

In the context of this PR, I think I'm OK with leaving this dependency as-is. Separately, though, and before 8.1.0, we'll need to get ome-codecs released and updated so that we don't have two different versions of aircompressor.

components/formats-gpl/src/loci/formats/filter/LZ4Filter.java
```diff
@@ -0,0 +1,75 @@
+package loci.formats.filter;
```
melissalinkert, line 26: This will need the copyright statement that is at the top of all other files in `components/formats-gpl/src/loci/formats/`.

components/formats-gpl/src/loci/formats/filter/LZ4Filter.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public int getId() {
+    return id;
+  }
+
+  @Override
+  public byte[] encode(byte[] dataIn) throws IOException {
+    // only decoding implementation needed for reading
+    byte[] dataOut = dataIn;
```
melissalinkert: Slight preference for throwing `UnsupportedOperationException` here instead of returning the input array, so that it's very clear this isn't supported.

components/formats-gpl/src/loci/formats/services/NetCDFServiceImpl.java
```diff
@@ -45,13 +45,16 @@
...
 import ucar.nc2.NetcdfFile;
+import ucar.nc2.NetcdfFiles;
 import ucar.nc2.Variable;
 
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.KryoSerializable;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
 
+import com.google.common.collect.ImmutableList;
```
melissalinkert: It looks like this import is unnecessary?

melissalinkert (changes requested):
Thanks, @marcobitplane. I can now run `showinf` on the test files after building `bioformats_package.jar` with Ant, and see images displayed without error.

I have re-included this for automated testing, together with the new public sample files uploaded to Zenodo. I don't expect any test failures at this point, but we'll know for sure tomorrow.

The few in-line comments I have are fairly minor, but should be addressed before we merge.

marcobitplane:
Thank you again for the help and feedback, I added a few commits to address the requested changes.

melissalinkert (approved):
Latest changes look good to me, and most recent build is passing: https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/264/

components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
```diff
@@ -467,14 +490,92 @@ private Object getImageData(int no, int x, int y, int width, int height)
...
+        if (zct[2] != lastT) {
+          lastT = zct[2];
+        }
+
+        int blockNumber = zct[0] / blockSizeZPerResolution[resolutionIndex];
+        int[] dims = new int[] {blockSizeZPerResolution[resolutionIndex], getSizeY(), getSizeX()};
+        int[] idcs = new int[] {blockNumber * blockSizeZPerResolution[resolutionIndex], 0, 0};
+        try {
+          String path;
+          for (int ch = 0; ch < getSizeC(); ch++) {
```
sbesson, line 512: While I understand the rationale for caching an entire block for a given channel given the way the data is stored, my understanding is that different channels are stored separately.

While there are scenarios where preloading all channels makes sense, would this add some unnecessary overhead in some others e.g. when only one channel is requested or in the case of a large number of channels where only a few are rendered?

sbesson (commented):
Thanks @marcobitplane for sharing the samples using the Zenodo community (https://zenodo.org/records/14197622). We have copied them to our curated QA repository so that they can be integrated in the nightly CI repository tests. As indicated by @melissalinkert above, these have been passing for the last execution under all Java LTS versions.

Thanks for the changes to the Ant build infrastructure. I don't know off-hand of a better solution that to manually unzip and combine as you have done. Confirmed that the new filter is now included in the JAR bundle created either by Maven or Ant

```
sbesson@Sebastiens-MacBook-Pro-3 bioformats % unzip -qc artifacts/bioformats_package.jar META-INF/services/ucar.nc2.filter.FilterProvider
ucar.nc2.filter.Blosc$Provider
ucar.nc2.filter.Deflate$Provider
ucar.nc2.filter.Checksum32$Fletcher32Provider
ucar.nc2.filter.Checksum32$Adler32Provider
ucar.nc2.filter.Checksum32$CRC32Provider
ucar.nc2.filter.ScaleOffset$Provider
ucar.nc2.filter.Shuffle$Provider
loci.formats.filter.LZ4Filter$LZ4FilterProvider
sbesson@Sebastiens-MacBook-Pro-3 bioformats % unzip -qc  components/bundles/bioformats_package/target/bioformats_package-8.1.0-SNAPSHOT.jar  META-INF/services/ucar.nc2.filter.FilterProvider
loci.formats.filter.LZ4Filter$LZ4FilterProvider
ucar.nc2.filter.Blosc$Provider
ucar.nc2.filter.Deflate$Provider
ucar.nc2.filter.Checksum32$Fletcher32Provider
ucar.nc2.filter.Checksum32$Adler32Provider
ucar.nc2.filter.Checksum32$CRC32Provider
ucar.nc2.filter.ScaleOffset$Provider
ucar.nc2.filter.Shuffle$Provider
```

On the functional testing of the LZ4 compression, I confirm that the two sample LZ4 compressed Imaris files cannot be opened with Bio-Formats 8.0.1 with a `RuntimeException` due to an unknown filter. With this PR included both samples can be initialized using the command-line utility and the binary data can be read and compared to the ground truth as shown in Imaris Viewer.

Do you have specific samples and testing scenarios that would allow us to test the performance/caching improvements?

components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
marcobitplane: Different channels are indeed stored separately. 

If, when reading an image, planes were requested sequentially from a single channel (therefore moving along z before moving along channels) then the buffer could simply hold a stack of planes from that single channel. However, in my testing i saw that Bio-Formats (or Fiji?) requests the same image plane for all channels before moving to the next plane (moving along channels before moving along z). If the buffer just cached a stack from a single channel, the buffer would need to be updated every time, making it useless.

Let me know if I misunderstood your comment

marcobitplane:
Thank you @sbesson! 

With any 3D dataset you should be able to see a significant performance boost, for example an image of a few GBs should be read in a few seconds instead of a few minutes. How much faster depends on the chunk z size. For smaller images it might not be as obvious, I can look to provide an example of a larger ims file if you don't have one in the QA repository, just let me know.

components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
sbesson: The order in which planes are being retrieved including the index of the channels to be retrieved is fundamentally the decision of the caller. In the case of Fiji, I assume a single reader will indeed iterate through all the active channels for the given timepoint/Z-section before moving onto the next ones.

Other applications might make different decisions or even use concurrent requests in which case the benefit of caching all the channels might be limited. We will need to assess the cost/benefit of the caching logic across these scenarios as part of the testing.

sbesson:
>  For smaller images it might not be as obvious, I can look to provide an example of a larger ims file if you don't have one in the QA repository, just let me know.

If you could share some sample IMS files that you have used for validating the cache improvements, it would be really useful to review these changes.

marcobitplane:
I uploaded an example dataset to Zenodo (I created a new submission and set the access to restricted since this is not one of our demo images and I would again have to ensure it can be uploaded to a public repository). It is a typical 200MB, 3D image with two channels which should be read about 10 times faster with the caching mechanism.

marcobitplane:
The new dataset on Zenodo is now open access.

sbesson (commented):
The latest uploaded sample image, now available at https://zenodo.org/records/14235726, was used for testing the caching improvements. Measurements were made using the command-line tools using the different options available for reading different ranges of data.

Plane read times were compared between Bio-Formats built with this PR and Bio-Formats 8.0.1 using the average/standard deviation computed from three consecutive executions of the `showinf` command. In all scenarios, the reader `setId` time was unmodified by this PR. The output of the command-line utility e.g. `2.825s elapsed (14.413265ms per plane)` was used as the source of the times.

|                                                                    	| Bio-Formats 8.0.1 	| This PR           	|
|--------------------------------------------------------------------	|-------------------	|-------------------	|
| all planes / first resolution (no options)                         	| 17.74s +/- 0.21s  	| 3.41s +/- 0.08s   	|
| 1/4 of all planes / first resolution (-crop 512,512,512,512)       	| 6.95s +/ 0.02s    	| 2.84s +/- 0.03s   	|
| first plane / first resolution (-range 0 0)                        	| 0.289s +/- 0.003s 	| 0.172s +/- 0.001s 	|
| first plane / second resolution (-noflat -resolution 1 -range 0 0) 	| 0.200s +/- 0.001s 	| 0.195s +/- 0.003s 	|
| 51st plane / first resolution (-range 50 50)                       	| 0.264s +/- 0.001s 	| 0.341s +/- 0.007s 	|
| 101st plane  / first resolution (-range 100 100)                   	| 0.299s +/- 0.003s 	| 0.182s +/- 0.004s 	|
| last plane / first resolution (-range 195 195)                     	| 0.221s +/- 0.005s 	| 0.279s +/- 0.001s 	|

As shown in the two first rows, the benefit of the caching is most visible in the scenario where multiple / all planes are being read using the same initialized reader. In the most advantageous scenario, each block will only be read once and the internal cache re-used for all subsequent reads. In practice, this should be the behavior of the Fiji/ImageJ Bio-Formats plugin when using the `Hyperstack` view without the `Use virtual stack` option selected since the stack order is `XYCZT`. 

For individual planes, the read times with this PR included are also consistent with the measurements using the latest release of Bio-Formats taking ~200ms per plane. It is worth noting the table above show an increase in the read times variance with this PR included. Although this might result from the different way the underlying `netcdf` library is used, my proposal would be to follow-up with a more systematic measurement of the `openBytes` time for every plane index of the sample dataset.

melissalinkert:
Thanks, @sbesson. Overall that looks like a positive result.

> my proposal would be to follow-up with a more systematic measurement of the `openBytes` time for every plane index of the sample dataset

With this bit of test code:

```
$ cat ../PixTest.java 
import loci.common.DebugTools;
import loci.formats.FormatTools;
import loci.formats.in.ImarisHDFReader;

public class PixTest {
    DebugTools.enableLogging("WARN");
    try (ImarisHDFReader reader = new ImarisHDFReader()) {
      reader.setId(args[0]);

      byte[] buf = new byte[FormatTools.getPlaneSize(reader)];

      for (int plane=0; plane<reader.getImageCount(); plane++) {
        long t0 = System.currentTimeMillis();
        reader.openBytes(plane, buf);
        long t1 = System.currentTimeMillis();
        System.out.println("plane #" + plane + " time: " + (t1 - t0) + " ms");
      }
    }
  }
}
```

I see a dramatic improvement in most planes with this change compared to develop, but every 16th plane is worse with this change. The total time across all planes is still better with this change. Output of test code (3 runs each of merge build and develop), plus combined CSV of all plane times: [perf-results.tar.gz](https://github.com/user-attachments/files/17984404/perf-results.tar.gz)

Every 16th plane being worse matches up with the cache being refreshed; `blockSizeZPerResolution[0]` appears to be 16. I think the increase in variance makes sense in that context? That may also explain the worse behavior when reading a single plane that isn't the first one - since `getImageData(0, ...)` gets called by `initFile`, reading with `-range 50 50` or similar means a full cache refresh in the `openBytes` call.

sbesson:
Thanks @melissalinkert, everything makes sense and especially

> That may also explain the worse behavior when reading a single plane that isn't the first one - since getImageData(0, ...) gets called by initFile, reading with -range 50 50 or similar means a full cache refresh in the openBytes call.

which I had missed and explains a lot of the trends above.

I used a slightly modified version of your example script https://github.com/ome/bioformats/pull/4249#issuecomment-2513034254 which measures performance in sequential access to get a feeling of the performance in random access.

```java
import loci.common.DebugTools;
import loci.formats.FormatTools;
import loci.formats.in.ImarisHDFReader;

public class PixTestRandom {
    DebugTools.enableLogging("WARN");
    int nplanes;
    try (ImarisHDFReader reader = new ImarisHDFReader()) {
      reader.setId(args[0]);
      nplanes = reader.getImageCount();
    }
    for (int plane=0; plane<nplanes; plane++) {
      try (ImarisHDFReader reader = new ImarisHDFReader()) {
         reader.setId(args[0]);

        byte[] buf = new byte[FormatTools.getPlaneSize(reader)];
        long t0 = System.currentTimeMillis();
        reader.openBytes(plane, buf);
        long t1 = System.currentTimeMillis();
        System.out.println("plane #" + plane + " time: " + (t1 - t0) + " ms");
      }
    }
  }
}
```

Below are the scatter plots aggregating the results of 3 separate executions of the scripts using Bio-Formats 8.0.1 and this PR

## Sequential access
sequential

There is no debate here and sequential access is clearly improved by the caching. The overhead of regenerating the cache at the end of every block is largely compensated by the benefit of using the cache for all the other reads.

## Random access
random

There are really two behaviors here:
- if the accessed plane is part of the block that was cached during `initFile` (planes 0-15 and 98-112), the read times are improved with this change
- if the accessed plane is outside this block, the read times are degraded with this change

I retested the above on another [public sample Imaris file](https://downloads.openmicroscopy.org/images/Imaris-IMS/eliana/2016-07-26_17.11.03_Sample%20Prefix_5P298C2_Protocol.ims) with one channel and less Z planes and got a very similar result

random2

Assuming the overhead comes from the cache update, would it make sense to disable the caching during the reader initialization in https://github.com/ome/bioformats/blob/862723aa15ca004fb19e0cabbca702d0f012a5ac/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java#L308 call in `initFile`? This means the cache would only be populated during the first `openBytes` call after initialization. For sequential access, this should only defer and not change the total access time while for random access, this would make the behavior consistent across all planes.

melissalinkert:
> Assuming the overhead comes from the cache update, would it make sense to disable the caching during the reader initialization in
> 
> https://github.com/ome/bioformats/blob/862723aa15ca004fb19e0cabbca702d0f012a5ac/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java#L308
> 
> call in `initFile`? This means the cache would only be populated during the first `openBytes` call after initialization. For sequential access, this should only defer and not change the total access time while for random access, this would make the behavior consistent across all planes.

I think that makes sense. It also means `initFile` would be a little faster, which is helpful for use cases that are only looking at used files and metadata without reading images.

@marcobitplane, would you be OK to make this small change?

marcobitplane:
Thank you both for the detailed performance testing!

I added a commit to avoid caching when determining pixel type, I considered a few options and ended up preferring a dedicated getSampleData() method but feel free to suggest something different.

sbesson (approved):
Thanks @marcobitplane, the nightly CI repository tests have been passing with the latest changes. I recomputed the read times when opening every plane both in sequential access and in random access against the sample file:

sequential
random

The sequential access results are unchanged except for the first plane index where the caching happens. The random access results now show a consistent behavior across all plane indices.

As shown in both results, the caching introduce an overhead everytime a new block is read primarily to the additional amount of data retrieved from the storage (blockSize * sizeC more than before). In that regard, accessing a single plane in random access is probably the most challenging test for the strategy implemented in this proposal. Still the response time of a single plane remains acceptable and mostly lower than 200ms.

I looked at a last variant of this test keeping the random access but reading the planes for all the channels of a given Z sections e.g. as one would do for rendering a multi-channel 2D slice.

```java
import loci.common.DebugTools;
import loci.formats.FormatTools;
import loci.formats.in.ImarisHDFReader;

public class PixTest2ChannelsRandom {
    DebugTools.enableLogging("WARN");
    int sizeZ;
    int sizeC;
    try (ImarisHDFReader reader = new ImarisHDFReader()) {
      reader.setId(args[0]);
      sizeZ = reader.getSizeZ();
      sizeC = reader.getSizeC();
    }
    for (int z=0; z<sizeZ; z++) {
      try (ImarisHDFReader reader = new ImarisHDFReader()) {
        reader.setId(args[0]);
        
        byte[] buf = new byte[FormatTools.getPlaneSize(reader)];
        long t0 = System.currentTimeMillis();
        for (int c=0; c<sizeC; c++) {
          reader.openBytes(reader.getIndex(z, c, 0), buf);
        }
        long t1 = System.currentTimeMillis();
        System.out.println((t1 - t0));
      }
    }
  }
}
``` 

Using the 2-channel sample file, we can see the benefit of the caching as the numbers are now matching those of Bio-Formats 8.0.1:

random2channels

The series of measurements above convinced me that the performance changes make substantial improvements for the sequential access scenario that was definitely suboptimal while retaining acceptable performance in the random access scenario.
From my perspective, this is ready for inclusion in the next minor release of Bio-Formats although I will leave it open until next Monday for additional comments from the OME team.

sbesson:
Merging. This will be included in the next Bio-Formats minor release early 2025, we'll review contributions and have a clearer release date after the break. Thanks @marcobitplane for your contribution.

PR #4253, opened 2024-11-07
Operetta/Phenix: calculate physical Z size in micrometers

### LINKED ISSUE #4240: OperettaReader: use smaller units when calculating physical Z size
With https://downloads.openmicroscopy.org/images/PerkinElmer-Operetta/omer/:

```
$ showinf -no-upgrade -nopix -omexml Index.idx.xml
...
<Pixels BigEndian="false" DimensionOrder="XYCZT" ID="Pixels:0" Interleaved="false" PhysicalSizeX="0.29898804047838085" PhysicalSizeXUnit="µm" PhysicalSizeY="0.29898804047838085" PhysicalSizeYUnit="µm" PhysicalSizeZ="9.999999999999994E-6" PhysicalSizeZUnit="dm" SignificantBits="16" SizeC="6" SizeT="1" SizeX="1080" SizeY="1080" SizeZ="18" Type="uint16">
...
```

As @mabruce noticed, after import to OMERO the `PhysicalSizeZ` is rounded to 0.0dm for display. That's not fully representative of the true value, and incorrectly suggests that the physical Z size is missing.

The Z positions being used to calculate `PhysicalSizeZ` are:

```
$ grep "<PositionZ" Index.idx.xml  | sort | uniq
      <PositionZ Unit="m">5.7E-05</PositionZ>
      <PositionZ Unit="m">5.8E-05</PositionZ>
      <PositionZ Unit="m">5.9E-05</PositionZ>
      <PositionZ Unit="m">6.1E-05</PositionZ>
      <PositionZ Unit="m">6.2E-05</PositionZ>
      <PositionZ Unit="m">6.3E-05</PositionZ>
      <PositionZ Unit="m">6.4E-05</PositionZ>
      <PositionZ Unit="m">6.5E-05</PositionZ>
      <PositionZ Unit="m">6.6E-05</PositionZ>
      <PositionZ Unit="m">6.7E-05</PositionZ>
      <PositionZ Unit="m">6.8E-05</PositionZ>
      <PositionZ Unit="m">6.9E-05</PositionZ>
      <PositionZ Unit="m">6E-05</PositionZ>
      <PositionZ Unit="m">7.1E-05</PositionZ>
      <PositionZ Unit="m">7.2E-05</PositionZ>
      <PositionZ Unit="m">7.3E-05</PositionZ>
      <PositionZ Unit="m">7.4E-05</PositionZ>
      <PositionZ Unit="m">7E-05</PositionZ>
```

and the relevant logic is:

https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/OperettaReader.java#L726
https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/OperettaReader.java#L1043

I suspect that simply performing the calculations in micrometers instead of meters will give a more useful result.

melissalinkert:
Fixes #4240 (cc @mabruce).

`showinf -nopix -omexml` on `curated/perkinelmer-operetta/public/omer/` without this change should show a `PhysicalSizeZ` that is very small, and in `dm` units. With this change, the `PhysicalSizeZ` should be in `µm` units, and large enough to not round to 0.

This does require a configuration update, which does flag that two sample plates now have a `PhysicalSizeZ` of `1µm`. See forthcoming configuration PR.

Adding to 8.0.1 for initial evaluation, but this can easily be moved to a later milestone.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -729,11 +729,10 @@ public class OperettaReader extends FormatReader {
             FormatTools.getPhysicalSizeY(first.resolutionY), i);
 
           if (getSizeZ() > 1 && last != null) {
-            Unit<Length> firstZUnit = first.positionZ.unit();
-            double firstZ = first.positionZ.value().doubleValue();
-            double lastZ = last.positionZ.value(firstZUnit).doubleValue();
+            double firstZ = first.positionZ.value(UNITS.MICROMETER).doubleValue();
+            double lastZ = last.positionZ.value(UNITS.MICROMETER).doubleValue();
             double averageStep = (lastZ - firstZ) / (getSizeZ() - 1);
-            store.setPixelsPhysicalSizeZ(FormatTools.getPhysicalSizeZ(averageStep, firstZUnit), i);
+            store.setPixelsPhysicalSizeZ(FormatTools.getPhysicalSizeZ(averageStep, UNITS.MICROMETER), i);
           }
         }
       }


sbesson (approved):
While this file format internally supports different length units and nanometers are typically used for excitation wavelength, several quantities including `ImageResolutionX`, `ImageResolutionY`, `PositionX`, `PositionY`, `PositionZ` are notoriously stored as meters using scientific notations for their values. This statement applies for all the representative samples in the curated OME QA repository.

The reader internally converts the values of `ImageResolutionX` and `ImageResolutionY` into micrometers when populating the `Pixels.PhysicalSize{X,Y}` values since the introduction of the reader in https://github.com/ome/bioformats/pull/22. Note this logic might also reflect the vestige of a time where the OME data model did not support alternative units and the physical sizes had to be stored as micrometers. However, looking at all representative samples, this matches the actual scale of every  data acquired by this instrument and received so far and provides a much better user experience for any tool aiming to display this metadata.

In that sense, the change proposed  in this PR simply updates the handling of the physical size alongside Z to match the exact same assumptions as the other physical sizes and improves both the reader consistency as well as the user experience.

PR #4254, opened 2024-11-13
Memoizer: ask any ImageReader instances to remove all but the current reader

melissalinkert:
Ported from a private draft, for wider discussion (cc @joshmoore). https://github.com/glencoesoftware/omero-ms-image-region/issues/148 provides some additional context for where this idea is coming from.

To minimally demonstrate the difference, run something like `showinf -cache -minmax -expand -separate` without this change, and note size of memo file (in same directory as data). Run the same thing twice with this change - once to generate the new memo file (note the size again), and once after a minute or so to verify that the new memo file is used and not overwritten. Note `-minmax -expand -separate` mimics the reader stack used in OMERO.

However, this has the potential to impact anything that uses `Memoizer`, including but not limited to OMERO. So ideas of what might break, why this change might be bad in general, or other approaches to consider would all be welcome.

One potential problem identified is what happens when there is an existing memo file generated by this change, but new reader (or better type detection) is introduced such that the underlying reader that should be used for the file changes.

### DIFF
--- a/components/formats-bsd/src/loci/formats/Memoizer.java
+++ b/components/formats-bsd/src/loci/formats/Memoizer.java
@@ -208,6 +208,17 @@ public class Memoizer extends ReaderWrapper {
 
     @Override
     public void saveReader(IFormatReader reader) {
+      // clean up reader list in any instances of ImageReader
+      IFormatReader r = reader;
+      while (r instanceof ReaderWrapper || r instanceof ImageReader) {
+        if (r instanceof ImageReader) {
+          ImageReader ir = (ImageReader) r;
+          ir.cleanupReaderList();
+          r = ir.getReader();
+        }
+        else r = ((ReaderWrapper) r).getReader();
+      }
+
       kryo.writeObject(output, reader.getClass());
       kryo.writeObject(output, reader);
     }
--- a/components/formats-api/src/loci/formats/ImageReader.java
+++ b/components/formats-api/src/loci/formats/ImageReader.java
@@ -151,6 +151,24 @@ public class ImageReader implements IFormatReader {
   // -- ImageReader API methods --
 
   /**
+   * Intended for use prior to caching an initialized reader.
+   * This removes everything except the current reader from the
+   * list of possible readers to use. Be very careful, and do
+   * not call this method on an ImageReader that may be reused
+   * for multiple different files.
+   *
+   * If the current reader index is less than 0 (i.e. not initialized),
+   * this does nothing.
+   */
+  protected void cleanupReaderList() {
+    if (current >= 0) {
+      IFormatReader currentReader = readers[current];
+      current = 0;
+      readers = new IFormatReader[] {currentReader};
+    }
+  }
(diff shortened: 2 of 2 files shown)


joshmoore:
> why this change might be bad in general

:) Nope. I'm generally supportive. I remember having considered this, but probably didn't spend the time to think it through well enough as you have.

> but new reader (or better type detection) is introduced such that the underlying reader that should be used for the file changes.

This is a good and interesting point, but I imagine that a similar edge case could be found even without dropping instances from ImageReader. Since only a breaking change in the storage layout (new field) would trigger an invalidation, even just a change to the logic that *fills* a field could actually be something should invalidate the cache.

> other approaches to consider                                                                                          

Combined with the previous point, the only additional option I can think of is that we more fully define the semantics of when [`VERSION`](https://github.com/ome/bioformats/blob/384153da5ad54224f02a9003b967049c522eacef/components/formats-bsd/src/loci/formats/Memoizer.java#L353) should be bumped.

melissalinkert:
Thanks, @joshmoore.

> Combined with the previous point, the only additional option I can think of is that we more fully define the semantics of when [VERSION](https://github.com/ome/bioformats/blob/384153da5ad54224f02a9003b967049c522eacef/components/formats-bsd/src/loci/formats/Memoizer.java#L353) should be bumped.

That's a good point. Maybe at minimum, adding a new reader means updating the version number?

joshmoore:
No immediate objections to that other than that it might lead to a bit more churn then we've had to date. On the flip side, we can't necessarily control who has added a reader to ImageReader at runtime. I could imagine additionally allowing OMERO to override the version itself (passing a parameter? environment variable?) since that's a more "controlled" environment, but the general case in B-F land is definitely tricky.

melissalinkert:
As discussed in today's formats meeting, a next step would be to make the behavior here optional (but not default) so we can start testing in OMERO.

components/formats-bsd/src/loci/formats/Memoizer.java
```diff
@@ -208,6 +208,17 @@ public void saveRevision(String revision) {
 
     @Override
     public void saveReader(IFormatReader reader) {
+      // clean up reader list in any instances of ImageReader
+      IFormatReader r = reader;
+      while (r instanceof ReaderWrapper || r instanceof ImageReader) {
```
gerring, line 213: Suggest it would be cleaner to have ReaderWrapper and ImageReader both implement an interface 'ReaderContainer' (or use existing interface) for the method:
`IFormatReader getReader();`
Then test r for being an instance of that. Slightly bigger change so understand why not done :)

Since ImageReader requires the clean method the interface could have something like:
`IFormatReader getCleanReader();`

Which is implemented in ImageReader as `cleanupReaderList(); return getReader();` and in ReaderWrapper to just return `getReader()`.

Not a major point though.

melissalinkert:
Including as discussed in today's formats meeting, so we can start to evaluate impact.

PR #4255, opened 2024-12-02
Add Hamamatsu DCIMG Reader

zacsimile:
I found myself in the fortunate position of regularly needing to investigate bead stacks acquired as `.dcimg` files. My goal was to load them and reslice in xz. This seemed easiest to do in ImageJ, so I ported a merge of some existing DCIMG readers to Java for use as a local tool. The result is presented here, in case there is broader interest in the reader.

There are a few things to note:

1. On a Mac (works fine on Windows), while the stacks load fine, reslicing does not work on the loaded stack. I receive the error

```
java.lang.IllegalArgumentException: Invalid bitDepth: 0
  at ij.gui.NewImage.createImage(NewImage.java:397)
  at ij.plugin.Slicer.createOutputStack(Slicer.java:462)
  at ij.plugin.Slicer.resliceRectOrLine(Slicer.java:442)
  at ij.plugin.Slicer.reslice(Slicer.java:115)
  at ij.plugin.Slicer.run(Slicer.java:80)
  at ij.IJ.runPlugIn(IJ.java:216)
  at ij.Executer.runCommand(Executer.java:152)
  at ij.Executer.run(Executer.java:70)
  at java.lang.Thread.run(Thread.java:750)
```

If I duplicate the stack, reslicing works fine. Does anyone know what is likely to cause such an error just on a Mac?

2. I’ve made a design choice, where multiple frames within a single DCIMG file are loaded in `T` and multiple DCIMG files are loaded along `Z`. 

3. I’ve done the bare minimum on loading metadata.

4. I have only tested the 4 pixel correction on older DCIMG 0x7. I don’t have any examples of DCIMG 0x1000000 or 0x2000000 that need the 4 pixel correction, though I would be happy to take a look at one.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DCIMGReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DCIMGReader.java
@@ -5,8 +5,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import com.sun.media.jfxmedia.logging.Logger;
-
 import loci.common.Location;
 import loci.common.RandomAccessInputStream;
 import loci.formats.CoreMetadata;
@@ -101,6 +99,7 @@ public class DCIMGReader extends FormatReader {
       stream.skipBytes(byteFactor*(getSizeX() - w - x));
     }
 
+    stream.close();
     return buf;
   }
 
--- a/components/formats-gpl/src/loci/formats/in/DCIMGReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DCIMGReader.java
@@ -19,6 +19,7 @@ import loci.formats.meta.MetadataStore;
  * 
  * Follows spec at https://github.com/python-microscopy/python-microscopy/blob/master/PYME/IO/dcimg.py
  * and https://github.com/lens-biophotonics/dcimg/blob/master/dcimg.py.
+ * 
  */
 public class DCIMGReader extends FormatReader {
 
@@ -34,11 +35,21 @@ public class DCIMGReader extends FormatReader {
 
   // -- Fields --
 
+  private long version;
   private long headerSize;
   private long dataOffset;
   private long pixelType;
-  private int bytesPerImage;
+  private long bytesPerImage;
+  private long bytesPerRow;
   private int byteFactor;
+  private long offsetToFooter;
+  private boolean fourPixelCorrectionInFooter = false;
+  private long offsetToFourPixels;
+  private long fourPixelOffsetInFrame;
+  private int fourPixelCorrectionLine;
+  private long fourPixelCorrectionOffset;
+  private long currentStreamPosition;
+  private long frameFooterSize;
 
   private List<String> companionFiles = new ArrayList<String>();
   private String[] uniqueFiles;
@@ -99,8 +110,26 @@ public class DCIMGReader extends FormatReader {
     // DCIMG is stored column major
     stream.seek(headerSize + dataOffset + tp*bytesPerImage + byteFactor*y*getSizeX());
     for (int row=h-1; row>=0; row--) {
-      stream.skipBytes(byteFactor*x);
-      stream.read(buf, byteFactor*row*w, byteFactor*w);
+      if (fourPixelCorrectionInFooter && (row == fourPixelCorrectionLine) && (x < 4)) {
+
+        // mark the current position, as we want to pick up from here
+        currentStreamPosition = stream.getFilePointer();
+
+        // go get the four pixel offset
+        stream.seek(fourPixelCorrectionOffset);
+        stream.skip(byteFactor*x);
+        stream.read(buf, byteFactor*row*w, byteFactor*(4-x));
+
+        // go back to our current position, plus four pixels
+        stream.seek(currentStreamPosition);
+        stream.skipBytes(byteFactor*4);
+
+        // continue reading
+        stream.read(buf, byteFactor*(row*w+4), byteFactor*(w-4));
+      } else {
+        stream.skipBytes(byteFactor*x);
+        stream.read(buf, byteFactor*row*w, byteFactor*w);
+      }
       stream.skipBytes(byteFactor*(getSizeX() - w - x));
     }
 
@@ -122,7 +151,7 @@ public class DCIMGReader extends FormatReader {
 
     in.seek(8);
 
-    long version = in.readUnsignedInt();  // DCIMG version number
+    version = in.readUnsignedInt();  // DCIMG version number
     LOGGER.info("DCIMG Version: {}", version);
     if ((!(version == DCIMG_VERSION_0)) && (!(version >= DCIMG_VERSION_1))) {
       throw new FormatException(String.format("Unknown DCIMG version number %d.", version));
@@ -134,8 +163,8 @@ public class DCIMGReader extends FormatReader {
 
     in.skipBytes(20);
 
-    long numSessions = in.readUnsignedInt();
-    long numFrames = in.readUnsignedInt();
+    in.skipBytes(4);  // long numSessions = in.readUnsignedInt();
+    in.skipBytes(4);  // long numFrames = in.readUnsignedInt();
     headerSize = in.readUnsignedInt();
     in.skipBytes(4);
     long fileSize = in.readUnsignedInt();
@@ -143,7 +172,7 @@ public class DCIMGReader extends FormatReader {
     long fileSize2 = in.readUnsignedInt();
     if (fileSize != fileSize2) throw new FormatException("Improper header. File sizes do not match.");
     in.skipBytes(16);
-    long mystery1 = in.readUnsignedInt();  // 1024 in all examples
+    // long mystery1 = in.readUnsignedInt();  // 1024 in all examples
 
     CoreMetadata m = core.get(0);
 
@@ -155,10 +184,14 @@ public class DCIMGReader extends FormatReader {
     m.falseColor = false;
     m.metadataComplete = true;
     m.thumbnail = false;
-    m.sizeC = 1;
+    m.sizeC = 1;  // AFAIK DCIMG are always grayscale
 
+    // One shot header reading to avoid lots of control flow statements (because of
+    // the different DCIMG versions). Unfortunately, a bit less clean than parsing
+    // each parameter individually.
     if (version == DCIMG_VERSION_0) {
       parseDCAMVersion0Header(in);
+      parseDCAMVersion0Footer(in);
     } else if (version == DCIMG_VERSION_1) {
       parseDCAMVersion1Header(in);
     }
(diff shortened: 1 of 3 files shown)


melissalinkert:
Thank you, @zacsimile.

Before we proceed with review, there are two things we will need:

- a signed Contributor License Agreement (see https://ome-contributing.readthedocs.io/en/latest/cla.html)
- example datasets we can use to test this reader (uploaded to https://zenodo.org/communities/bio-formats)

Once we have those two things, we'll assign reviewers to this pull request. Please note we'll do our best to review in a timely fashion, but can't guarantee a timeline at this point. Please see https://forum.image.sc/t/bio-formats-update/101734 for more information about the current state of Bio-Formats maintenance.

> 1. On a Mac (works fine on Windows), while the stacks load fine, reslicing does not work on the loaded stack. I receive the error
> 
> ```
> java.lang.IllegalArgumentException: Invalid bitDepth: 0
>   at ij.gui.NewImage.createImage(NewImage.java:397)
>   at ij.plugin.Slicer.createOutputStack(Slicer.java:462)
>   at ij.plugin.Slicer.resliceRectOrLine(Slicer.java:442)
>   at ij.plugin.Slicer.reslice(Slicer.java:115)
>   at ij.plugin.Slicer.run(Slicer.java:80)
>   at ij.IJ.runPlugIn(IJ.java:216)
>   at ij.Executer.runCommand(Executer.java:152)
>   at ij.Executer.run(Executer.java:70)
>   at java.lang.Thread.run(Thread.java:750)
> ```
> 
> If I duplicate the stack, reslicing works fine. Does anyone know what is likely to cause such an error just on a Mac?

I don't have any immediate ideas, but once we have example data I expect we would try to reproduce this error as part of the review process.

melissalinkert:
Apologies, it looks like we do already have a signed CLA. So we'll just need example data in order to proceed here.

zacsimile:
No worries! I apologize, after reading that forum post, it seems this is exactly the kind of PR you are not looking for at the moment. Thanks for having a look anyway, and please take your time. I have compiled a local version of bioformats that works for me at the moment. 

I will upload some data ASAP. I just have to email a few people because a couple of the files I tested on are not mine and I'm not sure if uploading them is OK. In this case, I will have to record a couple more files in different DCIMG formats.

zacsimile:
Please find examples for DCIMG `0x7` and `0x1000000` at https://zenodo.org/records/14281237 and https://zenodo.org/records/14268554, respectively. The first one is a stack with multiple frames in a single DCIMG file and the second is a sequence of single frames, both monochromatic. I still do not have a DCIMG version `0x2000000` or a DCIMG with multiple color channels, but I do not claim to be able to load these with the current code. Happy to update it later if someone provides one or both of these examples.

zacsimile:
Fixed 4-pixel correction for DCIMG 0x1000000 upon receiving example at https://zenodo.org/records/14287640.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
+package loci.formats.in;
```
melissalinkert, line 26: This file should have the same copyright statement that is at the top of the other readers in the `formats-gpl` component.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+  {
+    return FormatTools.CAN_GROUP;
+  }
+
+  @Override
+  public String[] getSeriesUsedFiles(boolean noPixels) 
+  {
+    FormatTools.assertId(currentId, true, 1);
+
+    if (!isGroupFiles() || noPixels) return null;
```
melissalinkert: This will produce an invalid used files list any time file grouping turned off, even on a single-file dataset. `showinf -nogroup  Cell07_642_000_000.dcimg` demonstrates the issue. When `isGroupFiles()` is `false`, the used file list should contain the one file that is being read.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+
+  }
+
+  @Override
+  public boolean isThisType(RandomAccessInputStream stream) 
+    throws IOException 
+  {
+    String desc = stream.readString(5);
+    if (!desc.equals("DCIMG")) return false;
+    return true;
```
melissalinkert: These two lines could be simplified to `return desc.equals("DCIMG");`?

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+  @Override
+  public byte[] openBytes(int no, byte[] buf, int x, int y, int w, int h) 
+    throws FormatException, IOException
+  {
+    FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
+
+    int zp = no / getSizeT();
+    int tp = no % getSizeT();
+
+    RandomAccessInputStream stream = new RandomAccessInputStream(uniqueFiles[zp]);
```
melissalinkert: The stream should be opened and used in a try-with-resources, which means the explicit call to `stream.close()` isn't needed. https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/MicroCTReader.java#L148 is one similar example of this pattern.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+    stream.seek(headerSize + dataOffset + tp*bytesPerImage + byteFactor*y*getSizeX());
+    for (int row=h-1; row>=0; row--) {
+      if (fourPixelCorrectionInFooter && (row == fourPixelCorrectionLine) && (x < 4)) {
+
+        // mark the current position, as we want to pick up from here
+        currentStreamPosition = stream.getFilePointer();
+
+        // go get the four pixel offset
+        stream.seek(fourPixelCorrectionOffset);
+        stream.skip(byteFactor*x);
```
melissalinkert: Here and in lines 124-125, it would be a good idea to combine the seek and skip into a single seek call if possible. Reducing the number of seeks/skips is a known way to improve read performance in general.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+    if (version == DCIMG_VERSION_0) {
+      parseDCAMVersion0Header(in);
+      parseDCAMVersion0Footer(in);
+    } else if (version == DCIMG_VERSION_1) {
+      parseDCAMVersion1Header(in);
+    }
+
+    if (pixelType == DCIMG_PIXELTYPE_MONO8) {
+      m.pixelType = FormatTools.UINT8;
+      byteFactor = 1;
```
melissalinkert: It's not necessary to store the number of bytes per pixel separately. This can be retrieved from the pixel type using `FormatTools.getBytesPerPixel(getPixelType())`.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+    } else if (pixelType == DCIMG_PIXELTYPE_MONO16) {
+      m.pixelType = FormatTools.UINT16;
+      byteFactor = 2;
+    }
```
melissalinkert, line 227: Logging or throwing `FormatException` if an unexpected pixel type is found would be useful.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+      addFileToList(file);
+    }
+  }
+
+  // Logic copied from DicomReader 
+  private void addFileToList(String file)
+    throws FormatException, IOException
+  {
+    String ext = getExtension(file);
+    if (!ext.equals("dcimg")) {
```
melissalinkert: `checkSuffix(file, "dcimg")` (inherited from `FormatHandler`) could be used here instead.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+
+  @Override
+  public boolean isThisType(RandomAccessInputStream stream) 
+    throws IOException 
+  {
+    String desc = stream.readString(5);
+    if (!desc.equals("DCIMG")) return false;
+    return true;
+  }
+
```
melissalinkert, line 119: The `close(boolean)` method should be overriden here to reset all fields. https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/MicroCTReader.java#L126 is an example.

components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -0,0 +1,408 @@
...
+  // Logic copied from DicomReader 
+  private void addFileToList(String file)
+    throws FormatException, IOException
+  {
+    String ext = getExtension(file);
+    if (!ext.equals("dcimg")) {
+      LOGGER.warn("File {} with extension {} failed extension check", file, ext);
+      return;
+    }
+    RandomAccessInputStream stream = new RandomAccessInputStream(file);
```
melissalinkert: As in `openBytes` above, this stream should be opened using a try-with-resources.

melissalinkert (changes requested):
Thanks for providing test data, @zacsimile. I've done a first round of code review - let us know when you have had a chance to address all of the comments, and then we can make a plan for further testing.

zacsimile:
Thanks so much for the feedback! Addressed.

melissalinkert (approved):
https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/293/ continues to pass with this pull request included, and the last set of changes looks good to me.

Propose to get this in for 8.1.0, but will hold off on merging until the new year in case any final comments.

sbesson:
Only last question while re-reading the code changes: can I double check the reader is expected to be added to `formats-gpl` - the other option being `formats-bsd` ?

zacsimile:
One of the codes I based this off of is licensed under GPL-3.0 and the other is MIT. I went more permissive. I didn't explicitly copy code, since it was all Python, so there might be some flexibility here (although I'm not sure, I'm still confused by these licenses). Is there a reason to prefer BSD in this case?

sbesson:
>  Is there a reason to prefer BSD in this case?

Nothing specific from my side. Just wanted to make sure the licensing aspect had been discussed and decided on your side rather than copy-n-pasted from other reader examples. Relicensing a piece of code after it has been released is doable but also additional work so better to have things sorted before merging.

melissalinkert:
This is now merged, so will be released in Bio-Formats 8.1.0. Thanks again, @zacsimile!

PR #4259, opened 2024-12-10
OIR: parse time step from TIMELAPSE axis when possible

melissalinkert:
See https://forum.image.sc/t/bio-formats-reads-wrong-time-interval-from-oir-time-lapse-images/105684/

Without this change, `showinf -nopix -omexml` on the provided file (in `curated/olympus-oir/imagesc-105684/`) should show `Plane.DeltaT` values spaced ~2.17 seconds apart. With this change, `Plane.DeltaT` values should be spaced 30 seconds apart.

The time step is now being parsed the same way that the Z step is parsed, so I think this makes sense. I would expect some test failures overnight in the `DeltaT` test for existing .oir data; I can update configs as needed later in the week, or we can exclude temporarily.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -36,6 +36,7 @@ import java.util.Map;
 import java.util.Objects;
 import javax.xml.parsers.ParserConfigurationException;
 
+import loci.common.Constants;
 import loci.common.DataTools;
 import loci.common.DateTools;
 import loci.common.Location;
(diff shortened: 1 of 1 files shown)


sbesson:
A few OIR samples failed the tests for the plane timestamp values and will need to be reconfigured accordingly - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/45268/console

melissalinkert:
As noted above, this change does modify the timestamps for a few existing files, including https://downloads.openmicroscopy.org/images/Olympus-OIR/gh-4205/zenodo-12773657/DAPI_mCherry_4T_5Lambda-420-630-w10nm-s50nm.oir.

@NicoKiaru, since that file was submitted by you, any thoughts on this change or the expected plane timestamps for the file would be welcome.

NicoKiaru:
I confirmed that there was an issue in the timestamps which is fixed by these commits, thanks @melissalinkert !

I checked on the Fluoview software, and the delay between frames is of 20 seconds, and thanks to this PR the frame interval is now detected to be of 20 secs with the file you linked.

sbesson (approved):

PR #4263, opened 2025-01-10
Bump logback version to 1.3.15

sbesson:
See also https://github.com/ome/ome-common-java/pull/96


melissalinkert (approved):
Merging as discussed in today's formats meeting.

PR #4265, opened 2025-01-16
Bump ome-common, ome-model and ome-metakit to their latest releases

sbesson:
See https://github.com/ome/ome-model/releases/tag/v6.4.0, https://github.com/ome/ome-common-java/releases/tag/v6.0.25 and https://github.com/ome/ome-metakit/releases/tag/v5.3.8 for more details


melissalinkert (approved):
:+1: Can merge tomorrow assuming builds continue to pass.

PR #4268, opened 2025-01-22
Remove outdated entry from plugins.config

tferr:
See discussion at https://forum.image.sc/t/housecleaning-fiji-for-the-java-21-update/107452


melissalinkert:
Thanks, @tferr. I don't have any objections to removing this menu item, since it has been 11 years since it was introduced in #944. Adding the `include` label so this gets tested by tonight's CI jobs.

To merge this as-is, we would need a signed Contributor License Agreement:

If there is any problem with signing that, let us know.

tferr:
Thanks a lot for the super-fast reply @melissalinkert! Just emailed the signed CLA. Let me know if something is missing.

sbesson (approved):

PR #4270, opened 2025-01-28
InCell: allow mix of 2D and 3D channels to be treated as a Z stack

melissalinkert:
See https://forum.image.sc/t/incell-plates-with-z-stacks-cannot-be-imported-on-omero/107802 and data copied to `inbox/imagesc-107802`.

As indicated in the forum post, one channel in the plate has a Z stack (16 slices), while the remaining channels have a single Z slice. The changes in lines 795 and later are to pick up the dimensions of each field as 16 Z x 4 C, with most planes blank.

The changes in lines 467-468 are necessary because the total image count is 152, and the number of planes per field is 64 - so the `expectedSeries` calculation results in 2, when there are actually 8 series (4 fields x 2 wells). This change in particular might break some older data, so we'll see what the test results are, and may need to exclude/rework this a bit.

Adding to 8.1.1 milestone instead of 8.1.0, anticipating that some extra work will be needed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -464,7 +464,8 @@ public class InCellReader extends FormatReader {
       }
       int expectedSeries = totalImages / (getSizeZ() * getSizeC() * getSizeT());
       if (expectedSeries > 0) {
-        seriesCount = (int) Math.min(seriesCount, expectedSeries);
+        LOGGER.warn("Using series count {} but plane count indicates {}",
+          seriesCount, expectedSeries);
       }
     }
     else seriesCount = totalImages / (getSizeZ() * getSizeC() * getSizeT());
@@ -791,7 +792,7 @@ public class InCellReader extends FormatReader {
     private int wellRow, wellCol;
     private int nChannels = 0;
     private boolean doT = true;
-    private boolean doZ = true;
+    private Boolean doZ = null;
     private Image lastImage = null;
 
     @Override
@@ -908,7 +909,11 @@ public class InCellReader extends FormatReader {
         String fusion = attributes.getValue("fusion_wave");
         if (fusion.equals("false")) ms0.sizeC++;
         String mode = attributes.getValue("imaging_mode");
-        if (mode != null) {
+
+        // different wavelengths (channels) may have different imaging modes
+        // we want to allow a Z stack if one or more "3-D" imaging modes
+        // are encountered, even if some are variations on "2-D"
+        if (mode != null && (doZ == null || !doZ)) {
           doZ = mode.equals("3-D");
         }
       }
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -67,6 +67,9 @@ public class InCellReader extends FormatReader {
 
   // -- Constants --
 
+  public static final String DUPLICATE_PLANES_KEY = "incell.duplicate_missing_planes";
+  public static final boolean DUPLICATE_PLANES_DEFAULT = true;
+
   public static final String INCELL_MAGIC_STRING = "IN Cell Analyzer";
   public static final String CYTELL_MAGIC_STRING = "Cytell";
 
@@ -125,6 +128,17 @@ public class InCellReader extends FormatReader {
     hasCompanionFiles = true;
     datasetDescription = "One .xdce file with at least one .tif/.tiff or " +
       ".im file";
+  }
+
+  // -- InCellReader API methods --
+
+  public boolean duplicatePlanes() {
+    MetadataOptions options = getMetadataOptions();
+    if (options instanceof DynamicMetadataOptions) {
+      return ((DynamicMetadataOptions) options).getBoolean(
+       DUPLICATE_PLANES_KEY, DUPLICATE_PLANES_DEFAULT);
+    }
+    return DUPLICATE_PLANES_DEFAULT;
   }
 
   // -- IFormatReader API methods --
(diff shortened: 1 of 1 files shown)


jburel:
See https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/52803/

components/formats-gpl/src/loci/formats/in/InCellReader.java
```diff
@@ -908,8 +922,19 @@ else if (qName.equals("Wavelength")) {
         String fusion = attributes.getValue("fusion_wave");
         if (fusion.equals("false")) ms0.sizeC++;
         String mode = attributes.getValue("imaging_mode");
+
+        // different wavelengths (channels) may have different imaging modes
+        // we want to allow a Z stack if one or more "3-D" imaging modes
+        // are encountered, even if some are variations on "2-D"
         if (mode != null) {
-          doZ = mode.equals("3-D");
+          boolean is3D = mode.equals("3-D");
```
sbesson, line 961: An interesting realisation while browsing the xdce files that we have received is that every `Wavelength` element that includes some `imaging_mode` attribute also includes `z_section/z_slice/z_index` attributes.

At first look, these would seem to be a more reliable way of identifying images with varying acquired Z-section. However, looking specifically at the XML of the public dataset uploaded to Zenodo

```xml
        <Wavelengths>
            <Wavelength aperture_rows="50" dof_factor="8.00" edge_confocal_mode="false" edge_confocal_width="-1" fusion_wave="false" imaging_mode="2-D Deconvolution" index="0" label="" laser_power="100" name="FITC_511" num_frames_to_avg="1" open_aperture="true" show="false" z_section="0.000" z_slice="1" z_start_index="0" z_step="0.000">
            </Wavelength>
            <Wavelength aperture_rows="50" dof_factor="8.00" edge_confocal_mode="false" edge_confocal_width="-1" fusion_wave="false" imaging_mode="3-D" index="1" label="" laser_power="100" name="DAPI" num_frames_to_avg="1" open_aperture="true" show="false" z_section="64.000" z_slice="16" z_start_index="-4" z_step="4.000">
            </Wavelength>
            <Wavelength aperture_rows="50" dof_factor="8.00" edge_confocal_mode="false" edge_confocal_width="-1" fusion_wave="false" imaging_mode="2-D" index="2" label="" laser_power="100" name="Cy3" num_frames_to_avg="1" open_aperture="true" show="false" z_section="4.000" z_slice="3" z_start_index="-1" z_step="4.000">
            </Wavelength>
            <Wavelength aperture_rows="50" dof_factor="8.00" edge_confocal_mode="false" edge_confocal_width="-1" fusion_wave="false" imaging_mode="2-D" index="3" label="" laser_power="100" name="Brightfield" num_frames_to_avg="1" open_aperture="true" show="false" z_section="0.000" z_slice="1" z_start_index="0" z_step="0.000">
            </Wavelength>
        </Wavelengths>
```

the `Cy3` channel is marked as `imaging_mode="2-D"` with `z_slice="3"`

sbesson (commented):
Using the reference dataset uploaded at https://zenodo.org/records/14777242, confirmed that the latest released version of Bio-Formats 8.1.0 fails to initialize the `xdce` file with `java.lang.IllegalArgumentException`.

With these changes include, the reader initializes successfully and detects a plate with 8 images of dimensionality 1024x1024x16x4x1

Screenshot 2025-02-25 at 09 15 05

The handling of mixed 2D/3D acquisitions is consistent with the decisions that have been made for other HCS readers, notably CV7000Reader - see https://github.com/ome/bioformats/pull/4085 for the relevant changes.

As for CV7000, the outstanding question is what should be displayed for channels acquired in 2D when Z>0. The current implementation returns black planes. Inline with the changes made in https://github.com/ome/bioformats/pull/4085, would be be feasible to introduce an option allowing to either replicate the plane for Z=1 or return black planes and have the first behavior being the default?

melissalinkert:
> As for CV7000, the outstanding question is what should be displayed for channels acquired in 2D when Z>0. The current implementation returns black planes. Inline with the changes made in https://github.com/ome/bioformats/pull/4085, would be be feasible to introduce an option allowing to either replicate the plane for Z=1 or return black planes and have the first behavior being the default?

Added in c9b7c30.

components/formats-gpl/src/loci/formats/in/InCellReader.java
melissalinkert: Agreed, the `z_*` attributes here aren't really usable for setting the Z size for the channel. Maybe `z_start_index` could be used when the duplicate planes option is turned off, to pick where the single plane sits within the stack? But I don't know if that's actually useful, or just adds confusion.

sbesson (approved):
Tested against the first and last field of view of the supporting sample with the two choices for the duplicate plane option

```
$ ./tools/showinf ~/Downloads/Training_Demo_1/Training_Demo_1.xdce -autoscale
$ ./tools/showinf ~/Downloads/Training_Demo_1/Training_Demo_1.xdce -autoscale -series 5
$ ./tools/showinf ~/Downloads/Training_Demo_1/Training_Demo_1.xdce -autoscale -option incell.duplicate_missing_planes false
$ ./tools/showinf ~/Downloads/Training_Demo_1/Training_Demo_1.xdce -autoscale -series 5 -option incell.duplicate_missing_planes false
 ```
 
 By default, the plane at Z=0 is replicated for all Z indices for channels 0, 2 and 3. When setting `-option incell.duplicate_missing_planes false` all planes at Z>0 for channels 0,2 and 3 contain zero values. In both cases, the channel 1 loads the relevant binary data for each Z index.

Screenshot 2025-02-26 at 11 14 57

Including for the next Bio-Formats patch release (this will also need a documentation PR to list the new option)

PR #4276, opened 2025-02-19
ND2: preserve pixel type when a multi-position dataset is found

melissalinkert:
See https://forum.image.sc/t/reconstructed-multiposition-files-from-nikon-sim-open-like-static-noise/108834 and data uploaded to https://zenodo.org/records/14893791.

`showinf` on that dataset with 8.1.0 shows that the pixel type is detected as `uint16`; NIS Elements Viewer confirms that the pixel type should be `float`.

With this change, `showinf -normalize` should also show a pixel type of `float`, and the image should correspond with the screenshot of NIS Elements in the forum thread. Omitting the `-normalize` option is expected not to work, but ImageJ can also be used to check that the image looks correct.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Handler.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Handler.java
@@ -791,6 +791,7 @@ public class ND2Handler extends BaseHandler {
               int tSize = ms0.sizeT;
               int c = ms0.sizeC;
               String order = ms0.dimensionOrder;
+              int pixelType = ms0.pixelType;
               core = new CoreMetadataList();
               for (int i=0; i<numSeries; i++) {
                 CoreMetadata ms = new CoreMetadata();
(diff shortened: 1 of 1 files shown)


melissalinkert:
Wasn't fast enough to get this included in tonight's build, so we'll know on Friday morning if this passes (in which case I'll assign reviewers) or if more work is needed. cc @will-moore

sbesson (commented):
Without this change, all images of the dataset are detected as `uint16` and with this dataset, all images are detected as `float` with reasonable values when displaying the planes using `showinf -normalize` as described.

It looks like the NIS Elements Viewer is (temporarily? permanently?) no longer downloadable for Mac OS X so I was unable to confirm the ground truth using the reference software but the changes look good. Waiting for a complete run of the repository tests with this changes included including the new received dataset

sbesson (approved):

melissalinkert:
https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-repo/353/ is passing with this change and new test data - see https://github.com/openmicroscopy/data_repo_config/pull/634 for additional details of new data.

will-moore (approved):

PR #4277, opened 2025-02-24
Make type cast explicit when calculating image count

melissalinkert:
This is meant to test fixing an alert from CodeQL, and shouldn't otherwise impact builds.

PR #4278, opened 2025-02-25
Add CodeQL workflow

sbesson:
Follow-up of #4277

The default CodeQL configuration has a few limitations including the inability to run the workflow on Pull Requests. This commit adds the configuration for executing CodeQL workflows on every push and pull request against the development branch as well as a scheduled build

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   analyze:
     name: Analyze (${{ matrix.language }})
-    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
+    runs-on: ubuntu-latest
     permissions:
       # required for all workflows
       security-events: write
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,14 +1,3 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL Advanced"
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,98 @@
...
+          'languages you are analyzing, replace this with the commands to build' \
+          'your code, for example:'
+        echo '  make bootstrap'
+        echo '  make release'
+        exit 1
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v3
+      with:
+        category: "/language:${{matrix.language}}"
```
melissalinkert, line 50: I think https://github.com/ome/bioformats/pull/4278/checks?check_run_id=37790317684 indicates that the `analyze` action is being run on the matrix of automatically detected languages, not the matrix defined in lines 44-47.

sbesson: Added the actions language manually and https://github.com/ome/bioformats/pull/4278/checks?check_run_id=37796169026 looks much better

melissalinkert: Seems kind of weird that's necessary, but the output does look fine now. Probably not a bad idea to run checks against `python` and `actions` anyway.

melissalinkert (approved):

joshmoore (approved):
MSTM. Only two quick comments:
 - I'm fine with dropping the comments (esp. those at the beginning)
 - Also happy to see this moved to https://github.com/ome/.github/tree/master/workflow-templates if that would reduce the overall maintenance burden

sbesson:
> I'm fine with dropping the comments (esp. those at the beginning)

 Good point, [f47afac](https://github.com/ome/bioformats/pull/4278/commits/f47afac09f55760511aa8a69120bdc88f11fc35a) and [97e303a](https://github.com/ome/bioformats/pull/4278/commits/97e303ad7bed11cbd019ef4cac9c9877c50e44bd) should reduce the size of the configuration file and make it more readable and maintainable
 
> Also happy to see this moved to https://github.com/ome/.github/tree/master/workflow-templates if that would reduce the overall maintenance burden
 
The obvious caveat is the language will need to be adjusted depending on the OME repository. I don't have a specific need for my immediate next steps so this is probably more a question for the OME team

melissalinkert:
In preparation for 8.1.1, I will merge this tomorrow (Tuesday) afternoon unless any further comments.

PR #4144, opened 2024-01-18
SVS: remove thumbnails and improve label/macro detection

melissalinkert:
Fixes #3757.

As discussed at the end of #3757, thumbnail images are now removed by default. The `svs.remove_thumbnail` option controls this behavior; it is `false` by default but setting to `true` should cause thumbnails to be included (matching behavior without this PR).

816a588 and 2fd4489 have related updates to label/macro detection, so that the corresponding image names are set correctly for either value of `svs.remove_thumbnail`. Those commits also add support for the case where a label or macro image has a non-null comment that does not include `label` or `macro`; this is typically the result of anonymization/de-identification.

Configuration changes are needed to remove the thumbnails; corresponding PRs are forthcoming.

PR #4085, opened 2023-08-18
CV7000 improvements

melissalinkert:
Backported from a private PR. Adds a few improvements for more complex acquisitions:

- Include the action index in channel names, so that it's easier to distinguish between the same channel acquired in multiple actions
- Set `Channel.Fluor`, if possible
- For mixed 2D/Z stack acquisitions, `SizeZ` was already to be the largest Z size across all channels, but that leaves potentially many planes with no data. 5cd0ea5 and 6058cec switch the default behavior to from returning a blank plane to returning the first plane in the channel, for any missing data. Setting `cv7000.duplicate_missing_planes` to `false` restores the previous behavior.

Any existing plate will demonstrate the channel name changes; I'll add an artificial sample that demonstrates the `Fluor` and plane duplication changes shortly.

This might be better suited to a minor release, as it adds a new reader option and changes all channel names. A corresponding configuration PR is forthcoming.

PR #4279, opened 2025-02-25
Override `getAvailableOptions()` in CV7000 and SVS readers

melissalinkert:
Noticed while adding an option in https://github.com/ome/bioformats/pull/4270.

This shouldn't impact builds, but just unifies option reporting across readers. I think this was just an oversight in #4085 and #4144.

Haven't assigned a milestone yet, as not sure whether this is best done in a minor release, or OK for a patch.


sbesson (approved):
Thanks, this probably reflects the fact the `getAvailableOptions` API is not really consumed anywhere at the moment.

We can discuss how to improve this so that we have some mechanism to test the API is updated accordingly. In the interim, this change is fixing the return value of an existing API so I don't think this needs to wait for a minor release

PR #4283, opened 2025-02-26
InCell: only record per-timepoint channel counts if timelapse enabled

### LINKED ISSUE #4271: Dataset from InCell 2200 microscope misread as a plate
## Bug description
Quick summary : Some datasets acquired by an InCell 2200 microscope are not correctly read as a plate by BioFormats. This seems to be due to a channel duplication. This duplication introduces a `wix 0` suffix in the name of each image.

More info in the dedicated [forum post](https://forum.image.sc/t/incell-plates-with-duplicate-channels-not-recognized-as-plate-on-omero/107800)

## Dataset to test
[Zenodo repo](https://zenodo.org/records/14769820)

## Bioformats version
8.0.1

melissalinkert:
Fixes #4271. See data in `curated/incell/public/zenodo-14769820/`, particularly `Dataset_Fail`.

In some cases, a complex timelapse with varying channel counts per timepoint will be defined but not actually used. Examining the `TimeSchedule` under `PlateMap` in `Dataset3_RedoAcq_1.xdce` will show that the timepoints defined do not match up with the .tif files (where only 1 timepoint is present), but the `TimeSchedule` has `enabled` set to `false` and so should be ignored.

Without this change, `showinf -omexml Dataset3_RedoAcq_1.xdce` should throw an exception. With this change, the same test should initialize without exception, and display OME-XML and the first well's images. The ZCT dimensions and plate metadata should match up with the .tif file names.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/InCellReader.java
+++ b/components/formats-gpl/src/loci/formats/in/InCellReader.java
@@ -867,7 +867,7 @@ public class InCellReader extends FormatReader {
           }
         }
       }
-      else if (qName.equals("TimePoint")) {
+      else if (qName.equals("TimePoint") && doT) {
         channelsPerTimepoint.add(nChannels);
         nChannels = 0;
       }


sbesson (approved):
Using the public sample dataset, I was able to reproduce the failure with Bio-Formats 8.1.0.
With a build from this PR, the reader initializes, the OME metadata is consistent with the content of the `xdce` file and images render with 5 channels as expected.

Issue #3919, opened 2022-12-05
CZI: Failed to read pre-stitched image with no pyramid


dgault:
This issue was raised on forum thread https://forum.image.sc/t/czi-pyramidal-with-bioformat/73215/10 and a set of sample files provided (https://wetransfer.com/downloads/5021c67300c122c50c8394172155318c20221201105414/b5cb3f, Note: David to upload sample files to repo inbox). When attempting to open the stitched image with Bio-Formats 6.11.1 you instead see 42 series (comes from 21 mosaics with 2 channels) and the pixel data appears skewed.

While debugging the issue, if you modify the values at https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java#L832, keeping the seriesCount at 1 and setting prestitched to `true`, that is enough to allow the file to be opened correctly a single stitched image. This can likely be fixed by correcting this section of code to recognise a prestitched image when no pyramids are present.

dgault:
Sample files are now located under `inbox/imagesc-73215`

melissalinkert:
Confirmed with 8.5.0 that all 3 files are now read as a single series, and the images appear to be consistent with what is shown in ZEN.

Issue #3323, opened 2019-02-14
CZI 8bit brightfield tiles stitched in Zen open incorrectly


Svidro:
A bit depth of 14 is standard for CZI images, and I have no problems opening original tiled images, or fused/unfused (according to Zen) stitched results at that bit depth.
When saving images in 8bit RGB, single images open just fine. Tiles open just fine as long as they have NOT been stitched. The problem arises when attempting to open stitched 8bit CZI images. All of the information is there, as shown if I uncheck stitching within the BioFormats CZI reader (FIJI) and can see each tile just fine. Selecting the "Fuse tiles" option within Zen does not seem to impact the result. With stitching checked, the image opens as shown on the left:
![image](https://user-images.githubusercontent.com/23145209/52819847-3e7e4600-305f-11e9-98a6-cf9bccde283b.png)
I would like to be able to read the 8bit stitched images for analysis purposes :)

Example files can be found here: https://drive.google.com/open?id=1QLl0nzGPzy2cxAhNIgwPCI5xGk3sDW5i
There are 3 sets of files, two 8bit examples, and one 14bit example (that opens normally). Each has an unstitched (original) version, and at least one stitched version, either fused or unfused.

Please let me know if I can provide any additional information!

dgault:
Hi @Svidro, thank you for raising the issue and providing the sample filesets. Apologies for the delayed response, the Bio-Formats 6.0.0 release has taken up a lot of time over the past week. I have been able to test and reproduce the problem as you described using this latest Bio-Formats release.

I have created a Trello card on the Bio-Formats inbox to track this problem: https://trello.com/c/sPGdYLaz/331-czi-fused-8-bit-tiles-skewed

Svidro:
Thanks for taking a look!

PR #4284, opened 2025-03-03
fix for garbled images with CZIs

ptahmose:
This is trying to address an issue which was seen with CZIs from e. g. AxioScan. Those CZIs contain pyramids, and under certain situations the resulting images are garbled like this:
image

It was found that the code seems to determine the extent of the sub-blocks by dividing the logical size by a scale-factor (corresponding to the pyramid-layer), which seems highly questionable, or this assumption of equality seems unjustified. For the very least, it may give inaccurate results. The patched version is using information for the sub-block's dimension-entry array, which unambiguously gives the pixel-size (or: physical size) of the sub-block instead. For the cases investigated, this gives the correct result:
image

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -4241,7 +4241,7 @@ public class ZeissCZIReader extends FormatReader {
         }
       }
 
-      throw new FormatException("DimensionEntry for 'X' not found");
+      throw new FormatException("DimensionEntry for 'Y' not found");
     }
 
     // -- Helper methods --
--- a/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ZeissCZIReader.java
@@ -405,8 +405,12 @@ public class ZeissCZIReader extends FormatReader {
         {
           int res = (int) Math.pow(scaleFactor, plane.resolutionIndex);
 
-          int realX = plane.x / res;
-          int realY = plane.y / res;
+          // The physical extent (i.e. the number of pixels) for the respective sub-block is unambiguously given
+          //  by the "storedSize" of the respective dimension-entry. Note that this may differ from dividing the
+          //  logical extent by the scale-factor, e.g. due to rounding errors (since we are dealing here with integers
+          //  only).
+          int realX = plane.getPhysicalWidth();
+          int realY = plane.getPhysicalHeight();
 
           if ((prestitched != null && prestitched) || validScanDim) {
             Region tile = new Region(plane.col, plane.row, realX, realY);
@@ -4218,6 +4222,26 @@ public class ZeissCZIReader extends FormatReader {
         return buf;
       }
       return data;
+    }
+
+    public int getPhysicalWidth() throws FormatException {
+      for (DimensionEntry entry : this.directoryEntry.dimensionEntries) {
+        if ("X".equals(entry.dimension)) {
+          return entry.storedSize;
+        }
+      }
+
+      throw new FormatException("DimensionEntry for 'X' not found");
+    }
+
+    public int getPhysicalHeight() throws FormatException {
+      for (DimensionEntry entry : this.directoryEntry.dimensionEntries) {
+        if ("Y".equals(entry.dimension)) {
+          return entry.storedSize;
+        }
+      }
+
+      throw new FormatException("DimensionEntry for 'X' not found");
     }
 
     // -- Helper methods --


melissalinkert:
Thanks for opening this, @ptahmose.

[As noted in our contributing policy](https://ome-contributing.readthedocs.io/en/latest/code-contributions.html#procedure-for-accepting-code-contributions), we will need a signed Contributor License Agreement before we can fully test this change:

It would also be very helpful to have a test dataset that demonstrates this fix. If possible, please upload a file to https://zenodo.org/communities/bio-formats.

#3323 and #3919 may also be related, which I can verify as part of the review process once we have a signed CLA.

ptahmose:
Thanks, and I sent in the CLA. I'll upload an example file and instructions how to verify the fix later on.

ptahmose:
I uploaded sample data -> https://zenodo.org/api/records/14968770/draft/files/2025_01_27__0007_offline_Zen_3_9_5.czi/content .

With this file and running 
`bfconvert -series 4 -channel 0 2025_01_27__0007_offline_Zen_3_9_5.czi output.png`

I get (w/o this PR)
image

and with this PR
image

NicoKiaru:
Could be similar to https://forum.image.sc/t/would-anyone-have-a-palm-czi-example-file/85900/12

NicoKiaru:
( I can's access the file yet, I guess it needs to be approved by the OME team for the moment )

sbesson:
I cannot see the request for the record either. @ptahmose was the file submitted to the [Zenodo Bio-Formats collection](https://zenodo.org/communities/bio-formats)??

NicoKiaru:
Maybe it's still in a draft ? The link has a draft in its url

ptahmose:
Sorry for that, I am new to Zenodo. To my understanding, it is now in "review state" -> https://zenodo.org/uploads/14968770
image
If I missed something or additional actions is needed from my side - I'd be glad to learn.

melissalinkert:
Thanks, @ptahmose; the file is now visible.

melissalinkert (commented):
This change was [tested last night](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/58347) against our full repository of data; a couple of public .czi files are noted as having different image data with this change compared to the `develop` branch. Files are available from:

https://downloads.openmicroscopy.org/images/Zeiss-CZI/zenodo-10577186/

in particular:

https://downloads.openmicroscopy.org/images/Zeiss-CZI/zenodo-10577186/2023_11_30__RecognizedCode-27-Background%20subtraction-08.czi
https://downloads.openmicroscopy.org/images/Zeiss-CZI/zenodo-10577186/2023_11_30__RecognizedCode-27-Deconvolution%20%28defaults%29-11.czi

I don't see particularly obvious differences when inspecting visually, so will need a little more investigation to understand if the change in image data is actually problematic, or OK in this case.

NicoKiaru:
I just checked quickly and I don't get the issue in the quick start reader. It's probably because I was using the block size as well. However @ptahmose can you check whether your fix does not create an issue with the resolution level 5 of https://zenodo.org/records/10577621/files/Young_mouse.czi?download=1 ? 

This thing was failing:
```
ZeissCZIReader r = new ZeissCZIReader();
r.setId("image path to \\Young_mouse.czi");
r.setSeries(5);
r.openPlane(0,0,0,5947,2168);
```

ptahmose:
> I just checked quickly and I don't get the issue in the quick start reader. It's probably because I was using the block size as well. However @ptahmose can you check whether your fix does not create an issue with the resolution level 5 of https://zenodo.org/records/10577621/files/Young_mouse.czi?download=1 ?
> 
> This thing was failing:
> 
> ```
> ZeissCZIReader r = new ZeissCZIReader();
> r.setId("image path to \\Young_mouse.czi");
> r.setSeries(5);
> r.openPlane(0,0,0,5947,2168);
> ```

@NicoKiaru 
Can you maybe eleborate what the issue with this is, or what to look out for? What I tried is to run above lines, and I found that the result (i.e. the byte[] returned from r.openPlane(...) ) has the exact same content with and without the changes in this PR.
If you have indication of something being at odds here, let me know.

NicoKiaru:
If there's no problem then it's great. I found the issue back:

https://forum.image.sc/t/would-anyone-have-a-palm-czi-example-file/85900/12

The issue starts from this message and spans over a few posts below. In short the pixel block size in the metadata do not match the pixel sizes obtained after decompression  - it's off by one pixel.

melissalinkert:
Re-including in tonight's tests, with updated configuration for the 4 files that had test failures. Investigating the affected files referenced in https://github.com/ome/bioformats/pull/4284#pullrequestreview-2665198326 more carefully, it appears that the differences are valid, and the behavior with this change is correct.

If builds are passing over the weekend, I propose to merge on Monday for inclusion in 8.2.0.

PR #4286, opened 2025-03-05
Force isThisType check for .arf files

melissalinkert:
CV7000 datasets can contain an .arf file that is some sort of analysis results. `ARFReader` tried to read everything with an .arf extension, so directory imports of certain CV7000 plates may fail because the .arf files fail to be read.

Without this change, a simple test file:

```
$ cat test.arf
test
```

should throw this exception:

```
$ showinf -debug test.arf
...
loci.formats.FormatException: Undefined endianness
	at loci.formats.in.ARFReader.initFile(ARFReader.java:105)
	at loci.formats.tools.ImageInfo.main(ImageInfo.java:1159)
```

With this change, `UnknownFormatException` should be thrown instead.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ARFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ARFReader.java
@@ -55,6 +55,7 @@ public class ARFReader extends FormatReader {
   public ARFReader() {
     super("ARF", "arf");
     domains = new String[] {FormatTools.UNKNOWN_DOMAIN};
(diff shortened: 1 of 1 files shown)


sbesson (approved):
The changes make sense and act defensively against unsupported files with an `.arf` extension. Confirmed this change does not cause any regression with the `.arf` files supported by the reader. These files have just been reincluded in the nightly CI builds so proposing we get a passing build before merging this

PR #4288, opened 2025-03-06
Fix a bunch of `Implicit type narrowing` warnings

melissalinkert:
I can split up if easier to review, but just wanted to get the easy CodeQL warnings sorted out.

I wouldn't expect this to impact builds at all.

### DIFF
--- a/components/formats-bsd/src/loci/formats/ResourceNamer.java
+++ b/components/formats-bsd/src/loci/formats/ResourceNamer.java
@@ -139,8 +139,14 @@ public class ResourceNamer {
   public static int alphabeticIndexCount(String index) {
     int count = 0;
     char[] letters = index.toCharArray();
+    if (letters.length > 6) {
+      // this will cause an integer overflow
+      throw new IllegalArgumentException("Cannot count more than 6 characters");
+    }
     for (int i = 0; i < letters.length; i++) {
-      count += (letters[i] - 64) * Math.pow(ALPHABET_LENGTH, i);
+      int letterIndex = (int) Math.pow(ALPHABET_LENGTH, i);
+      int letterValue = letterIndex * (letters[i] - 64);
+      count += letterValue;
     }
     return count;
   }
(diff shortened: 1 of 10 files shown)


components/formats-bsd/src/loci/formats/ResourceNamer.java
```diff
@@ -139,8 +139,14 @@ public String getLetter() {
   public static int alphabeticIndexCount(String index) {
     int count = 0;
     char[] letters = index.toCharArray();
+    if (letters.length > 6) {
```
sbesson, line 142: Seeing the only place where this API is used is https://github.com/ome/bioformats/blob/105c8990d2526085f791c7381ae42ca4696562ff/components/formats-bsd/src/loci/formats/in/FakeReader.java#L1389 i.e. translating the well row into an index, it almost feels like this could be thrown with letters.length >3.

But generally agreed with the limit chosen here as the code would have resulted in an Integer overflow

sbesson (approved):
The proposed changes add defensive checks in a couple of locations to prevent overflows replacing them when appropriate with `FormatException` with appropriate error messages.

So far no such overflow had been reported as an issue by the community and the nightly CI builds have been passing with this PR included. @melissalinkert do you have any specific functional testing in mind other than code review. Otherwise, this should be good to merge.

components/formats-bsd/src/loci/formats/ResourceNamer.java
melissalinkert: I debated being more restrictive, but chose the limit of 6 as it's public API and there's a chance that something external is using this method differently from `FakeReader`.

melissalinkert:
I didn't have any particular functional testing in mind beyond verifying that all builds pass. I don't know of data that would actually throw exceptions on any of the new overflow checks, but it seemed like a good idea to add now rather than later. None of the writers are impacted, so conversions shouldn't need to be tested either.

PR #4289, opened 2025-03-07
ImageJ export plugin: add some checks to prevent file appending

### LINKED ISSUE #4221: FIJI plugin: Running Bio-Formats Exporter as a macro appends to existing file by default
See imagesc threads [here](https://forum.image.sc/t/beginners-experiences-on-using-ome-tiff/98077/18) and [here](https://forum.image.sc/t/macro-behaviour-of-bio-formats-exporter/100376) 

When running the Bio-Formats Exporter via the Plugin GUI and selecting an existing file, the user is prompted to confirm that they wish to overwrite the existing file. This matches the behaviour when saving via the command line tools. However when running the Exporter via a macro such as below the data is instead appended to the end of the existing file.

```
run("Bio-Formats Exporter", "save=/tmp/test.ome.tif export compression=Uncompressed");
```

melissalinkert:
This is one approach to fixing #4221.

I could reproduce the original issue when running a macro like:

```
run("Gel");
run("Bio-Formats Exporter", "save=test.ome.tif export compression=Uncompressed");
waitForUser("check file size", "check file size");
run("Gel");
run("Bio-Formats Exporter", "save=test.ome.tif export compression=Uncompressed");
```

When prompted to check the file size, it's ~132KB, at the end it's ~261 KB. With this change, the file size should be unchanged, as it's deleted due to 57ec2b6.

The last two commits add logging/erroring for multi-file datasets (e.g. one timepoint per file), so that files beyond the first are not appended to if the `windowless` flag is set to `true`.

This seemed like the least disruptive change, but other possible approaches include:

- add an `overwrite` option
  * only delete any file (including the first) if `overwrite` is `true` OR `windowless` is `false` and a prompt to delete files was accepted
 - only delete any file (including the first) if `windowless` is `false` and a prompt to delete files was accepted, otherwise stop the export
 - log but automatically delete files if `windowless` is `true` and this is a multi-file export (instead of stopping the export)

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -302,6 +302,9 @@ public class Exporter {
             outfile = new File(dir, name).getAbsolutePath();
             if (outfile == null) return;
         }
+        else {
+          f = new File(outfile);
+        }
 
         if (windowless) {
             if (splitZ == null) splitZ = Boolean.FALSE;
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -686,6 +686,7 @@ public class Exporter {
             if (outputFiles.length > 1) {
                 for (int i = 0; i < outputFiles.length; i++) {
                     if (new File(outputFiles[i]).exists()) {
+                        IJ.error(outputFiles[i] + " already exists");
                         in = true;
                         break;
                     }
--- a/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
+++ b/components/bio-formats-plugins/src/loci/plugins/out/Exporter.java
@@ -705,6 +705,10 @@ public class Exporter {
                     new File(outputFiles[i]).delete();
                 }
             }
+            else if (in) {
+              IJ.error("Some files already exist; delete them and try again.");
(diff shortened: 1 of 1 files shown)


joshmoore:
@tischi, do you want to give you a try? (I can't add you as reviewer since you aren't in any of the org teams.)

sbesson (approved):
Tested with the sample macro from the PR description. Confirmed that with the current Bio-Formats 8.1.1 release, the file size increases every time the exporter is executed. With this PR included, the file is deleted first and the file size is consistent for each execution.

For multi-file exports, I tested the following macro which will create an OME-TIFF fileset with two 17M TIFF files, one per channel.

```
run("Mitosis (5D stack)");
run("Bio-Formats Exporter", "save=/tmp/test.ome.tiff write_each_channel compression=Uncompressed");
waitForUser("check file size", "check file size");
run("Mitosis (5D stack)");
run("Bio-Formats Exporter", "save=/tmp/test.ome.tiff write_each_channel compression=Uncompressed");
```

The second export fails with the following error message

Screenshot 2025-05-19 at 12 06 28

Agreed, more advanced scenarios could be implemented with additional user control depending on the flags. Additional feedback and input would be valuable here to define these workflows.

Overall, the new logic introduced by this PR fixes an erroneous behavior of the exporter functionality and prevents silent appending to existing files. Under some conditions (multi-file exports), the additional checks might result lead to new errors for existing macros.

PR #4291, opened 2025-03-12
bfopen: fix array allocations

### LINKED ISSUE #3254: Colormaps extraction inconsistent in bfopen for MATLAB
In https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/matlab/bfopen.m, there are multiple problems relating to the reading of colormaps.

In particular:
* [line 144](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/matlab/bfopen.m#L144): Initialises the `colorMaps` variable to be `numImages`-by-`numImages`, which doesn't make sense; it should be `numImages-by-1` (following example of `imageList`)
* [line 152 onwards](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/matlab/bfopen.m#L152): stores the color maps in the row of `colorMaps` corresponding to the current series, leaving the other rows of `colorMaps` unused.
* [line 222](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/matlab/bfopen.m#L222): stores the entire `colorMaps` variable for each series, when only the row with index `s` is actually used.

I can create a pull request to fix these and make it consistent with `imageList`, but it will break backwards compatibility. Is this OK?

If so, then I'm not sure how to update the [unit test](https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-gpl/test/matlab/TestBfopen.m).

### LINKED ISSUE #3882: MatLab: Bfopen preallocating cell of wrong size
Issue was raised by a user on forum thread https://forum.image.sc/t/bfopen-in-matlab-preallocates-a-cell-variable-of-the-wrong-size/72045

In the below line of `bfopen` in the matlab toolbox the cell size is being allocated with `colorMaps = cell(numImages); ` which allocates a cell of `numImages * numImages` which is larger than required. In reality this should likely be numImages * numSeries

https://github.com/ome/bioformats/blob/cca8143ffb219d23700ea41d6349052d2ba07e41/components/formats-bsd/matlab/bfopen.m#L152

melissalinkert:
Fixes #3882.

Without this change, testing `bfopen` with any file that has multiple planes should confirm that `colorMaps` is an n x n array, not n x 1. https://www.mathworks.com/help/matlab/ref/cell.html also confirms that `cell(n)` produces a 2D array, not a 1D array.

This also exposes that the indexing of `colorMaps` is incorrect. `colorMaps` is allocated within the loop over series, so one `colorMaps` should be allocated per series (and assignment to `result` later on confirms this). That means that the only index into `colorMaps` should be `i` (the plane index), never `s` (the series index).

Looking more closely at `result` also exposed that the initial allocation can be 4 elements per series, rather than allocating 2 elements and later expanding to 4 to include metadata.

With this change, testing with any file that has multiple series and multiple planes should confirm that the `result` is now as expected, and in particular `result{*, 3}` which is the `colorMaps` for a particular series. Testing with a file that has a lookup table and `isFalseColor() == true` (e.g. `curated/tiff/wayne/Rockes-8bit-indexed.tif`) should show that a color map is still accessible without issue.

### DIFF
--- a/components/formats-bsd/matlab/bfopen.m
+++ b/components/formats-bsd/matlab/bfopen.m
@@ -133,7 +133,7 @@ if planeSize/(1024)^3 >= 2,
 end
 
 numSeries = r.getSeriesCount();
-result = cell(numSeries, 2);
+result = cell(numSeries, 4);
 
 globalMetadata = r.getGlobalMetadata();
 
@@ -146,7 +146,7 @@ for s = 1:numSeries
     bppMax = power(2, bpp * 8);
     numImages = r.getImageCount();
     imageList = cell(numImages, 2);
-    colorMaps = cell(numImages);
+    colorMaps = cell(numImages, 1);
     for i = 1:numImages
         if mod(i, 72) == 1
             fprintf('\n    ');
@@ -156,16 +156,16 @@ for s = 1:numSeries
 
         % retrieve color map data
         if bpp == 1
-            colorMaps{s, i} = r.get8BitLookupTable()';
+            colorMaps{i} = r.get8BitLookupTable();
         else
-            colorMaps{s, i} = r.get16BitLookupTable()';
+            colorMaps{i} = r.get16BitLookupTable();
         end
 
         warning_state = warning ('off');
-        if ~isempty(colorMaps{s, i})
-            newMap = single(colorMaps{s, i});
+        if ~isempty(colorMaps{i})
+            newMap = single(colorMaps{i});
             newMap(newMap < 0) = newMap(newMap < 0) + bppMax;
-            colorMaps{s, i} = newMap / (bppMax - 1);
+            colorMaps{i} = newMap / (bppMax - 1);
         end
         warning (warning_state);
 
--- a/components/formats-bsd/matlab/bfopen.m
+++ b/components/formats-bsd/matlab/bfopen.m
@@ -155,10 +155,11 @@ for s = 1:numSeries
         arr = bfGetPlane(r, i, varargin{:});
 
         % retrieve color map data
+        % transpose tables for compatibility with things like imshow
         if bpp == 1
-            colorMaps{i} = r.get8BitLookupTable();
+            colorMaps{i} = r.get8BitLookupTable()';
         else
-            colorMaps{i} = r.get16BitLookupTable();
+            colorMaps{i} = r.get16BitLookupTable()';
         end
 
         warning_state = warning ('off');


sbesson (commented):
With the current Bio-Formats 8.1.1, the behavior is the following

```
 
>> data=bfopen('10-01.lsm');
Reading series #1
    ....
>> data

data =

  1x4 cell array

    {4x2 cell}    {379 java.util.Hashtable}    {4x4 cell}    {1x1 loci.formats.ome.OMEPyramidStore}

>> data{3}

ans =

  4x4 cell array

    {65536x3 single}    {65536x3 single}    {65536x3 single}    {65536x3 single}
    {    0x0 double}    {    0x0 double}    {    0x0 double}    {    0x0 double}
    {    0x0 double}    {    0x0 double}    {    0x0 double}    {    0x0 double}
    {    0x0 double}    {    0x0 double}    {    0x0 double}    {    0x0 double}
```

With this PR included

```
>> bfCheckJavaPath();
>> data=bfopen('10-01.lsm');
Reading series #1
    ....
>> data

data =

  1x4 cell array

    {4x2 cell}    {379 java.util.Hashtable}    {4x1 cell}    {1x1 loci.formats.ome.OMEPyramidStore}

>> data{3}

ans =

  4x1 cell array

    {3x65536 single}
    {3x65536 single}
    {3x65536 single}
    {3x65536 single}
```

The historical usage of a NxN cell array is certainly incorrect and it makes sense to update this representation to a single array.

In addition to the matrix dimensions, this PR includes 2 additional changes:
- previously the colormaps for a given series would be on a given row of the cell array. With the proposed change, each colormap would be a different row of the Nx1 cell array. From my perspective, this is consistent with the representation of the planes in the first element of the cell array.
- the colormaps returned by the API are now directly stored in the cell array. In the example above, this means the matrix dimensionsa re `3x65536` instead of  `65536x3`. The primary impact with this change is that some of the standard image operation like `imshow` expect a c-by-3 matrix for the colormap - see https://uk.mathworks.com/help/matlab/ref/imshow.html#bvmnrxi-1-map. So `imshow(data{1}{1},data{3}{1});` will fail wiht the current state of this PR but `imshow(data{1}{1},data{3}{1}');` will succeed. Any objection to keeping the transposition in place so that the colormap is in a MATLAB compatible format?

melissalinkert:
> Any objection to keeping the transposition in place so that the colormap is in a MATLAB compatible format?

Should be fixed with 7944f29.

sbesson (approved):

PR #4293, opened 2025-03-14
Remove unnecessary Python script for bumping component versions

sbesson:
The original intent of the script was to simplify the bump of the POM versions as well as the static STABLE_VERSION field in the UpgradeChecker class on release. In the context of the IDR fork of Bio-Formats, additional functionality was added to manage versions for a custom groupId.

As the UpgradeChecker.STABLE_VERSION has been deprecated and the IDR fork is no longer active, this script is redundant and can be fully replaced by `mvn versions:set` which is the current documented workflow for Bio-Formats releases.

https://github.com/search?q=org%3Aome%20bump_maven_version&type=code indicates there are 2 downstream documentation places which refer to this script and which I can update if this PR is approved


melissalinkert:
Generally :+1:, but note too https://github.com/search?q=org%3Aopenmicroscopy%20bump_maven_version&type=code and in particular https://github.com/openmicroscopy/management_tools/blob/master/ci/jobs/BIOFORMATS-DEV-latest-version-bump_0001.sh. I don't know what the impact there might be.

melissalinkert (approved):

PR #4302, opened 2025-04-25
Olympus ScanR: check getCurrentFile is the first value of getUsedFiles

sbesson:
See https://github.com/ome/bioformats/pull/4172


melissalinkert (approved):
All makes sense, and tests look good: https://bf-testing-results.s3.amazonaws.com/2025/2025-04-28/build-status.log

PR #4303, opened 2025-04-25
Fix Unidata repository URLs

melissalinkert:
This is consistent with https://docs.unidata.ucar.edu/netcdf-java/current/userguide/using_netcdf_java_artifacts.html. We have been seeing build errors with Gradle in other repositories that use `https://artifacts.unidata.ucar.edu/content/repositories/unidata-releases`, so this is just a preventative measure.


sbesson (approved):
A few other repositories in the OME organization are making use of the former URL style and might need to be updated similarly - https://github.com/search?q=org%3Aome%20https%3A%2F%2Fartifacts.unidata.ucar.edu%2Fcontent%2Frepositories%2Funidata-releases&type=code /cc @jburel

PR #4304, opened 2025-04-30
Leica LIF: map offsets to images by memory block ID

### LINKED ISSUE #4296: Lif image distorted
See [Image.sc - 110520](https://forum.image.sc/t/lif-sometimes-fine-sometimes-not/110520).
The first few image series look good, but from series 8 on it's just random noise.
Files in `inbox/imagesc-110520`.

melissalinkert:
Fixes #4296.

Data from #4296 includes some extra non-image memory blocks in between the image memory blocks, which causes some images to be read from incorrect offsets. Using the memory block IDs allows the extra blocks to be skipped correctly.

Without this change, `showinf -series 7` or viewing all thumbnails in ImageJ for `inbox/imagesc-110520/AMR1.lif` should show a clearly incorrect image. With this change, the same checks should show an image that corresponds to stitching the previous 6 series.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -127,6 +127,7 @@ public class LIFReader extends FormatReader {
 
   /** Offsets to memory blocks, paired with their corresponding description. */
   private List<Long> offsets;
+  private transient List<String> offsetBlockIDs;
 
   private int[][] realChannel;
   private int lastChannel = 0;
@@ -400,7 +401,7 @@ public class LIFReader extends FormatReader {
     else {
       posInFile = dataOffset + no * planeSize;
     }
-    
+
     // seek instead of skipBytes to prevent dangerous int cast
     in.seek(posInFile);
   }
@@ -443,6 +444,7 @@ public class LIFReader extends FormatReader {
     super.close(fileOnly);
     if (!fileOnly) {
       offsets = null;
+      offsetBlockIDs = null;
       realChannel = null;
       lastChannel = 0;
       lutNames.clear();
@@ -496,6 +498,7 @@ public class LIFReader extends FormatReader {
     in = new RandomAccessInputStream(id);
     in.setEncoding(ENCODING);
     offsets = new ArrayList<Long>();
+    offsetBlockIDs = new ArrayList<String>();
 
     in.order(true);
 
@@ -559,7 +562,10 @@ public class LIFReader extends FormatReader {
       int descrLength = in.readInt() * 2;
 
       if (blockLength > 0) {
-        offsets.add(in.getFilePointer() + descrLength);
+        String memBlockID = DataTools.stripString(in.readString(descrLength));
+        offsets.add(in.getFilePointer());
+        offsetBlockIDs.add(memBlockID);
+        descrLength = 0;
       }
 
       in.seek(in.getFilePointer() + descrLength + blockLength);
@@ -1162,12 +1168,11 @@ public class LIFReader extends FormatReader {
 
     List<Element> imageNodes = new ArrayList<Element>();
     Long[] oldOffsets = null;
-    if (images.getLength() > offsets.size()) {
+    if (images.getLength() != offsets.size()) {
       oldOffsets = offsets.toArray(new Long[offsets.size()]);
       offsets.clear();
     }
 
-    int nextOffset = 0;
     for (int i=0; i<images.getLength(); i++) {
       Element image = (Element) images.item(i);
       Element grandparent = (Element) image.getParentNode();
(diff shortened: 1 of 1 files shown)


sbesson:
See https://bf-testing-results.s3.amazonaws.com/2025/2025-05-01/build-status.log, these changes caused `setId` failures for several LIF datasets

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -127,6 +127,7 @@ private static ImmutableMap<String, Integer> createChannelPriorities() {
 
   /** Offsets to memory blocks, paired with their corresponding description. */
   private List<Long> offsets;
+  private transient List<String> offsetBlockIDs;
```
sbesson: While reading this PR, my understanding that there should be a one-to-one mapping between `offsets` and `offsetBlockIDs` which could have been represented as a `Map<String,Long>`.
Is that accurate? And this would be a larger change that would impact other locations in the reader?

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -559,7 +562,10 @@ protected void initFile(String id) throws FormatException, IOException {
       int descrLength = in.readInt() * 2;
 
       if (blockLength > 0) {
-        offsets.add(in.getFilePointer() + descrLength);
+        String memBlockID = DataTools.stripString(in.readString(descrLength));
+        offsets.add(in.getFilePointer());
+        offsetBlockIDs.add(memBlockID);
+        descrLength = 0;
```
sbesson: Took me several iterations to read as either `descrLength` or `blockLength` will be `0` in the following `in.seek()` call.

Wonder if we should always call `String memBlockID = DataTools.stripString(in.readString(descrLength));` and have `in.seek(in.getFilePointer() + blockLength)`

sbesson (commented):
A few questions from a code review perspective, mindful of the fact this is a large reader and I appreciate the aim to keep changes minimal.

Functionally tested using Fiji. With the current 8.1.1 release of Bio-Formats, I can confirm some of the latest series of the sample files appear as corrupted. With this PR included, the same data loads correctly and the pixel data is consistent with screenshots using Leica LAS X.

components/formats-gpl/src/loci/formats/in/LIFReader.java
melissalinkert: Yes, this could be represented as a `Map<String, Long>` instead of a pair of lists. I wasn't totally sure if this block ID approach would be usable initially, so was going for a minimal change that didn't require changing everywhere that touches `offsets`. Given that tests etc. are passing, I'm OK with spending a little more time here to switch to `Map<String, Long>`, if you think it's worthwhile.

components/formats-gpl/src/loci/formats/in/LIFReader.java
melissalinkert: Fair point, and done in 8aeece0139.

components/formats-gpl/src/loci/formats/in/LIFReader.java
melissalinkert: See what you think of d3683cedb1. I went back and forth a bit on whether `Map<String, Long>` really makes sense, and ultimately for the `openBytes` use case I think it doesn't. `openBytes` doesn't know or care about memory block IDs, as the offsets are meant to already be in the correct order according to series index. Switching `offsets` to a `List` of objects that captures both the memory block ID and actual offset seemed like a reasonable compromise, but happy to hear other thoughts.

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -126,7 +126,7 @@ private static ImmutableMap<String, Integer> createChannelPriorities() {
   // -- Fields --
 
   /** Offsets to memory blocks, paired with their corresponding description. */
```
sbesson: Could be updated  to something along the lines of `Memory blocks including offsets and ID`

components/formats-gpl/src/loci/formats/in/LIFReader.java
```diff
@@ -126,7 +126,7 @@ private static ImmutableMap<String, Integer> createChannelPriorities() {
...
-  private List<Long> offsets;
+  private List<MemoryBlock> offsets;
```
sbesson: Should this internal variable be renamed as 

```
private List<MemoryBlock> blocks;
```
or

```
private List<MemoryBlock> memoryBlocks;
```
?

This means updating a few tens of references across the reader but arguably most of these lines are already updated by this PR.

sbesson (commented):
Last commit introduces the concept of a `MemoryBlock` and converts the internal `List<Long>` capturing the offsets  into a`List<MemoryBlock>`. The rationale is discussed in  https://github.com/ome/bioformats/pull/4304#discussion_r2087088687. Given the requirement for accessing these memory blocks sequentially in `openBytes`, I agree a simple `Map` has implementation shortcomings  and using `List<MemoryBlock>` feels like a more robust solution.

The nightly tests running against the curated QA repository are still passing with this change included.

Thinking of future maintenance, the only remaining question is whether we should also rename the `offsets/oldOffsets/storedOffsets` variables to `memoryBlocks/oldMemoryBlocks/storedMemoryBlocks` to match the type change.

components/formats-gpl/src/loci/formats/in/LIFReader.java
melissalinkert: Done in 18ed364dcc.

components/formats-gpl/src/loci/formats/in/LIFReader.java
melissalinkert: Done in 18ed364dcc, along with a few other places that reference offsets where (memory) blocks is now more appropriate.

sbesson (approved):

PR #4305, opened 2025-05-01
Add reader for Molecular Devices JDCE

melissalinkert:
Backported from private PRs. 

Test plates are in `curated/jdce/`. The `*.jdce` file is the target for `showinf` or import.

Note the `WellContainer` class is new; if adding that as public API (or adding in a minor release) isn't comfortable, I am OK with reworking. The intent of that class was to be able to reduce some of the duplicated well handling across multiple HCS readers, but obviously none of the other readers in this repository use it.

### DIFF
--- a/components/formats-api/src/loci/formats/WellContainer.java
+++ b/components/formats-api/src/loci/formats/WellContainer.java
@@ -1,0 +1,158 @@
+/*
+ * #%L
+ * Top-level reader and writer APIs
+ * %%
+ * Copyright (C) 2020 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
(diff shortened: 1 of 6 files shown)


sbesson (approved):
Code-wise, the reader has gone through several rounds of internal reviews and is identical to the version that was announced in https://www.glencoesoftware.com/blog/2025/02/19/whats-new-in-omero-plus-bioformats-8-0-1-761.html.

The reader has been extensively tested against a set of representative plates. A subset of these have been added to the curated OME QA repository and included in the nightly builds - see https://bf-testing-results.s3.amazonaws.com/2025/2025-05-05/jdce.log for the last successful execution.

`WellContainer` is the same as the version available in https://github.com/glencoesoftware/bioformats2raw/blob/v0.10.0-rc2/src/main/java/com/glencoesoftware/bioformats2raw/WellContainer.java. This gives us at least one other downstream reader (`BioTekReader`) which makes use of this API and possibly other HCS readers in this repository in the future.

joshmoore:
:100: from my side on having repetitive code encapsulated in classes like `WellContainer`. I'd personally consider a different package for it, but I don't know of one that "Just Fits"(TM) so can see deferring that, though as we know, it'll get harder as it's used more.

melissalinkert:
Following some offline discussion, `loci.formats.internal` proposed as the new location for `WellContainer`. 4095848719 renames, and adds a `package.html` that tries to capture the intent of the package, as discussed.

sbesson (approved):

joshmoore (approved):

PR #4308, opened 2025-05-13
Add support for Zeiss XRM (.txm/.txrm) data

melissalinkert:
Backported from private PRs. Data forthcoming, opening this now to check that no other tests are impacted.

This is a single file/single series format that makes use of https://github.com/ome/ome-poi. See also https://www.zeiss.com/microscopy/us/products/x-ray-microscopy.html.

Note that in addition to the reader, there is a new constant added to `POIServiceImpl` to aid file type detection. If that's concerning, can move it elsewhere.

Tentatively scheduling for 8.3.0 to leave plenty of time for review, but happy to hear other opinions.

### DIFF
--- a/components/formats-api/src/loci/formats/readers.txt
+++ b/components/formats-api/src/loci/formats/readers.txt
@@ -109,6 +109,7 @@ loci.formats.in.XLEFReader            # xlef
 loci.formats.in.OlympusTileReader     # omp2info
 loci.formats.in.DCIMGReader           # dcimg
 loci.formats.in.JDCEReader            # jdce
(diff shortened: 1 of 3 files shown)


sbesson (approved):
Successfully tested with the sample `.txm/.txrm` files available from the curated QA repository. After initialization, the reader detect images  as multi-Z images with the ModuloAlongZ annotation which is consistent with the specification described in https://ome-model.readthedocs.io/en/stable/developers/6d-7d-and-8d-storage.html. All original metadata available from the acquisition software is populated in the metadata table and the binary data also reads as expected for all planes.

PR #4309, opened 2025-05-16
ND2Reader Changed the padding for a specific condition to pad to 4 byte boundaries.

odinsbane:
This relates to issue https://github.com/ome/bioformats/issues/3661 The problem appears to be caused by a padding at the end of each scanline. The padding is incorrectly calculated so the bytes shift across the image.

This fix calculates the scanline padding to achieve 4 bytes aligned scanlines. 

The scanline padding is used at one other place in the code. I haven't checked that but it could probably use the same calculation.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Reader.java
@@ -25,6 +25,7 @@
 
 package loci.formats.in;
 
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


melissalinkert:
Thanks, @odinsbane. In order to consider this, we will need a signed [Contributor License Agreement](https://ome-contributing.readthedocs.io/en/latest/cla.html). Could you please send the CLA as indicated, and comment here once that's been done?

odinsbane:
I believe I have sent in a CLA.

odinsbane:
There is another issue that is related to the same reader https://forum.image.sc/t/large-nd2-file-one-system-can-open-but-nobody-else/108565/19 I changed the code so that particular image will not cause a problem. Should I include that in a commit?

Right now a buffer that is too large is trying to be created. For the particular example the buffer is too large, but it doesn't need to be. The required array is not too large.

Should I include those changes in a commit?

melissalinkert:
If the two sets of changes don't conflict, a separate pull request for the other issue will be a little easier to review. If the changes can't be merged together without conflict, though, then feel free to add commits here.

odinsbane:
They would conflict. The second commit is less intrusive. It makes the buffer allocation only be a scanline. Then, when it iterates over rows, the buffer is re-used. I think it is a pretty general performance increase, but it is dealing with the other fix.

melissalinkert:
If there are no more planned changes here, @odinsbane, I'll plan to review later this week.

odinsbane:
@melissalinkert there are no planned changes. There are more complaints about a similar issue so...if it updates Ill just tack it on. :D

melissalinkert:
Including in tonight's test run so we can see if any existing data is impacted by this change. If there are any test failures, I'll remove the `include` label tomorrow, and add details in a review comment.

sbesson:
Note this contribution was not included in the nightly CI builds as it was waiting on approval of the GitHub actions builds. 
After approval, one of the Ant builds is unfortunately failing due the known intermittent issue with the OME  artifactory while retrieving the `jhdf5` artifacts.

```
/home/runner/work/bioformats/bioformats/ant/toplevel.xml:136: The following error occurred while executing this line:
/home/runner/work/bioformats/bioformats/ant/java.xml:50: Could not resolve artifacts: Could not transfer artifact cisd:jhdf5:jar:19.04.1 from/to ome (https://artifacts.openmicroscopy.org/artifactory/maven/): Premature end of Content-Length delimited message body (expected: 8,467,645; received: 7,727,577)
```

The changes from https://github.com/ome/bioformats/pull/4311 are one approach to address this problem

melissalinkert:
> Note this contribution was not included in the nightly CI builds as it was waiting on approval of the GitHub actions builds.
After approval, one of the Ant builds is unfortunately failing due the known intermittent issue with the OME artifactory

Sorry, thanks for catching. Restarting the builds once worked for now, so this should be in tonight's build.

melissalinkert (approved):
This has been included in builds and passing for a few days, most recently https://bf-testing-results.s3.amazonaws.com/2025/2025-05-30/build-status.log.

The relevant test files all look good with this change; I will add them to the test repo separately.

Merging for inclusion in 8.3.0 - thanks again @odinsbane.

PR #4314, opened 2025-05-30
Operetta/Harmony: fix case where index XML is in the parent of the image directory

melissalinkert:
Backported from a private PR.

Operetta/Harmony v7 supposedly can place the `index.idx.xml` file in a directory above the image data. `curated/perkinelmer-operetta/samples/artificial-v7-layout` is an artificial example of the file structure that we have seen with some private data.

Using the example plate, `showinf index.idx.xml` without this change will run without an exception, but the used files list will not contain any TIFFs, the images will be entirely blank, and the pixel types will be `int8`. With this change, all TIFF files in each subdirectory should be on the used files list, there should be no warnings about missing files, and the image data/pixel type should be sensible.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OperettaReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OperettaReader.java
@@ -79,6 +79,8 @@ public class OperettaReader extends FormatReader {
   private MinimalTiffReader reader;
   private ArrayList<String> metadataFiles = new ArrayList<String>();
 
+  private transient Location imagesDir;
+
   // -- Constructor --
 
   /** Constructs a new Operetta reader. */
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Using the artificial v7 sample and Bio-Formats 8.2.0, `showinf` only returns one file, displays many `Could not find valid TIFF file` warnings and only returns black/empty planes independently of the selected series/plane.

With this PR included, all TIFF files are included in the used files list, the pixel type is correctly detected and the plane are properly read 

Screenshot 2025-06-02 at 08 49 56

The artificial sample dataset has been configured for inclusion in the nightly repository tests and has been successfully tested - https://bf-testing-results.s3.amazonaws.com/2025/2025-06-02/perkinelmer-operetta.log

Issue #4313, opened 2025-05-29
Error parsing schema


caylamason:
Hi there,

I am using bftools version 8.2.0. I attempted to validate the XML after successfully converting a czi file to ome.tiff and received the following error:

$bftools/xmlvalid lab_processed/images/sample.ome.tiff
Parsing schema path
http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd
Validating lab_processed/images/sample.ome.tiff
Error parsing schema at http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd
org.xml.sax.SAXParseException: s4s-elt-character: Non-whitespace characters are not allowed in schema elements other than 'xs:appinfo' and 'xs:documentation'. Saw '301 Moved Permanently'.
  	at org.apache.xerces.util.ErrorHandlerWrapper.createSAXParseException(Unknown Source)
	at org.apache.xerces.util.ErrorHandlerWrapper.error(Unknown Source)
	at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
	at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
	at org.apache.xerces.impl.xs.opti.SchemaDOMParser.characters(Unknown Source)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanContent(Unknown Source)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(Unknown Source)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanDocument(Unknown Source)
	at org.apache.xerces.impl.xs.opti.SchemaParsingConfig.parse(Unknown Source)
	at org.apache.xerces.impl.xs.opti.SchemaParsingConfig.parse(Unknown Source)
	at org.apache.xerces.impl.xs.opti.SchemaDOMParser.parse(Unknown Source)
	at org.apache.xerces.impl.xs.traversers.XSDHandler.getSchemaDocument(Unknown Source)
	at org.apache.xerces.impl.xs.traversers.XSDHandler.parseSchema(Unknown Source)
	at org.apache.xerces.impl.xs.XMLSchemaLoader.loadSchema(Unknown Source)
	at org.apache.xerces.impl.xs.XMLSchemaLoader.loadGrammar(Unknown Source)
[...]

PR #4200, opened 2024-06-20
Allow processing multiple files sequentially in a single run, with file names read from stdin.

tstoeter:
This reduces JVM startup overhead when processing thousands of files. If -- is passed as file argument then file names are read line by line from stdin and processed sequentially. Usage information for reading file names from stdin are added with the second commit.

Code compiles with `ant clean jars tools`, compilation with Maven not tested.
All tests using `ant test` have passed.

PR #4316, opened 2025-06-02
xmlvalid: consume OMEXMLServiceImpl.SCHEMA_CLASSPATH_READER

sbesson:
Originally opened as a proposed remediation to #4313 

This allows to map the OME schemas from their URL to the corresponding XSD files available as resources from the specification JAR and perform the XML validation offline / without making HTTP requests.

With the batch improvements added in #4200, it should be possible to test this utility against all the sample OME-XML files from https://github.com/ome/ome-model/tree/master/specification/samples e.g

```
find /path/to/specification/samples -type f -exec ./tools/xmlvalid {} \;
```

Without this PR and with the current HTTP -> HTTPS redirect for the OME schemas URL, the command above should fail. With this PR included, it should work both online and offline

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/XMLValidate.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/XMLValidate.java
@@ -45,6 +45,7 @@ import loci.common.RandomAccessInputStream;
 import loci.common.xml.XMLTools;
 import loci.formats.in.TiffReader;
 import loci.formats.tiff.TiffParser;
+import loci.formats.services.OMEXMLServiceImpl;
 
 /**
  * Attempts to validate the given XML files.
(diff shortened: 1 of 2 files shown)


melissalinkert (approved):

melissalinkert:
With the removal of the https redirect referenced in https://github.com/ome/bioformats/issues/4313#issuecomment-3020458675, I'd expect this change is no longer strictly necessary to fix the validation issue. Do we want to merge anyway though, given that it was working as intended and should help anyone trying to do fully offline validation?

sbesson:
Agreed this change is no longer a requirement to address #4313 (description updated).
I would think this PR still has some value as it allows the `xmlvalid` to be used offline / behind a proxy but happy to hear counterarguments

melissalinkert:
Confirmed that while in airplane mode, I can still successfully validate sample files:

```
$ xmlvalid ~/hcs.ome.xml 
Failed to reach the update site
Parsing schema path
Validating /home/melissa/hcs.ome.xml
No validation errors found.
```

PR #4321, opened 2025-06-04
ND2: swap channels when splitting RGB data

### LINKED ISSUE #3761: ND2: Inverted LUT for RGB images
Issue was raised on forum thread https://forum.image.sc/t/nd2-rgb-import-issue/60901 and a sample file reproducing the issue is available on Zenodo - https://zenodo.org/record/5772393#.YbNTrPnP1Zc

When the ND2 file is opened using Bio-Formats 6.8.0 and Composite mode via the FIJI plugin the resulting image does not match the colour seen when importing into NIS Elements Viewer

melissalinkert:
Fixes #3761.

See also notes in https://github.com/openmicroscopy/data_repo_config/pull/656.

Additional test data is https://zenodo.org/records/12734440 and https://zenodo.org/records/10277961. For all new datasets, comparing `showinf -merge ...` or similar option in ImageJ against screenshot of NIS Elements should show clearly incorrect colors without this change. With this change, `showinf -merge`/ImageJ should match Elements.

This is expected to cause some test failures, which will need to be carefully evaluated to determine if this PR needs to be refined or if configurations need to be updated.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Reader.java
@@ -295,7 +295,7 @@ public class ND2Reader extends SubResolutionFormatReader {
 
         if (split) {
           pix = ImageTools.splitChannels(pix, lastChannel, getEffectiveSizeC(),
-            bpp, false, true);
+            bpp, true, true);
         }
         System.arraycopy(pix, 0, buf, 0, pix.length);
       }
@@ -334,9 +334,8 @@ public class ND2Reader extends SubResolutionFormatReader {
           LOGGER.warn("data line not fully read" + r + " of " + destLength);
         }
         byte[] p2 = ImageTools.splitChannels(pix, lastChannel, getEffectiveSizeC(),
-                bpp, false, true);
+                bpp, true, true);
         System.arraycopy(p2, 0, buf, row*p2.length, p2.length);
-        
       }
 
     }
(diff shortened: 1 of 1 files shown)


melissalinkert:
Another way to approach this would be to not swap channels, but to just set channel colors explicitly to blue/green/red. I'm of multiple minds on which approach is more correct/less confusing. Swapping is a little more consistent with how NIS Elements presents the channels, though, which is red/green/blue order.

sbesson:
https://bf-testing-results.s3.amazonaws.com/2025/2025-06-05/nd2.log - 522 test failures across ~130 ND2 samples (50% of the total number in the curated QA repository). Definitely worth evaluating carefully giving the impact

melissalinkert:
Comparing https://downloads.openmicroscopy.org/images/ND2/aryeh/rfp_h2a_cells_02.nd2 and https://downloads.openmicroscopy.org/images/ND2/zenodo-15579371/823_Clarkia%20pulchella_09-1_DIC_SG.nd2 in particular, the channel splitting occurs on line 336 in both cases. The former file has 2 channels, and is clearly incorrect with the change. The latter file displays correctly with this change. That suggests that reversing the channels isn't a good approach in general.

For `rfp_h2a_cells_02.nd2`:

image

and with this change:

image

note the same pixel is highlighted in both cases, showing the difference in values.

Given that evidence, I am going to close this PR, and focus on setting channel colors in a separate PR instead. Happy to discuss if any disagreement though.

PR #4323, opened 2025-06-06
showinf: configure logging for `-omexml-only` after all arguments parsed

### LINKED ISSUE #4310: Incorrect handling of -omexml-only errors
[This line](https://github.com/ome/bioformats/blob/32a5c82137daa5709ea6679b2e9c896da193a87b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java#L233) makes some errors not be displayed depending on the position of the arguments.

ie. Notice that `-no-pix` is an incorrect flag and therefore should error
`showinf -no-pix -omexml-only image.ome.tif` <- Will display "Invalid parameter -no-pix"
`showinf -omexml-only -no-pix image.ome.tif` <- Will error but without any message

Perhaps the output should be redirected to STDERR or somewhere else. Disabling it completely can lead to confusing situations like the one mentioned above.
Another potential solution is redefining the flag, where the output is omexml only if no errors are found.

Perhaps a note on the documentation is an alternative too?

Thanks,
Alex

melissalinkert:
Fixes #4310.

As noted in the issue, compare `showinf -omexml-only -no-pix test.fake` against `showinf -no-pix -omexml-only test.fake` and note the difference in behavior without this change. Similarly for `showinf -omexml-only -debug test.fake` and `showinf -debug -omexml-only test.fake`.

With this change, the same sets of options should behave the same, independent of the order in which they are specified. `-debug`, `-trace`, and `-omexml-only` are the only options that change the log level from the default `INFO`.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageInfo.java
@@ -230,7 +230,6 @@ public class ImageInfo {
         else if (args[i].equals("-omexml-only")) {
           omexmlOnly = true;
           omexml = true;
-          DebugTools.setRootLevel("OFF");
         }
         else if (args[i].equals("-preload")) preload = true;
         else if (args[i].equals("-ascii")) ascii = true;
@@ -296,6 +295,13 @@ public class ImageInfo {
           return false;
         }
       }
+    }
+
+    // if the -omexml-only flag was used, only print OME-XML, no other logging
+    // the log level is set at the end of argument parsing, so that unknown
+    // any unknown arguments are still logged correctly
+    if (omexmlOnly) {
+      DebugTools.setRootLevel("OFF");
     }
     return true;
   }


sbesson (approved):
Without this change, the output of `showinf` depends on the input arguments for the two examples described: 
- `-omexml-only` + invalid`-no-pix` option
- `-omexml-only` + `-debug`
- `-omexml-only` + `-trace`

With this PR included the output of the command is now independent of the order of the arguments:
- when using an invalid option,  `Found unknown command flag:` is always printed
- when using `-omexml-ony` and `-debug` or `-omexml-ony` and `-trace`, only the OME-XML is printed

There is some intrinsic conflict between the `-debug`, `-trace` and `-omexml-only` options. If the command-line utilities were migrated to use a framework like picocli, these would be candidates for converting into mutually exclusive options. This is outside the scope of this fix which addresses an outstanding inconsistency.

PR #4339, opened 2025-07-15
Fake: add `labelPlanes` option to add human-readable labels to planes

melissalinkert:
Discussed with @sbesson that in the same line as `MakeTestOmeTiff`, it might be useful for some test scenarios to have `FakeReader` show readable ZCT indexes and other metadata included in each plane's pixels. This is encoded already in the "special pixels" in the top left corner, but human-readable values would be nice as well.

`showinf test&labelPlanes=true&...fake` should draw the relevant metadata in white, as `MakeTestOmeTiff` does. This doesn't attempt to wrap the image name if it spans multiple lines. The default behavior without the `labelPlanes` key should remain unchanged.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -38,6 +38,11 @@ import static ome.xml.model.Pixels.getPhysicalSizeXUnitXsdDefault;
 import static ome.xml.model.Pixels.getPhysicalSizeYUnitXsdDefault;
 import static ome.xml.model.Pixels.getPhysicalSizeZUnitXsdDefault;
 
+import java.awt.Font;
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested with a few example of synthetic images e.g. 

```
% ./tools/showinf "test&labelPlanes=true&sizeZ=20&sizeC=3&sizeT=3&series=2.fake" -series 1
```

Screenshot 2025-08-14 at 13 57 09

The primary goal from my side was to simplify the validation of plane reading in the context of https://github.com/glencoesoftware/omero-zarr-pixel-buffer/pull/28. Coupled with `bioformats2raw`, I think this change is sufficient to easy generate test datasets with various modalities and dimensionalities.

Proposing to merge unless @jburel @dominikl can think of features introduced in their investigation worth bringing here.

PR #4294, opened 2025-03-19
Miscellaneous fixes to SlideBook7Reader.java

nicolapapp:
Added a Long Type to fix errors in the metadata reader caused by reading integers greater than 2^31
Enclosed in  try/catch all decoding of numbers from the yaml files
Added support for sldyz and npyz files, which store the data compressed, using the zstd codecs
Added support of 2D timelapse data stored in a single file (SFMT -> Single File Multi Timpoint)
Fixed GetVoxelSize to accont for magnification
Fixed bugs in the usage of the string substring method
Fixed allocation of byte buffer to read a plane. Allocation was twice as large as needed.
Added code to import ROI's annotations when Metadata level is not Minimum
Fixed adding Plane Exposure Time to metadata store
Fixed error: "YAMLException: The incoming YAML document exceeds the limit: 3145728 code points." for very large yaml files by using yaml LoaderOptions setCodePointLimit.

PR #4340, opened 2025-07-23
Update SlideBook7Reader.java to fix handling of negative Interplane Spacing

nicolapapp:
There was a problem setting PhysicalSizeZ if the value was negative. Changed to use the absolute value of Interplane Spacing Uploaded a test file to Zenodo:
https://zenodo.org/records/16355849
Old version will not display PhysicalSizeZ  in the OME Metadata window. New version will show it with a value of 100 for this test case.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/SlideBook7Reader.java
+++ b/components/formats-bsd/src/loci/formats/in/SlideBook7Reader.java
@@ -47,6 +47,7 @@ import java.util.HashMap;
 import java.util.TreeMap;
 import java.util.List;
 import static java.lang.Integer.max;
+import java.lang.Math;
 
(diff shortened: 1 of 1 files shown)


melissalinkert:
As indicated in #4294, and in particular in https://github.com/ome/bioformats/pull/4294#issuecomment-3028168240, we are no longer accepting changes to the `SlideBook7Reader` as it is scheduled for migration out of this repository ahead of the upcoming 8.3.0 release. The pull request to migrate this reader out is forthcoming, and any further information will be communicated there.

melissalinkert:
Please see update in https://github.com/ome/bioformats/pull/4294#issuecomment-3353287414.

PR #4343, opened 2025-07-29
Upgrade from commons-lang:commons-lang to org.apache.commons:commons-lang3

sbesson:
The latest release of commons-lang 2.6 is now over a decade old and is affected by the recent https://nvd.nist.gov/vuln/detail/CVE-2025-48924. As a mitigation, this upgrades the dependency to the latest release of org.apache.commons:commons-lang3.

Testing wise, there are no functional change, this should be a straightforward replacement and the nightly CI builds should keep passing

### DIFF
--- a/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
+++ b/components/bio-formats-tools/test/loci/formats/tools/ImageConverterTest.java
@@ -58,7 +58,7 @@ import loci.formats.out.OMETiffWriter;
 import loci.formats.services.OMEXMLService;
 import loci.formats.tiff.TiffParser;
 
-import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang3.ArrayUtils;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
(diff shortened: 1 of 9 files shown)


melissalinkert:
Makes sense to me. Waiting for a passing build with this included just to be safe, but expect to be fine to merge tomorrow. I did kind of wonder if we could remove this dependency given the very narrow scope in which we're using it, but that's probably a separate discussion for another time.

sbesson:
>  I did kind of wonder if we could remove this dependency given the very narrow scope in which we're using it, but that's probably a separate discussion for another time.

Thanks for raising this question. I am not strongly opinionated but I had a similar thought about the requirement for this dependency. To be specific, the usage of `commons-lang{3}` is limited to `StringUtils.isEmpty`, `StringUtils.isBlank`,  `ArrayUtils.EMPTY_STRING_ARRAY` and `ArrayUtils.isEmpty`.

Note from the OMERO integration perspective, `commons-lang` is declared and used in a few components so even if we reduced the Bio-Formats dependency stack, I assume this upgrade will need to be performed at this level.

melissalinkert (approved):
https://bf-testing-results.s3.amazonaws.com/2025/2025-07-30/build-status.log

PR #4347, opened 2025-07-31
Fix Metamorph .nd encoding

### LINKED ISSUE #4021: Metamorph reader reads .nd files using wrong encoding
Currently, the metamorph nd reader reads nd files as utf-8 (essentially, in MetamorphReader.java, initFile() calls `DataTools.readFile(ndFileName)`; readFile is implemented in ome-common-java/.../DataTools.java and uses Constants.ENCODING = "UTF-8" from ome-common-java/.../Constants.java.

But generating MDA acquisitions on Metamorph (v7.10.5.476 on Windows) with non-ascii stage position names shows that that encoding is incorrect: I believe that the correct encoding is actually Windows-1252, though I cannot find documentation that explicitly states that.  (In particular, € is encoded at 0x80, which distinguishes the encoding from latin-1, and trying to input some non-Windows-1252 characters in position names results in these being replaced by question marks.)  In practice, such characters currently end up being replaced in bioformats by a Unicode replacement character (U+FFFD �).

I would therefore suggest fixing the encoding, e.g. by providing a `DataTools.readFile(String id, Charset charset)` overload and using `readFile(ndFilename, "windows-1252")` in MetamorphReader.java.

melissalinkert:
Fixes #4021.

See data in `curated/metamorph/gh-4021`. That is the `test.nd` file attached to #4021, with a single artificial .tif to allow for easier testing.

Without this change, `showinf -nopix test.nd` should show an incorrect stage name in the original metadata, as reported. With this change, the name should be fixed.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -120,6 +120,8 @@ public class MetamorphReader extends BaseTiffReader {
   private static final String NDINFOFILE_VER1 = "Version 1.0";
   private static final String NDINFOFILE_VER2 = "Version 2.0";
 
+  private static final String ND_ENCODING = "windows-1252";
+
   // -- Fields --
 
   /** The TIFF's name */
@@ -484,7 +486,7 @@ public class MetamorphReader extends BaseTiffReader {
       boolean useWaveNames = true;
 
       ndFilename = ndfile.getAbsolutePath();
-      String[] lines = DataTools.readFile(ndFilename).split("\n");
+      String[] lines = DataTools.readFile(ndFilename, ND_ENCODING).split("\n");
 
       boolean globalDoZ = true;
       boolean doTimelapse = false;
@@ -2279,7 +2281,7 @@ public class MetamorphReader extends BaseTiffReader {
    */
   private String getNDVersionSuffix(Location ndfile) throws IOException {
     ndFilename = ndfile.getAbsolutePath();
-    String[] lines = DataTools.readFile(ndFilename).split("\n");
+    String[] lines = DataTools.readFile(ndFilename, ND_ENCODING).split("\n");
     boolean globalDoZ = true;
     String version = NDINFOFILE_VER1;
     StringBuilder currentValue = new StringBuilder();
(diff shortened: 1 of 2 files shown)


sbesson (approved):
Tested using the provided sample file. With Bio-Formats 8.2.0, I confirm the characters of the stage position name are definitely incorrectly read

```
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats % ~/Downloads/bftools/showinf -nopix -omexml ~/Desktop/bf-data-repo/automated-tests/curated/metamorph/gh-4021/test.nd | grep Stage
Series #0 -- Stage1 "���?":
DoStage: TRUE
NStagePositions: 1
Stage1: "���?"
   <Image ID="Image:0" Name="Stage1 &quot;���?&quot;">
               <Key>DoStage</Key>
               <Key>Stage1</Key>
               <Key>NStagePositions</Key>
               <Key>Stage1 "���?" Comment</Key>
```

With this PR included, the special characters are correctly parsed and stored in the metadata.

```
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats % ./tools/showinf -nopix -omexml ~/Desktop/bf-data-repo/automated-tests/curated/metamorph/gh-4021/test.nd | grep Stage
Series #0 -- Stage1 "µ€œ?":
DoStage: TRUE
NStagePositions: 1
Stage1: "µ€œ?"
   <Image ID="Image:0" Name="Stage1 &quot;µ€œ?&quot;">
               <Key>DoStage</Key>
               <Key>Stage1 "µ€œ?" Comment</Key>
               <Key>Stage1</Key>
               <Key>NStagePositions</Key>
```

It is worth noting that the impact of this change is not limited to the original metadata annotations as the stage position name is also stored in the `Image.Name` attribute as shown above.

As this attribute is tested as part of the automated repository tests via `FormatReaderTest.testImageNames`, this means we can also track future regressions of this behavior as part of our nightly tests against the Metamorph datasets of the curated QA repository. Based on the outcome of https://bf-testing-results.s3.amazonaws.com/2025/2025-08-01/metamorph.log, no other dataset is affected by this change. A closer examination confirms that the image names for all existing sample datasets received so far used alphanumerical characters.

In the absence of a clear documentation (as mentioned in the description of the original issue), the biggest assumption of this proposal is that `windows-1252` must be used for any Metamorph dataset. The alternative possibility is that different versions of the instrument/software/operating systems may write metadata using different encodings. Given this is arguably the first evidence of this encoding issue in over a decade, I think the risk is low and other encodings would have been read incorrectly up to now anyways. 
Happy to get this work released as part of Bio-Formats 8.3.0 and work on future iterations if we receive new datasets.

PR #4351, opened 2025-08-07
GH release workflow: add flag to automatically generate notes

sbesson:
This tries to take advantage of  the `gh release create` options to automatically populate the release created on tags with notes as per  https://ome-contributing.readthedocs.io/en/latest/bioformats-release-process.html#source-code-release


melissalinkert (approved):
Makes sense. Assuming nothing to test here, and we'll just need to double-check during the 8.4.0 release.

PR #4355, opened 2025-08-14
Fix DCIMG indexing over multiple files

zacsimile:
Selection of `z` and `t` was incorrect when multiple files were grouped as a virtual stack. Fixed here. While this is not a problem with the DCIMG reading, but with how I originally assumed `no` worked, I uploaded an example data set where there's drift, so it's easy to see the different between the correct (this PR) and incorrect (current code) indexing: https://zenodo.org/records/16875377.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/DCIMGReader.java
+++ b/components/formats-gpl/src/loci/formats/in/DCIMGReader.java
@@ -123,9 +123,12 @@ public class DCIMGReader extends FormatReader {
   {
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
-    int zp = no % getSizeZ();
-    int tp = (no / getSizeZ()) % getSizeT();
+    int[] zct = getZCTCoords(no);
+    int zp = zct[0];
+    // int cp = zct[1];  
+    int tp = zct[2];
 
+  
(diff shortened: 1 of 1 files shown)


components/formats-gpl/src/loci/formats/in/DCIMGReader.java
```diff
@@ -123,8 +123,8 @@ public byte[] openBytes(int no, byte[] buf, int x, int y, int w, int h)
   {
     FormatTools.checkPlaneParameters(this, no, buf.length, x, y, w, h);
 
-    int zp = no / getSizeT();
-    int tp = no % getSizeT();
+    int zp = no % getSizeZ();
+    int tp = (no / getSizeZ()) % getSizeT();
```
melissalinkert: Have you tried using something like:

```
int[] zct = getZCTCoords(no);
int zp = zct[0];
int tp = zct[2];
```

`getZCTCoords(int no)` is meant to do this calculation, and takes the dimension order into account as well without it being hard-coded. If I read correctly, the result would be equivalent.

zacsimile: Indeed, this works--thanks! I didn't know about this function. Updated.

melissalinkert (approved):
Build passing with latest change: https://bf-testing-results.s3.amazonaws.com/2025/2025-08-21/build-status.log

Thanks, @zacsimile.

PR #4357, opened 2025-08-20
Add reader for TissueFAXS (.tfcyto) data

melissalinkert:
Backported from several private PRs.

Data upload in progress, to `curated/tissuefaxs/`; as indicated in readme, this data is for internal testing only and cannot be made public. There are 4 datasets each of brightfield and fluorescence.

This is a slide format with pyramids/larger images, so expect to use `showinf -noflat` on the .tfcyto files, with cropping as needed.

### DIFF
--- a/components/formats-api/src/loci/formats/readers.txt
+++ b/components/formats-api/src/loci/formats/readers.txt
@@ -109,7 +109,7 @@ loci.formats.in.XLEFReader            # xlef
 loci.formats.in.OlympusTileReader     # omp2info
 loci.formats.in.DCIMGReader           # dcimg
 loci.formats.in.JDCEReader            # jdce
-loci.formats.in.TissuegnosticsReader  # tfcyto
+loci.formats.in.TissueFAXSReader      # aqproj
 loci.formats.in.ZeissXRMReader        # txm, txrm
 
 # multi-extension messes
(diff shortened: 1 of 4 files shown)


components/formats-api/src/loci/formats/readers.txt
```diff
@@ -109,6 +109,7 @@ loci.formats.in.XLEFReader            # xlef
 loci.formats.in.OlympusTileReader     # omp2info
 loci.formats.in.DCIMGReader           # dcimg
 loci.formats.in.JDCEReader            # jdce
+loci.formats.in.TissueFAXSReader      # aqproj
```
sbesson: This could probably be update to `aqproj, tfcyto` since both extensions will be supported

sbesson (changes requested):
Very nitpicky suggestion in the comment of `readers.txt` to document both supported extensions.

The content of the reader has been subject to extensive internal review against a set of datasets which cannot be shared publicly. All these datasets have been added to the curated QA repository for inclusion in the nightly repository tests - see https://bf-testing-results.s3.amazonaws.com/2025/2025-08-22/tissuefaxs.log for the latest execution.

Using a local build of Bio-Formats from the HEAD of this branch, I confirmed that `showinf -nopix` was able to read the file format, detect several multi-resolution images and read the binary data at different resolution levels.

components/formats-api/src/loci/formats/readers.txt
melissalinkert: Done in f9744ca8c5.

sbesson (approved):

PR #4361, opened 2025-08-27
Remove unused detector linking logic

### LINKED ISSUE #4211: break statement always gets executed
https://github.com/ome/bioformats/blob/bf2c4624b1230d9b3d4ecc2ae98bc2adfd74e560/components/formats-gpl/src/loci/formats/in/LeicaHandler.java#L224

No matter what the value of **index** is, the **break** statement will get executed and exit the **for loop**.

melissalinkert:
Fixes #4211.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LeicaHandler.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaHandler.java
@@ -96,7 +96,6 @@ public class LeicaHandler extends BaseHandler {
 
   private int nextChannel = 0;
   private Double zoom, pinhole;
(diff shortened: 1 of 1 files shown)


sbesson (approved):
As suggested in the issue, this logic is currently unused as the `detectorIndices` array is never filled and the bulk of the detector creation/population happens in https://github.com/ome/bioformats/blob/feea2160c779bf4546d33fc2920af1cdf6b23b5c/components/formats-gpl/src/loci/formats/in/LeicaHandler.java#L733-L831

The nightly repository tests were successful with this included. Merging for the next minor release of Bio-Formats

Issue #4093, opened 2023-09-09
bioformats misinterprets some nd files as HCS data if the stage labels match [a-z]\d+


anntzer:
Consider the test nd dataset provided at https://github.com/ome/bioformats/issues/4069#issue-1839504482.  Fiji's bio-formats importer opens it correctly, and inspecting it e.g. with showinf indicates `Series count = 2` and `Image count = 4` on both series, as expected.
Yet, if one opens test.nd and directly edits the position names from "position_a" and "position_b" to e.g. "f1" and "f2", then the dataset is no longer loaded correctly by Fiji (in particular it seems that some of the tiff files just don't show up at all in the final dataset), and showinf reports `Series count = 2` but `Image count = 1`, together with a "RGB mismatch" warning.  As far as I can see, this occurs because the position names match WELL_COORDS (`"\\b([a-z])(\\d+)"`), ultimately leading to taking the `isHCS = true` branch.
I can understand that some degree of guessing based on filenames and position names is needed, but I would guess that the `isHCS = true` branch could be skipped if 1) it leads to ignoring timepoints that are specified by the NTimePoints entry in the .nd file (and that are actually present), and/or 2) if it leads to RGB mismatched data (whereas the non-HCS interpretation is just fine).

dgault:
Thank you @anntzer for opening the issue and investigating the root cause, that is extremely helpful. This is perhaps something we can look at addressing alongside https://github.com/ome/bioformats/issues/4069 at the same time

PR #4363, opened 2025-08-29
Metamorph: read plane positions from multiple files

### LINKED ISSUE #4069: bfconvert misreports stage position when converting from metamorph tiff ND
Consider the attached ND dataset: [test.zip](https://github.com/ome/bioformats/files/12280486/test.zip)
If we delete the nd file (to force bioformats to consider each file individually) and use showinf to display the X, Y stage position (stageX[0], stageY[0]) for each frame individually, we see that the images have been taken at two different positions ((16204.6, 444.7) and (16204.6, -1429.1)).
Yet, after converting the file with bfconvert to an ome.tiff, the PositionY of *both* positions (position_a and position_b) is reported as 444.7, i.e. the second Y position has been lost.  This can also be seen by checking the XMLAnnotations (a bit further down in the ome metadata), which only shows `X position for position #1` and `Y position for position #1`, but not for position 2.
Screenshot 2023-08-07 at 15 47 48
Unless I am mistaken, the coordinates of the second position are thus completely lost by bfconvert.

-----

On a similar theme, the XMLAnnotations only seem to have kept metadata for `Channel #0` (491_Fast illumination), not for `Channel #1` (561_Fast).   (As a side point, perhaps the numbering of positions should be zero-based to be consistent with the channels.)

melissalinkert:
Fixes #4069.

As indicated in the commit message for a9166d7, the only way to get the correct position for a particular plane is to open the file containing the plane and read the relevant IFD tag. This change reads from the first file in each series, in an attempt to balance getting accurate metadata against the performance impact of always reading every file during initialization. For multi-file/multi-series data, though, this is going to slow down initialization (but not `openBytes`) - I really don't know of a way around this.

The dataset referenced in #4069 is already configured under `curated/metamorph/gh-4093`, so see forthcoming config PR for change in behavior.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MetamorphReader.java
@@ -1231,9 +1231,14 @@ public class MetamorphReader extends BaseTiffReader {
         }
 
         if (perSeriesStageX != null && i < perSeriesStageX.length &&
-          perSeriesStageX[i] != null && p < perSeriesStageX[i].length)
+          perSeriesStageX[i] != null && perSeriesStageX[i].length > 0)
         {
-          store.setPlanePositionX(perSeriesStageX[i][p], i, p);
+          if (p < perSeriesStageX[i].length) {
+            store.setPlanePositionX(perSeriesStageX[i][p], i, p);
+          }
+          else {
+            store.setPlanePositionX(perSeriesStageX[i][0], i, p);
+          }
         }
         else if (stageX != null && p < stageX.length) {
           store.setPlanePositionX(stageX[p], i, p);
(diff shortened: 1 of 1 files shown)


melissalinkert:
Build is now passing again with latest changes: https://bf-testing-results.s3.amazonaws.com/2025/2025-09-05/build-status.log

sbesson (approved):
The proposed changes make sense. At the level of the representative Metamorph datasets, the biggest impact is to populate extra plane position metadata as expected. 

Comparing the execution times of the nightly repository tests for the last two weeks with this PR included to the previous runs, I see no obvious performance degradation associated with the parsing. I agree using the first file is probably a fair compromise between metadata availability and initialization peformance.

For situations where quick initialization time is critical and plane metadata is not important, it is always possible to set the `MetadataLevel` to its minimum value e.g. using `-nometa` in the command-line utility.

PR #4368, opened 2025-09-24
Put the scanline padding to 4 bytes on the non-compressed reading.

### LINKED ISSUE #4367: ND2 skewed images
With 8.3.0, https://zenodo.org/records/17186598 displays incorrectly as indicated in https://forum.image.sc/t/nd2-files-still-opening-in-a-skewed-diagonal-format/116644/.

odinsbane:
A previous fix made the scanline padding at 4 byte boundaries. That was only for a specific reading branch. Another branch also requires 4 byte boundaries, but it is not clear why. The method 'getScanlinePadding' should probably be fixed to use 4 byte padding, but I don't know what conditions it worked, so maybe that will break something.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Reader.java
@@ -341,6 +341,7 @@ public class ND2Reader extends SubResolutionFormatReader {
 
     }
     else {
+      scanlinePad = w%4;
       // plane is not compressed
       readPlane(in, x, y, w, h, scanlinePad, buf);
     }
--- a/components/formats-gpl/src/loci/formats/in/ND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Reader.java
@@ -342,7 +342,7 @@ public class ND2Reader extends SubResolutionFormatReader {
     }
     else {
       //The scanline pad is set at 4 byte boundaries described by the pixel size in bytes.
-      scanlinePad = ((bpp * w)%4) / bpp;
+      scanlinePad = ((bpp * getSizeX())%4) / bpp;
       // plane is not compressed
       readPlane(in, x, y, w, h, scanlinePad, buf);
     }
--- a/components/formats-gpl/src/loci/formats/in/ND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Reader.java
@@ -341,7 +341,8 @@ public class ND2Reader extends SubResolutionFormatReader {
 
     }
     else {
-      scanlinePad = w%4;
+      //The scanline pad is set at 4 byte boundaries described by the pixel size in bytes.
+      scanlinePad = ((bpp * w)%4) / bpp;
       // plane is not compressed
       readPlane(in, x, y, w, h, scanlinePad, buf);
     }


odinsbane:
This issue was brought up in a forum thread https://forum.image.sc/t/nd2-files-still-opening-in-a-skewed-diagonal-format/116644

melissalinkert:
Thanks, @odinsbane. The `include` label has been added so this will be included in tonight's regression tests. We should know tomorrow if any existing test data is impacted by this change.

sbesson (changes requested):
This caused several regressions in the nightly repository tests against the ND2 files from the curated QA repository - see https://bf-testing-results.s3.amazonaws.com/2025/2025-09-25/nd2.log

The failures can be reproducing using a build of the command-line utilities from the current HEAD of this PR and one of the public failing samples e.g. [control002.nd2](https://downloads.openmicroscopy.org/images/ND2/jonas/control002.nd2) 

Screenshot 2025-09-25 at 08 55 40

odinsbane:
@sbesson I'm a little confused by github here. You are just requesting changes, you're not proposing any right?  Also, did you write the original "getScanLinePad"? Is there a spec for that to follow or is it just, check things and use a padding that works?

sbesson:
@odinsbane that's correct, my GitHub review only highlights that the proposed fix is currently insufficient as it causes regression while reading other representative ND2 files but I don't have a concrete suggestion.

> Also, did you write the original "getScanLinePad"? Is there a spec for that to follow or is it just, check things and use a padding that works?

As mentioned in https://bio-formats.readthedocs.io/en/latest/formats/nikon-nis-elements-nd2.html, we have several sample files for this file format but no specification document. All the historical changes have been typically made by reverse engineering using the publicly available ND2 Viewer as the ground truth for validation where possible. Note Nikon recently limited the download of this software to its Windows version only.

odinsbane:
I "improved" the criteria. So it works for both cases, but I don't have the whole suite to go through.

odinsbane:
I went through and tested 3 of the images that failed (I think) and they all worked fine now. I couldn't access the other ones though.

- control002.nd2
- JJ1473_conrol_24h_02.nd2
- Exception9_e3.nd2

sbesson:
Thanks @odinsbane, all included PRs are tested nightly against OME's curated QA repository. We'll review the outcome of the non-regression tests and let you know if additional changes are necessary before we can perform code and functional testing.

sbesson:
As per https://bf-testing-results.s3.amazonaws.com/2025/2025-09-26/nd2.log we're down to 4 failures for 2 sample ND2 files. Unfortunately we don't have permissions to share these files submitted more than a decade ago.

Interestingly, the failing tests correspond to the reading of tiles while the entire plane loading is working correctly. I can reproduce this locally using Bio-Formats command-line utility and these files as `showinf -autoscale` works but `showinf -autoscale -crop 0,0,512,512` displays a skewed image.

I wonder if your proposed fix just needs to be adjusted to always compute the scanline padding from the full image width rather than the width of the requested tile i.e.

```diff
-      scanlinePad = ((bpp * w)%4) / bpp;
+      scanlinePad = ((bpp * getSizeY()) % 4) / bpp;
````

odinsbane:
So it looks like two images are still failing. I cannot access them though. 

- /mnt/bioformats-data/curated/nd2/michael/10um skewed.nd2:
- /mnt/bioformats-data/curated/nd2/michael/50um_004 skewed.nd2

odinsbane:
That makes sense, though I suspect it should be `getSizeX()`.

odinsbane:
Looks good now, https://bf-testing-results.s3.amazonaws.com/2025/2025-09-27/nd2.log

sbesson (commented):
Thanks, the changes are indeed passing all the nightly tests against the ND2 samples of the curated OME QA repository. I have added the sample files uploaded to https://zenodo.org/records/17186598 to this repository and added the configuration so that it should be included into the next CI runs. 
With the tests passing, we'll schedule the review of this contribution for inclusion in the next Bio-Formats release.

melissalinkert (approved):

PR #4370, opened 2025-10-01
ND2: don't reset pixel type when the bits per pixel doesn't make sense

### LINKED ISSUE #3667: ND2: Incorrect bits per pixel
This issue was raised through forum post https://forum.image.sc/t/49494/9
A sample file has been received and is available at `inbox/eric-griffis-2021-03-04`

The issue was tested with BF 6.6.0 and occurs the same with and without the chunkmap and native reader format options.

Opening the image planes shows the top quarter of each image displayed compared to NIS Elements:
Screen Shot 2021-03-11 at 15 39 54

In NIS the image is seen as RGB 8bit 1608 x 1608 whereas when read with BF it is read as 16bit 1608 x 1608

melissalinkert:
Fixes #3667.

Without this change, `showinf neuron.nd2` will show incorrect images, channel count, and pixel type as described in #3667. With this change, there should be 3 channels, pixel type of `uint8`, and reasonable images. See screenshots on forthcoming configuration PR.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Handler.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Handler.java
@@ -485,7 +485,7 @@ public class ND2Handler extends BaseHandler {
           bits++;
         }
         try {
-          if (ms0.pixelType == 0) {
+          if (ms0.pixelType == 0 && bits <= 16) {
             ms0.pixelType =
               FormatTools.pixelTypeFromBytes(bits / 8, false, false);
           }


sbesson (approved):
Works as expected. Without this change, the pixel type of the sample data is incorrectly detected and as a result `openBytes` returns a wrong image. With this simple change, the pixel type and the images are consistent with what is displayed in the ND2 Viewer.

No other sample ND2 files were affected by this change as per the nightly repository tests.

PR #4371, opened 2025-10-03
Fix and refactor ModuloC handling in ImageJ plugins

### LINKED ISSUE #3446: Insonsistent / incorrect handling of OME modulo structured annoations
I am looking at the modulo-Datasets here:
https://docs.openmicroscopy.org/ome-model/6.0.0/ome-tiff/data.html#modulo-datasets

Which use the XML structured annotation documented here:
https://docs.openmicroscopy.org/ome-model/6.0.0/developers/6d-7d-and-8d-storage.html

Opening the files with FIJI via the Bioformats plugins seems to show some support for the annotations but it seems to be mostly broken / inconsistent.

issue1
Looking at `FLIM-ModuloAlongC.ome.tiff`with the `Hyperstack` viewer; the lifetime index displayed above the image does not correspond with the `Sub Channel` displayed in the image itself. Using the `Data Browser` viewer a `lifetime`-Slider is displayed that corresponds with the `lifetime`-Index displayed above the image but does also not correspond with the `Sub Channel` displayed in the image (e.g. lifetime 5/8; Sub Channel 1/8).

issue2
Looking at `FLIM-ModuloAlongT-TSCPC.ome.tiff` there is no `lifetime`-Index above the image nor is there a `lifetime`-Slider in the `Data Browser`.

melissalinkert:
Fixes the primary issue in #3446.

With https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/modulo/FLIM-ModuloAlongC.ome.tiff and 8.3.0, open in ImageJ with the `Data Browser` option. Note that:

- there are 4 sliders as expected, but one has no label
- the bottom 2 sliders are for the channels, but the position of these 2 sliders will not match the positions recorded on the images

With the same dataset and this change, all sliders should have a suitable label, and scrolling through each slider should result in a position that matches the position recorded on the images.

Note though that only ModuloC has ever been supported in ImageJ plugins. Testing with any of the ModuloZ or ModuloT examples in https://docs.openmicroscopy.org/ome-model/6.0.0/ome-tiff/data.html#modulo-datasets will not show an extra slider for the modulo dimension. Adding ModuloZ and ModuloT is a lot more UI work in ImageJ, which I am inclined to think is outside the scope of what we can currently prioritize. If any strong feeling that we should keep an issue open for that, then it would make sense to split off as a separate issue from #3446 (which documents an actual bug).

Note also that the signature of `getChannelDimType` in `LociFunctions` has changed, as that method should never have returned `Double` values. Open to discussing whether it's better to deprecate the existing macro function and introduce something new, but `getChannelDimType` without this change is effectively unusable.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java
+++ b/components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java
@@ -41,6 +41,7 @@ import java.util.List;
 import java.util.Vector;
 import java.util.regex.Matcher;
 
+import loci.common.Constants;
 import loci.common.DateTools;
 import loci.common.Location;
 import loci.common.Region;
@@ -213,6 +214,38 @@ public class ImagePlusReader implements StatusReporter {
     }
 
     return stack;
+  }
+
+  public static int[] getSubC(Modulo moduloC, int sizeC) {
+    if (moduloC.length() > 1) {
+      // ordering of lengths and types in subC and subCTypes matters
+      // use the step size of the modulo to determine correct ordering
+      if (Math.abs(moduloC.step - 1) < Constants.EPSILON) {
+        return new int[] {moduloC.length(), sizeC / moduloC.length()};
+      }
+      else {
+        return new int[] {sizeC / moduloC.length(), moduloC.length()};
+      }
+    }
+    return new int[] {sizeC};
+  }
+
+  public static String[] getSubCTypes(Modulo moduloC) {
+    if (moduloC.length() > 1) {
+      String parentType = moduloC.parentType;
+      if (parentType == null) {
+        parentType = FormatTools.CHANNEL;
+      }
+      // ordering of lengths and types in subC and subCTypes matters
+      // use the step size of the modulo to determine correct ordering
+      if (Math.abs(moduloC.step - 1) < Constants.EPSILON) {
+        return new String[] {moduloC.type, parentType};
+      }
+      else {
+        return new String[] {parentType, moduloC.type};
+      }
+    }
+    return new String[] {FormatTools.CHANNEL};
   }
 
   // -- Helper methods - image reading --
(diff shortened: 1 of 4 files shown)


components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java
```diff
@@ -215,6 +216,38 @@ public static ImageStack createStack(List<ImageProcessor> procs,
...
 
+  public static int[] getSubC(Modulo moduloC, int sizeC) {
+    if (moduloC.length() > 1) {
+      // ordering of lengths and types in subC and subCTypes matters
+      // use the step size of the modulo to determine correct ordering
+      if (Math.abs(moduloC.step - 1) < Constants.EPSILON) {
+        return new int[] {moduloC.length(), sizeC / moduloC.length()};
+      }
+      else {
+        return new int[] {sizeC / moduloC.length(), moduloC.length()};
```
gerring, line 227: sizeC (int) / moduloC.length() (int) will take the integer part. Assume this is intentional but is protection needed? I guess not if sizeC % moduloC.length == 0 but it's not clear that is always the case?

gerring (approved):

components/bio-formats-plugins/src/loci/plugins/in/ImagePlusReader.java
melissalinkert: Yes, this is intentional. `sizeC` should always be a multiple of the Modulo length, as the Modulo describes a sub-dimension of the channel count. Any mismatch between the two values is going to cause other problems outside the scope of viewing in ImageJ. This pull request is not attempting to change any actual dimensions, just the ordering of the sub-dimension and true dimension for channels.

@gerring, are you working with Bio-Formats in a particular context, or mostly familiarizing yourself with the code? If the latter, you may wish to read more in-depth about the modulo feature and how it is used, starting with:

- https://ome-model.readthedocs.io/en/stable/developers/6d-7d-and-8d-storage.html
- https://github.com/ome/bioformats/blob/develop/components/formats-api/src/loci/formats/Modulo.java
- https://github.com/ome/bioformats/blob/develop/components/formats-bsd/src/loci/formats/in/ICSReader.java

gouttegd (approved):
/!\ Disclaimer: I discovered the whole moduloC/T/Z concept merely a few hours ago and have never worked with any image that used it.

But testing with the `ModuloAlongC` file from the [Modulo datasets](https://docs.openmicroscopy.org/ome-model/6.0.0/ome-tiff/data.html#modulo-datasets), the behaviour is as expected (contrary to the current behaviour). The “lifetime“ dimension is correctly identified and navigating using the “lifetime” and “channel” sliders display the expected slice.

joshmoore (approved):
Confirmed with @gouttegd that his review here includes a code review without comment (i.e. ✅)

sbesson (approved):

PR #4380, opened 2025-10-28
Calculate resolution counts when generating pyramids with `bfconvert`

### LINKED ISSUE #3644: bfconvert command line tool: Smarter resolution handling
This request for first raised on the imagesc thread: https://forum.image.sc/t/bfconvert-size-of-smallest-pyramid-level/46282/4

There are 2 main requests from this thread:

- The option to 'smart' create pyramid resolutions when only the option `-pyramid-scale`. In this scenario the number of resolutions would not be provided by the user. Instead the converter would create resolutions until a sensible base line (proposed as the tile size) is reached.

- To better handle the scenario when too many resolutions are requested and the below exception is thrown:
`Exception in thread "main" java.lang.IllegalArgumentException: 0 must be non-null and strictly positive`

melissalinkert:
Fixes #3644.

35227e7 fixes the case where `-pyramid-resolutions` may be too large for the `-pyramid-scale` and XY image size. Something like `bfconvert -pyramid-scale 2 -pyramid-resolutions 15 "test&sizeX=2048&sizeY=2048.fake" test.ome.tiff` will show the difference. Without this PR, that should throw an exception as the smallest few resolutions would be 0x0. With this PR, there should be a logged warning that the resolution count was changed, and the conversion should complete without an exception.

b2c1d0d fixes the case where `-pyramid-resolutions` is not specified but `-pyramid-scale` is. Something like `bfconvert -pyramid-scale 2 "test&sizeX=2048&sizeY=2048.fake" test.ome.tiff` should show the difference. Without this PR, no pyramid is generated. With this PR, a whole pyramid with smallest resolution 1x1 should be generated without error.

### DIFF
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -485,7 +485,7 @@ public final class ImageConverter {
     // TODO: there may be other options not compatible with -precompressed
     if (precompressed &&
       (width_crop > 0 || height_crop > 0 ||
-      pyramidResolutions > 1 ||
+      pyramidScale > 1 || pyramidResolutions > 1 ||
       fillColor != null ||
       autoscale
       ))
@@ -589,6 +589,19 @@ public final class ImageConverter {
     } else {
       width = Math.min(reader.getSizeX(), width_crop);
       height = Math.min(reader.getSizeY(), height_crop);
+    }
+
+    // want to generate a pyramid but no resolution count specified
+    // calculate based upon the scale and tile size
+    if (pyramidScale > 1 && pyramidResolutions == 1) {
+      pyramidResolutions = 0;
+      int checkWidth = width;
+      int checkHeight = height;
+      while (checkWidth > 0 && checkHeight > 0) {
+        pyramidResolutions++;
+        checkWidth /= pyramidScale;
+        checkHeight /= pyramidScale;
+      }
     }
 
     if (channel >= reader.getEffectiveSizeC()) {
--- a/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
+++ b/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
@@ -599,7 +599,9 @@ public final class ImageConverter {
       pyramidResolutions = 0;
       int checkWidth = width;
       int checkHeight = height;
-      while (checkWidth > saveTileWidth && checkHeight > saveTileHeight) {
+      while (checkWidth >= (int) Math.max(saveTileWidth, pyramidScale) &&
+        checkHeight >= (int) Math.max(saveTileHeight, pyramidScale))
+      {
         pyramidResolutions++;
         checkWidth /= pyramidScale;
         checkHeight /= pyramidScale;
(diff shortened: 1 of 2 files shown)


sbesson (changes requested):
I can reproduce both behaviors described in the original issue (runtime exception and conversion without pyramid generation) using Bio-Formats 8.3.0 and the example commands provided.

With these changes, both examples produce pyramids with 12 resolution levels in total up to a 1x1 tile. Setting `-pyramid-scale 8` in the second command creates 4 resolution levels.

The second scenario (`-pyramid-scale` without `-pyramid-resolutions`) modifies `bfconvert` to behave more closely the `bioformats2raw` implementation. It raises the question of whether the downsampling should continue until 1x1 or whether the default behaviour should match some of decisions made there i.e.
1. compute the resolutions levels until the smallest resolution size reaches a minimum target size
2. (optionally) add a command-line option to control this minimal target size

In addition to the open implementation questions above, could `ImageConverterTest` be updated to also include some unit tests covering both conversion scenarios?

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
```diff
@@ -591,6 +591,21 @@ public boolean testConvert(IFormatWriter writer, String[] args)
...
 
+    // want to generate a pyramid but no resolution count specified
+    // calculate based upon the scale and image size
+    // the tile size has not been yet been set on the writer,
+    // and the writer needs a MetadataRetrieve with resolutions populated first
+    if (pyramidScale > 1 && pyramidResolutions == 1) {
+      pyramidResolutions = 0;
+      int checkWidth = width;
+      int checkHeight = height;
+      while (checkWidth > saveTileWidth && checkHeight > saveTileHeight) {
```
sbesson: With a command along the lines of `bfconvert -pyramid-scale 2 "test&sizeX=2048&sizeY=2048.fake" test2.ome.tiff -tilex 1024 -tiley 1024` this results in a single resolution.

For tiled based writing, the following should create 2 resolution levels in this scenario and align with the behavior of `bioformats2raw "test&sizeX=2048&sizeY=2048.fake" --target-min-size=1024 test.zarr`

```suggestion
      while (checkWidth >= saveTileWidth && checkHeight >= saveTileHeight) {
```

Maybe this should also check the height/width are greater than `pyramidScale` as well to avoid `0x0` image in the case of stripped images

sbesson (changes requested):
Latest changes look good. One suggestion to align the boundary condition for the smallest resolution size when `-pyramid-resolutions` is unspecified.

It is worth noting there is a difference in the level calculation between writing TIFFs in strips vs tiles so the two following commands yield different number of pyramidal levels

```
bfconvert  "test&sizeX=2048&sizeY=2048.fake" test_tiled.ome.tiff -tilex 1024 -tiley 1024 -pyramid-scale 2 
bfconvert "test&sizeX=2048&sizeY=2048.fake" test_strips.ome.tiff -pyramid-scale 2 
```

Mostly capturing here as we might want to clarify the expectation in the documentation

components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java
melissalinkert: 1e6928e/01b2e7a is one approach to both checks - that will omit any resolutions with a dimension smaller than `pyramidScale` (e.g. a `1x1` resolution), but since `pyramidScale` is expected to be quite small anyway I think that is reasonable.

sbesson (approved):

PR #4392, opened 2025-12-04
A few more ND2 dimension fixes

### LINKED ISSUE #4344: ND2 timelapse plus Z stack not detected correctly
> This issue has been reported by another user in https://forum.image.sc/t/unable-to-open-nikon-nd2-with-bio-formats/38904/5
> 
> 
> 
> A sample file is available at https://zenodo.org/record/7439387#.Y9fA9OzP10t which replicates the issue with Bio-Formats 6.11.1. None of the ND2 reader options impact the behaviour.
> 
> 
> 
> In the case of this sample file there are expected to be 44 timepoints, when read with Bio-Formats the timepoints are split into series, but even then only 20 series are displayed and the remaining timepoints are dropped. 

 _Originally posted by @dgault in [#3313](https://github.com/ome/bioformats/issues/3313#issuecomment-1408615365)_

Splitting from #3313 as this specific file wasn't fixed by #4341, unlike the rest of the data referenced in #3313. In this case, the series count is detected as 20 (it should be 1), the `SizeT` is 1 (it should be 44), while the `SizeZ` (3) and `SizeC` (5) counts are correct. The difference between 20 and 44 suggests that there is some other plane count calculation issue occurring.

melissalinkert:
Fixes #4344.

Both public datasets referenced in #4344 should now have correct ZCT dimensions and series counts with this change. See forthcoming configuration PR for reference screenshots from NIS Elements.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ND2Reader.java
+++ b/components/formats-gpl/src/loci/formats/in/ND2Reader.java
@@ -2402,9 +2402,14 @@ public class ND2Reader extends SubResolutionFormatReader {
             stampIndex = getIndex(coords[0], !split ? coords[1] : 0, 0);
           }
           stampIndex += (coords[2] * getSeriesCount() + i) * zcPlanes;
-          if (tsT.size() == getImageCount()) stampIndex = n;
+          if (tsT.size() == getImageCount()) {
+            stampIndex = n;
+          }
           else if (tsT.size() == getSizeZ()) {
             stampIndex = coords[0];
+          }
+          else if (tsT.size() == getSizeZ() * getSizeT()) {
+            stampIndex = n / getEffectiveSizeC();
           }
           if (stampIndex < tsT.size()) {
             double stamp = tsT.get(stampIndex).doubleValue();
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Tested against the set of sample ND2 files included in the curated QA repository notably the public ones.

In terms of dimensions, https://zenodo.org/records/16921650 is detected as 3 channels, 1 Z-section and 405 timepoints with Bio-Formats 8.3.0. With this PR included, the file is detected as 3 channels, 81 timepoints and 5 Z sections as expected

The plane timestamps of several ND2 sample files from the curated QA repository are also modified by these changes. The new values of `Plane.DeltaT` have been compared and validated against the ground truth values exported from NIS Elements.

The sample datasets from https://zenodo.org/records/7439387 and https://zenodo.org/records/16921650 uploaded as part of the original issue have been successfully added to the curated repository and configured and are now included in the nightly repository tests for non-regression testing - see https://bf-testing-results.s3.amazonaws.com/2026/2026-01-12/nd2.log for the latest execution

Issue #4264, opened 2025-01-16
.vmic reader


kwiechen:
Hello,

before making a PR of: https://forum.image.sc/t/vmic-whole-slide-image-reader/106485/3

I am using code from https://github.com/usnistgov/pyramidio to read the deepzoom image pyramid. Where should these files go?

Best regards,

Kai

melissalinkert:
Thank you for asking this, @kwiechen.

Please do not copy in files from pyramidio; instead, add it as a dependency in the appropriate pom.xml file. The dependency declaration is provided by pyramidio:

https://github.com/usnistgov/pyramidio?tab=readme-ov-file#maven-dependency

but replace the `...` with the correct version number. This declaration should go in `components/formats-bsd/pom.xml` if your reader is added under `components/formats-bsd/` (and thus BSD-licensed), or in `components/formats-gpl/pom.xml` if your reader is added under `components/formats-gpl/` (and thus GPL-licensed).

Before opening a pull request with a new reader, please make sure to read the following:

- https://ome-contributing.readthedocs.io/en/latest/third-party-policy.html
- https://forum.image.sc/t/bio-formats-update/101734
- [A successful pull request that adds a new reader](https://github.com/ome/bioformats/pull/4255)

As all of those pages indicate, we will do our best to evaluate and include your pull request. However, if the proposed changes need substantial work, or are too complex to review and test in a reasonable amount of time, we will let you know and close the PR. This is not to discourage you from opening a pull request, but rather to be clear that we may not have the capacity to spend substantial time on this new reader. If that turns out to be the case, we’ll do our best to provide suggestions for how to proceed before closing.

kwiechen:
Thank you for the information, @melissalinkert 

The signed Contributor License Agreement was sent to contributors@openmicroscopy.org

An example .vmic dataset was uploaded to Zenodo.org

The .vmic wsi files (https://openslide-users.andrew.cmu.narkive.com/f3mFAKMI/precipoint-vmic-driver-and-deepzoom) are nested zip archives with a deepzoom image pyramid. In order to read pixel data and metadata without unzipping, I have used the java.nio.file.FileSystem. In addition .vmic files may be generated in sparse scan mode. Therefore the pyramidio deepzoom reader was (slightly) modified to work with the zip file system provider and missing image data.

Best regards,

Kai

melissalinkert:
As a place to start then, how about:

- push a branch with your changes to pyramidio
[...]

PR #4397, opened 2026-01-08
.vmic WSI reader

kwiechen:
Hello,

the .vmic WSI file (#4264) is a nested Zip archive including a deep zoom pyramid. .vmic format information is from:

https://github.com/openslide/openslide/issues/168
https://lists.andrew.cmu.edu/pipermail/openslide-users/2015-December/001164.html

The VmicReader reads the deep zoom pyramid using:

https://github.com/usnistgov/pyramidio

Reading data from WSI files < 2 GB is possible directly using java.nio.file.FileSystems. The reader unzips the outer zip container into a temporary file if size is > 2 GB.

Example data: https://zenodo.org/records/14714280

Best regards,

Kai


kasper-cancilico:
@sbesson @melissalinkert following up on Melissa's note connecting this to #4453. Kai and I have agreed to consolidate onto #4397 (mine adds a synthetic-fixture unit test and a few robustness fixes on top of his reader), and since it's milestoned for 9.0.0 I wanted to ask whether it could get a first review. That would tell us what to change before I open the follow-up against Kai's branch, and it avoids duplicating effort. No urgency on merging. Thanks!

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+
+    private FileSystem inner_zipfs;
+    private File innerZipFile;
+    boolean unzippedOuterZip = false;
+
+    // -- Constructor --
+
+    public VmicReader() {
+        super("vmic", new String[]{"vmic"});
+        domains = new String[]{FormatTools.GRAPHICS_DOMAIN};
```
melissalinkert, line 93: Since this is whole slide data, I'd expect `FormatTools.HISTOLOGY_DOMAIN` here.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+import loci.common.Location;
+import loci.formats.*;
+import loci.formats.gui.AWTImageTools;
+import loci.formats.meta.MetadataStore;
+import ome.units.UNITS;
+import ome.units.quantity.Length;
+import org.w3c.dom.*;
```
melissalinkert, line 45: Here and in other imports above, we prefer to have necessary imports enumerated instead of using `*`.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+
+        if (inner_zipfs != null) {
+            if (unzippedOuterZip) {
+                Path tempFileToDelete = Paths.get(innerZipFile.getAbsolutePath());
+                if (Files.exists(tempFileToDelete)) {
+                    try {
+                        Files.delete(tempFileToDelete);
+                    }
+                    catch (IOException e) {
+                        System.out.println(e.getMessage());
```
melissalinkert, line 152: Please use the existing logging infrastructure instead of `System.out.println`.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+
+        return buf;
+    }
+
+    /* @see loci.formats.IFormatReader#close(boolean) */
+    @Override
+    public void close(boolean fileOnly) throws IOException {
+        super.close(fileOnly);
+
+        if (inner_zipfs != null) {
```
melissalinkert, line 144: If `close(false)` is called, all fields should be reset to their original value, including e.g. `filesFolder`. I would expect that lines 145-158 should only happen if `close(false)` was called (and not `close(true)`), but if there is a reason that this is always done then it would be helpful to clarify.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+            }
+
+            inner_zipfs = null;
+            innerZipFile = null;
+        }
+    }
+
+    /* @see IFormatReader#getResolutionCount() */
+    @Override
+    public int getResolutionCount() {
```
melissalinkert, line 164: Overriding `getResolutionCount()` should be unnecessary.

I note that `showinf -nopix -noflat vmic-sample-small-1.vmic` shows a single pyramid of 6 resolutions. I would then expect `showinf -nopix vmic-sample-small-1.vmic` to show 6 separate series (one resolution per series), but instead I see a single series.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+        FormatTools.assertId(currentId, true, 1);
+        return core.get(0, 0).resolutionCount;
+    }
+
+    /* @see IFormatReader#setResolution(int) */
+    @Override
+    public void setResolution(int no) {
```
melissalinkert, line 171: Similar to `getResolutionCount()`, I would expect overriding `setResolution(int)` to be unnecessary.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+    /* @see loci.formats.FormatReader#getThumbSizeY() */
+    @Override
+    public int getThumbSizeY() {
+        return core.get(0, 0).thumbSizeY;
+    }
+
+    /* @see loci.formats.FormatReader#initFile(String) */
+    @Override
+    public void initFile(String id) throws FormatException, IOException {
+        setFlattenedResolutions(false);
```
melissalinkert, line 206: This will definitely result in some unexpected behavior. A reader that supports reading pyramids is expected to support both flattened (one resolution per series) and unflattened (all resolutions as part of a single series), without removing resolutions in the flattened case.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+
+        try (ZipFile outerZipFile = new ZipFile(Location.getMappedId(id))) {
+            ZipEntry innerZipEntry = outerZipFile.getEntry(INNER_CONTAINER);
+
+            // unzip outer container only if inner container > 2 GB
+            if (innerZipEntry.getSize() > Integer.MAX_VALUE - 1) {
+                innerZipFile = File.createTempFile("Image", ".vmici");
+                innerZipFile.deleteOnExit();
+                try (InputStream innerZip = outerZipFile.getInputStream(innerZipEntry)) {
+                    OutputStream out = new FileOutputStream(innerZipFile);
```
melissalinkert, line 219: Any reason not to use the same `try (OutputStream out = ...)` pattern as is used elsewhere?

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+            for (int i = 0; i < length; i++) {
+                Node node = childNodes.item(i);
+                if ("Size".equals(node.getNodeName())) {
+                    NamedNodeMap attributes = node.getAttributes();
+                    w = attributes.getNamedItem("Width").getNodeValue();
+                    h = attributes.getNamedItem("Height").getNodeValue();
+                }
+            }
+            width = Integer.parseInt(w);
+            height = Integer.parseInt(h);
```
melissalinkert, line 279: I imagine it's unlikely, but this should check if `w` and `h` are non-null.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+        }
+
+        initBasicMetadata();
+        initExtendedMetadata();
+    }
+
+    private void initBasicMetadata() throws IOException {
+        try {
+            Path entry = inner_zipfs.getPath(DZI_FILE);
+            InputStream is = Files.newInputStream(entry);
```
melissalinkert, line 252: `try (InputStream is = ...`) could also be used here.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+        MetadataTools.populatePixels(store, this);
+
+        HashMap<String, String> metaDataMap = new HashMap<>();
+        metaDataMap.put("ShortName", null);
+        metaDataMap.put("Magnification", null);
+        metaDataMap.put("PixelPerMicron", null);
+
+        try {
+            Path entry = inner_zipfs.getPath(EXTENDED_METADATA);
+            InputStream is = Files.newInputStream(entry);
```
melissalinkert, line 343: As above, `try (InputStream is = ...)` can be used.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+                        }
+                    }
+                }
+            }
+
+            Length pixelsize = FormatTools.createLength(1 / Double.parseDouble(metaDataMap.get("PixelPerMicron")), UNITS.MICROMETER);
+            store.setPixelsPhysicalSizeX(pixelsize, 0);
+            store.setPixelsPhysicalSizeY(pixelsize, 0);
+            // IDs must not contain white spaces ??
+            store.setInstrumentID("PreciPoint", 0);
```
melissalinkert, line 377: `store.set*(...)` methods follow the OME-XML schema. See https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Instrument for `Instrument`, and https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Instrument_ID for the expected format of `ID` in particular.

A more standard way to model that would be something like:

```
store.setInstrumentID("Instrument:0", 0);
store.setMicroscopeManufacturer("PreciPoint", 0);
```

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+            store.setObjectiveSettingsID(metaDataMap.get("ShortName"), 0);
+            store.setObjectiveID(metaDataMap.get("ShortName"), 0, 0);
```
melissalinkert, line 379: Similar to how `Instrument` is described, see https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Objective. A more standard way to model this would be:

```
String objectiveID = MetadataTools.createLSID("Objective", 0, 0);
store.setObjectiveID(objectiveID, 0, 0);
store.setObjectiveModel(metaDataMap.get("ShortName"), 0);
store.setObjectiveSettingsID(objectiveID, 0);
```

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+
+                        String nn = innerNode.getNodeName().replace("ObjectScanConfig:", "");
+                        if (metaDataMap.containsKey(nn)) {
+                            metaDataMap.put(nn, innerNode.getTextContent());
+                        }
+                    }
+                }
+            }
+
+            Length pixelsize = FormatTools.createLength(1 / Double.parseDouble(metaDataMap.get("PixelPerMicron")), UNITS.MICROMETER);
```
melissalinkert, line 373: I expect it's unlikely, but this should check for null when calling `metaDataMap.get("PixelPerMicron")`.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+            store.setPixelsPhysicalSizeX(pixelsize, 0);
+            store.setPixelsPhysicalSizeY(pixelsize, 0);
+            // IDs must not contain white spaces ??
+            store.setInstrumentID("PreciPoint", 0);
+            store.setObjectiveSettingsID(metaDataMap.get("ShortName"), 0);
+            store.setObjectiveID(metaDataMap.get("ShortName"), 0, 0);
+            store.setObjectiveNominalMagnification(Double.parseDouble(metaDataMap.get("Magnification")), 0, 0);
```
melissalinkert, line 380: As with the `PixelPerMicron` key, the value for `Magnification` could possibly be null.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+                                           int column, int row) throws IOException {
+        File levelFolder = new File(filesFolder, Integer.toString(level));
+        File tile = new File(levelFolder, column + "_" + row + "." + format);
+
+        String s = tile.getPath();
+        Path entry = inner_zipfs.getPath(s);
+
+        if (Files.exists(entry)) {
+            InputStream is = Files.newInputStream(entry);
+            try (ImageInputStream iis = ImageIO.createImageInputStream(is)) {
```
melissalinkert, line 512: Is there a reason to re-implement ImageIO logic here, instead of making use of existing Bio-Formats readers that use ImageIO (e.g. `JPEGReader`)? That would also allow removal of most of the `BufferedImage` handling, since the existing ImageIO-based readers already unpack `BufferedImage` to a byte array.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+ * https://lists.andrew.cmu.edu/pipermail/openslide-users/2015-December/001164.html
+ */
+
+public class VmicReader extends SubResolutionFormatReader {
+    private static final String INNER_CONTAINER = "Image.vmici";
+    public static final String DZI_FILE = "dzc_output.xml";
+    private static final String EXTENDED_METADATA = "VMCF/config.osc";
+    public static final String DZI_FILES = "dzc_output_files";
+
+    private File filesFolder;
```
melissalinkert, line 77: `File` variables will break memoization, which `showinf -nopix -noflat -cache /path/to/file.vmic` can demonstrate. For `filesFolder`, it should be enough to remove this line, and replace use of `filesFolder` with `DZI_FILES`.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+    private int tileSize;
+    private int overlap;
+    private String format;
+    private int width;
+    private int height;
+    private int maxLevel;
+
+    private FileSystem inner_zipfs;
+    private File innerZipFile;
```
melissalinkert, line 86: Similar to `filesFolder` above, the `File` variable is a problem for memoization. If `innerZipFile` is needed outside of `initFile`, you could store the `String` path instead.

components/formats-gpl/src/loci/formats/in/VmicReader.java
```diff
@@ -0,0 +1,564 @@
...
+
+    private File filesFolder;
+    private int tileSize;
+    private int overlap;
+    private String format;
+    private int width;
+    private int height;
+    private int maxLevel;
+
+    private FileSystem inner_zipfs;
```
melissalinkert, line 85: This will also cause a problem for memoization, but is a little less straightforward to fix. The most consistent with other readers would be to make this field `transient`, then override the `reopenFile()` method to set `inner_zipfs` appropriately. `ZipReader` and `MinimalTiffReader` are two examples that follow a similar pattern.

melissalinkert (changes requested):
Thanks @kwiechen and @kasper-cancilico for reviving this discussion, and apologies for the delay.

Please see initial inline review comments; also please revert the move of `SubResolutionExample.java`.

In places where there is some unexpected/undesired behavior, I've tried to note commands you can run on your own to reproduce that behavior. Another test you can run is to take the files from https://zenodo.org/records/18241389, put them in a directory, and then run (from the root of this repo):

```
$ cd components/test-suite
$ ant -Dtestng.memory=2g -Dtestng.directory=/path/to/vmic/data -Dtestng.configDirectory=/path/to/vmic/data test-config
$ ant -Dtestng.memory=2g -Dtestng.directory=/path/to/vmic/data -Dtestng.configDirectory=/path/to/vmic/data test-automated
```

See https://bio-formats.readthedocs.io/en/latest/developers/commit-testing.html#automated-tests for details. That is what we would do to get data into our test repo, so a prerequisite for merging is that both of those `ant ...` commands run without errors.

Finally, if you are planning to add commits here, @kasper-cancilico, we would need a signed [Contributor License Agreement](https://ome-contributing.readthedocs.io/en/latest/cla.html) (we already have one for @kwiechen).

PR #4394, opened 2026-01-06
Bump org.openmicroscopy:ome-model from 6.5.1 to 6.5.3

sbesson:
See https://github.com/ome/ome-model/releases/tag/v6.5.2 and https://github.com/ome/ome-model/releases/tag/v6.5.3

PR #4401, opened 2026-01-16
Fix image ROI reference population in Zeiss LSM and MIAS readers

sbesson:
See #4394 and the failures reported in https://github.com/ome/bioformats/pull/4394#issuecomment-3718441504

When populating the OME metadata via `setImageRoiRef`, the Zeiss LSM and MIAS reader currently make use of the same index for the ROIRef as the ROI index. For multi-series, this means the setters are typically called as:

```
setImageROIRef("ROI:0",0,0);
setImageROIRef("ROI:1",0,1);
setImageROIRef("ROI:2",0,2);
setImageROIRef("ROI:3",1,3);
setImageROIRef("ROI:4",1,4);
setImageROIRef("ROI:5",1,5);
```

The sequence above was working with the current `ome-xml` implementation which was largely ignoring the `roiRefIndex` in `setImageROIRef` and using the order of the calls. With the changes proposed in https://github.com/ome/ome-model/pull/226 and released in `org.openmicroscopy:ome-xml:6.5.2` the index is now taken into account.

This change updates the readers to separate the `roiIndex` from the `roiRefIndex` and ensure each series starts with index 0.

Together with #4394, I expect this should address the failures in `mias` and `zeiss-lsm` although we likely still want to fix the underlying OME-XML library to support the scenario above which has been functional for years - see https://github.com/ome/ome-model/pull/229.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/MIASReader.java
+++ b/components/formats-gpl/src/loci/formats/in/MIASReader.java
@@ -910,6 +910,7 @@ public class MIASReader extends FormatReader {
 
           int[] position = getPositionFromFile(file);
           int well = position[0];
+          int roiRefIndex = 0;
 
           if (name.endsWith("detail.txt")) {
             String data = DataTools.readFile(file);
@@ -925,7 +926,8 @@ public class MIASReader extends FormatReader {
 
             for (int j=start+1; j<lines.length; j++) {
               populateROI(columnNames, lines[j].split("\t"), well,
-                nextROI++, position[1], position[2], store);
+                nextROI++, position[1], position[2], store, roiRefIndex);
+              roiRefIndex++;
             }
           }
           else if (name.endsWith("AllModesOverlay.tif")) {
@@ -942,11 +944,15 @@ public class MIASReader extends FormatReader {
               store.setChannelColor(colors[position[3]], s, position[3]);
             }
             if (position[3] == 0) {
-              nextROI += parseMasks(store, well, nextROI, file);
+              int nOverlays = parseMasks(store, well, nextROI, file, roiRefIndex);
+              nextROI += nOverlays;
+              roiRefIndex += nOverlays;
             }
           }
           else if (name.endsWith("overlay.tif")) {
-            nextROI += parseMasks(store, well, nextROI, file);
+            int nOverlays = parseMasks(store, well, nextROI, file, roiRefIndex);
+            nextROI += nOverlays;
+            roiRefIndex += nOverlays;
           }
         }
       }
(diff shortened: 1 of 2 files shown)


components/formats-gpl/src/loci/formats/in/MIASReader.java
```diff
@@ -910,6 +910,7 @@ else if (name.endsWith("overlay.tif")) {
 
           int[] position = getPositionFromFile(file);
           int well = position[0];
+          int roiRefIndex = 0;
```
melissalinkert: I think this should be initialized to the existing ROI ref count for the relevant `Image` (with index `well`). There can be multiple analysis files for the same well, e.g. one `*detail.txt` and one `*AllModesOverlay.tif` per channel.

components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
```diff
@@ -1614,6 +1614,7 @@ protected void parseOverlays(int series, long data, String suffix,
 
       String roiID = MetadataTools.createLSID("ROI", i);
       String shapeID = MetadataTools.createLSID("Shape", i, 0);
+      int roiRefIndex = i - totalROIs;
```
melissalinkert: I'm not sure this is quite right. `parseOverlays` can be called multiple times per series, which would mean that `roiRefIndex` starts at 0 for every call, potentially overwriting previous references that were set on the same series. Using the existing ROI ref count for `Image` index `i` might be the easiest?

melissalinkert (changes requested):

components/formats-gpl/src/loci/formats/in/ZeissLSMReader.java
sbesson: [0a8f829](https://github.com/ome/bioformats/pull/4401/commits/0a8f829627da36692b09f2693c1258b21fc85bcc) is one way to address this by defining a transient `roiRefIndex` and resetting it to `0` for each series before the calls to `parseOverlays`. Is that what you were thinking of?

sbesson:
After re-reading carefully the MIASReader logic, I was reminded of the  following logic https://github.com/ome/bioformats/blob/eb8fd308b0a1004596899247bc0505b9ee6a97f4/components/formats-gpl/src/loci/formats/in/MIASReader.java#L1160-L1170 which was truly designed for OMERO imports and consumed in https://github.com/ome/omero-blitz/blob/5bf40c91b650586f4fdad4eb8827e682cc4eb683/src/main/java/ome/services/blitz/repo/ManagedImportRequestI.java#L895-L910

The MIAS ROI handling will require more careful attention including some validation win OMERO. To avoid creating a bottleneck, I am proposing the following progression:
- close this PR
- review if https://github.com/ome/ome-model/pull/229 alone suffices to fix all repository failures initially reported in https://github.com/ome/bioformats/pull/4394#issuecomment-3718441504
- open a follow-up PR updating the Image ROI reference indexing in Zeiss LSM reader 
- investigate separately the MIAS ROI parsing both using Bio-Formats command-line tools and OMERO and test the impact of ROI reference indexing changes.

PR #4402, opened 2026-01-19
Review the list of cdm-core exclusions in formats-gpl

sbesson:
Most of these exclusions correspond to older versions of the library, likely before the netcdf/cdm-core split and are no longer applicable.

To test this PR, compare `mvn dependency:list` with and without this PR  and confirm the dependencies are identical.

### DIFF
--- a/components/formats-gpl/pom.xml
+++ b/components/formats-gpl/pom.xml
@@ -69,10 +69,6 @@
       <version>5.6.0</version>
       <exclusions>
         <exclusion>
-          <groupId>com.amazonaws</groupId>
-          <artifactId>aws-java-sdk-s3</artifactId>
-        </exclusion>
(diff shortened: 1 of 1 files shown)


sbesson:
As a side note, I tested the upgrade of `cdm-core` to the latest release `5.9.1`. The only difference is the migration from `commons-math:commons-math` to `org.apache.commons:commons-math3`. Upgrading this might be a candidate for a follow-up PR

melissalinkert (approved):
Makes sense, and comparing `mvn dependency:list` shows no difference. Just holding off on merging until this has been in a passing nightly build.

melissalinkert:
https://bf-testing-results.s3.amazonaws.com/2026/2026-01-20/build-status.log

PR #4406, opened 2026-01-27
Fix Javadoc warnings

sbesson:
Backported from a private PR

Reviews all the Javadoc warnings currently printed as a results of the `ant docs` target, currently 73 warnings as of https://github.com/ome/bioformats/actions/runs/21302506093

With these changes, the generated Javadoc should be unchanged and no additional warning should be printed in the command output


melissalinkert (approved):
No more warnings for Java 11, which is great. Java 8 does have some warnings, but I imagine we're OK with ignoring those right now.

Java 17 has just 2 warnings left, which I think we can address separately at a later date:

- `[javadoc] warning: URL http://docs.oracle.com/javase/8/docs/api/element-list was redirected to https://docs.oracle.com/en/java/javase/25/ -- Update the command-line options to suppress this warning.`
- `[javadoc] /home/runner/work/bioformats/bioformats/components/formats-api/src/loci/formats/FormatReader.java:1014: warning: Tag {@link}: nested link
  [javadoc]    * {@link #setFillColor(Byte)}, then 0 is returned.`

Probably we'll want to give the other components a similar review ahead of 9.0.0, and looking at build logs it might be good to review deprecation warnings again as well. I'll open a separate issue for that in the 9.0.0 milestone.

PR #4414, opened 2026-02-17
ImageReader: do not check file handle for in-memory sources

### LINKED ISSUE #4413: ImageReader with mapFile logs an unexpected error
I am currently facing an issue when attempting to load an image using the `ImageReader` with the `mapFile` procedure from a bytes array. The error that is logged to the console is
```bash
14:35:05.401 [main] ERROR loci.formats.ImageReader -- File has length 0 and may be corrupt
```
This does not stop the processing pipeline, but it does not reflect the real behaviour of the `ImageReader`. Indeed the code continues without any issues afterwards. In other words, I am able to load the data from the image file.

I narrowed down the error to be coming from this piece of code. 
https://github.com/ome/bioformats/blob/eddfa670652145530db288bc8d4aa2f3b245c75a/components/formats-api/src/loci/formats/ImageReader.java#L180-L187
When the `mapFile` procedure is used, the `Location.checkValidId(id)` does not raise an IOException since the `getMappedFile(id)` returns something.  When the code continues, the `Location(id).length()` gets called and this returns 0 since the file does not exist on the disk. In this case, the error log is emitted. I believe that this log should not be emitted and that there should rather be a second check that the length of the mapped file is not equal to 0. 

For instance, I think that this piece of code should look something like (sorry about the pythonic implementation I am not familiar with Java)
```python
# New check to ensure the file is correctly mapped before attempting to read
if Location(filename_hint).length() == 0:
    # Ensure that mapped file length is not zero, which would indicate a problem with the mapping
    mapped_file = Location.getMappedFile(filename_hint)
    if mapped_file is not None and mapped_file.length() == 0:
        print("Warning: Both Location.getMappedFile and Location(filename_hint) report length 0. This may indicate a problem with the file mapping.") # Or simply raise an exception
```

Here is the python code that I am using to load the image into memory and read it using bioformats. I am using `scyjava`. I do not think that this is an issue with `scyjava` which is why I am reporting the problem here. 
```python
filename = 'image.czi'
with open(filename, 'rb') as f:
    data = f.read()

filename_hint = os.path.basename(filename)  # Bio-Formats uses the filename hint to determine the file format

Location = scyjava.jimport("loci.common.Location")
ByteArrayHandle = scyjava.jimport("loci.common.ByteArrayHandle")

java_bytes = jpype.JArray(jpype.JByte)(data)  # data is bytes/bytearray
handle = ByteArrayHandle()
handle.write(java_bytes)
handle.seek(0)  # Reset position to the beginning

Location.mapFile(filename_hint, handle)

ImageReader = scyjava.jimport("loci.formats.ImageReader")
MetadataTools = scyjava.jimport("loci.formats.MetadataTools")

reader = ImageReader()
reader.setId(filename_hint) # This raises the error

series_count = reader.getSeriesCount()
for serie in range(series_count):
    reader.setSeries(serie)
    size_x = reader.getSizeX()
    size_y = reader.getSizeY()
    size_z = reader.getSizeZ()
    size_c = reader.getSizeC()
    size_t = reader.getSizeT()
    print(f"Series {serie}: SizeX={size_x}, SizeY={size_y}, SizeZ={size_z}, SizeC={size_c}, SizeT={size_t}")
print(f"Number of series in the image: {series_count}")

Location.mapFile(filename_hint, None)
```

I am using these versions:
scyjava: 1.12.2
jpype: 1.6.0
bioformats: 8.4.0

sbesson:
Fixes #4413 

If a file has been mapped into memory via the `Location.mapFile()` API, skip the location checks entirely instead of printing an ERROR level warning

### DIFF
--- a/components/formats-api/src/loci/formats/ImageReader.java
+++ b/components/formats-api/src/loci/formats/ImageReader.java
@@ -179,7 +179,7 @@ public class ImageReader implements IFormatReader {
 
    // NB: Check that we can generate a valid handle for the ID;
    // e.g., for files, this will throw an exception if the file is missing.
-   if (!fake && !omero) {
+   if (!fake && !omero && !Location.getIdMap().containsKey(id)) {
      Location.checkValidId(id);
      if (new Location(id).length() == 0) {
        LOGGER.error("File has length 0 and may be corrupt");


melissalinkert (commented):
I think I'm OK with this change, but in the case of non-fake files with non-zero length (as in #4413), I wouldn't have expected `new Location(id).length()` to return 0.

Looking at the `Location` constructor:

https://github.com/ome/ome-common-java/blob/master/src/main/java/loci/common/Location.java#L234

and:

https://github.com/ome/ome-common-java/blob/master/src/main/java/loci/common/Location.java#L370

that's accounting for String -> String mapping, but not String -> IRandomAccess. That's almost certainly incorrect, and probably worth fixing as well.

sbesson:
Proposing to close this fix in favour of https://github.com/ome/ome-common-java/pull/113 which allows the `Location.length()` API to work on in-memory files.

PR #4415, opened 2026-02-24
Remove explicit aircompressor dependency from formats-gpl

melissalinkert:
This should be inherited from ome-codecs. No impact expected, but watch for any test failures, especially in the Imaris HDF LZ4 sample data.


sbesson (approved):
https://bf-testing-results.s3.amazonaws.com/2026/2026-02-25/build-status.log all green. Merging and we'll review the impact of the dependency version bump from https://github.com/ome/ome-codecs/pull/46

Issue #3645, opened 2020-12-09
NetCDF: thread-safety issue


sbesson:
While converting a fileset composed of multiple `.ims` files combined with a `.pattern` files using [bioformats2raw](https://github.com/glencoesoftware/bioformats2raw), I found the ImarisHDFReader was hanging on the initialization of one of the IMS files. The IMS file differed from execution to execution but the hanging was reproducible. Setting the number of concurrent workers to 1 is sufficient to work through the entire conversion pipeline. 

Separately, the non-regression tests were executed against the representative files of the three file format readers using the NetCDF service, both in single-threaded and multi-threaded mode:

- ImarisHDFReader [multi-threaded](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/86970/console) vs [single-threaded](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/86971/console)
- MincReader [multi-threaded](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/86972/console) vs [single-threaded](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/86973/console)
- VeecoReader [multi-threaded](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/86974/console) vs [single-threaded](https://latest-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-test-folder/86975/console)

In all three cases, the tests completed successfully when run single-threaded. When run multi-threaded, the tests fail silently and the suite is not completed but no error is raised.

In all cases, this highlight an issue with the thread-safety of the readers using `NetCDF` as the service to access the HDF data. 

See also https://github.com/glencoesoftware/bioformats2raw/pull/76.

sbesson:
Came back to this in the context of the testing of https://github.com/ome/bioformats/pull/4409. @melissalinkert established the issue is not as severe as originally described. The `FormatReader` tests are successfully executed against all Imaris samples and reported in the corresponding ThreadPool log files 

```
ubuntu@ip-172-31-27-233:~/bioformats/components/test-suite$ grep PASSED target/bio-formats-test-pool-1-thread-*-2026-02-13_02-13-2*.log | wc
   3392   22016  433939
```

Updated the issue description accordingly.

PR #4421, opened 2026-03-23
Imaris HDF: allow paths to be in a subdirectory

melissalinkert:
See https://forum.image.sc/t/ims-files-cannot-open-in-bio-formats-again/118877/10.

Relevant data is https://zenodo.org/records/18852087 and https://zenodo.org/uploads/19090243.

Without this change, `showinf -debug` on either file should show:

```
loci.formats.FormatException: Unknown pixel type: null
        at loci.formats.in.ImarisHDFReader.initFile(ImarisHDFReader.java:339)
```

`h5ls` on either file will show something like:

```
DataSet                  Soft Link {/Workflows/InitialImages/DataSet}
DataSetEvents            Soft Link {/Workflows/InitialImages/DataSetEvents}
DataSetInfo              Soft Link {/Workflows/InitialImages/DataSetInfo}
DataSetTimes             Soft Link {/Workflows/InitialImages/DataSetTimes}
Thumbnail                Soft Link {/Workflows/InitialImages/Thumbnail}
Workflows                Group
```

netcdf-java does not appear to automatically resolve soft links, so to get the correct path prefix (`Workflows/InitialImages`) this change looks for the `ImarisVersion` attribute and captures the parent path.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
@@ -289,8 +289,7 @@ public class ImarisHDFReader extends SubResolutionFormatReader {
     if (seriesCount > 1) {
       for (int i=1; i<seriesCount; i++) {
         CoreMetadata ms = new CoreMetadata();
-        String groupPath =
-          getPath("DataSet/ResolutionLevel_" + i + "/TimePoint_0/Channel_0");
+        String groupPath = getPlanePath(i, 0, 0);
         ms.sizeX =
           Integer.parseInt(netcdf.getAttributeValue(groupPath + "/ImageSizeX"));
         ms.sizeY =
@@ -319,13 +318,13 @@ public class ImarisHDFReader extends SubResolutionFormatReader {
     // read block sizes for caching
     blockSizeZPerResolution = new int[seriesCount];
     for (int res = 0; res < seriesCount; res++) {
-      String datasetPath = getPath("DataSet/ResolutionLevel_" + res + "/TimePoint_0/Channel_0/Data");
+      String datasetPath = getPlaneDataPath(res, 0, 0);
       Hashtable<String, Object> table = netcdf.getVariableAttributes(datasetPath);
       String chunkSizesString = (String) table.get("_ChunkSizes");
       String[] sizes = chunkSizesString.split(" ");
       blockSizeZPerResolution[res] = Integer.parseInt(sizes[0]);
     }
-    
+
     // determine pixel type - this isn't stored in the metadata, so we need
     // to check the pixels themselves
 
@@ -513,7 +512,7 @@ public class ImarisHDFReader extends SubResolutionFormatReader {
         try {
           String path;
           for (int ch = 0; ch < getSizeC(); ch++) {
-            path = getPath("/DataSet/ResolutionLevel_" + resolutionIndex + "/TimePoint_" + zct[2] + "/Channel_" + ch + "/Data");
+            path = getPlaneDataPath(resolutionIndex, zct[2], ch);
             buffer[ch] = netcdf.getArray(path, idcs, dims);
           }
         }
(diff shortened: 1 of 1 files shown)


melissalinkert:
No build failures in https://bf-testing-results.s3.amazonaws.com/2026/2026-03-24/build-status.log, but #3645 still applies.

sbesson (commented):
In the absence of native support for symlink resolution in NetCDF-Java which is unfortunate, detecting the path prefix as proposed here might be our best workaround.

Testing using using the sample IMS file from https://zenodo.org/records/18852087, a few high-level questions/comments :

- this proposal is using the `ImarisVersion` attribute. Interesting, this attribute is undefined in https://imaris.oxinst.com/support/imaris-file-format and the `FormatVersion` attribute expected as per this specification is missing from the sample. Would it make sense to use `ImarisDataSet` instead to have some consistency with the official specification document?
- in the provided IMS samples, there are two `ImarisVersion` (or `ImarisDataset`) attributes: one at the top-level and one under the hierarchy. Is it correct that the current logic works parses the groups and attributes recursively starting with the root group so the last occurence of the attribute will be the one in the leaf group of interest?
- extending the above, I assume in the hypothetical case of multiple group hierarchies containing an `ImarisDataset`, the last parsed group would take precedence independently of the top-level symlink.

melissalinkert:
> * this proposal is using the `ImarisVersion` attribute. Interesting, this attribute is undefined in https://imaris.oxinst.com/support/imaris-file-format and the `FormatVersion` attribute expected as per this specification is missing from the sample. Would it make sense to use `ImarisDataSet` instead to have some consistency with the official specification document?

Done in b417b7bc90. That seems to work locally, so hopefully tests continue to pass overnight.

> * in the provided IMS samples, there are two `ImarisVersion` (or `ImarisDataset`) attributes: one at the top-level and one under the hierarchy. Is it correct that the current logic works parses the groups and attributes recursively starting with the root group so the last occurence of the attribute will be the one in the leaf group of interest?

Yes, the reader is iterating over the list of attributes recursively parsed from https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/services/NetCDFServiceImpl.java#L113 and https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/services/NetCDFServiceImpl.java#L243.

> * extending the above, I assume in the hypothetical case of multiple group hierarchies containing an `ImarisDataset`, the last parsed group would take precedence independently of the top-level symlink.

Exactly, yes. I could see a case for only setting the `pathPrefix` if it is empty, which means that if a file had something like:

- `/ImarisDataSet`
- `/Workflows/InitialImages/ImarisDataSet`
- `/Workflows/Workflow0/ImarisDataSet`

then `/Workflows/InitialImages` would be chosen (not `/Workflows/Workflow0` as is the case now). We don't have data that demonstrates this case though, so that might be premature.

components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
```diff
@@ -288,7 +290,7 @@ protected void initFile(String id) throws FormatException, IOException {
       for (int i=1; i<seriesCount; i++) {
         CoreMetadata ms = new CoreMetadata();
         String groupPath =
-          "DataSet/ResolutionLevel_" + i + "/TimePoint_0/Channel_0";
+          getPath("DataSet/ResolutionLevel_" + i + "/TimePoint_0/Channel_0");
```
sbesson: As most of the group paths are under with `Dataset/ResolutionLevel_`, would it be useful to introduce some `getResolutionPath(int)` API ? Or is it creating more noise than necessary?

sbesson (commented):
Retested with https://zenodo.org/records/18852087 and I can confirm that Bio-Formats 8.5.0 fails to open this file while the latest version of this PR successfully opens an image with two resolutions and 330x330x12x2x121 dimensions. When setting the log level to DEBUG, the initialization also indicates the path prefix is updated

```
Set path prefix to 
Set path prefix to Workflows/InitialImages
```

Looking at the image metadata, the IMS file was derived from a Slidebook file and processed using Imaris 11

```
OriginalFormat: Intelligent Imaging Innovations: SlideBook
OriginalFormatFileIOVersion: ImarisFileIO x64 11.0.0
```

The workflows features introduced in https://imaris.oxinst.com/support/imaris-release-notes/11-0-0 are consistent with the HDF layout of the representative sample. These changes should allow the reader to support the scenario where a single dataset has been processed via these workflows. As discussed during the review, supporting HDF files will multiple groups under `Workflows` is outside the scope of this PR and we would need representative samples as well as upstream information on how to resolve this scenario anyways.

components/formats-gpl/src/loci/formats/in/ImarisHDFReader.java
melissalinkert: c0e8d88 is one approach.

sbesson (approved):

PR #4429, opened 2026-05-13
Check for null when calling TiffParser.checkHeader()

### LINKED ISSUE #4232: NULL being returned for a Boolean object
https://github.com/ome/bioformats/blob/678fd01d88b24208188683db77f293b0ff34e11f/components/formats-bsd/src/loci/formats/tiff/TiffParser.java#L199-L220

Referring to above, there are 3 places in method **checkHeader()** a null is returned.  This will cause a Null Pointer Exception for those callers that are returned the result to a boolean data type and not the Boolean object.

Here are 2 examples of callers expecting boolean data type and not Boolean object:

https://github.com/ome/bioformats/blob/e4ffbe89ba6cb9fd44f92fb4dfa4cc64f3918ba5/components/formats-bsd/src/loci/formats/out/PyramidOMETiffWriter.java#L129

https://github.com/ome/bioformats/blob/0a895ae7e7895e9193c02139ca51189fe5b9c106/components/formats-gpl/src/loci/formats/in/LeicaReader.java#L775

Here’s an example of code checking for NULL and taking appropriate action:

https://github.com/ome/bioformats/blob/678fd01d88b24208188683db77f293b0ff34e11f/components/formats-bsd/src/loci/formats/in/MinimalTiffReader.java#L448-L452

If it’s thought that this is an issue, my Eclipse seems to indicate there are about 10 other callers with this NPE situation.

melissalinkert:
Fixes #4232.

Most calls to `checkHeader` already check for a `null` return value, or follow a call to `isValidHeader` which checks for `null` itself. In the few remaining cases here, the changes are mostly throwing a more informative exception, or logging an error when that isn't reasonable. `checkHeader` returning `null` implies that the TIFF file is invalid, so throwing an exception of some kind is usually appropriate.

This should not be a breaking change, and is not expected to have any impact on tests.

### DIFF
--- a/components/formats-bsd/src/loci/formats/out/PyramidOMETiffWriter.java
+++ b/components/formats-bsd/src/loci/formats/out/PyramidOMETiffWriter.java
@@ -126,7 +126,12 @@ public class PyramidOMETiffWriter extends OMETiffWriter {
       long[] allOffsets = null;
       try (RandomAccessInputStream in = new RandomAccessInputStream(id)) {
         TiffParser parser = new TiffParser(in);
-        littleEndian = parser.checkHeader();
+        Boolean writtenLittleEndian = parser.checkHeader();
+        if (writtenLittleEndian == null) {
+          // this would be extremely unexpected
+          throw new IOException("Endian check failed on written data");
+        }
+        littleEndian = writtenLittleEndian;
         allOffsets = parser.getIFDOffsets();
       }
 
--- a/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
+++ b/components/formats-bsd/src/loci/formats/tiff/TiffParser.java
@@ -114,7 +114,10 @@ public class TiffParser implements Closeable {
     doCaching = true;
     try {
       long fp = in.getFilePointer();
-      checkHeader();
+      Boolean littleEndian = checkHeader();
+      if (littleEndian == null) {
+        LOGGER.error("Endian check failed, this may be an invalid file");
+      }
       in.seek(fp);
     }
     catch (IOException e) { }
--- a/components/formats-bsd/test/loci/formats/utests/tiff/TiffParserTest.java
+++ b/components/formats-bsd/test/loci/formats/utests/tiff/TiffParserTest.java
@@ -73,6 +73,7 @@ public class TiffParserTest {
   @Test
   public void testHeader() throws IOException {
     assertTrue(tiffParser.isValidHeader());
+    // isValidHeader above makes sure that checkHeader is non-null
     assertTrue(tiffParser.checkHeader());
     assertFalse(tiffParser.isBigTiff());
   }
@@ -178,6 +179,8 @@ public class TiffParserTest {
   public void testNonUniformRowsPerStrip() throws IOException, FormatException {
     mock = new NonUniformRowsPerStripMock();
     tiffParser = mock.getTiffParser();
+    assertTrue(tiffParser.isValidHeader());
+    // isValidHeader above makes sure that checkHeader is non-null
     assertTrue(tiffParser.checkHeader());
     tiffParser.getFirstIFD().getRowsPerStrip();
     mock.close();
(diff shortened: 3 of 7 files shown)


sbesson (commented):
The proposed changes are consistent with other readers like `FlowSightReader.java` or `MinimalTiffReader` which check the output of `TiffParser.checkHeader` and throw a `FormatException` if the return value is `null`.

To assess the functional impact on the readers, I constructed a synthetic Metamorph dataset derived from a [public sample](https://zenodo.org/records/13642395), replacing  individual TIFF files by empty files. When the first TIFF file in the list (`test_timelapse_20240816_s1_t1.TIF`) is corrupted, using both Bio-Formats 8.5.0 or this PR, the reader initialization would fail on the file with:

```
loci.formats.FormatException: Invalid TIFF file: /Users/sbesson/Downloads/Bio-Formats_testdata/test_timelapse_20240816_s1_t1.TIF
	at loci.formats.in.MinimalTiffReader.initFile(MinimalTiffReader.java:452)
	at loci.formats.in.BaseTiffReader.initFile(BaseTiffReader.java:609)
	at loci.formats.in.MetamorphReader.initFile(MetamorphReader.java:425)
```

When corrupting another file e.g. `test_timelapse_20240816_s1_t2.TIF`, the latest release of Bio-Formats 8.5.0 initializes the dataset, does not includes `test_timelapse_20240816_s1_t2.TIF` in the used files list and renders the corresponding plane as empty. With this change, the reader fails at initialization time with:

```
loci.formats.FormatException: Invalid TIFF file: /Users/sbesson/Downloads/Bio-Formats_testdata/test_timelapse_20240816_s1_t2.TIF
	at loci.formats.in.MetamorphReader.initFile(MetamorphReader.java:1175)
```

I will look into constructing similar examples for LeicaReader and ZeissLSMReader.
Based on the above, the outstanding question is whether increasing the strictness in terms of file validity is part of the contract/expectation of this PR or whether we want to retain the leniency of the current implementation.

sbesson (changes requested):
See https://github.com/ome/bioformats/pull/4429#pullrequestreview-4357092519

melissalinkert:
6c7e3d6 should fix the Metamorph case, to warn instead of throwing an exception if an invalid file is encountered. `MetamorphReader` previously did not use the return value of `TiffParser.checkHeader()` at all, so that should be consistent with previous behavior.

`ZeissLSMReader` and `LeicaReader` both already made use of the return value prior to this change, so should just be changing the exception that is thrown and not introducing new behavior. If `null` was returned, `LeicaReader` would have thrown `NullPointerException` on line 775 (where now it would throw `FormatException` on line 777). Similarly, `ZeissLSMReader` would have thrown `NullPointerException` on line 501 (since `CoreMetadata.littleEndian` is `boolean`), and will now instead throw `FormatException` on line 495.

sbesson (approved):
Retested the last commit using the modified Metamorph dataset derived from https://zenodo.org/records/13642395
- the original dataset reads with and without this PR with both the metadata and the planes being equivalent
- when replacing the first TIFF files of any series by an empty file, the dataset fails to initialize with and without this PR
- when replacing TIFF files other than the first of any series by an empty file, the dataset initializes correctly, the metadata and pixel data are consistent and the planes corresponding to corrupted TIFF render as black as expected.

The presence of invalid TIFF files is also communicated by the logging statement:

```
Checking file format [Metamorph STK]
Initializing reader
MetamorphReader initializing /Users/sbesson/Downloads/Bio-Formats_testdata3/test_timelapse_20240816_s1_t1.TIF
Reading IFDs
Populating metadata
...
Endian check failed, this may be an invalid file
Found invalid TIFF file: /Users/sbesson/Downloads/Bio-Formats_testdata3/test_tim
elapse_20240816_s1_t2.TIF
Endian check failed, this may be an invalid file
Found invalid TIFF file: /Users/sbesson/Downloads/Bio-Formats_testdata3/test_tim
elapse_20240816_s1_t4.TIF
Endian check failed, this may be an invalid file
Found invalid TIFF file: /Users/sbesson/Downloads/Bio-Formats_testdata3/test_tim
elapse_20240816_s2_t3.TIF
Parsing schema path
Validating OME-XML
No validation errors found.
Initialization took 0.323s
```

PyramidOMETiffWriter, LeicaReader and ZeissLSMReader all had existing logic relying on the return value of `checkHeader`. Reviewing the code, I agree a `null` value would have resulted on downstream errors albeit possibly opaque NullPointerException. The new logic now handles `null` value consistently with an informative error message.

The nightly builds have been passing with this PR included - see https://bf-testing-results.s3.amazonaws.com/2026/2026-07-26/build-status.log.
Merging this PR based on above. The only addition suggestion from the testing would be to create a synthetic sample as described above i.e. derived from a complete Metamorph dataset with some corrupted TIFF files, add it to the curated QA repository and configure it so that we have some coverage of the reader behavior for this use case.

PR #4430, opened 2026-05-13
Remove unnecessary dead store to local variable

### LINKED ISSUE #4235: Dead store to local variable
https://github.com/ome/bioformats/blob/682bc7e318b88f159d3c6e2603fb82fb9d613268/components/formats-bsd/src/loci/formats/dicom/DicomTag.java#L353

**location** is an argument to this constructor and so it being incremented won’t be seen outside this constructor. 
 **location** getting updated doesn't seem to be used anywhere else in the constructor.

Was a member variable wanted??

Is **location** really not needed after the constructor?

melissalinkert:
Fixes #4235.

I am pretty sure this is not breaking, but we'll want to keep an eye out for test failures in tonight's build. It's probably also worth double-checking that converting one or two datasets to DICOM works as before; I tried `bfconvert -noflat CMU-1.svs CMU-1.dcm -compression JPEG` without issue, but an extra check can't hurt.

A more-breaking change would be to deprecate/remove the `location` variable from the constructors, but I'm not 100% certain that's what we want. Leaving the possibility of using `location` at a later date may still be useful.

### DIFF
--- a/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
+++ b/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
@@ -350,7 +350,6 @@ public class DicomTag implements Comparable<DicomTag> {
         if (in.getFilePointer() + skipCount <= in.length()) {
           in.skipBytes(skipCount);
         }
-        location += elementLength;
         value = "";
       }
     }


sbesson (approved):
As noted in the original issue, the `location` variable is not used anywhere so removing the update should be harmless.
Tested functionally by converting CMU-1.svs to DICOM both with JPEG compression and the `-precompressed` mode and comparing the output to the original.

PR #4437, opened 2026-06-08
Allow slices to be deleted from a virtual stack

### LINKED ISSUE #4004: Plugin: Exception when deleting slice with Virtual Stack
This issue was reported on forum thread https://forum.image.sc/t/ffmpeg-plugin-does-not-support-virtual-stack-slice-deletion/81112. The exception occurs after importing a file using the Bio-Formats plugin as a virtual stack and then running the `Image > Stacks > Delete Slice` command. The initial report was with an AVI file. Testing other file types, and although the exception is not thrown, the delete slice functionality does not appear to work. The issue only seems to be with Virtual Stacks, the functionality works as expected with a standard import.

melissalinkert:
Fixes #4004.

To test in ImageJ, open any file with multiple planes using `Plugins > Bio-Formats > Bio-Formats Importer` with `Use virtual stack` checked. Move the slider to any plane, and select `Image > Stacks > Delete Slice`. The stack size should decrease appropriately, and moving the slider through all planes should work as expected without error. Deleting additional slices from the same stack should also work.

As an initial test, it might be worth using one of the datasets in https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/bioformats-artificial/, as those have plane indexes visible on each image, so it should be easy to see if the correct planes are being read after deletion.

### DIFF
--- a/components/bio-formats-plugins/src/loci/plugins/util/BFVirtualStack.java
+++ b/components/bio-formats-plugins/src/loci/plugins/util/BFVirtualStack.java
@@ -274,9 +274,31 @@ public class BFVirtualStack extends VirtualStack {
   @Override
   public int getSize() {
     if (reader.getCurrentFile() == null) return 0;
+    if (planeIndexes != null) {
+      return planeIndexes.length;
+    }
     reader.setSeries(series);
     if (merge) return new ChannelMerger(reader).getImageCount();
     return planeIndexes == null ? reader.getImageCount() : planeIndexes.length;
+  }
+
+  @Override
+  public void deleteSlice(int n) {
+    if (planeIndexes == null) {
+      planeIndexes = new int[getSize() - 1];
+      for (int i=0; i<planeIndexes.length; i++) {
+        planeIndexes[i] = i;
+        if (i >= n - 1) {
+          planeIndexes[i]++;
+        }
+      }
+    }
+    else {
+      int[] newPlaneIndexes = new int[planeIndexes.length - 1];
+      System.arraycopy(planeIndexes, 0, newPlaneIndexes, 0, n - 1);
+      System.arraycopy(planeIndexes, n, newPlaneIndexes, n - 1, planeIndexes.length - n);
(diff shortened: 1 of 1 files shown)


sbesson (changes requested):
Using the suggested sample, I can reproduce the original error with the latest Fiji. After deleting a slice from a 5D image opening using the virtual stack, I get the following error:

Image

With this PR included, the slice is deleted as expected, the dimension alongside Z is reduced by one and the plane assignment is correct

One remaining issue is the slice title where only the first part (prior to the parenthesis) has been updated with the new size and indices creating some confusion:

Image

melissalinkert:
If you select `Plugins > Bio-Formats > Bio-Formats Plugins Configuration`, what is the `Slice Label Pattern` on the `General` tab?

sbesson:
`%c%z%t- %n`

Screenshot 2026-08-12 at 16 48 32

melissalinkert:
:+1: I had `%n` set, so did not see the issue immediately. With `%c%z%t- %n` I can reproduce the same issue when deleting a slice from a normal (not virtual) stack as well. Everything inside the parentheses is the slice label, and these labels are not recalculated when a slice is deleted.

I definitely see the potential for confusion, but I can also see a few arguments for keeping the current behavior:

- behavior is consistent with virtual and non-virtual stacks, with no reported issue that I could find for non-virtual stacks
- slice label pattern is configurable, so if too confusing it can just be set to `%n`
- it may be useful to know what the original ZCT index was for a particular plane

Happy to discuss further though.

sbesson (approved):
Thanks for the clarification. I think you're correct and we would need to know more about the perceived usage/expectations of this slice label to make a decision on the behavior. Arguably the usage of `%c%z%t` is redundant and the default pattern could also be `%n`.

In all cases, this is addressing the core issue i.e. the inability to delete slices for a multi-dimensional image opened as a virtual stack. Merging this and new issues can be opened to discuss the label behavior if needed.

PR #4438, opened 2026-06-08
Remove unnecessary negative checks

### LINKED ISSUE #4393: A check for negativity may be incorrect
Regarding:

https://github.com/ome/bioformats/blob/3679b522c0bf6356b91daad370616d2f2ffd7d17/components/formats-bsd/src/loci/formats/dicom/DicomTag.java#L497-L502

I don’t think the check for negative is ever checked for correctly.

Variables n1 and n2 are ints but populated by shorts.  Thus, if the shorts are negative they will be sign extended to negative for ints n1 and n2.
Via the code “&= 0xffff;”, the most significant 16 bits of the ints n1 and n2 are set to 0, thus losing the negativity of the number.
Thus, the checks for negativity (n1< 0 and n2<0) are never true.

Should n1 and n2 be changed to short and the two “&=” statement be removed?

        short n1 = DataTools.bytesToShort(b, 2, 2, in.isLittleEndian());
        short n2 = DataTools.bytesToShort(b, 2, 2, !in.isLittleEndian());
        if (n1 < 0 || n1 + in.getFilePointer() > in.length()) return n2;
        if (n2 < 0 || n2 + in.getFilePointer() > in.length()) return n1;

melissalinkert:
Fixes #4393.

The fix proposed in #4393 results in several files failing to initialize. This change preserves the existing working behavior, but removes the negativity checks on `n1` and `n2` that are expected to always fail.

This passed tests on all of the DICOM files I have locally, but waiting for a successful nightly build before assigning review.

### DIFF
--- a/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
+++ b/components/formats-bsd/src/loci/formats/dicom/DicomTag.java
@@ -497,8 +497,8 @@ public class DicomTag implements Comparable<DicomTag> {
         int n2 = DataTools.bytesToShort(b, 2, 2, !in.isLittleEndian());
         n1 &= 0xffff;
         n2 &= 0xffff;
-        if (n1 < 0 || n1 + in.getFilePointer() > in.length()) return n2;
-        if (n2 < 0 || n2 + in.getFilePointer() > in.length()) return n1;
+        if (n1 + in.getFilePointer() > in.length()) return n2;
+        if (n2 + in.getFilePointer() > in.length()) return n1;
         return n1;
       case RESERVED:
         int len = DataTools.bytesToInt(b, 0, 4, in.isLittleEndian());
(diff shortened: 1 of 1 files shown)


melissalinkert:
https://bf-testing-results.s3.amazonaws.com/2026/2026-06-09/build-status.log

sbesson (commented):
As noted in the upstream issue, the checks for negative values are unnecessary as per the usage of `&= 0xffff`. The nightly builds have been passing across all DICOM datasets in the curated QA repository suggesting there is no edge case where this would cause a regression.

I was slightly confused by the usage of `DataTools.bytesToShort` with the returned value being immediately being casted to `int`. Out of curiosity, why are we not using `DataTools.bytesToInt` in lines 496/497?

melissalinkert:
> I was slightly confused by the usage of `DataTools.bytesToShort` with the returned value being immediately being casted to `int`. Out of curiosity, why are we not using `DataTools.bytesToInt` in lines 496/497?

Fair point - updated in 57a8abe164. That passes tests on the subset of files I have locally.

sbesson (approved):
The nightly DICOM tests are still passing with this PR included - https://bf-testing-results.s3.amazonaws.com/2026/2026-06-12/dicom.log

PR #4439, opened 2026-06-11
Leica SCN: update list of scanner models without color correction

### LINKED ISSUE #4420: the return of the Leica SCN reader pink image bug, for SCN files generated by the Leica FL 120 instrument
Leica has recently release a new slide scanner hardware that outputs SCN format files, the Aperio FL 120. We just shifted over from the Aperio VERSA instrument that also outputs SCN format files.

We use Bio-Formats to make use of SCN images in ImageJ. We're working with Bio-Formats version 8.1.1.

We now see the same pink images as were earlier reported in #3510 when we attempt to open SCN files created by the Aperio FL 120 in ImageJ using Bio-Formats. The underlying reasons are likely the same, in that this is connected to the YCbCr photometric interpretation.

Looking at the PR for that earlier issue, https://github.com/ome/bioformats/pull/3562/changes it looks like the YCbCr correction is being applied or not applied in LeicaSCNReader.java according to a string related to the instrument designation. I believe that a designation for the new Leica FL 120 instrument now needs to be added to the list.

melissalinkert:
Fixes #4420.

Without this change, `showinf -noflat` on https://zenodo.org/records/19009239 should show pink images. With this change, the images should look normal.

This doesn't seem to impact existing Leica SCN test data when checking locally, but waiting for a passing nightly build before assigning for review.

I debated going ahead and adding a reader option to override the list of models as well, but I'm not sure how useful that would end up being in general. Thoughts welcome.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LeicaSCNReader.java
@@ -74,7 +74,9 @@ public class LeicaSCNReader extends BaseTiffReader {
   private static final String SCHEMA_2010_10 =
     "http://www.leica-microsystems.com/scn/2010/10/01";
 
-  private static final String[] MODELS_WITHOUT_CORRECTION = {"versa", "leica scn400"};
+  private static final String[] MODELS_WITHOUT_CORRECTION = {
+    "versa", "leica scn400", "aperio leica biosystems fl"
+  };
 
   // -- Fields --
   LeicaSCNHandler handler;


sbesson (approved):
The proposed change is in the spirit of https://github.com/ome/bioformats/pull/3562 where we are maintaining a list of models which must include a YCbCr correction.

Using the provided sample SCN file uploaded as part of the issue, with Bio-Formats 8.5.0, any resolution (e.g. third resolution of the second series below) renders as pink

Image

With this PR included, the image renders as expected:

Image

melissalinkert:
Anything else you would like to see here @sbesson, or is this OK to merge?

sbesson:
I cannot remember why I didn't merge this back in June :)

PR #4447, opened 2026-06-29
Fix LIF row padding with mixed memory blocks

cgohlke:
This PR improves plane reading robustness especially for LIF files where non-image memory blocks (for example, timestamps) are interleaved with image data:

* Prefer explicit per-row stride from BytesInc metadata when available, instead of inferring row padding from adjacent memory block offsets.
* Keep inferred-gap logic only as a legacy fallback when row stride metadata is unavailable.
* Restrict fallback padding inference to rows that are not naturally 4-byte aligned (based on row bytes, not pixel width).
* Clamp inferred fallback padding to non-negative values.

See https://forum.image.sc/t/problems-reading-some-widefield-leica-lif-files-with-bioformats-in-fiji/121584 for the original bug report and a test file.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/LIFReader.java
+++ b/components/formats-gpl/src/loci/formats/in/LIFReader.java
@@ -9,15 +9,15 @@
  * %%
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as
- * published by the Free Software Foundation, either version 2 of the 
+ * published by the Free Software Foundation, either version 2 of the
  * License, or (at your option) any later version.
- * 
+ *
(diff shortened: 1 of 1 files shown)


cgohlke:
Note that I have changed the fallback condition from `getSizeX() % 4` to `(getSizeX() * bpp) % 4`. I didn't quite understand the original intent, neither am I sure if the fallback is still necessary...

will-moore:
Looks like this is the same issue https://forum.image.sc/t/some-images-within-a-lif-file-become-garbled/76807

EDIT: apologies, I missed that issue was linked to https://github.com/ome/bioformats/issues/3836 which has been fixed now

PR #4451, opened 2026-07-03
Make `DicomTile` objects more accessible

melissalinkert:
Adds a bit of API to `DicomReader` to make it easier to get `DicomTile` objects corresponding to a particular resolution and/or plane. This also improves indexing of tiles, which should hopefully make image reading faster overall.

Local tests pass, but will exclude this temporarily if tests fail with it included. Since this is expected to improve performance, will need to compare total runtime of passing tests with and without this change, for several days each.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/DicomReader.java
+++ b/components/formats-bsd/src/loci/formats/in/DicomReader.java
@@ -172,7 +172,6 @@ public class DicomReader extends SubResolutionFormatReader {
   public byte[] openCompressedBytes(int no, int x, int y) throws FormatException, IOException {
     FormatTools.assertId(currentId, true, 1);
 
-    // TODO: this will result in a lot of redundant lookups, and should be optimized
     int tileWidth = getOptimalTileWidth();
     int tileHeight = getOptimalTileHeight();
     Region boundingBox = new Region(x * tileWidth, y * tileHeight, tileWidth, tileHeight);
@@ -860,7 +859,7 @@ public class DicomReader extends SubResolutionFormatReader {
 
         for (int p=0; p<imagesPerFile; p++) {
           DicomTile tile = new DicomTile();
-          tile.fileIndex = tilePositions.get(0).size();
+          tile.fileIndex = getTiles(0).size();
           if (m.sizeZ == 1 && fileZOffset != null) {
             tile.zOffset = fileZOffset;
           }
@@ -1069,6 +1068,13 @@ public class DicomReader extends SubResolutionFormatReader {
     }
 
     watch.start();
+
+    int currentCoreIndex = getCoreIndex();
+    for (Integer key : tilePositions.keySet()) {
+      setCoreIndex(key);
+      calculateTilePlaneIndexes();
+    }
+    setCoreIndex(currentCoreIndex);
 
     // The metadata store we're working with.
     MetadataStore store = makeFilterMetadata();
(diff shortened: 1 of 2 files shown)


components/formats-bsd/src/loci/formats/in/DicomReader.java
```diff
@@ -1741,6 +1740,23 @@ private List<DicomTile> getTileList(int no, Region boundingBox, boolean firstTil
     return tileList;
   }
 
+  private void calculateTilePlaneIndexes() {
+    List<Double> zs = zOffsets.get(getCoreIndex());
```
sbesson: Should this be `getZOffsets(getCoreIndex())` to use the new API?

sbesson (commented):
This PR introduces a couple of public APIs allowing to access the internal tiles associated with each series and precomputes the index of plane associated to every single tile.

Looking at the timing of the integration tests

```
for i in $(seq 1 30); do curl -s -o log https://bf-testing-results.s3.amazonaws.com/2026/2026-06-$(printf %02d i)/dicom.log && grep "Total time:" log; done 
for i in $(seq 1 30); do curl -s -o log https://bf-testing-results.s3.amazonaws.com/2026/2026-07-$(printf %02d i)/dicom.log && grep "Total time:" log; done 
```

the average completion time  is 260 min +/- 14 min between 2026-06-01 and 2026-07-03 (before this PR was introduced) vs 249 min +/- 18 min between 2026-07-04 and 2026-07-30.

Is it correct that the major performance improvements will be seen when calling `openBytes` or `openCompressedBytes` as the tile lookup should benefit from the preliminary computation? In that case, is it pertinent to also measure the timing of a conversion from DICOM to OME-TIFF for some sample WSI DICOM?

components/formats-bsd/src/loci/formats/in/DicomReader.java
melissalinkert: See c3997d639f

melissalinkert:
> Is it correct that the major performance improvements will be seen when calling `openBytes` or `openCompressedBytes` as the tile lookup should benefit from the preliminary computation? In that case, is it pertinent to also measure the timing of a conversion from DICOM to OME-TIFF for some sample WSI DICOM?

Yes, I'd expect `openBytes`/`openCompressedBytes` to be a bit faster overall, so converting is probably the easiest way to iterate over all tiles to test that assumption.

sbesson (approved):
I ran `bfconvert -noflat` and `bfconvert -noflat -precompressed` with and without this PR using https://downloads.openmicroscopy.org/images/DICOM/wsi/2023-04-28/PARNED-0BNNF4_B2_Q30/ as the input using 3 repeats.

Average conversion times are the following:

|                                  | Bio-Formats 8.5.0 | PR 4451     |
|----------------------------------|-------------------|-------------|
| bfconvert -noflat                | 559s +/- 12s      | 548s +/- 9s |
| bfconvert -noflat -precompressed | 143s +/- 3s       | 139s +/- 2s |

In summary, this PR is improving access to the internal DICOM tiles via a new public API and refactors a few places in the reader. The nightly repository tests confirm the reading is unaffected and the conversion tests confirm no regression and a possible slight performance improvement. Were you expecting a more substantial writing change and would you like me to perform additional conversion testing on another sample?

Otherwise, no objections to merging from my perspective

melissalinkert:
> Were you expecting a more substantial writing change and would you like me to perform additional conversion testing on another sample?

I didn't have a specific performance change in mind, other than "not worse, maybe slightly better". I think the test you have done is sufficient to confirm that.

PR #4453, opened 2026-07-15
Add reader for PreciPoint .vmic whole-slide images

kasper-cancilico:
## Summary

Adds `VmicReader` for the PreciPoint `.vmic` whole-slide image format — the
long-standing format request tracked in openslide/openslide#168.

`.vmic` is an outer ZIP whose single `Image.vmici` entry is itself a ZIP containing a
Deep Zoom (DZI) pyramid of JPEG tiles (`dzc_output.xml` +
`dzc_output_files/<level>/<col>_<row>.jpg`) plus optional PreciPoint scan metadata
(`VMCF/config.osc`). The reader exposes it as a tiled, sub-resolution pyramid with
physical pixel size and objective magnification.

## Implementation

- Streams the inner `Image.vmici` container to a temporary file and reads it via a zip
  `FileSystem`, keeping heap usage low regardless of container size (these files are
  routinely > 1 GB).
- Parses `dzc_output.xml` and `config.osc` via `XMLTools`; `config.osc` (physical size /
  magnification) is optional.
- Decodes JPEG tiles with `ImageIO` and assembles requested regions across tile
  boundaries.
- Registered in `readers.txt` immediately after `ZipReader`.

## Testing

`VmicReaderTest` (TestNG, registered in the formats-gpl suite) runs entirely against a
**synthetically generated** nested-zip Deep Zoom fixture — no proprietary sample data.
It covers core metadata, pyramid resolutions, tile reads spanning tile boundaries,
optional-metadata handling, temp-file cleanup on close, and rejection of a ZIP lacking
the inner container. Also verified end-to-end by opening real PreciPoint `.vmic` slides
in QuPath through the Bio-Formats reader at modest heap.

## Attribution

The reader originated in @kwiechen's fork
(https://github.com/kwiechen/bioformats-vmic) and is credited as co-author. Opening
here to bring it upstream — happy to adjust attribution/approach per maintainer
preference, and to loop in the original author.

## Follow-ups

- Draft: opened for review before finalizing.
- A supported-formats documentation page will follow separately (the docs live outside
  this repository).
- Sample data: can provide a small synthetic `.vmic` if useful; production samples are
  patient-derived and cannot be shared.

### DIFF
--- a/components/formats-api/src/loci/formats/readers.txt
+++ b/components/formats-api/src/loci/formats/readers.txt
@@ -8,6 +8,7 @@ loci.formats.in.URLReader[type=external] # urlreader
 
 # readers for compressed/archive files
 loci.formats.in.ZipReader             # zip
+loci.formats.in.VmicReader            # vmic
 
 # javax.imageio readers
 loci.formats.in.APNGReader            # png [javax.imageio]
--- a/components/formats-gpl/src/loci/formats/in/VmicReader.java
+++ b/components/formats-gpl/src/loci/formats/in/VmicReader.java
@@ -1,0 +1,564 @@
+/*
+ * #%L
+ * OME Bio-Formats package for reading and converting biological file formats.
+ * %%
+ * Copyright (C) 2005 - 2017 Open Microscopy Environment:
+ *   - Board of Regents of the University of Wisconsin-Madison
+ *   - Glencoe Software, Inc.
+ *   - University of Dundee
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


melissalinkert:
@kasper-cancilico, this appears to be the same reader as #4397?

kasper-cancilico:
Apologies for the duplicate; it's the same reader. I built on @kwiechen's fork and opened this before I noticed #4397 already existed.

This version adds some fixes on top of the original reader:

- streams the inner `Image.vmici` to a temp file instead of reading the whole container into memory (the current version OOMs on the >1 GB files we work with)
- parses the XML through `XMLTools`
- throws checked `FormatException`s instead of runtime exceptions
- builds tile paths with `/` so they resolve on Windows
- treats `config.osc` as optional
- adds a unit test that runs against a synthetically generated `.vmic`, so there's no proprietary sample data

I don't want to step on @kwiechen's PR, though. @kwiechen, would you rather I send these as a PR against your branch so they land in #4397? If you'd prefer I carry this one instead, that's fine too. I'll leave it to you and the maintainers to decide which one to keep.

melissalinkert:
Closing this, given that https://github.com/ome/bioformats/pull/4397#issuecomment-4989386203 proposes to add changes here onto #4397.

PR #4454, opened 2026-07-16
Add API and implementations to provide axis orientation

melissalinkert:
See https://github.com/glencoesoftware/bioformats2raw/issues/329 and https://ngff.openmicroscopy.org/rfc/4.

As outlined in https://github.com/glencoesoftware/bioformats2raw/issues/329, this implements a new API that allows readers to provide axis orientation information. Right now, per RFC-4, only anatomical orientation is implemented; the intent is for this API to be flexible enough to support other orientation types in the future, if needed.

Tested on at least one dataset in each format, but unfortunately none are public, and none are true reference datasets where the correct values in RFC-4 terminology are known. Opening this is a non-draft PR to make sure no existing tests break, but once we have reference data (cc @thewtex) this will need more careful testing and possibly some updates as a result.


thewtex:
We have some testing data here:

  https://huggingface.co/fideus-labs/ome-zarr-rfc4-data
  
still a WIP.

PR #4455, opened 2026-07-22
JPEG: reshape internal tiles if calculated tile width exceeds max short value

melissalinkert:
Backported from a private PR.

No test data at the moment, but opening to see the impact on existing tests.

### DIFF
--- a/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
+++ b/components/formats-bsd/src/loci/formats/services/JPEGTurboServiceImpl.java
@@ -214,9 +214,14 @@ public class JPEGTurboServiceImpl implements JPEGTurboService {
       }
     }
 
+    LOGGER.info("restartInterval = {}", restartInterval);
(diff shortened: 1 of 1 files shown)


melissalinkert:
Artificial test dataset now in `curated/hamamatsu/samples/large-restart-interval/`. Test code that generated the .ndpi file from https://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/openslide/CMU-2/ is included.

I have so far been unsuccessful in creating a file that 100% replicates the original problem (full resolution all black, but still viewable in NDP.view2). However, using the created test file should still demonstrate this fix as `showinf -noflat -crop 0,0,1024,1024` without this change will show clearly broken image data, and with this change should match the original `CMU-2.ndpi`.

Using a8c93ce we have established that no existing test file had a restart interval large enough to require this fix; in most cases, the restart interval is 256. The problem occurs when the restart interval multiplied by 8 (MCU width) exceeds the maximum value of a `short`.

a0e3226067 updates the log level of statements introduced in a8c93ce.

PR #4457, opened 2026-07-30
NDPI: Add writer support

ebenolson:
Adds support for writing NDPI files.

  - Writes pyramid levels, macro images, and tissue maps.
  - Supports restart-indexed JPEG data, AuthCode generation, associated-image metadata, and files larger than 4 GiB.
  - Sequential bounded-memory writing for large images.
  - Derives NDPI tags from OME metadata when mapping is clear, with writer options for NDPI-specific fields.
  - Includes structural, metadata, pixel, cleanup, and writer-discovery coverage.

Mainly generated by Codex. Outputs verified against NDPIReader and OpenSlide, and confirmed readable by Hamamatsu NDP.view2 and Aperio ImageScope

### DIFF
--- a/components/formats-api/src/loci/formats/writers.txt
+++ b/components/formats-api/src/loci/formats/writers.txt
@@ -5,6 +5,7 @@
 loci.formats.out.OMEXMLWriter  	# ome
 loci.formats.out.PyramidOMETiffWriter # ome.tif, ome.tiff
 loci.formats.out.OMETiffWriter  # ome.tif, ome.tiff
+loci.formats.out.NDPIWriter     # ndpi
 loci.formats.out.TiffWriter     # tif, tiff
 loci.formats.out.JPEGWriter     # jpg, jpeg
 loci.formats.out.JPEG2000Writer # jp2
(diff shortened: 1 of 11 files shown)


ebenolson:
rebased on current `develop` and fixed a crash found while running the PR tests (all complete):

```verify that the branch merges cleanly into develop
verify that the branch compiles with the clean jars tools Ant targets
verify that the branch compiles using Maven
verify that the branch does not use syntax or API specific to Java > 11
run the unit tests (ant test) and correct any failures
test at least one file in each affected format, using the showinf command
```

melissalinkert:
While we can see the potential use cases for being able to save data as NDPI, we do not feel that we can accept the responsibility of maintaining this writer in Bio-Formats.

Bio-Formats’ primary objective is to convert data from proprietary formats into open formats, thereby making data more accessible. Adding export support for a proprietary format is not in line with this objective. This is alluded to (though unfortunately not explicitly stated with respect to new writers) in blog posts and other documentation (https://www.openmicroscopy.org/2019/06/25/formats.html, https://bio-formats.readthedocs.io/en/latest/about/index.html).

We encourage you to make this writer available via a separate repository that you maintain independently.

PR #4463, opened 2026-08-01
Quicktime fixes

### LINKED ISSUE #4091: QuickTime: images may not be readable when using Memoizer
Importing the file from `curated/quicktime/qa-11074` into OMERO succeeds and the first 3 planes can be viewed, but subsequent planes fail to display with errors like this in the logs:

```
java.lang.NullPointerException: null
        at loci.formats.in.NativeQTReader.openBytes(NativeQTReader.java:231)
        at loci.formats.DelegateReader.openBytes(DelegateReader.java:227)
        at loci.formats.ImageReader.openBytes(ImageReader.java:465)
        at loci.formats.ChannelFiller.openBytes(ChannelFiller.java:167)
        at loci.formats.ChannelFiller.openBytes(ChannelFiller.java:159)
        at loci.formats.ChannelSeparator.openBytes(ChannelSeparator.java:200)
        at loci.formats.ReaderWrapper.openBytes(ReaderWrapper.java:348)
        at ome.io.bioformats.BfPixelsWrapper.getTile(BfPixelsWrapper.java:350)
        at ome.io.bioformats.BfPixelBuffer.getTile(BfPixelBuffer.java:499)
```

I could not immediately reproduce this with `showinf` or `showinf -cache`. Some simple test code does show the same exception though:

```
$ cat Test.java 
import java.util.Arrays;
import loci.formats.Memoizer;

public class Test {
    try (ImageReader r = new ImageReader()) {
      r.setId(args[0]);

      for (int i=0; i<r.getImageCount(); i++) {
        Memoizer m = new Memoizer();
        m.setId(args[0]);
        byte[] regularPlane = r.openBytes(i);
        byte[] memoPlane = m.openBytes(i);
        System.out.println("plane #" + i + " equal: " + Arrays.equals(regularPlane, memoPlane));
      }
    }
  }
}
$ java Test hekdfcpgfpmcherry2xfyvelivestarve_p0.mov
```

The source of the issue seems to be that the `prevPixels` array in `NativeQTReader` (`QTReader` in 7.0.0+) changes after every `openBytes` - as expected, this is for handling the case when the stored plane is actually the difference between the current and previous planes. However, updates to `prevPixels` do not get written back to the memo file - also more or less as expected. This means that `prevPixels` can be in the wrong state for a given plane when the `Memoizer` is reinitialized on every plane read.

This may be slightly tricky to fix; getting `prevPixels` into the correct state for a particular plane can require reading all previous planes.

As far as I can tell, this was not recently introduced, and has likely been a problem for a long time. We might want to consider adding a repository test that demonstrates this use case.

Darlokt:
Hej,
while investigating weird memoization behaviors, I came upon the old issue ome/bioformats#4091. It was on the way, and only partially related to the memoization problems.

These commits now fix QuickTime parsing for variable-sized samples and random access to QTRLE delta frames after memoization, by which it fixes ome/bioformats#4091

## Changes

- Correct QuickTime sample-size and chunk-offset table parsing.
- Handle leading atoms and final sample boundaries correctly.
- Rebuild preceding QTRLE reference frames when accessing delta frames non-sequentially.
- Add regression tests covering variable sample sizes and memoized random access.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/QTReader.java
+++ b/components/formats-bsd/src/loci/formats/in/QTReader.java
@@ -178,7 +178,7 @@ public class QTReader extends FormatReader {
     preparePreviousFrame(no, code);
 
     int offset = offsets.get(no).intValue();
-    int nextOffset = (int) pixelBytes;
+    int nextOffset = (int) (pixelOffset + pixelBytes);
 
     scale = offsets.get(0).intValue();
     offset -= scale;
@@ -472,11 +472,12 @@ public class QTReader extends FormatReader {
         if (atomType.equals("mdat")) {
           // we've found the pixel data
           pixelOffset = in.getFilePointer();
-          pixelBytes = offset + atomSize;
+          long pixelEnd = offset + atomSize;
 
-          if (pixelBytes > in.length()) {
-            pixelBytes = in.length();
+          if (pixelEnd > in.length()) {
+            pixelEnd = in.length();
           }
+          pixelBytes = pixelEnd - pixelOffset;
         }
         else if (atomType.equals("tkhd")) {
           // we've found the dimensions
@@ -537,9 +538,9 @@ public class QTReader extends FormatReader {
           if (numPlanes != getImageCount()) {
             int off = in.readInt();
             offsets.add(off);
-            for (int i=1; i<getImageCount(); i++) {
-              if (i - 1 < chunkSizes.size()) {
-                rawSize = chunkSizes.get(i - 1).intValue();
+            for (int i=0; i<getImageCount() - 1; i++) {
+              if (chunkSizes.size() > 0 && i < chunkSizes.size()) {
+                rawSize = chunkSizes.get(i).intValue();
               }
               else i = getImageCount();
               off += rawSize;
--- a/components/formats-bsd/src/loci/formats/in/QTReader.java
+++ b/components/formats-bsd/src/loci/formats/in/QTReader.java
@@ -182,8 +182,9 @@ public class QTReader extends FormatReader {
     offset -= scale;
 
     if (no < offsets.size() - 1) {
-      nextOffset = offsets.get(no + 1).intValue() - scale;
+      nextOffset = offsets.get(no + 1).intValue();
     }
+    nextOffset -= scale;
 
     if ((nextOffset - offset) < 0) {
       int temp = offset;
@@ -448,10 +449,10 @@ public class QTReader extends FormatReader {
         if (atomType.equals("mdat")) {
           // we've found the pixel data
           pixelOffset = in.getFilePointer();
-          pixelBytes = atomSize;
+          pixelBytes = offset + atomSize;
 
-          if (pixelBytes > (in.length() - pixelOffset)) {
-            pixelBytes = in.length() - pixelOffset;
+          if (pixelBytes > in.length()) {
+            pixelBytes = in.length();
           }
         }
         else if (atomType.equals("tkhd")) {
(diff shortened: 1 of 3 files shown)


components/formats-bsd/src/loci/formats/in/QTReader.java
```diff
@@ -448,10 +472,10 @@ else if (atomSize > in.length()) {
         if (atomType.equals("mdat")) {
           // we've found the pixel data
           pixelOffset = in.getFilePointer();
-          pixelBytes = atomSize;
+          pixelBytes = offset + atomSize;
 
-          if (pixelBytes > (in.length() - pixelOffset)) {
-            pixelBytes = in.length() - pixelOffset;
+          if (pixelBytes > in.length()) {
+            pixelBytes = in.length();
```
melissalinkert: Is there a reason to change the value of `pixelBytes` (meant to be the number of bytes of pixel data), instead of adjusting `nextOffset` in `openBytes`? Were the `nextOffset` and `openBytes` calculations only a problem for the first plane?

components/formats-bsd/src/loci/formats/in/QTReader.java
```diff
@@ -511,12 +535,11 @@ else if (atomType.equals("stco")) {
...
           int numPlanes = in.readInt();
           if (numPlanes != getImageCount()) {
-            in.seek(in.getFilePointer() - 4);
             int off = in.readInt();
             offsets.add(off);
             for (int i=1; i<getImageCount(); i++) {
-              if ((chunkSizes.size() > 0) && (i < chunkSizes.size())) {
-                rawSize = chunkSizes.get(i).intValue();
+              if (i - 1 < chunkSizes.size()) {
+                rawSize = chunkSizes.get(i - 1).intValue();
```
melissalinkert: The intent of this would be easier to follow if the `i` index into `chunkSizes` was kept, and the loop instead changed to start at 0 and end at `getImageCount() - 1`. Please keep the check for `chunkSizes.size() > 0`.

components/formats-bsd/test/loci/formats/utests/QTReaderTest.java
```diff
@@ -0,0 +1,134 @@
...
+    byte[][] expected;
+    try (QTReader direct = new QTReader()) {
+      direct.setId(movie.getAbsolutePath());
+      expected = new byte[direct.getImageCount()][];
+      for (int plane = 0; plane < expected.length; plane++) {
+        expected[plane] = direct.openBytes(plane);
+      }
+    }
+
+    Memoizer seed = new Memoizer(new QTReader(), 0, memoDirectory);
```
melissalinkert: Why not use `try (Memoizer ...)` similar to line 103?

components/formats-bsd/test/loci/formats/utests/QTReaderTest.java
```diff
@@ -0,0 +1,134 @@
...
+    seed.setId(movie.getAbsolutePath());
+    assertTrue(seed.isSavedToMemo());
+    seed.close();
+
+    for (int plane = 0; plane < expected.length; plane++) {
+      Memoizer memoized = new Memoizer(new QTReader(), 0, memoDirectory);
```
melissalinkert: Again, why not use `try (Memoizer ...)`?

melissalinkert (changes requested):

components/formats-bsd/src/loci/formats/in/QTReader.java
Darlokt: Not really, sorry. The original value included the `mdat` header, while my change made it an absolute endpoint, neither strictly matched the documented byte count. I’ve updated `pixelBytes` to hold the actual payload length and moved the final-boundary calculation into `openBytes`. This issue affects the final plane, the earlier planes use the next sample offset.

components/formats-bsd/src/loci/formats/in/QTReader.java
Darlokt: Yeah, sure!

components/formats-bsd/test/loci/formats/utests/QTReaderTest.java
Darlokt: Sorry, I overlooked that, I forgot about the closeable.

Darlokt:
This should address it all, I saw that I also forgot the closeable in ome/bioformats#4464 so i will clean it up.

PR #4464, opened 2026-08-01
Memoizer fixes

### LINKED ISSUE #3034: Memoizer should use drive letter to determine folder on Windows
Memoizer strips off the drive letter on Windows, but this creates a potential situation where the the memo file may be mistakenly used for the a similarly named file under a similar path but on a different drive.

Suppose the directory property of Memoizer is set to C:\temp
File A is located at C:\somedir\foo.nd2
File B is located at D:\somedir\foo.nd2

Memoizer will use the same memo file for A and B located at C:\temp\somedir\.foo.nd2.bfmemo
This is because the drive is simply stripped:
https://github.com/openmicroscopy/bioformats/blob/master/components/formats-bsd/src/loci/formats/Memoizer.java#L794

I suggest that the memo files should be created at
C:\temp\somedir\C\.foo.nd2.bfmemo
C:\temp\somedir\D\.foo.nd2.bfmemo

That can be accomplished by just removing ":" rather than stripping off the drive letter.

Darlokt:
Hej,

While testing Bio-Formats memoization with the Java 25 build from ome/bioformats#4450, several independent memoization problems surfaced. I did not take care of my branch and now its quite stacked, so this combines all the fixes here. The `MetadataOnly` schema issue is now being addressed upstream in ome/ome-model#233, so that workaround from ome/bioformats#4459 is intentionally excluded here.

### Keep standard data tests on direct readers

Avoid implicit memoization in standard data tests. An empty cache path behaves differently across Java versions, including Java 25.
Fixed in aaa50efe6a651ceb236a706d586e434344c0d14e.

### Invalidate memos when effective reader options change

Compare complete metadata-option state and load the current `.bfoptions` file before accepting a cached reader.
Fixed in 39a7fe100e722ea4517915f44d975d08c63716a4.

### Invalidate memos when any used file changes

Store a versioned manifest for every `getUsedFiles()` entry, including relocatable paths, existence, size, and modification time. Reject stale memos when any dependency changes.

Fixed in 96f3128157046d117b0fda3682270d5766fb5713.

### Report memo saves only after successful installation

Install temporary memo files atomically where supported, fall back when necessary, and correctly report installation failures through `saveMemo()` and `isSavedToMemo()`.

Fixed in 0a38ed2ef9a8c78a4e42c4c96a358b96b392a258.

### Keep Windows volume identity in memo cache paths

Preserve Windows drive letters and UNC share names so different volumes cannot map to the same memo.

Fixes ome/bioformats#3034

Fixed in 787e3988a91beee4c43e19733be6f3a1390c2a7f.

### Fix memo cache path expectations on Windows

Update regression tests for platform-independent drive and UNC cache-path behavior.

Fixed in d126395581ef4064e5e049ea01d3da84644f3b26.

## Tests

Add focused coverage for metadata-option changes, `.bfoptions` changes, companion-file invalidation, relocation, failed installation, and Windows drive/UNC paths.

The normal focused `MemoizerTest` execution passes. The configured `test-no-hdf` execution still reports two `Memoizer(0)` save failures, but this is not related to these changes.

### DIFF
--- a/components/formats-api/src/loci/formats/FormatTools.java
+++ b/components/formats-api/src/loci/formats/FormatTools.java
@@ -1267,9 +1267,7 @@ public final class FormatTools {
       return false;
     }
 
-    if (!a.getMetadataOptions().getMetadataLevel().equals(
-      b.getMetadataOptions().getMetadataLevel()))
-    {
+    if (!a.getMetadataOptions().equals(b.getMetadataOptions())) {
       return false;
     }
 
--- a/components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java
+++ b/components/formats-api/src/loci/formats/in/DefaultMetadataOptions.java
@@ -80,4 +80,23 @@ public class DefaultMetadataOptions implements MetadataOptions {
     validate = validateMetadata;
   }
 
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (other == null || getClass() != other.getClass()) {
+      return false;
+    }
+    DefaultMetadataOptions options = (DefaultMetadataOptions) other;
+    return metadataLevel == options.metadataLevel &&
+      validate == options.validate;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = metadataLevel == null ? 0 : metadataLevel.hashCode();
+    return 31 * result + (validate ? 1 : 0);
+  }
+
 }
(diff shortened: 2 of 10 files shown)


Darlokt:
b91d1483811d5fa5cccf663b58dae3c81864c671 adds documentation and improved best-effort companion file checks as well as improved testing coverage.

Darlokt:
As in https://github.com/ome/bioformats/pull/4463 this now properly uses the closeable.

melissalinkert:
We are not able to accept these changes at this time. Any changes to memoization have serious implications for OMERO and other tools, and require a great deal of care during implementation, review, and testing. The extensive set of changes here, including changes to the memo file format itself, are not in line with our current priorities. They also conflict with ongoing work on [#4254](https://github.com/ome/bioformats/pull/4254).

We further note that this is one of several large-scale changes made without prior communication with the OME team. We strongly encourage you to start a discussion on a relevant issue prior to making such changes.

PR #4360, opened 2025-08-27
Fix several OIR issues

melissalinkert:
Fixes #4132, #3808, and #3748.

Without this change:

- `showinf` on QA 40683 should throw an exception (#4132)
- `showinf` on `imagesc-65533/Wrong*` should detect a single timepoint (#3808)
- `showinf -nopix -omexml` on pretty much any public .oir data should show a `Laser` set in the OME-XML, but no `Power` attribute set

With this change:

- QA 40683 should initialize and display 3 channels
- all files in `imagesc-65533` should have 10Z x 15T
- `Power` should be set on `Laser`

Units for laser power are assumed to be milliwatt, but don't appear to be explicitly specified in the file, so open to other opinions.

Initialization times and test run times for the `olympus-oir` directory should be checked with 47ae817 in particular - this didn't seem to impact have much impact locally, but would be good to verify that it doesn't increase times on the test machine.

Issue #3808, opened 2022-04-12
Olympus OIR: Undetected PixelBlocks/Incorrect dimensions


dgault:
This issue was raised in the forum thread https://forum.image.sc/t/wrong-dimensions-for-some-images-when-reading-olympus-oir-files/65533

Sample files have been provided which reproduce the issue with the latest Bio-Formats (6.9.0)

The sample files are showing incorrect dimensions, missing the T dimension. The size should be 10Z and 15T but instead is detected as 10Z and 1T. From debugging one of the sample files the number of pixel blocks detected by the reader only accounts for the first 10 frames and the remainder have not been parsed. Looking into the file itself and the uid's for the undetected blocks are present in the file as expected.

Requires further investigation as to why the pixel blocks for the T dimension are not picked up.

dgault:
sample file is now located at `scratch/inbox/imagesc-65533`

Issue #3748, opened 2021-11-24
OIR: Laser metadata not being populated


dgault:
Currently a Laser element is created with an ID in the OME-XML, however additional metadata values which are parsed are simply included in the global metadata.

The reader is already parsing values for ID, data ID, power, transmissivity and wavelength. These values could be populated on the laser element.

melissalinkert:
As discussed in #4360, it's clear that laser power metadata is present in .oir files, but it's unclear how this should be mapped into the OME instrument model. We would need guidance on what the stored laser power value actually represents and how it's used in order to proceed.

Closing as not planned, but can re-evaluate if we receive definitive information on how to handle this metadata. See also https://github.com/ome/bioformats/pull/4360#issuecomment-3327307855.

PR #4465, opened 2026-08-04
OIR: fix two parse failures on Olympus LEXT (OLS) acquisitions

### LINKED ISSUE #4132: Olympus OIR: MetadataTools.populatePixelsOnly: 0 must be non-null and strictly positive.
See [QA 40683](https://www.openmicroscopy.org/qa2/qa/feedback/40683/) 

```
java.lang.IllegalArgumentException: 0 must be non-null and strictly positive.
	at ome.xml.model.primitives.PositiveInteger.(PositiveInteger.java:48)
	at loci.formats.MetadataTools.populatePixelsOnly(MetadataTools.java:310)
	at loci.formats.MetadataTools.populateMetadata(MetadataTools.java:262)
...
```

Note: QA used older Insight/Bioformats. But I can replicate the issue with the latest build too:
```
(bf) ➜  bftools ./showinf -version
Version: 7.1.0
Build date: 11 December 2023
VCS revision: 05c7b2413cfad19a73b619c61ddf77ca2d038ce7
(bf) ➜  bftools ./showinf -nopix ~/40683/Image_10.oir
Checking file format [Olympus OIR]
Initializing reader
OIRReader initializing /Users/dom/40683/Image_10.oir
Reading pixel blocks from /Users/dom/40683/Image_10.oir
Exception in thread "main" java.lang.IllegalArgumentException: 0 must be non-null and strictly positive.
	at ome.xml.model.primitives.PositiveInteger.<init>(PositiveInteger.java:48)
	at loci.formats.MetadataTools.populatePixelsOnly(MetadataTools.java:310)
	at loci.formats.MetadataTools.populateMetadata(MetadataTools.java:262)
```

PEEKPerformer:
Refs: [Opening POIR Olympus files](https://forum.image.sc/t/opening-poir-olympus-files/26823) on image.sc.

`OIRReader` cannot open `.oir` files produced by Olympus/Evident LEXT (OLS) instruments. Nothing structural is wrong: the reader already locates the pixel blocks, resolves the channels and reads the metadata correctly once it gets past two `NumberFormatException`/`FormatException` sites during initialisation.

Neither case is LEXT-specific in principle. Both are places where a field that is present, or scaled a particular way, in every FV-series file behaves differently elsewhere.

No open issue tracks either of these. #4132, #3808 and #3748 were the recent OIR reports and all were closed by #4360.

### 1. Missing `order` attribute in `parseImageProperties`

Introduced in bc3a9492e4 (2017), "Use `order` attribute to set channel ordering":

```java
int index = Integer.parseInt(channelNode.getAttribute("order")) - 1;
```

`Element.getAttribute` returns `""`, not `null`, for an absent attribute, so this throws `NumberFormatException` out of `setId` and the file cannot be opened at all.

LEXT files omit it: the `commonphase:channel` elements under `commonimage:imageInfo` carry only an `id`, and the ordering lives on the `lsmimage:LSMChannel` elements elsewhere in the document. Those ids appear under `imageInfo` in the same sequence as their recorded order, so document order is the correct index when the attribute is missing.

For comparison, in the two public sample sets under `downloads.openmicroscopy.org/images/Olympus-OIR/` the same elements read `order="2"` and `order="3"`, so the new branch is unreachable for them.

### 2. RGB depth divided below one byte, in `parseFrameProperties` and `parseImageProperties`

This one extends recent work. 76eb916f5c ("OIR: support data with component channels", #4360, fixes #4132) added component channel handling, which splits an RGB channel into one OME channel per component and divides the recorded depth by three:

```java
int bytes = Integer.parseInt(depth.getTextContent());
if (rgb) {
  bytes /= 3;
}
m.pixelType = FormatTools.pixelTypeFromBytes(bytes, false, false);
```

That is right where `depth` is the total across all three components. The LEXT
camera image is a genuine component channel case, one `commonphase:channel`
with `order="1"` and three `ElementChannelInfo` children, but it records depth
**per component**, so the division yields 0:

```
FormatException: Unsupported byte depth: 0
```

Correcting something I said earlier in this description: I had claimed the file was self-inconsistent here. It is not. The `commonframe:imageDefinition` for the 1024x1024 image is flat and unambiguous, `colorType RGB` with `depth 1` and `bitCounts 8`. The `depth 3` / `bitCounts 24` I was comparing it against belongs to a different, 1024x755 image in the same file. So the file uses the per-component convention for one image and the total convention for another, and divisibility tells them apart correctly.

A total across three components is always a multiple of three, so the fix only divides when the value divides evenly, with the same test applied to `bitCounts`. A `depth` that is not a multiple of three cannot be a total, so this also stops `bitCounts 8` on an RGB component being reported as 2 valid bits per pixel.

This is a heuristic. Since it sits on top of very recent work, I would rather defer on it: if there is a better way to tell the two conventions apart (component count, `bitCounts`, an explicit variant check) I am happy to rework this half.

One limit of the divisibility test worth recording. It is sound for byte depths, since the valid per-component values are 1, 2 and 4 and none of those divide by three. It is weaker for `bitCounts`, where 12 is both a plausible per-component depth and a multiple of three, so a 12-bit-per-component RGB file recorded per-component would be halved to 4. I have no such file to test against, so I have not tried to special-case it.

## Code

+15 -5 in `OIRReader.java`, three hunks, split across two commits so each bug can be reviewed and reverted independently. Both commits build on their own.

## Validation

Against the README PR testing checklist:

- This branch merges cleanly into the current `develop` (`git merge-tree`, 0 conflicts).
- `ant clean jars tools`: BUILD SUCCESSFUL.
- Maven build: `mvn -DskipTests install` completes successfully.
- `ant test`: 1897 tests, 0 failures, 0 skips, 5m01s.
- No syntax or API beyond Java 11 (plain `if`/`String.trim`/`isEmpty`).
- `showinf` on every affected format, below.

### Regression against the public corpus

`showinf -nopix -novalid` output on both public Olympus-OIR sample sets is
**byte-identical before and after** this branch (1343 and 857 lines
respectively, `diff` clean):

| file | result |
|---|---|
| `imagesc-105684/1202-interval_10sec_sequence_frame.oir` | identical to stock `develop` |
| `etienne/amy slice z stack_0001.oir` | identical to stock `develop` |

This is the expected outcome given the analysis above: fix 1's fallback is unreachable when `order` is present, and fix 2 only alters inputs that currently raise `FormatException`.

### LEXT files, previously unopenable

These come from a single OLS5500 session in our lab on 2026-08-03. Both `.poir` files below are deposited under CC-BY 4.0, see the note at the end, so this is reproducible once the deposit clears review. Each row is a distinct case rather than a distinct dataset: two of the three come out of the same `.poir`.

| inner `.oir` | from | size | before | after |
|---|---|---|---|---|
| `260803_1437_000_LSM3D^3D_LSM.oir` | `001.poir` | 12.5 MB | `NumberFormatException` | 256×256, SizeC=3, uint16 |
| `260803_1448_000_LSM3D^3D_LSM.oir` | `002.poir` | 15.4 MB | `NumberFormatException` | 1024×1024, SizeC=3, uint16 |
| `260803_1448_000_COLOR3D^XY_Camera.oir` | `002.poir` | 12.3 MB | `Unsupported byte depth: 0` | 1024×1024, SizeC=3, uint8 |

`LSM3D` is the laser confocal result, three channels of colour, height and intensity for a single plane. `COLOR3D` is the companion colour camera image, and is the component channel case discussed above. Every `.poir` here except `001.poir` contains one of each.

Metadata now resolves as expected, e.g.

```
<Objective ID="Objective:0:0" LensNA="0.3" NominalMagnification="10.0"
           WorkingDistance="10.4" WorkingDistanceUnit="mm"/>
<Pixels PhysicalSizeX="5.0" PhysicalSizeXUnit="µm"
        PhysicalSizeY="5.0" PhysicalSizeYUnit="µm" .../>
```

5 µm/px at 10× over 256 px is a 1.28 mm field, which matches the objective.

End-to-end, `bfconvert` produced OME-TIFF for **79 inner `.oir` files** with 0 failures: 9 from five single acquisitions, and 70 from one 35-field mosaic. All from the same instrument and session, so this is depth of coverage on one LEXT variant rather than breadth across the family.

One thing that looks wrong but is not: `showinf` reports `Valid bits per pixel = 7` on one LSM3D channel. The file really does record `bitCounts 7` there, and that channel's pixel values top out at 90.

## Notes

The instrument here is an OLS5500; files came off it as `.poir` and `.mpoir`. Those are Zip64 wrappers. `.poir` contains one or more `.oir`, and `.mpoir` contains `matl.omp2info` plus one `.poir` per field (entries stored, not deflated). **This PR does not add readers for either wrapper**; it only makes the inner `.oir` readable once extracted, which is what the image.sc thread ran aground on six years ago.

If a `.poir`/`.mpoir` reader would be welcome I'm happy to work on it, but per the external-readers guidance I'd rather ask first whether you'd want it in core or as an external reader. Happy to go either way.

I don't have access to the curated data repo, so `ant test-automated` is needed.

Five `.poir` files from this session, including the two used above, are deposited under CC-BY 4.0 as [10.5281/zenodo.21792018](https://doi.org/10.5281/zenodo.21792018) and submitted to the [Bio-Formats Zenodo community](https://zenodo.org/communities/bio-formats), where they are awaiting approval.

Between them they cover both parse paths this PR touches, so the change should be testable against the curated corpus if you want to add them. Polymer composite surface metrology, no biological content. Happy to deposit the `.mpoir` mosaic as well if that case is useful, though it is 323 MiB and exercises nothing the single acquisitions do not.

One sibling worth noting, not changed here: `parseImageProperties` also does an unguarded `Integer.parseInt(component.getAttribute("order"))` on `elementChannel` components, added by the same commit as case 2. The LEXT files do populate `order` there, so it is not reachable for this data, but it is the same pattern as case 1.

### DIFF
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -874,14 +874,18 @@ public class OIRReader extends FormatReader {
       }
       if (depth != null) {
         int bytes = Integer.parseInt(depth.getTextContent());
-        if (rgb) {
+        // 'depth' is normally the total byte count across all components, but
+        // some RGB frames record the per-component count instead. A total
+        // across three components is always a multiple of three, so only
+        // divide when it divides evenly.
+        if (rgb && bytes % 3 == 0) {
           bytes /= 3;
         }
         m.pixelType = FormatTools.pixelTypeFromBytes(bytes, false, false);
       }
       if (bitCount != null) {
         m.bitsPerPixel = Integer.parseInt(bitCount.getTextContent());
-        if (rgb) {
+        if (rgb && m.bitsPerPixel % 3 == 0) {
           m.bitsPerPixel /= 3;
         }
       }
@@ -1000,14 +1004,15 @@ public class OIRReader extends FormatReader {
 
           if (depth != null) {
             int bytes = Integer.parseInt(depth.getTextContent());
-            if (rgb) {
+            // see the equivalent block in parseFrameProperties
+            if (rgb && bytes % 3 == 0) {
               bytes /= 3;
             }
             m.pixelType = FormatTools.pixelTypeFromBytes(bytes, false, false);
           }
           if (bitCount != null) {
             m.bitsPerPixel = Integer.parseInt(bitCount.getTextContent());
-            if (rgb) {
+            if (rgb && m.bitsPerPixel % 3 == 0) {
               m.bitsPerPixel /= 3;
             }
           }
--- a/components/formats-gpl/src/loci/formats/in/OIRReader.java
+++ b/components/formats-gpl/src/loci/formats/in/OIRReader.java
@@ -951,7 +951,13 @@ public class OIRReader extends FormatReader {
         Element channelNode = (Element) channelNodes.item(i);
 
         c.id = channelNode.getAttribute("id");
-        int index = Integer.parseInt(channelNode.getAttribute("order")) - 1;
+        // LEXT (OLS) variants omit the 'order' attribute on the channel
+        // elements under commonimage:imageInfo; fall back to document order.
+        String order = channelNode.getAttribute("order");
+        int index = i;
+        if (order != null && !order.trim().isEmpty()) {
+          index = Integer.parseInt(order.trim()) - 1;
+        }
 
         Element name = getFirstChild(channelNode, "commonphase:name");
         if (name != null) {


components/formats-gpl/src/loci/formats/in/OIRReader.java
```diff
@@ -874,14 +874,17 @@ private void parseFrameProperties(Element root) throws FormatException {
       }
       if (depth != null) {
         int bytes = Integer.parseInt(depth.getTextContent());
-        if (rgb) {
+        // 'depth' is normally the total byte count across all components, but
+        // some RGB frames record the per-component count instead; only divide
+        // when the result would still be a valid byte depth.
+        if (rgb && bytes >= 3) {
```
melissalinkert: Here and in the other places where a check for `>= 3` is introduced, would it make more sense to check for something like `bytes % 3 == 0`?

PEEKPerformer: Yes, that is much better, thanks Melissa! I changed at all four sites in 503f590. Appreciate your speedy feedback!

melissalinkert:
> If a `.poir`/`.mpoir` reader would be welcome I'm happy to work on it, but per the external-readers guidance I'd rather ask first whether you'd want it in core or as an external reader. Happy to go either way.

Are you able to provide a wider range of `.poir`/`.mpoir` data that covers a range of dimensions (channels, time, Z) and modalities? Is `.poir`/`.mpoir` the default for OLS5500 acquisitions, or is this an option that needs to be enabled?

PEEKPerformer:
Yeah, the instrument manual describes .poir as the 'OLS5000/OLS5100/OLS5500 standard file format' and the save dialogs default to it. Stitching acquisitions produce .mpoir so it's what these instruments emit unless SDF/JPEG/BMP/PNG is manually exported by the user.

On breadth, more than happy to spend more time on the instrument. I'll check which modes our instrument supports and will deposit as much variation as I can. Happy to try and fill special requests if you have any!

Edit: will ping you when i get the files :)

PEEKPerformer:
@melissalinkert submitted, pending approval. 10.5281/zenodo.22017632

I found some files that trigger the bug from the description above. Four snapshot tiles in the set hit the unguarded elementChannel order parseInt (no order attribute on their ElementChannelInfo). A near-identical passing file is included. Happy to fix it like the imageInfo site, here or as a follow-up. Just let me know :)

Would be happy to take a stab at the reader as well with your blessing.

Nothing has a SizeT > 1 because the instrument lacks a time series mode.

PR #4468, opened 2026-08-05
Add `omitIniFromUsedFiles` option to `FakeReader`

melissalinkert:
This makes it easy to create a fake dataset that will fail to initialize or have inconsistent dimensions after importing to OMERO.

Options provided in a .fake.ini file override the options provided in the matching .fake file name. Simple example:

```
$ cat "test&sizeX=0.fake"
$ cat "test&sizeX=0.fake.ini"
sizeX=1024
omitIniFromUsedFiles=true
$ showinf -no-upgrade -nopix test\&sizeX\=0.fake
Checking file format [Simulated data]
Initializing reader
FakeReader initializing test&sizeX=0.fake
Initialization took 0.198s

Reading core metadata
filename = test&sizeX=0.fake
Used files = [/home/melissa/bad-test/test&sizeX=0.fake]
Series count = 1
Series #0 :
	Image count = 1
	RGB = false (1) 
	Interleaved = false
	Indexed = false (true color)
	Width = 1024
	Height = 512
	SizeZ = 1
	SizeT = 1
	SizeC = 1
	Tile size = 1024 x 512
	Thumbnail size = 128 x 64
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata

Reading metadata
$ rm "test&sizeX=0.fake.ini"
$ showinf -no-upgrade -nopix test\&sizeX\=0.fake
...
loci.formats.FormatException: Invalid sizeX: 0
	at loci.formats.in.FakeReader.initFile(FakeReader.java:874)
```

The idea is that OMERO import should succeed with a dataset like this, but only the .fake file would be put under ManagedRepository. Attempting to initialize the .fake file after import would then throw an exception. This can also be used to get data which would initialize after import, but have different dimensions, pixel type, etc.

Discussed with @knabar in the context of https://github.com/ome/omero-web/issues/657 that something like this might be useful. I was not successful in finding a way to do something similar with other readers, so this is an alternative.

### DIFF
--- a/components/formats-bsd/src/loci/formats/in/FakeReader.java
+++ b/components/formats-bsd/src/loci/formats/in/FakeReader.java
@@ -230,6 +230,7 @@ public class FakeReader extends FormatReader {
   private int sleepOpenBytes = 0;
   private int sleepInitFile = 0;
   private boolean labelPlanes = false;
+  private boolean omitIniFromUsedFiles = false;
 
   static void sleep(String msg, int ms) {
     if (ms <= 0) return; // EARLY EXIT
@@ -524,8 +525,12 @@ public class FakeReader extends FormatReader {
       FormatTools.assertId(currentId, true, 1);
       List<String> files = new ArrayList<String>();
       fakeSeries.clear();
-      if (!noPixels) files.addAll(listFakeSeries(currentId));
-      if (iniFile != null) files.add(iniFile);
+      if (!noPixels) {
+        files.addAll(listFakeSeries(currentId));
+      }
+      if (iniFile != null && !omitIniFromUsedFiles) {
+        files.add(iniFile);
+      }
       return files.toArray(new String[files.size()]);
   }
 
@@ -857,6 +862,8 @@ public class FakeReader extends FormatReader {
         sleepInitFile = intValue;
       } else if (key.equals("labelPlanes")) {
         labelPlanes = boolValue;
+      } else if (key.equals("omitIniFromUsedFiles")) {
+        omitIniFromUsedFiles = boolValue;
       }
     }
 


joshmoore:
Love interesting new flags in FakeReader :heart:

sbesson (commented):
With the current HEAD of develop
```
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats % echo "omitIniFromUsedFiles=true" > test.fake.ini
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats % ./tools/showinf -nopix test.fake
Initializing reader
Initialization took 0.043s

Reading core metadata
filename = test.fake
Used files:
	/Users/sbesson/Documents/GitHub/bioformats/test.fake
	/Users/sbesson/Documents/GitHub/bioformats/test.fake.ini
Series count = 1
Series #0 :
	Image count = 1
	RGB = false (1) 
	Interleaved = false
	Indexed = false (true color)
	Width = 512
	Height = 512
	SizeZ = 1
	SizeT = 1
	SizeC = 1
	Tile size = 512 x 512
	Thumbnail size = 128 x 128
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata

Reading metadata
```

With this PR included

```
sbesson@Sebastien-GS-MacBook-Pro-2025 bioformats % ./tools/showinf -nopix test.fake
Initializing reader
Initialization took 0.034s

Reading core metadata
filename = test.fake
Used files = [/Users/sbesson/Documents/GitHub/bioformats/test.fake]
Series count = 1
Series #0 :
	Image count = 1
	RGB = false (1) 
	Interleaved = false
	Indexed = false (true color)
	Width = 512
	Height = 512
	SizeZ = 1
	SizeT = 1
	SizeC = 1
	Tile size = 512 x 512
	Thumbnail size = 128 x 128
	Endianness = intel (little)
	Pixel type = uint8
	Valid bits per pixel = 8
	Metadata complete = true
	Thumbnail series = false
	-----
	Plane #0 <=> Z 0, C 0, T 0


Reading global metadata

Reading metadata
```

So definitely works as described. I am unclear how this is intended to be used for creating test OMERO scenario with mismatching dimensions. As the contract of the importer is to upload the return value of `getUsedFiles`, a synthetic dataset with `omitIniFromUsedFiles=true` would only upload a single `.fake` file and use it for creating the metadata in the database.

melissalinkert:
> So definitely works as described. I am unclear how this is intended to be used for creating test OMERO scenario with mismatching dimensions. As the contract of the importer is to upload the return value of `getUsedFiles`, a synthetic dataset with `omitIniFromUsedFiles=true` would only upload a single `.fake` file and use it for creating the metadata in the database.

I guess I was working under the assumption that the original initialization (from .fake + .fake.ini that had not been transferred yet) was used to set metadata. If that assumption is incorrect (which makes sense on further reflection), then I really don't know how we'd accomplish the original goal of creating data that imports correctly, but fails to initialize and/or has conflicting metadata when trying to view post-import.

I am OK with closing this, or at least excluding from rc1 consideration.

PR #4482, opened 2026-08-24
Remove unused and unpackaged `TJExample` and `TJUnitTest`

### LINKED ISSUE #4480: turbojpeg: TJExample and TJUnitTest are in the default package
`components/forks/turbojpeg/src/TJExample.java` and `TJUnitTest.java` have no `package` declaration, so `ome:turbojpeg` ships two classes in the default package:

```
$ unzip -l turbojpeg-8.5.0.jar | grep -E '^[A-Za-z0-9_$]+\.class$'
TJExample.class
TJUnitTest.class
```

That makes the jar unusable as an automatic module, so any downstream build that puts it on the module path fails. Concretely, maven-javadoc-plugin 3.11.2 running on a Java 9+ JDK aborts with:

> Unable to derive module descriptor for .../ome/turbojpeg/8.5.0/turbojpeg-8.5.0.jar: TJExample.class found in top-level directory (unnamed package not allowed in module)

Could the two classes be moved into a package (e.g. `org.libjpegturbo.turbojpeg`), or simply be excluded from the published jar since they are only an example and a test driver?

Workaround downstream, for reference: `-DlegacyMode=true` for the javadoc plugin.

melissalinkert:
Fixes #4480.

I'm not entirely opposed to moving these to a package or updating the pom to exclude instead, but I also don't think there is a good reason to keep these classes. Since they are not in a package, I'd expect that no one is extending either of them, and they are not used internally either. Happy to discuss further though if needed.


sbesson (approved):
Agreed with removing these files which are only examples of using the Java bindings of libjpeg-turbo and are outside the package. Waiting for a successful execution of the nightly repository tests

sbesson:
Merging as per https://bf-testing-results.s3.amazonaws.com/2026/2026-08-26/build-status.log

PR #4486, opened 2026-09-17
Upgrade check cleanup

### LINKED ISSUE #3390: UpgradeChecker: proposed deprecations
https://forum.image.sc/t/bioformat-upgrade-matlab/26099 revealed a few issues with the `UpgradeChecker` installation logic affecting callers like `bfUpgradeCheck`.

The most prominent issue is that the CI logic  within the class is no longer functional. Having the library auto-installing CI artifacts is probably no longer a requirement:

https://github.com/openmicroscopy/bioformats/blob/a28a7cb6e7aa07862dc5a788720ff8032a51fd2e/components/formats-bsd/src/loci/formats/UpgradeChecker.java#L69-L82

For Bio-Formats 6.2.0, a proposal would be to:
- deprecate `TRUNK_BUILD`, `DAILY_BUILD` and all the currently broken codepaths using the CI logic to retrieve the latest stable library instead
- review the `STABLE_BUILD`and `STABLE_VERSION` logic. It feels like we should be able to use the latest URL and there is no longer a need for encoding the `STABLE_VERSION` directly in the class.
- review all consumer code including ImageJ and Bio-Formats

melissalinkert:
Fixes #3390.

This deprecates several fields as proposed in #3390, and removes several long-deprecated fields that are no longer maintained. The ImageJ plugin updater and `UpgradeCheck` command-line tool have been updated accordingly, so that only the current release is checked/downloaded as applicable.


components/formats-bsd/src/loci/formats/UpgradeChecker.java
```diff
@@ -63,23 +63,25 @@ public class UpgradeChecker {
...
   /**
-   * @deprecated  As of release 6.6.0
+   * Location of the OME continuous integration server.
+   *
+   * @deprecated since 9.0.0
    */
-  public static final String STABLE_VERSION = "6.6.0";
-
-  /** Location of the OME continuous integration server. */
   public static final String CI_SERVER = "http://ci.openmicroscopy.org";
```
sbesson, line 72: These deprecated fields could be annotated with `@Deprecated` to communicate the expectations and silence the warnings

sbesson (changes requested):
Once the Javadoc markup for the deprecated fields is supplemented with the corresponding Java annotation, this should be ready to do.

components/formats-bsd/src/loci/formats/UpgradeChecker.java
melissalinkert: Done in f760334.

sbesson (approved):

